PR #107, opened 2013-01-27
Experimental implementation of a modified ESA model.

piskvorky:


oroszgy:
Hi @piskvorky, I might be interested in finishing this PR. Would it be still useful for the project?

piskvorky:
@oroszgy I don't know about this PR, but the feature itself is still interesting. It's a nice simple technique :)

PR #175, opened 2014-03-02
[Word2Vec] Support for reading calculated skip ngrams

josephcc:
This adds support for training from precompiled skip ngrams, instead of raw text. Each ``sentence" from the corpus reader now consists of the following format ([str:word1, str:word2], int:distance, int:count). This is useful if you are trying to train a word2vec model from a ngram model, for example the Google Web1T dataset. Both the Python and Cython implementations were updated.

New functionalities are controlled by the skip_ngram parameter, with the default value the behavior of existing functionalities should remain the same.

I wasn't sure how to change the matrix manipulation part, so each skip ngrams with counts will update the matrix multiple times. This is likely to be very inefficient, and should be improved in the future.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -228,6 +228,7 @@ class Word2Vec(utils.SaveLoad):
         sentence_no, vocab = -1, {}
         total_words = 0
         for sentence_no, sentence in enumerate(sentences):
+            count = 1
             if skip_n_gram:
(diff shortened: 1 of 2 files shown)


tmylk:
@josephcc Would you have time to work on this over next 2 months? If not, suggest closing as abandoned PR

tmylk:
Closing as abandoned.

PR #221, opened 2014-07-16
changed preprocessing step in tut1.rst from a set build with a count() c...

brummell:
...all in the comprehension to a slightly more verbose dict build -> match as the former methods runtime did not scale to even relatively large data sets. This hopes to prevent copying/pasting distress in users less familiar with python.


tmylk:
@piskvorky Is this still relevant?

piskvorky:
This particular tutorial example was fixed in https://github.com/piskvorky/gensim/pull/279, which subsumed this PR.

It's already "live": http://radimrehurek.com/gensim/tut1.html#from-strings-to-vectors

Thanks @brummell !

PR #233, opened 2014-09-11
Static compile of word2vec

piskvorky:
Avoid weird problems with dynamic `pyximport` compilation.

Instead, bundle the cython-generated C file and try to compile that, at gensim install time.

If installation fails, show a warning but install gensim anyway (=the compilation is optional).


boyangli:
Could anyone tell me what's the status of this issue? I am still getting that "Cython compilation failed, training will be slow." I tried all kinds of python configurations. Can't get any to work.

piskvorky:
@boyangli: this issue was fixed about a year ago. Your troubles must be unrelated.

To know more, try posting your gensim installation log (to its own issue, or to the gensim mailing list, not here). From the log you should be able to see what the problem was = why the compilation failed.

boyangli:
Thank you, @piskvorky . I switched to scipy 0.15.1, and the problem seems resolved. It must be a compatibility issue with 0.16. I will post a most detailed description as a separate issue.

gojomo:
@boyangil – sounds like known issue #382, patched in #415.

boyangli:
@gojomo If it's been fixed, I'm still seeing it, even after downloading the source code from github directly.

boyangli:
A new issue has been filed https://github.com/piskvorky/gensim/issues/429

PR #248, opened 2014-10-18
Add compressing secondary files

cscorley:
The utils.smart_extension function is used to add new extensions to
a filename, but to do it smart when concerning compressed filenames.

For example, when a LdaModel is saved and also needs to save the state,
use smart_extension to re-order the extensions so that the state is also
saved as a compressed file if the model is being saved as compressed.


piskvorky:
Going through open pull requests, not sure why this wasn't merged. Looks good to me @cscorley !

To finish this up:

Is this backward compatible? Will existing code continue to work without modifications?
Under what conditions will NEW `load` of objects stored with OLD `save` break? 
Under what conditions will OLD `load` of objects stored with NEW `save` break? 
Do you think this warrants some explicit doc mentions / examples, or should we treat this functionality as completely opaque?

cscorley:
I think that most code should work without modification. However, things like LdaModel default to loading with `mmap='r'`, and that would be something to change. The default behavior would fail to load files saved as compressed. So, updating the default behavior in the load methods to not use mmap is something still to-do and would definitely cause a behavior change.

- New load would fail on old saves because the files are saved to different locations (previously always as npy, now sometimes as npz if compression is used) -- it may be possible to 'fallback' to swapping, but I don't think this is safe in case the fallback file is for a different model. Could cause unexpected consequences.
- Old load would fail on new saves for the same reason. However, this case should be much less likely than the previous :)

Mentioning the change in behavior would definitely need to change in the documentation.

piskvorky:
Could you explain a bit more about the "new load on old save" combo?

What happens if I have an old model saved, and try to load it with the new code? Does it fail outright? Or only in some cases?

cscorley:
The old saved npy files will load fine *if* the model wasn't saved into a compressed file (or has then since uncompressed). It will also load fine in either case where separate npy files weren't saved. However, it will throw an exception in the case that the model is saved as compressed and had the NumPy arrays saved separately.

Essentially, it expects whenever a model is saved as .bz2 or .gz. that the separate NumPy files are saved compressed in NumPy's npz format. Otherwise, if the model is not .bz2 or .gz, it expects the uncompressed npy format.

Additionally, NumPy doesn't care if it is saved compressed or not. The exception is raised because NumPy will *not* load a compressed file with mmap, as discussed earlier.

piskvorky:
I see, thanks.

I think this is fine, not that many cases fall under this "error scenario". We'll put a big fat warning during release that any saved models should be updated, just in case.

Merging!

piskvorky:
Oh, cannot merge, conflicts... can you sort that out?

cscorley:
Yup, I'll update the PR and merge it tonight. Thanks!

PR #277, opened 2015-01-08
add some test codes for DtmModel

hyhan:
I have added some test codes for pull request #272. All the test cases are ok.


piskvorky:
Thanks, that great!

Travis CI complains that two DTM tests fail though -- can you have a look?

hyhan:
Yes, I have checked it. First, there is a problem when de-serializing the test corpus. I simply use the MmCorpus format instead. Second, Travis CI doesn't use the dtm binary file currently. So the related test cases will be skipped if the environmental variable ```$DTM_PATH``` is not properly set up to specify the location of the executable file.

piskvorky:
OK. Cheers, merging.

PR #279, opened 2015-01-14
use faster and more idiomatic frequency counting in tutorial

larsmans:
Uses a `defaultdict` instead of linear scanning through documents to do frequency counting. This takes the complexity down from quadratic to linear, which is important if inexperienced programmers blindly copy-paste this into their own program, and it's more obvious (IMHO) then the `sum` hack used previously.


piskvorky:
Yeah, I've seen the code copy&pasted way too many times, despite the warning.

Merging, thanks.

PR #280, opened 2015-01-15
fix TypeError in LdaMulticore model

chyikwei:
Hi,
Just find a minor issue in `LdaMulticore` model.
When I set parameter `eval_every` to `None`, it will raise TypeError.
So I send this pull request to fix it.


piskvorky:
Thanks!

PR #281, opened 2015-01-15
speed up doc2bow by ~40%

larsmans:
I found `doc2bow` taking a rather large amount of time. This patch speeds it up by ca. 40% on my data:

* use hash tables, not sorting + grouping
* loops in conditional blocks, not conditional blocks in loops
* cache attribute/method lookups
* inline `to_unicode`

Surprisingly, the final optimization was the most important one. Function call overhead in Python is horrible.

### DIFF
--- a/gensim/corpora/dictionary.py
+++ b/gensim/corpora/dictionary.py
@@ -17,7 +17,7 @@ with other dictionary (:func:`Dictionary.merge_with`) etc.
 
 from __future__ import with_statement
 
-from collections import Mapping
+from collections import Mapping, defaultdict
 import sys
 import logging
 import itertools
@@ -124,7 +124,7 @@ class Dictionary(utils.SaveLoad, Mapping):
                 logger.info("adding document #%i to %s", docno, self)
 
             # update Dictionary with the document
-            _ = self.doc2bow(document, allow_update=True) # ignore the result, here we only care about updating token ids
+            self.doc2bow(document, allow_update=True) # ignore the result, here we only care about updating token ids
 
         logger.info("built %s from %i documents (total %i corpus positions)",
                      self, self.num_docs, self.num_pos)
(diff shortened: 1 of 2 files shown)


piskvorky:
Thanks Lars. Tests failing on all Pythons, can you fix?

Out of curiosity -- under what circumstance is doc2bow a bottleneck? What is the use case?

I'm mild -1 on inlining `to_unicode`. I'd like to keep this functionality in one place, in case I change the default params / add more logic in the future.

But if your scenario is a common one, and the speed up is big (how much from the 40%?), I'll live with it :)

piskvorky:
Ping @larsmans -- let me know if you plan to fix the tests. I'd like to make a new release this month and this would be a great addition.

piskvorky:
ping @larsmans

cscorley:
Hi @piskvorky, I'd also like to see this get in. I can look into fixing the tests by next release if needed. Looks like it's mostly a generator issue, and possibly with how the returning bow is sorted.

piskvorky:
Hmm, maybe. The generator error seems simple (make a list out of the input document).

But the different bow order is unpleasant. Not just because of all the tests that need to be changed, but also documentation. Lars mentions the "most important optimization" was just inlining `to_unicode`.

So I wonder if we can avoid all the trouble and just do that one thing, for most of the performance gain...

piskvorky:
@cscorley @larsmans I plan to release next week. Please let me know if you plan to finish this and merge.

cscorley:
Ok, I've fixed the tests (locally). Changes are on branch [pr/281](https://github.com/piskvorky/gensim/tree/pr/281).

Times for interest:
- Master branch: Ran 9 tests in 39.084s
- PR: Ran 9 tests in 12.120s
- PR without inline to_unicode: Ran 9 tests in 34.875s

Many of the failing tests (around the models and similarities) were expecting the Dictionary to assign ids in a certain order. Fixed these tests by hand (only a couple values swapped order). 

The last failing one is an LSI test, @piskvorky can you take a look at those test values? I suspect it is due to the same reason (new dictionary order).

larsmans:
I'm afraid I don't have time to help, so I'll leave this to @cscorley.

gojomo:
Rather than inlining `to_unicode` up, maybe most of the benefit would also be achieved by making a variant of `to_unicode` that takes a list/iterator, and does the tight loop itself? 

(Also: perhaps the biggest speedup is in cases where it's a no-op, because the tokens are already unicode. I'm guessing in most batches, either all the tokens are already unicode, or none, as they'll all have arrived via the same encoding path. So a batch to_unicode could possibly even take the risk of returning what it was passed, when the first token passes `isinstance(text, unicode)`.)

larsmans:
That's not reliable since mixed lists do occur in practice. In particular, LXML produces them (probably as an optimization for Py2):

```py
>>> from lxml import etree
>>> etree.fromstring(u'<x><t>foo</t><t>bàr</t></x>').xpath('//text()')
['foo', u'b\xe0r']
```

gojomo:
That's too bad. Still, pushing the loop into a utility method may help.

piskvorky:
Great, thanks Chris. I've pushed a fix for the last failing unittest.

I'm going to close this and open PR for `pr_281` instead.

PR #283, opened 2015-01-18
Fix docstring for WikiCorpus

jbochi:
After 16h trying to run the commands in the docstring, I got a message saying that 

    AttributeError: 'WikiCorpus' object has no attribute 'saveAsText'

The method does not exist anymore.


piskvorky:
Aah, sorry about that! The docstring must have slipped through the cracks during refactoring.

Thanks a lot, merging.

PR #284, opened 2015-01-20
Feature/sharded corpus

hajicj:
Hello Radim,

as we talked about briefly after the Machine Learning Meetup in Prague, I wrote an IndexedCorpus that serializes (pretty) fast and supports efficient batch retrieval as numpy arrays/scipy sparse matrices. 

The ShardedCorpus class is in gensim/corpora/sharded_corpus.py. Unit tests are in a separate file (gensim/tests/test_sharded_corpus.py) and I added two functions to utils for creating mock gensim-style data.

I also have a very superfluous benchmarking script for serialization and retrieval that I wrote just to be sure that ShardedCorpus is reasonably faster than MmCorpus and other IndexedCorpus implementations. Would you be interested in that as well?

Best wishes,
Jan Hajic jr.


piskvorky:
Hello Jan, thanks, I'll have a look!

In the meanwhile, can you give some high level motivation / use cases for this?

For example, what is the advantage of using shards over a plain memory map = `matutils.Dense2Corpus(numpy.load(mmap='r'))`? Under what circumstances is this new corpus a win? When do you yourself use it?

There may be some overlap with joblib IIRC (but that package never really worked for me).

CC @larsmans

piskvorky:
@hajicj can you fix the Travis errors, so the PR can be merged?

I'm pointing people toward this pull request, and I'd like to include this in the next release of gensim, to test it out. Cheers.

hajicj:
@piskvorky Thanks for the heads-up & sorry for the delay, I had quite some catching up to do elsewhere. I have trouble getting travis to show me test results, I'll fix it as soon as I find a way to see what's wrong.

Back to the original questions:

I use ShardedCorpus for iterative training of a theano model, where I need lots of numpy minibatches fast from a larger-than-memory dataset. (I wrote the first version of the class when I found out that calling __getitem__ on a MmCorpus was a major bottleneck.) I could use a huge mmapped numpy array for that, but I'd have to create it in the first place; ShardedCorpus takes care of that. Plus, ShardedCorpus also supports serializing to scipy.sparse shards, I'm not sure joblib can deal with that.

The serializer is designed for a situation where you need the dense representation of the corpus, and you need it repeatedly. The 'gensim' init arg of ShardedCorpus is set to True by default only to make gensim-style output the default -- but the point of the serializer is getting the *numpy* (or sparse scipy) representation of a large corpus, for whatever purpose (like, say, neural network training).

Comparing to joblib: I don't really know joblib (though I tried it out for a bit to answer your question), so these are more impressions than rigorously tested ideas, but:
- ShardedCorpus is more convenient for dealing with larger-than-memory-datasets in gensim, as it only interates over the input corpus and expands by itself, without having to know the length of the data.
- Nowhere is there a single larger-than-memory numpy array allocated, not even on disk.
- You currently cannot control the location of the shard files, but it should be easy to add; that way, you can distribute your data any way you want over the filesystem. I won't pretend I really understand map-reduce, but couldn't that be an advantage in a MR setting?
- Another possible advantage is that the shard is actually loaded into memory, so accessing items that lie in the same shard requires no disk reads. (Although there may be a joblib mechanism for that, actually.)

It's true that reading from a ShardedCorpus with a pattern of accesses that requires a lot of shard-changing can be slow. Joblib can probably beat ShardedCorpus easily in that scenario; depends on the joblib indexing scheme, I guess. Unless it's a really bad case of shard-jumping, you should be able to combat it by increasing the shard size (although loading a larger shard will take longer, so this is *not* a universal cure).

Note that it should be possible to add joblib or any other memoize-pattern caching over ShardedCorpus.__getitem__ if it turns out that it will significantly speed things up.

I hope this clarifies how ShardedCorpus can be useful.

piskvorky:
Great, thanks!

Regarding Travis: you can access its test results under the [Details](https://travis-ci.org/piskvorky/gensim/builds/47695241) link, at the bottom of this PR page (where it says "X Failed - The Travis CI build failed").

piskvorky:
Thanks.

Are we good to merge this for the release, or is there any other functionality you wanted to add/change?

hajicj:
I'm ready if you are. (I do plan to add the option of serializing directly in gensim sparse vector format, but it's not critical for the primary use case and definitely will not be ready this week.)

piskvorky:
Ok, thanks, merging!

PR #292, opened 2015-02-09
Override Dictionary save method; allow passing filename or file handle.

### LINKED ISSUE #291: Allow dictionary.save to take an open file handle.
It would improve composability if the `save` method of a `Dictionary` instance could take an open file handle instead of a filename. In particular, there are certain frameworks which only pass around file-like objects and do not use filenames, such as `luigi`. This change would increase compatibility from frameworks like these and provide room for other custom file-like objects to be used.

macks22:
Seeing that the inherited `save` method includes a great deal of special cases that are irrelevant for a dictionary, the most efficient resolution is to override. I considered including the same method parameters, but none of them are really relevant. `ignore` could be used, but leaving out any dictionary attributes would likely cause errors after loading again. Resolves #291.

### DIFF
--- a/gensim/utils.py
+++ b/gensim/utils.py
@@ -20,7 +20,7 @@ except ImportError:
 try:
     import cPickle as pickle
 except ImportError:
-    import pickle as pickle
+    import pickle
 
 import re
 import unicodedata
--- a/gensim/utils.py
+++ b/gensim/utils.py
@@ -228,19 +228,22 @@ class SaveLoad(object):
             setattr(obj, attrib, None)
         return obj
 
-    def save(self, fname, separately=None, sep_limit=10 * 1024**2, ignore=frozenset()):
+    def _smart_save(self, fname, separately=None, sep_limit=10 * 1024**2, ignore=frozenset()):
         """
         Save the object to file (also see `load`).
 
-        If `separately` is None, automatically detect large numpy/scipy.sparse arrays
-        in the object being stored, and store them into separate files. This avoids
-        pickle memory errors and allows mmap'ing large arrays back on load efficiently.
+        If `separately` is None, automatically detect large
+        numpy/scipy.sparse arrays in the object being stored, and store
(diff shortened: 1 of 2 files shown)


piskvorky:
Thanks. The same `file_or_handle` functionality could be useful for other classes, too. I'd prefer having it in the base class (=not override Dictionary only), it's more future-proof.

Also, use the highest protocol (-1) for pickling, to save space & memory.

macks22:
I moved the alternate functionality to the `SaveLoad` base class and changed the pickle protocol to `pickle.HIGHEST_PROTOCOL`, which reads more transparently than -1. I also renamed the pickling helper functions to `smart_pickle` and `smart_unpickle` and altered the pickle import to use the name `pickle` rather than `_pickle`. This will reduce the overhead required for other developers to understand the code in the future.

piskvorky:
-1 on the pickle renaming. The idea is OK, but it's not backward compatible and there's tons of code that relies on `utils.pickle`. We can't just change the API like that.

The rename also seems to be the reason why all Travis tests fail.

macks22:
Roger that. I reverted the pickle renaming.

gensim/utils.py
```diff
@@ -20,7 +20,7 @@
 try:
     import cPickle as _pickle
 except ImportError:
-    import pickle as _pickle
+    import _pickle
```
piskvorky: Where does this `_pickle` come from? Travis fails on Python3, I think it's related.

macks22:
Hmm... I guess the tests missed this case. I assumed all tests passing locally covered the bases. It's been fixed in the most recent commit. Sorry about that.

piskvorky:
Great, merging. Much thanks, Mack!

Re. local tests: depends which Python you run them with. If you ran them with Python3 locally, you'd see the errors. If you only run them with Python 2 (as I assume you did), you'd miss those.

That's why we use Travis, for peace for mind :)

PR #295, opened 2015-02-18
Move wrappers for external modeling programs into a submodule

cscorley:
This change moves all wrappers for external modeling programs into a submodule.

This is for clarity and organization. I've noticed several mailing list posts with confusion on what these wrappers require, i.e., the external program. Hopefully moving these models into a submodule makes it clearer that they are not 'out-of-the-box' gensim models like the others.

### DIFF
--- a/docs/src/apiref.rst
+++ b/docs/src/apiref.rst
@@ -23,7 +23,6 @@ Modules:
     corpora/indexedcorpus
     models/ldamodel
     models/ldamulticore
-    models/ldamallet
     models/lsimodel
     models/tfidfmodel
     models/rpmodel
@@ -35,8 +34,9 @@ Modules:
     models/lda_worker
     models/word2vec
     models/doc2vec
-    models/dtmmodel
     models/phrases
+    models/wrappers/ldamallet
+    models/wrappers/dtmmodel
     similarities/docsim
     similarities/simserver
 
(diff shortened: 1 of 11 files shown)


piskvorky:
Good idea, thanks Chris!

Do we allow imports under the old path (without `.wrappers.`) too? Or only an explicit import, i.e. `gensim.models.wrappers.ldamallet`?

If the latter, we need to change the doc examples too (I see `models.ldamallet` without `wrappers` in `test_lee.py` and `ldamallet.py`, maybe elsewhere as well).

cscorley:
I think it should be explicit. Which is how I thought it was when I submitted the PR:

```python
>>> import gensim
>>> gensim.models.ldamallet
AttributeError: 'module' object has no attribute 'ldamallet'
>>> gensim.models.wrappers.ldamallet
<module 'gensim.models.wrappers.ldamallet' from 'gensim/models/wrappers/ldamallet.pyc'>
```

I'll work on updating the bits of documentation for both LdaMallet and DtmModel to reflect the changes.

cscorley:
@piskvorky Any further input towards this change? I'll squash and merge whenever the thumbs up is given :)

piskvorky:
Sorry, github didn't notify me.

Looks good, let's merge :+1:

piskvorky:
Oh oh, Travis now reports a broken build on `master`.

cscorley:
Only after I changed the CHANGELOG.txt :-/ The commit themselves passed. Looks like downloading the software for 2.6 timed out

cscorley:
I restarted the build and it passes again! :smiley_cat:

Issue #301, opened 2015-03-08
adding support to read doc-topics.txt generated by LdaMallet


chiragnagpal:
Hi! I'm back again.
So I have a corpus of about 600,000 documents over which I have trained an LDA model using Mallet.

when I try to extract the topic distributions using lda[corpus], it appears as if LDA is re-trying to infer the distribution for each document, leading me to memory issues.

Since mallet is already generating the doc-topics.txt would it not be simpler enough to parse output from the file, to infer topic distribution atleast for the training corpus? 

Would adding a function to the wrapper to do this be helpful?

piskvorky:
In general, gensim avoids loading the training corpus (or any structures the size of training corpus) into RAM, to avoid memory issues.

But since Mallet loads everything into RAM anyway, I guess it makes no difference here. A wrapper will be welcome.

Chirag, if you add that wrapper, can you also check it gives the same results as the current inference? It would be weird if these two gave different output.

chiragnagpal:
yup WILCO

chiragnagpal:
@piskvorky so i checked the results of the current inference with the doc-topics.txt and it does pass the sniff test, the results are consistent

tmylk:
Functionality added

PR #302, opened 2015-03-08
Issue #301 Adds a wrapper to read Mallet's topic distribution file for t...

chiragnagpal:
...raining corpus

added a method to class LdaMallet, load_document_topics(), that reads the doctopics.txt using the function read_doctopics()


cscorley:
Hey @chiragnagpal,

Looks like this would be fine, but there are mixed indentation (tab and spaces), which Python 3 happily rejects as a syntax error. Could you convert this PR to using 4-spaces per tab? Thanks!

PR #303, opened 2015-03-09
Master

mataddy:


piskvorky:
Hey @mataddy , what's this?

mataddy:
Hi @piskvorky, the pull req was a key slip.   I tried to close it before it got in your way.  apologies! 

in case you are curious: I forked gensim and am making some additions in proof of concept for a project we're working on.  The ideas is to train multiple w2v on sentences corresponding to some class labels.  So if you have 2  lables (e.g., pos/neg), you'd train 2 w2v, one on the sentences corresponding to each. Then new un-labeled sentences are scored for each class by their probability according to the skipgram model used in training, and you label them according to whichever gives higher probability.

the only additional code needed is to score the sentences, so I just adapted your training functions to do so.

In theory this works very nicely.  in practice, we'll see...

piskvorky:
Hey, that sounds interesting! Please do follow up with your results.

mataddy:
I will do for sure.  my little notebook example showing it works as well as my best case model for small a small dataset (< 1 mil sentences, many neutral) is at http://nbviewer.ipython.org/github/taddylab/deepir/blob/master/w2v-inversion.ipynb.  Since the interactions w2v finds get useful for bigger datasets, I expect that's where the gains occur.  we're scaling and I'll give you and update.

thanks for the awesome code to work from!  gensim is great.

piskvorky:
Good luck, looking forward!
I tweeted out your notebook, hope you don't mind.

mataddy:
that's awesome. thanks for the publicity!

mataddy:
hey @piskvorky the short paper at http://arxiv.org/abs/1504.07295 is going to be in ACL 2015.  Let me know if you have interest in pulling the likelihood scoreing functions I use into gensim; it is basically your 'train' function adapted to return likelihood scores and not change the model.  I have them coded only for huffman trees at this point and would want to do some more robustness checking before sending anything your way.

piskvorky:
Hmm, how complex is it? Maybe just open the pull request (even if incomplete), so I know what we're talking about.

Ping @gojomo -- thoughts?

mataddy:
OK, I'll re-open right now so you can see.  If you want to go ahead I'd undo any changes to the doc2vec stuff that I added while evaluating that for the paper, so the only differences worth paying attention to would be in the word2vec and word2vec_inner files.  I also need to document, etc.

piskvorky:
Yes, as focused PR as possible would be great. @gojomo has made many changes to doc2vec in #356, let's avoid potential conflicts as much as possible. Thanks!

mataddy:
actually I just merged with your develop branch; let me go through that to make sure everything works good before sending a pull your way.  I'll definitely avoid any changes at all in doc2vec.

traveling right now, will do this in the next week or so.

gojomo:
Very interesting! I changed the title to be more descriptive. 

For Doc2Vec inference, I've changed the train_ methods to take extra parameters, controlling to what extent the train-process actually changes different parts of the model (if at all; for inference it only changes the passed-in candidate vector). This feature might fit in similarly, if some optional error/probability-reporting function could be slotted in.

I imagine there's no support for negative-sampling because the theoretically-justified 'probability' value would require a massive loop over all possible predicted words.. But maybe for the specific case of binning a sentence to the best of N alternate models, it would be good enough to find some efficiently-calculable non-probability 'score' that's still possible with neg-sampled models?

mataddy:
I think the analogue for negative sampling would be to calculate a subsampled likelihood, where the subsampling is the same as whatever scheme was used for the original negative sampling.  I didn't code this for the paper because it's a bit hard to justify to my audience, but i can see how its worthwhile for the purposes of uniformity (i.e., so that you can score sentences regardless of how you trained the model).

PR #304, opened 2015-03-10
Added wrapper class around Vowpal Wabbit's online LDA implementation.

davechallis:
This adds a new class which acts as a wrapper around Vowpal Wabbit's LDA implementation (https://github.com/JohnLangford/vowpal_wabbit/wiki).

Vowpal Wabbit works on files, so this wrapper reads/writes files from/to a temporary directory while it's around.

The API for this was based on Genesis's LdaModel (and to a lesser extent LdaMallet class), and arguments/methods chosen to match that where possible.

Tested under python 2.6, 2.7, and 3.4, and with Vowpal Rabbit 7.10 (the current release). I'll aim to keep this working with the latest release.

### DIFF
--- a/gensim/models/wrappers/ldavowpalwabbit.py
+++ b/gensim/models/wrappers/ldavowpalwabbit.py
@@ -277,6 +277,10 @@ class LdaVowpalWabbit(utils.SaveLoad):
             with open(self._model_filename, 'rb') as fhandle:
                 self._model = fhandle.read()
 
+        if not self._topics and os.path.exists(self._topics_filename):
+            # ensure topics are read from file before serialisation
+            self._load_vw_topics()
+
         super(LdaVowpalWabbit, self).save(fname, *args, **kwargs)
 
     @classmethod
--- a/gensim/models/wrappers/ldavowpalwabbit.py
+++ b/gensim/models/wrappers/ldavowpalwabbit.py
@@ -128,7 +128,7 @@ class LdaVowpalWabbit(utils.SaveLoad):
         used by this wrapper. Setting to False can be useful for debugging,
         or for re-using Vowpal Wabbit files elsewhere.
 
-        `prefix` used to prefix temporary working directory name.
+        `tmp_prefix` used to prefix temporary working directory name.
         """
         # default parameters are taken from Vowpal Wabbit's defaults, and
         # parameter names changed to match Gensim's LdaModel where possible
@@ -270,6 +270,9 @@ class LdaVowpalWabbit(utils.SaveLoad):
     def save(self, fname, *args, **kwargs):
         """Serialise this model to file with given name."""
         if os.path.exists(self._model_filename):
+            # Vowpal Wabbit uses its own binary model file, read this into
+            # variable before serialising this object - keeps all data
+            # self contained within a single serialised file
             LOG.debug("Reading model bytes from '%s'", self._model_filename)
             with open(self._model_filename, 'rb') as fhandle:
                 self._model = fhandle.read()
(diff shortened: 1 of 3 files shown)


cscorley:
Dave, this looks great! Thanks so much for working on this. Going through and adding a couple of comments. Will merge in once everything looks good :)

gensim/models/wrappers/ldavowpalwabbit.py
```diff
@@ -0,0 +1,537 @@
...
+    def show_topic(self, topicid, topn=10):
+        topics = self._get_topics()
+        topic = topics[topicid]
+        bestn = numpy.argsort(topic)[::-1][:topn]
+        return [(topic[t_id], self.id2word[t_id]) for t_id in bestn]
+
+    def save(self, fname, *args, **kwargs):
+        """Serialise this model to file with given name."""
+        if os.path.exists(self._model_filename):
+            LOG.debug("Reading model bytes from '%s'", self._model_filename)
```
cscorley: Why does .save() read and .load() write?

gensim/models/wrappers/ldavowpalwabbit.py
```diff
@@ -0,0 +1,537 @@
...
+                 cleanup_files=True, tmp_prefix='tmp'):
+        """`vw_path` is the path to Vowpal Wabbit's 'vw' executable.
+
+        `corpus` is an iterable training corpus. If given, training will
+        start immediately, otherwise the model is left untrained (presumably
+        because you want to call `update()` manually).
+
+        `num_topics` is the number of requested latent topics to be extracted
+        from the training corpus.
+        Corresponds to VW's '--lda <num_topics>' argument.
```
cscorley: Thanks so much for adding these lines to the documentation. Makes it very clear how the parameters translate.

gensim/models/wrappers/ldavowpalwabbit.py
```diff
@@ -0,0 +1,537 @@
...
+        Corresponds to VW's '--epsilon <eps>' argument.
+
+        `random_seed` sets Vowpal Wabbit's random seed when learning.
+        Corresponds to VW's '--random_seed <seed>' argument.
+
+        `cleanup_files` whether or not to delete temporary directory and files
+        used by this wrapper. Setting to False can be useful for debugging,
+        or for re-using Vowpal Wabbit files elsewhere.
+
+        `prefix` used to prefix temporary working directory name.
```
cscorley: Shouldn't this be `tmp_prefix`?

davechallis: Ah yes, you're correct, I'll fix this tomorrow morning.

gensim/models/wrappers/ldavowpalwabbit.py
davechallis: Vowpal Wabbit uses its own binary model format, so when saving, this is read into a variable, so that it's serialised as part of this of this class (this means there's just a single serialised file containing both the model and the wrapper class).

The same (but opposite) goes for load(), when the wrapper is loaded, it writes the binary Vowpal Wabbit model back to disk again, so that VW can start using it right way (i.e. for predictions, further updates, etc.).

It is a bit confusing though, so I can add some comments around that.

cscorley: Ah, I see. Thanks for clarifying.

piskvorky:
Are we good to merge here?

@cscorley Chris, any comments?

Looks good to me. l plan to change all the `open()`s to use [smart_open](https://github.com/piskvorky/smart_open) instead, but I'll probably do it en-masse, across all of gensim in a separate PR, so this is not blocking here in any way.

piskvorky:
And thanks again Dave -- it is rare we get such an all rounded pull request :+1:

davechallis:
Great, glad it made it in :)  I've been using it for a few weeks without any problems now, only thing I'd like to improve in future is shrinking the serialised model size a bit (some of the Vowpal Wabbit generated files can be large, but can also be compacted).

piskvorky:
Well, `smart_open` can open compressed files transparently, so if that's an option, working with zip/bz2 serialized models could be straightforward!

davechallis:
Ah, good to know, that might definitely be a good option to start with! There's one VW file that's basically a matrix encoded as ascii text, so converting that to a numpy matrix would probably shrink the model by quite a bit, so that's also something I'll look into.

Issue #273, opened 2014-12-31
Pickle issue with LdaMulticore


cwharland:
I'm trying to use LdaMulticore to speed up model training and I'm running into a pickle error:

```
PicklingError: Can't pickle <function worker_e_step at 0x0000000036D26748>: it's not found as gensim.models.ldamulticore.worker_e_step
```

The error occurs when running:

```
lda = gensim.models.ldamulticore.LdaMulticore(corpus=corpus, id2word=dictionary, num_topics=10)
```

Single threaded works just fine with no error:

```
lda = gensim.models.ldamodel.LdaModel(corpus=corpus, id2word=dictionary, num_topics=10)
```

I'm running this on Windows with gensim 0.10.3.  Haven't tried on another platform so can't rule out platform problem.  I can pickle other things just find so I don't think my pickle library is messed up.  It looks like this is some kind of serialization issue possibly related to [this](http://matthewrocklin.com/blog/work/2013/12/05/Parallelism-and-Serialization/) post on pickle vs. dill

maifeng:
Got the same error on Windows.

az0:
I also have the same error on Windows 7 and Python 2.7.9 64-bit

piskvorky:
Can you guys move the inner `worker_e_step` function in `ldamulticore.py` (or any other function that fails on Windows) into module scope, and open a pull request?

maifeng:
I just tested the suggestion under Windows and it is working now. I also opened a pull request.

piskvorky:
I've merged @maifeng 's fix.

@az0 @cwharland can you test, using the `develop` branch?

cwharland:
Sorry for the late response but it worked for me too from develop branch

Issue #256, opened 2014-11-01
setup Error "Type Error: an integer is required"


lvqian:
I was tried hundreds times...   But something are still wrong.
My system is centOS. 
My python version is 2.6.6
I use yum install numpy scipy

Because the network problem, I can't use "easy_install" or "pip" to get gensim.
So, I down the gensim.tar.gz, and unzip it, change location,
then, "sudo python setup.py install" 
the information as follows:

running install
running bdist_egg
running egg_info
writing requirements to gensim.egg-info/requires.txt
writing gensim.egg-info/PKG-INFO
writing top-level names to gensim.egg-info/top_level.txt
writing dependency_links to gensim.egg-info/dependency_links.txt
reading manifest file 'gensim.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching '*.sh' under directory '.'
no previously-included directories found matching 'docs/src*'
writing manifest file 'gensim.egg-info/SOURCES.txt'
installing library code to build/bdist.linux-x86_64/egg
running install_lib
running build_py
creating build
creating build/lib.linux-x86_64-2.6
creating build/lib.linux-x86_64-2.6/gensim
copying gensim/utils.py -> build/lib.linux-x86_64-2.6/gensim
copying gensim/interfaces.py -> build/lib.linux-x86_64-2.6/gensim
copying gensim/matutils.py -> build/lib.linux-x86_64-2.6/gensim
copying gensim/nosy.py -> build/lib.linux-x86_64-2.6/gensim
copying gensim/__init__.py -> build/lib.linux-x86_64-2.6/gensim
creating build/lib.linux-x86_64-2.6/gensim/models
copying gensim/models/ldamallet.py -> build/lib.linux-x86_64-2.6/gensim/models
copying gensim/models/tfidfmodel.py -> build/lib.linux-x86_64-2.6/gensim/models
copying gensim/models/tst.py -> build/lib.linux-x86_64-2.6/gensim/models
copying gensim/models/ldamulticore.py -> build/lib.linux-x86_64-2.6/gensim/models
copying gensim/models/test.py -> build/lib.linux-x86_64-2.6/gensim/models
copying gensim/models/lsimodel.py -> build/lib.linux-x86_64-2.6/gensim/models
copying gensim/models/dtmmodel.py -> build/lib.linux-x86_64-2.6/gensim/models
copying gensim/models/ldamodel.py -> build/lib.linux-x86_64-2.6/gensim/models
[...]

PR #305, opened 2015-03-11
Update ldamulticore.py

maifeng:
Moved the inner worker_e_step function into module scope of ldamulticore.py, so LdaMulticore is working properly in Windows. See Pickle issue with LdaMulticore #273


piskvorky:
Great, thanks Feng Mai!

Merging.

piskvorky:
This also fixes the anniversary issue #256.

PR #306, opened 2015-03-11
Issue #301 Adds a wrapper to read Mallet's topic distribution file for t...

chiragnagpal:
...opic distribution


piskvorky:
Thanks Chirag!

Can you add a unit test for this new functionality too? The place for that would be inside the [TestLdaMallet](https://github.com/piskvorky/gensim/blob/develop/gensim/test/test_models.py#L369) class.

PR #309, opened 2015-03-31
fixed 3.x-incompatible use of dict.keys()

jordanm:


piskvorky:
Why close? What happened @jordanm ?

PR #311, opened 2015-04-02
added topic ranking method

jtmcmc:
added a method to the LdaModel class that ranks topics based on the umass algorithm for topic coherence.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -754,19 +754,22 @@ class LdaModel(interfaces.TransformationABC):
         coherence_scores = []
         str_topics = self.show_topics(num_topics=self.num_topics,
                 num_words=num_words,formatted=False)
-        topics = [ [x[1] for x in topic] for topic in str_topics]
-        topics = [self.id2word.doc2bow(topic) for topic in topics]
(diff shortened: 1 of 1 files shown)


jtmcmc:
Thanks smargh! I changed 2 of the 3 list comprehensions to generators the last one needs to be randomly accessed and won't actually be larger than the results of show_topics so it may not have a large memory problem. I realized that at 764 binary_corpus was a possibly disastrous list comprehension which I've replaced by just using the corpus directly. the resulting code isn't as nice, however.

cscorley:
Excellent! Thanks.

On line 758, you use `self.id2word.doc2bow`. Unfortunately, we cannot assume that the dictionary given to the model was a `gensim.corpora.Dictionary`. Can you replace this with something that doesn't make this assumption? It may also to help simplify those generators out entirely if you did not rely on `self.show_topics` and retrieved the topic-term vectors directly

jtmcmc:
Ah  I didn't know that. I've removed the usage of doc2bow and retrieved the topic-term vectors directly.

cscorley:
Thanks!

piskvorky:
@jtmcmc would you be interested in writing a short exemplification / article about this new functionality?

Please ping me at me@radimrehurek.com :)

PR #315, opened 2015-04-08
Allow custom `ignore` parameter in LdaModel.save

mattilyra:
In `LdaModel.save` the `ignore` list is explicitly passed to the parent method, making it impossible to add anything to that list. I propose that the `ignore` list be handled in the following way instead.

    if 'ignore' in kwargs:
        kwargs['ignore'] = list(set(['state', 'dispatcher']) | set(kwargs['ignore']))
    else:
        kwargs['ignore'] = ['state', 'dispatcher']
    super(LdaModel, self).save(fname, *args, **kwargs)


piskvorky:
Travis reports fails in Python 3; we use `six` for dual compatibility.

For example, `basestring` in Python 2 becomes `six.string_types` for Py 2 & 3.

mattilyra:
I somehow forgot that this was still open. I fixed it now and the tests pass. Upstream develop has been merged into the latest commit.

piskvorky:
Looks good. Thanks @mattilyra !

Issue #319, opened 2015-04-09
Word2vec not training


fccoelho:
I am trying to use the latest gensim with python 3 but it simply does not start training the nnet after ingesting the sentences:

```
2015-04-09 08:53:38,848 : INFO : PROGRESS: at sentence #40000, processed 11825151 words and 176854 word types
2015-04-09 08:53:42,090 : INFO : collected 200158 word types from a corpus of 14837441 words and 50000 sentences
2015-04-09 08:53:42,163 : INFO : total 17671 word types after removing those with count<50
2015-04-09 08:53:42,163 : INFO : constructing a huffman tree from 17671 words
2015-04-09 08:53:42,699 : INFO : built huffman tree with maximum node depth 18
2015-04-09 08:53:42,714 : INFO : resetting layer weights
2015-04-09 08:53:43,103 : INFO : training model with 8 workers on 17671 vocabulary and 500 features, using 'skipgram'=1 'hierarchical softmax'=1 'subsample'=0 and 'negative sampling'=0
2015-04-09 08:53:43,124 : INFO : reached the end of input; waiting to finish 0 outstanding jobs
2015-04-09 08:53:43,125 : INFO : training on 0 words took 0.0s, 0 words/s
2015-04-09 08:53:43,125 : INFO : saving Word2Vec object under MediaCloud_d2v, separately None
2015-04-09 08:53:43,125 : INFO : not storing attribute syn0norm
2015-04-09 08:53:43,421 : INFO : precomputing L2-norms of word weight vectors
```

 Notice that it does not throw any exceptions so I really don't know what may be happening.

piskvorky:
What type are the input sentences?

fccoelho:
sentences are  lists of strings, for example:
```python
['argentino', 'fase', 'final', 'recuperação', 'lesão', 'muscular', 'ainda', 'espera', 'alta', 'médica', 'voltar', 'gramados']
```

fccoelho:
I have found the problem:

I was passing the sentences to word2vec as a generator, but that apparently is not acceptable.

```python
def train_w2v_model(model_name="MediaCloud_w2v"):
    print("Training model...")
    sentences = sentence_gen(50000)
    model = Word2Vec(sentences, workers=num_workers, \
                     size=num_features, min_count=min_word_count, \
                     window=context, iter=3)
```

if I pass a list of sentences, it works.

```python
def train_w2v_model(model_name="MediaCloud_w2v"):
    print("Training model...")
[...]

PR #320, opened 2015-04-09
Added testing for generator being passed to Word2Vec, and added a unitte...

fccoelho:
Added testing for generator being passed to Word2Vec, and added a unittest for it.

Related to issue #319

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -89,6 +89,7 @@ logger = logging.getLogger("gensim.models.word2vec")
 from gensim import utils, matutils  # utility fnc for pickling, common scipy operations etc
 from six import iteritems, itervalues, string_types
 from six.moves import xrange
+from types import GeneratorType
 
 
 try:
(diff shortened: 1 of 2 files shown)


gensim/test/test_word2vec.py
```diff
@@ -257,6 +257,13 @@ def models_equal(self, model, model2):
         self.assertTrue(numpy.allclose(model[most_common_word], model2[most_common_word]))
 #endclass TestWord2VecModel
 
+    def test_sentences_should_not_be_a_generator(self):
+        """
+        Is sentences a generator object?
+        """
+        gen = (s for s in sentences)
+        self.assertRaises(TypeError, word2vec.Word2Vec(gen))
```
piskvorky: This won't work -- by the time `assertRaises` gets its arguments, `Word2Vec(gen)` has already been called and failed. the generator must be an extra argument.

piskvorky:
Thanks a lot Flávio, merging! :+1:

gensim/models/word2vec.py
```diff
@@ -309,6 +310,8 @@ def __init__(self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
         self.hashfxn = hashfxn
         self.iter = iter
         if sentences is not None:
+            if isinstance(sentences, GeneratorType):
+                raise TypeError("You can't pass a generator as the sentences argument. Try an iterator.")
```
piskvorky: There's a typo here that slipped the code review -- "iterator" should read "iterable". Can you fix this?

Issue #313, opened 2015-04-08
LdaModel 2 to 3 save/load fails if alpha='auto'


mattilyra:
An LdaModel that has been trained with `alpha='auto'` or a preset `alpha` or `eta` under Python 2 and consequently saved using `LdaModel.save` can not be loaded in Python 3 because the small `numpy` arrays are pickled in with the model. This raises a `UnicodeDecodeError` because `pickle` is trying to do silly things with the `numpy` arrays. 

There's a fairly trivial fix to this, which is to call `LdaModel.save(separately=['alpha', 'expElogbeta'])`, however this should at least be documented somewhere. Notice that `expElogbeta` is usually always dumped separately using `numpy` because it exceeds the `sep_limit=10 * 1024**2` set in `gensim.utils.LoadSave.save` but `alpha` is unlikely to exceed that.

I'll put in a pull request for improved documentation unless someone has a better idea.

mattilyra:
Slight correction to the above, the save call has to be `LdaModel.save(separately=['alpha', 'expElogbeta', 'sstats'])`, otherwise `LdaModel.state` will be pickled with sstats in the object, which will again fail trying to load the object in Python 3

piskvorky:
I'm not terribly sure we want to commit to persistence compatibility between Python 2/3.

I mean, it would be nice to have, but I wonder how much effort it would be to do it properly (not just for `LdaModel`), and to maintain it going forward.

CC @cscorley : Chris have you come across this scenario? Is this something users need? Why do you need it, Matti?

mattilyra:
I was just trying to load some models I had on disk in Python 3 and noticed that it doesn't quite work. It's not essential that I get them loaded into 3, which is why I suggested amending the documentation so that if someone actually has to do this, they'll at least have somewhere to start from.

cscorley:
Pickles are in general usable across Python versions, but as noted also generally only work on native Python structures. It may be a good default for Numpy structures to save in their own format and that change should be straightforward. Likewise, similar issues may arise with corpora.Dictionary. It would be a good wishlist item, I think, but for now noting the volatility of Python 2/3 persistence in the docs is definitely needed. 

@mattilyra Mind updating the docs with exactly situations what fail and an example on how to avoid it? The Wiki Recipes may be a good place for the latter.

mattilyra:
@cscorley Sorry for not getting back earlier. I'll update the docs, and put in some examples on how to do the migration.

cscorley:
[...]

PR #331, opened 2015-04-28
Fix for issue 313

mattilyra:
Added a comment to LdaModel.save about the difficulties in migrating saved models from Python 2 to 3. Fixes issue #313


piskvorky:
Thanks Matti!

PR #334, opened 2015-05-06
Trivial changes

midnightradio:
I found that gensim already has some useful tools for preprocessing and tokenizing and I was using external librarys for the same purpose while I'm only focusing heavier modules in gensim. However, the usage was not clear and led me to read the codes in the module, utils and parsing, and put some words  on them.

I wonder if there's any specific reason for not loading the module 'parsing' while importing 'utils' under the top module 'gensim'. I think it gives better usability to load 'parsing' instead of 'utils', as users can access namespaces in 'utils' in both ways of 'gensim.utils' and 'gensim.parsing.utils', as well as having the package 'gensim.parsing' without importing it explicitly. Of course it won't affect to users already put the namespace 'gensim.utils' in their codes.

### DIFF
--- a/gensim/__init__.py
+++ b/gensim/__init__.py
@@ -3,7 +3,7 @@ This package contains interfaces and functionality to compute pair-wise document
 similarities within a corpus of documents.
 """
 
-from gensim import utils, matutils, interfaces, corpora, models, similarities
+from gensim import parsing, matutils, interfaces, corpora, models, similarities
 import logging
 
 try:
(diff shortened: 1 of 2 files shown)


piskvorky:
That makes sense, thanks @midnightradio !

The reason `parsing` is a bit 2nd class citizen is because it was contributed by someone else, and the code is not as well tested and designed as the rest of gensim.

But if it works for you, no problem. Thanks for the doc updates!

midnightradio:
Yeah, I can see what you meant while getting more curious with finding some tricky behaviors as I read the code more, which is not a bug but one might have to be aware of before using it. I understand the reason for placing it as 2nd class citizen rather than disclosing. Maybe it's better to promote using it and contributing to it. And I would gladly be one of users and contributors starting from this merge. :-)

piskvorky:
Great :) Looking forward to it!

PR #335, opened 2015-05-07
Fixes to string/byte handling with python 3 and smart_open.

davechallis:
This fixes an issue with using smart_open in the ldavowpalwabbit wrapper in python 3.

smart_open reads/writes as bytes rather than string, so data being
written or read is now of the expected type (i.e. string encoded as bytes before writing, and assumed to be bytes rather than string when being read).

Tested with python 2.6, 2.7, 3.4.


piskvorky:
Thanks Dave! Merging.

Automated unit testing of the wrapper modules leaves a lot to be desired...

PR #338, opened 2015-05-11
a single string label for LabeledSentence in doc2vec

zygmuntz:
...without wrapping in a sequence.

### DIFF
--- a/gensim/__init__.py
+++ b/gensim/__init__.py
@@ -3,7 +3,7 @@ This package contains interfaces and functionality to compute pair-wise document
 similarities within a corpus of documents.
 """
 
-from gensim import utils, matutils, interfaces, corpora, models, similarities
+from gensim import parsing, matutils, interfaces, corpora, models, similarities
 import logging
 
 try:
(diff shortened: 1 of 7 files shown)


cscorley:
I can be ok with this change.

ping @gojomo , what do you think? Will this change in your fork?

cscorley:
By the way, @zygmuntz, can you squash this into a single commit, since it's fairly straightforward?

piskvorky:
@cscorley What fork?

This is fine by me, +1.

Also, +1 on commit squash before merging. Thanks!

zygmuntz:
Since I'm mostly unfamiliar with `rebase`, I squashed them by brute force into https://github.com/piskvorky/gensim/pull/339, closing dis.

gojomo:
I don't expect any problems from this small convenience change (auto-tupling a single string passed to the LabeledSentence constructor). I'm trying to maintain the option of multiple-labels-per-doc, both because it might be important to the original Doc2Vec author @temerick, and I can envision some interesting potential uses (like one of the slots identifying an author). But in practice I've only been testing with a single label per doc. 

@piskvorky – @cscorley likely just means my working fork.

cscorley:
Okay, thanks! I'll merge in the other PR.

PR #339, opened 2015-05-13
Allow "single string label" for LabeledSentence in doc2vec

zygmuntz:
...without wrapping it in a sequence!


piskvorky:
What's with the dollar sign? Why did you name your commit that way @zygmuntz ?

zygmuntz:
I like to stick emojis into my commits, this time :dollar: caught my eye in the [list](http://www.emoji-cheat-sheet.com/). I hope you aren't offended.

PR #341, opened 2015-05-14
handle iter inside train() for consistency

gojomo:
as reported/discussed at https://groups.google.com/forum/#!topic/gensim/HKE8Y5MeC-k


piskvorky:
+1 @gojomo ; if this works for you, go ahead and merge.

PR #343, opened 2015-05-20
Update word2vec.py

aykutfirat:
Python's split, when used with no parameter, uses 'space, tab, newline, return, formfeed' to split a string. Mikolov's word2vec, however, may produce words that include non-whitespace spaces. Explicitly putting whitespace, and eliminating the newline, gets rid of annoying errors.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -606,7 +606,7 @@ class Word2Vec(utils.SaveLoad):
                     result.syn0[line_no] = fromstring(fin.read(binary_len), dtype=REAL)
             else:
                 for line_no, line in enumerate(fin):
-                    parts = utils.to_unicode(line).split()
+                    parts = utils.to_unicode(line.rstrip()).split(" ")
                     if len(parts) != layer1_size + 1:
                         raise ValueError("invalid vector on line %s (is this really the text format?)" % (line_no))
                     word, weights = parts[0], list(map(REAL, parts[1:]))


gensim/models/word2vec.py
```diff
@@ -606,7 +606,7 @@ def load_word2vec_format(cls, fname, fvocab=None, binary=False, norm_only=True):
                     result.syn0[line_no] = fromstring(fin.read(binary_len), dtype=REAL)
             else:
                 for line_no, line in enumerate(fin):
-                    parts = utils.to_unicode(line).split()
+                    parts = utils.to_unicode(line.rstrip()).split(" ")
```
piskvorky: Do we need the `rstrip`? Sounds like it introduces the same problem you're trying to solve, no? Maybe just remove the final byte (newline).

piskvorky:
Ah, good point, thanks!

What was the "annoying error"? Maybe paste it here explicitly, so people can find your fix when googling for the problem.

gensim/models/word2vec.py
aykutfirat: Without rstrip we have one extra token in the end. The error was 
"invalid vector on line %s (is this really the text format?)"
and the line had a word, which seemingly had a space in it, but nevertheless produced by word2vec C code.

piskvorky: I think `" "` won't work in Python3 -- split on byte string expects a byte string (`b" "`). Did you try in Python3?

piskvorky: `rstrip` will have the same problems as `strip`, it's the same code. If it's the newline that creates problems, remove the newline only.

aykutfirat: I have not tried in Python 3, and I am no Python expert. The problem is not with the new line. It is with the more exotic spaces, such as non-breaking space(formfeed?). If you stick to the same set of separation characters as in word2vec.c -> (if ((ch == ' ') || (ch == '\t') || (ch == '\n')) ) in some way, the problem can be fixed.

piskvorky: Are you still talking about `rstrip`? I'm getting confused.

`rstrip` will remove ANY whitespace from the end of a string. We DON'T want that. We want to remove only `\n` (or whatever character C word2vec puts at the end of each line). Want want to keep any other (weird) whitespace intact.

Btw simulating split ala `(ch == ' ') || (ch == '\t') || (ch == '\n')` is just `[part for part in re.split((r"[ \t\n]", x) if part]` in Python.

aykutfirat: The space problem is within the word, which is the first token. You can for example have something like 

hello world 0.234 0.345 ....

and assume that the space between hello and world is not in [ \t\n] 

rstrip is inconsequential. I needed it because I only used " " to split. If you use more than whitespace it may not be needed.

piskvorky: Aha, OK. I don't remember the format anymore -- didn't realize there can be no strings at the end.

Anyway, can you split on the same chars as the original C? Since we're fixing the compatibility, we might as well fix it 100%. Cheers!

PR #344, opened 2015-05-20
Update word2vec.py

aykutfirat:
Hi Radim,

No tabs or new lines are allowed in a word. So a split based on whitespace is enough (after a newline removal with rstrip). I added the whitespace with a b prefix as you suggested for python 3 compatibility(have not tested with python 3)

I also added errors="ignore" because I had errors like UnicodeDecodeError: 'utf8' codec can't decode byte 0xe2 in position 97: invalid continuation byte,(again this is with output from word2vec.)

I tested the code with my data file, and everything worked fine.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -606,7 +606,7 @@ class Word2Vec(utils.SaveLoad):
                     result.syn0[line_no] = fromstring(fin.read(binary_len), dtype=REAL)
             else:
                 for line_no, line in enumerate(fin):
-                    parts = utils.to_unicode(line).split()
+                    parts = utils.to_unicode(line.rstrip(),errors="ignore").split(" ")
                     if len(parts) != layer1_size + 1:
                         raise ValueError("invalid vector on line %s (is this really the text format?)" % (line_no))
                     word, weights = parts[0], list(map(REAL, parts[1:]))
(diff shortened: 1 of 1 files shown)


sildar:
Hi,

I juste wanted to point out that the important change you made in your commit was the addition of the "errors='ignore'" rather than the splitting. While I think your modification of the split is legit, I think that leaving unicode errors silents would lead to hard to detect errors.

piskvorky:
Yes, we cannot merge any silencing of errors.

If you use non-utf8 inputs in the C word2vec, we could add support for an extra `load_word2vec_format(encoding=...)` parameter. Default would be `encoding='utf8'`, for backward compatibility.

aykutfirat:
Ok, I think it is a legitimate point.

On the other hand, I used an output file that I wrote in java as UTF-8, ran
word2vec, and got the encoding error in one or two places when I tried to
load it with gensim.(I process a whole day of Twitter, so all kinds of
weird stuff may exist.)

If the word2vec C code is screwing up the encoding, then passing the errors
via load_word2vec_format with a default value of strict could be a
compromise?

piskvorky:
@aykutfirat did you get a chance to fix this? Let's finish the PR and merge, the splut is a good catch :)

piskvorky:
Word2vec has been refactored since, so this PR didn't merge.

I implemented your proposed changes in a new commit: ad12ec97

@aykutfirat Please let me know if the `develop` branch fixes your issues now! I also implemented `encoding` parameter to `load_word2vec_format`, in case your words are not in UTF8.

PR #345, opened 2015-05-21
Added documentation clarification to use of bytes in Phrases delimiter.

davechallis:
Added comment to documentation regarding use of byte string in Phrases (as discussed in https://github.com/piskvorky/gensim/issues/342).


piskvorky:
Thanks Dave!

PR #348, opened 2015-05-25
Amended comments for "most_similar" method.

KCzar:
Corrected an incomplete sentence, and explained what happens if topn is False.


piskvorky:
Thanks!

PR #350, opened 2015-05-28
Sped up most_similar and accuracy

KCzar:
I was surprised how long it took to run the "accuracy" function (almost 2 hours with the default settings using the google-trained vectors), so I figured I'd speed it up.  It now runs 54x faster (about 2 minutes).

There were a variety of things slowing it down:
1) it didn't take advantage of the "restricted_vocab" reduction in necessary computation (you don't need to do a dot product of the full array, just the "restricted_vocab" number of vectors
2) you don't need to sort entire arrays (nlogn) just partial arrays (nlogk for most_similar, and for accuracy, just n because you only need the best answer) regardless of whether you're using restricted vocab.

I've left in both the old functions for most_similar and accuracy under their old names, and added most_similar_v2 and accuracy_v2 separately for now for anyone who wants to compare.  I'll clean it up before we merge (there's some more functionality I want to add to most_similar anyway, but one step at a time).  As of now I haven't touched most_similar_cosmul, though the same techniques can speed it up just as much.

most_similar_v2 on its own is about 30% faster than most_similar for the default of topn=10 when using the pre-trained google vectors.  Profiling showed my helper function argTopNSorted_nda is about 10 times faster than argsort, but the performance of most_similar is dominated by matrix multiplication.  

Let me know what you think, and I can move forward with it.  

The other addition I want to make after implementing these speed ups is to add vector negation functionality.  It's not as useful for doing analogies, but it's more useful in document/keyword/concept retrieval than subtracting vectors.  I'll expand on that later but I think you'll like it.  

Cheers

P.S. I forgot to remove some extraneous files before committing, but some of them contain the results of my time tests anyway in case you're interested...


sebastien-j:
@KCzar , for the analogies, would you mind comparing with pull request #340 ? I made this pull request a few days ago, but didnt complete the unit tests yet.

On my desktop, for the GoogleNews model with default parameters, I get a runtime of ~13.5 seconds. Most of this is spent in reordering the vectors once (from largest to smallest count). If this isn't necessary (do `model.indexed_by_count=True`), it takes ~1.4 seconds.

KCzar:
Sure, I'll have a look tomorrow.  Is there any chance you can summarize for me where the gains are coming from in your method?  I'm guessing from a cursory overview that you're computing everything all at once from two large matrices...

Looks like we had slightly different goals in mind, though there is certainly overlap.  From what I see you did a pretty heavy rewrite, but didn't touch most_similar and instead removed it entirely from the accuracy function.  I suppose one question is what we want out of the "accuracy" function.

If it is important or useful that "most_similar" be a parameter so that one can experiment with different similarity functions, then my method would be useful, and perhaps worth including in addition to yours.  If the purpose of the "accuracy" function is only to test a trained model, then no question yours is much faster.

Either way, if you haven't changed anything with the most_similar function, then my changes there would be worth adding I think.  I got a modest but useful speedup in that function alone.  

I'll have a closer look tomorrow, but let me know if my analysis is off base.

sebastien-j:
@KCzar , as I didn't modify the `most_similar` function, I agree that the changes you made may indeed be useful.

My general approach is to use matrix-matrix operations to get arrays of size `num_questions x restrict_vocab` that contain the scores (similarities) for each question. The speed-up comes from doing (mini-)batch computations and from dealing with the two bottlenecks you have already identified.

I replaced `most_similar` with `non_lin`, which transforms the cosine similarity with an arbitrary non-linearity. While reasonably powerful, this approach is a little bit less general.

KCzar:
Ok - just got a chance to look over the code, and I'm about to clone and test it out.  How important is it to break it up into batches?  I would think the size of the model itself would take up much more memory relative to the additional memory needed to run accuracy.  

At the moment I'm thinking one alternative that would combine our approaches, keep the code simple, and potentially get the best of both worlds would be to modify most_similar so that it can also accept a list of lists for batch processing, and then we can still have accuracy call most_similar.  That may be useful functionality to people even outside of the accuracy function, and I'm sure we could make it run with performance similar to what you're getting.  

What do you think?

sebastien-j:
For large values of `restrict_vocab`, using smaller mini-batches may be useful. For example, I think a batch of 2000 questions with a 1,000,000 word vocabulary would use about 8 additional GB (and 24GB with 3CosMul).

I haven't thought much about modifying `most_similar`, but it could very well be a fine alternative.

KCzar:
I just ran your version on my machine.  Certainly fast but I didn't get quite the results you get.  Initial run (unsorted) was about 19.5 seconds, but after it's already sorted, it was still ~6.4 seconds average.  I have a pretty powerful computer so I'm a little surprised our results differ by ~4x.

Just finished running accuracy with the full 3,000,000 word Google vocab, batchsize=2000.  Took 18 minutes, and at peak my memory went up by ~33 extra GB.  That seems quite excessive though.  I would think the code must be copying something unnecessarily.  Numpy dot frequently makes a copy of input arrays before multiplying, but I don't see how that explains the memory usage.  Let me tinker with this a bit.

KCzar:
Ah - yes I see what's going on.  It is making unnecessary copies, which is also slowing it down a lot.  The batches shouldn't be necessary.  Let me go ahead and see if I can write up a version where most_similar has the option of taking a list of lists or tuples as I think that will be the simplest way of achieving the best of both worlds.  Might need a couple days though.

KCzar:
Just to update - almost done.  I was able to put in a little time over the weekend and can probably wrap it up tomorrow.  So far so good.  Cheers

KCzar:
Haven't fully cleaned everything up, but would you mind testing out the version I just pushed?  The functions you'll want are most_similar_v5 and accuracy_v5.

For me, accuracy ran in roughly comparable time to your version, and mostly comparable memory (posts I had found like this were apparently a pack of lies!  ...or just outdated: http://wiki.scipy.org/PerformanceTips).  I think my version may have a brief memory spike that yours does not, however.  Perhaps you could help me there.  

In the version I'm submitting, most_similar_v5 should also be faster than most_similar when topn > 1 (which doesn't come into play during accuracy testing.  The other changes I made to most_similar_v5 are that it can take (and return) lists of lists for bulk processing, though it should be fully backwards compatible.  

I had originally gone the rout of trying to make all of my solutions generalizable so that they didn't necessarily require "indexed_by_count," though ultimately I decided the code was getting too complex, and would be unnecessarily difficult for future contributors to interpret.  

Let me know what you think and we can work from there.

sebastien-j:
Nice progress! It is still a bit slower, but it's great not to have to change the API.

Using the full 3,000,000 word vocabulary, and `batchsize=200` to avoid memory overuse, my version ran in just over 200 seconds (min 202, max 209 over 3 tries). Yours took from 472 to 474s. In both cases, the total system memory use was about 11 GB.

As you mentioned earlier that you had a reasonably powerful computer, I also find it strange that my tests are much quicker than yours.

In any case, it's nice that your approach is slightly more general, helps simple nearest neighbour search and preserves the user interface. On a different note, it would be preferable if the accuracy logs for each category were displayed as soon as they become available.

KCzar:
Ah - yeah I haven't considered much about the logging b/c I guess I've had it turned off anyway.  Certainly something we can change though.  

I have no explanation for the speed differences - either between our two computers or between the two methods.  My processor is:

model name      : Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz
cpu MHz         : 1200.000
cache size      : 12288 KB
cpu cores       : 6

I currently have 64 GB of RAM on this machine too.

I would think the only explanations for how yours is so much faster would be A) maybe you have a lot more cache or B) you are using multithreaded BLAS or something.  

My version took about 1600s last time I ran it, though I was running it with cprofile, which slows it down a bit (shouldn't be THAT much though).  About 2/3 of that time was spent in numpy dot, and the other 1/3 was spent in argmax, so it's further surprising that your method was over 2x faster than mine.  I think I must be doing argmax inefficiently - perhaps you can help me there.  

I'll run each of our versions with the settings you used for comparison and report back, but otherwise I won't have any time until at least the weekend probably.  

What do you think about the general approach though?  I wonder if we should check with Radim regarding what he thinks about my proposed changes to most_similar.

KCzar:
My memory info, btw - I would think this is fairly fast memory.  What are your specs?


Handle 0x0032, DMI type 16, 23 bytes
Physical Memory Array
	Location: System Board Or Motherboard
	Use: System Memory
	Error Correction Type: Multi-bit ECC
	Maximum Capacity: 128 GB
	Number Of Devices: 8

Handle 0x0034, DMI type 17, 34 bytes
Memory Device
	Total Width: 72 bits
	Data Width: 64 bits
	Size: 16384 MB
	Type: DDR3
	Type Detail: Registered (Buffered)
	Speed: 1866 MHz
	Configured Clock Speed: 1866 MHz

sebastien-j:
Like I said before, I think your approach has multiple benefits. I didn't look carefully at the code yet, but I would hope that `most_similar_v*` could be simplified a little.

Specs:
Kubuntu 14.04, multithreaded OpenBLAS
i7-4790k @ 4-4.4 Ghz
16GB DDR3 RAM @ 2400 MHz

KCzar:
Ah - yeah, you're just flat outclassing me in terms of hardware, plus multithreaded BLAS.  I'm currently using Anaconda python, so I guess I have to either figure out a way to make that work with multithreaded BLAS (haven't found much in initial google searches) or move off of Anaconda.  

I can certainly simplify it - some parts are unnecessarily convoluted b/c of the evolutionary path it's taken through my initial effort to avoid necessarily having to index_by_count.  I'll see about cleaning it up + commenting over the weekend if I can.

I just ran cprofile versions of each of our methods.  They spend roughly the same amount of time in numpy dot (1211s to 1154s), but mine is spending way too much time in argmax, as I suspected ( 352s to 26s), even though argmax is being called about the same number of times.  The reason for that undoubtedly has to do with the order (c contiguous vs f contiguous) of the array that is being input into argmax.  I will have to rethink my code to make sure I get an array in the correct format before running argmax on it.  

Overall, your version ran in 1197s compared to mine, which ran in 1567s with no vocab restrictions and batchsize = 200.

The argmax issue would also explain why you're getting > 2x speedup, because given that you're using multithreaded BLAS, numpy dot is probably a much lower proportion of your computational expense.  

I'll get back to you on or after the weekend.

KCzar:
What instructions did you use when installing OpenBlas?  I went with these, but I'm only getting a little over a 2x speedup for numpy dot when using OpenBlas.  I assume it might not be compiling optimally:

http://gromgull.net/blog/2013/07/multithreaded-scipynumpy-with-openblas-on-debian/

sebastien-j:
@KCzar ,
I think I simply used the command at the top of `http://deeplearning.net/software/theano/install_ubuntu.html`. For some operating systems older than *ubuntu 14.04, this may limit the number of threads to 2 however.

gensim/models/word2vec.py
```diff
@@ -909,6 +913,879 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=most_similar):
         return sections
 
 
+    def argTopNSorted_nda(self, inArray_nda, topn=1):
```
piskvorky: How is this different to `gensim.matutils.argsort`?

KCzar:
I'm not sure - I'll have a look a little later or tomorrow when I get some time.  I didn't know that existed.  Previously, gensim.matutils.argsort was not being used - argsort was being imported from numpy.  Was gensim.matutils.argsort created recently?  I see that some updates within the last day or so have pertained to gensim.matutils.argsort.

Haven't had any time this week, but I can probably get back to this tomorrow.

gensim/models/word2vec.py
```diff
@@ -909,6 +913,879 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=most_similar):
...
+        """
+        if topn == 1:
+            return argmax(inArray_nda)
+        unsortedTopNIdxs_nda = argpartition(-inArray_nda, topn)[:topn]
+        subIdxs_nda = argsort(-inArray_nda[unsortedTopNIdxs_nda])
+        sortedTopNIdxs_nda = unsortedTopNIdxs_nda[subIdxs_nda]
+        return sortedTopNIdxs_nda
+
+
+    def argTopNSorted_2D_nda(self, inArray_nda, topn=1, axis=0):
```
piskvorky: Python uses `underscores_for_identifiers` by convention, not `camelCase` ([PEP 8](https://www.python.org/dev/peps/pep-0008/)).

piskvorky:
Hi guys, sorry for having gone MIA! I've gone through your comments now, great to see such speed ups!

Questions:

* How does this PR #350 relate to PR #340 ? Do we merge, use one, the other?
* I see a lot of test code, different versions for the same function etc. Do we pick one "best version" and delete the rest? Can you clean-up unneeded code paths / files?
* What else is needed to make this PR (or PR #350 ) merge-able?

I'm not that worried about performance, but I don't want to make the code base too unwieldy (complex, hard to read).

gensim/models/word2vec.py
KCzar: yeah, that's one area where I deviate from PEP 8 - I also like to label my variables with their type for convenience - but I'll clean these things up to conform with the package before I'm done.  And I'll also have a look at the built-in gensim.matutils.argsort to see if my version is duplicative (sounds like it may be).  

Do you like the idea of modifying the "most_similar" function to also accept a list of lists for bulk processing?  

My initial effort achieved considerable speed-ups and was designed to be "minimally invasive," but Sebastian showed me a more significant rewrite he had been working on with significantly greater speed-ups, which I agree are well worth the added complexity - so we've merged our approaches.  Do you approve of the path we're on generally?

seanaedmiston:
FWIW - I have been looking into these ideas for speeding up most_similar and accuracy a little bit and thought I would share a couple of observations:

* gensim.matutils.argsort is great. One small thing - the numpy argpartition needs a [::-1] right at the end to match the bottleneck implementation. Using this change alone more than doubles the speed of most_similar (from ~68ms to ~31ms on my system). The other beauty of this is that there are no API changes anywhere 
* Using a reduced set of indexes (essentially doing the distance calculations only on vectors in ok_vocab (corresponding to the restricted_vocab argument). This change is very tempting also - I have modified an experimental similarity measure I have to incorporate this. If my restricted_vocab is 1:10 of the total vocab then I get a speedup of a further factor of 2. However if I run this version with the full vocab then performance actually drops. So the actual performance boost depends on the ratio of restrict_vocab to vocab.size. Finally I found coding my similarity function to support this restricted index idea involved a bit more mental gymnastics which got in the way of understanding the similarity metric itself.

So my suggestion would just be to update most_similar to use matutils.argsort(dists, topn) in place of argsort(dists) and call it a day. I can do a PR but it seems too trivial.

piskvorky:
@KCzar @sebastien-j @seanaedmiston I plan to make a new gensim release this weekend. DO you think we can decide what to do here?

I see three proposals:

1. simple one from @seanaedmiston 
2. more complex #340 and #350 , both need work to finalize (see my comment above)

Sean, can you open your "simple PR"? If I understood correctly, that should be straightforward, so we can merge that independently of #340/#350, and merge those later on top.

Thanks a lot guys!

KCzar:
Hey - sorry I haven't been keeping up on this.  I've gotten behind on some other work and need to focus on that for the moment.  If you can hold the release until Monday, I can spend Saturday & Sunday (maybe some of Friday) trying to get a quality implementation in line with Gensim/PEP standards in time for the release.  I think this would be entirely doable.  

If we need to release it sooner rather than later, I think we could release a version with the simpler but not fully optimized improvements - the first version I submitted included dramatic speed improvements to both most_similar and accuracy with minimally invasive changes to the code, though the speed improvements were not nearly as great as Sebastien showed they can be.

piskvorky:
Sure, thanks @KCzar .

There's no rush -- we can get your & @seanaedmiston 's simple speedups in now, and do a more thorough rewrite for the next release. It's more important we merge something clean and finished, and improve on it later.

seanaedmiston:
I have created PR number #379 with the most minimal (and least performant) improvements to speed up most_similar. It does however give just over a 2x speedup in my tests.

Note that there is one other use of numpy.argsort in the accuracy function that I have not changed to matutils.argsort as it is unclear what a good value of 'topn' should be in this case (and I wanted to keep the PR minimalist.

piskvorky:
@KCzar @sebastien-j  any progress on this?

We'll be making a new release (0.12.1) soon, so perhaps this could go in?

KCzar:
I think for now we should go with the original changes I made, which were very simple and should be easy to add.  What do you need me to do to make that happen?

piskvorky:
OK. Just clean up the PR please -- something that merges cleanly with current gensim, doesn't have unneeded code / dead code paths, and is reasonably well tested. Cheers!

gojomo:
Note that currently, the Doc2Vec helper class DocvecsArray crudely duplicates a bunch of the `most_similar()`code – the same calculations are required amongst doc vectors, now in a separate ndarray. There wasn't an easy way to reuse the Word2Vec path, and I didn't want to do any further refactoring while these improvements were in-process. 

When any improvements here are stable, I'd want to either replicate them on DocvecsArray, or perhaps move all the functions to some single place but amenable to reuse from other instances. (Vector utils functions? Or a mix-in class? I don't know...)

piskvorky:
@KCzar I'll release 0.12.1 now, so let's aim for 0.12.2 with this.

piskvorky:
@KCzar 0.12.2 will be released soon -- any progress on this? Cheers!

piskvorky:
@KCzar ping

KCzar:
Hey - I'll see what I can do this weekend.  I'm hesitant to make
commitments/promises though because I've haven't been able to deliver on
the last few I made.  :/

piskvorky:
@KCzar sorry to bother again, but it would be great if we could get this in!

piskvorky:
Using only `restrict_vocab` matrix implemented in #458.

Closing this now; the other optimizations can be implemented on top of `develop`.

Issue #336, opened 2015-05-07
word2vec (& doc2vec) training doesn't benefit from all CPU cores with high `workers` values


jticknor:
I have been using word2vec with great success on a Macbook Pro, seeing nice increases with additional workers. However, when moving the code over to a MacPro (24 virtual cores), the code actually takes longer with the addition of more workers. When I run "gensim.models.word2vec.FAST_VERSION" I get a value of 0. This same code will give me a value of 1 on the Macbook Pro. Is there something I am missing moving over to the Xenon processors?

piskvorky:
Maybe your input sentence iterator gets saturated? What is the throughput of your sentence stream? Can it sustain that many cores?

jticknor:
I am using the LineSentence iterator that comes with the gensim code. I should have said that the code is slower for any number of cores beyond one using the MacPro. However, I see similar improvements as your blog post using a Macbook Pro (going all the way to 8 virtual cores). These improvements are seen with the same code and datasets. My concern was the output of the FAST_VERSION check being different for each machine.

piskvorky:
Hmm. `FAST_VERSION` reflects the type of BLAS interface return type (single vs. double). Both should be fine, it's an internal detail.

Can you post the output of:

```python
import scipy
scipy.show_config()
```

jticknor:
This is for the MacPro:

atlas_threads_info:
  NOT AVAILABLE
blas_opt_info:
    extra_link_args = ['-Wl,-framework', '-Wl,Accelerate']
    extra_compile_args = ['-msse3', '-I/System/Library/Frameworks/vecLib.framework/Headers']
    define_macros = [('NO_ATLAS_INFO', 3)]
atlas_blas_threads_info:
  NOT AVAILABLE
openblas_info:
  NOT AVAILABLE
lapack_opt_info:
    extra_link_args = ['-Wl,-framework', '-Wl,Accelerate']
    extra_compile_args = ['-msse3']
    define_macros = [('NO_ATLAS_INFO', 3)]
openblas_lapack_info:
  NOT AVAILABLE
atlas_info:
  NOT AVAILABLE
lapack_mkl_info:
  NOT AVAILABLE
blas_mkl_info:
  NOT AVAILABLE
atlas_blas_info:
  NOT AVAILABLE
mkl_info:
  NOT AVAILABLE

thomaskern:
[...]

PR #355, opened 2015-06-06
minimize thread-stalling np.random.randint() calls

gojomo:
Probable fix for #336 - something about the numpy randint() (perhaps its own use of NOGIL) makes calling it once per word (to generate `reduced_windows`) wreak havoc on multiple worker processing.


gensim/models/word2vec_inner.pyx
```diff
@@ -633,6 +632,9 @@ def train_sentence_sg(model, sentence, alpha, _work):
             else:
                 codelens[i] = 1
             result += 1
+    # single randint() call avoids a big thread-sync slowdown
+    for i, item in enumerate(np.random.randint(0, window, sentence_len)):
+        reduced_windows[i] = item
```
piskvorky: Why not just assign the result of `randint(low, high, size)` to `reduced_windows[:size]`? Is there a problem with that? 

Should be faster too.

gojomo: I hadn't even thought of that, being in C-thinking mode when I wrote it (and filling the `cdef np.uint32_t reduced_windows[MAX_SENTENCE_LEN]` array). But I just tried…

`reduced_windows[:sentence_len] = np.random.randint(0, window, sentence_len)`

…and got a couple cython errors. On the other hand, this variant…

`reduced_windows[:sentence_len] = <np.uint32_t *>(np.PyArray_DATA(np.random.randint(0, window, sentence_len)))`

…will compile, so I'm giving it a try now.

piskvorky: If it's too hairy or obscure then never mind. I haven't actually tried / profiled that change, it's just a minor nitpick.

gojomo:
Casting the return ndarray from randint to be the new `reduced_windows` is sufficient and eliminates the loop/copy. (The stack-local preallocation for `reduced_windows` goes away too.)

piskvorky:
Hmm, and when does the array get de-allocated? Is that handled by cython automagically, somehow?

gojomo:
Hmm; the ndarray returned from randint() is a python object, so automatically managed as usual. But I don't know if the reference to the PyArray_DATA value is enough to keep that python object alive... so if GC (and reallocation) is possible before the cython function returns, there might be a risk of the array being clobbered by other uses. I don't yet know enough about CPython GC to know if that's a risk and a few searches haven't yet turned up a sure answer; I'll take a closer look tomorrow.

piskvorky:
In that case, I propose to keep the simple loop for now :)

I'd really like us to focus on finalizing the doc2vec PR.

gojomo:
OK, rolled back other variant, simple approach that's now in this branch/PR should be zero-risk/all-benefits, if anyone's eager to test.

piskvorky:
Merging, thanks @gojomo !

PR #356, opened 2015-06-10
big doc-vector refactor/enhancements

gojomo:
Ready for review/testing!

The headline changes are optimized doc-vector inference, and separating doc-vectors (during training or comparison) from the word-vocabulary – allowing many more docs (via memmap-backing) than words, and avoiding some confusion.

Many smaller changes include a DM-concatenative-context mode (as recommended in the Paragraph Vectors paper), an optional 'lock_factor' to attenuate training of some vectors, and other optimizations/cleanup.

See gensim/test/test_doc2vec.py to do a quick check on a new system or survey some API possibilities. See docs/notebook/doc2vec-IMDB.ipynb for a walkthrough of reproducing the PV paper IMDB sentiment experiment. 

If you used Doc2Vec previously, a few key changes to note: 
* `LabeledSentence` is now `TaggedDocument`, and the (one or more) document labels that correspond to vectors are now referred to as 'tags'. Greatest memory efficiency is possible by using only int tags, contiguous and ascending from 0.
* Doc vectors are stored, accessed, and compared through a consituent '.docvecs' field of the Doc2Vec model, rather than the old accessors/comparison methods that still work for words. So, `d2v_model.docvecs[doc_tag]` or `d2v_model.docvecs.most_similar(doc_tag)`, rather than on `d2v_model` directly.

Some top needs:
* testing on diverse, larger datasets and systems - though the ability to handle doc-vector sets much larger than RAM is theoretically there, I haven't forced an overflow yet
* improving save() to externalize numpy arrays as with the prior models – maybe a recursive utils.SaveLoad?
* refactoring the DocvecsArray similarity-testing methods – currently, just a quick-and-dirty copy-and-adapt from Word2Vec, but would ideally share code (and the optimizations pending in other @sebastien-j /  @KCzar PRs) with word-vecs


gojomo:
A few more thoughts:

After I get a few ack's that the Doc2Vec cython code is working on other diverse systems, I'd like to apply the same code-duplication-reduction refactoring that I did in doc2vec_inner.pyx to word2vec_inner.pyx. At the same time, I'd make the small changes in word2vec[_inner].py[x] that would respect the 'syn0_lockf' values when doing word2vec training. (Right now, even though the field lives in Word2Vec, only the Doc2Vec paths respect it.)

Practical testing/benchmarking on a topical-similarity task is a big need; most of my testing has been on the IMDB-sentiment task, which may not be representative of most uses. I recently noticed this other PV assessment paper:

"Document Embedding with Paragraph Vectors"
https://fb56552f-a-62cb3a1a-s-sites.googlegroups.com/site/deeplearningworkshopnips2014/68.pdf
 
(Some discussion at: http://building-babylon.net/2015/06/03/document-embedding-with-paragraph-vectors/)

They get interesting results for a doc-similarity metric based on a autogenerated dataset of Wikipedia articles & Wikipedia categories. And, as luck would have it, their "simple but effective trick to improve Paragraph Vector" is in fact what Mikolov's minimally-patched word2vec.c already did, and what these Doc2Vec patches also allow with the `dbow_words=1` parameter. So I'm going to try reproducing their results. 

However, the DBOW+Words mode, with a window of N, already takes about N times longer than plain DBOW - so on the 26M word IMDB dataset and a window of 8, ~40s/pass becomes ~300s/pass. On the 5B+ Wikipedia dataset, that could mean ~2 weeks for a 20-pass session. :( 

So I'm thinking of ways to optimize or truncate that same technique. Any other ideas for a small, demo/testing-scale doc-similarity task on public data would be appreciated!

gensim/models/doc2vec.py
```diff
@@ -220,29 +542,113 @@ def _vocab_from(sentences):
...
+            alpha = ((alpha - min_alpha) / (steps - i)) + min_alpha
+
+        return doctag_vectors[0]
 
     def __str__(self):
-        return "Doc2Vec(vocab=%s, size=%s, alpha=%s)" % (len(self.index2word), self.layer1_size, self.alpha)
+        return "Doc2Vec(%id, sg=%i, hs=%i, negative=%i, dm_concat=%i)" % (self.vector_size, self.sg, self.hs, self.negative, self.dm_concat)
+
+    @property
+    def compact_name(self):
```
piskvorky: Interesting idea! Why not simply make this `__str__()`?

gensim/models/doc2vec.py
```diff
@@ -42,48 +42,67 @@
...
     FAST_VERSION = -1
 
-    def train_sentence_dbow(model, sentence, lbls, alpha, work=None, train_words=True, train_lbls=True):
+    def train_sentence_dbow(model, word_vocabs, doctag_indices, alpha, work=None,
+                            train_words=False, learn_doctags=True, learn_words=True, learn_hidden=True,
+                            word_vectors=None, word_locks=None, doctag_vectors=None, doctag_locks=None):
         """
         Update distributed bag of words model by training on a single sentence.
 
         The sentence is a list of Vocab objects (or None, where the corresponding
```
piskvorky: Why the `sentence` rename? +the change should be reflected in the docs too (doc still says "sentence").

gensim/models/doc2vec.py
gojomo: Sure! (I'd then still lead with the class name and abbreviate all else inside the parens.)

gensim/models/doc2vec.py
gojomo: At this point, it's been transformed from what's usually a 'sentence' (either strings or even indexes) into a list of the Vocab objects; I thought it would help to make that clear. I'll update the doc comment soon. 

In other places I've begun to prefer 'document' over 'sentence' just for parallelism with the class name, and to emphasize these are often many-sentences of arbitrary length. Perhaps, these train_sentence_MODE methods should be train_document_MODE instead?

piskvorky: I see. Yes, that makes sense.

And yes, consistency in naming will make the code / API easier to follow :+1:

akhudek:
I just tried this branch on a file that previously failed due to memory problems. It did not go well, though I'm unsure if that is a fault of my OS or the library. I'm running OS X 10.10.3 on a 2.6ghz i7 with 16gb of ram (mac mini). My computer nearly locked up in memory swapping death. I was able to train this same dataset with the C word2vec, so I doubt it's the word vectors causing issues. The data set has 125 million strings with a total of 3.9 billion words.

gojomo:
Thanks for testing, @akhudek ! Note that none of these changes affect plain Word2Vec memory usage, or the Doc2Vec word vocabulary memory costs. You'll still need enough main memory to hold the model.vocab dict (string token -> Vocab objects) and the syn0 and syn1 (/syn1neg) arrays, each taking vector_dimensions*vocabulary_length*4 bytes. 

This PR should be able to help with the doc-vector overhead, but mainly if you (1) use contiguous plain integers, counting up from 0, as document tags (avoiding a tag->object dict with as many entries as your document count); and (2) use the `docvecs_mapfile` parameter to specify a backing file for the giant (vector_dimensions*document_count*4 bytes) docvecs.doctag_syn0 array. 

If you have 125 million documents and (say) 100 dimensions, then no matter how small the word vocabulary, you'll need (125 million * 100 * 4=) 50GB of main memory just for holding the document-vectors in training. (Looking at Mikolov's patched word2vec.c with --sentence-vectors option, I have a hard time seeing how that could complete on a 16GB machine without swapping-hell... but maybe its packed structures are more swap-efficient?)

If you can try the (1) and (2) above, let me know. In any case it'd also be interesting to know what initialization options you're using for your model and the size of effective word vocabulary.

akhudek:
Ah, I assumed that you had built in some sort of disk backing for the document vectors due to the comment "though the ability to handle doc-vector sets much larger than RAM is theoretically there, I haven't forced an overflow yet". When I said that I've used this data set with word2vec, I meant the original, not the one that trains document vectors.

gojomo:
These latest commits wrap up some loose ends:

* DocvecsArray will now write its large ndarrays to separate files as other models do, and theutils.SaveLoad save/load methods have been updated to recognize when an attribute is itself a SaveLoad, and recursively allow it to ignore/separate/restore its attributes
* word2vec_inner.pyx code has been slimmed of redundancies (790 lines reduced to 454) in the same fashion as doc2vec_inner was a while ago
* word2vec's CBOW/sum mode gets the same backprop-error-division as helped word training in doc2vec DM/sum
* word2vec's cython routines all now respect the syn0_lockf array of training-suppression factors

It's been usable for a while and now it's arguably 'complete', so I'd like to merge soon, if only to make it more findable and testable from the 'develop' branch. (I'm OK with anyone pressing that green 'merge pull request' button at any time!)

cscorley:
I've had success in using this pull request for some early work on an information retrieval task (in software engineering research domain). In general, it's working well from both Linux and OS X 10.9.5. Excited to see these changes get merged in. Going to go through and code review now.

gensim/models/doc2vec.py
```diff
@@ -181,68 +496,175 @@ def __init__(self, sentences=None, size=300, alpha=0.025, window=8, min_count=5,
...
+        self.docvecs = docvecs
+        if not self.docvecs:
+            self.docvecs = DocvecsArray(docvecs_mapfile)
+        self.comment = comment
+        if documents is not None:
+            self.build_vocab(documents)
+            self.train(documents)
+
+    def clear_sims(self):
+        Word2Vec.clear_sims(self)
```
cscorley: Is there a reason for avoiding super()?

gensim/models/word2vec.py
```diff
@@ -625,6 +657,53 @@ def load_word2vec_format(cls, fname, fvocab=None, binary=False, norm_only=True):
...
+        counts = None
+        overlap_count = 0
+        logger.info("loading projection weights from %s" % (fname))
+        with utils.smart_open(fname) as fin:
+            header = utils.to_unicode(fin.readline())
+            vocab_size, vector_size = map(int, header.split())  # throws for invalid file format
+            if not vector_size == self.vector_size:
+                logger.error("incompatible vector sizes")
+                # TOCONSIDER: maybe truncation/smaller vectors still useful enough to merge?
+                return # TODO raise ValueError()?
```
cscorley: TODO

gensim/models/doc2vec.py
gojomo: Mere oversight, will remedy.

cscorley: Ok. Note that this pattern exists in a few other places, e.g., the next following couple of methods.

gensim/models/word2vec.py
gojomo: Yes, the TODO idea – raising a ValueError – is likely a better way for this method to fail. 

(There were also a few TODO comments predating any of my edits... so I may yet leave some as pointers to 'ripe spots for future work', rather as than indicators of must-fix/must-do gaps. Unless, of course, that's a disfavored practice on this project, or some other convention should take precedence.)

gojomo:
Thanks for the testing & review! Especially if your use-case is extreme in any interesting dimension – data volume, parameter choices, etc – it'd be great to hear confirmation things are working and useful (or alternatively, any ways that it's hitting limits).

akhudek:
Some more large scale feedback. I tried using docvecs_mapfile, but it still died (in memory death?) right when it got to "reseting layer weights". Also, I didn't see that it ever created the mapfile.

model = Doc2Vec(alpha=0.025, min_alpha=0.025, docvecs_mapfile='mapfile') 

It was strange though, the process was using 4gb, and then suddenly I got the beach ball and couldn't interact with the computer. Activity monitor was still active and reporting only 4gb. Even an ssh session into the machine from another froze.  I suppose it is possible that it called for some huge malloc and the computer went nuts trying to fill it by swapping things, but didn't update the memory usage.

As a sanity check, I just ran gensim's word2vec with size=300 (doc2vec's default) on the same dataset. It starts training successfully using just 3.5-4gb of ram (I didn't wait for it to finish). Doc2vec code is as of today.

piskvorky:
@gojomo you used Python3 for the notebook, right?

I've made some minor py2 and coding style fixes and sent you a pull request. All works fine (OS X, Python 2.7.6) :+1:

gojomo:
@akhudek - the Doc2Vec code does allocate the vocabulary arrays (which can't be practically mem-mapped) first. No memmap file appearing makes it sound like that step is triggering the problem state. But, the fact build_vocab succeeds with your pure Word2Vec test also suggests that memmapped-operation ought to succeed: the Doc2Vec path should only be using a very small amount of extra memory until the point of divergence. 

How big is your vocabulary, before and after min_count trimming? Are you sure the items presented to the Doc2Vec instance return a '.tags' which is a list with plain-integers inside, sparing the growth of the DocvecsArray.doctags dict and DocvecsArray.index2doctag list? Is there a chance that your composition of the (words=[...], tags=[...]) data for presentation to Doc2Vec is consuming a lot of memory? 

(I'm also looking at other ways to reduce memory use and assist config choices. This should probably involve splitting build_vocab into at least two distinct steps. First, the survey-scan, after which the user can see a report of discovered vocabulary and the projected memory use with different min_count/size/etc choices. Then, the actual freezing of choices and allocation of structures - which could actually warn or fail if the demand is larger than main memory.)

gojomo:
@piskvorky the notebook & most other work was done in py3.4/OSX 1st, but I've also regularly run the notebook and other tests on py2.7/ubuntu. Thanks for the PR, merged!

akhudek:
@gojomo The vocabulary phase definitely isn't the problem, I had logging on during the attempt and saw that it completed building the vocabulary with about the same memory usage as word2vec. From memory, (I lost the exact numbers due to the crash, should have logged to a file), there are about 4 million words before pruning, much less afterwards. The crash happened immediately after seeing a log about resetting the layer weights.

The training script is very straightforward, as you can see below. A single line from the input data definitely won't blow memory.

```python
import sys
import gzip
import logging
from gensim.models.doc2vec import TaggedDocument, Doc2Vec

class TaggedLineSentence(object):
   def __init__(self, filename):
      self.filename = filename
 
   def __iter__(self):
      for uid, line in enumerate(gzip.open(self.filename, 'rb')):
         yield TaggedDocument(words=line.split(), tags=[uid])

sentences = TaggedLineSentence(sys.argv[1])

model = Doc2Vec(alpha=0.025, min_alpha=0.025, docvecs_mapfile='mapfile')  # use fixed learning rate
model.build_vocab(sentences)
 
for epoch in range(10):
   print epoch
   model.train(sentences)
   model.alpha -= 0.002  # decrease the learning rate
   model.min_alpha = model.alpha  # fix the learning rate, no decay

# store the model 
model.save(sys.argv[2])
```

gojomo:
Hmm. If build_vocab() completes I think you would see 'mapfile' appear in the working directory: Doc2Vec should have written random initialization vectors across its entire extent. (Do you see the '0' epoch print?)

akhudek:
No, it doesn't make it that far. I've just tried with a small subset of the data. After finishing the vocab with the small data, it's using ~100mb of memory. Somewhere between `resetting layer weights` and starting training, memory usage jumps to 1.3gb. This probably makes sense since OS X is no doubt loading the entire mmap'd file into memory. In this case the mmap'd file does get created.

I'm beginning to suspect that this might be OS X behaving poorly with the mmap'd file, trying to load the entire thing into it's concept of virtual memory resulting in swap file death. I've tried to limit the processes memory via ulimit, but it doesn't seem to work. 

I'm going to try this on a linux system to see if it's OS related.

```
2015-06-28 17:10:58,509 : INFO : collected 143273 word types from a corpus of 31045724 words and 1000000 documents
2015-06-28 17:10:58,571 : INFO : total 29599 word types after removing those with count<5
2015-06-28 17:10:58,572 : INFO : constructing a huffman tree from 29599 words
2015-06-28 17:10:59,766 : INFO : built huffman tree with maximum node depth 23
2015-06-28 17:10:59,795 : INFO : resetting layer weights
0
2015-06-28 17:11:44,514 : INFO : training model with 1 workers on 29599 vocabulary and 300 features, using 'skipgram'=0 'hierarchical softmax'=1 'subsample'=0 and 'negative sampling'=0
2015-06-28 17:11:45,519 : INFO : PROGRESS: at 0.73% words, alpha 0.02500, 225655 words/s
2015-06-28 17:11:46,523 : INFO : PROGRESS: at 1.46% words, alpha 0.02500, 225018 words/s
2015-06-28 17:11:47,532 : INFO : PROGRESS: at 2.19% words, alpha 0.02500, 223860 words/s
```

piskvorky:
Let me merge this PR into `develop`.

We can continue the discussion here, as well as open new PRs for fixes / improvements.

piskvorky:
And it goes without saying -- massive thanks to @gojomo for his epic refactor!

craigpfeifer:
Great changes! When will these be available via pip?

gojomo:
I don't know @piskvorky's plans for a numbered release, but you can always pip install from a github branch. For example, this should do the trick (because 'develop' is the default branch for /piskvorky/gensim):

    pip install git+https://github.com/piskvorky/gensim.git

e9t:
Are you sure you want to totally remove `LabeledSentence`?

Some versioning schemes (ex: [semver](http://semver.org/)), suggest minor version changes to be backwards-compatible, and those who are used to this may have some confusions because this feature merge is a minor version up.
Rather than raising an AttributeError,

    AttributeError: 'module' object has no attribute 'LabeledSentence'

how would it be to use a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning)?

gojomo:
Remaining @cscorley & @e9t suggestions handled on #373 – thanks for review!

e9t:
@gojomo, I got your notebook code to work on my Macbook and reproduced similar results.
Even when I tried using my own dataset, the code ran just fine with sequential integer tags as you suggested. However, when I replaced the tags with strings (I wanted the tags to be actual document IDs), I got a `Segmentation Fault` during the docvec training phase as shown below:

<pre>
# Load corpus
labeledTrainData_clean.tsv
testData_clean.tsv
unlabeledTrainData_clean.tsv
# Set-up Doc2Vec Training & Evaluation Models
Doc2Vec(dm/c,d100,n5,w5,mc2,t8)
Doc2Vec(dbow,d100,n5,mc2,t8)
Doc2Vec(dm/m,d100,n5,w10,mc2,t8)
# Bulk Training
START 2015-07-01 01:20:52.256642
*0.392200 : 1 passes : Doc2Vec(dm/c,d100,n5,w5,mc2,t8) 54.5s 0.5s
*0.392000 : 1 passes : Doc2Vec(dm/c,d100,n5,w5,mc2,t8)_inferred 54.5s 2.1s
Segmentation fault: 11
</pre>

I'm using a Macbook, so this is probably related to https://github.com/numpy/numpy/issues/4007.
Do you think we can fix this?

gojomo:
I wouldn't be confident it's related to the numpy issue – I've not had random/numpy segfaults while developing this on OSX. (All the many segfaults I've worked through have been traceable to my own genuine bugs.)

I didn't think any Macbooks had 8 cores; did you manually set that number of workers? (I'd recommend just the number of real cores, though that shouldn't risk a segfault.)

Does it happen every time, at roughly the same time, when using string tags? And never when using int IDs? 

The string tags should definitely work – they just may prove unwieldy if you get to tens-of-millions of docs. Are your tags relatively short, and just one tag to a document? 

Is there a chance any of your documents are > 10,000 words? (That's currently the hard limit, though I hope to loosen it, and going over it should only cause truncation rather than crashes.)

If you can reliably trigger segfaults with a small test case, I'll have other ideas for details to collect. (First among them: enabling core dumps and checking exactly where the fault is happening.)

e9t:
> I didn't think any Macbooks had 8 cores; did you manually set that number of workers? (I'd recommend just the number of real cores, though that shouldn't risk a segfault.)

My Macbook has a i7 CPU with 4 physical cores but enabling hyperthreading gives me 8 logical cores, which probably is the reason `multiprocessing.cpu_count()` returned 8.
I just manually set the number of workers to 4.

> Does it happen every time, at roughly the same time, when using string tags? And never when using int IDs?

Until now, yes. (The sample isn't so big, but I have run each 5+ consecutive times.)

> The string tags should definitely work – they just may prove unwieldy if you get to tens-of-millions of docs. Are your tags relatively short, and just one tag to a document?

Yes and yes, max length of the tags is 6 and just one tag per document.

> Is there a chance any of your documents are > 10,000 words? (That's currently the hard limit, though I hope to loosen it, and going over it should only cause truncation rather than crashes.)

No, document lengths are all shorter than 10,000 words.

> If you can reliably trigger segfaults with a small test case, I'll have other ideas for details to collect. (First among them: enabling core dumps and checking exactly where the fault is happening.)

I'll see if I can create a toy case.

gojomo:
I'd somehow overlooked the i7 had hyperthreading (usually doing this dev in a VM with 4 virtual cores)... turns out 8 virtual cores do incrementally help, about a 14% speedup over my previous configuration of 4 in a current test. So by all means keep using 8. 

Unless you're doing c/cython stuff, any bug is far more likely in the optimized cython/blas code, which can easily ignore boundaries or hold pointers past memory reuse. So please keep me posted any triggering patterns you find. (There is one other report of a segfault on the discussion list.)

e9t:
You are referring to [this discussion](https://groups.google.com/d/msg/gensim/WvLk1_rySeI/1aqIDnwQDAMJ) right?

I looked at my crash logs and found the same exception type and codes EXC_BAD_ACCESS (SIGSEGV) and KERN_INVALID_ADDRESS.
However the trackback looks a bit different from the previous segfault report:

<pre>
Thread 2 Crashed:
0   libBLAS.dylib                 	0x00007fff9325bc95 cblas_sdot + 976
1   libBLAS.dylib                 	0x00007fff9322ee8d SDOT + 16
2   doc2vec_inner.so              	0x00000001102b71ea __pyx_f_5trunk_6gensim_6models_13doc2vec_inner_our_dot_double + 10 (doc2vec_inner.c:1455)
3   doc2vec_inner.so              	0x00000001102c41fa __pyx_f_5trunk_6gensim_6models_13doc2vec_inner_fast_document_dbow_hs + 202 (doc2vec_inner.c:1678)
4   doc2vec_inner.so              	0x00000001102c3909 __pyx_pw_5trunk_6gensim_6models_13doc2vec_inner_1train_document_dbow + 13769 (doc2vec_inner.c:4102)
5   org.python.python             	0x000000010086091a PyEval_EvalFrameEx + 21166
6   org.python.python             	0x00000001007e39e5 gen_send_ex + 169
7   org.python.python             	0x00000001007c86f1 PyIter_Next + 16
8   org.python.python             	0x0000000100859c31 builtin_sum + 378
9   org.python.python             	0x0000000100860869 PyEval_EvalFrameEx + 20989
10  org.python.python             	0x000000010085b4b5 PyEval_EvalCodeEx + 1622
11  org.python.python             	0x0000000100863b38 fast_function + 321
12  org.python.python             	0x00000001008606fb PyEval_EvalFrameEx + 20623
13  org.python.python             	0x000000010085b4b5 PyEval_EvalCodeEx + 1622
14  org.python.python             	0x00000001007e922f function_call + 372
15  org.python.python             	0x00000001007c8e2a PyObject_Call + 103
16  org.python.python             	0x0000000100860daf PyEval_EvalFrameEx + 22339
17  org.python.python             	0x0000000100863ac2 fast_function + 203
18  org.python.python             	0x00000001008606fb PyEval_EvalFrameEx + 20623
19  org.python.python             	0x0000000100863ac2 fast_function + 203
20  org.python.python             	0x00000001008606fb PyEval_EvalFrameEx + 20623
21  org.python.python             	0x000000010085b4b5 PyEval_EvalCodeEx + 1622
22  org.python.python             	0x00000001007e922f function_call + 372
23  org.python.python             	0x00000001007c8e2a PyObject_Call + 103
24  org.python.python             	0x00000001007da54c method_call + 136
25  org.python.python             	0x00000001007c8e2a PyObject_Call + 103
26  org.python.python             	0x00000001008631c4 PyEval_CallObjectWithKeywords + 93
27  org.python.python             	0x000000010089445b t_bootstrap + 70
28  libsystem_pthread.dylib       	0x00007fff8f102268 _pthread_body + 131
29  libsystem_pthread.dylib       	0x00007fff8f1021e5 _pthread_start + 176
30  libsystem_pthread.dylib       	0x00007fff8f10041d thread_start + 13
</pre>

I've ran the code three times in a row, and the crashes logs all occurred in the same place.
So perhaps numpy was the problem after all?

gojomo:
Yes, that's the discussion. That is a different crash location... but if the bug is some other code, running a little earlier clobbering unintentional addresses with illegal values, the ultimate crash could happen in a variety of places. (However, the other thread reports not seeing the crash again, perhaps specifically since not using the 'sample' parameter... which it doesn't appear you're using at all. So still unclear whether the incidents are related.) 

In your original output, it looked like at least one pass in one training mode (dm/c)  completed, but then the crash occurred during the 1st pass in the second (dbow) mode. So, is that the repeated 'same place' a crash occurs: always that mode when run second? If you put logging to DEBUG does it indicate about the same amount of progress through the data each crash? How about if you only run the DBOW mode – still crashing same proportion of the way through? How about if you leave out DBOW mode entirely – does it crash another place? 

(I'm somewhat skeptical the numpy issue is related – that would seem to necessarily trigger earlier if present.  If you step through the triggering steps outlined in the original gensim-related-report, at https://github.com/piskvorky/gensim/issues/131#issuecomment-27670984, can you get that error?)

gojomo:
@e9t – FYI, I found a way to reliably trigger segfaults locally, and fixed the bug responsible – see the commits about "str doctags trigger bad indexes" on PR #373. Specifically, if both using string doc tags, *and* some tags repeat before all tags are discovered, some tags could be assigned too-high indexes into the vector array (because of a bad assumption of exactly one training example per tag). That'd eventually lead to out-of-bounds accesses or writes. 

Using repeated tags, while not exactly the mode described in the PV paper, is definitely a supported use-case. It might be reasonable to do so to create tag vectors for tags representing metadata/set-membership that many documents share. But also, feeding the training process "[tag-A] 1 2 3 . 4 5 6 ." as one large example, or "[tag-A] 1 2 3 ." then "[tag-A] 4 5 6 ." as two smaller examples, will cause approximately the same training for [tag-A] to occur. (In pure DBOW, the training is essentially identical; in other modes where the 'window' setting comes into play, there will be differences related to when the window reaches across the sentence boundary.)

Can you check if the branch backing that PR resolves your crash?

e9t:
Well, I've enabled DEBUG, and found that the crash does not occur exactly at the "same place".
Running the code three times all crashed during the 1st pass in the third (dm/m) mode, however each at job 3, 7, and 11.

The GOOD NEWS is that yes, that bug fix #380 did the trick!
My code now runs perfectly.

piskvorky:
@gojomo I think this warrants a version bump to 0.12 (rather than just 0.11.2)... what do you think?

gojomo:
:+1: – version increments are free, and features (and API changes) warrant it. (About to make a few small CHANGELOG tweaks.)

vierja:
@gojomo is the `infer_vector` only applicable to a doc2vec model? Or can it be used with word2vec?

gojomo:
It only works for Doc2Vec. (And, if presenting text for inference that contains new words, those words are treated like any other unknown words – dropped before analysis.) 

It's possible a similar mechanism could be offered for inferring word vectors. In many ways the work in #435 supports similar goals.

vierja:
@gojomo thanks for the reply.

I meant like, is it possible to generate a paragraph vector representation using a Word2Vec model, using the same technique/code in `infer_vector`. I have a Word2Vec model, can I use it to infer vectors on a list of paragraphs without training them with Doc2Vec from scratch?

gojomo:
The Doc2Vec algorithms (from the 'Paragraph Vectors' paper) do not start with word vectors, then create doc vectors. Rather, they train doc vectors from text (and only sometimes, in some training methods, generate word vectors as part of that process).

So: no. The inference requires a trained up Doc2Vec model. Having preexisting word vectors isn't a typical/required input for creating doc-vectors (in this algorithm). 

(It's intuitively plausible that seeding some kinds of Doc2Vec models with pre-existing word vectors *might* offer a benefit, but a few small experiments I've done in that direction have had mixed results. There's other research about ways to create sentence/document vectors which do in fact require word vectors first; those algorithms aren't currently in gensim.)

PR #357, opened 2015-06-11
Update doc comment to match implementation

lukaselmer:
Document pageid in extract_pages. It was introduced here: https://github.com/piskvorky/gensim/commit/6783b813408acc4e04ebe0603192c0d76508b048#diff-eece52d95c280dabe57c803c95d6bb96

### DIFF
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -184,7 +184,7 @@ def extract_pages(f, filter_namespaces=False):
     """
     Extract pages from MediaWiki database dump.
 
-    Return an iterable over (str, str) which generates (title, content) pairs.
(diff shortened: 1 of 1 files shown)


gensim/corpora/wikicorpus.py
```diff
@@ -184,7 +184,7 @@ def extract_pages(f, filter_namespaces=False):
     """
     Extract pages from MediaWiki database dump.
 
-    Return an iterable over (str, str) which generates (title, content) pairs.
+    Return an iterable over (str, str, str) which generates (title, content, pageid) pairs.
```
piskvorky: They're not "pairs" anymore then, they're triplets :)

lukaselmer: You're absolutely right! :)

piskvorky:
Merged, thanks @lukaselmer !

Issue #359, opened 2015-06-18
Support custom pickle version in utils.SaveLoad


ccwang002:
I was trying to load a word2vec model saved in Python 3.4 at Python 2.7 environment. Currently it fails to load due to the unrecognized pickle protocol version 4.  

After digging it down, the `Word2Vec.save()` is defined by `utils.SaveLoad` and it always calls the `utils.pickle` with default argument, which pass the protocol version `-1` implying the highest protocol available.

```python
# utils.py
def pickle(obj, fname, protocol=-1):
     """Pickle object `obj` to file `fname`."""
     with smart_open(fname, 'wb') as fout: # 'b' for binary, needed on Windows
         _pickle.dump(obj, fout, protocol=protocol)

class SaveLoad(object):
    def save(self, fname, separately=None, sep_limit=10 * 1024**2, ignore=frozenset()):
        # ...
        pickle(self, fname)
```

To make saved models work on both Python 2.x and 3.x, the pickle protocol version should be set at 2, the highest version Python 2.x. If user can supply their desired protocol version, it would be easier to switch between Python versions.

Naively, I came up with the following api change:

```python
# utils.py
class SaveLoad(object):
    def save(..., pickle_protocol=-1):
         pickle(self, fname, pickle_protocol)  # same for all places calling pickle
```

Rests of the model implementation (subclass of `SaveLoad`) remains unchanged. Take word2vec as example,

```python
# models/word2vec.py
class Word2Vec(utils.SaveLoad):
     # ...
     def save(self, *args, **kwargs):
         kwargs['ignore'] = kwargs.get('ignore', ['syn0norm']) # don't bother storing the cached normalized vectors
         super(Word2Vec, self).save(*args, **kwargs)
     save.__doc__ = utils.SaveLoad.save.__doc__
```

Whether passing `word2vec.save(..., pickle_protocol=2)` or not doesn't break existed code.

Hope this ease the transition between Python 2 and 3. I could file a pull request if this is feasible.

### EDIT: Current solution

In Python 3.4, run the following code before saving will help convert the existed models

```python
import gensim
import pickle as _pickle

def pickle_py2k_compat(obj, fname, protocol=2):  # change the version number you like
    """Pickle object `obj` to file `fname`."""
    print('Saving %s' % fname)
    with gensim.utils.smart_open(fname, 'wb') as fout: # 'b' for binary, needed on Windows
        _pickle.dump(obj, fout, protocol=protocol)

[...]

PR #362, opened 2015-06-21
Set default pickle protocol version to 2

ccwang002:
To make pickled objects compatible across Python 2 and 3, max pickle protocol version should be set at 2.

I'll file another PR for specifying pickle version as option in`utils.SaveLoad.save(...)`, so user can still set their desired pickle protocol.

For full report see issue #359.


piskvorky:
Merged, thank you Liang!

PR #363, opened 2015-06-21
Add pickle protocol as option in utils.SaveLoad

ccwang002:
Users now can specify desired pickle version when saving their models.
Before the change, pickle version is set to the highest version
available, now defaults to 2 to make models compatible across Python 2
and 3.

#### Example
Before:

```python
model = Word2Vec(sentences, size=100, window=5, min_count=5, workers=4)
model.save(fname)  # (mostly) hard-coded by utils.pickle using pickle version 2
```

Also, some operations call `_pickle.dump()` directly which uses different pickle version on Python 2 and 3, that may cause unexpected behavior.

Now:

```python
model.save(fname, pickle_protocol=2)
```

I'm not sure if this API change is this best approach. Any comments are welcomed :)

For full report see issue #359.

### DIFF
--- a/gensim/utils.py
+++ b/gensim/utils.py
@@ -299,7 +299,7 @@ class SaveLoad(object):
         return obj
 
     def _smart_save(self, fname, separately=None, sep_limit=10 * 1024**2,
-                    ignore=frozenset()):
+                    ignore=frozenset(), pickle_protocol=2):
         """
         Save the object to file (also see `load`).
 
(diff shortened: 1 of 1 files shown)


piskvorky:
Looks good to me, thanks! Let's hear from @cscorley and @gojomo to see if there see any potential issues, before merging.

cscorley:
Looks ok to me, as long as load still works for models saved previous to this change (i.e., have protocol != 2).

piskvorky:
Yeah, load is not affected, only save.

Merging, thanks again @ccwang002 !

PR #365, opened 2015-06-22
Online Word2Vec

rutum:
Adding functions:
- update vocab: updates the vocabulary with new words
- update weights: uses the weights of the old vocabulary, and reseting the weights of new vocabulary

Usage: 

model = Word2Vec() # sg and hs are the default parameters
model.build_vocab(sentences)
model.train(sentences)
model.save("base_model")

model.update_vocab(new_sentences)
model.train(new_sentences)
model.save("updated_model")

Then you can compare the 2 models to see whether the new vocabulary is learning the way it is supposed to. 

I tried an experiment with learning a model without "queen", and adding it in the subsequent set of sentences. The updated model learned "queen" as being similar to "king", "duke" etc. So that was a huge success. I would love to hear of any other ideas you might have to test this.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -187,6 +187,7 @@ def train_sg_pair(model, word, word2, alpha, labels, train_w1=True, train_w2=Tru
             model.syn1neg[word_indices] += outer(gb, l1)  # learn hidden -> output
         neu1e += dot(gb, l2b)  # save error
     if train_w2:
+        neu1e += dot(gb, l2b)  # save error
         model.syn0[word2.index] += neu1e  # learn input -> hidden
     return neu1e
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -632,8 +632,10 @@ class Word2Vec(utils.SaveLoad):
         similarity, negative words negatively.
 
         This method computes cosine similarity between a simple mean of the projection
-        weight vectors of the given words, and corresponds to the `word-analogy` and
+        weight vectors of the given words and the vectors for each word in the model. The method corresponds to the `word-analogy` and
         `distance` scripts in the original word2vec implementation.
+        
+        If topn is False, most_similar returns the vector of similarity scores.
 
         Example::
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -978,18 +978,6 @@ class Word2Vec(utils.SaveLoad):
         super(Word2Vec, self).save(*args, **kwargs)
     save.__doc__ = utils.SaveLoad.save.__doc__
 
-class Sentences(object):
-    def __init__(self, filename):
-        self.filename = filename
-
-    def __iter__(self):
-        for line in utils.smart_open(self.filename):
-            line = utils.to_unicode(line)
-            line = line.strip()
-            words = [token.lower() for token in line.split(" ")]
-            if not words:  # don't bother sending out empty sentences
-                continue
-            yield words
 
 class BrownCorpus(object):
     """Iterate over sentences from the Brown corpus (part of NLTK data)."""
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -184,8 +184,9 @@ def train_sg_pair(model, word, word2, alpha, labels, train_w1=True, train_w2=Tru
         fb = 1. / (1. + exp(-dot(l1, l2b.T)))  # propagate hidden -> output
         gb = (labels - fb) * alpha  # vector of error gradients multiplied by the learning rate
         if train_w1:
+            model.syn1neg[word_indices] += outer(gb, l1)  # learn hidden -> output
             neu1e += dot(gb, l2b)  # save error
-    if train_w2 and model.syn0lock[word2.index] == 1:
+    if train_w2:
         model.syn0[word2.index] += neu1e  # learn input -> hidden
     return neu1e
 
@@ -213,7 +214,7 @@ def train_cbow_pair(model, word, word2_indices, l1, alpha, labels, train_w1=True
         if train_w1:
             model.syn1neg[word_indices] += outer(gb, l1)  # learn hidden -> output
         neu1e += dot(gb, l2b)  # save error
-    if train_w2 and model.syn0lock[word2.index] == 1:
+    if train_w2:
         model.syn0[word2_indices] += neu1e  # learn input -> hidden, here for all words in the window separately
     return neu1e
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -187,7 +187,6 @@ def train_sg_pair(model, word, word2, alpha, labels, train_w1=True, train_w2=Tru
             model.syn1neg[word_indices] += outer(gb, l1)  # learn hidden -> output
         neu1e += dot(gb, l2b)  # save error
     if train_w2:
-        neu1e += dot(gb, l2b)  # save error
         model.syn0[word2.index] += neu1e  # learn input -> hidden
     return neu1e
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -164,12 +164,38 @@ except ImportError:
 def train_sg_pair(model, word, word2, alpha, labels, train_w1=True, train_w2=True):
     l1 = model.syn0[word2.index]
     neu1e = zeros(l1.shape)
-
     if model.hs:
         # work on the entire tree at once, to push as much work into numpy's C routines as possible (performance)
         l2a = deepcopy(model.syn1[word.point])  # 2d matrix, codelen x layer1_size
         fa = 1.0 / (1.0 + exp(-dot(l1, l2a.T)))  # propagate hidden -> output
         ga = (1 - word.code - fa) * alpha  # vector of error gradients multiplied by the learning rate
+        if train_w1:
+            model.syn1[word.point] += outer(ga, l1) # learn hidden -> output
+        neu1e += dot(ga, l2a) # save error
+
+    if model.negative:
+        # use this word (label = 1) + `negative` other random words not from this sentence (label = 0)
+        word_indices = [word.index]
+        while len(word_indices) < model.negative + 1:
+            w = model.table[random.randint(model.table.shape[0])]
+            if w != word.index:
+                word_indices.append(w)
+        l2b = model.syn1neg[word_indices] # 2d matrix, k+1 x layer1_size
+        fb = 1. / (1. + exp(-dot(l1, l2b.T))) # propagate hidden -> output
+        gb = (labels - fb) * alpha # vector of error gradients multiplied by the learning rate
+        if train_w1:
+            neu1e += dot(gb, l2b) # save error
+    if train_w2 and model.syn0lock[word2.index] == 1:
+        model.syn0[word2.index] += neu1e # learn input -> hidden
+    return neu1e
+
+
+def train_cbow_pair(model, word, word2_indices, l1, alpha, labels, train_w1=True, train_w2=True):
+    neu1e = zeros(l1.shape)
+    if model.hs:
+        l2a = model.syn1[word.point]  # 2d matrix, codelen x layer1_size
+        fa = 1. / (1. + exp(-dot(l1, l2a.T)))  # propagate hidden -> output
+        ga = (1. - word.code - fa) * alpha  # vector of error gradients multiplied by the learning rate
         if train_w1:
             model.syn1[word.point] += outer(ga, l1)  # learn hidden -> output
         neu1e += dot(ga, l2a)  # save error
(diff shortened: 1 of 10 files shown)


gensim/models/word2vec.py
```diff
@@ -164,15 +164,14 @@ def train_sentence_cbow(model, sentence, alpha, work=None, neu1=None):
...
     if model.hs:
         # work on the entire tree at once, to push as much work into numpy's C routines as possible (performance)
         l2a = deepcopy(model.syn1[word.point])  # 2d matrix, codelen x layer1_size
         fa = 1.0 / (1.0 + exp(-dot(l1, l2a.T)))  # propagate hidden -> output
         ga = (1 - word.code - fa) * alpha  # vector of error gradients multiplied by the learning rate
         if train_w1:
-            model.syn1[word.point] += outer(ga, l1)  # learn hidden -> output
-        neu1e += dot(ga, l2a)  # save error
+            model.syn1[word.point] += outer(ga, l1) # learn hidden -> output
+        neu1e += dot(ga, l2a) # save error
```
piskvorky: At least two spaces before inline comments ([PEP8](https://www.python.org/dev/peps/pep-0008/#inline-comments)).

gensim/models/word2vec.py
```diff
@@ -181,27 +180,25 @@ def train_sg_pair(model, word, word2, alpha, labels, train_w1=True, train_w2=Tru
             w = model.table[random.randint(model.table.shape[0])]
             if w != word.index:
                 word_indices.append(w)
-        l2b = model.syn1neg[word_indices]  # 2d matrix, k+1 x layer1_size
-        fb = 1. / (1. + exp(-dot(l1, l2b.T)))  # propagate hidden -> output
-        gb = (labels - fb) * alpha  # vector of error gradients multiplied by the learning rate
+        l2b = model.syn1neg[word_indices] # 2d matrix, k+1 x layer1_size
```
piskvorky: ditto

gensim/models/word2vec.py
```diff
@@ -181,27 +180,25 @@ def train_sg_pair(model, word, word2, alpha, labels, train_w1=True, train_w2=Tru
...
 
 
 def train_cbow_pair(model, word, word2_indices, l1, alpha, labels, train_w1=True, train_w2=True):
     neu1e = zeros(l1.shape)
-
     if model.hs:
-        l2a = model.syn1[word.point] # 2d matrix, codelen x layer1_size
-        fa = 1. / (1. + exp(-dot(l1, l2a.T))) # propagate hidden -> output
-        ga = (1. - word.code - fa) * alpha # vector of error gradients multiplied by the learning rate
+        l2a = model.syn1[word.point]  # 2d matrix, codelen x layer1_size
```
piskvorky: I see here you add the comment spaces... strange :)

gensim/models/word2vec.py
```diff
@@ -918,6 +977,19 @@ def save(self, *args, **kwargs):
...
     save.__doc__ = utils.SaveLoad.save.__doc__
 
+class Sentences(object):
+    def __init__(self, filename):
+        self.filename = filename
+
+    def __iter__(self):
+        for line in utils.smart_open(self.filename):
+            line = utils.to_unicode(line)
+            token_tags = [t.split('/') for t in line.split() if len(t.split('/')) == 2]
```
piskvorky: What is this class about? `token_tags` doesn't seem to be used in anyway here.

gensim/models/word2vec.py
```diff
@@ -918,6 +977,19 @@ def save(self, *args, **kwargs):
...
+            # ignore words with non-alphabetic tags like ",", "!" etc punctuation, weird stuff)
```
piskvorky: Where are these ignored?

gensim/models/word2vec.py
rutum: I have added the 2 spaces in most of the code where there are inline comments.

gensim/models/word2vec.py
rutum: Argh.. my latest commit didn't go through

piskvorky:
@rutum I had a look at the code -- if I understand correctly, after calling `update_vocab`, training can only continue on **new** words exclusively (not old+new), right?

What is the reason for this? Why not continue training on all vocabulary?

gensim/models/word2vec.py
```diff
@@ -185,23 +184,21 @@ def train_sg_pair(model, word, word2, alpha, labels, train_w1=True, train_w2=Tru
         fb = 1. / (1. + exp(-dot(l1, l2b.T)))  # propagate hidden -> output
         gb = (labels - fb) * alpha  # vector of error gradients multiplied by the learning rate
         if train_w1:
-            model.syn1neg[word_indices] += outer(gb, l1)  # learn hidden -> output
```
piskvorky: Where did the update to `syn1neg` go? I don't understand this refactor. Is it really equivalent?

rutum: It is a bug, and I just fixed it. 
Just FYI, I have made changes only to the skip gram model with hierarchical sampling

rutum:
@piskvorky I have remove the word freeze feature for now. However, the idea is, that if we have a good enough model to start with, we don't want to change it too much because of the introduction of a small amount of new data. Each new learning iteration starts with a default alpha of 0.025, which would make learning very aggressive with new words.

piskvorky:
Ah, I see what you mean. It is an interesting question what should happen to "old" words. Maybe freezing makes sense. Or perhaps we could do some per-word learning rate? Just an idea :)

gensim/models/word2vec.py
```diff
@@ -186,7 +185,7 @@ def train_sg_pair(model, word, word2, alpha, labels, train_w1=True, train_w2=Tru
         gb = (labels - fb) * alpha  # vector of error gradients multiplied by the learning rate
         if train_w1:
             model.syn1neg[word_indices] += outer(gb, l1)  # learn hidden -> output
-        neu1e += dot(gb, l2b)  # save error
+            neu1e += dot(gb, l2b)  # save error
```
piskvorky: Is this also a bug? Careful with the whitespace -- Python is picky :)

gensim/models/word2vec.py
```diff
@@ -918,6 +978,18 @@ def save(self, *args, **kwargs):
...
 
+class Sentences(object):
+    def __init__(self, filename):
+        self.filename = filename
+
+    def __iter__(self):
+        for line in utils.smart_open(self.filename):
+            line = utils.to_unicode(line)
+            line = line.strip()
+            words = [token.lower() for token in line.split(" ")]
```
piskvorky: Probably better to split on all whitespace (not just `" "`).

But then this class looks like a duplicate of the existing `word2vec.LineSentence` class -- what is the difference?

rutum:
Yup! That is the next piece I am working on - using a per word learning rate using [ADAGRAD](http://www.magicbroom.info/Papers/DuchiHaSi10.pdf)

gensim/models/word2vec.py
rutum: The only difference is the removal of newlines etc before and after a sentence. Also, it expects one sentence per line

rutum:
@piskvorky You might be interested in checking out this site: http://rutumulkar.com/blog/2015/word2vec/
I have done some evaluation between the difference in bulk training vs online training. The performance does drop in online training. Your thoughts would be awesome!

gojomo:
It looks like this would work (perhaps already does?) for skip-gram negative-sampling, and could work for CBOW too. 

There's some collision with changes in my pending doc2vec PR – but nothing major, and some changes there make this easier. The The syn0_lockf ('lock-factor') arrays in the bigdocvec PR serve as exactly the word-specific training-dampener (that was also your removed syn0lock) – though only the doc2vec training paths fully respect them. A 1.0 value (the default) means full backpropagated errors are applied, a 0.0 means no-error correction (locked). Whether ADAGRAD or similar would need that full parallel array of factors, or can use some other more local/temporary value, I don't yet understand ADAGRAD enough to say. 

The cost of reallocating-and-copying syn0/syn1/syn1neg each vocab-expansion may be an issue in volume use – and could  be avoided by going to a segmented-representation. That is, syn0 would be a list of ndarrays, rather than one, and a word's coordinate would be 2d rather than 1. (The segments *could* be equal sized – a true 2d ndarray – but a list of ragged-sized segments is probably just as efficient and more flexible.)

Balancing the influence of new-examples and prior-training may be a big factor in the quality of incremental changes. Locking all old vectors in place is one simple, defensible approach – and if the old vectors have already been ported off to downstream applications, where they can't 'drift' to new values without other re-deployment costs, maybe is the dominant approach. But letting the old vectors improve a little, in proportion to how much info about each word the new examples bring, might be optimal...

gensim/models/word2vec.py
piskvorky: That is no different to `LineSentence`. It removes newlines (all whitespace) as well as and expects one sentence per line.

gensim/models/word2vec.py
cscorley: yep, PEP8 compliance!

piskvorky: @cscorley the "strangeness" was in removing such spaces at places, and adding them in others (see comments above).

rutum:
@gojomo Good points. You are correct about the code already working with negative sampling and cbow. In fact, you can definitely see improvement in accuracy *after* adding more vocabulary, when you are using negative sampling, as opposed to hierarchical sampling. I think using hierarchical sampling intuitively doesn't feel right to me, because the binary huffman tree will change with the new frequencies, changing the syn0 to syn1 matrix mappings. 

Anyway, let me know what are the next steps for this!

phdowling:
@rutum @piskvorky Awesome work, I'd be really interested in this feature. What is the status here, will this be included in a release anytime soon?

rutum:
@phdowling We are working on some tests, and then cython implementation before doing the merge

rutum:
I have resolved the merge conflicts with the `develop` branch, but there still seem to be some issues. Any thoughts? @piskvorky

rutum:
Nevermind! Figured it out.

gensim/test/test_word2vec.py
```diff
@@ -44,19 +44,33 @@ def __iter__(self):
...
 
+    def testOnlineLearning(self):
+        """Test that the algorithm is able to add new words to the
+        vocabulary and to a trained model"""
+        model = word2vec.Word2Vec(sentences, min_count=1)
+        model.update_vocab(new_sentences)
+        model.train(new_sentences)
+        self.assertEqual(len(model.vocab), 14)
+        self.models_equal(model, word2vec.Word2Vec.load(datapath(
+            "gensim_word2vec_update.tst")))
```
piskvorky: Where does this file come from? Also, a more direct test of success (things become locked?) may be preferable here, rather than comparing against a pre-generated model.

piskvorky:
@rutum what is the progress here?

I'm thinking what we need now is:

* resolve conflicts, so this PR is mergeable
* I'll try to find someone to help out with cythonization of relevant parts
* merge
* release as part of 0.12.2 (~end of August)
* fix and improve later, depending on reception :)

rutum:
@piskvorky I am working on the merge conflicts right now. The last release has significant changes, and I plan to open up a new PR with this update by next week. I think we should be able to meet the end of August deadline.

piskvorky:
Perfect :)

gensim/test/test_word2vec.py
rutum: @piskvorky Do you have any recommendations on what to test? I am testing the vocabulary length to make sure it has increased.

piskvorky: Well, you can get inspiration from the other surrounding tests :)

But here I meant: where does the file `gensim_word2vec_update.tst` come from? I don't see it in the repository.

And this test should probably be finer anyway -- relying on bit-for-bit equality against a pregenerated model seems to brittle. Any future change to the RNG, or any other parameter, will result in a fail. My guess is this won't work even now, across different Python version -- did you run the tests pass on Python 2.6 & 2.7 & 3.4? Once the conflicts are resolved, Github will launch Travis tests and we will be able to see the test results.

rutum: Agreed. Will change the tests in the new PR with merged changes from develop.

piskvorky:
Superceded by #435 . See there for continued discussion.

sasikum:
AttributeError: 'Word2Vec' object has no attribute 'update_vocab

sasikum:
AttributeError: 'Word2Vec' object has no attribute 'update_vocab'

sasikum:
AttributeError: 'Word2Vec' object has no attribute 'update_vocab'

chez8990:
@sasikum I think update_vocab(sentences) has been replaced by build_vocab(sentences, update=True), this requires the model to be pre-trained.

Zayme249Shaw:
@rutum So how do you use hierarchical sampling in online word2vec right now?I'm interested in the ideas about that

> @gojomo Good points. You are correct about the code already working with negative sampling and cbow. In fact, you can definitely see improvement in accuracy _after_ adding more vocabulary, when you are using negative sampling, as opposed to hierarchical sampling. I think using hierarchical sampling intuitively doesn't feel right to me, because the binary huffman tree will change with the new frequencies, changing the syn0 to syn1 matrix mappings.
> 
> Anyway, let me know what are the next steps for this!

penelope24:
> Adding functions:
> 
> * update vocab: updates the vocabulary with new words
> * update weights: uses the weights of the old vocabulary, and reseting the weights of new vocabulary
> 
> Usage:
> 
> model = Word2Vec() # sg and hs are the default parameters
> model.build_vocab(sentences)
> model.train(sentences)
> model.save("base_model")
> 
> model.update_vocab(new_sentences)
> model.train(new_sentences)
> model.save("updated_model")
> 
> Then you can compare the 2 models to see whether the new vocabulary is learning the way it is supposed to.
> 
> I tried an experiment with learning a model without "queen", and adding it in the subsequent set of sentences. The updated model learned "queen" as being similar to "king", "duke" etc. So that was a huge success. I would love to hear of any other ideas you might have to test this.

that's lovely but how can I use it in my gensim implementation? thx!

Issue #293, opened 2015-02-11
Memory Error - Word2Vec


dav009:
Im running a fairly simple script[1] calling word2vec on a 15G corpus which is already tokenized.I have tried running on a 30G machine, and then on a 60G machine.
Both leading to the following error:

```
Traceback (most recent call last):
  File "word2vec.py", line 13, in <module>
    read_corpus("/mnt/data/corpus")
  File "word2vec.py", line 9, in read_corpus
    model = gensim.models.Word2Vec(sentences, min_count=10, size=500, window=10, sg=1, workers=4)
  File "/usr/local/lib/python2.7/dist-packages/gensim/models/word2vec.py", line 312, in __init__
    self.build_vocab(sentences)
  File "/usr/local/lib/python2.7/dist-packages/gensim/models/word2vec.py", line 414, in build_vocab
    self.reset_weights()
  File "/usr/local/lib/python2.7/dist-packages/gensim/models/word2vec.py", line 524, in reset_weights
    self.syn1 = zeros((len(self.vocab), self.layer1_size), dtype=REAL)
MemoryError
```

Any clues? Does the script contains something which is not meant to be done ?

[1] https://gist.github.com/dav009/fb9a42890d3048b3b745

sebastien-j:
How big is your vocabulary?

If I recall correctly, with a vocabulary of size |V|, the memory usage
should be approximately 8 * size * |V| bytes (plus some overhead).

For |V|=10^7 and size=500, this is 40 GB.

The simplest solution is probably to increase `min_count`.

DortelDuck:
@dav009 , could you find the problem? gensim is also simply failing for me with a Memory Error although plenty of memory is available.

piskvorky:
As @sebastien-j says, it's best to give a more detailed report. A link (gist) to the log of your run would be ideal.

DortelDuck:
The actual problem was that on my cluster node a 32Bit Python (in an Anaconda framework) was installed, so that had nothing to do with gensim.

piskvorky:
@dav009 did you figure out the cause in your case? Was it 32bit Python as well?

Let's try to conclude & close this ticket.

tmylk:
Closing as abandoned.

shirish93:
[...]

PR #370, opened 2015-06-28
Improves peak memory usage of Word2Vec on vocabulary creation

phdowling:
There's no need to keep the `vocab` object after we have `self.vocab`, so that space can be freed. For me, this basically solves #293.


gojomo:
This is a good idea. Wondering: is the `gc.collect()` strictly necessary in your case, or might the `del vocab` be sufficient?

piskvorky:
The idea is nice, but this implementation isn't good.

How about simply turning `vocab` into a list and then `pop`ing from it? That way, it will end up being empty naturally, no `del` or `collect` needed?

phdowling:
I considered the `list.pop()` pattern, I can change it to that if you think it's better - I thought del is a bit more explicit. About the `gc.collect()` call: I'm testing right now to find out if it actually makes a difference, but I expect it probably is not strictly needed.

gojomo:
I've started looking at bigger changes here, including replacing the elastic Vocab objects with more-compact namedtuples, or eliminating them entirely after initial setup, in favor of vocabulary-sized lists/arrays (which are also then more amenable to direct use by the cython routines). 

A possible limitation of peeling entries away (as if by `pop()`) one-by-one from a dict or list: do these structures' capacities shrink-on-removes as fast as they grow-on-inserts?

gojomo:
(Also: efficiency in the vocab-dict handover may be irrelevant, because the next step is allocation of weight arrays that are often much larger than both temp and permanent dicts combined. As long as the temporary surge in memory use is completed/released, its peak shouldn't matter as long as it's still less than the next demand...)

phdowling:
So the `gc.collect()` call in fact does not matter, `del` is enough. For your other point: My situation is that I am building a huge vocabulary, which is then pruned substantially. On my machine I have enough RAM to hold _either_ the larger dict _or_ the smaller dict plus the two matrices built in the next step. Therefore I need to free the larger dicts memory before allocating the weights matrices.

gojomo:
Good to know CPython GC is smart enough to collect before failing the allocation. 

If you're using (or would consider) negative-sampling, see PR #373 for a reimplementation of the sampling-table that may save you 399+MB per model – and speeds sampling as well. 

I tried a couple tactics for minimizing HS memory use, too... but so far changes I thought *might* have a small positive performance effect have instead been causing a surprise, significant decrease in performance, perhaps due to some CPython internals I don't understand, or CPU on-chip cache interactions.

piskvorky:
@gojomo: reallocations for growing and shrinking happens in discrete steps. I think there's a constant multiplier (2x?) = when the number of items in the list is less than half (?) the allocated size, the list is shrunk.

I think the dynamic `pop` solution is preferable, because as `self.vocab` grows, `vocab` gets smaller, so it saves a little more memory. With `del`, we have two full sized lists at the end of the loop, before `del`.

But let's wait for @gojomo's namedtuple / vocab elimination experiments, these sound interesting :)

gojomo:
I'm not sure a key-deletion (even when crossing some low-load threshold) triggers a dict-resize down; if I recall correctly, resize may only happen after a future insert (which in this scenario wouldn't come). 

But also, *any* resize (up or down) allocates a new hashtable, and requires a move/rehash of all surviving items. (That is, even on a triggered hashtable-shrink, there's a temporary burst of new memory demand.) 

So the 'pop one by one' approach may not free memory as smoothly as hoped.. Since the filter-copy is succeeding, and it's easy to de-reference the whole old dict just before the array allocation demand arrives, I think it's a worthwhile mini-fix, even while considering other improvements...

piskvorky:
What hashtable? We're talking about `pop` from a list.

gojomo:
The `vocab` local that's taking space even after all its min_count-surviving entries have been copied to `self.vocab` is a dict.

piskvorky:
Well my original suggestion was using a list instead, and popping from that.

In fact, even the act of using a list itself may be enough, never mind the `pop`.

gojomo:
If using a list to drive what's discarded from the original dictionary, that list may need to be much larger than a second (survivors) dictionary would be. So I believe that whether a discard-list approach needs more or less temporary memory will depend on the relative counts of words that do or don't make the min_count cut. 

However, the discarding-step isn't where @phdowling's memory is exhausted. It's the next step, the big array allocations. I don't see how using a discard-list can leave more free memory, at that next big allocation moment – compared to the one-liner that just discards the old dictionary. (It looks to me like that reaches the optimal endpoint, regardless of whether the interim steps use more or less memory.)

Here's an example of a dict not shrinking as fast as it grows:

    d = {}
    getsizeof(d)  # 288
    d.update(zip(range(10000),range(10000)))
    getsizeof(d)  # 393312
    [_ for _ in range(9000) if d.popitem() is None]
    len(d)  # 1000
    getsizeof(d)  # 393312
    e = {}
    e.update(zip(range(1000),range(1000)))
    len(e)  # 1000
    getsizeof(e)  # 49248

piskvorky:
I think we're talking past each other here.

What me & Philipp were considering is a one-liner, where `vocab` becomes `list(vocab.items())` (or perhaps `_from_vocab()` to return a list directly).

Then, during iteration over `vocab`, we can delete (`pop`) each item, since it's only ever needed once (stream of 2-tuples).

Let me try some pseudo code to avoid misunderstandings:

```python
vocab = list(self._vocab_from(sentences).items())  # temporary dict copy
while vocab:
     word, v = vocab.pop()
     ...
```

piskvorky:
Thanks @phdowling !

I assume you tried this new code and solves your memory issues too? Are we ready to merge?

phdowling:
I did some initial tests and it seems so, but give me another 2h or so to verify. I'll post here once I have results!

piskvorky:
Sure :) Thanks.

phdowling:
Alright! Everything is fine from my side, I've run a test and I can build the vocab just as with the `del` version. I think we can merge.

gojomo:
The pseudocode (and commit) lets me see what Radim meant, but the approach doesn't likely save any peak memory usage over the `del` (or other) approaches, or free any more memory for the critical moment that comes later (array allocation). 

In fact, for the common case (where the surviving vocabulary is much, much smaller than the discovered vocabulary), creating that list is going to be *worse* in its peak memory use. At the moment the list is filled (before any `pop()`s and before the supplied source dict (return val from _vocab_from()) is implicitly dereferenced), the heap of live objects will include both the full list, *and* the full source dict. 

In comparison, the current implementation may not use that much memory up through the moment the creation of the surviving dictionary is populated. (In its peak moment, the full source dict exists, *and*  a survivors dict – that survivors dict often smaller than the temp list would have been.)

Also notable: `pop()`s face some of the same temporary-surge-in memory use on resize-downs that afflict dicts. The array backing a list only shrinks when the list becomes less than half its prior capacity – see <https://github.com/python/cpython/blob/829b49cbd2e4b1d573470da79ca844b730120f3d/Objects/listobject.c#L26>. Since the new smaller array leaves a little headroom for potential future additions, at the moment a list shrinks, it temporarily needs a little more than 1.5x its current footprint.

The true effect of the list approach is that it hides the (necessary) dereference of the temporary dict earlier. If that's aesthically preferred over any explicit `del`, the same effect can be achieved by just eliminating the local `vocab` and putting the call to `_vocab_from` into the `iteritems()`. Then the implicit-dereference happens at the moment the source dict is no longer needed – the end of the loop.

phdowling:
I'll have to trust that what you're saying is accurate - in that case, I think just moving the `_vocab_from` call into the for loop definition is probably the most elegant solution. However, the second part of your first statement seems incorrect to me. I think the memory is definitely freed by the time we get to the matrix allocation step, in fact I'd have to assume it is just from the empirical results. The only problem that really might remain with the impementation as it is now is that it possibly allocates both the list and the dict (which I personally do not know for sure).

Radim, what do you think?

piskvorky:
That's a long analysis, thanks Gordon :)

I think it saves a lot though -- lists are very small and compact, compared to dicts. That's what I meant by *"In fact, even the act of using a list itself may be enough, never mind the pop"*.

These structures only hold object references, so whatever objects are in the list/dict are shared (not made a copy of). It's not like resizing a list is making a deep copy -- only the "slots" are reallocated, and these are relatively small, compared to the list contents.

Probably just bike shedding by now anyway -- both versions seem to work. I like this version better (IMO cleaner, less memory), but I don't feel strongly about it. So, whatever works for you guys.

phdowling:
Well he has a point in that we technically still build the whole dict, so we can't really be saving much space by converting it into a list (I would imagine this gives a temporary spike, but in practice it doesn't seem problematic).

Maybe I'm missing something, but why can't we just `del` individual items from the dict? Surely the dict will shrink _eventually_, at least by the time it's empty, correct? In that case, we definitely save as much memory as with the list, since at one point we have the whole dict allocated anyway. 

In terms of practical use once again though: what's implemented now works for me, and since the really critical step only comes later anyway and this approach seems to free the memory by then, I think this can be left as-is. If the problem does come up again, we can optimize further.

gojomo:
Yes, this is a lot of discussion over a 1-liner either way! But the memory footprints are a tricky subject. 

Yes, a list is usually going be in the range of 1/3 to 1/2 the size of a dict with the same number of entries. But that's no net savings if the temp list must (at outset) hold 100% of the dict, while the built-up fresh-dict might only ever hold 33% or less of the source-dict. That is: 

    dict_of_1,000,000 + list_of_1,000,000 > dict_of_1,000,000 + dict_of_200,000. 

And, neither free up space as gracefully as you might hope as individual items are removed. Shrinks never happen one-item-a-time. They only happen when certain free-capacity thresholds are passed and a full-resize is triggered. It's especially tricky in the dict case: it turns out that dicts not only require a new insert for resizing to be considered, but the insert must cause the live-item-count *plus* the leftover-deleted-entry-placeholder-count to exceed the same load factor that would have also caused a grow. So counterintuitively, removes from a python dict, alone without some number of following inserts, never cause a resize-down.

The resize itself then allocates a new array of the now-needed capacity, and moves everything over. So there's a burst of extra memory needed, often around 1/2 the current capacity, so that less will be used at the end of the resize.

Still, in this case (and many cases), it hardly matters how much temporary memory is used – as long as no allocations fail, being extravagant can be the fast/clear choice. What matters is what lives at the end of the process. And in this case, any of the fixes that make sure the temp dict has been dereferenced are equivalent in that respect. If using the list made the code shorter/clearer, that'd be great – but it shouldn't be thought of as a sure win in either temporary memory use (where it will often be worse) or end-point memory use (where all suggestions that dereference the original dict are equivalent).

Practically, can we hold this aside for a day or so? Some in-progress changes I'm considering for `build_vocab()` – to allow analyzing 'min_count' and 'sample' choices before the cull/freeze/array-allocation – should make the particulars of dict-lifetime in build_vocab very different.

piskvorky:
Yes. Simply pruning the vocabulary periodically (just like in the existing Phrases module, or the original word2vec implementation) is probably the simplest solution, but this is an interesting discussion :)

Re. resizing -- `list` only hold references => resizing overhead is tiny, no deep copy.

Re. dict vs. list size -- I don't know the exact CPython numbers. Are you sure list (a simple serial structure) is only 2x more memory efficient than a dict (=hash table)? That sounds surprising. But then again, the tuples coming from `.items()` inside the list have overhead too...

I think, if it boils down to managing individual object allocations and counting bytes, Python is probably the wrong tool for the job... or we're using the wrong algo. Looking forward to your `build_vocab` changes!

piskvorky:
Superceded by #385, which does memory pruning. @phdowling can you check that branch, see if it works for you?

phdowling:
Hey @piskvorky : That PR is nice and definitely a bug help in terms of memory consumption, but actually the trouble I was having potentially still persists. 

In my local branch, I also made changes that implement early vocab pruning using code almost the same as that in the phrase detection. Still, I was pruning fairly little to not offset the global counts by too much (in the new code, corresponding to a large `max_vocab_size`), but then after all counts were collected, I would prune more aggressively (higher `min_count`) to be able to manage the matrix allocations in the next step. 

In the current implementation, we still keep both the larger and the scaled back vocab in memory, so the matrix allocation may still fail as it previously did in my case - worse, now the larger version is not even collected as it falls out of scope, so it actually always stays in RAM, even though it might still be substantially large. 

Early pruning does not necessarily imply a small vocabulary, though it does in theory enable the user to make it small enough. Arguably though, if the user wants more precise counts, they should only use early pruning as something like a last resort, in which case it would be very beneficial to free up the `self.raw_vocab` space as soon as we no longer need it.

Any thoughts, am I missing something?

piskvorky:
`self.raw_vocab` is small, it's only a string=>int mapping. Even with 20 million entries, it's about 2GB of RAM.

This dynamic pruning will likely affect only low-frequency words, which are then pruned away with min_count anyway, so this shouldn't have any noticeable impact on the final counts.

`self.vocab`, the dict of Vocab objects that consumes a lot of RAM and was the bottleneck, is now constructed only for those tokens needed = for tokens that are >= min_count.

We could throw away `self.raw_vocab` once we have `self.vocab`, yes. But Gordon added a lot of code around decoupling count-aggregation (requires a corpus pass, `scan_vocab`) and min_count pruning (fast, `scale_vocab`, uses the precomputed results from `scan_vocab`). So I kept it in, in case you wanted to re-run `scale_vocab` again, with a different min_count, without having to do another corpus scan. I don't know how important that scenario is.

phdowling:
Ah, I understand. I didn't look closely enough to notice that `self.raw_vocab` does not use `Vocab`objects. I guess it's unlikely that it will actually cause problems this way, plus given the benefit of being able to scale the vocab again I doubt it's worth throwing it out then after all.

piskvorky:
No problem. And of course, an explicit check that this fix works in your code wouldn't hurt :)

PR #373, opened 2015-06-30
smaller&faster neg-sampling table; reduce cython duplication; feedback tweaks

### LINKED ISSUE #367: top-level __init__.py confuses cython; can we remove?
The top-level `__init__.py` seems to confuse cython – which apparently climbs the dir tree to find the 'package' directory. When my gensim clone is named something with a '-' (as is common, eg 'gensim-develop') cython fails with a message like "'gensim-develop.gensim.models.doc2vec_inner' is not a valid module name". My workaround has been to temporarily rename the directory for each compilation. 

But that leads to path cruft in the compiled C, and cython remains confused about the proper project-relative paths to things – which makes sharing definitions/functions (even between the sibling files word2vec_inner & doc2vec_inner) impossible. 

I was baffled by the behavior for a long time, and only after a post to cython-users did someone point me at the `__init__.py` as the issue. (See <https://groups.google.com/d/msg/cython-users/d-3Clm2EkHs/G7CJish_c4kJ>.) After locally removing `__init__.py`, I can skip the workaround and finally cimport a function from word2vec_inner to doc2vec_inner (rather than cut & paste). 

It looks like the file is a [workaround from ~4 years ago](https://github.com/piskvorky/gensim/commits/develop/__init__.py); the commit comment from @dedan says: "fix the module import when linking to the git root instead of module – for some application I need to link to the gensim folder which is also the root of the repository. This script helps python to find the actual sourcecode of the module and had to be changed because radim moved the source within the repo". 

That seems a pretty narrow need that may have expired. (Any thoughts, @dedan?) But I'd also be interested to know if there's another workaround others who've been preparing cython-compiled C files have used...

gojomo:
Replaces 400MB negative-sampling table (which took ~30 seconds to build no matter how small the vocabulary) with a cumulative-distribution table which is proportional to the vocabulary size. For a ~100K word vocabulary, will take 400KB (1000x smaller) and is created almost instantly (making each class's tests finish 3+ minutes faster). Draws use a binary-search into this table; in my tests, sampling has also been *faster* for vocabularies at least up through 250K words (and probably into the millions). 

Uses PXD file to share declarations from word2vec_inner to doc2vec_inner, reducing duplication (fixes #367 ).

Integrates small changes suggested on prior PR #356 from @cscorley and @e9t.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -1,4 +1,0 @@
-# script by dedan: helps him to symlink gensim
-import os
-dirname = __path__[0]		# Package's main folder
-__path__.insert(0, os.path.join(dirname, "gensim"))
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -36,6 +36,7 @@ The model can also be instantiated from an existing file on disk in the word2vec
 
 import logging
 import os
+import warnings
 
 try:
     from queue import Queue
@@ -241,6 +242,11 @@ class TaggedDocument(namedtuple('TaggedDocument','words tags')):
     """
     def __str__(self):
         return '%s(%s, %s)' % (self.__class__.__name__, self.words, self.tags)
+
+# for compatibility
+class LabeledSentence(TaggedDocument):
+    def __init__(self, *args, **kwargs):
+        warnings.warn('LabeledSentence has been replaced by TaggedDocument', DeprecationWarning)
 
 
 class DocvecsArray(utils.SaveLoad):
@@ -510,7 +516,7 @@ class Doc2Vec(Word2Vec):
         doc-vector training; default is 0 (faster training of doc-vectors only).
 
         """
-        Word2Vec.__init__(self, size=size, alpha=alpha, window=window, min_count=min_count,
+        super(Doc2Vec, self).__init__(size=size, alpha=alpha, window=window, min_count=min_count,
                           sample=sample, seed=seed, workers=workers, min_alpha=min_alpha,
                           sg=(1+dm) % 2, hs=hs, negative=negative, cbow_mean=dm_mean, 
                           null_word=dm_concat, **kwargs)
@@ -526,7 +532,7 @@ class Doc2Vec(Word2Vec):
             self.train(documents)
 
     def clear_sims(self):
-        Word2Vec.clear_sims(self)
+        super(Doc2Vec,self).clear_sims()
         self.docvecs.clear_sims()
 
     def reset_weights(self):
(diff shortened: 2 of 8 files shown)


gojomo:
Forgot to mention: this also creates a one-time-seeded numpy.random.RandomState that's local to the model, to isolate the model's RNG from anything else using the (shared global) numpy random instance. And, for seeded_vector, it draws from a deterministically-seeded one-time-use RandomState. (Previous behavior was a bit off – all randoms from the global source, which happened to be re-seeded on each seeded_vector call – but that just happened to work alright in the usual one-model-at-a-time case.)

piskvorky:
Looks great!

How much of the changed functionality is covered by unit tests? Any expected incompatibilities?

gojomo:
No new functionality here: just smaller/faster/clearer implementations – so baseline is, all unit tests still pass (and much faster!) 

Only incompatibility risks would be if someone was reaching into the old bulky neg-sampling table themselves for their own reasons, or was somehow dependent on perfect reproducibility of the prior flaky randomization approach. (Both very unlikely.)

piskvorky:
Great. Let's merge then :+1:

gojomo:
Done, thanks!

piskvorky:
Doc2vec no longer compiles for me now.

The error is:

```
./gensim/models/doc2vec_inner.c:1153:8: error: 'inline' can only appear on functions
static CYTHON_INLINE unsigned PY_LONG_LONG (*__pyx_f_6gensim_6models_14word2vec_inner_bisect_left)(__pyx_t_5numpy_uint32_t *, unsigned PY_LONG_LONG, u...
       ^
./gensim/models/doc2vec_inner.c:174:27: note: expanded from macro 'CYTHON_INLINE'
    #define CYTHON_INLINE __inline__
                          ^
```

(OS X, python 2.7.6, latest `develop` at 3cdc43c8)

piskvorky:
Btw was there a rebase? Normally after merging, the pull request is automatically closed by github. But I see it's still open here..?

piskvorky:
Never mind, must be some weird github caching, it's gone now again.

piskvorky:
Removing the inline statement from `bisect_left` fixes the error. How important is that `inline` for performance?

I'm thinking if it gives trouble to (some) compilers, and the performance penalty is not too big, probably safer to leave it out.

gojomo:
Hmm, I'm on OSX and it's working for me. But, definitely want a setup that works everywhere. Can you try removing the `inline` in the word2vec_inner.pxd file, but leave it in the pyx, and see if that works (or generates a new error)?

(Re: github stale caches - yes, I've even had clones/fetches grab stuff that was hours/days old... but a few retries usually find a fresh version.)

piskvorky:
Yes, pxd is enough. I've pushed the commit to `develop`: 570f08a8 .

PR #379, opened 2015-07-02
Use matutils.argsort in most_similar and most_similar_cosmul

seanaedmiston:
Minimal PR to speed up most_similar (and most_similar_cosmul). Detailed discussion in PR#350

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -70,7 +70,7 @@ except ImportError:
             return numpy.argsort(x)[:topn]
         # numpy >= 1.8 has a fast partial argsort, use that!
         biggest = numpy.argpartition(x, topn)[:topn]
-        return biggest.take(numpy.argsort(x.take(biggest)))  # resort topn into order
+        return biggest.take(numpy.argsort(x.take(biggest))[::-1])  # resort topn into order
 
 
 logger = logging.getLogger("gensim.matutils")
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -748,7 +748,7 @@ class Word2Vec(utils.SaveLoad):
         dists = dot(self.syn0norm, mean)
         if not topn:
             return dists
-        best = argsort(dists)[::-1][:topn + len(all_words)]
+        best = matutils.argsort(dists, topn + len(all_words))[::-1][:topn + len(all_words)]
         # ignore (don't return) words from the input
         result = [(self.index2word[sim], float(dists[sim])) for sim in best if sim not in all_words]
         return result[:topn]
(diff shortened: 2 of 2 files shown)


gensim/matutils.py
```diff
@@ -70,7 +70,7 @@ def argsort(x, topn=None):
             return numpy.argsort(x)[:topn]
         # numpy >= 1.8 has a fast partial argsort, use that!
         biggest = numpy.argpartition(x, topn)[:topn]
-        return biggest.take(numpy.argsort(x.take(biggest)))  # resort topn into order
```
piskvorky: This doesn't look right -- the sign is already changed 4 lines above. Are you sure this works?

gensim/models/word2vec.py
```diff
@@ -748,7 +748,7 @@ def most_similar(self, positive=[], negative=[], topn=10):
         dists = dot(self.syn0norm, mean)
         if not topn:
             return dists
-        best = argsort(dists)[::-1][:topn + len(all_words)]
+        best = matutils.argsort(dists, topn + len(all_words))[::-1][:topn + len(all_words)]
```
piskvorky: `matutils.argsort` already returns the greatest elements... the logic here looks too convoluted.

piskvorky:
In fact, let me change `matutils.argsort` to sort by "lowest first", just like `numpy.argsort`.

Having the two do opposite sorting may be too confusing, let's align the APIs.

seanaedmiston:
Aligned APIs sounds like a good improvement :-)

gensim/matutils.py
seanaedmiston: The -1 is needed to give the expected results when using numpy's argpartition and the below (minimal) changes to most_similar. (I did not test with bottleneck so didn't notice any API incompatability there.)

seanaedmiston:
Hi Radim - I have just discovered that Travis is failing because my change breaks the use of argsort in matutils.full2sparse_clipped. Sorry for not picking this up earlier. Let me know if you want me to address this. Sean

piskvorky:
I've refactored `matutils.argsort` to use API consistent with `numpy.argsort`, plus changed all gensim code to use it everywhere: d2f48bb8

Can you please check `develop` branch now works as intended for you?

The refactor should subsume this PR as well.

seanaedmiston:
Hi Radim,

Confirmed - develop branch now gives me correct looking results for
most_similar with a speed up of just over 2x

Thanks heaps

Sean

piskvorky:
Thank you Sean -- your PR made me realize how confusing the API was, clean it up and use it consistently. Invaluable!

Please let me know if you have other such good ideas :)

PR #380, opened 2015-07-02
Split build_vocab to scan, scale, finalize; train() loop/locking refactor; downsampling into cython

gojomo:
Multiple tweaks for speed/clarity/consistency. 

(1) `build_vocab()` now calls three distinct sub-steps: 
* `scan_vocab()` – discovers all words, but does no discarding
* `scale_vocab()` – applies the `min_count` and `sample` settings, but now has extra logging and a `dry_run` mode that just previews the new vocab counts, effective corpus size, and estimated memory use
* `finalize_vocab()` – considers the vocabulary frozen, builds the binary-tree or sampling-table and allocates model ndarrays

Everything can still be done in one step, or the `build_vocab()` then `train()`, but there's also now the option to `scan_vocab()` first, then look at what's been discovered and see what different `min_count`/`sample` settings would do before settling on particular values, then continue. 

(2) `train()`/`worker_train()` refactored to eliminate cross-thread locking other than via the job_queue and a progress_queue (reporting completed tallies), and move more of the sentence/document handling to worker threads, so there's slightly less chance the master thread could become a bottleneck. Also, train() gets optional keyword parameters for making the queues larger (`queue_factor`) and progress-logging less frequent (`report_delay`). And, using a `workers` value of 0 will now let all work happen in the main thread, which can make debugging or profiling easier. 

(3) frequent-word downsampling and Vocab-instance-lookup has moved into the `train_*` methods (rather than being applied beforehand), and thus also into the cython code, for a slight speedup and increased potential for parallelism. (Thus `train_*` methods now take plain lists-of-strings where they were taking Vocab objects.)


piskvorky:
I've made a quick review, looks reasonable to me, thanks @gojomo !

I especially appreciate the doc and naming fixes -- we've been pushing a lot of new functionality/changes, but I wonder how much of it is discoverable by users from the docs/tutorials.

My only nitpick here is code style: there should always be a space after comma in parameters/tuples etc (`x, y`, not `x,y`).

Do you see any problems with pruning the vocabulary periodically (`_vocab_from`), ala [Phrases](https://github.com/piskvorky/gensim/blob/develop/gensim/models/phrases.py#L154) / original C word2vec?

gojomo:
I sometimes slip on the space-after-commas style when mentally in a 'compact' mode. Will fix before any merge.

But, just found a recipe to trigger segfaults locally – in this branch, but quite possibly a problem in the earlier changes (and perhaps matching the couple of prior segfault reports). It will be a day or so before I can definitely drill into exactly what's going wrong.

gojomo:
Re: pruning vocabulary

That'd be a worthwhile option, but I expect anyone with the RAM would want to put it off until the end, so that their word-counts (and pruning decisions) aren't affected by order-of-tallying.

piskvorky:
Alright, segfaults are serious stuff. Let's postpone the release for until we know more. How difficult / common is it to trigger them?

Thanks for investigating -- I'll refactor to add an option for the periodic vocab prunes in the meanwhile.

gojomo:
I found & fixed the problem: indexes into the docvecs array were being set incorrectly (too large), if and only if using string-doctags *and* some doctags repeated before all were discovered. With changes, I can't reproduce the faults. This likely caused both the user-reported segfaults, as well. 

I made a quick attempt at merging in 'develop' to make this PR apply cleanly – but that triggered a new failure importing from doc2vec_inner (thus fallback to "-1" pure-py implementation), and the reason was not immediately clear. (I'll be able to take a closer look tomorrow.)

gojomo:
The issue yesterday must've been some local slip-up; trying a merge again today went without a hitch.

But, there's still an issue with the LDA tests. All 4 travis builds passed OK after my merge commit, but then after the PEP8 commit, 2 of the 4 failed. So I hit 'restart' on the build box and 3/4 passed. Hit 'restart' again and got lucky: 4/4 passing. So the test looks too fragile to either random-numbers or Python 3 hash/dict-order randomization.

piskvorky:
Yes, not happy with the tests after the recent #281 merge, I'll have to check them again.

Are we good to merge here?

gojomo:
+1 merge this!

piskvorky:
Awesome :)

This means we're good for the new release! I'll look into the flaky unit tests first, and perhaps the min_count pruning, then release 0.11.2 later today.

gojomo:
Looking forward to it! Just noticed another few lines of code for nicer logging I've been running with that wasn't yet in this... see as PR #384.

gojomo:
Let me drop here for future reference a interesting possibility this PR enables, because `scale_vocab()` also returns its internal-sizing-calculations as a dict. Lets say you estimate you can afford to allot 4GB main-memory to your model. Whereas you used to have to pick a likely-suitable `min_count` in advance, with steps like...

    model = Doc2Vec(..., min_count=n)
    model.build_vocab(corpus)
    model.train(corpus)

You should now be able to do the following...

    model = Doc2Vec(...,min_count=1)
    model.scan_vocab(corpus)
    while model.scale_vocab(dry_run=True)['memory'] > (4 * 1024**3):
        model.min_count += 1
    model.scale_vocab()
    model.finalize_vocab()
    model.train(corpus)

...and you'd find the minimum `min_count` that fits in your RAM budget. (Note that the memory-estimate for vocabulary structures is very rough, based on just a small number of my local trials with natural-English corpora. But that size is usually dominated by the exactly-sized ndarrays, so the estimate will be in the right range.)

Of course such constraint-fitting could eventually be moved into `scale_vocab()` itself, but one step at a time...

piskvorky:
Looks cool :)

I think that if this is made robust enough (some memory sizes will depend on 32bit / 64bit python etc), we can even make this the default behaviour. That is, search for the right vocabulary size automatically, given allotted RAM.

I don't think anyone cares for `min_count` directly. It's more a nuisance parameter. As long as each word count is large enough to give the algorithms a fighting chance to learn its embedding, the API should focus on #vocab_size or #RAM, not #min_count. It feels more natural.

PR #384, opened 2015-07-05
Nicer progress-logging during vocab scan

gojomo:
Shows tally of discovered tags (doc vec names/IDs) and words-per-second rate.


piskvorky:
Thanks -- feel free to merge when finished with this PR.

PR #385, opened 2015-07-05
Prune vocab

piskvorky:
Use less memory during word2vec and doc2vec vocabulary building.

Refactored Phrases to use the same function for pruning the count dictionary, which was moved under `utils.prune_vocab`.

The main win comes from keeping raw counts in a plain dict (defaultdict) and only creating the expensive Vocab objects for words with count >= `min_count`.

The additional option of trimming the defaultdict to be <= `max_vocab_size` is likely not even necessary (I set the default `max_vocab_size=10,000,000`, which consumes only about 1GB of RAM).

I also removed the stderr flushes (belongs to logging, not core gensim -- logs can go to files etc, stderr is just incidental) and made some minor PEP8 fixes.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -587,8 +587,8 @@ class Doc2Vec(Word2Vec):
         for document_no, document in enumerate(documents):
             if document_no % progress_per == 0:
                 interval_rate = (total_words - interval_count) / (default_timer() - interval_start)
-                logger.info("PROGRESS: at example #%i, processed %i words (%i/s), %i word types, %i tags" %
-                            (document_no, total_words, interval_rate, len(vocab), len(self.docvecs)))
+                logger.info("PROGRESS: at example #%i, processed %i words (%i/s), %i word types, %i tags",
+                            document_no, sum(itervalues(vocab)) + total_words, interval_rate, len(vocab), len(self.docvecs))
                 interval_start = default_timer()
                 interval_count = total_words
             document_length = len(document.words)
@@ -604,8 +604,8 @@ class Doc2Vec(Word2Vec):
                 min_reduce += 1
 
         total_words += sum(itervalues(vocab))
-        logger.info("collected %i word types and %i unique tags from a corpus of %i examples and %i words" %
-                    (len(vocab), len(self.docvecs), document_no + 1, total_words))
+        logger.info("collected %i word types and %i unique tags from a corpus of %i examples and %i words",
+                    len(vocab), len(self.docvecs), document_no + 1, total_words)
         self.corpus_count = document_no + 1
         self.raw_vocab = vocab
 
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -688,7 +688,7 @@ class Doc2Vec(Word2Vec):
             segments.append('s%g' % self.sample)
         if self.workers > 1:
             segments.append('t%d' % self.workers)
-        return 'Doc2Vec(%s)' % ','.join(segments)
+        return '%s(%s)' % (self.__class__.__name__, ','.join(segments))
 
 
 class TaggedBrownCorpus(object):
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -477,8 +477,8 @@ class Doctag(namedtuple('Doctag', 'index, word_count, doc_count')):
 class Doc2Vec(Word2Vec):
     """Class for training, using and evaluating neural networks described in http://arxiv.org/pdf/1405.4053v2.pdf"""
     def __init__(self, documents=None, size=300, alpha=0.025, window=8, min_count=5,
-                 sample=0, seed=1, workers=1, min_alpha=0.0001, dm=1, hs=1, negative=0,
-                 dbow_words=0, dm_mean=0, dm_concat=0, dm_tag_count=1,
+                 max_vocab_size=10000000, sample=0, seed=1, workers=1, min_alpha=0.0001,
+                 dm=1, hs=1, negative=0, dbow_words=0, dm_mean=0, dm_concat=0, dm_tag_count=1,
                  docvecs=None, docvecs_mapfile=None, comment=None, **kwargs):
         """
         Initialize the model from an iterable of `documents`. Each document is a
@@ -505,6 +505,9 @@ class Doc2Vec(Word2Vec):
 
         `min_count` = ignore all words with total frequency lower than this.
 
+        `max_vocab_size` = limit RAM during vocabulary building; if there are more unique
+        words than this, then prune the infrequent ones. Set to `None` for no limit.
+
         `sample` = threshold for configuring which higher-frequency words are randomly downsampled;
                 default is 0 (off), useful value is 1e-5.
 
@@ -530,10 +533,11 @@ class Doc2Vec(Word2Vec):
         doc-vector training; default is 0 (faster training of doc-vectors only).
 
         """
-        super(Doc2Vec, self).__init__(size=size, alpha=alpha, window=window, min_count=min_count,
-                                      sample=sample, seed=seed, workers=workers, min_alpha=min_alpha,
-                                      sg=(1+dm) % 2, hs=hs, negative=negative, cbow_mean=dm_mean,
-                                      null_word=dm_concat, **kwargs)
+        super(Doc2Vec, self).__init__(
+            size=size, alpha=alpha, window=window, min_count=min_count, max_vocab_size=max_vocab_size,
+            sample=sample, seed=seed, workers=workers, min_alpha=min_alpha,
+            sg=(1+dm) % 2, hs=hs, negative=negative, cbow_mean=dm_mean,
+            null_word=dm_concat, **kwargs)
         self.dbow_words = dbow_words
         self.dm_concat = dm_concat
         self.dm_tag_count = dm_tag_count
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -43,7 +43,7 @@ try:
 except ImportError:
     from Queue import Queue
 
-from collections import namedtuple
+from collections import namedtuple, defaultdict
 from timeit import default_timer
 
 from numpy import zeros, random, sum as np_sum, add as np_add, concatenate, \
@@ -53,7 +53,7 @@ from numpy import zeros, random, sum as np_sum, add as np_add, concatenate, \
 from gensim import utils, matutils  # utility fnc for pickling, common scipy operations etc
 from gensim.models.word2vec import Word2Vec, Vocab, train_cbow_pair, train_sg_pair, train_sentence_sg
 from six.moves import xrange, zip
-from six import string_types, integer_types
+from six import string_types, integer_types, itervalues
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


piskvorky:
@gojomo whaddaya think?

I left the default max vocab size at 10 million (phrases has 40m), so it shouldn't affect "normal" corpora in any way. The pruning only kicks in for corpora with more than 10m distinct word types.

gensim/models/word2vec.py
```diff
@@ -541,19 +552,19 @@ def scale_vocab(self, min_count=None, sample=None, dry_run=False):
             threshold_count = int(sample * (3 + sqrt(5)) / 2)
         downsample_total, downsample_unique = 0, 0
         for w in retain_words:
-            v = self.vocab[w]
-            word_probability = (sqrt(v.count / threshold_count) + 1) * (threshold_count / v.count)
+            v = self.raw_vocab[w]
+            word_probability = (sqrt(v / threshold_count) + 1) * (threshold_count / v)
```
piskvorky: @gojomo looks like this does integer divisions (both `v` and `threshold_count` are integers). I don't remember the algo; is this on purpose?

gojomo: I've previously added "from __future__ import division" to the top for "true division"/Py3-style throughout.

piskvorky: Ah I missed that, thanks/

gensim/models/word2vec.py
```diff
@@ -473,25 +479,26 @@ def build_vocab(self, sentences):
...
-                    vocab[word].count += 1
-                else:
-                    vocab[word] = Vocab(count=1)
+                vocab[word] += 1
+
+            if self.max_vocab_size and len(vocab) > self.max_vocab_size:
+                total_words += utils.prune_vocab(vocab, min_reduce)
+                min_reduce += 1
+
+        total_words += sum(itervalues(vocab))
```
gojomo: Looks like only updating total_words at end might leave progress-report above wrong?

gensim/models/word2vec.py
```diff
@@ -562,14 +573,6 @@ def scale_vocab(self, min_count=None, sample=None, dry_run=False):
...
 
-        if not dry_run:
-            new_vocab = {}
-            for w in retain_words:
-                new_vocab[w] = self.vocab[w]
-                new_vocab[w].index = len(self.index2word)
-                self.index2word.append(w)
-            self.vocab = new_vocab
-            logger.info("vocabulary min_count & sample applied, and indexes assigned")
         return report_values
```
gojomo: I don't see `raw_vocab` being dropped anywhere, which is probably the intent. A separate `raw_vocab` is definitely a win for memory (by deferring Vocab creation) and clarity, but in this approach I might miss the opportunity to apply `scale_vocab()` (with effect) more than once. (Of course, only tightening min_count works but sample could vary either way.) Perhaps non-`dry_run` invoking w/ Vocab-filled vocab already in place should reconstitute a raw_vocab, to allow another effective edit?

gensim/models/word2vec.py
```diff
@@ -327,9 +328,10 @@ class Word2Vec(utils.SaveLoad):
     compatible with the original word2vec implementation via `save_word2vec_format()` and `load_word2vec_format()`.
 
     """
-    def __init__(self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
-                 sample=0, seed=1, workers=1, min_alpha=0.0001, sg=1, hs=1, negative=0,
-                 cbow_mean=0, hashfxn=hash, iter=1, null_word=0):
+    def __init__(
+            self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
+            max_vocab_size=10000000, sample=0, seed=1, workers=1, min_alpha=0.0001,
```
gojomo: I'd prefer a `None` default – it matches prior behavior and avoids correctness surprises like, "word X is missing even though I know it has more than min_count occurences in my corpus!" Meanwhile, choosing to turn it on means you're accepting the compromises it involves.

gensim/models/doc2vec.py
```diff
@@ -585,18 +592,22 @@ def _vocab_from(self, documents, progress_per=10000):
...
-                total_words += 1
-                if word in vocab:
-                    vocab[word].count += 1
-                else:
-                    vocab[word] = Vocab(count=1)
+                vocab[word] += 1
+
+            if self.max_vocab_size and len(vocab) > self.max_vocab_size:
+                total_words += utils.prune_vocab(vocab, min_reduce)
+                min_reduce += 1
```
gojomo: Constantly escalating `min_reduce` is likely overkill, discarding more words than strictly necessary. (I suspect it'd thus also increase the inaccuracy of final results – how different the N picked by this tend to be, compared to the true N most common words.)

piskvorky: This is more or less how the original works, following Zipf's law. What alternative do you propose?

We cannot afford to be calling `prune_vocab` every single time line 602 is hit => need an escalating parameter.

gensim/models/word2vec.py
piskvorky: Yes, I was thinking about that too.

But I think I prefer a sane default that "just works". Maybe I'll increase it to 40m (like in phrases), the defaultdict doesn't take up that much space. Not creating unnecessary Vocab objects was the main memory win.

gensim/models/word2vec.py
piskvorky: Not sure I follow -- why wouldn't you be able to call `scale_vocab` again?

gensim/models/word2vec.py
piskvorky: Ah yes, bug. Fixed in 599b6aea.

piskvorky:
@gojomo the doc2vec unit test seems to [fail](https://travis-ci.org/piskvorky/gensim/jobs/69740166) on

```
FAIL: test_dmc_hs (gensim.test.test_doc2vec.TestDoc2VecModel)
Test DM/concatenate doc2vec training.
  File "/home/travis/build/piskvorky/gensim/gensim/test/test_doc2vec.py", line 176, in test_dmc_hs
    self.model_sanity(model)
  File "/home/travis/build/piskvorky/gensim/gensim/test/test_doc2vec.py", line 126, in model_sanity
    self.assertTrue(fire2 in [match[0] for match in sims])
AssertionError: False is not true
Ran 225 tests in 282.772s
```

when using Python3.4.

I don't think this PR changes anything in that area, so I'll merge it. But we'll have to remove this flakiness, not sure what's going on (maybe multiple documents with the same score, so that fire2 gets pushed past 20?).

piskvorky:
@gojomo the failing test is `test_dmc_hs`, consistently. I wonder if this parameter combination is special w.r.t. `fire2`?

gensim/models/word2vec.py
gojomo: I was assuming that not dropping `raw_vocab` was an oversight. Yes, if it's kept users will even be able to re-scale up. 

But, holding that extra large dict makes out-of-memory errors *more* likely for people who don't use any `min_count` trimming. I could see making this a user-controlled option, perhaps as another visible option on `scale_vocab` (which is where people would be told to look/tinker if hitting vocab-size related issues). My preference would be for the default to be the "old behavior" (a min_count trim, once done, is irreversible – minimizing risks of OOM) but the new retain-raw-frequencies capability is there for experts/code-reviewers to discover.

piskvorky: Alright! I've postponed the release for until we settle this. I will increase the max_vocab_size to 2m and "forget" `raw_vocab` after `scale_vocab`.

And then finally release :)

We can add more options / functionality later of course.

gensim/models/word2vec.py
gojomo: Unfortunately, any default means inaccuracy for those with plenty of memory – and a behavior change from prior versions they might not expect/notice. 

Yet, any default (10m-40m) still leaves problems (1GB-4GB) for those not thinking about memory at all, trying to do too much in 4/8/16GB. They'll hit an allocation error sooner or later, and need a guide to what they can try then.  I'd rather help them with better error messages/docs, than new defaults that power-users might not like. 

As a power-user, I'll always want to run with `max_vocab_size=0`, preferring to see a explicit failure if my vocabularies ever grow unexpectedly large – and then making a conscious adaptation, either in preprocessing or via `max_vocab_size`. So just having to add this new param to my instantiations to get the old behavior will be a minor annoyance.

piskvorky: OK, good point. I'll default it to None before the release.

gensim/models/word2vec.py
piskvorky: Added in commit 7476ed23.

piskvorky:
@gojomo regarding the "minor fixes": does your editor support removing trailing whitespace automatically?

Keeping in trailing whitespace leads to harder-to-read commits, with changes to unrelated lines (such as [this one](https://github.com/piskvorky/gensim/commit/0e6c6475f2c58110313a57437c81fedbf637494d)), so it's better to trim it before committing / on every save. Cheers!

gensim/models/doc2vec.py
gojomo: There's surely a point where escalation is necessary – but only when a prune of the minimum number isn't removing very many words, making another call imminent. In many situations, repeated calls with the same floor will be effective and still practical. 

Especially think of cases where `true_vocab_size < (2 * max_vocab_size)`, and where eliminating all true count=1 words would be enough. A rapidly-escalating floor likely manages just two prunes – good! – but drops a lot more true count=2 words. (And, by creating all that extra headroom, a bunch of true count=1 words, that arrive after the last prune, also survive.) A slower-escalating floor might need ~4-5 prunes, but is much more precise: drops far fewer count=2 words, and lets fewer count=1 words sneak through in the final unpruned batch. 

The first thought that comes to mind is making `min_reduce` escalate if and only if the prune doesn't remove 'enough' words, where 'enough' is something like 1/10th of `max_vocab_size`. Another relevant option might be to apply @mfcabrera's PR #270 count-min-sketch.

gojomo:
Hmm; `test_dmc_hs` is (like some other tests) sensitive to initial randomization but for a long while the results seemed sufficiently stable that it was passing locally on and the build box. (And, I don't see relevant changes since that 'long while'.) I'll keep an eye on it.

gojomo:
I'll look into automating my whitespace-removal. (I've been catching it locally on code when I review `git diff` before commits... but this TXT edit I made via the github web editor.)

gensim/models/doc2vec.py
piskvorky: Or, stop escalating (as much) when reaching `min_count`. Basically linking `min_reduce` to `min_count` -- whatever ends up below `min_count` will get thrown away later anyway.

I should add this is all rather hypothetical: default is no pruning, and when a power users wants very large vocabulary (small `min_count`), they'll have enough RAM for very large `max_vocab_size` as well.

Now that I think of it, `max_vocab_size` may not be necessary at all -- the bottleneck is always `self.vocab`, with its size determined by `min_count`. Size of `self.raw_vocab` only comes into play with 1) super large corpora (tens of millions of distinct word types) AND 2) small RAM implying small `max_vocab_size` AND 3) large `min_count` for a small `self.vocab`. Only then does the pruning from 2) have a change to affect 3)...

gojomo: Agreed that it will be somewhat peculiar for the size of `raw_vocab` to be a blocking issue: except in weird cases, it's smaller than what will come later. And, exactly the corner cases where pruning might help, the user will have to apply more thought – and might want to graduate to extra corpus pre-processing as a distinct step, or some other more-efficient/probabilistic structure.

Farther out, I wonder if something like the count-min-sketch code (possibly still with luxurious amounts of RAM) could enable simultaneous vocabulary and bigram/trigram survey. That might argue for moving the word-survey out of W2V/D2V, where it might be a really simple implementation or very-sophisticated, and W2V/D2V just works with the results. (Have to think about that some more...)

PR #386, opened 2015-07-07
fix CBOW/negative-only regression

gojomo:
I hate when this happens, but: some of the refactoring in #380 turned one word2vec_inner.pyx code path, CBOW+negative-sampling-only, into a training no-op. (It didn't completely clean out an obsolete convention that 'codelens[n] == 0' indicated a word should be skipped.) There was no unit test of pure CBOW+negative-sampling that really verified that meaningful training occurred. 

First commit here includes new tests, one of which should fail to indicate the problem. As soon as I confirm expected fails on build-box builds, I'll add the fix commit.

### DIFF
--- a/docs/notebooks/doc2vec-IMDB.ipynb
+++ b/docs/notebooks/doc2vec-IMDB.ipynb
@@ -393,7 +393,7 @@
       "            best_indicator = '*' \n",
       "        print(\"%s%f : %i passes : %s %ss %ss\" % (best_indicator, err, epoch + 1, name, duration, eval_duration))\n",
       "\n",
-      "        if (epoch % 5) == 0:\n",
+      "        if ((epoch + 1) % 5) == 0 or epoch == 0:\n",
       "            eval_duration = ''\n",
       "            with elapsed_timer() as eval_elapsed:\n",
       "                infer_err, err_count, test_count, predictor = error_rate_for_model(train_model, train_docs, test_docs, infer=True)\n",
@@ -1538,10 +1538,10 @@
       "doc_id = np.random.randint(simple_models[0].docvecs.count)  # pick random doc, re-run cell for more examples\n",
       "model = random.choice(simple_models)  # and a random model\n",
       "sims = model.docvecs.most_similar(doc_id, topn=model.docvecs.count)  # get *all* similar documents\n",
-      "print('TARGET (%d): \u00ab%s\u00bb\\n' % (doc_id, ' '.join(alldocs[doc_id].words)))\n",
-      "print('SIMILAR/DISSIMILAR DOCS PER MODEL %s:\\n' % model)\n",
+      "print(u'TARGET (%d): \u00ab%s\u00bb\\n' % (doc_id, ' '.join(alldocs[doc_id].words)))\n",
+      "print(u'SIMILAR/DISSIMILAR DOCS PER MODEL %s:\\n' % model)\n",
       "for label, index in [('MOST', 0), ('MEDIAN', len(sims)//2), ('LEAST', len(sims) - 1)]:\n",
-      "    print('%s %s: \u00ab%s\u00bb\\n' % (label, sims[index], ' '.join(alldocs[sims[index][0]].words)))\n"
+      "    print(u'%s %s: \u00ab%s\u00bb\\n' % (label, sims[index], ' '.join(alldocs[sims[index][0]].words)))\n"
      ],
      "language": "python",
      "metadata": {},
@@ -1602,8 +1602,8 @@
       "    word = random.choice(word_models[0].index2word)\n",
       "    if word_models[0].vocab[word].count > 10:\n",
       "        break\n",
-      "# or just pick a word from the relevant domain:\n",
-      "word = 'comedy/drama'\n",
+      "# or uncomment below line, to just pick a word from the relevant domain:\n",
+      "#word = 'comedy/drama'\n",
       "similars_per_model = [str(model.most_similar(word, topn=20)).replace('), ','),<br>\\n') for model in word_models]\n",
       "similar_table = (\"<table><tr><th>\" +\n",
       "    \"</th><th>\".join([str(model) for model in word_models]) + \n",
@@ -1721,7 +1721,7 @@
       "# note: this takes many minutes\n",
       "for model in word_models:\n",
       "    sections = model.accuracy('questions-words.txt')\n",
-      "    correct, incorrect = (len(sum((s['correct'] for s in sections), [])), len(sum((s['incorrect'] for s in sections),[])))\n",
+      "    correct, incorrect = len(sections[-1]['correct']), len(sections[-1]['incorrect'])\n",
       "    print('%s: %0.2f%% correct (%d of %d)' % (model, float(correct*100)/(correct+incorrect), correct, correct+incorrect))"
      ],
      "language": "python",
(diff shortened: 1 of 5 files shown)


piskvorky:
OK.

Should we do a bug fix release, 0.12.0-1, when this is resolved? CBOW+neg is a fairly common scenario, I think.

gojomo:
Yes, we definitely want this mode to work for people who just grab the latest official release from 'pip'. 

I only found because I decided to try exactly the data & mode in the word2vec.c 'demo-word.sh' – which uses exactly this parameter combination. I'm still looking into a unexpected big shortfall comapred to word2vec.c in analogy-accuracy rate on that same toy text8 dataset, so please hold off a little on a new release – will update soon.

gojomo:
There's still an unexplained shortfall between the word2vec.c analogy accuracy on text8 (~53%) and the same data and CBOW/negative parameters in gensim (~35%), in my tests. Not yet sure if this is a regression or longstanding discrepancy... will be checking that and other possible explanations later today. 

(Commit 100ae13 fixes another small issue with incompletely-applied refactoring, but that would only affect a less-relevant path, CBOW/sum.)

seanaedmiston:
FWIW - I have recently seen a small discrepency between word2vec.c and
gensim in the 'big-data' case. (Using Mikolov's suggested big-data corpus
of about 8.5B words.) I am able to reproduce his 78% on the word-analogy
task quoted here:
https://code.google.com/p/word2vec/source/browse/trunk/demo-train-big-model-v1.sh
This specifically uses the CBOW/negsamp algorithm. However when I attempt
to reproduce this with Gensim I max out at 75%. Not a big difference
admittedly but one that makes me think that there may be some small long
standing discrepancy....

Happy to run a few extra tests if that helps at all...


On Wed, Jul 8, 2015 at 11:55 PM, Gordon Mohr <notifications@github.com>
wrote:

> There's still an unexplained shortfall between the word2vec.c analogy
> accuracy on text8 (~53%) and the same data and CBOW/negative parameters in
> gensim (~35%), in my tests. Not yet sure if this is a regression or
> longstanding discrepancy... will be checking that and other possible
> explanations later today.
>
> (Commit 100ae13
> <https://github.com/piskvorky/gensim/commit/100ae13a94b703483970d3408f55613ded13f6f6>
> fixes another small issue with incompletely-applied refactoring, but that
> would only affect a less-relevant path, CBOW/sum.)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/piskvorky/gensim/pull/386#issuecomment-119587178>.
>



-- 
Sean Edmiston
----------------------
On twitter: @seanaedmiston <https://twitter.com/#!/seanaedmiston>
Blog: www.edmiston.id.au

piskvorky:
Hmm, that's worrying.

I remember that at the time of porting, the accuracy numbers were the same (in fact, slightly higher for gensim).

Perhaps Tomas made qualitative changes to his the C version later that didn't make it into gensim?

Do you get the same 78% with the original version of C word2vec? (and the same 75% with gensim 0.10?)

I'd like to pinpoint where the divergence happened.

sebastien-j:
A few months ago, I reproduced that 78% result with a custom version of Gensim's word2vec. The changes in my code only added some functionality, without changing the basic algorithms.

I built the phrases using the C code though, and never tried Gensim for this.

gojomo:
Rolling back to June 1 code, I was able to match demo-word.sh accuracy with equivalent gensim parameters – so it's a recent regression, probably something I mis-optimized, and should be easy to locate. (Searching for the offending commit now...)

gojomo:
The root of the shortfall was a mistaken alpha calculation introduced in 626225b, and the calculation was especially far off when using the `iter` parameter. With 204cd4a, fixing the calculation, analogy accuracy rates on the demo-words.sh parameters/data are again similar to word2vec.c. (@seanaedmiston – if your 75% peak on the 'big' data was in the last 9 days, with 626225b, it could be worth another try. If it was earlier, it must be some other issue.)

In the course of testing, also extended LineSentence to enable splitting overlong examples (as in the weird text8 dataset) in the same way word2vec.c does.

gojomo:
There's no further fix needed here; an updated release could be offered at any time. 

By my understanding of PEP440 (and other conventions like Semantic Versioning), it's recommended that any substantive bugfixes to increment the micro/patch number – so "0.12.1" rather than "0.12.0-1".

gojomo:
(Actually just remembered a couple small fixes that could be made to the demo notebook, especially helpful in py2.7; I'll have time to apply them about +8 hours from now...)

gojomo:
Notebook tweaks in, good to go.

piskvorky:
Great, merging! Thanks @gojomo .

Were there any more fixes we wanted to include in this release? Did you find the cause for the failing `fire` vs `fire2` tests?

gojomo:
I don't know any other urgent fixes; I also haven't noticed any failing fire-vs-fire2 tests (in the doc2vec sanity checks) for a while. (Were there some recently I overlooked?)

piskvorky:
Last fail was 4 days ago: https://travis-ci.org/piskvorky/gensim/builds/69754288

(`test_dmc_hs` this time)

OK, I'll release 0.12.1 this weekend -- if you or @cscorley want anything else in, there's still time :)

piskvorky:
... and the latest build as well: [test_cbow_hs fails](https://travis-ci.org/piskvorky/gensim/builds/70347591)

gojomo:
Hmm; I picked the rank-required-to-pass after confirming the probed word's appearance in over 200 different-seeded runs. So it almost seems like the build machine is extra-"lucky" in achieving outlier results. I'll look at loosening the test again, and possibly printing more info about margin-of-failure, in failure-cases...

piskvorky:
The doc2vec tests started [failing again](https://travis-ci.org/piskvorky/gensim/jobs/73551188):

```
FAIL: test_dmc_neg (gensim.test.test_doc2vec.TestDoc2VecModel)
Test DM/concatenate doc2vec training.
  File "/home/travis/build/piskvorky/gensim/gensim/test/test_doc2vec.py", line 202, in test_dmc_neg
    self.model_sanity(model)
  File "/home/travis/build/piskvorky/gensim/gensim/test/test_doc2vec.py", line 128, in model_sanity
    self.assertLess(f2_rank, 30)
AssertionError: 43 not less than 30
```

Not sure if it's related -- can you have a look?

gojomo:
Since that's a "near-miss" (43rd-closest isn't far from the threshold of 30), I don't believe this is either a regression or a matter of much concern – just a 'bad draw' in a test that includes some indeterminacy. So some options are:

* hit re-run, and as long as it passes on a manual retry, wait for the next occurrence to consider extra steps
* increase the tolerance, say to top-50, and wait until that fails (if ever)
* adapt the test to assure full determinism – which might still vary across pythons, unless py3 dict-hash-randomization is also turned off, but we could pick seeds that 'pass for all 4 test setups'. This would be good in most respects but also includes a risk that whatever seeding we pick could just be 'lucky in the passing direction', and thus never trigger on some current or future bug.

piskvorky:
That Travis test is already a re-run (first one was 35 IIRC). In fact, the majority of recent builds have failed: https://travis-ci.org/piskvorky/gensim/builds (even with re-runs).

As long as you think there's no reason to worry, we can increase the threshold to top-50. I defer to your judgement.

Thanks!

piskvorky:
Same error again:
https://travis-ci.org/piskvorky/gensim/jobs/74477416

I'd suggest disabling the failing tests (or the entire failing functionality?) until this is resolved. It makes development and merging on github chaotic.

gojomo:
OK, I just noticed another issue in that test: the ranking test was being done against the leftover inferred-sims (rather than the line-before bulk sims). Fixed in f607d46, which might help. If it doesn't resolve we can change the threshold-rank on the next line (L128) to 50 rather than 30 – all the failures (against the wrong set of sims) have still been near misses, >30 but still <50.

piskvorky:
Great, thanks.

gojomo:
Good news: in 8 re-runs, no failures of the top-30 check across all four configs. 

But, there was a single random failure of the test a few lines down, L139, that the tennis-story should be the odd-one-out compared to two fire-stories. When I have a chance I'll run some local bulk tests to see if throwing more fire-stories into the set can make that more-unlikely.

Issue #387, opened 2015-07-07
the summarization example doesn't pass


Fil:
In PR #324 "Jerry…" has only 6 sentences, and the summarizer rejects it because `INPUT_MIN_LENGTH=10`.

I understand I can do `summarization.summarizer.INPUT_MIN_LENGTH=0`, but I don't like the idea of breaking if one document is too short (a warning might be good enough).

piskvorky:
@fedelopez77 @fbarrios can you have a look?

fbarrios:
@Fil At the moment we are forcing the input text to be longer than 10 sentences because the summary might not make sense otherwise. 
We haven't really considered the option of just raising a warning, I agree now that it's a better alternative. I'll do the changes and make a PR with the fix.

tmylk:
Fixed in https://github.com/piskvorky/gensim/pull/390

PR #390, opened 2015-07-08
Fix issue #387: summarization on short texts

fbarrios:
This fix allows summarizing documents with less than ten sentences, logging a warning in that case.


piskvorky:
Thank you for the quick fix @fbarrios ! Merging :)

PR #393, opened 2015-07-11
Misc fixes for 0.12.1 (ready)

gojomo:
Misc for next release.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -48,7 +48,7 @@ from timeit import default_timer
 
 from numpy import zeros, random, sum as np_sum, add as np_add, concatenate, \
     repeat as np_repeat, array, float32 as REAL, empty, ones, memmap as np_memmap, \
-    sqrt, newaxis, ndarray, dot, vstack, dtype
+    sqrt, newaxis, ndarray, dot, vstack, dtype, divide as np_divide
 
 from gensim import utils, matutils  # utility fnc for pickling, common scipy operations etc
 from gensim.models.word2vec import Word2Vec, Vocab, train_cbow_pair, train_sg_pair, train_sentence_sg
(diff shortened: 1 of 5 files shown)


gojomo:
@piskvorky – these changes are all good-to-go. 

I'm still tracking down one last issue: one of the new Word2Vec tests (test_sg_neg) is failing, but only in the pure-python implementation. I think it's a regression, having a 'git bisect' look for the offending commit now.

piskvorky:
Great, thanks!

Did you find the cause for the failing `fire2` tests?

gojomo:
Changes in 632802e make that fire-docs-near-each-other test more reliable: using a higher initial alpha for the DM/* modes (0.05, in accordance with the original word2vec.c's default for the similar CBOW), and a somewhat higher tolerance for acceptable ranking (top-30 rather than top-20). 

A bunch of tests aren't completely deterministic, though, so there may still be a chance that extremely 'unlucky' (rather than truly buggy) runs show failures – but the failure message for the 'expected in top N' tests will now show the margin-of-failure, which should help guide next steps if/when failures occur.

piskvorky:
OK, thanks Gordon.

Are we ready for the bugfix release, or do you want to wait for that `git bisect` for `test_sg_neg` first?

gojomo:
`bisect` has so far found already-known (and fixed) bugs, so still digging... it'd be good to wait, probably a simple fix at the end.

gojomo:
I've been checking a lot of things, but haven't yet figured out why the pure-python paths, when given the same parameters and number of training cycles on some toy datasets (like the LeeCorpus referenced in the test_word2vec.py tests), aren't reaching the same useful states as the cython-paths. 

Allowing the python paths a larger starting alpha and/or more iterations brings their results closer to expectations. So they "work": there's training happening, and improvement over training cycles. But the implementations are only slightly different, so the results from identical configurations/data shouldn't vary this much. 

As a small example: 

    model = Word2Vec(LeeCorpus(), sg=1, window=4, negative=5, hs=0, min_count=5, workers=2, iter=10)
    print( [word for word, score in model.most_similar('war',topn=len(model.index2word))].index('terrorism') )  # rank of target word

With cython extensions, in a few seconds, that similarity-ranking will usually be top-10 (and almost always be top-40). Disabling the cython extensions, as if by deleting the .so files or forcing the ImportError 'except:' path to be taken atop word2vec.py, makes that training take about 15 minutes (expected) but then report a similarity-ranking hardly distinguishable from random (unexpected). In fact, the rank is often behind hundreds of other words, or even in the bottom-half of the toy-sized 1750 word vocabulary.  

So if the (similar to the above example but not identical) `TestWord2VecModel.test_sg_neg()` test is run in pure-python, it fails. (Right now all other tests succeed even in pure-python, but other modes show a similar discrepancy in the 'tightness' of post-training relationships – just not enough to fail the tests.)

This may not be a concern; all tests only usually run with the cython paths, where they're passing. And any serious users will also be using the cython paths. So maybe this isn't a blocker. 

But the idea the pure-python variants might have some flaw is concerning, even if their main role is as illustrations. I suspect I'm overlooking something obvious.

piskvorky:
Let's release now, and not wait for the `sg_neg` fix. Like you say, hopefully not many users use the pure-Python code path anyway.

gojomo:
I have come back to looking for the reason for this discrepancy, and (again) think I may be close to finding the real reason. So if you can hold 1 more day, I'll update soon.

If I can't figure it out, it need not stop the release - but it'll probably justify a release-note bullet and tracking issue that really emphasizes to people that the cython path is not just faster but better tested and free of (at least) one pending bug.

piskvorky:
Sure, we can postpone till the end of weekend. Thanks for investigating!

gojomo:
Finally found the issue that was causing poorer training on the pure-python path, fixed in #404.

PR #395, opened 2015-07-11
Modified word2vec.py __getitem__() to handle phrases

tpsatish95:
Modified the __getitem__(self,word) method to handle phrases.
ex,
trained_model["office products"] to return the mean vector of "office" and "products"

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1140,6 +1140,10 @@ class Word2Vec(utils.SaveLoad):
           array([ -1.40128313e-02, ...]
 
         """
+        if isinstance(words, string_types):
+            # allow calls like trained_model['office'], as a shorthand for trained_model[['dog']]
+            return self.syn0[self.vocab[words].index]
+            
         return array([self.syn0[self.vocab[word].index] for word in words])
 
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1118,22 +1118,30 @@ class Word2Vec(utils.SaveLoad):
 
     def __getitem__(self, words):
 
-      """
-      Return a numpy array of N word representations in vector space, as a N X VEC_LENGTH 2D numpy array.
+        """
+        Accepts a single word (or) a list of words as input. 
 
-      Example::
+        Returns a word's representations in vector space, as a 1 X VEC_LENGTH 1D numpy array, IF a word is passed as an argument (or)
+        Returns N word's representations in vector space, as a N X VEC_LENGTH 2D numpy array, IF a list of words are passed as an argument.
 
-      >>> trained_model[['office','products']]
+        This is a common usecase of the __getitem__() method to support a list of words too.
+
+        Example::
+
+        >>> trained_model['office']
+        array([ -1.40128313e-02, ...])
+
+        >>> trained_model[['office','products']]
         array([ -1.40128313e-02, ...]
               [ -1.70425311e-03, ...]
-              ...)
+               ...)
 
-      """
-      if isinstance(words, string_types):
-          # allow calls like trained_model['office'], as a shorthand for trained_model[['dog']]
-          return self.syn0[self.vocab[words].index]
+        """
+        if isinstance(words, string_types):
+            # allow calls like trained_model['office'], as a shorthand for trained_model[['office']]
+            return self.syn0[self.vocab[words].index]
 
-      return vstack([self.syn0[self.vocab[word].index] for word in words])
+        return vstack([self.syn0[self.vocab[word].index] for word in words])
 
 
     def __contains__(self, word):
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1116,27 +1116,32 @@ class Word2Vec(utils.SaveLoad):
         dists = dot(vectors, mean)
         return sorted(zip(dists, words))[0][1]
 
-    def __getitem__(self, word):
+    @classmethod
+    def sum(self, vecs):
+        '''
+        To compute the mean of N vectors
+        Example:
+        >>> sum(trained_model['office','products']
+          array([ -1.40128313e-02, ...]
+
+        '''
+        if len(vecs) > 1:
+            return matutils.unitvec(array(vecs).mean(axis=0)).astype(REAL)
+        else:
+            return vecs[0]
+
+    def __getitem__(self, words):
         """
-        Return a word's representations in vector space, as a 1D numpy array.
+        Return a list of N word representations in vector space, as a N D numpy array.
 
         Example::
 
-          >>> trained_model['woman']
+          >>> trained_model['office','products']
           array([ -1.40128313e-02, ...]
 
         """
-        if len(word.split()) > 1:
-            mean = []
-            for w in word.split():
-                if w in self.vocab:
-                    mean.append(self.syn0norm[self.vocab[w].index])
-                else:
-                    raise KeyError("word '%s' not in vocabulary" % w)
-            mean = matutils.unitvec(array(mean).mean(axis=0)).astype(REAL)
-            return mean
-        else:
-            return self.syn0[self.vocab[word].index]
+        return array([self.syn0[self.vocab[word].index] for word in words])
+
 
     def __contains__(self, word):
         return word in self.vocab
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1126,7 +1126,17 @@ class Word2Vec(utils.SaveLoad):
           array([ -1.40128313e-02, ...]
 
         """
-        return self.syn0[self.vocab[word].index]
+        if len(word.split()) > 1:
+            mean = []
+            for w in word.split():
+                if w in self.vocab:
+                    mean.append(self.syn0norm[self.vocab[w].index])
+                else:
+                    raise KeyError("word '%s' not in vocabulary" % w)
+            mean = matutils.unitvec(array(mean).mean(axis=0)).astype(REAL)
+            return mean
+        else:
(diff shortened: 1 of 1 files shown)


phdowling:
Is there a reason mean would be the "correct" way to combine w2v vectors? I usually just sum them. More generally, there are a lot of methods to combining word vectors, including other neural network models. I'm not sure this is really needed, plus users might naturally assume that this is the "proper" way to combine the vectors.

gojomo:
This might make an OK utility method – with a name/comment describing it as one possible/experimental way to guess at a phrase vector. But putting it into the __getitem__ accessor encodes too many assumptions. (Among other things, it would conflict with anyone who's put tokens with whitespace in their model.) 

Expanding __getitem__ with 'advanced indexing'-like options might be nice, so that `sum(model['office', 'products'])` etc would work.

tpsatish95:
I made this patch because this is the naive way used by distance.c in C implementation to handle phrases and predict the top N words/phrases to it. So when i was trying to get that vector representation of a phrase in python, i faced a constraint of only being able to apply words in __getitem()__.
Yeah i accept its not the correct way, but i guess it is a baseline way of handling phrases.

tpsatish95:
I have made a recent patch to make __getitem__ with 'advanced indexing'-like option, so that sum(model['office', 'products']) would work.

tpsatish95:
Sorry, didn't notice that, I guess now `_getitem__` should work fine as before and also be handle word lists at the same time.

piskvorky:
Sorry, I'm -1 on this one.

If I understood your purpose correctly, you can do the same with just `vecs = np.vstack(model[word] for word in words)`, no?

Or `np.mean(np.vstack(model[word] for word in words), axis=0)` for the mean vector.

They're such simple one-liners that I don't think we need to clutter core gensim API with it. Maybe if it was a super-common scenario, but I'm not sure it is.

tpsatish95:
Why I wanted this change to go in, is because, I have been using the distance.c tool in the C implementation that computes the mean of all the words in the phrase entered to get the top N similar words to that phrase. Though calculating the mean of words to get the phrase vector is a naive approach it seems to work fine in the distance.c tool. I have been trying to get this phrase vector that it uses, in the python implementation too, like we do in `__getitem__()` in the case of words.

That is why in my initial commit, i added a couple of lines that returned the same phrase vector that distance.c uses to compute distances when we enter a phrase into `__getitem__()`. But later changed it on request, we could even rollback to that commit.

Moreover, I feel it is not going to clutter the core gensim API, as all I have added is just a couple of lines to make it return the same phrase vector that the distance.c tool uses (in the initial commit). Or in the case of the last commit, it just makes the core API more generic and flexible by accepting a list of words and returning a list of word vectors.

gojomo:
The pseudo-phrase splitting is such a specific policy (with risks of interfering with other uses) that it wouldn't fit in the `__getitem__` indexed-accessor. 

I mentioned accepting things like lists-of-indexes not so much as a request, but as an interesting idea for discussion. It would also happen to make the proposed pseudo-phrase operation more compact –  `mean(model[words.split()])` rather than `mean([model[word] for word in words.split())`. 

I suspect it could help elsewhere, too – the numpy 'advanced indexing' is pretty convenient and intuitive – but it's a small thing either way.

piskvorky:
OK, we can support the `model[sequence_of_wordids]` syntax (`model[["x", "y"]]`, not `model["x", "y"]`). I think it's consistent with other gensim models, which support both `model[single_vector]` and `model[sequence_of_vectors]` (like LSI, LDA...).

The result of `model[list_of_ids]` would be a 2d matrix, while the result of `model[id]` a 1d vector.

I'll have a look at your code & comment there :)

gensim/models/word2vec.py
```diff
@@ -1116,17 +1116,36 @@ def doesnt_match(self, words):
         dists = dot(vectors, mean)
         return sorted(zip(dists, words))[0][1]
 
-    def __getitem__(self, word):
+    @classmethod
```
piskvorky: This `sum` doesn't belong to core gensim -- it's a single `numpy.mean` command, it can't get simpler than that.

gensim/models/word2vec.py
```diff
@@ -1116,17 +1116,36 @@ def doesnt_match(self, words):
...
+    def sum(self, vecs):
+        '''
+        To compute the mean of N vectors
+        Example:
+        >>> sum(trained_model['office','products']
+          array([ -1.40128313e-02, ...]
+
+        '''
+        if len(vecs) > 1:
```
piskvorky: What is the purpose of this `if`? Why should `sum` have different semantics for inputs of length > 1?

Also, empty inputs will result in a hard-to-interpret exception.

gensim/models/word2vec.py
```diff
@@ -1116,17 +1116,36 @@ def doesnt_match(self, words):
...
 
-          >>> trained_model['woman']
+          >>> trained_model['office','products']
           array([ -1.40128313e-02, ...]
 
         """
-        return self.syn0[self.vocab[word].index]
+        if isinstance(words, string_types):
+            # allow calls like trained_model['office'], as a shorthand for trained_model[['dog']]
+            return self.syn0[self.vocab[words].index]
```
piskvorky: Better to do `words = [words`] and share the same code path below. Cleaner and less room for error.

gensim/models/word2vec.py
```diff
@@ -1116,17 +1116,36 @@ def doesnt_match(self, words):
...
+            
+        return array([self.syn0[self.vocab[word].index] for word in words])
```
piskvorky: `numpy.vstack` better (plus need to be careful about the result's `dtype` too -- should be `REAL`, for compatibility).

gensim/models/word2vec.py
tpsatish95: Sorry, Yeah I was so much into the distance tool that I wanted to mimic it in the python version to support phrases. I split my `__getitem__()` function in my initial commit into mean and `__getitem__()`. A new function is not necessary for this one liner `np.mean(model[["office","products"]],axis=0)`, I will remove it in my next commit.

gensim/models/word2vec.py
tpsatish95: It would be clean if we did `words = [words]` BUT it would result in returning a 2D array in all cases then. This `trianed_model["x"] = [[1.22323, ]]`, would be the result then. Which is against the default `__getitem__()`'s functionality. We need to have it the above way to make it return 1D and 2D arrays depending on the input or else it would end up in returning 2D arrays always.

gensim/models/word2vec.py
tpsatish95: Yeah making it `vstack` sounds better, will update it.The dtype is REAL internally, i have not made any changes in code to modify it.

gensim/models/word2vec.py
piskvorky: Aha, yes, that's true.

piskvorky:
Btw we want to release 0.12.1 soon, so this PR will already make it in, hopefully :)

tpsatish95:
That's nice to hear :) 
I have made the changes specified and committed once again.

gensim/models/word2vec.py
```diff
@@ -1116,17 +1116,25 @@ def doesnt_match(self, words):
...
+      Example::
+
+      >>> trained_model[['office','products']]
+        array([ -1.40128313e-02, ...]
+              [ -1.70425311e-03, ...]
+              ...)
+
+      """
+      if isinstance(words, string_types):
+          # allow calls like trained_model['office'], as a shorthand for trained_model[['dog']]
```
piskvorky: This is a common, supported use case -- please mention it explicitly in the docstring too. Also replace `'dog'` by `'office'` here, the comment is a little confusing now.

tpsatish95:
Are the comments fine now ?

piskvorky:
Looks OK, let's wait for @gojomo 's verdict before merging.

piskvorky:
Merging; @gojomo let me know if you see any issues with this. I'll release 0.12.1 tmr (Saturday).

@tpsatish95 : thanks for your patience and code!

gensim/models/word2vec.py
```diff
@@ -1116,17 +1116,33 @@ def doesnt_match(self, words):
...
 
-    def __getitem__(self, word):
+    def __getitem__(self, words):
+
         """
-        Return a word's representations in vector space, as a 1D numpy array.
+        Accepts a single word (or) a list of words as input. 
+
+        Returns a word's representations in vector space, as a 1 X VEC_LENGTH 1D numpy array, IF a word is passed as an argument (or)
+        Returns N word's representations in vector space, as a N X VEC_LENGTH 2D numpy array, IF a list of words are passed as an argument.
```
gojomo: Minor doc comment tweaks: should hard-wrap for easier reading; possessive plural `N words'`; example below (L1134) should have space after comma.

gojomo:
Looks correct and I believe the functionality will be intuitive and useful. 

One more far-out thought to ponder: I know some people are applying word2vec to non-natural-language domains. Their atomic tokens – the 'vocabulary' – may not always be strings, although they can probably always represent as canonical string symbols. But perhaps the the API should be tolerant of non-string 'words' – indeed anything that's hashable (and thus could be a key in the vocab dict) could be a 'word'. I'm thinking especially tuples but maybe other immutable objects as well. This would require a change here and in a few other places that check for a string type. 

Not recommending this change just yet, just planting a seed so we'll recognize the potential generalization if the need comes up...

gensim/models/word2vec.py
piskvorky: I also changed the docs a bit after merge: 4a751935.

PR #404, opened 2015-07-18
Pure-py train_sentence_sg fix (etc)

gojomo:
Fix loop-index mismatch causing corrupted training in the pure-python skip-gram path. (Now, all tests pass with equal tolerances both with and without cython-extensions.) 

Also: fix an issue with the 0-worker mode and match Word2Vec.__getitem__'s new ability to take a list of keys on DocvecsArray, too.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -319,7 +319,19 @@ class DocvecsArray(utils.SaveLoad):
             return i_index
 
     def __getitem__(self, index):
-        return self.doctag_syn0[self._int_index(index)]
+        """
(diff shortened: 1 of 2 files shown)


piskvorky:
Great; feel free to merge.

I'll release 0.12.1 tonight / tmr evening. @cscorley anything else you wanted in?

cscorley:
@piskvorky can't say I have anything code-wise!

gojomo:
The issue that just came up on the list – followup training after reloading a model from an old version – deserves a fix, too. Looking at it now.

PR #405, opened 2015-07-19
upgrade reloaded w2v models; total_examples/total_words for more training

gojomo:
A question on the list highlighted that old (<=0.11.1-1) Word2Vec models, after re-load into 0.12.0, will give errors when attempting continued training. This patches such Word2Vec models on `load()` to have the new expected properties. 

But, there were also some incorrect assumptions in `train()` around progress-calculation (and thus also alpha-updating) when performing any `train()` with data different than that used for vocabulary-building. 

So `train()` has been updated to consistently use a number the caller can supply – either raw word count (`total_words`), or raw example count (`total_examples`) – as the basis for progress & alpha calculations. If the caller supplies neither, the remembered sentence count `corpus_count` from vocabulary-building is used as `total_examples`, but a warning will be generated if that turns out not to match the supplied number of examples.

(Converting old Doc2Vec models on `load()` would be harder to do automatically... so nothing here for that.)

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -555,7 +555,6 @@ class Doc2Vec(Word2Vec):
 
         """
         super(Doc2Vec, self).__init__(
-#        super(self.__class__, self).__init__(
             size=size, alpha=alpha, window=window, min_count=min_count, max_vocab_size=max_vocab_size,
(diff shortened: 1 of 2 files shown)


piskvorky:
All tests pass, merging. Thanks @gojomo !

Issue #400, opened 2015-07-14
word2vec vocab building is not multi-threaded


btym:
I've been running a couple billion words through word2vec, using a 32-core EC2 instance. Actually training the model is blazing fast (relatively), but populating the vocab dict is painfully slow.

I forked and added multi-threading, but max_vocab_size pruning and progress logging were difficult to implement (you can't prune the vocab list of a single worker, since there could be more instances of a word in a different worker's sentences, and pruning after all the vocab counts have been merged defeats the purpose entirely). 

Because I had to disable those two things, I'm not going to submit a pull request. Here's what I changed, perhaps @piskvorky or someone with more time could get vocab size pruning working?

https://github.com/btym/gensim/commit/c45ec90ba79755392ddc33a8844b3a4db576ac96

piskvorky:
Can you say a little more about your use case?

What is the performance now, what performance do you need, and why?

We can certainly optimize this part if there's a compelling reason to :)

Note: threading doesn't generally help much with Python, because of the global interpreter lock (GIL). And it looks like you're putting all sentences into plain lists, in RAM, which wouldn't scale (no streaming).

btym:
It's going through a somewhat large history of github events (the past 7 months, ~1.8 billion words). Memory is not an issue in my case, but using your implementation building the vocab list is almost as slow as training the network, since it's only running on one core out of 32 (to be clear, threads will run properly on separate cores if core affinity is set correctly, but what I added could have been done just as easily with multiprocessing).

I realize the way I'm doing it is specific to my use case (not streaming, no max vocab size), but there's no reason for vocab building not to be muti-thread/multi-process, since it's such a huge performance loss. Streaming could be solved by having one iterator shared between threads and have the iterator handle locking. Not sure how max vocab size would work.

piskvorky:
Thread affinity is not related to GIL. I'm surprised you saw an improvement with multithreading at all -- do you have any numbers?

Anyway, yes, parallelizing the vocab building sounds useful. If it clashes with the max vocab size pruning, I'd rather drop that and keep the parallel version. The parallelization optimization scenario sounds more commonly useful than pruning... although it brings more maintenance headaches (multiprocessing works differently on Windows). Do you have a way of testing on Windows?

If you find a way to build the vocab in parallel, while streaming the input (not everything in RAM), can you please open a PR @btym?

For inspiration, see the actual training a few lines lower, which is also streamed.

btym:
[...]

PR #406, opened 2015-07-21
Parallel scan_vocab

fortiema:
Related to Issue #400 

In Word2Vec.build_vocab, scan_vocab not supports multi-threaded workers, in the same fashion as train.

The compromise is that I had to remove prune_vocab, as its logic is much harder to parallelize. I think this feature is more important, but I guess this is highly debatable.

Each worker temp vocab is a Counter instead of a dict, which eases merging of word counts.

Also modified tests to add multiple workers and different batch sizes.

Currently passing all unit tests, but so far only tested on Linux as other OSes are not easily accessible to me. Maybe someone can do this quicker than me? That would be appreciated.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -546,7 +546,8 @@ class Word2Vec(utils.SaveLoad):
                 while done_jobs < (job_no+1) or not push_done:
                     sentences, work_vocab = progress_queue.get(push_done)  # only block after all jobs pushed
                     sent_count += sentences
-                    vocab = {x: vocab.get(x, 0) + work_vocab.get(x, 0) for x in set(vocab).union(work_vocab)}
+                    for k in work_vocab:
+                        vocab[k] += work_vocab[k]
                     done_jobs += 1
                     elapsed = default_timer() - start
                     if elapsed >= next_report:


fortiema:
Apparently doc2vec had it's own scan_vocab method, but didn't have a build_vocab method, making it use word2vec's scan_vocab. I guess this wasn't the intended behavior so I copied the build_vocab method over to doc2vec, making it use its own scan_vocab.

fortiema:
Apparently, most of the time the test is passing, but occasionally it fails. I'm currently investigating possible race condition preventing vocabulary to fully build.

fortiema:
Intermittently failing test has been fixed, now only problem remaining is support of Counters in Python 2.6

gensim/models/word2vec.py
```diff
@@ -472,39 +473,106 @@ def create_binary_tree(self):
...
 
-    def build_vocab(self, sentences, keep_raw_vocab=False):
+    def build_vocab(self, sentences, keep_raw_vocab=False, scan_size=10000):
         """
         Build vocabulary from a sequence of sentences (can be a once-only generator stream).
         Each sentence must be a list of unicode strings.
 
         """
-        self.scan_vocab(sentences)  # initial survey
+        if isinstance(sentences, tuple):
```
piskvorky: What is this for? Is it necessary?

piskvorky:
Thanks @fortiema !

The changes around doc2vec will need @gojomo's sharp-eyed review.

Regarding Counter: I'm +1 on using the plain `defaultdict`. If I see correctly, the only operation you need is merging counts, which is pretty trivial. No need to bring in an extra backport dependency for that. Dropping 2.6 is not an option.

gensim/models/word2vec.py
fortiema: With the new logic in place, if sentences is a generator (which it shouldn't) it's the best way I found to prevent it, making sure the code is passing the unit test for the generator.

If someone has a better way to deal with that, I'll be more than happy to apply it!

fortiema:
Sure thing for Counter, I'll revert to dict as soon as I have some time tomorrow morning.

gojomo:
What sort of a speedup are you seeing, after these changes, on a large-corpus scan?

fortiema:
@gojomo This is also what I'm wondering, didn't have big enough corpus to bench it correctly today.

Plan to run it on 1B+ corpus tomorrow and get some real numbers, I'll be sure to share them here.

fortiema:
2.6 build also passing now. Will post training results later today.

fortiema:
@gojomo Right now, there doesn't seem to be any significant speedup. I also tried using `multiprocessing.Process` thinking the GIL was the problem, but performance is actually worse (although cpu load is a lot more uniform across cores). I tried with various chunksize, no significant difference.

I need to think more about this and see what prevents this. Any idea why?

gojomo:
In the threading scenario, the GIL is probably the main issue. I don't see any big blocks of work where the GIL would be released, and thus multi-threading possible. (Threads *might* offer some benefit if the 'upstream' source of the token-lists were the result of some costly parsing process that for some stages could release the GIL, or maybe if the text was coming from multiple laggy IO sources, so one thread could proceed while another was awaiting input. But there the multithreading would have to be elsewhere, supporting the iterator.)

In the multiprocessing case, there's bigger communication overhead moving the data read (in the 'master' process) to workers, then back again. That serialize/deserialize 'out' (then serialize/deserialize of the totals back 'in') may be outweighing any benefit from having multiple cores tallying into separate dicts at once. 

*If* the data source were easily/equally split among N files (or file-ranges), then spawning N independent processes to scan each part, then report back the totals, *might* get a noticeable speedup. It'd eliminate the larger 'out' side of the serialize/deserialize mentioned above, and also presumably move any other decompress/tokenization from the one (bottlenecked) master process to N processes (independent until final tally reporting). (Or, it might just reveal the source disk to be the next/real bottleneck.)

Still, that parallelization would require bigger assumptions about the real format of data sources, compared to the current API of "just feed us an iterator [of token lists]". So my hunch is it'd be most appropriate for such format-specific parallelism to live outside the Word2Vec class, as one or more utility classes that are specialized for the data's raw format(s) – compressed/uncompressed, one-file-or-many, etc. There could be support in Word2Vec for receiving (and merging) the summary-tallies of these external utilities.

What's your usual raw corpus format and source disk(s)?

fortiema:
Thanks for the clarification about threads/process and GIL, parallel python is new for me so still trying to get a good understanding of how everything works!

My raw corpus are single-files, 10s of GB in size and sit on normal HDDs. I agree with you on dealing with that problem outside word2vec (or even gensim), as it does not make much sense to implement that kind of input splitting logic in here.

I am also working with Hadoop, and am starting to consider using that instead to build the vocabulary, then feed it back to Word2Vec by putting together a utility class.

piskvorky:
Ha, finally a practical use case for Hadoop's WordCount example? :)
But if you mean vanilla Hadoop, the overhead there is very large, so I doubt you'd see much speed up anyway.

And yes, threading won't help because of GIL (as discussed in the original issue #400).

A good first step could be profiling the current code (single node), to see where the bottlenecks are. There's a line profiler `kernprof` that is very convenient for this.

It's quite possible that for such trivial tasks, (de)serialization is too slow. Then the solution would be to go Cython and use threads.

fortiema:
@piskvorky Thanks for the pointers, will try to profile code and see where the culprit is!

fortiema:
@gojomo What about  `doc2vec` missing a build_vocab and as such skipping its own implementation of scan_vocab. Should I submit a separate PR request for that as it seems to be a bug indeed?

gojomo:
Yes, my interpretation of that profiling data is that most of the time is spent in the `enumerate(sentences)` work – which would include raw IO, decompression, and (I suspect) character-coding and splitting-on-whitespace. Doing that on (say) 8 equally-sized part files (or slightly harder, non-overlapping ranges) on an 8-core machine might offer a significant speedup. Though, if the disk rather than the computation is the real blocker and all ranges are on the same disk, maybe not any speedup at all, until the ranges are moved to independent/faster disks. (Is a CPU core utilization totally pegged with the current implementation? If not, disk may already be the bottleneck.)

Even if the for-loop – looking-up-words and incrementing-their-counts – took zero time the speedup would be only 30%. There's probably some potential win there from the compactness and lesser-indirection of C/cython, but I can't guess how much. (If going down that road, you'd likely want to do the whole process, starting with the IO/encoding/splitting, so there's never python strings/lists in the mix. The needed raw result from the scan is so simple – list of words and their frequencies – it could be a separate tool.)

If just generally wanting a 10s-of-GB scan to go faster, my main thoughts would be: (1) SSD; (2) split to N (= number of cores) parts (ideally on different devices), run all parts in parallel, merge final results.

`Doc2Vec` shouldn't have to override `build_vocab'; on a Doc2Vec instance, the (inherited) `build_vocab`'s call to `scan_vocab` should use the instance's own implementation. And that's the behavior I'm seeing locally – progress updates have the 'tags' count characteristic of the `Doc2Vec.scan_vocab` implementation. Are you sure you've seen otherwise? (And if so, might it have been the result of some local weirdness, a typo or class-auto-reload thwarting usual behavior?)

fortiema:
@gojomo You are right about `Doc2Vec`, this change was only necessary after I modified `Word2Vec.scan_vocab` and changed its signature, which wasn't matching the one in `Doc2Vec` anymore. My bad!

I will investigate further and try to come up with a simple and clean way to speedup vocab building. Not so straightforward after all :-p

piskvorky:
Yes, the loop is fairly easy to cythonize (unless you're on Windows, where compilation is generally weird). It's essentially the exact same python code, just compiled to C auto-magically using Cython.

I mean, if you've never used Cython before, there's some learning curve in "what file goes where, how do I write it, how do I load it", but it's [really simple](http://docs.cython.org/src/tutorial/cython_tutorial.html) in this case.

And even if the Cython code is without all the robustness bells and whistles (=not ready to merge), it'd be an interesting comparison point against the pure Python version. Since the bottleneck seems to be file iteration itself, it may also be worth cythonizing your iterator (not just the word count for-loop here). The lower bound here is obviously your disk speed: if your disk can read 50MB/s, and you have 5GB of disk data, then 100 seconds is the best you can do, no matter the language :)

fortiema:
Okay, I just took quick measurements with basic cython code (no crazy optimization):

|| Setup || 1M sent. || 2M sent. ||
| Full Python | 12.7s | 24.0s |
| Cython scan_vocab | 11.8s | 23.2s |
| Cython scan_vobab + iterator | 11.0s | 21.8s |

Not the most rigorous test there is, but we can see ~10% speedup with both the scan_vocab and the iterator cythonized.

piskvorky:
Thanks a lot for investigating!

Yeah, 10% doesn't sound worth it. Can you share your code so we can check as well?

fortiema:
I guess I could open up a new PR since I started from scratch on 'develop', but let me just paste the code here verbrose:

*Added to word2vec_inner.pyx:*
```python
def scan_vocab(model, sentences):
    logger.info("using cythonized wordcount")
    logger.info("collecting all words and their counts")
    cdef int sentence_no = -1
    cdef int total_words = 0
    cdef int min_reduce = 1
    vocab = dict()
        if sentence_no % 10000 == 0:
            logger.info("PROGRESS: at sentence #%i, encountered %i unique words", sentence_no, len(vocab))

        for word in sentence:
            if word not in vocab:
                vocab[word] = ONE
            else:
                vocab[word] += ONE

        if model.max_vocab_size and len(vocab) > model.max_vocab_size:
                total_words += utils.prune_vocab(vocab, min_reduce)
                min_reduce += 1

    logger.info("PROGRESS: at sentence #%i, encountered %i unique words", sentence_no, len(vocab))
    model.corpus_count = sentence_no + 1
    model.raw_vocab = vocab
```

*Beginning of word2vec.py now looks like this:*
```python
try:
    from gensim.models.word2vec_inner import scan_vocab, train_sentence_sg, train_sentence_cbow, FAST_VERSION
except ImportError:
    # failed... fall back to plain numpy (20-80x slower training than the above)
    FAST_VERSION = -1

    def scan_vocab(model, sentences, progress_per=10000):
        """Do an initial scan of all words appearing in sentences."""

        if FAST_VERSION < 0:
            import warnings
            warnings.warn("C extension not loaded for Word2Vec, wordcount will be slow. "
                          "Install a C compiler and reinstall gensim for fast worccount.")

        print "PYTHON VERSION"

        logger.info("collecting all words and their counts")
        sentence_no = -1
        total_words = 0
        min_reduce = 1
        vocab = defaultdict(int)
                logger.info("PROGRESS: at sentence #%i, processed %i words, keeping %i word types",
                            sentence_no, sum(itervalues(vocab)) + total_words, len(vocab))
            for word in sentence:
                vocab[word] += 1

            if model.max_vocab_size and len(vocab) > model.max_vocab_size:
                total_words += utils.prune_vocab(vocab, min_reduce)
                min_reduce += 1

        total_words += sum(itervalues(vocab))
        logger.info("collected %i word types from a corpus of %i raw words and %i sentences",
                    len(vocab), total_words, sentence_no + 1)
        model.corpus_count = sentence_no + 1
        model.raw_vocab = vocab
...
```

piskvorky:
Thanks again @fortiema .

How large (in MB) are these three testing corpora?

I'll try to profile too when I get some time, because Python's IO is actually pretty fast, so with cythonization added, I'd expect the performance to be closer to HW limits.

So, assuming your "1M sentences" is 100 MB of data, and it takes you 11s to process, that's 9MB/s. For SSD, I think that's too slow.

What iterator do you use to iterate the sentences? Is it `TaggedLineDocument`?

fortiema:
I have actually started playing with the [dask](github.com/ContinuumIO/dask) library for another project, and am starting to think it would be a good fit to solve this particular problem here. I'm gonna put up a quick prototype as soon as I have some spare time this week.

Any objection to integrate this into gensim if it were to work?

piskvorky:
Not at all, I've had my eyes on dask / blaze for a while.

The "streaming" abstraction in gensim is very powerful, but the guys at Continuum are doing lots of great work, so why not tap into it. Similarly with a Spark bridge.

honnibal:
@fortiema , see my code snippet in #400 : I think the big problem is actually that the counts are stored in a Python dictionary, which gets slow when it gets very large.

I'm working on proper benchmarks, over large jobs. So far my process is at 3.5bn words counted after 30 minutes. Of course, there could be bugs :)

menshikh-iv:
Ping @fortiema

menshikh-iv:
Connected with #1446

PR #408, opened 2015-07-22
Modify install problem on gensim/summarization/keywords.py

AppleHolic:
On my mac(Mac OS X Yosemite, python 2.7.6), there is a install problem.

In 12th line (from six.moves.queue import Queue as _Queue), it incurs import problem.

But I change this line and '_init_queue' function. And then, I can install this library.

I glad to use this library for my research. Thanks.

### DIFF
--- a/gensim/summarization/keywords.py
+++ b/gensim/summarization/keywords.py
@@ -9,7 +9,7 @@ from gensim.summarization.textcleaner import tokenize_by_word as _tokenize_by_wo
 from gensim.summarization.commons import build_graph as _build_graph
 from gensim.summarization.commons import remove_unreachable_nodes as _remove_unreachable_nodes
 from itertools import combinations as _combinations
-from six.moves.queue import Queue as _Queue
+from six.moves import queue
 from six.moves import xrange
 
 
@@ -61,11 +61,11 @@ def _process_first_window(graph, tokens, split_text):
 
 
 def _init_queue(split_text):
-    queue = _Queue()
+    q = queue.Queue()
     first_window = _get_first_window(split_text)
     for word in first_window[1:]:
-        queue.put(word)
-    return queue
+        q.put(word)
+    return q
 
 
 def _process_word(graph, tokens, queue, word):


midnightradio:
As far as I understand, it has a good reason for the single underscore leading the module name Queue. I guess the importing error caused by your own environmental problem or has to be fixed in other ways but not like this. Simply, this change would not always go well with python 3 , and for the same reason I guess the contributor put the single underscore for python 2. FYI, "Queue" in python 2 has been renamed to "queue" in python 3.

AppleHolic:
Thanks for your comment. I understood how appear the problem.

piskvorky:
@AppleHolic As far as I can tell, your change in this PR is exactly equivalent to the original code. Just renaming a variable.

I don't understand how this change could make a difference for you -- can you explain? How come gensim didn't work before, but worked after?

If it helps something, we can certainly change it, but I don't understand the difference.

AppleHolic:
kakaotalk_photo_2015-07-30-15-17-46_76

Before I did edit code lines like commit, on the another mac book(python 2.7.6, OS X yosemite 10.10.4 ... ), it also occur import error likes above. As you saw in image, six.moves.queue module cannot be imported. So then, I did changed this line.

To solve this problem, I did search on google 'python six moves queue import error', I can find same import problem on google in the other python library. So I tried them, and it was solved.

I wish it will help to solve this problem.

Regards & thanks.

piskvorky:
@AppleHolic what is your version of `six`?

AppleHolic:
@piskvorky 
Version is 1.9.0!

piskvorky:
OK, that's a new version. That means I still don't understand what's happening here :(

I mean, if your fix really helped, we can certainly use it (merge this PR). But I'd like to understand what's happening, because this is pretty weird. CC @cscorley @gojomo any ideas?

AppleHolic:
@piskvorky 
I'm so sorry to check the version observantly. 

I checked again what is the version of six occurs this problem. The versions (1.2, 1.3) occurs other problem, and The veision 1.4 occurs exactly same problem. And after then there is no problem while importing lib.

So I recommand the requirement of the version of six library is greater than 1.5. I think it occurs again this problem when the six (version 1.4.0) is already installed.

piskvorky:
Sure, that would explain it!

Thanks a lot for investigating, I will merge your pull request from #412.

GirllunarExplorer:
I am also running on 1.9 and having the same issue.

piskvorky:
Then it must be a different issue -- this one was resolved by upgrading six.

@GirllunarExplorer can you check your gensim & six versions? To avoid potential confusion around multiple installations / virtual environments / python paths, can you do it like this:

```python
import gensim
import six
print gensim.__version__, six.__version__
```

GirllunarExplorer:
Ah that was it, I had an old version running (1.4.1)

dushshantha:
I had the same issue in Mac OSX Yosemite. I uninstalled gensim and six and installed them back using code in github. Worked just fine. Below are my versions that works together just fine.

gensim : 0.12.1 six : 1.9.0

iownsecrecy:
I was facing the exact issue, OS X Yosemite python 2.7.6. I'm using python via PyCharm and use pip to install everything. Both PyCharm package manager and pip said I had six 1.9.0 but when tried ```six.__version__``` it said 1.4.1. 

So, I downloaded and installed six manually. Solved my problem. My guess is that it had something to do with the environment. Hope this helps in finding the real cause.

ganesshkumar:
When I use pip to get the version six, it tell me 1.9.0

    $ pip list | grep six
    six (1.9.0)

But when I run check the same in python shell, it tell me 1.4.1

    >>> six.__version__
    '1.4.1'

piskvorky:
@ganesshkumar apparently a lot of people share this schizophrenic Python installation. Can you say a little more about how you got there? How did you install Python? How did you install pip? What OS do you use? Where is pip? (`which pip` + *cat `which pip`*)

ganesshkumar:
OS: Os X 10.10.5
Python version: 2.7.10 (I guess python came installed, as I don't remember installing it manually)
I installed pip using `sudo easy_install pip`
pip version 7.1.2

    $ which pip
    /usr/local/bin/pip

Let me know if you need any other information

piskvorky:
Can you post the output of the following:

```bash
$ which python
$ cat `which pip`
```

xyhu:
I have the same issue as @ganesshkumar
```
$ which python
/usr/bin/python

cat `which pip`
#!/usr/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'pip==7.1.2','console_scripts','pip'
__requires__ = 'pip==7.1.2'
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('pip==7.1.2', 'console_scripts', 'pip')()
    )
```

piskvorky:
My recommendation would be to start using virtual environments, and ignore all the global packages. 

Install [virtualenvwrapper](https://virtualenvwrapper.readthedocs.org/en/latest/) and start a new Python environment. Install everything you need into that environment.

I think these days, virtualenv defaults to *not* using global site packages (please check), so you don't even need to configure anything special when starting a new environment.

Basically, you don't want to mix packages from different places, from multiple site-packages folders and paths. Installing virtualenvwrapper should do that for you.

Let me know if that fixes it.

xyhu:
Thank you so much for your reply! I'll tried the virtualenvwrapper and it works!! Before your reply, I found a way to solve the problem by using Spyder and use the python environment through it. In order to use gensim, I added ```/Library/Python/2.7/site-packages``` to the PYTHONPATH manager in Spyder and everything works.

shinyke:
>>> import six
>>> print six.__version__
1.9.0
>>> import gensim
  File "C:\kxmApps\develop\Anaconda\lib\site-packages\gensim-0.12.2-py2.7-win32.
egg\gensim\__init__.py", line 6, in <module>
    from gensim import parsing, matutils, interfaces, corpora, models, similarit
ies, summarization
  File "C:\kxmApps\develop\Anaconda\lib\site-packages\gensim-0.12.2-py2.7-win32.
egg\gensim\summarization\__init__.py", line 4, in <module>
    from .keywords import keywords
  File "C:\kxmApps\develop\Anaconda\lib\site-packages\gensim-0.12.2-py2.7-win32.
egg\gensim\summarization\keywords.py", line 13, in <module>
    from six.moves.queue import Queue as _Queue
ImportError: No module named queue
>>>

I check my six version but still can not import gensim，Is there something wrong？

my environment：
windows 7
python 2.7
gensim 0.12.2 
six 1.9.0

piskvorky:
@shinyke that's weird, six 1.9.0 should have that queue module.

Can you try importing manually, i.e.

```python
import six
print six.__version__
from six.moves.queue import Queue as _Queue
```

And if that still fails maybe try asking the developers of `six`. It's supposed to be a simple package, I don't know why it keeps giving people so much trouble, out of all the dependencies.

aliostad:
How come I am getting this error in 0.12.4? The code changes by @AppleHolic fixed the issue for me.

piskvorky:
@aliostad see my reply above to @shinyke .

aliostad:
Thank you @piskvorky Long time due. 
Should I ask questions/possible issues here or google groups?

piskvorky:
General or open-ended questions on the mailing list; bug reports on Github (but see our [CONTRIBUTING.md](https://github.com/piskvorky/gensim/blob/develop/CONTRIBUTING.md)).

PR #409, opened 2015-07-27
Added --lda parameter for testing

alexandreyc:
This micro patch fixes a segmentation fault when VW is running in test only mode (tested on 7.7.0 and 8.0.0). This bug occurs on OS X (10.10.3). I can give you more details if needed.

### DIFF
--- a/gensim/models/wrappers/ldavowpalwabbit.py
+++ b/gensim/models/wrappers/ldavowpalwabbit.py
@@ -327,6 +327,7 @@ class LdaVowpalWabbit(utils.SaveLoad):
         """Get list of command line arguments for running prediction."""
         cmd = [self.vw_path,
                '--testonly', # don't update model with this data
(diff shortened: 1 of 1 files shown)


piskvorky:
Yes, more details always help :)

Especially when people find this PR later from CHANGELOG, or there's a regression, etc.

piskvorky:
@alexandreyc: can you prepare a short description (1-2 sentences) for what this PR fixed, when it's useful, how to use it?

alexandreyc:
This PR adds `--lda`parameter to VW command line when used for testing (i.e prediction). Without this parameter VW segfault on my laptop. I think this option should always be passed when doing LDA with VW (not only when fitting model, but also when testing). However it looks like the problem comes from VW and not gensim.

piskvorky:
Let's try to ping Matt Hoffman, author of the VW implementation. Perhaps he know more :) @heywhoah

Issue #382, opened 2015-07-03
word2vec_inner compile fail due to import scipy.linalg.blas with scipy 0.17


ylhsieh:
I notice that when compiling `word2vec_inner.pyx` it fails at `from scipy.linalg.blas import fblas`. Looking at scipy's [documents](http://docs.scipy.org/doc/scipy/reference/release.0.12.0.html#fblas-and-cblas) I see that this is deprecated since scipy 0.12. Recently I updated my scipy and maybe that caused this error. I changed it to `import scipy.linalg.blas as fblas` to get it to work without modifying other codes. Did anyone encounter this problem?

gojomo:
My working environments are at scipy 0.15.1, and I hadn't noticed any such warning (or failing error), either via the automated install or my manual invocations of cython-compilation. 

How are you compiling that file?

ylhsieh:
Hi, I put `word2vec.py` and `word2vec_inner.pyx` in the same folder, and use the following (modified some imports to make it easier to copy and test):
```python
logger.info("Building word2vec_inner")
import pyximport, os
from numpy import get_include
models_dir = os.getcwd()
pyximport.install(setup_args={"include_dirs": [models_dir, get_include()]})
from word2vec_inner import train_sentence_sg, train_sentence_cbow, FAST_VERSION
logger.info("Built and imported word2vec_inner")
```

It results in 
`ImportError: Building module word2vec_inner failed: ['ImportError: cannot import name fblas\n']`
Perhaps the reason roots from the 0.17 version of scipy? That they officially took out the `fblas`. If I have the time, I would try to downgrade scipy to 0.15 and try again. Thanks.

ylhsieh:
I am deeply sorry to clicked the wrong button! Reopening this.

piskvorky:
Yes, I've seen that deprecation, but forgot to investigate.

@ylhsieh can you investigate? If the fix is as simple as deleting `f` from `fblas`, then let's do it :)

ylhsieh:
Will do. But I suggest to add some version checking in the .pyx files instead of changing it altogether, as some might still be using scipy older than 0.12. For me, changing from `from scipy.linalg.blas import fblas` to `import scipy.linalg.blas as fblas` works, and the rest of the .pyx file can remain the same. This method works for both `word2vec_inner.pyx` and `doc2vec_inner.pyx`.

piskvorky:
Yes, can you check on scipy 0.10.1 please?

[...]

PR #414, opened 2015-07-31
Added support for np.ndarrays for SlicedCorpus

mattilyra:
The SlicedCorpus currently only supports slicing using Python lists. It would be quite useful if the slicing could be done similar to NumPy fancy indexing. I think the changes needed are quite simple.

Not sure how to handle the serialisation of the `index` though (turn it into a list - seems silly).

### DIFF
--- a/gensim/corpora/indexedcorpus.py
+++ b/gensim/corpora/indexedcorpus.py
@@ -20,7 +20,7 @@ of each document.
 import logging
 import shelve
 
-import numpy as np
+import numpy
 
 from gensim import interfaces, utils
 
@@ -48,6 +48,11 @@ class IndexedCorpus(interfaces.CorpusABC):
             if index_fname is None:
                 index_fname = utils.smart_extension(fname, '.index')
             self.index = utils.unpickle(index_fname)
+            # change self.index into a numpy.ndarray
+            # this will break code like
+            #     if corpus.index:
+            #         do stuff
+            self.index = numpy.asarray(self.index)
             logger.info("loaded corpus index from %s" % index_fname)
         except:
             self.index = None
@@ -96,6 +101,10 @@ class IndexedCorpus(interfaces.CorpusABC):
                 serializer.__name__)
 
         # store offsets persistently, using pickle
+        # we shouldn't have to worry about self.index being a numpy.ndarray as the serializer will return
+        # the offsets that are actually stored on disk - we're not storing self.index in any case, the
+        # load just needs to turn whatever is loaded from disk back into a ndarray - this should also be
+        # backwards compatible
         logger.info("saving %s index to %s" % (serializer.__name__, index_fname))
         utils.pickle(offsets, index_fname)
 
@@ -115,7 +124,7 @@ class IndexedCorpus(interfaces.CorpusABC):
         if self.index is None:
             raise RuntimeError("cannot call corpus[docid] without an index")
 
-        if isinstance(docno, slice) or isinstance(docno, np.ndarray):
+        if isinstance(docno, (slice, numpy.ndarray)):
             return utils.SlicedCorpus(self, docno)
 
         return self.docbyoffset(self.index[docno])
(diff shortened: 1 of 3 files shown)


gensim/corpora/indexedcorpus.py
```diff
@@ -113,7 +115,7 @@ def __getitem__(self, docno):
         if self.index is None:
             raise RuntimeError("cannot call corpus[docid] without an index")
 
-        if isinstance(docno, slice):
+        if isinstance(docno, slice) or isinstance(docno, np.ndarray):
```
piskvorky: `isinstance` also accepts multiple class options = tuple as its 2nd parameter.

piskvorky:
Thanks @mattilyra !

Can you also add some tests around this new functionality? These would go into [test/test_corpora.py](https://github.com/piskvorky/gensim/blob/develop/gensim/test/test_corpora.py).

mattilyra:
Sure. I'll add the tests and other missing bits and update the PR once I've got all that done. I was thinking of changing the `index` permanently to a `numpy.array`, are you ok with that? Also are you using `numpy` instead `np` as the NumPy reference? Need to update that as well.

piskvorky:
Great :)

`numpy`/`np`: whatever works for you. Same with `index` as as np array. Just make sure we don't break anything horribly. Also, previously serialized corpora should continue to work (backward compatibility), even if new ones work a bit differently underneath.

mattilyra:
There is one thing I could think of that might break backwards compatibility. If anyone is relying on `IndexedCorpus.index` being a `list` by doing something like `if corpus.index: do_thing()` then that will break.

I thought it's better to break that _always_, instead of changing the `index` on the fly if someone uses fancy indexing. That could lead to extremely weird behaviour, consider the following

    c = IndexedCorpus('dummy.txt') # indexed corpus
    slice_of_c = c[:100:2] # index remains a list
    if slice_of_c.index: # everything is fine because index is a list
        do_something()
    slice_of_c = c[[10, 20, 30, 40]] # index is now a ndarray
    if slice_of_c.index: # FAIL

I think it's better to consistently fail the `if` clauses by always having `index` be `ndarray` instead of it being dependent on runtime state.

gensim/corpora/indexedcorpus.py
```diff
@@ -113,10 +121,12 @@ def __getitem__(self, docno):
         if self.index is None:
             raise RuntimeError("cannot call corpus[docid] without an index")
 
-        if isinstance(docno, slice):
+        if isinstance(docno, (slice, list, numpy.ndarray)):
             return utils.SlicedCorpus(self, docno)
-
-        return self.docbyoffset(self.index[docno])
+        elif isinstance(docno, (int, numpy.int)):
```
piskvorky: `numpy.int` and `int` are the same thing; perhaps you meant `numpy.integer`? (to also cover int32, uint16 etc etc).

piskvorky:
I guess we'll have to fix the scipy fblas thingy from #382 first, otherwise the Travis tests just stall :(

mattilyra:
oh snap - I started testing the `scipy/word2vec_inner` changes yesterday but ran into issues installing `scipy-0.10.1`

piskvorky:
It's OK, I've fixed the Travis version to 0.15.1 just now, in commit 1cb6919.

So Travis should now work OK -- if you can merge `develop` into your branch.

mattilyra:
it seems there were some failing tests in https://github.com/piskvorky/gensim/commit/1cb6919b69a0c4bfc178c7b1e9a88e6bff6264a8

piskvorky:
The fails seem unrelated (doc2vec, #386, cc @gojomo ).

This PR seems good to go! Thanks a lot Matti :) Merging.

piskvorky:
Also, what is your preferred CHANGELOG entry for this change (1 short phrase / sentence)?

mattilyra:
Added support for NumPy style fancy indexing to corpus objects.

piskvorky:
Added in 591555e3.

PR #415, opened 2015-07-31
Add exception handling when importing from different versions of scipy

ylhsieh:
As per #382, add some exception handling when importing scipy linalg module.

### DIFF
--- a/gensim/models/doc2vec_inner.pyx
+++ b/gensim/models/doc2vec_inner.pyx
@@ -15,7 +15,12 @@ cimport numpy as np
 from libc.math cimport exp
 from libc.string cimport memset, memcpy
 
-from scipy.linalg.blas import fblas
+# scipy <= 0.15
+try:
+     from scipy.linalg.blas import fblas
+except ImportError:
+     # in scipy > 0.15, fblas function has been removed
+     import scipy.linalg.blas as fblas
 
 from word2vec_inner cimport bisect_left, random_int32, \
      scopy, saxpy, sdot, dsdot, snrm2, sscal, \
--- a/gensim/models/word2vec_inner.pyx
+++ b/gensim/models/word2vec_inner.pyx
@@ -15,7 +15,12 @@ from libc.math cimport exp
 from libc.math cimport log
 from libc.string cimport memset
 
-from scipy.linalg.blas import fblas
+# scipy <= 0.15
+try:
+    from scipy.linalg.blas import fblas
+except ImportError:
+    # in scipy > 0.15, fblas function has been removed
+    import scipy.linalg.blas as fblas
 
 REAL = np.float32
 


transcranial:
I tried that on scipy 0.16.0 and it doesn't work (it passes on travis because it's testing against 0.15.1), because all the blas functions are now raw function pointers that can be cimported: `cimport scipy.linalg.cython_blas`. See [cython_blas doc](https://docs.scipy.org/doc/scipy-dev/reference/linalg.cython_blas.html).

For example one cannot call `scipy.linalg.blas.sdot` which is what your code would do with `import scipy.linalg.blas as fblas`. 

To support both APIs, one needs to do conditional compilation I believe, but given the different scoping rules for python and cython, I'm having trouble figuring out how to do this.

Other ideas?

piskvorky:
No ideas :( Can you perhaps ask the scipy developers? I guess they know best what a valid replacement is, after removing the old functionality.

ylhsieh:
Hmm.. The one version that I did not try is 0.16. The PR works on 0.17dev. and 0.15. Also, according to the [scipy BLAS docs](https://docs.scipy.org/doc/scipy-dev/reference/linalg.blas.html#module-scipy.linalg.blas) one can in fact call `scipy.linalg.blas.sdot` and other BLAS related functions. I am not sure about the difference between `import` and `cimport`, but since these two .pyx files use `import`, I guess we don't need `cimport`?

piskvorky:
Yeah we're doing tricks around function pointers there, so this will needs good testing.

@transcranial you say this PR doesn't work, but it works for @ylhsieh . Can you find the difference? Or what is the error?

transcranial:
@ylhsieh thanks for the link and clarification.

@piskvorky I apologize, it does indeed work in 0.16.0 - I forgot to regenerate doc2vec_inner.c and word2vec_inner.c with cython.

Incidentally, in 0.16.0 this worked due to direct access to the function pointers (though due to the issues I mentioned before, it may be difficult to use this in addition to the older scipy API):

```
from scipy.linalg.cython_blas cimport scopy, saxpy, sdot, dsdot, snrm2, sscal
```

in place of

```
from scipy.linalg.blas import fblas

cdef scopy_ptr scopy = <scopy_ptr>PyCObject_AsVoidPtr(fblas.scopy._cpointer)  # y = x
cdef saxpy_ptr saxpy = <saxpy_ptr>PyCObject_AsVoidPtr(fblas.saxpy._cpointer)  # y += alpha * x
cdef sdot_ptr sdot = <sdot_ptr>PyCObject_AsVoidPtr(fblas.sdot._cpointer)  # float = dot(x, y)
cdef dsdot_ptr dsdot = <dsdot_ptr>PyCObject_AsVoidPtr(fblas.sdot._cpointer)  # double = dot(x, y)
cdef snrm2_ptr snrm2 = <snrm2_ptr>PyCObject_AsVoidPtr(fblas.snrm2._cpointer)  # sqrt(x^2)
cdef sscal_ptr sscal = <sscal_ptr>PyCObject_AsVoidPtr(fblas.sscal._cpointer) # x = alpha * x
```

piskvorky:
Excellent! So are we good to merge here? Or anything else missing?

I'll revert the `travis.yml` commit that fixes scipy version at 0.15.1 afterward.

piskvorky:
I tried to open a PR against your PR @ylhsieh , but github doesn't recognize your fork... weird. Looks like a bug on github.

Anyway, let's merge this and finish this PR. Is there anything else you wanted done before merge?

ylhsieh:
I am very new to Github so I cannot figure out what is going on here, sorry. What is the meaning of 'open a PR against your PR'? Anyways, nothing needed to be done here. Thank you!

piskvorky:
I wanted to add a new commit to this PR (your develop branch). But github didn't allow me. Probably a bug -- I filed a bug report.

Anyway, thanks for the fix @ylhsieh ! Merged now :)

ylhsieh:
Maybe I have to grant some kind of access to you in order for you to commit? Dunno.
Thanks to you and all other authors for providing such a great tool as gensim. I am glad to contribute!

piskvorky:
They replied from GitHub about the PR bug. It's a "feature" -- when there are more than 500 forks, they clip the number to 500, for browser performance reasons.

So, some forks are not available for pull requests.

The work around is to construct the pull request URL manually, not relying on github's "pull request" button, like so:
https://github.com/ylhsieh/gensim/compare/develop...piskvorky:ylh_fix

Irrelevant here now; just documenting the workaround, for archival and for clarity :)

Issue #407, opened 2015-07-21
AttributeError: 'Doc2Vec' object has no attribute '_prepare_items'


ryankohl:
After training a Doc2Vec model (with the Brown corpus), I went ahead and attempted scoring.  However, I got the following error:
AttributeError: 'Doc2Vec' object has no attribute '_prepare_items'

And indeed, neither Word2Vec nor Doc2Vec seem to have that method.  In an older version of gensim (0.10.3), there was reference to a '_prepare_sentences', which was indeed a method in Doc2Vec.

Guessing the method just got lost in a refactor?

gojomo:
Yes – I seemed to have refactored away a method @mataddy's scoring work relied upon. 

You can probably restore functionality by finding the `_prepare_items` method that existed at the same time as the  line calling it. 

More generally, the `score*` methods could parallel the refactorings in the `train*` paths, and a unit test would help guard against future breakage.

ryankohl:
Turns out that there may be another path - restoring the _prepare_items method lead to a cascade of other tweaks needed, so I looked around and found the infer_vector() method, which seems to do the job.  However, that method was present back in 0.10.3, so perhaps the score() method was doing something different.  At the end of the day, I'm only looking for word/paragraph vectors, so perhaps infer_vector() is the right call?

mataddy:
hi @gojomo and @ryankohl  there is no `score` method for Doc2Vec, so I'm a bit confused.  But regardless if  `_prepare_items` has indeed been removed from Word2Vec after the last release then I'll make sure to parallel the changes in `train` asap.  I'm travelling right now but will take a look next chance I get.

also @gojomo I will add a unit test to `TestWor2VecModel` in the near future.

ryankohl:
They @mataddy - the score method is inherited from the Word2Vec class.

mataddy:
thanks @ryankohl, that makes sense.

gojomo:
I'm not sure 'score' has a strong/straightforward interpretation for Doc2Vec – except indirectly because of the similarity of many of the training modes.... which might lead to plausibly-useful results even without intentional design. 

[...]

PR #417, opened 2015-08-02
fix scoring without __prepare_items, add unit test

mataddy:
This fixes the problems pointed out in #407, which occurred due to a refactoring of the Train methods.

I went through the new train methods and adapted things related to score() accordingly.  Re-running my standard yelp example looks fine: http://nbviewer.ipython.org/github/taddylab/deepir/blob/master/deepir.ipynb

Also added a testScoring() so that we won't be suprised by this type of thing again.

Thanks!

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -249,7 +249,8 @@ except ImportError:
         if model.negative:
             raise RuntimeError("scoring is only available for HS=True")
 
-        for pos, word in enumerate(sentence):
+        word_vocabs = [model.vocab[w] for w in sentence if w in model.vocab]
+        for pos, word in enumerate(word_vocabs):
             if word is None:
                 continue  # OOV word in the input sentence => skip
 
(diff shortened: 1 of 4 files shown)


piskvorky:
Alright! The test is rather superficial, but it will do for now :)

Merging. Thanks @mataddy !

mataddy:
no prob, thank you.  yep, the test is a bit cheeky I'll update it I promise ;-)

gojomo:
That test would have saved me from inadvertently breaking this – sorry! – so it's a great start in my view. :)

mataddy:
hi @piskvorky, sorry for the long delay but there is finally a more introductory demo up at https://github.com/TaddyLab/deepir/blob/master/deepir.ipynb.  the `docprob` wrapper function in the demo is something I could add to gensim if you're into it.

piskvorky:
Looks nice, thanks Matt!

Do we integrate this notebook into the [docs/notebooks](https://github.com/piskvorky/gensim/tree/develop/docs/notebooks) directory of gensim, or just link to it somehow?

I won't have time to review in detail until November; CC @tmylk.

mataddy:
great idea; so long as those docs/notebooks aren't run when the project is built this will work great.  I'll create a pull request with it and ping @tmylk

gojomo:
Hmm, testing the notebooks each build is an interesting idea... but for now, they're not being run/tested every build.

mataddy:
cool, thanks @gojomo. FYI the only reason it wouldn't work for my notebook here is because the data is on kaggle and the user needs to sign-up to get it.  but that could be worked around later if you guys decide to change policies and run the notebooks at each build

PR #418, opened 2015-08-03
Support unknown words by providing (trained) default vector/index

sjebbara:
Modified word2vec file that allows to dynamically replace unseen words with specified token, e.g. "UNK" by using a defaultdict.

During the training of the Word2vec object, all words that appear less than min_count in the dataset are actually treated as occurrences of the unknown_word.
During application, all new (unseen) words are treated as the unknown_word, thus, queries for word vectors or indices of these unseen words return the respective value for the unknown_word.


piskvorky:
Thanks @sjebbara .

IIRC, this version performs worse (lower accuracies) than just ignoring the words. Can say a little more about your use case? In what scenario is this UNK change beneficial?

Btw most of the formatting changes are a regression (PEP8 compliance). We'd have to get rid of them (keep the current version) and resolve conflicts, before merging.

sjebbara:
Thanks for your response.
I just started to do some sequence learning with recurrent neural nets (sentiment prediction) for which I wanted to use pretrained word vectors. I implemented this functionality mainly to stay close to my literature (e.g. Luong, Minh-Thang, Richard Socher, and Christopher D. Manning. (2013), Yao, Kaisheng, et al. 2013) and I thought that was a rather common way of dealing with unseen words. Additionally, my (untested) assumption was, that having an identifier for an unknown word in a sequence provides more information than dropping it. 

The formatting issues result from my PyCharm IDE's auto-formatting. I assumed this formatting would be PEP8 compliant. Sorry for that.

I will perform some tests on my implementation and let you know if it is actually useful or not :D

piskvorky:
Great :)

FYI, I think the improvement of dropping words (rather than replacing them) comes from effectively increasing the window size. Words dropped => farther words get into a window => better modeling of long distance relationships.

But of course, an experiment is always better. And it looks like your change is fully backward compatible (is it?), so there's no risk.

gojomo:
I can understand thinking why a stand-in `UNK` token might help – it provides some place for the presence of unknown words to be registered in the model, and preserves the relative positional-relationships between known words. 

But in practice I suspect several things work against it working that way: (1) as Radim notes, dropping the words allows a de facto expansion of the context window in exactly those places where the rare/unknown word means there's less prior knowledge; (2) most of the training methods don't care about exact word positions, so dropping words doesn't introduce position-confusion; (3) the rare/unknown words, rather than being importantly like each other, may instead represent the *most* diverse/conflicting meanings, so collapsing them into one shared stand-in token just pushes broad 'noise' into the window (while pushing other real specialized context-words out). 

(Other slightly-more-sophisticated policies might help remedy such limitations: replacing unknown-words with a part-of-speech specific token, or maybe even something generated from its letters/phonemes – thus simulating a person's ability to work from roots/sounds/etymologies.)

Unless there's an advantage to this approach in some known cases, I'd lean against integrating support for it as another parameter, because (1) the extra branches/special-casing makes the existing code harder to read/maintain; (2) some users tend to assume every parameter is there to help, and so may be attracted to use it to their detriment; (3) it shouldn't be too hard to simulate this treatment with corpus-preprocessing, keeping such policy customizations separate from the `scale_vocab()` method or individual `train_*` methods. 

To the end of supporting such experiments without lots of special-casing in the main code, perhaps it would be better to (1) add a generic facility for marking tokens as 'must keep' (despite not appearing in, or being extremely rare in, the initial vocab-scan); (2) make the `train_*` methods tolerate already-looked-up lists-of-vocabs, in addition to lists-of-tokens-to-be-looked-up. Then an unknown/rare-word handling policy would be easier to implement as a wrapper around the corpus. 

Looking at the implementation, the tallying of all rare words into the single stand-in token's occurrence count will affect frequent-word subsampling and negative-sampling in ways that may not logically fit with ways the `UNK` token might help. (Even if there are N million UNK occurrences, it might be better to pretend each was still rare, rather than both downsample it and negative-sample it as if it were a very common word.)

tmylk:
@gojomo Do I understand correctly that this PR should be closed and not merged?

gojomo:
In my opinion, yes – even *if* the approach offered tangible benefits, it could better be done as a preprocessing step.

piskvorky:
I agree. Thanks @sjebbara for the input and the discussion -- add UNK during preprocessing if needed.

PR #422, opened 2015-08-07
Fixed compatibility with S3 for gensim.utils.SaveLoad

ziky90:
Minor fix that enables loading objects that inherited from `gensim.utils.SaveLoad` also from Amazon S3. Bug was described in https://github.com/piskvorky/gensim/issues/421.

This implementation is necessary, before there will be fixed issues https://github.com/piskvorky/smart_open/issues/13 and https://github.com/piskvorky/smart_open/issues/27 in smart_open library.

### DIFF
--- a/gensim/utils.py
+++ b/gensim/utils.py
@@ -915,7 +915,7 @@ def pickle(obj, fname, protocol=2):
 def unpickle(fname):
     """Load pickled object from `fname`"""
     with smart_open(fname) as f:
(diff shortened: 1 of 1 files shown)


piskvorky:
Merged, thanks.

This may lead to inefficiencies, in case pickle proceeds in smaller chunks for file-like inputs by default, to save memory... but I don't think so. Should be ± the same memory.

PR #424, opened 2015-08-11
TransformedCorpus should still be `slice`able.

mattilyra:
Added support for slicing a `TransformedCorpus`. For instance, the corpus returned from applying `TfIdfModel` can still be randomly indexed, not just iterated over.

PR #425, opened 2015-08-11
TransformedCorpus should still be `slice`able.

mattilyra:
Fixes a bug in PR #424 where the transformation was not being applied to a sliced TransformedCorpus. Added test case, and an entry in the CHANGELOG.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,12 +1,10 @@
 Changes
 =======
 
-0.12.3
-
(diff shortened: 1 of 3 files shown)


piskvorky:
Thanks Matti. The next release will be 0.12.2 though, not 0.12.3.

mattilyra:
ok, I think I'm done with this one

piskvorky:
Thanks again, merging :)

PR #430, opened 2015-08-14
[doc2vec] train new doc tags with old words vocab

qiwang321:
Often we want to train doc vecs incrementally for new docs, without changing the word vecs (and old doc vecs). This train_new_doc interface enables us to do that

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -37,6 +37,7 @@ The model can also be instantiated from an existing file on disk in the word2vec
 import logging
 import os
 import warnings
+import itertools
 
 try:
     from queue import Queue
@@ -288,7 +289,8 @@ class DocvecsArray(utils.SaveLoad):
             self.count = max(self.count, key+1)
         else:
             if key in self.doctags:
-                self.doctags[key] = self.doctags[key].repeat(document_length)
+                #self.doctags[key] = self.doctags[key].repeat(document_length)
+                return
             else:
                 self.doctags[key] = Doctag(len(self.index2doctag), document_length, 1)
                 self.index2doctag.append(key)
@@ -354,7 +356,7 @@ class DocvecsArray(utils.SaveLoad):
         """Estimated memory for tag lookup; 0 if using pure int tags."""
         return 60 * len(self.index2doctag) + 140 * len(self.doctags)
 
-    def reset_weights(self, model):
+    def reset_space(self, model):
         length = max(len(self.doctags), self.count)
         if self.mapfile_path:
             self.doctag_syn0 = np_memmap(self.mapfile_path+'.doctag_syn0', dtype=REAL,
@@ -366,10 +368,35 @@ class DocvecsArray(utils.SaveLoad):
             self.doctag_syn0 = empty((length, model.vector_size), dtype=REAL)
             self.doctag_syn0_lockf = ones((length,), dtype=REAL)  # zeros suppress learning
 
+    def init_weight(self, model, i):
+        # construct deterministic seed from index AND model seed
+        seed = "%d %s" % (model.seed, self.index2doctag[i] if len(self.index2doctag) > 0 else str(i))
+        self.doctag_syn0[i] = model.seeded_vector(seed)
+       
+    def reset_weights(self, model):
+        self.reset_space(model)
+        length = max(len(self.doctags), self.count)
         for i in xrange(length):
-            # construct deterministic seed from index AND model seed
-            seed = "%d %s" % (model.seed, self.index2doctag[i] if len(self.index2doctag) > 0 else str(i))
-            self.doctag_syn0[i] = model.seeded_vector(seed)
+            self.init_weight(model, i)
+
+    def add_new_docs(self, documents, model):
+        old_length = len(self.index2doctag)
+        for document in documents:
+            document_length = len(document.words)
+            for tag in document.tags:
+                self.note_doctag(tag, 0, document_length)
+        # cache old
+        doctag_syn0_tmp = self.doctag_syn0
+        doctag_syn0_lockf_tmp = self.doctag_syn0_lockf
+        self.reset_space(model)
+        # restore old
+        for i in xrange(old_length):
+            self.doctag_syn0[i] = doctag_syn0_tmp[i]
+            self.doctag_syn0_lockf[i] = doctag_syn0_lockf_tmp[i]
+        # init new
+        length = max(len(self.doctags), self.count)
+        for i in xrange(old_length, length):
+            self.init_weight(model, i)
 
     def init_sims(self, replace=False):
         """
@@ -630,11 +657,22 @@ class Doc2Vec(Word2Vec):
         self.corpus_count = document_no + 1
         self.raw_vocab = vocab
 
+    def train_new_doc(self, documents, total_words=None, word_count=0, chunksize=100, total_examples=None, queue_factor=2, report_delay=1):
+        # add new docs and initialize their weights
+        doc_iter1, doc_iter2 = itertools.tee(documents)
+        self.docvecs.add_new_docs(doc_iter1, self)
+        self.train(doc_iter2, total_words, word_count, chunksize, total_examples, queue_factor, report_delay)
+        
     def _do_train_job(self, job, alpha, inits):
         work, neu1 = inits
         tally = 0
         raw_tally = 0
         for doc in job:
+            # add the doc to dict if it's not already there
+            #document_length = len(doc.words)
+            #for tag in doc.tags:
+                #self.docvecs.note_doctag(tag, 0, document_length)
+
             indexed_doctags = self.docvecs.indexed_doctags(doc.tags)
             doctag_indexes, doctag_vectors, doctag_locks, ignored = indexed_doctags
             if self.sg:
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,8 +1,11 @@
 Changes
 =======
 
+
 0.12.2
 
+* added support for sliced TransformedCorpus objects, so that after applying (for instance) TfidfModel the returned corpus remains randomly indexable. (Matti Lyra, #425)
+* changed the LdaModel.save so that a custom `ignore` list can be passed in (Matti Lyra, #331)
 * added support for NumPy style fancy indexing to corpus objects (Matti Lyra, #414)
 
 0.12.1, 20/07/2015
--- a/gensim/interfaces.py
+++ b/gensim/interfaces.py
@@ -120,6 +120,12 @@ class TransformedCorpus(CorpusABC):
         else:
             for doc in self.corpus:
                 yield self.obj[doc]
+
+    def __getitem__(self, docno):
+        if hasattr(self.corpus, '__getitem__'):
+           return self.obj[self.corpus[docno]]
+        else:
+            raise RuntimeError('Type {} does not support slicing.'.format(type(self.corpus)))
 #endclass TransformedCorpus
 
 
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -41,6 +41,7 @@ from itertools import chain
 from scipy.special import gammaln, psi  # gamma function utils
 from scipy.special import polygamma
 from six.moves import xrange
+import six
 
 # log(sum(exp(x))) that tries to avoid overflow
 try:
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -850,6 +850,9 @@ class LdaModel(interfaces.TransformationABC):
         """
         if self.state is not None:
             self.state.save(utils.smart_extension(fname, '.state'), *args, **kwargs)
+        
+        # make sure 'state' and 'dispatcher' are ignored from the pickled object, even if
+        # someone sets the ignore list themselves
         if 'ignore' in kwargs:
             ignore = kwargs['ignore']
             if isinstance(ignore, basestring):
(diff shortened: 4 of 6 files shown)


piskvorky:
I don't really understand what this does, let's wait for @gojomo 's verdict.

In any case, we don't want to commit any commented-out code, and we do want unit tests :)

gojomo:
It looks like the @mattilyra code is unrelated to the @qiwang321 code – and maybe landed in this PR by mistake? 

If I understand correctly, this is somewhat of a doc2vec analogue to what #365 does for word2vec. It lets you supply a fresh batch of documents, and continues training on them, also expanding the underlying record of doc vecs (unlike inference). 

I understand the interest in this sort of continued-training, but I'm also concerned it will be subject to a lot of misunderstanding: people seeing the method, using it without understanding the tradeoffs. New examples not interleaved with old are almost certain to be gradually diluting/ruining what was learned from earlier training... even essentially invalidating the relative meanings of older vectors relative to newer vectors. Maybe the right choice of alpha mitigates this. Or locking some parts of the model... but then you are practically doing inference. 

I have roughly the same concerns with the word2vec updating... so I'd like to see more description/testing of how these theoretically 'should' be used to get some tangible benefit. And even then, the functionality likely deserves a big disclaimer, "you probably don't want to do this if it is at all possible for you to do full dataset retraining, and the incremental benefits may come with offsetting decay/drift costs". 

For example, for both words and docs, it might make more theoretical sense to extend a big/good starting-model primarily by freezing it, and then only generating new vectors via inference. (This would in some cases require a word-inference technique analogous to the doc technique.) Then, if and only if the gensim structures/APIs are the preferred means of storage/access/similarity, append these inferred vectors into the models for that purpose.

gensim/models/doc2vec.py
```diff
@@ -366,10 +368,35 @@ def reset_weights(self, model):
             self.doctag_syn0 = empty((length, model.vector_size), dtype=REAL)
             self.doctag_syn0_lockf = ones((length,), dtype=REAL)  # zeros suppress learning
 
+    def init_weight(self, model, i):
+        # construct deterministic seed from index AND model seed
```
piskvorky: Comment should be a method docstring.

piskvorky:
@gojomo Is that a final -1 or are there concrete steps @qiwang321 can take to get this merged?

I understand this adds more complexity, and word2vec/doc2vec have become too bulky and hard to follow as it is.

On the other hand, online training is cool, and this doesn't seem to add too much code. So I'm not sure.

I leave this one to your judgement Gordon -- if there are things that must be fixed (apart from the already mentioned commented code, missing tests, Matti's unrelated commits etc), let's add a checklist. If not, let's thank @qiwang321 and close this PR.

gojomo:
My desired is still first and foremost a "desire for discussion/context" rather than "-1". I'd like to know more about the cases where this works, or is expected to work, to get improved results. 

Any discussion about that would then help both structure/polish the code, and also add docs/disclaimers that avoid users rushing to use this without understanding the tradeoffs. (For example, the supporting methods might be clearly segregated, perhaps even in a subclass, and marked as 'experimental' or linked to a FAQ on the unknowns/tradeoffs involved.)

As an example of the sort of evaluation that'd be ideal: take some existing Doc2Vec task, on which its performance on a full dataset can be quantified. Then, try instead just using half the data (as an initial training run), then 5 more incremental `train_new_doc()` updates (each 1/10th of the original data). At each of those steps, and the very end, how well does the 'online' updated model perform? 

(My hunch is, "worse than the original, and quite possibly worse-rather-than-better with each new update", but I of course can't be sure, and maybe very well-considered choices of alpha/iterations/locking can help.)

mattilyra:
Yeah my commits shouldn't be in this pull request, not sure how they even ended up here. They are related to PR #315 which I pushed upstream earlier this week, and which @piskvorky merged and closed 5 days ago.

piskvorky:
Agreed. @qiwang321 , what is your opinion? Can you tell us a little more about your use case, where (and how much) this feature helped etc? Thanks!

qiwang321:
Thanks for your discussion.

The PR is related to a specific use case in my work - in our system we have new documents coming everyday, we want to train vec representations of these new docs, while keeping the old trained documents unchanged. Since there will be many applications built upon these doc representations in the future, it would be better if the old trained documents remain unchanged, so that the whole data stream remains consistent.

For a more concrete example, we will use doc vecs to cluster docs into topics, and as new docs appears, we want to classify them into existing clusters, without changing the original clustering result. Although we could work around by doing a fresh new training each time, this 1) requires a refresh of the whole datastream for upper level applications, which adds to the possibility of fault, 2) takes a long time. I believe that the semantic space should be have a relatively stable structure, especially the word vecs, therefore it makes more sense to keep an initial, thorough training result as a persistent storage, and infer new doc vecs incrementally.

This used to be hard of older versions of gensim, when doc vecs and word vecs were mixed together, and the doc tags will affect the Huffman Tree structure. Then I realized that doc vecs and word vecs are separated in 0.12.1, which make this task easier.

I've tested the commit on my dataset, it performed well, the doc vecs incrementally trained preserves good indication of semantic similarities, not much different from the results obtained by training with a fresh start. This is expected, since when we do an initial training with large enough corpus, we will effectively extract the relatively stable (word_vec) semantic structure, which is useful for inferring new docs.

I'll try to fix the problems mentioned ASAP, and add some unit test and example data as I can. Thanks again for your active consideration and discussion :)

gojomo:
Thanks for the details, that makes good sense and I can also see it being of general use. (It's also good to hear the words-vs-docs refactor made it easier... while that was necessary for a lot of reasons, I know it's made a few things people wanted to do with mixed doc-and-word analysis harder. So nice to know of other work being benefitted by the split.)

It seems that if the 'foundational' batch is enough to create a sufficiently-broad and -representative semantic space, and all docs from that batch (and each further batch) are having their representations frozen, then you'd really want to lock the hidden weights for each incremental batch, as well. Otherwise, the model is 'drifting' towards more recent documents. Eventually older docs, induced against the latest model, could have different optimal representations. 

You could conceivably monitor for this condition over time to see how much of an issue it is in practice, but you could also forestall it entirely by treating your updates more as 'large batch inference', with (or optionally without) appending into the docvecs structure. In such a case, I'd expect your custom `train_*()` wouldn't necessarily use the `infer_vector()` method, but would somehow alter the `train_hidden` value used by the mode-specific training methods, during the batch operation, to make it more like inference. 

Let me know as you need any other suggestions/review!

qiwang321:
@gojomo My application case is exactly a "large batch inference" as you described, but yet it's not that large. It seems that infer_vector() is enough for now (I didn't know there's such a method before...). Thanks a lot for letting me know !
I'll close the PR for now. Thanks again for all the comments and suggestions.

piskvorky:
@qiwang321, what would have made `infer_vector` better discoverable for you? Where would you look for such functionality, to see if it exists or how to use it? Cheers!

qiwang321:
@piskvorky Usually I'll search for or grep in the code the expected keywords for the desired functionality, in this case it's just that I though the functionality should be described by parameters passed to train(), and didn't think of "infer", which is actually a better name. So I just tried to search for train_XXX and got no luck.

I think more examples in the doc could help it better discoverable. However generally I believe the api doc is good enough, I should probably take more responsibility for not discovering it :)

Thanks a lot

dineshbvadhia:
I know this PR is closed but I'd like to add to what @qiwang321 said:

In most production environments, new documents are coming in continuosly. The need is to create new doc vecs and add to the model, and to create word vecs for any new words (in the new docs) and add these to the model too. The option available today is to re-train the entire data set of old plus new documents frequently. It would be much better to be able to re-train the new documents and the new words only, as and when required.

@gojomo suggests above that theoretically it could be possible by freezing the model, and at first sight, I don't see why this would be a problem.

gojomo:
@dineshbvadhia – if the model is already rich/expansive (from a large representative starting base), and then frozen, then what you'd essentially be doing is performing inference on new documents, which already works well. However, newly-inferred documents are not added to the in-memory structure, and so not mixed in for use by the `similarity`/`most_similar`/etc methods. 

My perhaps-incorrect assumption has been that if people are working with an arbitrarily-large stream of new documents, they're likely to persist the vectors some other way – for example, a database rather than ever-larger numpy arrays. (They'd also then be using their own scans or possibly indexes, like the ANNOY index, for operations like nearest-neighbors.)

The code *could* be extended to allow growing the in-memory doc structures to add newly-inferred vectors, but a true "perpetual updates" scenario might hit in-core implementation limits pretty quickly. Also, the easiest way to grow the structures would essentially use at least double-the-memory each expansion – allocate the new larger objects, copy everything over, delete the old – so the resource ceiling under which this is practical is actually smaller than the count of documents that could be handled in one bulk step. (A segmented-structure could get around this, at some extra code/time complexity on each access.) Ultimately I believe most very-large/very-long-lived datasets will want to use gensim to calculate, but some other code to store/access.

Growing the words has slightly different tradeoffs; incremental updates may need a lot of tuning (of alphas, iterations, prior-vector-locking-against-change, etc) to get close to the representative power that's possible with a full, even. balanced training on a whole-dataset. Still, there's ongoing work to add that capability; see #435 and its predecessor. It currently does the 'temporary-double-to-grow' policy. Tests of its results show some ability to learn new words, but haven't been evaluated for overall impact to model quality. (I would use carefully, experimentally – not just with "it exists so it must be wise" confidence.) 

My current hunch is that the practice that's most likely to work well, without significant extra design/testing, would be:

(1) do a big Doc2Vec session with as much representative text as possible, to get a strong base model
(2) use inference on this base model for as long as practical, even with the unfortunate policy of simply dropping new, unknown words
(3) when the proportion of unknown words (or believed new shades-of-meaning in documents) in the document stream gets "too large", then start again from scratch, with a larger bulk text baseline. Your new words – of which there may have been dribs and drabs the whole way, but never enough to get a good vector – may now have enough examples to induce a good vector, and by interleaved cotraining with all other words, the model won't be strongly biased by the luck of training order. Note that after such a fresh-start, old vectors and new vectors aren't likely to be directly comparable. Anything that tried to enforce compatibility – locking some parts of the model in place – would also prevent all the data from having equal effect on the model's power.

piskvorky:
Let me add a  note regarding indexing: ANNOY doesn't support incremental indexing (yet). Gensim's `Similarity` class does. Both are effectively in-core but using virtual memory (mmap).

Similarity search (via ANNOY or [otherwise](https://github.com/piskvorky/gensim/wiki/Ideas-&-Feature-proposals#integrate-a-fast-k-nn-library)) definitely falls within gensim's area of interest.

gojomo:
I haven't yet looked closely at class Similarity's applicability to the w2v/d2v arrays, but its shards seem similar to the segmented-approach I've mentioned for smoother append-growth (but slower access).

There's already some unwanted c&p duplication in the w2v & d2v similarity methods; it may make long-term sense to actually break out the "vector-storage" functionality to a reused class, with support for:

- (optional) name-to-index bidirectional maps (as 'index2word', 'vocab', 'index2doctag', 'doctags')
- graceful grow-in-place via segments (& merging segments if/when desired to take the hit)
- indexing that allows users to be oblivious to the segmenting details
- utility similarity methods, including optimizations like in #340/#350 (which might benefit from an ability to occasionally re-sort vectors by frequency/importance, even when holding names/external-index-numbers constant)

Even the raw training routines could use a segmented vector-source, at some small cost in more-complicated indexing.

Just thinking out loud, for now. I don't yet know how much this would overlap with, or just reuse and be-compatible-with, the Similarity class.

dineshbvadhia:
@gojomo I always learn something new or get my incorrect train-of-thought corrected with your notes. As you say, inference on new documents already works well and secondly, if people are working with large stream of documents (old and new), the vectors should be persisted in some other way. Additionally, for large datasets, calculate the vectors with gensim, and use other code to store/access.

gojomo:
Yes, inference requires the model in memory. It doesn't actually require all the other precalculated doc-vectors, though... so in a memory pinch, if loading a model *just* for inference, parts of the 'docvecs' field (or maybe the whole thing, I'd have to think about it more) could be discarded.

As @piskvorky points out,indexing big vector sets is in-scope for gensim... and with giant-RAM machines, or memmapped files to fast storage, maybe some arbitrarily-big datasets can remain practical within gensim (especially if more work in this area is done). I've just assumed so far that really-big, ever-growing projects will often choose something custom...

dineshbvadhia:
@gojomo If inference of new documents doesn't require all the other pre-calculated doc-vectors then loading a light-weight model to support infer_vector() is a good option to have.

It would be of value to support bigger vector sets. Some projects will choose the custom option as data sizes grow.

gojomo:
@dineshbvadhia – Yes, I can see the usefulness of that – especially since once the model is frozen, further inference can be  easily parallelized/distributed-over-machines. My first thought would be that it could be an optional parameter to `save()` – indicating `inference_only=True` or maybe `save_docvecs=False`. The resulting model would be (much) smaller but after reloading, only be suitable for inference. 

I wouldn't be needing this anytime soon, but if you wanted to prep a PR adding it, I'd be able to review!

piskvorky:
Looks related to the existing `init_sims(replace=True)` API.

That's an API for discarding parts of the model **after** training, in favour of a smaller model that still supports "read-only" functionality (such as finding similar documents, or in this case, inference).

rutum:
@gojomo Fair point that any online training must be used with caution. I have some scores [here](http://rutumulkar.com/blog/2015/word2vec/) (related to [#435](https://github.com/piskvorky/gensim/pull/435)) that compare online training vs training in one go. There is definitely a difference in performance. However, this feature is often needed when the data is arriving piecemeal, so is useful to have.

gojomo:
@rutum I can definitely see the value and demand for such functionality! 

I'm concerned that some may rush to use it without understanding the tradeoffs, even if their data size (and update process) realistically *could* do a complete re-train, or some sort of interleaving of old-and-new, to get more balanced/robust final vectors. (Freezing the older, well-characterized vectors against changes, via the `syn0_lockf` mechanism, may help address such issues... but also introduce other ones.) 

It's good that your blogpost highlights that incremental updates aren't sure to improve all vectors (with regard to the analogy-evaluation), at least not in the same way that a larger/fuller dataset usually would.

PR #433, opened 2015-08-14
Fix dispatching in Python3

spacecowboy:
Following the instructions on setting up a distributed LSI, I ran into 

    TypeError: 'dict_items' object does not support indexing

This one-liner seemed to do the trick. I would have added a unit-test as well but there doesn't
seem to be any for distributated computing.


piskvorky:
Hmm, could have sworn this was fixed long ago. But apparently not. Thanks a lot @spacecowboy !

CC @gojomo : that word2vec random test fail reared its head again: https://travis-ci.org/piskvorky/gensim/jobs/75598327

Issue #423, opened 2015-08-11
Word2Vec: Make it possible to add more complex rules for vocabulary trimming


phdowling:
For some use cases, it may be necessary to keep certain words in the vocabulary even if their support is less than `min_count`. In my example, we're training vectors for DBpedia entities, and we'd like those to remain in the vocabulary even if their count is too low, or possibly still trim them but at a lower count than regular vocabulary words.

My proposal would be an optional parameter (`trim_rule`?) which could be perhaps a callable that returns a boolean, or some container that implements the `__contains__` method, and which is checked by word2vec when the preprocessing trims down the vocabulary. So far I've just been editing the gensim source locally and doing something like this, but that's of course a bit of a pain, and I think other might be interested in this kind of feature as well.

What do you think? Would this be a good feature to add? I'd probably be willing to implement this myself if there is interest.

gojomo:
Some scattered thoughts: 

I've seen some cases where making a word or words immune to frequency-based trimming could be helpful – most recently, as a possible way to support the same effect as #418 with less case-specific code. 

In the realm of simple parameterizable policies, a 'must keep' list would be an easy possibility. Per-word min_count thresholds that override the global would be another. Perhaps also, similarly, a 'boost' for certain words, to treat them 'as if' they occurred N times more frequently. (How such a false-frequency would then affect other calculations, like subsampling/neg-sampling/progress-estimation, would be something to consider.)

A callable offers the most flexibility; might introduce extra serialization/deserialization gotchas. It could conceivably be called once, after the 'default' trimming has occurred but before any useful info has been discarded, to do a bulk edit of the surviving vocabulary (maybe both adding & removing words). Or it could be called once per word, and in such a case might offer a sort of 3-valued decision: 'must keep', 'must discard', or 'do default'. 

Does your current local patch take advantage of the scan_vocab() / scale_vocab() / finalize_vocab() split?

piskvorky:
If we extend this functionality, we'll definitely go with something as general as possible (callable).

It doesn't look complex, and seems fully backward compatible, so I'm in favour of this.

Re. the callable API: again, something as generic as possible. How about a signature of `(word=unicode, freq=int) -> bool`? Or the 3-valued decision @gojomo proposes: `(word=unicode, freq=int) -> {KEEP=0, DISCARD=1, DEFAULT=2}`.

phdowling:
I personally think 3-valued output would be good, it makes it easier to add special case rules without having to also handle the normal case each time.

[...]

PR #434, opened 2015-08-15
Adds an optional vocab trimming rule parameter to word2vec

phdowling:
This addresses #423, basically implementing the per-word three-return-value option @gojomo described. I've tested this locally and it seems to be working fine, but I think the test for saving a model was still failing. Opening this PR now so I can get some input on it from you guys.

Example usage (wiki2vec for dbpedia spotlight):
```python
def read_corpus(path_to_corpus, output_path, min_count=10, size=500, window=10, entity_min_count=5):
    sentences = gensim.models.word2vec.LineSentence(path_to_corpus)

    def rule(word, count, min_count):
        if word.startswith("DBPEDIA_ID/") and count >= entity_min_count:
            return 2 # Keep this word
        else:
            return 0 # Don't care. Do whatever the default is (i.e. min_count is applied later)

    model = gensim.models.Word2Vec(None, min_count=min_count, size=size, window=window, sg=1, trim_rule=rule)
    model.build_vocab(sentences)
    model.train(sentences)
    model.trim_rule = None  # so we can save
    model.save(output_path)
```

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -499,7 +499,7 @@ class Doc2Vec(Word2Vec):
     def __init__(self, documents=None, size=300, alpha=0.025, window=8, min_count=5,
                  max_vocab_size=None, sample=0, seed=1, workers=1, min_alpha=0.0001,
                  dm=1, hs=1, negative=0, dbow_words=0, dm_mean=0, dm_concat=0, dm_tag_count=1,
-                 docvecs=None, docvecs_mapfile=None, comment=None, **kwargs):
+                 docvecs=None, docvecs_mapfile=None, comment=None, trim_rule=None, **kwargs):
         """
         Initialize the model from an iterable of `documents`. Each document is a
         TaggedDocument object that will be used for training.
@@ -569,7 +569,7 @@ class Doc2Vec(Word2Vec):
         self.docvecs = docvecs or DocvecsArray(docvecs_mapfile)
         self.comment = comment
         if documents is not None:
-            self.build_vocab(documents)
+            self.build_vocab(documents, trim_rule=trim_rule)
             self.train(documents)
 
     @property
@@ -597,7 +597,7 @@ class Doc2Vec(Word2Vec):
         self.docvecs.borrow_from(other_model.docvecs)
         super(Doc2Vec, self).reset_from(other_model)
 
-    def scan_vocab(self, documents, progress_per=10000):
+    def scan_vocab(self, documents, progress_per=10000, trim_rule=None):
         logger.info("collecting all words and their counts")
         document_no = -1
         total_words = 0
@@ -622,7 +622,7 @@ class Doc2Vec(Word2Vec):
             total_words += len(document.words)
 
             if self.max_vocab_size and len(vocab) > self.max_vocab_size:
-                utils.prune_vocab(vocab, min_reduce)
+                utils.prune_vocab(vocab, min_reduce, trim_rule=trim_rule)
                 min_reduce += 1
 
         logger.info("collected %i word types and %i unique tags from a corpus of %i examples and %i words",
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -553,6 +553,13 @@ class Doc2Vec(Word2Vec):
         `dbow_words` if set to 1 trains word-vectors (in skip-gram fashion) simultaneous with DBOW
         doc-vector training; default is 0 (faster training of doc-vectors only).
 
+        `trim_rule` = vocabulary trimming rule, specifies whether certain words should remain
+         in the vocabulary, be trimmed away, or handled using the default (discard if word count < min_count).
+         Can be None (min_count will be used), or a callable that accepts parameters (word, count, min_count) and
+         returns either word2vec.RULE_DISCARD, word2vec.RULE_KEEP or word2vec.RULE_DEFAULT.
+         Note: The rule, if given, is only used prune vocabulary during build_vocab() and is not stored as part
+          of the model.
+
         """
         super(Doc2Vec, self).__init__(
             size=size, alpha=alpha, window=window, min_count=min_count, max_vocab_size=max_vocab_size,
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -556,7 +556,7 @@ class Doc2Vec(Word2Vec):
         `trim_rule` = vocabulary trimming rule, specifies whether certain words should remain
          in the vocabulary, be trimmed away, or handled using the default (discard if word count < min_count).
          Can be None (min_count will be used), or a callable that accepts parameters (word, count, min_count) and
-         returns either word2vec.RULE_DISCARD, word2vec.RULE_KEEP or word2vec.RULE_DEFAULT.
+         returns either util.RULE_DISCARD, util.RULE_KEEP or util.RULE_DEFAULT.
          Note: The rule, if given, is only used prune vocabulary during build_vocab() and is not stored as part
           of the model.
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -77,7 +77,7 @@ from copy import deepcopy
 from collections import defaultdict
 import threading
 
-from gensim.utils import _keep_vocab_item
+from gensim.utils import keep_vocab_item
 
 try:
     from queue import Queue, Empty
@@ -393,7 +393,7 @@ class Word2Vec(utils.SaveLoad):
         `trim_rule` = vocabulary trimming rule, specifies whether certain words should remain
          in the vocabulary, be trimmed away, or handled using the default (discard if word count < min_count).
          Can be None (min_count will be used), or a callable that accepts parameters (word, count, min_count) and
-         returns either word2vec.RULE_DISCARD, word2vec.RULE_KEEP or word2vec.RULE_DEFAULT.
+         returns either util.RULE_DISCARD, util.RULE_KEEP or util.RULE_DEFAULT.
          Note: The rule, if given, is only used prune vocabulary during build_vocab() and is not stored as part
           of the model.
 
@@ -546,7 +546,7 @@ class Word2Vec(utils.SaveLoad):
         drop_unique, drop_total, retain_total, original_total = 0, 0, 0, 0
         retain_words = []
         for word, v in iteritems(self.raw_vocab):
-            if _keep_vocab_item(word, v, min_count, trim_rule=trim_rule):
+            if keep_vocab_item(word, v, min_count, trim_rule=trim_rule):
                 retain_words.append(word)
                 retain_total += v
                 original_total += v
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -514,7 +514,7 @@ class Word2Vec(utils.SaveLoad):
         sentence_no = -1
         total_words = 0
         min_reduce = 1
-        _modified_trim_rule = lambda (word, count, min_count): keep_vocab_item(word, count, min_count, self.trim_rule)
+        _modified_trim_rule = lambda word, count, min_count: keep_vocab_item(word, count, min_count, self.trim_rule)
         vocab = defaultdict(int)
         for sentence_no, sentence in enumerate(sentences):
             if sentence_no % progress_per == 0:
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -338,7 +338,8 @@ class Word2Vec(utils.SaveLoad):
     def __init__(
             self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
             max_vocab_size=None, sample=0, seed=1, workers=1, min_alpha=0.0001,
-            sg=1, hs=1, negative=0, cbow_mean=0, hashfxn=hash, iter=1, null_word=0):
+            sg=1, hs=1, negative=0, cbow_mean=0, hashfxn=hash, iter=1, null_word=0,
+            trim_rule=None):
         """
         Initialize the model from an iterable of `sentences`. Each sentence is a
         list of words (unicode strings) that will be used for training.
@@ -413,6 +414,21 @@ class Word2Vec(utils.SaveLoad):
         self.null_word = null_word
         self.train_count = 0
         self.total_train_time = 0
+
+        RULE_DEFAULT = 0
+        RULE_DISCARD = 1
+        RULE_KEEP = 2
+
+        if trim_rule:
+            def keep_vocab_item(word, count, min_count_):
+                rule_res = self.trim_rule(word, count)
+                default_res = count >= min_count_
+                return True if (rule_res == RULE_KEEP) else (False if (rule_res == RULE_DISCARD) else default_res)
+        else:
+            def keep_vocab_item(word, count, min_count_):
+                return count >= min_count_
+        self.keep_vocab_item = keep_vocab_item
+
         if sentences is not None:
             if isinstance(sentences, GeneratorType):
                 raise TypeError("You can't pass a generator as the sentences argument. Try an iterator.")
@@ -499,7 +515,7 @@ class Word2Vec(utils.SaveLoad):
                 vocab[word] += 1
 
             if self.max_vocab_size and len(vocab) > self.max_vocab_size:
-                total_words += utils.prune_vocab(vocab, min_reduce)
+                total_words += utils.prune_vocab(vocab, min_reduce, rule=self.keep_vocab_item)
                 min_reduce += 1
 
         total_words += sum(itervalues(vocab))
@@ -535,7 +551,7 @@ class Word2Vec(utils.SaveLoad):
         drop_unique, drop_total, retain_total, original_total = 0, 0, 0, 0
         retain_words = []
         for word, v in iteritems(self.raw_vocab):
-            if v >= min_count:
+            if self.keep_vocab_item(word, v, min_count):
                 retain_words.append(word)
                 retain_total += v
                 original_total += v
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -76,7 +76,9 @@ from timeit import default_timer
 from copy import deepcopy
 from collections import defaultdict
 import threading
-import time
+
+from gensim.utils import _keep_vocab_item
+
 try:
     from queue import Queue, Empty
 except ImportError:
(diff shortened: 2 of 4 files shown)


piskvorky:
Thanks @phdowling !

Python cannot pickle anonymous functions (nested methods, lambdas etc), which explains the errors. Move the `rule` function to top (non-anymous) scope to fix this.

gensim/models/word2vec.py
```diff
@@ -413,6 +414,21 @@ def __init__(
         self.null_word = null_word
         self.train_count = 0
         self.total_train_time = 0
+
+        RULE_DEFAULT = 0
+        RULE_DISCARD = 1
+        RULE_KEEP = 2
+
+        if trim_rule:
+            def keep_vocab_item(word, count, min_count_):
```
piskvorky: same here; the `self.trim_rule` attribute is not pickleable if it contains such anonymous functions or lambdas.

Also, why the `_` at the end of `min_count_`?

gensim/models/word2vec.py
```diff
@@ -413,6 +414,21 @@ def __init__(
...
+                rule_res = trim_rule(word, count, min_count)
+                default_res = count >= min_count_
+                return True if (rule_res == RULE_KEEP) else (False if (rule_res == RULE_DISCARD) else default_res)
```
piskvorky: Break into simpler individual `if`s, for better readability.

phdowling:
Done, also fixed a subtle bug related to the trailing underscore. About the pickleability: You're saying I should leave the param as it is, but instead of having `self.trim_rule`, the function should be defined on the module level and not accessed through `self`?

piskvorky:
That's right. Pickle only stores function references (names), so if the function is anonymous, it doesn't know how to store it and throws an exception.

This is true for any functions you want to store as attributes, not just `self.trim_rule`. They have to have a globally importable name.

phdowling:
@piskvorky I changed some stuff and added a test case for persistence and vocabulary membership with the trimming rule. All tests seem to be passing now, I haven't built a whole model yet but I have no reason to assume things would have broken now. Could you quickly check out the few changes in the last two commits and let if know if you see any problems?

phdowling:
The tests are basically fine. I think I don't totally understand what you are picturing for moving the rule to a non-anonymous scope, I don't really see an easy way to do that without either losing the ability to actually save the rule persistently (which is sort of what my current solution does, I just override a global function with the anonymous definition you see, which makes everything pickleable since the reference is well-defined but probably won't re-load the rule later on), or going with a `(KEEP, DISCARD)` implementation rather than `(KEEP, DISCARD, DEFAULT)`, since we need to combine the standard `min_count` functionality with the user specified rule, but want to avoid the user having to worry about applying min_count themselves. 

Could you elaborate a bit on how this should be implemented? Also, do we really need the rule to be persisted, since it's only really used at the time of vocabulary creation?

phdowling:
Here is an idea: we persist the function that is passed in using the `trim_rule` parameter (which just requires it to be a normally defined module-level function, but in order to be able to pass the "modified" version that combines the user-specified rule and the `min_count` functionality, we create an anonymous function dynamically and pass that to the vocab trimming functions.

phdowling:
Ah, I think I've got it now. I changed the implementation now such that it is as described in the last comment, but with a function on module level instead on an anonymous function, and I think the way it is now is probably close to what you had in mind. Tests are passing, let me know if this works for you!

piskvorky:
If we don't persist the decision function, then pickling is a non-issue, yes. It would also mean we could accept lambda functions for the rule (not just named, importable functions).

If we go with the "store the function too, for later (re)use" option, it has to be a non-anonymous function:

```python
def default_top_level_fnc(param):
    return X

class ClassA(object):
    def __init__(self, pickleable_fnc=default_top_level_fnc):
        self.pickleable_fnc = pickleable_fnc

    def compute_default(self, x):
        y = self.pickleable_fnc(x)
        if y == 2:
             ...

default_a = SomeClass()

...

def another_top_level_fnc_somewhere(param):
    return Y

custom_a = SomeClass(pickleable_fnc=another_top_level_fnc_somewhere)
```

phdowling:
Okay, yeah this is what I meant by the "can't return DEFAULT" problem, since the user specified function would also need to handle that case. I think I got around the issue, but there was another failing test just now so I'll take another look.

phdowling:
Tests are looking good now, what do you think?

gensim/models/word2vec.py
```diff
@@ -327,6 +327,26 @@ def __str__(self):
...
+RULE_DISCARD = 1
+RULE_KEEP = 2
+
+
+def keep_vocab_item(word, count, min_count, trim_rule=None):
+    if trim_rule:
+        rule_res = trim_rule(word, count, min_count)
+    else:
+        rule_res = RULE_DEFAULT
+    default_res = count >= min_count
```
piskvorky: no need to evaluate this for RULE_KEEP or RULE_DISCARD?

gensim/models/word2vec.py
```diff
@@ -338,7 +358,8 @@ class Word2Vec(utils.SaveLoad):
     def __init__(
             self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
             max_vocab_size=None, sample=0, seed=1, workers=1, min_alpha=0.0001,
-            sg=1, hs=1, negative=0, cbow_mean=0, hashfxn=hash, iter=1, null_word=0):
+            sg=1, hs=1, negative=0, cbow_mean=0, hashfxn=hash, iter=1, null_word=0,
+            trim_rule=None):
```
piskvorky: new param `trim_rule` needs good docs in the docstring

gensim/test/test_word2vec.py
```diff
@@ -50,6 +50,11 @@ def testfile():
     # temporary data will be stored to this file
     return os.path.join(tempfile.gettempdir(), 'gensim_word2vec.tst')
 
+def rule_for_testing(word, count, min_count):
+    if word == "human":
+        return 1  # throw out
```
piskvorky: use the defs from word2vec module (`RULE_DEFAULT` etc); that's what they're for :)

gojomo:
These pickle issues are the "extra serialization/deserialization gotchas" I'd worried about for the most-flexible option. 

(IIUC, for pickling the important thing is that the function that's assigned into the variable is defined via a global name; then the serialization is essentially: "find and use the global with this name". The configuration could even then be: supply the string name of the function to look-up in globals and use, rather than the function itself.)

I like flexibility but a little concerned if this approach makes the one most certain/demonstrated need – a firm list of words that should survive no matter their frequency – harder. Does this approach work OK for your DBpedia-driven use case? (Are you having to use two globals, one for the function and one for the data-driven list of protected words?)

phdowling:
@gojomo: My issues are solved by this implementation, see the example I posted in the very first comment. I think the scenario where there is a list is also solved by this without too much effort, the alternative (specifying a container) seems a bit messy to me. 

I guess the remaining consideration might be whether we want to serialize the rule function or not. I personally don't see a lot of scenarios where this would be necessary, and there is a slight, possibly unwanted side effect for the user in either implementation: either you can't use anonymous functions and lambdas, or you can't recover the rule callable after the serialization. I think the latter is probably less of a downside, I would think users very rarely save the model before even building the vocabulary.

@piskvorky: I took your comments into account, just need to fix the documentation issue. I'll hold on committing this stuff until there's a consensus on the whole serializiation issue (and whatever else remains).

piskvorky:
Thanks @phdowling . Re. storing the function or not -- if it's not needed, then we can avoid saving it, yes. Will give us more flexibility.

I can't think of scenarios where we might need the same function in the future, after `load`, but that may be just lack of my imagination... @cscorley @gojomo thoughts?

gojomo:
Things like the 'online word2vec' work need to expand the vocab at future times... and so might need to repeat the decisions of the initial `build`. (Though of course policies/thresholds would likely be different in such a situation.) Perhaps the presence of a non-pickleable value could be detected and generate a warning (rather than aborting the whole entire operation)?

piskvorky:
You mean generate a warning and then continue? That doesn't sound like a good idea. Better fail explicitly.

I'm leaning toward not storing the decision function. Like you say, with online training, the user has to solve this some other way anyway (or pass the same non-storable lambda again).

phdowling:
If we document the fact that the function is not saved well, I think that's better than generating a warning it will not be saved in some cases. That way the user has to write the same sort of code no matter what sort of callable they're passing in and we always have the same behaviour. The latter way seems like it would be more confusing for me, since the non-pickleability of lambdas and such is not something everyone would be intuitively aware of. Explicit instructions in the docs to re-supply the callable when running more training a) is more straightforward and b) probably affects less people. You could even give a warning on re-training that no rule callable was supplied if there was one in the previous round.

gojomo:
Note that "warn-and-proceed" is functionally the same as a policy of "just-don't-save" – but it generates a loud log event that the user should beware unexpected change in behavior after next-load, rather than counting on them reading docs. 

On the other hand, maybe the configuration should just be a "string: global function name" – that's always serializable, and would generate a clear error if deserialized into a (later) codebase where the expected function isn't globally-defined.

piskvorky:
Ah, I thought you meant "detect if unpickle: yes => warn; no => store". I'm -1 on that, that's too confusing. 

Or did you mean "detect if unpickleable: => yes warn; no => ?". Then what is the "?".

I'm -1 on refer-to-functions-as-strings. I don't see the benefit, I think the unpickle error is clear enough (`X object has no attribute Y` or somesuch).

gojomo:
A suggestion was made that maybe the function need not be serialized at all. My concern is that (even if documented) that might lead to surprise, if a serialize-deserialize cycle changes the previously-configured behavior. Generally, the assumption/contract is that if serialization silently succeeds, you get "the same" object back on deserialization. So this policy, "prevent even the possibility of failure by not even trying", seems bad to me. 

Compared to that policy, something that seems strictly more user-friendly and less error-prone to me would be: (1) check if serializable, and if so, serialize as requested; (2) if not serializable, loudly warn-log that it can't be serialized, but continue with serialization as if that property were unset. After all, in the case where it's thought OK to maybe not-serialize at all, that's because the user may not need the rule in the future. And thus the user's workaround, if they hit a blocking error, would probably be to undefine the rule and try again. The warn-and-serialize-what-you-can just does that, automatically, and so it's probably OK, or at least superior to not-even-trying. 

That leaves the hard-fail: try serialization, but if the rule is unserializable, just fail entirely until the user fixes the issue. This also has much to recommend it. I don't have a strong pref between this behavior (for its simplicity) and the plow-ahead-with-best-possible-after-warning (for its likely match with the common user need). 

One last thought: so that the case of a specific "white list" of untrimmable words can be supported without recourse to more globals, the rule-function signature woudl probably benefit from taking the 'model' itself as a parameter. Then something like a list-of-keep-words could be added as an optional model property, and be consulted by the rule. (And if the rule-function is a global, but the list a serializable property of the model, the functionality survives serialize-deserialize roundtrips.)

piskvorky:
I think the suggestion was to only pass this function to wherever it's explicitly needed (build vocab, online training etc), never serializing it. In other words, it's not part of the object contract, it's a runtime property and the user's responsibility to pass it in where needed, as needed. I like this option the best.

I still don't like doing two different things depending on whether the rule is pickleable or not (and just logging a warning if it's not). Many users don't have logging on (or even know it exists).

If we go down that route, it's probably better to add an extra dependency like [dill](https://pypi.python.org/pypi/dill) to the serialize code directly, even with all dill's caveats.

gojomo:
Ah, I see. I like pass-it-in-when-needed best, too, sidestepping the whole issue. (And, something composed and passed-in as needed can easily enclose its own necessary state, so no need for the signature to include the model or other sources of state.)

phdowling:
Sounds good to me also, that's probably the least confusing way to do it. I would add the optional `trim_rule` parameter to `build_vocab`, `scan_vocab`, `scale_vocab`, and keep it local to those functions (or passed to the next function in the case of `build_vocab`). 

Should we also keep it in the constructor? That way a user could conveniently add sentences and have the vocab built with a rule applied. Only downside I see would be that some users might believe the rule would be persisted, but I think that sort of argument could be made about a lot of things (`sentences` also gets passed into the constructor but obviously isn't persisted).

piskvorky:
Yes, I think explaining the non-storage in the constructor's docstring ought to be enough.

phdowling:
I've updated this to use a passed in rule now and added some test cases. Also wrote doc for it - is this good?

gensim/models/word2vec.py
```diff
@@ -387,6 +409,12 @@ def __init__(
 
         `iter` = number of iterations (epochs) over the corpus.
 
+        `trim_rule` = vocabulary trimming rule, specifies whether certain words should remain
+         in the vocabulary, be trimmed away, or handled using the default (discard if word count < min_count).
+         Must be a callable that accepts a tuple (word, count, min_count) and returns
```
piskvorky: It doesn't have to be a callable; in fact, the doc doesn't describe the most common option, which is leaving it at `None` (optional param).

gensim/models/word2vec.py
```diff
@@ -387,6 +409,12 @@ def __init__(
...
+         either word2vec.RULE_DISCARD, word2vec.RULE_KEEP or word2vec.RULE_DEFAULT.
+         Note: the rule is NOT serialized.
```
piskvorky: "NOT serialized" is too obscure. How about "the rule, if given, is only used prune vocabulary during `build_vocab()` and is not stored as part of the model."?

gensim/models/word2vec.py
```diff
@@ -474,22 +503,27 @@ def create_binary_tree(self):
...
-    def build_vocab(self, sentences, keep_raw_vocab=False):
+    def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None):
         """
         Build vocabulary from a sequence of sentences (can be a once-only generator stream).
         Each sentence must be a list of unicode strings.
 
         """
-        self.scan_vocab(sentences)  # initial survey
-        self.scale_vocab(keep_raw_vocab)  # trim by min_count & precalculate downsampling
+        if trim_rule is not None:
```
piskvorky: Is this `if` necessary? Wouldn't the branches work the same way, regardless of `None` or not?

phdowling: A doc2vec test fails since doc2vec uses this method. The alternative is to add the rule parameter to doc2vec as well, but I thought that would be best left to another PR.

gensim/models/word2vec.py
phdowling: How about ````trim_rule` = vocabulary trimming rule, specifies whether certain words should remain
         in the vocabulary, be trimmed away, or handled using the default (discard if word count < min_count).
         Can be None (min_count will be used), or a callable that accepts parameters (word, count, min_count) and
         returns either word2vec.RULE_DISCARD, word2vec.RULE_KEEP or word2vec.RULE_DEFAULT.
         Note: The rule, if given, is only used prune vocabulary during build_vocab() and is not stored as part
          of the model.```

gensim/models/word2vec.py
gojomo: I'd be happy to do the Doc2Vec adaptation, but since Doc2Vec shares so much of the vocab-scanning code, and it looks like it's only a few lines, it'd be good to keep it at parity in the same PR. Looking at changes that'd require (mainly `Doc2Vec.scan_vocab()`), how about moving `_keep_vocab_item` into utils, to eliminate the need for a `_modified_trim_rule` wrapper and ensure parameters named `trim_rule` have the same signature/return-values wherever they are used?

phdowling:
@gojomo @piskvorky I adjusted the doc2vec part as well now.

phdowling:
Hey, how are we looking here - any more issues to fix or is this good to go?

gensim/models/doc2vec.py
```diff
@@ -553,6 +553,13 @@ def __init__(self, documents=None, size=300, alpha=0.025, window=8, min_count=5,
         `dbow_words` if set to 1 trains word-vectors (in skip-gram fashion) simultaneous with DBOW
         doc-vector training; default is 0 (faster training of doc-vectors only).
 
+        `trim_rule` = vocabulary trimming rule, specifies whether certain words should remain
+         in the vocabulary, be trimmed away, or handled using the default (discard if word count < min_count).
+         Can be None (min_count will be used), or a callable that accepts parameters (word, count, min_count) and
+         returns either word2vec.RULE_DISCARD, word2vec.RULE_KEEP or word2vec.RULE_DEFAULT.
```
piskvorky: `RULE_*` moved from `word2vec` to `utils`

gensim/models/word2vec.py
```diff
@@ -387,6 +390,13 @@ def __init__(
 
         `iter` = number of iterations (epochs) over the corpus.
 
+        `trim_rule` = vocabulary trimming rule, specifies whether certain words should remain
+         in the vocabulary, be trimmed away, or handled using the default (discard if word count < min_count).
+         Can be None (min_count will be used), or a callable that accepts parameters (word, count, min_count) and
+         returns either word2vec.RULE_DISCARD, word2vec.RULE_KEEP or word2vec.RULE_DEFAULT.
```
piskvorky: Ditto.

gensim/utils.py
```diff
@@ -1099,9 +1100,29 @@ def prune_vocab(vocab, min_reduce):
...
                 old_len - len(vocab), min_reduce, old_len, len(vocab))
     return result
+
+
+RULE_DEFAULT = 0
+RULE_DISCARD = 1
+RULE_KEEP = 2
+
+
+def _keep_vocab_item(word, count, min_count, trim_rule=None):
```
piskvorky: Why the leading underscore? Let's make it a normal function.

gensim/utils.py
```diff
@@ -1099,9 +1100,29 @@ def prune_vocab(vocab, min_reduce):
...
+    if trim_rule is None:
+        return default_res
+    else:
+        rule_res = trim_rule(word, count, min_count)
+        if rule_res == RULE_KEEP:
+            return True
+        elif rule_res == RULE_DISCARD:
+            return False
+        else:
+            return default_res
```
piskvorky: Missing newline at the end of file :)

piskvorky:
A few minor code comments :)

But I think we're ready to merge, thanks! Let me know if you have time to fix the nitpicks, otherwise I'll merge now and fix them later (plus add a test for "rule entered via anonymous lambda").

phdowling:
Thanks, I took care of the fixes and added a test case. Let's see that all tests pass on Travis and we should be fine!

phdowling:
There is a failing test but it seems unrelated to my code from what I can see, correct?

gojomo:
Correct. I manually triggered a re-run and it passed. (It's still on my list to make the test more reliable.)

piskvorky:
Merging.

Thanks so much @phdowling ! I see you also added the lambda test, brilliant :)

PR #435, opened 2015-08-17
online word2vec

rutum:
Usage:

model = Word2Vec() # sg and hs are the default parameters
model.build_vocab(sentences)
model.train(sentences)
model.save("base_model")

model.build_vocab(new_sentences, update=True)
model.train(new_sentences)
model.save("updated_model")

Then you can compare the 2 models to see whether the new vocabulary is learning the way it is supposed to.

I tried an experiment with learning a model without "queen", and adding it in the subsequent set of sentences. The updated model learned "queen" as being similar to "king", "duke" etc. So that was a huge success. I would love to hear of any other ideas you might have to test this. 

Accompanying blogpost: http://rutumulkar.com/blog/2015/word2vec/


rutum:
Here is the accompanying blog post about how to use Online Word2Vec:
http://rutumulkar.com/blog/2015/word2vec/

robhawkins:
I see you consolidated update_vocab into build_vocab with another parameter.  I think that makes sense to keep code concise but I liked your original description.  For people new to gensim like myself, I think it'd be nice to have an update_vocab that wraps build_vocab(update=true).  Great work!

rutum:
Any ideas why this branch is failing on Python 3.3 only?

rutum:
@robhawkins Sure! But lets wait for the PR process to see what others recommend

piskvorky:
The fail is unrelated; @gojomo is working on fixing that.

piskvorky:
Btw six commit for resolving conflicts? That's scary. What's going on there?

Can you at least squash all these commits into one? We don't want such verbose git history.

gensim/models/word2vec.py
```diff
@@ -1444,7 +1510,6 @@ def __init__(self, init_fn, job_fn):
     def put(self, job):
         self.job_fn(job, self.inits)
 
-
```
piskvorky: There *should* be two empty lines between classes; why did you delete it?

gensim/models/word2vec.py
```diff
@@ -1409,6 +1474,7 @@ def save(self, *args, **kwargs):
 
     save.__doc__ = utils.SaveLoad.save.__doc__
 
+
```
piskvorky: Only one newline between methods.

gensim/models/word2vec.py
```diff
@@ -535,28 +539,51 @@ def scale_vocab(self, min_count=None, sample=None, dry_run=False, keep_raw_vocab
...
+                    original_total += v
+                    if not dry_run:
+                        self.vocab[word] = Vocab(count=v,
+                                    index=len(self.index2word))
+                        self.index2word.append(word)
+                else:
+                    drop_unique += 1
+                    drop_total += v
+                    original_total += v
+        elif update:
```
piskvorky: The condition is superfluous: if `not update` above evaluates to false, then `update` here is true for sure.

gensim/scripts/make_wiki.py
```diff
@@ -1 +0,0 @@
-make_wikicorpus.py
```
piskvorky: This is not good: you replaced all the symlinks by regular files. Can you put the symlinks back? This PR shouldn't touch the `make_wiki` functionality at all IMO.

rutum: Hmm! This is very strange. I have touched 3 files only, but it shows that I have updated 13 files. Possible merge issue?

piskvorky:
The failing Travis test is different this time: it looks like the online learning caused a segfault. That is a serious problem.

gensim/scripts/make_wiki.py
piskvorky: What is your merge process? I don't understand how all these whitespace, newline and unrelated file change issues can happen.

rutum:
Yeah! Looks like Python 3.4 is doing something differently. Any ideas?

gensim/scripts/make_wiki.py
rutum: I merge all conflicts manually, using Sublime, and then commit my changes. I have not come across any wiki related merge conflicts, so it happened automatically? The other space related ones are probably me, because of the manual merge conflict removal, but they seem harmless.

rutum:
All is well on Travis in the online Word2Vec land! Review away @piskvork, @gojomo and others

piskvorky:
I re-ran Travis and it still segfaults during the online training test. It looks like there is some subtle bug that causes a non-deterministic segfault.

gensim/scripts/make_wiki.py
piskvorky: When there's a conflict, git clearly shows you which parts are in conflict: it puts the pieces of code between `>>>>>` and `<<<<<<` markers.

I'm not sure what happened here or why you replaced the symlinks, but we definitely want to keep the original symlinks (and not full files, as in this PR).

This is also connected to debugging -- the fewer the changes, the smaller the surface where the bug may have been introduced. The segfault will be easier to debug with a smaller PR.

rutum: Agreed. However, I don't know how this changed happened with the sym links. I have not touched these files, and I am happy to reverse this change, if I knew how to.

piskvorky: OK, let's do that after fixing the segfault. Compared to the segfault, this is a trivial fix.

rutum:
I got some time to debug this, and it is a non-deterministic error. The `syn[0]` array ends up with `nan` values when training after new vocabulary is added. 

`word2vec.py:174: RuntimeWarning: overflow encountered in exp
  fa = 1.0 / (1.0 + exp(-dot(l1, l2a.T)))  # propagate hidden -> output
`
So clearly, there was something going on with the way the numpy arrays are referenced. It looks like I wrote some code to mane a shallow copy of the NP arrays, instead of a deep copy. I updated and tested it, and have not hit upon the segfault yet.

gojomo:
In the cython work I did, intermittent `nan`-related crashes of this type were sometimes due to earlier mis-addressed writes, which corrupted other arrays. 

If `deepcopy` has really prevented the errors, I'm concerned it may be covering them up rather than fixing the real error. (Perhaps it's coercing `nan` to 0?) The pre-deepcopy property-reassignment should have worked fine, and avoids both a large extra-allocation and a large memory-copy – so I'd revert to that so to be sure to trigger the problem until it's definitively fixed. 

I would focus on earlier steps, especially the initial expand/update-weights. (I see one potential problem L967 – will add line comment there.)

gensim/models/word2vec.py
```diff
@@ -921,6 +951,39 @@ def worker_loop():
...
+        newsyn0 = empty((len(self.vocab), self.vector_size), dtype=REAL)
+
+        # copy the weights that are already learned
+        for i in xrange(0, len(self.syn0)):
+            newsyn0[i] = deepcopy(self.syn0[i])
+
+        # randomize the remaining words
+        for i in xrange(len(self.vocab), len(newsyn0)):
+            # construct deterministic seed from word AND seed argument
+            self.syn0[i] = self.seeded_vector(self.index2word[i] + str(self.seed))
```
gojomo: Seems this should start at the *old* vocab length, and the newly-seeded vectors should be landing in the `newsyn0`? (Doubt this could cause segfaults as out-of-bounds access should be caught here, but true cause might be something similar elsewhere...)

rutum:
@gojomo The error is non-deterministic, which is why I think there was some `malloc` issue which was hard to reproduce. If I do not do a deep copy, then an index in `syn0` will still reference to an index in `newsyn0`, but I think python3 handles numpy references and numpy copies differently than 2.6 and 2.7. Anyway, this is what I think the issue was. Let me know if you can reproduce this.

gojomo:
If the issue is corruption from earlier bad writes, whether that corruption is fatal or harmless can be affected by the (effectively) arbitrary locations chosen by earlier mallocs, yes. So that could explain the indeterminism. But if that's the case, I don't yet see any way a deepcopy can fix the real problem.

All indexed-accesses to `syn0`, eg `syn0[i]`, will always reach into the current value of `syn0`, so there's no risk of misdirected accesses from doing `self.syn0 = newsyn0`. I'm pretty confident that's the right swap to make, for speed and memory efficiency.

If the deepcopy is seemingly resolving the problem (or just making it must less frequent), it may be by simply hiding it. The big reallocation might guarantee the array is further from other arrays, so out-of-bounds writes are damaging other memory. 

The tactic I mentioned the other day, to try the cython code with bounds-checkling back on (a directive atop the file), *might* help catch a real corrupting write when it happens, rather than when later calculations 'step' on it again. If you're not set up to modify/recompile the cython parts, I'll have a chance to try that in the next day or so...

tmylk:
@rutum Are you still working on this? Or should this be closed?
Thanks

skillachie:
This feature looks good.  Do you guys plan to merge it into the main dev/main branch soon ?

danintheory:
Hi,

I am very interested in this feature. What would be needed to get this or a new pull request accepted?

Thanks!
Dan

si74:
Hi I also wanted to know if you were planning on merging this soon! :)

piskvorky:
This PR is not mergeable yet:

* segfaults
* contains changes in unrelated files (symlinks)
* merge conflicts against the current `develop` branch
* needs a deeper review that it works -- see code comments by @gojomo in this PR

@danintheory the most important one is fixing the segfault. I think starting from a current `develop` checkout, and re-applying the relevant @rutum 's changes, may be easier than resolving the conflicts, due to the commit history.

gensim/scripts/make_wiki.py
zachmayer: See my update to this PR: https://github.com/piskvorky/gensim/pull/615

@rutum Something weird happened when you merged the dev branch into your branch, and you ended up replacing all the symlinks with their actual files.

zachmayer:
FYI: I rebased this PR and fixed merge conflicts here: https://github.com/piskvorky/gensim/pull/615.

I suggest we close this PR, and move the discussion over there.

piskvorky:
Continued in #615 .

adarshaj:
For posterity, this has been merged to develop on #900

PR #441, opened 2015-08-29
Summarisation optimisations

erbas:
Added speed improvements to keyword and summarisation methods.

### DIFF
--- a/gensim/summarization/keywords.py
+++ b/gensim/summarization/keywords.py
@@ -163,8 +163,7 @@ def _get_combined_keywords(_keywords, split_text):
                 result.append(word)   # appends last word if keyword and doesn't iterate
             for j in xrange(i + 1, len_text):
                 other_word = _strip_word(split_text[j])
-                # if other_word in _keywords and other_word == split_text[j].decode("utf-8"):
-                if other_word in _keywords and other_word == split_text[j]:
+                if other_word in _keywords and other_word == split_text[j].decode("utf-8"):
                     combined_word.append(other_word)
                 else:
                     for keyword in combined_word:
--- a/gensim/summarization/keywords.py
+++ b/gensim/summarization/keywords.py
@@ -32,6 +32,7 @@ def _get_words_for_graph(tokens, pos_filter):
         include_filters, exclude_filters = _get_pos_filters()
     else:
         include_filters = set(pos_filter)
+        exclude_filters = frozenset([])
     if include_filters and exclude_filters:
         raise ValueError("Can't use both include and exclude filters, should use only one")
 
--- a/gensim/summarization/keywords.py
+++ b/gensim/summarization/keywords.py
@@ -11,6 +11,8 @@ from gensim.summarization.commons import remove_unreachable_nodes as _remove_unr
 from itertools import combinations as _combinations
 from six.moves.queue import Queue as _Queue
 from six.moves import xrange
+from six import iteritems
+
 
 WINDOW_SIZE = 2
 
@@ -37,7 +39,7 @@ def _get_words_for_graph(tokens, pos_filter):
         raise ValueError("Can't use both include and exclude filters, should use only one")
 
     result = []
-    for word, unit in tokens.iteritems():
+    for word, unit in iteritems(tokens):
         if exclude_filters and unit.tag in exclude_filters:
             continue
         if (include_filters and unit.tag in include_filters) or not include_filters or not unit.tag:
@@ -116,7 +118,7 @@ def _extract_tokens(lemmas, scores, ratio, words):
 
 def _lemmas_to_words(tokens):
     lemma_to_word = {}
-    for word, unit in tokens.iteritems():
+    for word, unit in iteritems(tokens):
         lemma = unit.token
         if lemma in lemma_to_word:
             lemma_to_word[lemma].append(word)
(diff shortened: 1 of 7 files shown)


piskvorky:
Looks good! Isn't there a way to make use of sparsity of the adjacency matrix? That would save us a LOT of memory and allow processing much larger inputs.

@fbarrios is the adjacency matrix supposed to be symmetric or not? I didn't check the algo, but I see @erbas added some comments around complex numbers, which would suggest it's not symmetric...?

erbas:
Well, it's packed efficiently in the csr_matrix format, but then unpacked and made dense at full dimension by the addition of the `probability_matrix`.  I don't have good theory on this but debugging it looked like the `adjacency_matrix` was not symmetric.

To really scale this up I think you need a different algorithmic approach.  All I've done is some obvious optimisations.  Some form of count-min-sketch is probably the way to go.

piskvorky:
Yes, unless we find a way to make use of adjacency matrix' sparsity, there's no point in the CSR step. We should simply construct it as a dense array to start with, and avoid the pointless conversions (wasting memory and CPU).

Can you say a little more around what `WEIGHT_THRESHOLD` does? What are the tradeoffs here? Worth making it a tunable parameter (rather than a hardwired constant)? Maybe add docs?

piskvorky:
I looked at `graph` and as far as I can see, it should be undirected. Which means the adjacency matrix should be symmetrical, so we can use `eigsh`. @fedelopez77 @fbarrios can you please confirm?

erbas:
Yes, I agree with your analysis of the code.  The graph should be undirected, but when I step through the code the adjacency matrix is not symmetric.  Could be a bug?  Could be "conceptually symmetric", i.e. numerical issues mean it's not symmetric to many decimal places?  The latter seems unlikely to me as the differences were ~ 1, not 1.e-4.

You asked about `WEIGHT_THRESHOLD`.  The smallest non-zero entry in the matrix for a small sample of random texts was  ~ 0.1, so 0.001 seemed a reasonable threshold.  Perhaps it's safer to set it to zero?  I can't see much advantage in making it user tuneable.

piskvorky:
Let's wait for @fedelopez77's comment, so we don't get ahead of ourselves :)

fedelopez77:
I'm sorry I can't confirm this until I get home and run some tests, but I think the matrix **is not** symmetric. 
In the original TextRank paper, it is symmetric, but in the implementation that was merged, we use a variation of the [Okapi BM25 function](https://en.wikipedia.org/wiki/Okapi_BM25) to set the edge's weights. Given that this function uses the results of the tf-idf function on the words of only one of the BM25 arguments, the results aren't the same (i.e. if A and B are sentences, BM25(A, B) != BM25(B, A).
And that's is also why we need a directed graph.

piskvorky:
Ah ok, that explains it. Cheers @fedelopez77 :)

erbas:
@fedelopez77 My unit tests are failing on Travis because different hardware/library combinations give different keyword lists.  I would not have expected the numerical differences between platforms to make such a difference, any suggestions?

piskvorky:
@erbas Maybe you're hitting some singular inputs (degenerate matrix spectrum)?

Otherwise we want the algo to be stable of course = small numerical imprecisions should have only proportional (small) impact on the result.

erbas:
@piskvorky  I'm developing on OS X and Travis is running linux inside Docker,  so I would expect small differences from different BLAS libraries at least.  But emphasis on "small".  The really odd error is the mismatch number of keywords: ask for 15, get 16.  I don't understand :(

piskvorky:
@erbas what's the status here? We'll make a new release this week, would be great to get this in.

erbas:
Hi @piskvorky, so it all works on python 2.  I don't understand unicode on python 3 well enough to understand the issue there.  The unit tests I wrote are failing because of the previously noted instabilities in keywords.  Again I'm not sure what's going on, I would have expected the largest eigenvector to be pretty robust.  Seems more likely to me that there's something in the calculation of the scores which is platform dependent.

piskvorky:
@tmylk any ideas? Can you check / fix?

tmylk:
Merged just a Keywords test with Linux keywords in [PR-474](https://github.com/piskvorky/gensim/pull/474) It passes in Python 2 in pre-BM25, pre-PR441 and post-PR441 on my Fedora and Travis docker. This shows that this PR hasn't deviated.

The instability due to platform/BLAS version is a separate issue but pretty important. 
For example, the test_keywords_ratio passes on my Fedora but fails on Travis.

And of course @erbas actually gets different keywords on OSX.
@erbas What is the BLAS and numpy versions where the tests are passing? Will try to reproduce it in some OSX VM.

erbas:
Hi @tmylk, I'm on numpy 1.9.2 and Xcode 7.1 for the BLAS (contained in the OS X Accelerate Framework).  I really don't understand the instability as the largest eigenvector should be robust, is it possible the matrix has different values on different platforms?

tmylk:
@erbas do you have any advice on how to get the keywords in this file test_data/mihalcea_tarau.kw.txt ?

I didn't have any luck producing them. Everywhere I try I get one different set of keywords. I created a test with this new set in #474 and it passes on unix'es and [OSX](https://travis-ci.org/tmylk/gensim-wheels/jobs/83860907) (python=2.7.9 numpy=1.9.2 Xcode=7.0) 

Thinking about removing mihalcea_tarau.kw.txt changes from this PR and using the ones in #474 instead.

erbas:
Nice work! I'm afraid I really don't have any better ideas. In some ways it's reassuring to think this sample text is somehow pathological for this algo but that does beg the question of which other texts might be problematic. Seems reasonable to me to replace the tests.

tmylk:
@erbas  All merged in. Thanks for the optimisations!

Ratio parameters in tests update to values more resistant to numeric fluctuations.
The cut-off used to be in the middle of a flat patch at 0.08 and that made it very unstable. There were differences even between consequent Travis runs.
scores

erbas:
@tmylk Thanks for digging into this, and sorry for not figuring it out :(

PR #444, opened 2015-09-02
Word2Vec iterator/generator train failure warning

### LINKED ISSUE #443: Word2Vec w/ iterable input, make sure not to exhaust generator
In attempting to use an iterator/generator combination as input to my [Word2Vec model](https://radimrehurek.com/gensim/models/word2vec.html#gensim.models.word2vec.Word2Vec), I ran into a similar gotcha as [this person](http://comments.gmane.org/gmane.comp.ai.gensim/4656) where my iterator wasn't re-initializing my generator.

I had an iterator that read from a generator, like this (simplified):

    def generator():
        for i in range(10):
            yield i

    class MyIterator(object):
        def __init__(self):
            self.generator = generator()

        def __iter__(self):
            return self

        def next(self):
            return self.generator.next()

BUT, it needed to be like this:

    def generator():
        for i in range(10):
            yield i

    class MyIterator(object):
        def __iter__(self):
            self.generator = generator()
            return self

        def next(self):
            return self.generator.next()

because Word2Vec training re-uses the iterator.

Of course, this is user error, but the resulting error that gets thrown by gensim is:

`UnboundLocalError: local variable 'job_no' referenced before assignment`

and that doesn't make my mistake very obvious. I suggest putting some sort of check in place that validates that if an iterator has elements the first time but not the second time, some sort of error indicating so is thrown.

loisaidasam:
Fixes #443

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -737,6 +737,7 @@ class Word2Vec(utils.SaveLoad):
         done_jobs = 0
         next_alpha = self.alpha
         jobs_source = enumerate(utils.grouper(sentences, chunksize))
+        job_no = -1
         # fill jobs queue with (sentence, alpha) job tuples
         while True:
             try:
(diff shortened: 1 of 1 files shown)


piskvorky:
Thanks for raising the issue, the current error message is really terrible!

The fix seems fine, but the message is still too cryptic. The documentation clearly says where an iterable vs. generator is needed (`__init__()` vs. `train()`). I think we should simply draw attention to that in the message, and possibly point them to a tutorial on these topics (such as [here](http://rare-technologies.com/data-streaming-in-python-generators-iterators-iterables/)).

gojomo:
Initializing `job_no` to -1 is a good way to avoid the unintended, cryptic error. 

For the detected-no-examples case, might a 'warn' log message be better than an exception? 

I can imagine a case where some pipeline of new example batches, repeatedly sent to `train()`,  legitimately filters an occasional batch of training examples down to an empty iterable/generator – and might expect that to be a no-op, allowing their enclosing code to proceed to future (non-empty) batches. 

The message could be a warning: "train() called with empty iterator. (If not intended, be sure to provide a corpus that offers restartable iteration.)" 

The warning (or exception) might also be limited to only fire when `train_count` is 0. (That's when it's most likely to be a user error using `__init__()`; if greater than 0 it may be a hypothetical incremental-training situation like above.)

loisaidasam:
Updating the wording as per @gojomo's suggestion. Also just issuing a warning.

piskvorky:
Merging, thanks a lot @loisaidasam !

Issue #389, opened 2015-07-07
show_topics method is not consistent across HdpModel and LdaModel


lucasoldaini:
I've noticed that the show_topics method return topics in a different format depending on the model used when `formatted=False`. 

For example, HdpModel returns a list of list in the following format:

```python
>>> model = gensim.models.HdpModel(corpus=corpus, id2word=dictionary)
>>> topics = model.show_topics(-1, formatted=False)
>>> print topics
[
    [0,
        [
            (u'apple', 0.11383073569595759),
            (u'pie', 0.10697087154306592),
            (u'crust', 0.10290230477535894),
            …
        ]
    ],
…
 ]
```

while LdaModel does not return the topic number and specifies probabilities as first term of the tuples

```python
>>> model = gensim.models.LdaModel(corpus=corpus, id2word=dictionary)
>>> topics = model.show_topics(-1, formatted=False)
>>> print topics
[
    [
         (0.11383073569595759, u'apple'),
         (0.10697087154306592, u'pie'),
         (0.10290230477535894, u'crust'),
         …
    ],
…
 ]
```
(numbers and terms are fictitious)

Any reason for the discrepancies?

-Luca

cscorley:
Indeed, this was brought up in #354. If you'd like to submit a PR, that'd be good :)

lucasoldaini:
sure, I'll take care of it soon.

tmylk:
Fixed in #448

Issue #354, opened 2015-06-02
LDA show_topic() might show wordids, not words


bestxolodec:
I mean this line:
https://github.com/piskvorky/gensim/blob/develop/gensim/models/ldamodel.py#L757 
should be 

```
beststr = [(topic[id], id) for id in bestn]
```

Otherwise it is not possible to get similar output to tfidf model, which shows list of tuples such as this

```
 (98801, 0.008706198772450975),
```

including   id of the word and its tfidf metric. 
Maybe such a behavior should rely on some function parameter, so it dont ruin any other related code which uses default word representation of `show_topic`.
I mean something like this:

``` (python)
    def show_topic(self, topicid, topn=10, show_ids=False):                         
        """                                                                         
        Return a list of `(words_probability, word)` 2-tuples for the most probable
        words in topic `topicid`.                                                   

        Only return 2-tuples for the topn most probable words (ignore the rest). 

        """                                                                         
        topic = self.state.get_lambda()[topicid]                                    
        topic = topic / topic.sum() # normalize to probability dist                 
        bestn = numpy.argsort(topic)[::-1][:topn]                                   
        if show_ids:                                                                
            beststr = [(topic[id], id) for id in bestn]                             
        else:                                                                       
            beststr = [(topic[id], self.id2word[id]) for id in bestn]               
        return beststr       
```

Nevertheless, even this approach is contradictory in the order of `id, metric` inside a tuple - tfidf shows id as a first parameter  and `lda.show_topics()` does it as a second.

cscorley:
I think that is something desirable, especially making sure that the `(id, metric)` is consistent across modules. Mind opening a PR?

BTW, there's no need to change the linked line. It isn't returned in the end for that method, anyway. I think you meant a few lines up, in `show_topic`: https://github.com/piskvorky/gensim/blob/develop/gensim/models/ldamodel.py#L734

menshikh-iv:
Resolved in #448.

PR #448, opened 2015-09-12
Make show_topics more consistent across models

cscorley:
This PR contains changes to address issues #354 and #389. Note that this change breaks compatibility for anyone that has come to rely on the format for a particular model.

Additional changes include splitting up the increasingly growing test_models.py file into individual files :)


cscorley:
FYI, I think the latest numpy will make these tests failing (at least on my local Python 3 environment). Seems that, at least for `LdaModel`, that the index of numpy arrays will be `numpy.int64` and no longer `int`. There's also some deprecation warnings being thrown from numpy related to how `LdaModel` indexes into arrays.

piskvorky:
Great! Can you also add a short summary to README, explaining whom this affects and how?

Re. failing tests: what's the problem with int64?

Indexing indexing: I know this one. It's due to [this line](https://github.com/piskvorky/gensim/blob/master/gensim/models/ldamodel.py#L559): that `as_numpy=True` forces `utils.grouper` to return documents as floats:

```python
doc = [(1, 0.5), (2, 0.3)]
print numpy.array(doc)

array([[ 1. ,  0.5],
       [ 2. ,  0.3]])
```

(the topic indexes become floats).

This was an optimization for sending data via Pyro, so that they serialize to smaller chunks. I think it's no longer needed => let's `as_numpy=False`.

cscorley:
w.r.t., int64: It should be handled now in the test explicitly, but numpy was returning indexes as int in Python 2 while numpy.int64 in Python 3. Very strange.

piskvorky:
@cscorley does this need anything else, or are we ready to merge? (after resolving the conflict)

piskvorky:
@cscorley we'll make a new release again this week. Any chance we could get this in?

cscorley:
Sorry @piskvorky, only really have the chance to work on weekends :) Luckily, the merge conflict was only in CHANGELOG.txt, so this should be ready to go.

piskvorky:
Perfect, thanks.

@tmylk please review & merge. This is a major, breaking change, so we have to be very explicit about this change in the release notes / announcement!

Probably warrants a major version bump too.

piskvorky:
@cscorley I see `as_numpy` is still True... is there a reason we don't want the indices to be integer? Or how did you handle this?

Possibly related to #494 .

cscorley:
@piskvorky more or less, I didn't change it here because I don't know how that would affect the Pyro stuff, so I left it as is for this PR. No idea how to make a test for that, not familiar with Pyro at all.

piskvorky:
Alright! Let's change it in a separate PR. It will affect Pyro performance (=will need more memory in the distributed computing), but that use case is so marginal I can't say I care much. Let's just mention in visibly in the CHANGELOG, for people who do use the distributed functionality.

Maybe we can even add this is an LDA parameter (default `as_numpy=False`), rather than hardwiring True/False there. That way, users can choose the value themselves.

@mattilyra are you up for this PR?

mattilyra:
@piskvorky Yes I can take it. Having a parameter `as_numpy=False/True` for LDA isn't particularly clear though and needs a lot of explaining. Besides wouldn't `as_numpy=True` break (or generate a lot of warnings) for Pyro as well under Py3?

PR #454, opened 2015-09-16
remove setup.cfg file

kmike:
We shouldn't build universal wheels for gensim because gensim provides C extensions. 
See http://pythonwheels.com/: 

> Warning: If your project has optional C extensions, it is recommended not to publish a universal wheel, because pip will prefer the wheel over a source installation.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,0 @@
-[wheel]
-universal = 1


cscorley:
:+1: this is probably why some users have had such a wacky time with word2vec.FAST_VERSION == -1.

kmike:
Exactly, that's what I was trying to fix.

piskvorky:
Good catch, I didn't realize it was still there!

Thanks @kmike , merging.

PR #455, opened 2015-09-19
error-log instead of assert shape-check in save_word2vec_format

gojomo:
Change assert that to ERROR-level logging; some custom optimizations (loading vocab from elsewhere) might legitimately cause mismatch (which would normally be a problem).

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -956,7 +956,8 @@ class Word2Vec(utils.SaveLoad):
                 for word, vocab in sorted(iteritems(self.vocab), key=lambda item: -item[1].count):
                     vout.write(utils.to_utf8("%s %s\n" % (word, vocab.count)))
         logger.info("storing %sx%s projection weights into %s" % (len(self.vocab), self.vector_size, fname))
-        assert (len(self.vocab), self.vector_size) == self.syn0.shape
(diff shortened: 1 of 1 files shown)


piskvorky:
Won't word look up fail if there's a mismatch between matrix and vocab? I think the assert is the proper way?

gojomo:
Hmm, I may have misdiagnosed what's happening; I thought it involved the `$TOTAL_DOCS_` word, but maybe instead word2vec.c is emitting a duplicate word? A mismatch would only be fatal if the token->index mapping was wrong (or out-of-bounds)?

piskvorky:
Yes. Can you check what's going on?

gojomo:
It may not be a true blocker, as I don't think there's a true need to dsword2vec-load then dsword2vec-save at that place in the TrainWord2Vec task: the model is already on-disk from word2vec.c. See the slight rearrangement in batch-branch: the lines are either moved to the gensim block, or unnecessary.

piskvorky:
Sure, a quick fix is fine.

I'd still like to know what's going on though, it's a bit worrying. I'm thinking it may be connected to unicode normalization (words different in utf8, same in unicode)?

piskvorky:
Fixed in acf1c199 (already released).

PR #456, opened 2015-09-19
LDA: initialisation of expElogbeta fixed

tmylk:
Fixing the initial value of expElogbeta to be in line with original Hoffmann implementation.


cscorley:
The only difference I see in this PR and the current implementation is that `eta` eventually is added into `sstats` in the call to `get_lambda` (and hence `get_Elogbeta`/`sync_state`). Would this be a concern other places that use `get_lambda`/`get_Elogbeta`?

tmylk:
All other calls to sync_state, get_lambda and get_Elogbeta are from the M-step so they are ok. 
It is just the first initialisation should be drawn just from gamma. Following this standard makes it much easier to compare gensim with other implementations of Hoffmann.

PR #458, opened 2015-09-22
Speed up accuracy evaluation

piskvorky:
Speed up accuracy evaluation, by taking into account only those words from `restrict_vocab` (e.g., top 30k most common words).

Will only work when the vocabulary has already been sorted.

TODO @gojomo construct the "top 30k-only matrix" dynamically, so we don't have to care about sorting.


gojomo:
Renamed parameter, added parameter doc, added `sort_vocab()` method and initialization parameter `sorted_vocab` (default on) which will sort the vocabulary early in `finalize_vocab()`, before `syn0` is allocated/initialized. Good to merge when tests pass (as they did locally).

piskvorky:
Thanks, merging!

PR #463, opened 2015-09-23
Fix summarization.keywords for py3k. Fixes #459

### LINKED ISSUE #459: dict.iteritems() usage breaks "keywords" function in python 3
Calling `gensim.summarization.keywords` in python3 results in `AttributeError: 'dict' object has no attribute 'iteritems'`.

Looks like there are offenders [here](https://github.com/piskvorky/gensim/blob/485d024e119d0ff818ea6651042748a3bff21081/gensim/summarization/keywords.py#L35) and [here](https://github.com/piskvorky/gensim/blob/485d024e119d0ff818ea6651042748a3bff21081/gensim/summarization/keywords.py#L114). I assume these should be fixed with six.iteritems?

d10genes:
With basic test

### DIFF
--- a/gensim/summarization/keywords.py
+++ b/gensim/summarization/keywords.py
@@ -8,9 +8,11 @@ from gensim.summarization.textcleaner import clean_text_by_word as _clean_text_b
 from gensim.summarization.textcleaner import tokenize_by_word as _tokenize_by_word
 from gensim.summarization.commons import build_graph as _build_graph
 from gensim.summarization.commons import remove_unreachable_nodes as _remove_unreachable_nodes
+from gensim.utils import to_unicode, to_utf8
 from itertools import combinations as _combinations
 from six.moves.queue import Queue as _Queue
 from six.moves import xrange
+from six import iteritems
 
 
 WINDOW_SIZE = 2
@@ -32,7 +34,7 @@ def _get_words_for_graph(tokens):
         raise ValueError("Can't use both include and exclude filters, should use only one")
 
     result = []
-    for word, unit in tokens.iteritems():
+    for word, unit in iteritems(tokens):
         if exclude_filters and unit.tag in exclude_filters:
             continue
         if (include_filters and unit.tag in include_filters) or not include_filters or not unit.tag:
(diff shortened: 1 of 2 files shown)


gensim/test/test_summarization.py
```diff
@@ -128,3 +128,14 @@ def test_corpus_summarization_ratio(self):
             expected_summary_length = int(len(corpus) * ratio)
 
             self.assertEqual(len(selected_docs), expected_summary_length)
+
+    def test_keywords_runs(self):
+        pre_path = os.path.join(os.path.dirname(__file__), 'test_data')
+
+        with utils.smart_open(os.path.join(pre_path, "mihalcea_tarau.txt"), mode="r") as f:
```
piskvorky: `"r"` mode is default, no need to mention it explicitly

gensim/test/test_summarization.py
```diff
@@ -128,3 +128,14 @@ def test_corpus_summarization_ratio(self):
...
+            text = f.read()
+
+        kwds = keywords(text)
```
piskvorky: Can you try the same thing, but also using `keywords(to_unicode(text))`?

That way we test that both utf8 input and unicode input work.

gensim/summarization/keywords.py
```diff
@@ -186,11 +188,12 @@ def _format_results(_keywords, combined_keywords, split, scores):
         return [(word, _get_average_score(word, _keywords)) for word in combined_keywords]
     if split:
         return combined_keywords
-    return "\n".join(combined_keywords)
+    return to_utf8("\n".join(combined_keywords))
```
piskvorky: We use utf8 only on output (serializing to files, external APIs etc). Inside gensim, we try to keep all text as unicode.

d10genes: Possibly not, I just put it there since it's the final step for `keywords` output.

gensim/test/test_summarization.py
d10genes: Sure, I also put some contrived `assertTrue` statements in there since it doesn't feel like a *real* test without asserts somewhere. If you think it's fine to simply call the `keywords` function though I can take them out.

tmylk:
Thanks. Looks good. FYI More complicated asserts are in https://github.com/piskvorky/gensim/pull/441

PR #466, opened 2015-09-24
{load|intersect}_word2vec_format: allow non-strict unicode error handling

gojomo:
A word2vec.c-format file might not have perfectly legal unicode encodings. A reader might want to load them anyway. This adds a parameter to load_word2vec_format & intersect_word2vec_format, default 'strict', that is passed to the `utils.to_unicode()` method as its `errors` parameter. By choosing one of the more tolerant alternatives – `ignore` or `replace` – bad unicode won't cause a fatal ValueError. Instead, the problematic string will just ignore or replace the offending char(s), allowing progress to continue.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -989,7 +989,7 @@ class Word2Vec(utils.SaveLoad):
                     fout.write(utils.to_utf8("%s %s\n" % (word, ' '.join("%f" % val for val in row))))
 
     @classmethod
-    def load_word2vec_format(cls, fname, fvocab=None, binary=False, norm_only=True, encoding='utf8'):
+    def load_word2vec_format(cls, fname, fvocab=None, binary=False, norm_only=True, encoding='utf8', unicode_errors='strict'):
         """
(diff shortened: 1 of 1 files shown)


piskvorky:
@tmylk I don't see this (important) change in the CHANGELOG. Did you miss it in the last release notes?

gojomo:
Perhaps even 'ignore' should be the default, *or* we could catch the error and before re-throwing include a message pointing to the option?

piskvorky:
Hmm, yes, a clearer error message may help. Plus a re-throw. Certainly not ignore.

ReemSuwaileh:
How can we use this parameter of load_word2vec_format that skips or replaces the chars with different encoding? I can't find this on documentation!

tmylk:
@ReemSuwaileh This question is better suited for the mailing list.

‘strict’, ‘replace’ (add U+FFFD, ‘REPLACEMENT CHARACTER’), or ‘ignore’  description in [Python documentation](https://docs.python.org/2/howto/unicode.html).

PR #468, opened 2015-09-27
Batch sentences inside word2vec/doc2vec. Fixes #450

### LINKED ISSUE #450: Batch sentences inside word2vec/doc2vec
...for an additional speed up (GIL acquired only once per batch, not once per sentence).

This is especially critical for very short documents, where the GIL overhead becomes deadly.

olavurmortensen:
This pull request will (when it's finished) fix [issue #450](https://github.com/piskvorky/gensim/issues/450). Training on a lot of short sentences can be done more efficiently by sending them to the Cython code in batches. This is more efficient as GIL is acquired once per batch, not once per sentence, thus it reduces the GIL overhead.

(More information will follow when there is more progress.)


piskvorky:
...continued discussion from #450.

TODO:
1. provide a high-level description for this PR (not *"No description provided"*)
2. resolve git conflicts, so the PR is merge-able
3. attach a clearer protocol of your experiments (what you tried, input data stats, logs, timings)

gensim/models/word2vec.py
```diff
@@ -725,31 +731,43 @@ def worker_loop():
...
+                        else:
+                            # words-based decay
+                            pushed_words += self._raw_word_count(job_batch)
+                            next_alpha = self.alpha - (self.alpha - self.min_alpha) * (pushed_words / total_words)
+                        next_alpha = max(next_alpha, self.min_alpha)
+                    # Reset the batch, and save the missed sentence for the next iteration.
+                    job_batch = [sent]
+                    batch_size = len(sent)
+                    job_no += 1
+                except StopIteration:
```
piskvorky: What is this try-except for? I don't see any iteration inside the block.

gensim/models/word2vec_inner.pyx
```diff
@@ -343,6 +343,99 @@ def train_sentence_sg(model, sentence, alpha, _work):
...
+            if i == MAX_SENTENCE_LEN:
+                break  # TODO: log warning, tally overflow?
+        sentence_len[sent_idx] = i
+        num_sentences += 1
+
+        # single randint() call avoids a big thread-sync slowdown
+        for i, item in enumerate(model.random.randint(0, window, sentence_len[sent_idx])):
+            reduced_windows[sent_idx][i] = item
+
+    # release GIL & train on the sentence
```
piskvorky: sentence**s**

gensim/models/word2vec.py
```diff
@@ -725,31 +731,43 @@ def worker_loop():
...
-                    next_alpha = max(next_alpha, self.min_alpha)
-            except StopIteration:
-                logger.info(
-                    "reached end of input; waiting to finish %i outstanding jobs",
-                    job_no - done_jobs + 1)
-                for _ in xrange(self.workers):
-                    job_queue.put((None, 0))  # give the workers heads up that they can finish -- no more work!
-                push_done = True
+        for sent in sentences:
+            if batch_size + len(sent) < MAX_SENTENCE_LEN:
```
piskvorky: misnomer; this is not the maximum length of a sentence, but rather the maximum number of words in a batch (as @gojomo pointed out)

gensim/models/word2vec_inner.pyx
```diff
@@ -343,6 +343,99 @@ def train_sentence_sg(model, sentence, alpha, _work):
...
+        cum_table = <np.uint32_t *>(np.PyArray_DATA(model.cum_table))
+        cum_table_len = len(model.cum_table)
+    if negative or sample:
+        next_random = (2**24) * model.random.randint(0, 2**24) + model.random.randint(0, 2**24)
+
+    # convert Python structures to primitive types, so we can release the GIL
+    work = <REAL_t *>np.PyArray_DATA(_work)
+
+    vlookup = model.vocab
+    for sent_idx, sentence in enumerate(sentences):
```
piskvorky: What happens when `sent_idx` is greater than `MAX_NUM_SENTENCES`?

...as I think happens currently, because it doesn't look like the number of sentences passed from `job_batch` is limited in any way?

piskvorky: During debugging, it's best to turn on bounds checking etc, near the top of this .pyx file.

During profiling, turn the checks back off, for best performance (but only after functionality has been debugged).

gensim/models/word2vec.py
```diff
@@ -725,31 +731,43 @@ def worker_loop():
...
+                            # examples-based decay
+                            pushed_examples += len(job_batch)
+                            next_alpha = self.alpha - (self.alpha - self.min_alpha) * (pushed_examples / total_examples)
+                        else:
+                            # words-based decay
+                            pushed_words += self._raw_word_count(job_batch)
+                            next_alpha = self.alpha - (self.alpha - self.min_alpha) * (pushed_words / total_words)
+                        next_alpha = max(next_alpha, self.min_alpha)
+                    # Reset the batch, and save the missed sentence for the next iteration.
+                    job_batch = [sent]
```
piskvorky: What happends when there are no more sentences at this point? Doesn't this last `job_batch` get lost?

olavurmortensen:
@gojomo in response to your comment on #450.  The code on the PR does cause the segmentation fault, yes. I've added a script `gensim/test/test_word2vec_batching.py` that reliably causes the segmentation fault (needs nltk installed with brown corpus downloaded).

gojomo:
Can you see @piskvorky's code comment above, from 11 days ago, about an issue that risks memory corruption, and an approach that helps to debug such issues (setting `# cython: boundscheck=True` while testing)?

olavurmortensen:
@piskvorky @tmylk I've tested whether the batched vectors are the same as the original, and the conclusion is that they are not (difference is up to the order of 1e-3). However, this actually makes sense because the learning rate is updated per batch instead of per sentence.

piskvorky:
Hmmm true. I haven't thought of that. I'll try to freeze alpha and then check equality, as I refactor the job loop.

olavurmortensen:
I've spotted some errors I need to fix. Have you started refactoring the code @piskvorky?

piskvorky:
Yes, I've refactored the code a bit. See https://github.com/olavurmortensen/gensim/pull/1 .

gensim/models/word2vec_inner.pyx
```diff
@@ -343,6 +343,107 @@ def train_sentence_sg(model, sentence, alpha, _work):
...
+    cdef REAL_t *word_locks = <REAL_t *>(np.PyArray_DATA(model.syn0_lockf))
+    cdef REAL_t *work
+    cdef REAL_t _alpha = alpha
+    cdef int size = model.layer1_size
+
+    cdef int codelens[MAX_SENTENCE_LEN]
+    cdef np.uint32_t indexes[MAX_SENTENCE_LEN]
+    cdef np.uint32_t reduced_windows[MAX_SENTENCE_LEN]
+    cdef int sentence_len[MAX_NUM_SENTENCES]
+    cdef int sentence_indeces_c[MAX_NUM_SENTENCES]
```
gojomo: I don't believe 'indeces' is a typical pluralization of 'index' anywhere. FWIW, the use of 'indexes' vs 'indices' was mixed in this code (especially between the pure-python and cython files) when I first started optimizing/extending it. There's a reasonable case to be made for either spelling, but I prefer 'indexes' so that greps for 'index' also find related lists/arrays.

olavurmortensen:
# Report

This report summarizes the work I'ved done in this pull request.

## Motivation

We want to speed up the training of the Word2Vec model by sending batches of sentences (instead of single sentences) to the Cython code. When we send a sentence at a time, we have to switch from GIL to NOGIL everytime, and when we are training on short sentences, the overhead of this switch slow us down a lot.

## Strategy

Basically, the idea is that we send a list of sentences to the Cython code, and then we loop over the sentences inside the NOGIL zone. This way, we switch from GIL to NOGIL once per batch, instead of once per sentence.

The batches must have less than `MAX_SENTENCE_LEN` words in total, and less than `MAX_NUM_SENTENCES` sentences (100 000 and 10 000, respectively, at the moment). The bigger we can make the batches, the less time we spend on the GIL context switching overhead. 

Changes to the code were made in:

* gensim/models/word2vec_inner.pyx
    * Added `train_batch_sg()`
        * Does what `train_sentence_sg()` does, but for a batch of sentences.
        * Takes a list of sentences as input.
        * There are some arrays that need to be prepared (`codelens`, `indexes` and so on) before they are sent to the `fast_sentence_sg_hs()` and `fast_sentence_sg_neg()` functions. I need one of these arrays for each of the sentences, so I just make some long arrays and keep track of the indexes where the sentences start and end.
* gensim/models/word2vec.py
    * `train()`
        * Prepare job batches and keep the `MAX_SENTENCE_LEN` and `MAX_NUM_SENTENCES` limits. 
    * `_do_train_job()`
        * Send batches of sentences to `train_batch_sg()`.


**Note:** we create Cython arrays of size `MAX_SENTENCE_LEN` and `MAX_NUM_SENTENCES`; if we make these arrays very large we cause a *segmentation fault* when we run the code. This seems to happen when `MAX_SENTENCE_LEN = 10000000` (that is, 10^7), but it is possible it depends on the compiler and/or the computer hardware.

## Testing

To test the code I trained on sentences of the Brown corpus from NLTK, and measure the time it takes to train the model. The test script can be found [here](https://github.com/olavurmortensen/gensim/blob/batch_sentences/gensim/test/test_word2vec_batching.py). These sentences have about 20 to 40 words each, and they are typically batched into about 5000 sentences with just under 100 000 words total.

I compare the performance to the develop branch of Gensim. One CPU core was used in these tests.

#sentences | Time, batched | Time, non-batched
-----------|---------------|--------------------
10 000     | 6.3 s         | 6.8 s
50 000     | 30.5 s        | 35.6 s

## Profiling

To profile the Cython code, I use `line_profiler` ([link](https://github.com/rkern/line_profiler)). I want to profile `train_batch_sg()`, so I make the following modifications.

* setup.py
    * In setup() add `define_macros=[('CYTHON_TRACE', '1'), ('CYTHON_TRACE_NOGIL','1')])` to the `word2vec` extension in `ext_modules`.
* gensim/models/word2vec_inner.pyx:
    * Add the following lines in the start:
```
        # cython: profile=True
        # cython: linetrace=True
        # cython: binding=True
        # distutils: define_macros=CYTHON_TRACE=1
        # distutils: define_macros=CYTHON_TRACE_NOGIL=1
```
        
* gensim/models/word2vec.py:
    * Add these lines right before call to `train_batch_sg()`:
```
        import line_profiler
        profile = line_profiler.LineProfiler(train_batch_sg)
        temp_tally = profile.runcall(train_batch_sg, self, sentences, alpha, work)
        print 'temp_tally = %d' % temp_tally
        profile.print_stats()
        import pdb
        pdb.set_trace()
```

Now compile `word2vec_inner.pyx`, and run `python setup.py install`. When `word2vec.py` is run, the profile will be printed to output, and we will go into debug at `pdb.set_trace()`.

As the output from the profiler is a bit lengthy, it is shown at the bottom of this report (of both the original and the modified Cython code). 

Let us look at the profile of the NOGIL section of the code (line 332 of original code and line 427 of modified code). Remember, the difference between the two versions is that in one we loop over sentences outside the NOGIL, and in one we loop over sentences inside the NOGIL. Note: we must be careful in comparing the numbers directly, as the amount of data processed in the two versions is most likely not the same.

We can see in both profiles that the time spent on the GIL switch pales in comparison to the rest of the computation taking place. If we, for example, look at the modified code, 3\*10^(-6) s is spent on the GIL switch, while 7473863\*10^(-6) s is spent on `fast_sentence_sg_hs()` alone.


## Future work

At the moment, the learning rate (`alpha`) is only being updated once per batch, which in the tests discussed above means the learning rate is only updated once per 5000 sentences (roughly).

When the vectors are initilized with the same seed (and `workers=1`), if we freeze the learning rate (set `alpha` constant), the resulting vectors *should* be the same. This is not the case in this implementation, the results differ. If this pull request is to be merged, this needs to be solved.

At the moment, batching is only implemented in Word2Vec, and only for skip-gram. It needs to be implemented for CBOW, and for Doc2Vec and its methods: DM and DBOW.

## Conclusion

The tests show there is not a lot of speed-up to gain from this optimization. I believe this is because the GIL switch is a quite small portion of the computation taking place in the code.

## Profile of original code

Output from profiling `train_sentence_sg()`.

```
Timer unit: 1e-06 s

Total time: 0.11104 s
File: /path/to/file/gensim/gensim/models/word2vec_inner.pyx
Function: train_sentence_sg at line 263

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   263                                           def train_sentence_sg(model, sentence, alpha, _work):
   264         1            3      3.0      0.0      cdef int hs = model.hs
   265         1            1      1.0      0.0      cdef int negative = model.negative
   266         1            1      1.0      0.0      cdef int sample = (model.sample != 0)
   267                                           
   268         1            2      2.0      0.0      cdef REAL_t *syn0 = <REAL_t *>(np.PyArray_DATA(model.syn0))
   269         1            2      2.0      0.0      cdef REAL_t *word_locks = <REAL_t *>(np.PyArray_DATA(model.syn0_lockf))
   270                                               cdef REAL_t *work
   271         1            1      1.0      0.0      cdef REAL_t _alpha = alpha
   272         1            1      1.0      0.0      cdef int size = model.layer1_size
   273                                           
   274                                               cdef int codelens[MAX_SENTENCE_LEN]
   275                                               cdef np.uint32_t indexes[MAX_SENTENCE_LEN]
   276                                               cdef np.uint32_t reduced_windows[MAX_SENTENCE_LEN]
   277                                               cdef int sentence_len
   278         1            1      1.0      0.0      cdef int window = model.window
   279                                           
   280                                               cdef int i, j, k
   281         1            1      1.0      0.0      cdef long result = 0
   282                                           
   283                                               # For hierarchical softmax
   284                                               cdef REAL_t *syn1
   285                                               cdef np.uint32_t *points[MAX_SENTENCE_LEN]
   286                                               cdef np.uint8_t *codes[MAX_SENTENCE_LEN]
   287                                           
   288                                               # For negative sampling
   289                                               cdef REAL_t *syn1neg
   290                                               cdef np.uint32_t *cum_table
   291                                               cdef unsigned long long cum_table_len
   292                                               # for sampling (negative and frequent-word downsampling)
   293                                               cdef unsigned long long next_random
   294                                           
   295         1            1      1.0      0.0      if hs:
   296         1            1      1.0      0.0          syn1 = <REAL_t *>(np.PyArray_DATA(model.syn1))
   297                                           
   298         1            1      1.0      0.0      if negative:
   299                                                   syn1neg = <REAL_t *>(np.PyArray_DATA(model.syn1neg))
   300                                                   cum_table = <np.uint32_t *>(np.PyArray_DATA(model.cum_table))
   301                                                   cum_table_len = len(model.cum_table)
   302         1            1      1.0      0.0      if negative or sample:
   303                                                   next_random = (2**24) * model.random.randint(0, 2**24) + model.random.randint(0, 2**24)
   304                                           
   305                                               # convert Python structures to primitive types, so we can release the GIL
   306         1            0      0.0      0.0      work = <REAL_t *>np.PyArray_DATA(_work)
   307                                           
   308         1            1      1.0      0.0      vlookup = model.vocab
   309         1            0      0.0      0.0      i = 0
   310        21           17      0.8      0.0      for token in sentence:
   311        22           28      1.3      0.0          word = vlookup[token] if token in vlookup else None
   312        22           15      0.7      0.0          if word is None:
   313         2            2      1.0      0.0              continue  # leaving i unchanged/shortening sentence
   314        20           13      0.7      0.0          if sample and word.sample_int < random_int32(&next_random):
   315                                                       continue
   316        20           17      0.8      0.0          indexes[i] = word.index
   317        20           12      0.6      0.0          if hs:
   318        20           23      1.1      0.0              codelens[i] = <int>len(word.code)
   319        20           17      0.8      0.0              codes[i] = <np.uint8_t *>np.PyArray_DATA(word.code)
   320        20           16      0.8      0.0              points[i] = <np.uint32_t *>np.PyArray_DATA(word.point)
   321        20           13      0.7      0.0          result += 1
   322        20           18      0.9      0.0          i += 1
   323        20           16      0.8      0.0          if i == MAX_SENTENCE_LEN:
   324                                                       break  # TODO: log warning, tally overflow?
   325         1            0      0.0      0.0      sentence_len = i
   326                                           
   327                                               # single randint() call avoids a big thread-sync slowdown
   328        21           63      3.0      0.1      for i, item in enumerate(model.random.randint(0, window, sentence_len)):
   329        20           14      0.7      0.0          reduced_windows[i] = item
   330                                           
   331                                               # release GIL & train on the sentence
   332         2            6      3.0      0.0      with nogil:
   333         1            1      1.0      0.0          for i in range(sentence_len):
   334        20          626     31.3      0.6              j = i - window + reduced_windows[i]
   335        20           30      1.5      0.0              if j < 0:
   336         3            3      1.0      0.0                  j = 0
   337        20           24      1.2      0.0              k = i + window + 1 - reduced_windows[i]
   338        20          165      8.2      0.1              if k > sentence_len:
   339         2            1      0.5      0.0                  k = sentence_len
   340        20          967     48.4      0.9              for j in range(j, k):
   341       114         1268     11.1      1.1                  if j == i:
   342        20           25      1.2      0.0                      continue
   343        94         2778     29.6      2.5                  if hs:
   344        94       104594   1112.7     94.2                      fast_sentence_sg_hs(points[i], codes[i], codelens[i], syn0, syn1, size, indexes[j], _alpha, work, word_locks)
   345        94          248      2.6      0.2                  if negative:
   346                                                               next_random = fast_sentence_sg_neg(negative, cum_table, cum_table_len, syn0, syn1neg, size, indexes[i], indexes[j], _alpha
, work, next_random, word_locks)
   347                                           
   348         1            2      2.0      0.0      return result

```

## Profile of modified code

Output from profiling `train_batch_sg()`.

```
Timer unit: 1e-06 s

Total time: 9.66989 s
File: /path/to/file/gensim/gensim/models/word2vec_inner.pyx
Function: train_batch_sg at line 351

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   351                                           def train_batch_sg(model, sentences, alpha, _work):
   352         1            2      2.0      0.0      cdef int hs = model.hs
   353         1            2      2.0      0.0      cdef int negative = model.negative
   354         1            1      1.0      0.0      cdef int sample = (model.sample != 0)
   355                                           
   356         1            1      1.0      0.0      cdef REAL_t *syn0 = <REAL_t *>(np.PyArray_DATA(model.syn0))
   357         1            1      1.0      0.0      cdef REAL_t *word_locks = <REAL_t *>(np.PyArray_DATA(model.syn0_lockf))
   358                                               cdef REAL_t *work
   359         1            1      1.0      0.0      cdef REAL_t _alpha = alpha
   360         1            1      1.0      0.0      cdef int size = model.layer1_size
   361                                           
   362                                               cdef int codelens[MAX_SENTENCE_LEN]
   363                                               cdef np.uint32_t indexes[MAX_SENTENCE_LEN]
   364                                               cdef np.uint32_t reduced_windows[MAX_SENTENCE_LEN]
   365                                               cdef int sentence_len[MAX_NUM_SENTENCES]
   366         1            1      1.0      0.0      cdef int window = model.window
   367                                           
   368                                               cdef int i, j, k, m
   369         1            0      0.0      0.0      cdef long result = 0
   370         1            1      1.0      0.0      cdef int num_sentences = 0
   371         1            1      1.0      0.0      cdef int sent_idx = 0
   372         1            0      0.0      0.0      cdef int idx = 0
   373                                           
   374                                               # For hierarchical softmax
   375                                               cdef REAL_t *syn1
   376                                               cdef np.uint32_t *points[MAX_SENTENCE_LEN]
   377                                               cdef np.uint8_t *codes[MAX_SENTENCE_LEN]
   378                                           
   379                                               # For negative sampling
   380                                               cdef REAL_t *syn1neg
   381                                               cdef np.uint32_t *cum_table
   382                                               cdef unsigned long long cum_table_len
   383                                               # for sampling (negative and frequent-word downsampling)
   384                                               cdef unsigned long long next_random
   385                                           
   386         1            0      0.0      0.0      if hs:
   387         1            1      1.0      0.0          syn1 = <REAL_t *>(np.PyArray_DATA(model.syn1))
   388                                           
   389         1            1      1.0      0.0      if negative:
   390                                                   syn1neg = <REAL_t *>(np.PyArray_DATA(model.syn1neg))
   391                                                   cum_table = <np.uint32_t *>(np.PyArray_DATA(model.cum_table))
   392                                                   cum_table_len = len(model.cum_table)
   393         1            1      1.0      0.0      if negative or sample:
   394                                                   next_random = (2**24) * model.random.randint(0, 2**24) + model.random.randint(0, 2**24)
   395                                           
   396                                               # convert Python structures to primitive types, so we can release the GIL
   397         1            1      1.0      0.0      work = <REAL_t *>np.PyArray_DATA(_work)
   398                                           
   399         1            1      1.0      0.0      vlookup = model.vocab
   400      5338         3240      0.6      0.0      for sent_idx, sent in enumerate(sentences):
   401      5337         3345      0.6      0.0          i = 0
   402     93538        52269      0.6      0.5          for token in sent:
   403     99992        70275      0.7      0.7              word = vlookup[token] if token in vlookup else None
   404     99992        50088      0.5      0.5              if word is None:
   405     11791         6547      0.6      0.1                  continue  # leaving i unchanged/shortening sentence
   406     88201        44746      0.5      0.5              if sample and word.sample_int < random_int32(&next_random):
   407                                                           continue
   408     88201        50990      0.6      0.5              indexes[idx + i] = word.index
   409     88201        44749      0.5      0.5              if hs:
   410     88201        52049      0.6      0.5                  codelens[idx + i] = <int>len(word.code)
   411     88201        47031      0.5      0.5                  codes[idx + i] = <np.uint8_t *>np.PyArray_DATA(word.code)
   412     88201        50448      0.6      0.5                  points[idx + i] = <np.uint32_t *>np.PyArray_DATA(word.point)
   413     88201        44224      0.5      0.5              result += 1
   414     88201        44149      0.5      0.5              i += 1
   415     88201        43912      0.5      0.5              if i == MAX_SENTENCE_LEN:
   416                                                           break  # TODO: log warning, tally overflow?
   417                                           
   418      5337         2699      0.5      0.0          sentence_len[sent_idx] = i
   419                                                   # single randint() call avoids a big thread-sync slowdown
   420     93538        91428      1.0      0.9          for i, item in enumerate(model.random.randint(0, window, sentence_len[sent_idx])):
   421     88201        68955      0.8      0.7              reduced_windows[idx + i] = item
   422                                           
   423      5337         3033      0.6      0.0          idx += len(sent)
   424      5337         3531      0.7      0.0          num_sentences += 1
   425                                           
   426                                               # release GIL & train on the sentences
   427         2            3      1.5      0.0      with nogil:
   428         1            2      2.0      0.0          for sent_idx in range(num_sentences):
   429      5337         3372      0.6      0.0              m = 0 if sent_idx == 0 else sentence_len[sent_idx - 1]
   430      5337         3457      0.6      0.0              for i in range(sentence_len[sent_idx]):
   431     88201        61926      0.7      0.6                  j = i - window + reduced_windows[m + i]
   432     88201        55674      0.6      0.6                  if j < 0:
   433     17316        10970      0.6      0.1                      j = 0
   434     88201        60958      0.7      0.6                  k = i + window + 1 - reduced_windows[m + i]
   435     88201        57073      0.6      0.6                  if k > sentence_len[sent_idx]:
   436     16852        10375      0.6      0.1                      k = sentence_len[sent_idx]
   437     88201        60433      0.7      0.6                  for j in range(j, k):
   438    590350       393176      0.7      4.1                      if j == i:
   439     88201        56467      0.6      0.6                          continue
   440    502149       326425      0.7      3.4                      if hs:
   441    502149      7473863     14.9     77.3                          fast_sentence_sg_hs(points[m + i], codes[m + i], codelens[m + i], syn0, syn1, size, indexes[m + j], _alpha, work, word
_locks)
   442    502149       317992      0.6      3.3                      if negative:
   443                                                                   next_random = fast_sentence_sg_neg(negative, cum_table, cum_table_len, syn0, syn1neg, size, indexes[m + i], indexes[m 
+ j], _alpha, work, next_random, word_locks)
   444                                           
   445         1            3      3.0      0.0      return result
```

tmylk:
A question about Testing section: how many workers were used there? It's
not clear from the script if it's 1 or 3.

olavurmortensen:
1 worker was used in the tests.

gojomo:
The potential improvement here would most show up in multi-worker testing. 

In a single thread, the batching is just extra work – both rolling things together in GIL-code, and unrolling them in NOGIL-code. And without other threads contending for the GIL, releasing/re-acquiring it is an easier operation. That the single-worker case is faster at all is even a little surprising, and may be due to the new MAX_SENTENCE_LEN arrays having data densely packed (greater CPU cache efficiency?) compared to the old over-allocated C arrays where 9,980 indexes were empty for the usual 20-word sentences.

With multiple threads, every NOGIL-entry means picking a waiting thread as the new GIL holder, and every exit from the NOGIL section requires waiting for the GIL to become available, perhaps in line with other waiting threads. Triggering those contended handoffs 1,000x less often (for 10-word sentences packed into a 10,000-long C arrays), with multiple threads, might demonstrate a bigger speedup.

olavurmortensen:
I should have mentioned that multithreading does not seem to work with this
code (only seems to use one core even if I set `workers = 2`). I don't know
what causes this.

gojomo:
The best test would be to use as many workers as available cores, which is typical for users of this code. You might have a better chance of seeing multiple cores engaged, during the training phase, with such a setting. (If you don't, there is some more foundational problem which should be solved first, as multi-threaded operation is necessary for best/representative performance.)

piskvorky:
Travis reports fails; and what do you mean multithreading does not work @olavurmortensen ?

If it does not work then this is useless. It just means there are still bugs in the code. Profiling buggy code is a futile exercise, the first step is always to make the code bugless.

piskvorky:
We're not finished with the batching (it's not working correctly in single core nor multi core) -- please leave open.

piskvorky:
Continued in #535 .

PR #473, opened 2015-09-30
PRs 441 463

tmylk:
Combine PRs #441 and #463 to see if 463 fixed 441's unicode test failures


tmylk:
KeywordsTest fails in the same way in Python 2 and 3. This is an improvement over #441

PR #475, opened 2015-10-07
Update test_keywords.py

tmylk:
A part of #441 . Updated ratio parameters specifically for this example.
Previous cutoff values were sensitive to numerical issues i.e. in the plateaus of this graph.
scores


tmylk:
Already merged in #473

Issue #477, opened 2015-10-09
'corpus2dense' method doesn't work with small 'num_terms' for HDP


nikicc:
When using HDP model in an iteratively fashion (using update) and trying to get numpy matrix with small number of terms (`num_terms`) the call `gensim.matutils.corpus2dense(corpus, num_terms=3).T` raises the following error:
`     IndexError: index 3 is out of bounds for axis 1 with size 3`

The minimal example that causes this error is pasted below. I believe the problem is in the file `matutils.py` in line `209` (I am using `gensim (0.12.2)`). For small number of terms the `list(doc)` contains indices larger than the length of `result` and consequently it crashes.

*********************************************************************************
The code that reproduce this error is the following:
```    
import gensim
from gensim import corpora, models

documents = [
    "Human machine interface for lab abc computer applications.",
    "A survey of user opinion of computer system response time.",
    "The EPS user interface management system",
    "System and human system engineering testing of EPS",
    "Relation of user perceived response time to error measurement",
    "The generation of random binary unordered trees",
    "The intersection graph of paths in trees",
    "Graph minors IV Widths of trees and well quasi ordering",
    "Graph minors A survey",
]

text = []
for sent in documents:
    text.append(sent.split())

dictionary = corpora.Dictionary(text)
corpus = [dictionary.doc2bow(t) for t in text]

hdp = models.HdpModel(corpus=[[]], id2word=dictionary)
hdp.update(corpus)
corpus = hdp[corpus]

matrix = gensim.matutils.corpus2dense(corpus, num_terms=3).T
print(matrix)
```

piskvorky:
Why do you set `num_terms=3`?

If I recall correctly, the dictionary on that example toy corpus had more than 10 terms. (since you don't even lower case the texts, probably more than 20).

nikicc:
Maybe my understanding of HDP is wrong. Is not the `num_terms` like the number of topics that you want to use to represent your data (as the number of topics for LDA)?

As I understand it, after using the HDP model on the `corpus`, it is the matrix of documents and theirs associations with different topics. Is that right?

What I then want is to do is to get out the matrix of documents in their "topical representation".

piskvorky:
No, the `num_terms` parameter in `corpus2dense(corpus, num_terms)` just says how many features there are in your corpus. It has nothing to do with HDP (the same `corpus2dense` call can be used on tfidf corpora, LSI corpora etc).

[...]

PR #478, opened 2015-10-09
sparse2full: Issue #477 fixup

nikicc:
Added check in `matutils.py` method `sparse2full` to only allow indices up to `length`. Fixing issue #477.


piskvorky:
We definitely don't want to silence any errors.

Passing in documents that have out-of-bound indexes is an error, and should throw an explicit exception (like it does now).

PR #479, opened 2015-10-10
LDA eta parameter auto-learning

cscorley:
Adds auto-learning for the eta parameter on the LdaModel, a feature mentioned in Hoffman's Online LDA paper. 

This PR is for a commit I cherry-picked from a fork by @joshua2ua. I've added some tests to ensure it works (=values actually change).

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -35,6 +35,7 @@ The algorithm:
 
 import logging
 import numpy  # for arrays, array broadcasting etc.
+import numbers
 
 from gensim import interfaces, utils, matutils
 from itertools import chain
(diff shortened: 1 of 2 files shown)


tmylk:
Cool, tests do give something asymmetric.
To avoid code duplication, could update_alpha and update_eta call the same dimension-agnostic update_dir_prior function?

cscorley:
@tmylk Not a bad idea, I've updated the PR with that.

cscorley:
@tmylk While were here, would it be a bad idea to use the same setup procedures for eta that we do in alpha, i.e., enforce a non-scalar value during init?

tmylk:
Yes, that would make code much easier.

cscorley:
@tmylk, sorry for the delay! Got around to fixing this up. Added a bunch of tests for the init behaviors of both alpha and eta.

tmylk:
Looks pretty neat.

On a related note. Do you think that this should go on the wishlist: "support update_alpha in MultiCore mode"?
Requires changes to LdaModel to share alpha's numpy array between E-step worker threads. Multiprocessing does locks and shared memory.

cscorley:
I would defer that to @ziky90.

tmylk:
Regression test passed. (i.e. the tests that apply to both pre and post-PR functionality pass in both versions.)

Issue #470, opened 2015-09-28
ValueError: Maximum allowed dimension exceeded


iqbaleric:
Hi,

I have installed numpy and scipy to make gensim library running in my PC. I am getting below error in both 64-bit and 32-bit version of Windows 7. Any quick help is highly appreciated.

```
======================================================================
ERROR: testScoring (__main__.TestWord2VecModel)
Test word2vec scoring.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\gensim\src\gensim\test\test_word2vec.py", line 193, in testScoring
    scores = model.score(sentences)
  File "C:\Users\gensim\src\gensim\models\word2vec.py", line 880, in score
    sentence_scores = matutils.zeros_aligned(total_sentences, dtype=REAL)
  File "C:\Users\gensim\src\gensim\matutils.py", line 147, in zeros_aligned
    buffer = numpy.zeros(nbytes + align, dtype=numpy.uint8)  # problematic on win64 ("maximum allowed dimension exceeded")
ValueError: Maximum allowed dimension exceeded

----------------------------------------------------------------------
Ran 28 tests in 1413.341s

FAILED (errors=1)
```

Thank you

tmylk:
@gojomo Do you have ideas about fixing? This is now  blocking for Windows builds and wheels #492 

Struggling to understand how can there be a memory error on Windows with only 9 sentences in the corpus.

gojomo:
I know little of Windows implementation limits other than the errors people have reported when hitting them, but from a look at the code:

For some reason, `score()` is using a default example count `total_sentences=int(1e9)`?! The comment does warn a value should be provided; maybe the 1B value is left over from some intentional extreme testing? (cc @mataddy)

At least in this case, since `sentence` is a plain list, the call from `testScoring()` could be `scores = model.score(sentences, len(sentences))`.

mataddy:
oops, my bad!  I'll fix this right now.

mataddy:
@tmylk and @gojomo this should be fixed. I dropped the total_sentences default and made its role more clear in the documentation.   

I already have an open pull request at #500 (adding a notebook demo for scoring) so this will be included in that now.

mataddy:
I'm also adding more safety checks to stop related issues; will ping here when I'm done.

mataddy:
OK this is all done.  @tmylk all you need should be in #500

tmylk:
[...]

PR #492, opened 2015-10-19
Windows build and wheel

tmylk:
TODO:

-  [ ] upload wheel to a fileshare
-  [ ] add steps to release wiki to download wheel from the share and upload to PyPi using wheelhouse-uploader and twine

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,85 @@
+# AppVeyor.com is a Continuous Integration service to build and run tests under
+# Windows
+# https://ci.appveyor.com/project/tmylk/gensim
+
+environment:
+  global:
+    # SDK v7.0 MSVC Express 2008's SetEnv.cmd script will fail if the
+    # /E:ON and /V:ON options are not enabled in the batch script intepreter
+    # See: http://stackoverflow.com/a/13751649/163740
+    CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\continuous_integration\\appveyor\\run_with_env.cmd"
+    
+  matrix:
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.8"
+      PYTHON_ARCH: "32"
+
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.8"
+      PYTHON_ARCH: "64"
+
+    - PYTHON: "C:\\Python35"
+      PYTHON_VERSION: "3.5.0"
+      PYTHON_ARCH: "32"
+
+    - PYTHON: "C:\\Python35-x64"
+      PYTHON_VERSION: "3.5.0"
+      PYTHON_ARCH: "64"
+
+
+
+install:
+  # Install Python (from the official .msi of http://python.org) and pip when
+  # not already installed.
+  - "powershell ./continuous_integration/appveyor/install.ps1"
+  - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+
+  # Check that we have the expected version and architecture for Python
+  - "python --version"
+  - "python -c \"import struct; print(struct.calcsize('P') * 8)\""
+
+  # Install the build and runtime dependencies of the project.
+    # Install the build and runtime dependencies of the project.
+  - "%CMD_IN_ENV% pip install --timeout=60 --trusted-host 28daf2247a33ed269873-7b1aad3fab3cc330e1fd9d109892382a.r6.cf2.rackcdn.com -r continuous_integration/appveyor/requirements.txt"
+  - "%CMD_IN_ENV% python setup.py bdist_wheel bdist_wininst "
+  - ps: "ls dist"
+
+  # Install the genreated wheel package to test it
+  - "pip install --pre --no-index --find-links dist/ gensim"
+
+# Not a .NET project, we build scikit-learn in the install step instead
+build: false
+
+test_script:
+  # Change to a non-source folder to make sure we run the tests on the
+  # installed library.
+  - "mkdir empty_folder"
+  - "cd empty_folder"
+
+  - "python -c \"import nose; nose.main()\" -s -v gensim"
+  # Move back to the project folder
+  - "cd .."
+
+artifacts:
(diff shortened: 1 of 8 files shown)


tmylk:
Blocked by #470 

Tests fail with "ValueError: Maximum allowed dimension exceeded"
https://ci.appveyor.com/project/tmylk/gensim-ahs80

tmylk:
Python 2 builds.

Python 3 Fails with "UnicodeEncodeError: 'mbcs' codec can't encode characters in position 0--1: invalid character".

Can't find a file with non-Ascii name though. Regex search returns nothing:
```
find . -print0 | perl -n0e 'chomp; print $_, "\n" if /[[:^ascii:][:cntrl:]]/'
```

gojomo:
The `cfgdata` being encoded by the offending line comes from `get_initdata()`, which appears to include not just a filename but other package metadata, including author name: https://github.com/python/cpython/blob/7fae13e4889ca44bbf22b61af277aca6d5dc13b9/Lib/distutils/command/bdist_wininst.py#L209

tmylk:
@gojomo Thanks a lot, that was the problem. 

Next step is Python3.5 on 32-bit windows
```
=====================================================================
ERROR: Test DBOW doc2vec training.
  File "C:\Python35-x64\lib\site-packages\gensim\test\test_doc2vec.py", line 160, in test_dbow_hs
    model = doc2vec.Doc2Vec(list_corpus, dm=0, hs=1, negative=0, min_count=2, iter=20)
  File "C:\Python35-x64\lib\site-packages\gensim\models\doc2vec.py", line 584, in __init__
    self.build_vocab(documents, trim_rule=trim_rule)
  File "C:\Python35-x64\lib\site-packages\gensim\models\word2vec.py", line 501, in build_vocab
    self.finalize_vocab()  # build tables & arrays
  File "C:\Python35-x64\lib\site-packages\gensim\models\word2vec.py", line 631, in finalize_vocab
    self.reset_weights()
  File "C:\Python35-x64\lib\site-packages\gensim\models\doc2vec.py", line 604, in reset_weights
    super(Doc2Vec, self).reset_weights()
  File "C:\Python35-x64\lib\site-packages\gensim\models\word2vec.py", line 953, in reset_weights
    self.syn0[i] = self.seeded_vector(self.index2word[i] + str(self.seed))
  File "C:\Python35-x64\lib\site-packages\gensim\models\word2vec.py", line 965, in seeded_vector
    once = random.RandomState(uint32(self.hashfxn(seed_string)))
OverflowError: Python int too large to convert to C long
```

gojomo:
I've seen that error reported before by Windows users, and they found a workaround. (Maybe in old github issues or forum threads? I even think the `unint32()` call may be an attempted workaround.)

But also, that error makes me suspicious as to whether that build is really 64-bit in all important ways. Could some 32-bit libraries/steps have snuck in, and the 'x64' in the name only reflects intent and not actuality?

To really understand what's going wrong, I might also try unpacking that line into 3...

    h = self.hashfxn(seed_string) 
    ui = uint32(h)
    once = random.RandomState(ui)

...to know exactly which one is triggering the error, and (with an extra print or catch) exactly what Python int value is breaking things.

tmylk:
I guess that the built-in hash function returns more than uint32 number on a 64-bit platform.
We need a uint32 to pass into random.RandomState.

Is there anything wrong with this hack from the mailing list?
```
hashfxn = lambda value:hash(value) & 0xffffffff
```
ERROR: test_zero_workers_mode (gensim.test.test_word2vec.TestWord2VecModel)
  File "C:\Python35-x64\lib\site-packages\gensim\test\test_word2vec.py", line 124, in test_zero_workers_mode
  File "C:\Python35-x64\lib\site-packages\gensim\models\word2vec.py", line 435, in __init__
  File "C:\Python35-x64\lib\site-packages\gensim\models\word2vec.py", line 501, in build_vocab
    self.finalize_vocab()  # build tables & arrays
  File "C:\Python35-x64\lib\site-packages\gensim\models\word2vec.py", line 631, in finalize_vocab
    self.reset_weights()
  File "C:\Python35-x64\lib\site-packages\gensim\models\word2vec.py", line 953, in reset_weights
    self.syn0[i] = self.seeded_vector(self.index2word[i] + str(self.seed))
  File "C:\Python35-x64\lib\site-packages\gensim\models\word2vec.py", line 966, in seeded_vector
    ui = uint32(h)
OverflowError: Python int too large to convert to C long
-------------------- >> begin captured logging << --------------------

gojomo:
Yes, the built-in hash function will give larger-than-32-bit numbers by design, and `numpy.uint32` is already supposed to be doing the casting/clipping down to the range `numpy.random.RandomState` needs. 

So, the core problem is `numpy.uint32()` isn't working on this Windows config. On other 64-bit platforms, it works. For example, try `numpy.uint32(sys.maxsize+1)`.

We can do the range-clipping ourselves with a bit-mask like in that workaround. However, since allowing a user-defined `hashfxn` is intended, simply putting the fix there is fine for an individual user's workaround but non-ideal for the general case. I'd instead change the line in `seeded_vector()` from:

        once = random.RandomState(uint32(self.hashfxn(seed_string)))

...to...

        once = random.RandomState(self.hashfxn(seed_string) & 0xffffffff)

That eliminates the dependency on `uint32()` (there) entirely, and will keep working if the user has some reason to set a custom `hashfxn` that gives full-range hashes. 

(That `uint32()` can't be relied upon to do the conversion it's supposed to might still be a time-bomb for anyone using numpy on such a Windows setup, but this one particular function should be OK...)

piskvorky:
@gojomo if that is the case, it may be worth [reporting](https://github.com/numpy/numpy) to the NumPy devs as well.

gojomo:
I'm not the best reporter to Numpy, since I can't directly reproduce... but I gave it a try at https://github.com/numpy/numpy/issues/6637. (Based on discussion I've seen elsewhere that 64-bit Windows isn't officially supported by Numpy, unsure how much attention it'll get.)

tmylk:
Windows build is now green.
Thanks to @gojomo and @piskvorky for making it possible.

Windows wheels coming soon!

piskvorky:
Great :)

PR #495, opened 2015-10-21
fixes #491: string doctags at end of index-range

### LINKED ISSUE #491: Doc2Vec: mixing plain int and string tags will give unexpected results
For typical Doc2Vec use with a large corpus and only a single (ID) tag per document, the best approach is to use contiguous plain ints starting from 0. (It avoids creating a string tag->index dict.)

Docs can, however, have more than one tag, and it's quite natural to think such other tags will be symbolic, just as a category-name. However, mixing plain-int and string tags in the same corpus will lead to confusing results: the ints are taken as literal indexes into the array of docvecs-in-trainig, but any string is assigned the next-available-int when it is first encounterd. Unless the user has been careful not to use that same int as a plain int ID, the int and string will indvertently share the same vector. 

Using *only* ints or strings will work well in the current code, but the scan can probably be adapted to avoid this problem. (First idea: strings are given temporary provisional indexes upon first discovery: perhaps, negative indexes. Only at the end of the full corpus scan are all string tags – those in `model.docvecs.doctags` and `model.docvecs.index2doctag` given real final positive int indexes, at the end of all other indexes. In the case where only string tags are used, their assigned indexes will wind up as the same starting-from-zero ints as before.)

gojomo:
Avoids assigning any string-doctags provided to indexes that overlap with raw-int doctags provided. Not super-happy with the 'offset' indexing I've used here - could be confusing to people using mixed tags - but alternatives (like keeping int-indexed and string-keyed doctags in totally separate structures) seems worse and would complicate training routines.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - wget 'http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh' -O miniconda.sh
   - chmod +x miniconda.sh
   - ./miniconda.sh -b
-  - export PATH=$PREFIX/bin:$PATH
+  - export PATH=/home/travis/miniconda2/bin:$PATH
   - conda update --yes conda
 install:
   - conda create --yes -n gensim-test python=$TRAVIS_PYTHON_VERSION pip atlas numpy scipy
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - wget 'http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh' -O miniconda.sh
   - chmod +x miniconda.sh
   - ./miniconda.sh -b
-  - export PATH=/home/travis/miniconda/bin:$PATH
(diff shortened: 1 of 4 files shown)


piskvorky:
Looks good, but can you also commit an update into CHANGELOG?

Something that explains the API difference, for the next release. And summarize what behaviour that now works will break in the new release, so people know if/how to upgrade.

gojomo:
Though the internal list property mapping indexes->string doctags changes its name/interpretation slightly, other than that name change, everything that worked will work the same way... and the mixed-case that wasn't doing the right thing will now work. 

Bullets to that effect added to CHANGELOG.txt

gojomo:
The current Travis CI build failure seems unrelated to this content; rather the 'conda' command is missing when it's needed. It seems the Miniconda-latest-Linux-x86_64.sh install script may have changed its destination path slightly (to `/home/travis/miniconda2` from `/home/travis/miniconda`).

gojomo:
New PATH in .travis.yaml fixed build issue. Merging.

piskvorky:
Perfect, thanks!

PR #508, opened 2015-11-03
[WIP] Replace frequency counting using dict by a combination of hyperloglog and CountMinSketch

janrygl:
Re. #400 and related to #406.
### Changes:
- Pruning words with lower frequencies is removed.
- Approximative algorithms are used:
  - `hyperloglog` for counting vocabulary size (default vocabulary size error 1%)
  - `CountMinSketch` for frequency count (if F is true freq, result is in range [F, F \* (1+e)], default e = 0.01)
- Uses `defaultdicts` for chunk processing (it can be rewritten to parallel chunk building)
- May need to update constants:
  - threshold can be lower with higher vocabulary size
  - chunks can be bigger (sentence is expected to contain 20 words, can be 10)
  - allowed errors can be lower
### Comparison of old and new implementation:
- 100 000 000 randomly generated sentences (10 % stop words, 90 % random)
- max_vocab_size=40000000 (ignored in new implementation)

| Implementation | Old | New |
| --- | --- | --- |
| Time (h) | 6 h 9 min | 28 h 26 min |
| Time (ratio) | 100% | 462 % |
| Vocabulary size | 29 700 928 | 761 363 290 |
- 20 000 000 randomly generated sentences (10 % stop words, 90 % random)
- max_vocab_size=40000000 (ignored in new implementation)

| Implementation | Old | New |
| --- | --- | --- |
| Time (h) | 2h 26 min | 11 h 26 min |
| Time (ratio) | 100% | 469 % |
| Vocabulary size | 12 692 683 | 309 776 447 |
- 10 000 000 randomly generated sentences (10 % stop words, 90 % random)
- max_vocab_size=40000000 (ignored in new implementation)

| Implementation | Old | New |
| --- | --- | --- |
| Time (h) | 36 min | 2 h 53 min |
| Time (ratio) | 100% | 480 % |
| Vocabulary size | 4 077 048 | 81 486 060 |
- from nltk.corpus import gutenberg
- guttenber.sents() (98551 sentences)

| Implementation | Old | New |
| --- | --- | --- |
| Time (s) | 14 s | 34 s |
| Time (ratio) | 100% | 240 % |
| Vocabulary size | 618 421 | 619 564 |
- from nltk.corpus import brown
- brown.sents()  57340 sentences)

| Implementation | Old | New |
| --- | --- | --- |
| Time (s) | 7 s | 24 s |
| Time (ratio) | 100% | 340 % |
| Vocabulary size | 503 281 | 504 632 |

### DIFF
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -60,9 +60,10 @@ two tokens (e.g. `new_york_times`):
 import sys
 import os
 import logging
-import itertools
 import random
 from collections import defaultdict
+from itertools import islice
+
 
 import numpy as np
 from six import iteritems, string_types
@@ -254,7 +255,7 @@ class Phrases(interfaces.TransformationABC):
                 # TODO in parallel version, log information must be changed
                 logger.info("PROGRESS: at sentence #%i" % (sentence_no + start_position))
 
-            for token_bigram in itertools.izip(sentence, sentence[1:]):
+            for token_bigram in zip(sentence, sentence[1:]):
                 sub_vocab[token_bigram[0]] += 1
                 sub_vocab[delimiter.join(token_bigram)] += 1
                 total_words += 1
@@ -286,7 +287,7 @@ class Phrases(interfaces.TransformationABC):
 
         while True:
             try:
-                sentences_chunk = itertools.islice(sentences_iterator, step)
+                sentences_chunk = islice(sentences_iterator, step)
                 sub_vocab = Phrases.add_subvocab(
                     sentences_chunk, current_position)
                 current_position += step
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -72,15 +72,7 @@ import hyperloglog
 logger = logging.getLogger(__name__)
 
 
-class DictCounter(defaultdict):
-    def __init__(self):
-        defaultdict.__init__(self, int)
-
-    def increment(self, key, increase=1):
-        self[key] += increase
-
-
-class CountMinimalSketchCounter(object):
+class CountMinSketchCounter(object):
     """This class is using two algorithms:
      * count-min sketch (https://en.wikipedia.org/wiki/Count%E2%80%93min_sketch)
         is used for determining the approximate frequency of token.
@@ -93,7 +85,7 @@ class CountMinimalSketchCounter(object):
     """
     def __init__(
             self, overestimated_freq_count_error=0.01, approximation_factor=5E-6, vocabulary_size_error=0.01,
-            conservative=True, prime=66405897020462343733):
+            conservative=False, prime=66405897020462343733):
         """
         `approximation_factor`: (epsilon in c implementation)
             amount of tolerable error in the returned error count: 1 + approximation_factor * true_value
@@ -228,7 +220,7 @@ class Phrases(interfaces.TransformationABC):
         self.threshold = threshold
         self.max_vocab_size = max_vocab_size
         self.average_sentence_size = average_sentence_size
-        self.vocab = CountMinimalSketchCounter()  # mapping between utf8 token => its count
+        self.vocab = CountMinSketchCounter()  # mapping between utf8 token => its count
         self.min_reduce = 1  # ignore any tokens with count smaller than this
         self.delimiter = delimiter
 
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -60,13 +60,119 @@ two tokens (e.g. `new_york_times`):
 import sys
 import os
 import logging
+import itertools
+import random
 from collections import defaultdict
 
+import numpy as np
 from six import iteritems, string_types
-
 from gensim import utils, interfaces
+import hyperloglog
 
 logger = logging.getLogger(__name__)
+
+
+class DictCounter(defaultdict):
+    def __init__(self):
+        defaultdict.__init__(self, int)
+
+    def increment(self, key, increase=1):
+        self[key] += increase
+
+
+class CountMinimalSketchCounter(object):
+    """This class is using two algorithms:
+     * count-min sketch (https://en.wikipedia.org/wiki/Count%E2%80%93min_sketch)
+        is used for determining the approximate frequency of token.
+     * hyperloglog (https://pypi.python.org/pypi/hyperloglog)
+        is used for counting number of unique tokens in vocabulary
+
+    This version is refactored version of original Miguel Cabrera's version
+        https://github.com/piskvorky/gensim/pull/270
+    and expanded by hyperloglog library (suggested by Radim Řehůřek and Miguel Cabrera).
+    """
+    def __init__(
+            self, overestimated_freq_count_error=0.01, approximation_factor=5E-6, vocabulary_size_error=0.01,
+            conservative=True, prime=66405897020462343733):
+        """
+        `approximation_factor`: (epsilon in c implementation)
+            amount of tolerable error in the returned error count: 1 + approximation_factor * true_value
+            influence a width of the sketch
+        `overestimated_freq_count_error`: (delta in c implementation)
+            probability that the returned token frequency is not within the accepted error.
+            influence a depth of the sketch
+        `vocabulary_size_error`: allowed percentage difference between real and estimated unique token counts
+        `prime`: prime number used for hash function generation
+        """
+        if not 0 <= overestimated_freq_count_error <= 1:
+            raise ValueError("delta must be between 0 and 1, exclusive")
+        if not 0 <= approximation_factor <= 1:
+            raise ValueError("epsilon must be between 0 and 1, exclusive")
+        if not 0 < vocabulary_size_error < 1:
+            raise ValueError("`vocabulary_size_error` is a percentage error, it must be in open range (0, 1)")
+
+        # initialise hyperloglog counter for approximate unique word count
+        self.counter = hyperloglog.HyperLogLog(vocabulary_size_error)
+        self.conservative = conservative
+
+        self.prime = prime
+
+        self.width = int(np.ceil(2 / approximation_factor))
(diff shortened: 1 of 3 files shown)


gensim/models/phrases.py
```diff
@@ -242,7 +363,7 @@ def __getitem__(self, sentence):
         return [utils.to_unicode(w) for w in new_s]
 
 
-if __name__ == '__main__':
+if __name__ == '__main__' and 0:
```
piskvorky: What is this for?

gensim/models/phrases.py
```diff
@@ -106,17 +206,23 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
...
+            max_vocab_size / average_sentence_size sentences is used to build python
+            dictionary per thread.
+            Lowering `average_sentence_size` will increase memory demands.
         """
         if min_count <= 0:
-            raise ValueError("min_count should be at least 1")
-
+            raise ValueError("min_count should be 1")
+        if min_count > 1:
+            logger.warning("min_count should be 1")
```
gojomo: Why retain `min_count` if it now must be a constant 1?

gensim/models/phrases.py
```diff
@@ -130,54 +236,68 @@ def __str__(self):
...
-
-        logger.info("merging %i counts into %s", len(vocab), self)
-        self.min_reduce = max(self.min_reduce, min_reduce)
-        for word, count in iteritems(vocab):
-            self.vocab[word] += count
-        if len(self.vocab) > self.max_vocab_size:
-            utils.prune_vocab(self.vocab, self.min_reduce)
-            self.min_reduce += 1
+        sentences_iterator = iter(sentences)
+        step = int(self.max_vocab_size / self.average_sentence_size)
```
gojomo: The logic for using step-sizes that are `max_vocab_size` (previously a count of unique tokens) divided by `average_sentence_size` (which isn't really known without a full-scan, and includes non-unique token counts) isn't clear. If `step` is actually an important parameter for performance or memory use, maybe allow it to be specified directly?

gensim/models/phrases.py
janrygl: The only reason I left it is not to break API (it has limited functionality to subtract this count from all frequencies before counting final bi-gram score).
I would be grateful for example how to solve the parameter removal elegantly.

gensim/models/phrases.py
```diff
@@ -60,15 +60,115 @@
...
+        logging.info("Creating a Count Min-Sketch with dimension {0}x{1}".format(self.depth, self.width))
+
+        self.hash_functions = [self._generate_hash_function() for d_i in range(self.depth)]
+        self.count = np.zeros((self.depth, self.width), dtype='int32')
+
+    def increment(self, key, increase=1):
+        """Modified version of update, as the increment is done in a dict-like manner
+        """
+        self.counter.add(key)  # add key to unique vocabulary
+        chat = self["key"]
```
piskvorky: This looks suspicious.

piskvorky:
Starting to look up, good progress!

Next steps:

* separate the MinCount/HyperLogLog logic into a separate module, decouple it from Phrases
* make Phrases flexible in what counting logic it uses (an injection param that accepts a `counter` object, with an API that supports MinCount/plain dict with pruning/whatever else conforms to the `counter` API)
* use the same counter injection in word2vec / doc2vec
* look into optimizations. The 5x perf hit is actually not that bad, I was expecting worse. But it seems with a bit of C we could make this much faster still, perhaps even comparable. See e.g. [PyBloomFilterMmap](https://github.com/axiak/pybloomfiltermmap/blob/master/src/bloomfilter.h) for a set of fast hashing routines doing a very similar thing.

gensim/models/phrases.py
gojomo: In my opinion, when the underlying behavior changes, it can be more safe/honest to break the API (give a "no such parameter min_count" error) than maintain a superficial compatibility that's no longer having the original effect (silently altering the API) or indeed fails (with a thrown error) in most situations where the old values would have been supported. 

Perhaps here it's better for both implementations to co-exist, at least for a while, as different classes with slightly-different options? That may also make it easier to compare their performance, and allow any project that prefers the old precision, or needs reproducibility of prior runs, to stay with the original implementation unless/until they want the benefits of the new.

gojomo:
I'm pretty excited about an improved phrases-detection, so here are a bunch of random thoughts/comments:

*Re: benchmarks*

Can peak memory usage be captured, as well? (Might any of the up-to-5x-slowdowns be caused by swapping?)

Are the timings just for an initial 'survey' pass, or do they also include one 'convert-to-phrases' pass? 

Would be very useful to see differences in vocab/bigram count, inferred-phrases, and speed/memory performance on a real corpus, like Wikipedia. (I could possibly try this sometime next week.)

*Re: possible optimizations*

It seems the 'step' (40,000,000 / 20 = 2,000,000) chunks of sentences are used to minimize calls to the (expensive?) `increment()`. Two possible alternate optimizations to minimize redundant slot-hashing come to mind: 

(1) caching the list of `column` indexes for the last-N/most-common tokens; 

(2) rather than using 2M-batches for precise counts, and at the end doing `increment()` for all tokens then starting with a fresh precise count, use something a bit like the old capped-size dictionary but when hitting the max-size, the 'purge' does `increment()`s for the N purged keys. (I'm not quite sure if such a purge would be best prioritized by lowest-counts, highest-counts, or oldest-keys, in order to minimize the times the same key is `increment()`ed.)

Can CountMinSketch objects of the same shape be added together? If so that's one plausible path to parallelization. (More generally, here and other parts of gensim may benefit from the idea of a corpus that can be read from many files, or many start-points in a file, by separate processes, to get away from the one-linear-reader-handing-items-to-many-workers pattern that doesn't work well with the Python GIL & cross-process serialization overhead.)

*Re: misc*

It might be interesting to store the unigram and bigram counts (and overall unique tallies) in separate structures, for more visibility into what's happening and to give them separate precisions. 

It seems the tunable count parameters (or documentation of same) should include some relationship for 'expected unique inserts'. (Don't the actual error margins depend on how saturated the structure gets, like a Bloom Filter, and beyond a certain chosen load factor the errors go beyond target levels?)

gensim/models/phrases.py
piskvorky: Agreed with @gojomo -- we don't want to keep dead code around.

And in case of `Phrases`, we don't need the "old" behaviour either. The Phrases functionality is new and not widely used yet, so a clear release note saying things have changed seems enough here.

piskvorky:
I'll also push in other improvement to phrase detection (not related to counting)... probably next weekend.

mfcabrera:
Hi, nice too see someone continued the work. sadly, due to time/personal issues I couldn't continue. I just wanted to share, that  I did profile my original code, and I found out that one of the things hitting the performance was the calculation of the hash functions. I believe some Cython magic might come handy. Is @janrygl still working on this PR right? Let me know if I can help somehow.

piskvorky:
@janrygl has other duties now, so this PR is not "active" at the moment :) Any help welcome!

Btw there are good, fast hash functions in `pybloomfiltermmap` (item 4 in my list above). There's a lot of overlap here with that project, both conceptually and implementation-wise.

tmylk:
Pinging @mfcabrera @janrygl - do you think it can be part of gensim January release?

janrygl:
@tmylk It depends on priorities defined by @piskvorky , I got hurt my right hand in the beginning of January and I am behind the schedule with all my projects.

piskvorky:
@janrygl won't have time for this in January for sure; don't know about @mfcabrera .

It's a great little algorithmic project though, very pleasant. I wish I had time to tinker with this myself, would make for an exciting blog post / series!

piskvorky:
Relevant read:

[Extension to hyperloglog as used by Google](http://blog.acolyer.org/2016/03/17/hyperloglog-in-practice-algorithmic-engineering-of-a-state-of-the-art-cardinality-estimation-algorithm/)

piskvorky:
Another interesting article https://www.linkedin.com/pulse/from-count-min-sketch-tree-23-guillaume-pitel, by @PitZeGlide.

thescopan:
Any updates on this branch. Phrases implementation is so slow that it is making me switch to a different library for doc2vec. Any update will be helpful.

piskvorky:
@thescopan I don't think so -- feel free to contribute.

Also, a re-implementation of the existing phrases in C/Cython would be appreciated too. It's a really small and trivial change, but nobody did it yet. CC @tmylk .

tmylk:
Cythonising Phrases will be done this summer as part of GSOC

piskvorky:
Awesome! This is a much needed functionality.

A fast & scalable collocation (phrase) detection is sorely missing (even in our own non-open-source projects).

tmylk:
More specifically it's on @prakhar2b 's [timeline](https://github.com/numfocus/gsoc/blob/master/2017/proposals/prakhar_gsoc_17.md) for end of June.

piskvorky:
More discussion on hyperloglog as used inside reddit:
https://redditblog.com/2017/05/24/view-counting-at-reddit/

menshikh-iv:
Ping @janrygl, what status of this PR?

piskvorky:
@menshikh-iv isn't this one of our Google Summer of Code projects this year?

menshikh-iv:
@piskvorky [cythonising phrases](https://github.com/numfocus/gsoc/blob/master/2017/proposals/prakhar_gsoc_17.md) is a project of current GSoC, but, as I understand, this is not the same that the topic of this PR.

piskvorky:
Nearly the same -- efficient "counting" is the biggest challenge there. Phrases should definitely use this new counting functionality, among other modules.

It's an extremely common task, widely useful, and that's why I'd like this to be an independent library.

menshikh-iv:
Connected with #1446

PR #513, opened 2015-11-07
LDAModel depracation warning (issue 494)

mattilyra:
I decided to not call `utils.grouper` with `as_numpy=False` and break the Pyro stuff as I thought we can work around that. So now there's just a check in `.inference` that makes sure the indices are `ints` when a numpy array is passed in.

This seemed like a better choice as it doesn't introduce any breaking changes unnecessarily.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -368,6 +368,8 @@ class LdaModel(interfaces.TransformationABC):
         # to Blei's original LDA-C code, cool!).
         for d, doc in enumerate(chunk):
             if isinstance(doc, numpy.ndarray):
+                if len(doc) == 0:
+                    continue
                 ids = doc[:, 0].astype(numpy.int_)
                 cts = doc[:, 1]
             else:
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -367,8 +367,12 @@ class LdaModel(interfaces.TransformationABC):
         # Lee&Seung trick which speeds things up by an order of magnitude, compared
         # to Blei's original LDA-C code, cool!).
         for d, doc in enumerate(chunk):
-            ids = [id for id, _ in doc]
-            cts = numpy.array([cnt for _, cnt in doc])
+            if isinstance(doc, numpy.ndarray):
+                ids = doc[:, 0].astype(numpy.int_)
+                cts = doc[:, 1]
+            else:
+                ids = [id for id, _ in doc]
+                cts = numpy.array([cnt for _, cnt in doc])
             gammad = gamma[d, :]
             Elogthetad = Elogtheta[d, :]
             expElogthetad = expElogtheta[d, :]
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,18 @@ NEXT
   and `LdaMulticore` that may be reliant on the old tuple layout of
   `(probability, word)`.
 
+* Forwards compatibility for  NumPy > 1.10 (Matti Lyra, #494)
+  - LdaModel and LdaMulticore produce a large number of DepracationWarnings from
+    .inference() because the term ids in each chunk returned from utils.grouper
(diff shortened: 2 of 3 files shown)


gensim/models/ldamodel.py
```diff
@@ -367,7 +367,15 @@ def inference(self, chunk, collect_sstats=False):
         # Lee&Seung trick which speeds things up by an order of magnitude, compared
         # to Blei's original LDA-C code, cool!).
         for d, doc in enumerate(chunk):
-            ids = [id for id, _ in doc]
+            if isinstance(doc, numpy.ndarray):
+                if len(doc) == 0:
```
piskvorky: `len(array)` will return the length along the first axis (=number of rows). Is that always the case here?

gensim/models/ldamodel.py
```diff
@@ -367,7 +367,15 @@ def inference(self, chunk, collect_sstats=False):
...
+                    continue
+
+                # if utils.grouper was called with as_numpy=True all the
+                # term IDs will be floats, let's convert them to ints
+                ids = doc[:, 0].astype(numpy.int_)
```
piskvorky: What if `doc` is only 1-dimensional? Also, probably better to cast to plain `int` instead of `numpy.int_`. What is `numpy.int_`?

piskvorky:
Thanks Matti.

We can do these ad-hoc fixes in various places, though I still think it's cleaner to solve this at the place of origin, where the float values are generated (`as_numpy=False`).

Also, what's the advantage of your `if` solution vs. a direct `ids = [int(id) for id, _ in doc]`?

mattilyra:
If the `doc` is already a numpy array it's faster to cast it to `int_` (the default numpy integer, 64bit I guess on most machines) instead of calling `int()` in the list comprehension - at least on the machine I tried.

I didn't consider the case of 1-d documents, the `len(doc)` also won't be semantically consistent with them. I'll change the call to `utils.grouper` in the end that'll end up being much cleaner as you say.

I thought I was being so clever.

piskvorky:
:D

You are clever, just in this case I think simplicity and consistency are more important (it's not the bottleneck).

mattilyra:
Ok, so I decided to keep the check for the IDs being ints since I assume numpy will complain about it, pyro or no pyro. The default is now for the chunks to be not numpy, and there's an option on `.update` to make them numpy.

piskvorky:
That looks like a nice solution. Thanks.

If it later turns out this "make sure ids are ints" functionality is more widely used, we could factor it out into a separate function, something like `matutils.check_vector()`.

Travis reports all tests as failing though:

```
ERROR: testTransform (gensim.test.test_ldamodel.TestLdaMulticore)
  File "/home/travis/build/piskvorky/gensim/gensim/test/test_ldamodel.py", line 382, in setUp
    self.model = self.class_(corpus, id2word=dictionary, num_topics=2, passes=100)
  File "/home/travis/build/piskvorky/gensim/gensim/models/ldamulticore.py", line 141, in __init__
    gamma_threshold=gamma_threshold)
  File "/home/travis/build/piskvorky/gensim/gensim/models/ldamodel.py", line 331, in __init__
    self.update(corpus, chunks_as_numpy=use_numpy)
TypeError: update() got an unexpected keyword argument 'chunks_as_numpy'
```

piskvorky:
Ah I see, the fails are coming from `LdaMulticore`, which inherits from `LdaModel` class but probably overrides its `update()`. Can you add the same parameter there as well?

piskvorky:
Thank you very much Matti, ready to merge!

Can you add yourself + a short description of this change to CHANGELOG?

mattilyra:
I always forget the CHANGELOG

piskvorky:
Great, thanks!

piskvorky:
That's probably unrelated to this PR -- or do you see a connection?

tmylk:
It was passing on Sunday and it's the only change I can find around it.
Numpy and scipy versions remained the same.

Issue #485, opened 2015-10-16
Mallet 2.0.8 read_doc_topics format changed


tmylk:
Hi,

Seems like read_doctopics in gensim.models.wrappers.ldamallet needs to be changed.  Am I right?

In Mallet 2.0.7, doctopics.txt looks like:
#doc name topic proportion ...

0	0	158	0.1182170542635659	129	0.07170542635658915	111	0.04844961240310078 ...

So it makes sense to zip things in read_doctopics: zip(parts[::2], parts[1::2])

In Mallet 2.0.8RC2, doctopics.txt looks like:
0	0	2.8037970188132654E-5	1.9405424585156917E-5	2.8328585124176796E-5 ...

I am hoping topic probabilities follow the same order (topic1, topic2, ...) in all lines of this new doctopics.txt file.

Also This [line](https://github.com/piskvorky/gensim/blob/develop/gensim/models/wrappers/ldamallet.py#L51) needs to be changed additionally (no header in the new version).

From [mailing list](https://groups.google.com/d/msg/gensim/L4ME7M6kbP0/bmfg8QDHBwAJ)

ziky90:
Is anyone working on this already?
Would you like to keep the compatibility with Mallet 2.0.8RC2 as well? For example using some automatic mallet version determining or version parameter?

piskvorky:
@ziky90 yes, that sounds like the best option: detect dynamically what format it is (pre/post 2.0.8) and adjust parsing accordingly.

We definitely want to (have to) keep compatibility with the old Mallet too.

I don't know whether there have been any other changes in Mallet. If so, we may have to introduce a new explicit parameter into the wrapper (`mallet_version`), or detect the version explicitly somehow, rather than just adjust the format parsing. If it's just the parsing, probably no need to bother users with it, we can auto-detect.

tmylk:
@jeiranj had ideas for the PR on [the mailing list](https://groups.google.com/d/msg/gensim/L4ME7M6kbP0/bmfg8QDHBwAJ)

cscorley:
Closing, this issue should be fixed by @jeiranj!

PR #516, opened 2015-11-09
fixes #484 and #485: id2word does not need to be a Dictionary and making read_doc_topics compatible with Mallet2.0.8

### LINKED ISSUE #484: LDAMallet id2word shouldn't have to be a corpora.Dictionary with token2id attribute
```python
id2word_fails = dict(enumerate(words))
id2word_passes = gensim.corpora.Dictionary.from_corpus(corpus_lda, dict(enumerate(words)))
```

From [mailing list](https://groups.google.com/d/msg/gensim/oimusXhkinI/H-tpcMjGBwAJ)

Error:
``` python

  File "test_lda.py", line 571, in <module>

    main()

  File "test_lda.py", line 545, in main

    lda = gensim.models.wrappers.LdaMallet(mallet_path="/jj/mallet-2.0.7/bin/mallet", corpus=corpus_lda, num_topics=200, id2word=id2word, workers=4, optimize_interval=10, iterations=1000, pre

fix='mallet_')

  File "/usr/local/lib/python2.7/dist-packages/gensim/models/wrappers/ldamallet.py", line 101, in __init__

    self.train(corpus)

  File "/usr/local/lib/python2.7/dist-packages/gensim/models/wrappers/ldamallet.py", line 162, in train

    self.word_topics = self.load_word_topics()

  File "/usr/local/lib/python2.7/dist-packages/gensim/models/wrappers/ldamallet.py", line 193, in load_word_topics

    tokenid = self.id2word.token2id[token] if hasattr(self.id2word, 'token2id') else int(token)


ValueError: invalid literal for int() with base 10: 'year'
```

Solution suggested by jj:
change
``` python
tokenid = self.id2word.token2id[token] if hasattr(self.id2word, 'token2id') else int(token)
```
to 
``` python
word2id = dict((v, k) for k, v in iteritems(self.id2word))
tokenid = self.word2id[token]
```
This should work even if id2word is not provided as this line takes care of that at the very beginning.
https://github.com/piskvorky/gensim/blob/develop/gensim/models/wrappers/ldamallet.py#L83

jeiranj:
#484 :  to resolve this, created word2id dictionary by reversing id2word and from there it is easy to get tokenid by calling word2id on the particular token you want.  This should work even if id2word is not provided as this line (https://github.com/piskvorky/gensim/blob/develop/gensim/models/wrappers/ldamallet.py#L83) takes care of that at the very beginning.

#485 :  to resolve this, made `read_doctopics()` a def for LdaMallet class.  We read the first line of the file, and check whether it started with '#doc ' (Mallet 2.0.7), then we skip the header line as before.  If there is no header line however (Mallet 2.0.8), we do not skip the first line.  The next change involved checking the number of items in each line to see whether the file contains 'topic#  probability' ordering (Mallet 2.0.7) or just 'probabilities' (ordered by increasing topic# as in Mallet 2.0.8). After this check, doc probabilities are obtained accordingly.

#517 :  modified `top_topics()` in ldamodel.py to account for coherence score correctly.

### DIFF
--- a/gensim/models/wrappers/ldamallet.py
+++ b/gensim/models/wrappers/ldamallet.py
@@ -41,24 +41,7 @@ from gensim.utils import check_output
 
 logger = logging.getLogger(__name__)
 
-
-def read_doctopics(fname, eps=1e-6):
-    """
-    Yield document topic vectors from MALLET's "doc-topics" format, as sparse gensim vectors.
-
-    """
-    with utils.smart_open(fname) as fin:
-        next(fin)  # skip the header line
-        for lineno, line in enumerate(fin):
-            parts = line.split()[2:]  # skip "doc" and "source" columns
-            if len(parts) % 2 != 0:
-                raise RuntimeError("invalid doc topics format at line %i in %s" % (lineno + 1, fname))
-            doc = [(int(id), float(weight)) for id, weight in zip(parts[::2], parts[1::2]) if abs(float(weight)) > eps]
-            # explicitly normalize probs to sum up to 1.0, just to be sure...
-            weights = float(sum([weight for _, weight in doc]))
-            yield [] if weights == 0 else sorted((id, 1.0 * weight / weights) for id, weight in doc)
-
-
+            
 
 class LdaMallet(utils.SaveLoad):
     """
@@ -173,13 +156,14 @@ class LdaMallet(utils.SaveLoad):
         cmd = cmd % (self.fcorpusmallet() + '.infer', self.finferencer(), self.fdoctopics() + '.infer', iterations)
         logger.info("inferring topics with MALLET LDA '%s'" % cmd)
         check_output(cmd, shell=True)
-        result = list(read_doctopics(self.fdoctopics() + '.infer'))
+        result = list(self.read_doctopics(self.fdoctopics() + '.infer'))
         return result if is_corpus else result[0]
 
 
     def load_word_topics(self):
         logger.info("loading assigned topics from %s" % self.fstate())
         wordtopics = numpy.zeros((self.num_topics, self.num_terms), dtype=numpy.float32)
+        word2id = dict((v, k) for k, v in iteritems(self.id2word))
         with utils.smart_open(self.fstate()) as fin:
             _ = next(fin)  # header
             self.alpha = numpy.array([float(val) for val in next(fin).split()[2:]])
@@ -188,7 +172,7 @@ class LdaMallet(utils.SaveLoad):
             for lineno, line in enumerate(fin):
                 line = utils.to_unicode(line)
                 doc, source, pos, typeindex, token, topic = line.split(" ")
-                tokenid = self.id2word.token2id[token] if hasattr(self.id2word, 'token2id') else int(token)
+                tokenid = word2id[token]
                 wordtopics[int(topic), tokenid] += 1
         logger.info("loaded assigned topics for %i tokens" % wordtopics.sum())
         self.wordtopics = wordtopics
@@ -203,7 +187,7 @@ class LdaMallet(utils.SaveLoad):
         Return an iterator over the topic distribution of training corpus, by reading
         the doctopics.txt generated during training.
         """
-        return read_doctopics(self.fdoctopics())
+        return self.read_doctopics(self.fdoctopics())
 
 
     def show_topics(self, num_topics=10, num_words=10, log=False, formatted=True):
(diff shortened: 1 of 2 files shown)


cscorley:
Hi jj, the change related #517 looks fine to me. Is there a chance you can rebase these commits onto the latest develop HEAD? Thanks!

piskvorky:
Thanks so much @jeiranj !

Killing three flies at once, perfect :)

Re. #485: is there any way we can test this?

@cscorley @tmylk can you review?

piskvorky:
Actually, now that I think of it, #485 should be easy to test. The current Mallet tests accept a path to Mallet executable (by means of an environment var `MALLET_HOME`).

So if we run the existing unit tests twice, setting `MALLET_HOME` to 2.0.7 in one case and > 2.0.8 in the other, we should be checking both code paths. Right?

gensim/models/wrappers/ldamallet.py
```diff
@@ -173,13 +156,14 @@ def __getitem__(self, bow, iterations=100):
...
         check_output(cmd, shell=True)
-        result = list(read_doctopics(self.fdoctopics() + '.infer'))
+        result = list(self.read_doctopics(self.fdoctopics() + '.infer'))
         return result if is_corpus else result[0]
 
 
     def load_word_topics(self):
         logger.info("loading assigned topics from %s" % self.fstate())
         wordtopics = numpy.zeros((self.num_topics, self.num_terms), dtype=numpy.float32)
+        word2id = dict((v, k) for k, v in iteritems(self.id2word))
```
cscorley: wondering why the check for token2id attr was dropped (L191 previously)? seems like it would eliminate the need for a new duplicate one when possible. could be something like:

```
if hasattr(self.id2word, 'token2id'):
    word2id = self.id2word.token2id
else:
    word2id = # current line
```

jeiranj: you are absolutely right!

gensim/models/wrappers/ldamallet.py
```diff
@@ -244,3 +228,27 @@ def show_topic(self, topicid, topn=10):
...
+        fin.close()
+        with utils.smart_open(fname) as fin:
+            if first_line.startswith("#doc "):  # check the file header
+                next(fin)   # skip the header line
+            for lineno, line in enumerate(fin):
+                parts = line.split()[2:]  # skip "doc" and "source" columns
+                if len(parts) == 2 * self.num_topics:
+                    doc = [(int(id), float(weight)) for id, weight in zip(parts[::2], parts[1::2]) if abs(float(weight)) > eps]
+                elif len(parts) == self.num_topics:
+                    doc = [(int(id), float(weight)) for id, weight in list(enumerate(parts)) if abs(float(weight)) > eps]
```
cscorley: very clever solution! :smile_cat: 

I think it would be helpful to add a comment over denoting why there's a need for the two different if-statements (e.g., this is for mallet version x, which formats the lines as `...`) just to make it clear what changes happened between the two versions.

For this particular step, there shouldn't be a need to use `list(enumerate(parts))`. Things should be fine with it being a plain `enumerate(parts)` since we are iterating over it with a loop anyway. Plus, this will save a lil memory not indirectly duplicating the parts list itself

cscorley:
Haven't really followed the changes Mallet is requiring, but from a code standpoint it seems good to me! :+1:

gensim/models/wrappers/ldamallet.py
```diff
@@ -244,3 +228,27 @@ def show_topic(self, topicid, topn=10):
...
+
+
+    def read_doctopics(self, fname, eps=1e-6):
+        """
+        Yield document topic vectors from MALLET's "doc-topics" format, as sparse gensim vectors.
+
+        """
+        fin = utils.smart_open(fname)
+        first_line = next(fin)
+        fin.close()
```
cscorley: Feel iffy about opening and closing like this.

Would this accomplish the same or similar if we moved the check into the loop?

```
with utils.smart_open(fname) as fin:
      for lineno, line in enumerate(fin):
           if lineno == 0 and line.startswith("#doc "): # first line and it is the header line, no need to process it
                continue 
           parts = line.split()[2:]  # skip "doc" and "source" columns
           ...
```

piskvorky:
@jeiranj do you want to address @cscorley 's points ? Or do we merge as is. Apart from these details, the PR looks fine.

piskvorky:
@jeiranj oh actually I'd ask you for one more thing: could you add yourself + a short description of the change to the project CHANGELOG? Cheers!

jeiranj:
sorry, I am traveling abroad and got little time on my hands now.  i will
address his points most definitely as soon as I get back to the states.

– Jeiran

piskvorky:
@jeiranj no problem, thanks. Have a safe trip.

jeiranj:
it seems tough to have a *safe* trip these days.

– Jeiran

cscorley:
@piskvorky I'd like to make sure this gets in before next release. If needed, I think that this is fine as-is, I can patch in my comments before-hand.

piskvorky:
Alright, I'll merge. Thanks both!

PR #519, opened 2015-11-10
Updated Vowpal Wabbit LDA wrapper to support versions 7.10 to 8.1.1

davechallis:
Main changes are:

* improved detection for start of data area
* avoid repeated command line arguments that are serialised as part of the model


piskvorky:
Looks good, thanks!

Can you explicitly mention somewhere in the docs (module docstring?) what versions are supported?

davechallis:
Yup, good idea, will add that now.

piskvorky:
Thanks Dave.

Is this PR final or are the unit tests forthcoming as well?

davechallis:
I've got some time this Friday, and was planning to work on some tests then
(though I admit I'm not sure how these might work, since they'll depend on
an external executable, but I'll take a look and see if the Mallet wrapper
does something I can base things from).

I'll happy to add them to this pull request, or add them to a future one,
either is fine by me.

Thanks,
Dave

piskvorky:
I think something simple, ala Mallet tests, is enough to start with. The tester will need to have VW installed (and specify `VW_PATH` env var).

The goal is to make sure VW executes & API works, to catch glaring incompatibilities (like the one reported now).

davechallis:
Ah, great, that should be easy enough, I'll work on that tomorrow. I was
more thinking about tests run by CI/jenkins, and making VW in that
environment, but local tests with env variable set sounds like a good start.

Thanks,
Dave

davechallis:
I've added some tests now, they should run if the `VOWPAL_WABBIT_PATH` is set and points to a `vw` executable. Tests currently pass with python 2.7 and 3.4, and with Vowpal Wabbit 7.10 and 8.1.1.

piskvorky:
Perfect! Merging.

Thanks so much @davechallis , much appreciated.

Issue #482, opened 2015-10-15
add 'Word Mover's Distance' implementation to gensim?


gojomo:
"From Word Embeddings to Document Distances" (Kusner et al 2015: http://jmlr.org/proceedings/papers/v37/kusnerb15.pdf) introduces the "Word Mover's Distance" (WMD), a novel distance-between-text-documents measure. It is an adaptation of another distance metric, "Earth Mover's Distance" (EMD), introduced(?) in "A Metric for Distributions with Applications
to Image Databases" (Rubner et al 1998: http://www.cs.jhu.edu/~misha/Papers/Rubner98.pdf), that is useful for comparing images and can be calculated as a special case of much-older transportation problem optimizations. For text, the WMD leverages word2vec vectors of the documents' individual words, in a way that seems to outperform simple combinations (sum/mean) of those word vectors. 

There's a blog post from OpenTable with some impressive examples of "similar sentences" using WMD on restaurant reviews: http://tech.opentable.com/2015/08/11/navigating-themes-in-restaurant-reviews-with-word-movers-distance/

The paper reports strong kNN classifier results on document classification, and intuitively, the method seems amenable to the selection (or even synthesis) of 'canonical' or 'borderline/contrastive' text segments, perhaps to assist iterative classification tasks.

The author's code is available as a ZIP bundle (http://matthewkusner.com/#page2) – but its python wrappers depend on some older C EMD code of unclear licensing status. There's a bunch of other EMD code around for other projects (esp. image similarity measures) that also might be a good starting point. 

An implementation (perhaps in optimized cython) for gensim might be widely useful.

(Further out: it might be interesting to use WMD to induce a document-embedding. That is, train doc embeddings with random draws of 3 documents, nudging the embeddings so that the relative which-two-are-closest is the same in the embedding space as in the WMD metric.)

piskvorky:
CC @tmylk: backlog of open source tasks.

tmylk:
Added to https://github.com/piskvorky/gensim/wiki/Ideas-&-Feature-proposals

piskvorky:
@olavurmortensen has taken up this issue now.

Regarding the "unclear license": let's clarify with @mkusner. We definitely don't want to include anything without a license / LGPL non-compatible.

gojomo:
[...]

PR #521, opened 2015-11-10
[WIP] Implement Word Mover's Distance (WMD) in Gensim

olavurmortensen:
In response to issue #482.

We want to compute WMD between documents, based on "From Word Embeddings To Documents Distances" by Matt Kusner et al. ([link to paper](http://jmlr.org/proceedings/papers/v37/kusnerb15.pdf)). Using word2vec embeddings, WMD finds the minimum "traveling distance" (in a matter of speaking) of two documents. The method shows great promise in kNN classification, as shown by Vlad Niculae and Matt Kusner in a [blog post](http://vene.ro/blog/word-movers-distance-in-python.html) using Scikit-Learn.

Our implementation will be based on `pyemd`, which is a Python wrapper of a C implementation of EMD (Earth Mover's Distance) ([link to GitHub repo](https://github.com/wmayner/pyemd)) by Will Mayner, which will be directly included in Gensim.

After the simple implementation of a method that computes the WMD, a kNN classifier will be implemented, inspired by the aforementioned blog post of Vlad Niculae and Matt Kusner. Gensim's `docsim` module will be used for this purpose ([link to docsim](https://radimrehurek.com/gensim/similarities/docsim.html)).

At the moment, only WMD between two documents is implemented in this PR, and it's done by importing `pyemd`. This Python wrapper needs to be directly included in Gensim, with no need for this external module.

@tmylk @piskvorky


olavurmortensen:
@tmylk @piskvorky I'm done implementing the `wmdistance` method, using `pyemd` natively in Gensim.

tmylk:
Great. Could you please add unit tests for this new functionality?

piskvorky:
Good job! How do I use it? Please provide an example.

piskvorky:
Thanks @olavurmortensen ! We now have a pair-wise "exact" WMD metric, plus a tutorial.

I'll push some minor code formatting fixes later this week, but I think the first stage is finished. @tmylk do you agree?

For the next step, let's extend this pairwise 1:1 metric into a n:1 similarity search (a static corpus of `n` document queried by a dynamic query of one document).

The baseline option is to simply compute full WMD of each document pair (= `n` WMD calls = each `corpus_i` doc vs. `query` doc), sort by their distance and pick "top k" closest documents for output.

But we also want to explore the optimizations described in the paper: WCD (as a loose lower bound on the "full WMD" distance) and RWMD (as a tighter bound). Both seem straightforward to implement and promise great speed ups.

The expected output is a new class `WmdSimilarity` in `docsim.py` that:

* follows the same interface as the existing MatrixSimilarity class
* constructor accepts a *corpus of documents* (= a sequence of lists of words = same as word2vec `sentences` input) and a *word=>vector mapping* (such as a trained word2vec model)
* instances support the `instance[query]` syntax (query = a list of words again), which will return the ids of `instance.num_best` nearest documents from `corpus` to `query`. It will use the prefetch-and-prune algorithm from Section 4 to avoid computing "full WMD" for documents that cannot be in top-k due to failing lower bound.
* prefetch-and-prune will be parametrized by `m` as in the paper, so that we can set `m=k` for plain WCD and `m=len(corpus)` for full exact WMD.

This is an optimization task, so we want to be comparing performance of all optimizations against that plain "full WMD loop" (which would be something simple like `top_k = sorted(corpus, key=lambda doc: full_wmd(doc, query), reverse=True)[:k]`).

piskvorky:
Or does `pyemd` already offer an implementation of the WCD or RWMD functionality?

olavurmortensen:
I implemented a **pure Python** version of the WMD as well. It uses a brute force method by trying all combinations of word pairings. This is very slow, as the algorithm is O(n^2 m^2) where n and m are the lengths of the documents. This is so slow that it is ill advised to use this in any case. Just comparing two 10 word documents takes half a minute.

The unit tests fail *sometimes* because I'm testing that the WMD is within some bound, and from time to time this bound doesn't hold. This will be annoying it left as it is, as it will randomly fail and cause confusion.

gojomo:
By my understanding,  WMD/EMD as a measure is *not* approximated with any stochastic processes. (I didn't find any obvious references to `rand`/`random' in either the `pyemd` code or the underying Pele/Werman C++ code, except in its demos. And, the `pyemd` unit tests verify calculations match exact target values.) 

So any randomness in the tests is coming from some *other* step, and isn't purely testing the WMD path.

Which test fails, in which python version(s), and how often? 

I suspect it's `testNonzero`, which asserts a specific 'correct' value – but that value must just be chosen from some earlier run of the same code that was assumed to be just-right in its results? 

I see there's an attempt to get reproducible results from the Word2Vec step by using a seed and single-worker-thread. But, I'm not sure that guarantees identical results across python versions (especially with Python 3 string-hash-randomization) or potentially slightly-different math libraries. (So you might be able to pick a value that always succeeds on one python 2.x machine, but then fails always/sometimes in other configurations.)

gojomo:
Since `pyemd` is a PYPI-available package, I think it'd be better to use it from there – rather than copy & paste its code into gensim – unless and until some divergence is necessary.

There are continuing applications of EMD outside of text/WMD (like bitmap image comparisons), so using a shared package could win more eyes/contributions on potential new EMD API features (like access to the flows). 

(Only if @wmayner were to say `pyemd` is frozen/dead, and that status was blocking a pending improvement, would I think that a bulk copy of code elsewhere to be justified... and even then, it might better make sense as a successor EMD project.)

wmayner:
In general I'm in favor of the more distributed, open-source model rather than duplicating code, but as I noted in #482, I'm not planning on doing much work on `pyemd` except as maintainer; i.e., I'll happily look over PRs and try to merge them in a timely manner, but I'm not planning on working on exposing the flows, for example.

If you guys think you'll be doing a lot more with the EMD than `pyemd` currently does then it might make sense to fold it into your codebase, but only if it will really be largely unrecognizable. If there's just a few additions to the API, then I'd suggest forking and sending a PR my way, which I can then merge and publish so you can leave it as a dependency as @gojomo suggested.

piskvorky:
Sure. I'm not terribly worried about the integration, this PR is still a long way away from complete. I was more looking for feedback for the API and algorithmic side.

I suspect we'll need to adapt `pyemd` by the time the dust has settled, one way or another (at least use a Cython/C module for the optimized code paths).

olavurmortensen:
@piskvorky I timed the algorithm, as you suggested. The results are in the table below. As you can see, it explodes quite fast. For 10 000 words, it took a really long time, before my machine ran out of memory and I had to stop.

According to [this page](http://www.ariel.ac.il/sites/ofirpele/fastemd/) (by the original authors Ofir Pele and Michael Werman), the algorithm is O(n^3\*log(n)). In comparison, the brute force algorithm (checking all combinations) is O(n^4). The algorithm is *much* faster than brute force, but it's definitely not cheap.

#words | Time (sec.)
-----------|--------------
10         | 0.00198
100       | 0.123
1000     | 63.871

The page I linked to mentions a "simplified flow network", that is faster to compute. I'm not sure if this is something we would be able to utilize, nor if the implementations provides an interface for it.

The take home message is that one should refrain from computing the distances between documents that are several thousand words long. You typically wouldn't do that anyway, especially after removing stopwords.

piskvorky:
@olavurmortensen what is the `n` in that formula? I would expect the complexity to depend (at least) on two variables: length of document A, length of document B.

Or is `n` a combination of the two lengths (maximum)?

Btw checking all combinations has usually *exponential* complexity (not `O(n^4)`).

wmayner:
@olavurmortensen, the simplified flow network arises from the reduction in edges that can be achieved if a thresholded distance is used such that the average number of sub-threshold distances from a given histogram bin to the other bins is constant (this is the improvement that makes their version of the EMD “fast”, which they describe in the paper). 

So if you can use a thresholded distance, then the time complexity will be more like `O( N^2 * log(N) )`, where `N` is the maximum number of bins in either histogram (@piskvorky). If a non-thresholded distance is used, then the flow network cannot be simplified and the complexity will be `O( N^3 * log(N) )`.

And, as @piskvorky has mentioned, the naive algorithm is indeed exponential, not polynomial.

gensim/test/test_similarities.py
```diff
@@ -257,6 +270,47 @@ class TestMatrixSimilarity(unittest.TestCase, _TestSimilarityABC):
...
+        index.num_best = num_best
+        query = texts[0]
+        sims = index[query]
+    
+        if num_best is not None:
+            # Sparse array.
+            self.assertTrue(numpy.alltrue(sims != 0.0))
+        else:
+            self.assertTrue(sims[0] == 0.0)  # Similarity of a document with itself is 0.0.
+            self.assertTrue(numpy.alltrue(sims[1:] != 0.0))
```
tmylk: should it be < 0? As it is a similarity.
Additional tests that check num_best functionality would be good. That the order is  non-increasing for example.

docs/notebooks/WMD_tutorial.ipynb
```diff
@@ -0,0 +1,184 @@
...
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython2",
+   "version": "2.7.6"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 0
+}
```
tmylk: An application of WmdSimilarity to Yelp corpus is needed in the ipynb.

gensim/test/test_similarities.py
olavurmortensen: @tmylk maybe `_TestSimilarityABC` (which I just extend to WMD) should contain these `num_best` tests? Specifically in `testNumBest`.

wmayner:
Hi folks,

Just wanted to ask that if PyEMD does make it into your codebase, I'm sure it would be much appreciated by the original authors of the EMD algorithm if you could put a note in the relevant documentation saying that any published research that uses your WMD feature should cite the EMD papers (details on those at the end of the [PyEMD README](https://github.com/wmayner/pyemd/blob/develop/README.rst)).

Thanks!

olavurmortensen:
@wmayner That's a good point, thanks for the input.

docs/notebooks/WMD_tutorial.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "sent = 'the food in this restaurant is quite delicious'\n",
+    "sent = sent.split()\n",
+    "sims = index[sent]\n",
+    "print sims"
+   ]
```
tmylk: please add cell output

tmylk:
Planned items for this PR:

- show the text of most similar documents in Yelp corpus in the ipynb
- add references to EMD papers in all files in pyemd folder, and to WMD paper in word2vec.py
- test WCD, RWMD performance vs WMD. Report 3 timings on yelp corpus n=100
- Implement pre-fetch and prune in the similarity class

olavurmortensen:
@tmylk 

Hey everyone. Did some testing of WMD, WCD and RWMD that I just wanted to share. For those that don't know, WCD and RWMD are lower bounds to WMD that are fast to compute.

Below is a plot where I computed all three distances for random sentences (words randomly sampled from vocab), length between 4 and 9 words, in all 1 000 sentences. WMD is sorted, and WCD and RWMD are sorted according to WMD. We see that WCD and RWMD are both less than WMD, which is what we want, so that's good. RWMD is a much tighter bound than WCD, which is what we expect. 



Below is a plot of the running times when computing the distances, where the y axis is on a log scale. Everything is as expected, except that the lower bounds' running times, especially WCD, are on rare occasions quite large.



I will most likely be doing some optimizing at some point, so the latter of the plots is subject to change. Hopefully I can increase the gap between WMD and the lower bounds' running times, and make them less sporadic.

That's all, just wanted to share this.

3cos:
With regard to making these distance measures practical, I just found paper about approximating a knn graph you may be interested in.

http://wwwconference.org/proceedings/www2011/proceedings/p577.pdf

ABSTRACT K-Nearest Neighbor Graph (K-NNG) construction is an important operation with many web related applications, including collaborative filtering, similarity search, and many others in data mining and machine learning. Existing methods for K-NNG construction either do not scale, or are specific to certain similarity measures. We present NN-Descent, a simple yet efficient algorithm for approximate K-NNG construction with arbitrary similarity measures. Our method is based on local search, has minimal space overhead and does not rely on any shared global index. Hence, it is especially suitable for large-scale applications where data structures need to be distributed over the network. We have shown with a variety of datasets and similarity measures that the proposed method typically converges to above 90% recall with each point comparing only to several percent of the whole dataset on average.

tmylk:
@mkusner Should we expect RWMD to be much faster than WMD?
 It gives very little speed up in these tests, maybe because it is numpy vs C implementation of EMD. If we can't improve RWMD runtime then suggest not using it at all.

RWMD code here is almost the same as in [matlab RWMD code](https://github.com/mkusner/wmd/blob/master/compute_rwmd.m), just without a square root.

gensim/models/word2vec.py
```diff
@@ -1185,6 +1195,168 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None):
...
+            for i, token in dictionary.items():
+                X[:, i] = self[token]
+
+            dist = sqrt(np_sum((X.dot(d1) - X.dot(d2))**2))
+            return dist
+        if RWMD:
+            # Compute RWMD.
+            with warnings.catch_warnings():
+                # Ignore Numpy warning: "All-NaN axis encountered".
+                warnings.filterwarnings('ignore', r'All-NaN axis encountered')
```
tmylk: how much time does this take?

mkusner:
Hi all! Thanks a lot for spending so much careful time on this, I just
looked through the thread!

It depends a bit on the dataset how much RWMD speeds up WMD; we saw
speedups between 1.1x and 16x (in Figure 7 of the paper:
http://mkusner.github.io/publications/WMD.pdf) if EMD is in C and RWMD is
in numpy this could certainly shrink the speedup. RWMD is even more
effective at speeding up document distances when used to prune documents
for approximate WMD computation (in the prefetch and prune algorithm at the
end of Section 4 in the paper). I'd be happy to give more details about
this if this is something that people would want included.

In any case, thanks again!

gensim/similarities/docsim.py
```diff
@@ -569,6 +569,145 @@ def __str__(self):
...
+        self.chunksize = chunksize
+
+        # Normalize embeddings to sum to 1.
+        self.w2v_model.init_sims(replace=True)
+
+        # Normalization is not possible, as corpus is a list (of lists) of strings.
+        self.normalize = False
+
+        # index is simply an array from 0 to size of corpus.
+        self.index = numpy.array(range(len(corpus)))
```
tmylk: Where is this variable used later?

olavurmortensen: In `gensim.interfaces.SimilarityABC`, which `WmdSimilarity` extends. The Similarity base class uses `index` to sort the documents.

menshikh-iv:
Ping @olavurmortensen, what status of this PR?

olavurfargen:
I believe part of this PR was merged in #659 (just WMD), and that that optimizations (RWMD and WCD) were taken up by @RishabGoel in #800. 

I'm no longer working on this PR and haven't been for over a year. So the PR should be closed.

olavurmortensen:
Just want to note that I used the wrong account in my previous comment (olavurfargen), but it was me.

PR #536, opened 2015-11-19
Doc2Vec batching multiple docs to cython routines

gojomo:
Another minimalist approach to #450. Currently only does Doc2Vec – in fact unimplemented methods might make word2vec broken in its current state.


tmylk:
Implemented in #535 instead

Issue #534, opened 2015-11-19
match word2vec.c defaults (& option names? & command-line switches?) more closely


gojomo:
Per thread <https://groups.google.com/d/msg/gensim/ggCHGncd5rU/Z_pQDD69AAAJ>, we may want to reduce confusion for people transitioning to/from the original word2vec.c by matching its defaults exactly, or even further changing our option names to match theirs in the couple of cases where they disagree, or even giving word2vec.py a `main()` that allows it to be invoked exactly like `word2vec` to trigger the same training/saving.

piskvorky:
Thanks!

Btw `word2vec.py` already has a non-public `main`, used for testing. We can definitely replace that with something more flexible and robust.

davechallis:
I'd generally argue in favour of keeping gensim APIs more consistent with each other, than with external pieces of software, mostly because there's no guarantee about the stability of parameters/defaults of software you don't control (though I admit I don't know much about word2vec.c, i.e. whether it's updated often, whether it's stable, etc.).

I'd also argue than when using a module, internal consistency is also very important. E.g. if gensim uses 'workers' as a parameter in most of its classes and 'threads' in word2vec (to match word2vec.c), that could cause some confusion to those familiar with gensim, but new to word2vec.

Amongst many people I know that use gensim's word2vec, most haven't used the original C implementation, having gone straight to gensim/python for ease of use (though this is just anecdotal, I can't speak for wider usage of course :) ).

An easy starting point might just be to add some additional documentation, e.g. "If you want to replicate the original word2vec.c settings, use gensim parameters X."

Another alternative could be to add something like a `new_with_word2vec_c_settings` classmethod returning a Word2Vec instance using whatever the current word2vec.c settings.

I guess one thing to consider might be whether to try and optimise the experience of a) new or existing gensim users or b) those trying to replicate/compare directly with word2vec.c.

Hugo-W:
They are not exclusive ways that to optimize experience of new users and replicate word2vec.c results.
You would also optimize experience of new Gensim users by having consistent parameters. Obviously it is better to share names across the Gensim's API, but the defaults values are those you try at first.
And honestly, I found Gensim's word2vec after word2vec.c and I think that it is generalizable to must non-gensim users. As the classic way leading to it is either after reading Mikolov's paper which explicitly refers to its available code, or by googling "word2vec" and end up on the first link being "Google's code for Word2Vec"...
Then a day later you want a python version of it :)

[...]

Issue #531, opened 2015-11-17
intermittent failure: Test CBOW w/ hierarchical softmax - 


tmylk:
Need to restart Travis and Appveyor builds a couple of times in order to get this test set to pass. Is there a way to make it more robust?

-------------------- >> end captured logging << ---------------------
 
======================================================================
FAIL: Test CBOW w/ hierarchical softmax
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\gensim\test\test_word2vec.py", line 246, in test_cbow_hs
    self.model_sanity(model)
  File "C:\Python27\lib\site-packages\gensim\test\test_word2vec.py", line 226, in model_sanity
    self.assertLess(t_rank, 50)
AssertionError: 64 not less than 50
-------------------- >> begin captured logging << --------------------
gensim.models.word2vec: INFO: collecting all words and their counts

gojomo:
An odd thing here is that the tolerances were picked such that I saw no failures in 100-200 local test runs... but sometimes (and almost in a bursty fashion?!?!) the auto-testing systems seem far more prone to outlier results. 

Especially with regard to a Windows build that requires many tries to pass, compared to linux builds that "usually" passes, it might be indicative of a true training degradation on that platform, requiring a fix other than testing adjustments.

Have you noticed if it's always this exact test (`test_cbow_hs`) and python (2.7 as opposed to 2.6/3.x)? Or, are you just seeing at-least-one of the many similar tests failing across multiple configs?

More generally I see the problem as:

We're attempting quick tests on a necessarily-thin amount of bundled data on algorithms that have some inherent variance. 

We may need to seed the tests to achieve perfect run-for-run reproducibility. At least then, many true bugs will break the test that previously passed predictably. 

Simply expanding the tolerances on a still-random process can make the failures less-frequent – but might not eliminate them entirely. Then on the rare occasions test-failures still happen, we'd be unsure if the code really changed in some tangible way, or it was just a very-unlucky outlier run. Making the tolerances so generous that we "never" (practically) see a failure could also hide many other kinds of calculation-degradation bugs we'd usually want tests to catch.

piskvorky:
I'm -1 on seeding tests exactly. If the inherent variance is such that it requires a "test range of values" to become effectively meaningless, then we don't need the test in the first place (it's meaningless for the same reason).

So let's try to come up with other solutions. Would more data help? Different parameters (iterations)?

tmylk:
[...]

PR #538, opened 2015-11-23
cbow_mean default changed from 0 to 1.

akutuzov:
Taking average from input vectors in CBOW mode gives better results than their sum (see https://docs.google.com/spreadsheets/d/1dgr513AePh4EjCUQxQyeT9i6Xnig3SOtjLJwiIVmuu4 for experiments). It is also the default setting in Mikolov's word2vec. Therefore, cbow_mean=1 should be default in Gensim as well.


piskvorky:
It's a change in welcome direction, but I'd prefer this to be a pull request that fixes the compatibility fully, across variables, as discussed in #534 .

Piece-meal updates will only make the switch more painful and confusing.

akutuzov:
OK. I will try to return to this in a few days.  Should all Gensim's word2vec variables be aligned with Mikolov's tool, or only in the main()?

piskvorky:
Thanks @akutuzov !

I think we should keep the current names in the code, just change their default values. But the `main()` CLI names can match the C tool, sure, why not.

tmylk:
Hey @akutuzov do you have any updates? Planning a release and would be good to include this.

akutuzov:
Unfortunately not yet.
What is the planned date for the release? I could try to make it in time.

tmylk:
17 Jan is the tentative date

akutuzov:
I see. Then I hope to finish this by 15 Jan.

akutuzov:
I changed the default values of all hyperparameters to match those of Mikolov's wordvec. The only exception is default hierarchical softmax (word2vec has default negative sampling of 5 samples). If I change Gensim defaults to negative sampling, it immediately breaks a number of tests: particularly, word2vec training (which seems to rely on syn1 matrix) and scoring (which is not implemented for negative sampling at all). Thus, I left default hierarchical softmax as is.

I also changed the defaults in (main) and will implement CLI arguments processing mimicking those of word2vec in the next couple of days.

Not quite sure why checks are failed on this PR, and what conflicts it mentions. The code passes all the tests here.

akutuzov:
@tmylk, I am finished with this. Gensim word2vec defaults are now consistent with Mikolov's C tool (except for hierarchical softmax, see above), and the (main) now accepts the same command line arguments as the C tool. 
I had to change one value in the unit test for word2vec (model_sanity function, line 238). It tested for the word 'terrorism' to be in the first 50 synonyms of the word 'war', and this was not always the case, so I changed 50 to 60. The reason is that the default vector size is now 100 (as in Mikolov's word2vec), not 200. Thus, the resulting default models are slightly worse (but smaller).

gojomo:
Thanks for the contribution!

FYI, the default size is 100 before and after, so that's not specifically the cause for the test failure you saw. Rather, that's been an ornery test, more likely to fail on some systems (including the CI servers) than others. See #531 and potential improvement in PR #581. Note that merely loosening the tolerance only helps a little, leaving the potential for outliers over the new threshold – whereas I think the changes is #581 will more effectively tighten the range of results. 

Other comments coming in line notes.

gensim/models/word2vec.py
```diff
@@ -342,8 +342,8 @@ class Word2Vec(utils.SaveLoad):
     """
     def __init__(
             self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
-            max_vocab_size=None, sample=0, seed=1, workers=1, min_alpha=0.0001,
-            sg=1, hs=1, negative=0, cbow_mean=0, hashfxn=hash, iter=1, null_word=0,
+            max_vocab_size=None, sample=1e-3, seed=1, workers=12, min_alpha=0.0001,
```
gojomo: Since our multithreading doesn't parallelize as effectively as the C code (and in fact hits a point of diminishing returns even before reaching the CPU-core count), the `workers` parameter is one exception where we should pick a different value. I suggest 3 as a value likely to help without causing issues.

gensim/models/word2vec.py
```diff
@@ -357,8 +357,8 @@ def __init__(
         If you don't supply `sentences`, the model is left uninitialized -- use if
         you plan to initialize it in some other way.
 
-        `sg` defines the training algorithm. By default (`sg=1`), skip-gram is used.
-        Otherwise, `cbow` is employed.
+        `sg` defines the training algorithm. By default (`sg=0`), CBOW is used.
+        Otherwise (`sg=1`), SkipGram is employed.
```
gojomo: To match style previous and elsewhere, 'skip-gram' > 'SkipGram'.

gensim/models/word2vec.py
```diff
@@ -342,8 +342,8 @@ class Word2Vec(utils.SaveLoad):
     """
     def __init__(
             self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
-            max_vocab_size=None, sample=0, seed=1, workers=1, min_alpha=0.0001,
-            sg=1, hs=1, negative=0, cbow_mean=0, hashfxn=hash, iter=1, null_word=0,
```
gojomo: Word2Vec's switch from `hs=1, negative=0` to `hs=0, negative=5` is a fairly meaningful and helpful one as far as I can tell, motivated by improved performance, so I think it's an important place to match them. Any particular feature that now needs 'hs=1' to be specified (such as scoring) should explicitly do that, rather than forcing a less-preferred default on the larger userbase.

gensim/models/word2vec.py
```diff
@@ -376,16 +376,17 @@ def __init__(
         need about 1GB of RAM. Set to `None` for no limit (default).
 
         `sample` = threshold for configuring which higher-frequency words are randomly downsampled;
-            default is 0 (off), useful value is 1e-5.
+            default is 1e-3, useful value is 1e-5, 0 stands for off.
 
         `workers` = use this many worker threads to train the model (=faster training with multicore machines).
 
-        `hs` = if 1 (default), hierarchical sampling will be used for model training (else set to 0).
+        `hs` = if 1 (default), hierarchical sampling will be used for model training (if set to 0, negative sampling will be used).
```
gojomo: Technically `hs=0` is not enough to trigger the use of negative-sampling: `negative` must be set non-zero as well. (Also: line-length.)

gensim/models/word2vec.py
```diff
@@ -376,16 +376,17 @@ def __init__(
         need about 1GB of RAM. Set to `None` for no limit (default).
 
         `sample` = threshold for configuring which higher-frequency words are randomly downsampled;
-            default is 0 (off), useful value is 1e-5.
+            default is 1e-3, useful value is 1e-5, 0 stands for off.
```
gojomo: Should match word2vec.c recommendation – 'useful range is 0 to 1e-5` – rather than imply that the default is somehow not as useful as some other value.

akutuzov:
@gojomo , I fixed the code as per your comments. Though, I am not sure what to do with the non-deterministic tests (like whether `terrorism` is the first 50 neighbours of `war`). Right now I see that this test fails on two of Travis builds, but successfully passes on other three. Seems very chaotic. Should I just leave it as it is?

gojomo:
Thanks! 

For any of the tests that previously assumed HS behavior, the best fix would be to specify `hs=1, negative=0` in the preceding Word2Vec instantiation – so that the same functions under HS are actually tested, rather than skipped because of the new defaults. 

With regards to the specifically ornery test (`test_cbow_hs`) that was intermittently failing before these changes, you just want to make sure you haven't made things worse. To that end I suggest:

(1) returning the threshold to 50; then at least in cases when a failure happens, you'll get an indication of how-close was the miss (the actual position) from that earlier line. 

(2) bring another fix (from #581 but now also in the main gensim `develop` branch) into your branch. It should definitely help with the `test_cbow_hs` test, and maybe others. (I think a `git fetch upstream`, `git merge upstream/develop`, in your local project directory, should do the trick.) 

(3) *If* there are still any problems after this step, then (a) check to make sure any of the other default changes didn't make the tests worse – the `sample` parameter is the most tangible one I'd consider; but also (b) it'd be ok to throw your hands up & say, "I doubt these test-failures are related to my changes", and I'll take a closer look at any tests needing tuning or disabling.

akutuzov:
@gojomo , I returned the threshold to 50 and specified using of HS in all tests which depended on it. Now all the tests pass successfully (at last).

gojomo:
Looks good to me! Last step would be a bullet-item for the CHANGELOG.txt, describing the change. (I might even add extra emphasis to this item, by listing it first or adding some all-caps prefix like "DEFAULTS CHANGED", since it may require extra user attention to change any code that depended on the old defaults...)

akutuzov:
Done with Changelog. Though, one of five builds once again failed with CBOW+negative sampling test.

gojomo:
I ran some bulk tests and it seems the move from `sample=0` to the new default `sample=1e-3` is enough to turn that `test_cbow_neg` test from something that doesn't fail (in 1500+ runs) into one that fails intermittently. So, the test has caught a real change in code behavior – but not really anything to be worried about, because we know these tests (on tiny datasets and compressed timeframes) are unavoidably fragile. Explicitly testing the old default – adding `sample=0` to the `test_cbow_neg` Word2Vec instantiation – restored the test reliability for me. So that's my recommendation.

akutuzov:
OK, I've changed the downsampling parameter for this test. Now it passes fine.

gensim/test/test_word2vec.py
```diff
@@ -154,7 +154,7 @@ def testVocab(self):
         total_words = sum(len(sentence) for sentence in corpus)
 
         # try vocab building explicitly, using all words
-        model = word2vec.Word2Vec(min_count=1)
+        model = word2vec.Word2Vec(min_count=1, hs=1, negative=0)
```
tmylk: Could we please add a new test for the new defaults that are introduced in this PR? i.e. where model = word2vec.Word2Vec(min_count=1)

tmylk:
Waiting for new tests tofor the new defaults to be added before this can be merged.

akutuzov:
What can these tests possibly be?

tmylk:
The checks could be the same as for hs=1, negative=0 case, but with a different set of expected results.

gojomo:
In my opinion new tests aren't necessary. The intent of the prior set was adequate coverage of all supported options (not just the arbitrarily-chosen prior defaults), and in some cases the literal defaults aren't the best test within the time/data constraints of quick unit tests.

tmylk:
@gojomo Please correct my understanding: the goal of this PR is to make it easy for new users to replicate word2vec.c results in gensim.
An appropriate test for that is to run gensim defaults and check the output against word2vec.c either via wrapper or via copy-pasted results.

gojomo:
It' to reduce confusion when people expect the same defaults. Let's say hypothetically that we use the same defaults, *but* get markedly different quality results. That'd be an issue, but it's not *this* issue. This issue is: let's use the same defaults.

If someone wants to run a deep comparison of the quality of the results from gensim vs the original word2vec.c, more power to them. But that shouldn't be a blocking condition for this issue, which just switches gensim's default from one previously-supported mode to another previously-supported mode. 

(Again I'd emphasize: there's no new functionality here. SG-or-CBOW, HS-or-negative-sampling, downsampling-or-not were all already supported (and largely covered by unit tests).  So just toggling the defaults shouldn't impose a new quality bar to hurdle.

piskvorky:
This looks good to me. Two changes needed before merge:

- [ ] squash the multitude of git commits
- [ ] factor out the CLI script from `__main__` into `gensim/scripts` (the `__main__` was just undocumented helper code, now we want to be more formal about this functionality and promote it)

A unit test that the new script works would be nice too -- at least making sure it's runnable (python 2.6, python 3.3 etc.)

CHANGELOG.txt
```diff
@@ -3,6 +3,13 @@ Changes
 
 0.12.4, XX/XX/XXXX
 
+* Word2vec default values of hyperparameters now are identical to those of original C tool, generally improving performance (Andrey Kutuzov, #538)
+ - By default use CBOW with negative sampling and the mean of context vectors, instead of skip-gram with hierarchical softmax.
```
piskvorky: Unnecessarily verbose -- a github link to the diff between "old vs. new" defaults is enough IMO.

gensim/models/word2vec.py
```diff
@@ -1600,35 +1602,65 @@ def __iter__(self):
...
-# Example: ./word2vec.py ~/workspace/word2vec/text8 ~/workspace/word2vec/questions-words.txt ./text8
+# Example: ./word2vec.py -train data.txt -output vec.txt -size 200 -window 5 -sample 1e-4 -negative 5 -hs 0 -binary 0 -cbow 1 -iter 3
 if __name__ == "__main__":
+    import argparse
     logging.basicConfig(
         format='%(asctime)s : %(threadName)s : %(levelname)s : %(message)s',
         level=logging.INFO)
-    logger.info("running %s", " ".join(sys.argv))
-    logger.info("using optimization %s", FAST_VERSION)
+    logging.info("running %s", " ".join(sys.argv))
```
piskvorky: Why this change? Better use `logger`.

tmylk:
Standalone script manually tested successfully with minor code changes as in # 593.
Automated tests for the standalone script are still work in progress in #593 .

Plan to merge in #593 tonight - Andrey, @akutuzov, if you could review it would be great.

akutuzov:
@tmylk, seems OK for me.

tmylk:
Merged in # 539 - needed some changes to run.
Tests are still pending

piskvorky:
Thanks a lot @akutuzov ! Great work, much appreciated :)

PR #539, opened 2015-11-23
Provide INPUT_MIN_LENGTH in docstring

timClicks:
Give users quick access to the minimum number of documents needed 
by summarize_corpus. Use string formatting at runtime in case the value
of INPUT_MIN_LENGTH is changed.


piskvorky:
That style of string formatting won't work in Python2.6.

Otherwise I think this can be useful, I'm +1 on merging. Thanks!

timClicks:
@piskvorky Done. Sorry about the mix up, didn't realise that `gensim` also supports Python 2.6 (great!).

piskvorky:
Great, thanks :)

Issue #499, opened 2015-10-24
Don't allow saving incomplete word2vec objects


piskvorky:
From a [mailing list report](https://groups.google.com/forum/#!topic/gensim/1z0S8TBnkUA):

Currently, when a word2vec object is loaded from the C tool via `load_word2vec_format`, it can be saved with `model.save()`. But loading such saved model back with `load()` results in an error. Vital pieces of the object are missing, the C format is incomplete.

This is undesirable and unintuitive behaviour.

Solution:

* either refuse to save: throw an explicit exception that necessary structures are missing
* or (better) make those structures not necessary. If the object serves its purpose after `load_word2vec_format`, without those structures, it might as well serve its purpose after `load`, again without those structures.

ziky90:
Issue was fixed in #540 and #543 so I guess that we can close this.

PR #540, opened 2015-11-23
fixed #449 removed loading of structures that are not necessary

### LINKED ISSUE #449: Word2vec to use GPU
Add an option (likely subclassing `Word2vec`) to train word2vec model using GPU.

ziky90:
This PR proposes fix for the issue #499. Thanks to this modification it will be now possible to load even object that was pickled without at least for inference unnecessary `model.vocab.values[n].sample_probability` values. 

These kind of models can be obtained simply by loading the model via `load_word2vec_format("model.txt")`, `save("model.pkl")` and `load("model.pkl")` sequence of commands.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1466,7 +1466,7 @@ class Word2Vec(utils.SaveLoad):
         for v in model.vocab.values():
             if hasattr(v, 'sample_int'):
                 break  # already 0.12.0+ style int probabilities
-            else:
+            elif hasattr(v, 'sample_probability'):
                 v.sample_int = int(round(v.sample_probability * 2**32))
                 del v.sample_probability
         if not hasattr(model, 'syn0_lockf') and hasattr(model, 'syn0'):


gensim/models/word2vec.py
```diff
@@ -1466,7 +1466,7 @@ def load(cls, *args, **kwargs):
         for v in model.vocab.values():
             if hasattr(v, 'sample_int'):
                 break  # already 0.12.0+ style int probabilities
-            else:
+            elif hasattr(v, 'sample_probability'):
                 v.sample_int = int(round(v.sample_probability * 2**32))
```
ziky90: I did not find the `sample_probability` anywhere else, so I am not sure whether this is needed at all?

gojomo: Earlier versions used `sample_probability` (a float) instead of `sample_int` for this purpose; this allows models saved from those versions to load properly.

gojomo:
Curious, why would someone want to use the `load_word2vec_format("model.txt")`, `save("model.pkl")` and `load("model.pkl")` sequence? The resulting model will still just be the same as that available from `load_word2vec_format("model.txt")` alone.

piskvorky:
Faster load (mmap), more consistent save/load API. I can see how such conversion can be useful.

I'm merging this, thanks @ziky90 !

I suspect there are other things we could do to make loading older / incomplete versions of word2vec easier. We receive confused support requests on the mailing list around this regularly. Feel free to open a PR if you find something else.

piskvorky:
Maybe I merged too soon. Now I'm thinking an explicit unit test for "`load_w2v_format` => `model sanity check` => `save` => `load` => `model sanity check`" would have been in order here, to make sure #499 is really fixed.

PR #545, opened 2015-11-24
load_word2vec_format(): new parameter to skip init_sims() call

svenkreiss:
In certain use cases (custom doc2vec-type computations) only unnormalized vectors are used. The `init_sims()` call at the end of `load_word2vec_format` takes a lot of memory (even with `norm_only=True`) and is unnecessary in this scenario. This PR allows to skip the call which improves performance.


piskvorky:
Sounds useful and clean, +1.

Let's wait for @gojomo review & then merge.

piskvorky:
Also @svenkreiss , can you commit a brief description of this change in `CHANGELOG.txt`?

svenkreiss:
@piskvorky thanks for comments. Done.

piskvorky:
Perfect, thanks a lot @svenkreiss !

gojomo:
Looks fine, but another more radical simplification that reduces parameters & codepaths would also be worth considering: just don't do any automatic norming in the `load_word2vec_format()`. (That is, remove the `init_sims()` call rather than make it switchable.)

Then the (`syn0`) result of a load is really just a load, not a load-and-do-other stuff. This would also be roughly consistently with the `syn0` state after native gensim training: you have the raw vectors, what you do with them next is up to your explicit further steps. 

If the user starts making similarity calls, `syn0norm` would be automatically backfilled, as with natively-trained vectors.... but someone who doesn't need that would just choose not to trigger it. 

If they instead want to convert to a compact, normed-only model, they'd call `init_sims()` themself – just as if you'd trained the vectors with gensim.

piskvorky:
Oh yes, that makes sense too. I actually like @gojomo 's option better -- simpler is better.

svenkreiss:
I also agree with @gojomo. I am traveling till Sunday, but can prepare a PR that removes the init_sims and norm_only parameters. 

This would be api backwards incompatible and change the default behavior of this function.

piskvorky:
Yes, we'll need a prominent warning in CHANGELOG :)

Thanks again @svenkreiss , you're really helpful!

PR #560, opened 2015-12-09
fix string-doctags missing from most_similar results

gojomo:
bug was introduced by changes in #491 / 39b63e0


piskvorky:
Can you add some sort of test, so we catch similar problems in the future?

tmylk:
@gojomo Tests and CHANGELOG required before it can be merged.

gojomo:
Test & CHANGELOG note added – ready for merge.

Issue #537, opened 2015-11-20
FAIL: testTransform LdaMulticore on Windows 80% of the time 


tmylk:
Could we add more passess to allow LDA to converge here?

On Windows this test passed only once out of 5 re-runs in Python 2.7 32bit. It failed every time on Python 2.7 64 bit, Python 3.5 32 bit and 64 bit.

This test was passing last month.

It is blocking Windows wheel generation for gensim.

FAIL: testTransform (gensim.test.test_ldamodel.TestLdaMulticore)

https://ci.appveyor.com/project/tmylk/gensim-ewnfe

gojomo:
I see the word2vec CBOW/HS test also failed at the same time. (Just barely: expected result was 53rd against a tolerance of top-50.)

If the same tests (usually) pass on other platforms, or (usually) passed on Windows last month... adding passes (or loosening targets) may be hiding some other true regression with the Windows math.

(I notice that both of these tests are multi-threaded; maybe something about the math routines is more likely to have 'older' values clobber 'newer' values, in shared memory, on Windows... thus slowing convergence.)

tmylk:
@piskvorky Instead of commenting the failing Windows tests, maybe it is better to skip them on Windows but still run in Linux? (based on an environment variable set up in appveyor.yaml )

gojomo:
I strongly suspect that it is something different about thread-scheduling that's causing this (and the occasional Word2Vec/Doc2Vec failures), as that's the thing most likely to differ across platforms, and multithreading is definitely implicated in the tests that are failing. It might be further aggravated by the small size of these toy test examples, so data isn't being cut into as many chunks (with synchronized model updates per chunk) as in more realistic examples. 

I'm not very familiar with the LDA options, but perhaps enabling smaller chunks – via instantiation parameters like `batch=True, chunksize=50` in LdaMulticoreTest – would lead to results that are more stable across varied thread-scheduling.

menshikh-iv:
Unrelevant, because it works OK now.

PR #580, opened 2016-01-11
try smaller chunks

gojomo:
for #537. (PR to trigger AppVeyor Windows testing)


cscorley:
Since the test corpus is still smaller than the chunksize (9 documents in the corpus), I don't think this could change the outcome vs the default chunksize.

gojomo:
`chunksize=1` didn't help; 7 of 8 AppVeyor runs failed: 2 froze at `Test word2vec parallel training` and 5 failed on the same `testTransform (gensim.test.test_ldamodel.TestLdaMulticore)`.

Issue #571, opened 2015-12-31
Race Condition in Muli-threaded word2vec


elmarhaussmann:
The current development branch uses batch-based multi-threading. Some parts of the code are not synchronized properly and susceptible to race conditions. In particular, workers execute `self.jobs_left -= 1` and the job producer: `self.jobs_left += 1`. This is not an atomic operation. Depending on timing and release of the GIL this leads to a wrong number of jobs left. Since the termination condition for training checks whether jobs are left, training often never finishes properly.

I verified this by outputting the number of jobs left at the end of training. The longer the training before, the higher the chances of a corrupted value and training not terminating.

Access should be properly synchronized (via locking) or another thread-safe data structure could be used instead.

piskvorky:
I don't see how this could be non-atomic. Could you post (gist) your verification tests?

piskvorky:
OK, it looks like in-place increment is not atomic in Python, so this may indeed lead to a race condition.

Still curious to see your tests though. I've never seen the "training often never finishes" in runs or tests, so I'm curious what numbers you got.

elmar-haussmann:
As far as I know only single opcodes are guaranteed to be thread-safe and "+=" consists of reading, adding and writing (loosely speaking - didn't verify the exact opcodes) the value of the variable. I can whip up some code showing this tomorrow.

elmar-haussmann:
I ran some computations on huge data (>1billion) words and 12 threads - the process never finished (2 of 2 times) and was hanging at the end. I also experienced it on smaller input (~10 million words), but it's less likely to occur (maybe 1 in 4 times), which makes sense. 
I cannot reproduce this for certain. My current test/workaround consists of changing the termination condition like in this snippet: https://gist.github.com/elmar-haussmann/d7670cac549cafe0862c. I can clearly see in my output that there are jobs left, but nothing is being computed anymore.

piskvorky:
Sure, thanks for reporting! What version of gensim was that on? Develop or some release?

elmar-haussmann:
I used the current develop branch (from two days ago). Batch based multi-threading scales a lot better in my experiments! (not sure if it's in a release version already)

phdowling:
Acquiring a lock before the increment and releasing it after would probably be an easy fix, or do you see any problems there?

```
from multiprocessing import Lock
lock = Lock()

...
with lock:
    self.jobs_left -= 1
```

piskvorky:
[...]

PR #581, opened 2016-01-12
threadsafe job counts; configurable batch_size

gojomo:
Fix for #571. Uses local vars in each thread for counts. Adds class/init `batch_words` parameter to set a non-default value.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,8 @@ Changes
 * Word2vec allows non-strict unicode error handling (ignore or replace) (Gordon Mohr, #466)
 * On-demand loading of the `pattern` library in utils.lemmatize (Jan Zikes, #461)
   - `utils.HAS_PATTERN` flag moved to `utils.has_pattern()`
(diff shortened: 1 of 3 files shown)


gojomo:
@tmylk @piskvorky – any concerns about merging this into develop? I believe it will improve (and perhaps fix entirely) two causes of intermittent CI test failures (and thus save anyone else working on develop from thinking their changes may have caused a failure, as in #538).

piskvorky:
It looks quite elegant, I'm not sure why I messed around with shared state when a simple counter was enough.

Maybe I'm missing something here (reasoning about parallelism is tricky) but if all tests and sanity checks pass, it's LGTM from me.

piskvorky:
@tmylk I suspect a similar issue could be behind the failing tests in `LdaMulticore` (which uses processes instead of threads, but otherwise similar logic flow, AFAIR).

gojomo:
Added CHANGELOG & merged to make it easier for other in-progress/future branches to build on this.

Issue #520, opened 2015-11-10
Strange result when retrieving a not existing word embedding for a document


claudiogreco:
Hello,

after creating a doc2vec model using the code reported in the tutorial with **python 3.5** and **gensim 0.12.3**, I have wrongly tried to retrieve a missing document, but I have received a strange numpy.ndarray as a result whose shape is (1, 1235, 300), where 1235 is the number of documents and 300 is the size of embeddings. Why does it happen? Shouldn't be raised an exception? Eventually, how can we check if a document is missing or not?

Thank you in advance,
Claudio

gojomo:
This is caused by an unfortunate interaction between out internal method which converts a string-tag to an int-index (which returns None for not-present) and numpy array indexing, which returns that nested 1xCOUNTxSIZE result when passed None. 

Was a KeyError what you'd expected?

Until that's fixed, you can check whether a tag is in the trained set with `key in model.docvecs`.

claudiogreco:
Ok, I'll check whether a tag is in the trained set using your suggestion for now. Thank you for your help.

tmylk:
@gojomo should this be closed as a workaround exists?

gojomo:
@tmylk no, there should really be a KeyError (rather than a giant nested array) when the requested key isn't present. I'll prep a fix before next week.

gojomo:
Fixed by #582.

PR #582, opened 2016-01-12
fix for #520: raise KeyError when no matching doctag

gojomo:


tmylk:
Thanks, Gordon. Please add Changelog

tmylk:
Thanks!

PR #596, opened 2016-01-29
Separate Mallet serialization from import-file

piskvorky:
... so that `LdaMallet` subclasses can serialize input data in a more flexible way.


piskvorky:
@tmylk review & merge. Ready from my side. We definitely want this in the next release.

PR #600, opened 2016-02-02
wikicorpus: Refactor MediaWiki IGNORED_NAMESPACES

mrshu:
* Refactor ignore_namespaces that previously used just a split over a
  line of text into a list.

* Add a few more namespaces into the IGNORED_NAMESPACES list.

Signed-off-by: mr.Shu <mr@shu.io>

### DIFF
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -54,6 +54,12 @@ RE_P14 = re.compile('\[\[Category:[^][]*\]\]', re.UNICODE) # categories
 # Remove File and Image template
 RE_P15 = re.compile('\[\[([fF]ile:|[iI]mage)[^]]*(\]\])', re.UNICODE)
 
+# MediaWiki namespaces (https://www.mediawiki.org/wiki/Manual:Namespace) that
+# ought to be ignored
+IGNORED_NAMESPACES = ['Wikipedia', 'Category', 'File', 'Portal', 'Template',
+                      'MediaWiki', 'User', 'Help', 'Book', 'Draft',
+                      'WikiProject', 'Special', 'Talk']
+
 
 def filter_wiki(raw):
     """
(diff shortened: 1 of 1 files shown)


tmylk:
Thanks @mrshu !

It would be ideal to have tests for this change, specifically for 'python -m gensim.scripts.make_wikicorpus ~/gensim/results/small_test_sample_enwiki-latest-pages-articles.xml.bz2 ~/gensim/results/wiki_en'.

The test needs a ~10kb wikipedia sample as an input and should be similar to https://github.com/hanabi1224/gensim/commit/f118d465e4107fad09c21b1802e2804ff1c9f6ea#diff-94580fda7e6624c633df38580d4e3288R410

mrshu:
Thanks @tmylk, I am pretty sure it would be ideal but I do not clearly see how I would exactly test that, given that this PR adds a few more [MediaWiki namespaces](https://www.mediawiki.org/wiki/Manual:Namespace) that should be skipped when processing the Wikipedia dumps. I am not sure if I can get so many pages that should be skipped in one ~10kb Wikipedia sample (and there should probably also be some valid page).

Moreover I am not sure what to do with the output (files generated in `~/gensim/results/wiki_en`). I believe they should be removed in order to keep the testing directories clean. I unfortunately cannot see that in the example you linked to.

Which brings me to the last question I have: How should I verify that the generated files (in `~/gensim/results/wiki_en` given your example) are actually correct? Should I just check the output as in the example you posted above?

Thanks a ton!

mrshu:
Hey @tmylk, any progress on this? I would hate to implement some sort of tests without you being OK with what they are actually testing, let alone their structure.

I believe this change can be beneficial as it both cleans up the code a bit and also improves the internal wikipedia articles skipping. My sole reason for creating this PR was the last sentence of this footnote: http://radimrehurek.com/gensim/wiki.html#id7

    By the way, improvements to the Wiki markup parsing code are welcome :-)

If you find that sentence still holds true, please do let me know what are you answers the the questions I posted above so that we can get this change merged in.

Thanks again!

tmylk:
Hi @mrshu , apologies for late reply - back to online world now. Thank you for your contribution and for improving gensim!

File can be up to 2Mb if it works better, just not the whole of wikipedia. :)

The continuous integration envrionments Travis and Appveyor recreate the test directories from scratch so no need to delete the result.

Just checking the exit code of the script is a good test.

tmylk:
@mrshu Merging in, but raising https://github.com/piskvorky/gensim/issues/650 as we need tests for these popular scripts.

PR #605, opened 2016-02-10
Fix RP model loading for large Fortran-order arrays

piskvorky:
...seems like a left-over bug from when `SaveLoad` started storing large arrays separately.

RpModel overloads `__setstate__`, which expected all arrays to be loaded at once. The reason for this overload was to work around a bug in NumPy, which was causing segfaults on using loaded Fortran-order arrays. The fact that we're now loading some object attributes later broke this `__setstate__` functionality, causing an exception [reported by users on the mailing list](https://groups.google.com/forum/#!topic/gensim/BE3qDSNJs54).

The fix in this PR delays applying the bug workaround for when the loaded array is first *used* (as opposed to *loaded*).

Might also be worth checking if the NumPy bug (segfault) is still there in the first place, so we could remove the workaround altogether. In fact, NumPy has seen a lot of changes since 2010, so maybe we don't need to keep arrays in Fortran-order at all. Or maybe we still do -- check!

### DIFF
--- a/gensim/models/rpmodel.py
+++ b/gensim/models/rpmodel.py
@@ -6,10 +6,8 @@
 
 
 import logging
-import itertools
 
 import numpy
-import scipy
 
 from gensim import interfaces, matutils, utils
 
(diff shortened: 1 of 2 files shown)


mateuss-ntnu:
I've tested it now, and it works fine :+1:

piskvorky:
Thanks @mateuss-ntnu !

@tmylk , please review & merge. And make investigating these (unrelated) AppVeyor fails a priority... it's really annoying.

PR #607, opened 2016-02-10
Control whether to use lowercase for computing word2vec accuracy.

alantian:
Add options to control whether to use lowercase when computing the accuracy of word2vec model.


piskvorky:
@tmylk the merge caused Travis to fail -- the project status is now "failing".

Probably unrelated to the change here; can you check what's going on?

tmylk:
This re-build was caused by CHANGELOG commit https://github.com/piskvorky/gensim/commit/1be6dff55228ef484c177807cbab789e7f8563e9

The issue was with Python 3.3 which I see a lot recently. Always fixed on restart

PR #610, opened 2016-02-15
minor doc syntax

jrouquie:


cscorley:
I think that was the original intent: to show that either should work.  Maybe instead, for clarity, change it two sets: one with lists of tuples and another of tuples of lists?

jrouquie:
Thanks for the explanation.

Issue #527, opened 2015-11-13
Improving performance for function most_similar


anhldbk:
The current implementation of most_similar is NOT efficient. I've found that [Annoy](https://github.com/spotify/annoy) can be used to yield a better performance.

Here is my code:

```python
import logging
from gensim import matutils
from gensim.models.word2vec_inner import REAL
from numpy.core.multiarray import ndarray, array
from gensim.models import Doc2Vec
from six import integer_types
from six import string_types
from sklearn.externals import joblib
import numpy as np
from annoy import AnnoyIndex

FILE_DOC2VEC = 'dbow_5ns_100features_1mw_r20'
FEATURES_NUM = 100
ANNOY_TREE_SIZE = 300 # choose careful

dbow = Doc2Vec.load(FILE_DOC2VEC)
indexer_label_dict = dict()


def init_indexer(features_num, tree_size):
	"""
	Initialize the Annoy indexer
	"""
    annoy_indexer = AnnoyIndex(features_num)
    docvecs = dbow.docvecs
    docvecs.init_sims()
    size = len(docvecs.doctag_syn0norm)

    for i in range(size):
        annoy_indexer.add_item(i, docvecs.doctag_syn0norm[i])
        indexer_label_dict[i] = docvecs.index2doctag[i]

    annoy_indexer.build(tree_size)
    return annoy_indexer

def dev_most_similar(docvecs, positive=[], negative=[], topn=10, clip_start=0, clip_end=None):
    """
    Find the top-N most similar docvecs known from training. Positive docs contribute
    positively towards the similarity, negative docs negatively.

    This method computes cosine similarity between a simple mean of the projection
    weight vectors of the given docs. Docs may be specified as vectors, integer indexes
    of trained docvecs, or if the documents were originally presented with string tags,
    by the corresponding tags.

	('clip_start' and 'clip_end' do not play any role. I just keep it to maintain backward compatibility)
	The 'clip_start' and 'clip_end' allow limiting results to a particular contiguous
	range of the underlying doctag_syn0norm vectors. (This may be useful if the ordering
	there was chosen to be significant, such as more popular tag IDs in lower indexes.)
    """
    docvecs.init_sims()

    if isinstance(positive, string_types + integer_types) and not negative:
        # allow calls like most_similar('dog'), as a shorthand for most_similar(['dog'])
        positive = [positive]

    # add weights for each doc, if not already present; default to 1.0 for positive and -1.0 for negative docs
[...]

PR #548, opened 2015-11-25
Flexible API for nearest neighbour search

anhldbk:
This is an implementation for issue [#527](https://github.com/piskvorky/gensim/issues/527)

I've added neighbor_sim.py in directory gensim/models in which I defined NeighborIndexer classes.
```python
class NeighborIndexer(object):
    """
    Base class for k-NN libraries
    """

    def __init__(self, model=None, **kwargs):
        """
        Create a new indexer
        :param model:   Instance of Doc2Vec or Word2Vec
        :param kwargs:  Additional named parameters
        """
        if model is None:
            raise Exception("Invalid model parameter. Please provide model with an instance of Doc2Vec or Word2Vec")
        if type(model) is Doc2Vec:
            self._init_doc2vec_(model)
        elif type(model) is Word2Vec:
            self._init_word2vec_(model)
        else:
            raise Exception("Invalid model parameter. Please provide model with an instance of Doc2Vec or Word2Vec")
        self.start_indexing()

    def _init_doc2vec_(self, model):
        docvecs = model.docvecs
        docvecs.init_sims()
        size = len(docvecs.doctag_syn0norm)

        for i in range(size):
            self.add_item(docvecs.offset2doctag[i], docvecs.doctag_syn0norm[i])

    def _init_word2vec_(self, model):
        # raise Exception("Not supported at the moment")
        model.init_sims()
        size = len(model.syn0norm)

        for i in range(size):
            self.add_item(model.index2word[i], model.syn0norm[i])

    def get_item(self, label):
        """
        Get an item's vector by its label
        :param label: The label
        :return: The item's vector if have. Otherwise, returns None
        """
        pass

    def add_item(self, label, vector):
        """
        Add an item to index
        :param label: the label of the item (must be unique).
        :param vector: the item's vector.
        """
        pass

    def start_indexing(self):
        """
        Start the indexing operation. This method is supposed to run after all items are added
        It may take time to finish.
        """
        pass

    def get_nearest_items(self, vector, top_n=10):
        """
        Get nearest items for an item described by its vector
        :param	vector: The vector
        :param	top_n: Number of nearest items to get
        :return	A list of tuples (label, similarity) for the nearest ones
        """
        pass

    def save(self, file_name):
        """
        Dump internal data into disk
        :param file_name:   Output file path
        :return: Returns true on success. Otherwise, returns false
        """
        pass

    def load(self, file_name):
        """
        Load previously dumped info
        :param file_name:  Input file paht
[...]

PR #617, opened 2016-02-25
Flexible API for nearest neighbour search

anhldbk:
This is an implementation for issue [#527](https://github.com/piskvorky/gensim/issues/527)

I've added indexer.py in directory gensim/models in which I defined Indexer classes.

To make the integration process easier, I've modified most_similar() functions in doc2vec.py and word2vec.py by adding optional parameter *indexer*

For example: 
```python
# at line #409, doc2vec.py, the new function looks like this
def most_similar(self, positive=[], negative=[], topn=10, clip_start=0, clip_end=None, indexer=None):
```

To use the indexers, you may use the snippet below:
```python
from gensim.models.indexer import AnnoyIndexer
model = Doc2Vec.load(fname)

# create an indexer 
# features_num  = size of each document's vector
# You may need to optimize your *tree_size* parameter
indexer = AnnoyIndexer(model=model, features_num=100, tree_size=300)

# it may take time to finish the indexing operation

document = "Here comes the sun"
vector = model.features_num(document.lower().split())

# get most similar documents
similar_documents = model.most_similar([vector], indexer=indexer)
```

Currently, only Annoy-based Indexer is supported.

### DIFF
--- a/gensim/models/indexer.py
+++ b/gensim/models/indexer.py
@@ -1,0 +1,246 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2013 Radim Rehurek <me@radimrehurek.com>
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+
+
+"""
+Using 3rd party k-NN libraries to improve performance for method DocvecsArray.most_similar().
+More information can be found at: https://github.com/piskvorky/gensim/wiki/Ideas-&-Feature-proposals#integrate-a-fast-k-nn-library
+"""
+
+from doc2vec import Doc2Vec
+from word2vec import Word2Vec
+from annoy import AnnoyIndex
+
+class Indexer(object):
+    """
+    Base class for indexers
+    """
+
+    def __init__(self, model=None, **kwargs):
+        """
+        Create a new indexer
+
+        Args:
+            model:   Instance of Doc2Vec or Word2Vec
+            kwargs:  Additional named parameters
+        """
+        if model is None:
+            raise Exception("Invalid model parameter. Please provide model with an instance of Doc2Vec or Word2Vec")
+        if type(model) is Doc2Vec:
+            self._init_doc2vec_(model)
+        elif type(model) is Word2Vec:
+            self._init_word2vec_(model)
+        else:
+            raise Exception("Invalid model parameter. Please provide model with an instance of Doc2Vec or Word2Vec")
+        self.start_indexing()
+
+    def _init_doc2vec_(self, model):
+        docvecs = model.docvecs
+        docvecs.init_sims()
+        size = len(docvecs.doctag_syn0norm)
+
+        for i in range(size):
+            self.add_item(docvecs.offset2doctag[i], docvecs.doctag_syn0norm[i])
+
+    def _init_word2vec_(self, model):
+        model.init_sims()
+        size = len(model.syn0norm)
+
+        for i in range(size):
+            self.add_item(model.index2word[i], model.syn0norm[i])
+
+    def get_item(self, label):
+        """
+        Get an item's vector by its label
+
+        Args:
+            label: The label
+
+        Returns:
+            The item's vector if have. Otherwise, returns None
+        """
+        pass
+
+    def add_item(self, label, vector):
+        """
+        Add an item to index
+
+        Args:
+            label: the label of the item (must be unique).
+            vector: the item's vector.
+        """
+        pass
+
+    def start_indexing(self):
+        """
+        Start the indexing operation.
+        This method is supposed to run after all items are added
+        It may take time to finish.
+        """
+        pass
+
+    def get_nearest_items(self, vector, top_n=10):
+        """
+        Get nearest items for an item described by its vector
+
+        Args:
+            vector: The vector
+            top_n: Number of nearest items to get
+
+        Returns:
+            A list of tuples (label, similarity) for the nearest ones
+        """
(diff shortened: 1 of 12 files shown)


anhldbk:
@tmylk @piskvorky I'm terribly sorry for not mastering skills on using Github. 
I accidentally deleted my Gensim repo, so I had to close the PR [#548](https://github.com/piskvorky/gensim/pull/548) to create this one.
All things that @piskvorky  mentioned, I already fixed.

tmylk:
Hi @anhldbk , happy to talk in the chat channel if you wish to come back to this PR. https://gitter.im/piskvorky/gensim

anhldbk:
@tmylk I've already in the channel. What's for now?

tmylk:
Ping @piskvorky

gensim/models/indexer.py
```diff
@@ -0,0 +1,246 @@
...
+
+        Args:
+            file_name:   Output file path
+
+        Returns:
+            Returns true on success. Otherwise, returns false
+        """
+        pass
+
+    def load(self, file_name):
```
piskvorky: `load` should be a class method, returning the loaded object.

gensim/models/indexer.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+"""
+Using 3rd party k-NN libraries to improve performance for method DocvecsArray.most_similar().
+More information can be found at: https://github.com/piskvorky/gensim/wiki/Ideas-&-Feature-proposals#integrate-a-fast-k-nn-library
+"""
+
+from doc2vec import Doc2Vec
+from word2vec import Word2Vec
+from annoy import AnnoyIndex
```
piskvorky: We cannot depend on `annoy` being installed (=must not be a mandatory dependency).

gensim/models/indexer.py
```diff
@@ -0,0 +1,246 @@
...
+        Constructor
+
+        Args:
+            features_num: Number of features
+            tree_size: Number of trees. More trees gives higher precision when querying.
+
+        Raises:
+            TypeError: not provide enough parameters 
+        """
+        if "features_num" not in kwargs:
```
piskvorky: Why not simply name the parameter inside `__init__(model=None, feature_num=None)` ? Looks identical but more idiomatic.

gensim/models/indexer.py
```diff
@@ -0,0 +1,246 @@
...
+        Get nearest items for an item described by its vector
+
+        Args:
+            vector: The vector
+            top_n: Number of nearest items to get
+
+        Returns:
+            A list of tuples (label, similarity) for the nearest ones
+        """
+        nearest = self.annoy_indexer.get_nns_by_vector(vector, top_n, include_distances=True)
```
piskvorky: Can you unpack the `nearest` tuple into named vars? `nearest[0]` and `nearest[1]` are too opaque.

piskvorky:
Just a few minor comments & API changes, but looks very nice & clean overall. Great job @anhldbk !

Are you planning to add some unit tests?

gensim/models/indexer.py
anhldbk: @piskvorky You're right! I should have AnnoyIndexer implemented in a separate module right? 
Is a good idea to have a sub package named `indexer` in pacakge `gensim.models` ?

anhldbk:
@piskvorky It's my pleasure to work with you! I really appreciate your beautiful code. :)

anhldbk:
@piskvorky So I'm gonna have my test in package `gensim.test` right?

I'm thinking of a test scenario in the following way:
+ Training a Word2Vec model 
+ Creating and tuning an indexer based on Annoy
+ The test's considered to be passed if `model.most_similar` and `indexer.most_similar` yield the same result.

Do you think it's reasonable?

gensim/models/indexer.py
piskvorky: We already have a subpackage for indexing, called `gensim.similarities` -- I think the indexers belong there.

It's not a problem to import `annoy`, but the import cannot happen automatically, on `import gensim` or `import gensim.similarities` (=in `__init__.py`). That would fail for people who don't have Annoy installed, and we definitely don't want to make Annoy a mandatory dependency for gensim.

But we can let users import a module that needs annoy manually: `from gensim.similarities.annoy_index import AnnoyIndex`, that should be fine.

piskvorky:
Yes, I think it's a reasonable test. Just made sure the test is skipped when there's no Annoy installed (import fails).

karlhigley:
Since Annoy returns approximate nearest neighbors, testing for an exact match may lead to a flaky test. To get more consistent test behavior, it may be better to check that either:
* The distances of returned neighbors are below some threshold, or
* The precision of the returned neighbors vs the actual neighbors is above some threshold

gensim/models/indexer.py
karlhigley: `num_features` isn't used by the base `Indexer` class, so maybe it makes sense to hold off on adding `num_features` as an `__init__` param until more sub-classes also use it?

gensim/models/indexer.py
karlhigley: If `load` becomes a class method, there are a few issues that crop up:
* The initialization/indexing code based on the model is not needed when loading the index from disk.
* The `index_label_dict` and `label_index_dict` would need to saved/loaded in order to maintain the mapping between labels, ids, and vectors.

It seems like the process of building an index from a model could go in a class method, and `__init__` could take the index instead of the model. Not sure what to do about persisting the mappings, though.

tmylk:
Implemented in #784

PR #620, opened 2016-02-28
Add alpha=50 so Mallet unittests pass in 2.0.8

tmylk:
Breaking change for Mallet 2.0.8

Compatible in Mallet 2.0.7


piskvorky:
Mention the change in ctr defaults in CHANGELOG too?

Issue #611, opened 2016-02-17
Runtime error while running deeper.ipynb 'Deep Inverse Regression with Yelp reviews'


aliozcan:
I am trying to get deeper.ipynb done. But I got an error while running line above.
```
# get the probs (note we give docprob a list of lists of words, plus the models)
probs = docprob( [r['x'] for r in revtest], starmodels )
```
error looks like 

> ---------------------------------------------------------------------------
> RuntimeError                              Traceback (most recent call last)
> <ipython-input-11-cae934a7f464> in <module>()
>       1 # get the probs (note we give docprob a list of lists of words, plus the models)
> ----> 2 probs = docprob( [r['x'] for r in revtest], starmodels )
> 
> <ipython-input-9-a066fbad7f8d> in docprob(docs, mods)
>      13     sentlist = [s for d in docs for s in d]
>      14     # the log likelihood of each sentence in this review under each w2v representation
> ---> 15     llhd = np.array( [ m.score(sentlist, len(sentlist)) for m in mods ] )
>      16     # now exponentiate to get likelihoods,
>      17     lhd = np.exp(llhd - llhd.max(axis=0)) # subtract row max to avoid numeric overload
> 
> /usr/local/lib/python2.7/site-packages/gensim/models/word2vec.pyc in score(self, sentences, total_sentences, chunksize, queue_factor, report_delay)
>     892 
>     893         if not self.hs:
> --> 894             raise RuntimeError("we have only implemented score for hs")
>     895 
>     896         def worker_loop():
> 
> RuntimeError: we have only implemented score for hs
> 

I installed gensim on my MBP via pip. Does anyone has an idea what might be the problem?

piskvorky:
"hs" stands for hierarchical softmax... though the notebook should be working of course.

CC @mataddy -- were there any changes that could have broken the notebook?

mataddy:
hmmm, I'll check this out.

mataddy:
I can replicate the bug.  I'll dig into it and fix this.

mataddy:
the error message is totally correct; the problem is that the default for word2vec changed from `hs` to `negative` and I didn't have this option fixed in the notebook. 

It is fixed now, and I've updated the doc to highlight that score is only implemented for hs.  A pull request is forthcoming.

As an aside, I think that I could devise a score scheme for the negative sampling trained models.  It's tougher to justify a connection to a likelihood but I could just proceed by analogy and it would probably work.  I'll create an issue for it in my fork.

[...]

PR #622, opened 2016-02-29
fix deepir notebook to account for new defaults in w2v

mataddy:
this is to fix the issue pointed out in #611 ; also includes a small change to the docs.


piskvorky:
Thanks :)

@tmylk please review and merge.

tmylk:
Tested in https://github.com/piskvorky/gensim/issues/611

PR #624, opened 2016-03-03
Allow build_vocab to take progress_per parameter

zer0n:
Otherwise, it would print out too many lines when training on datasets such as Wikipedia.


tmylk:
@zer0n Thanks, that is a useful addition. Sent a PR for changelog note to your branch

zer0n:
Thanks @tmylk . Merged your changelog.

PR #627, opened 2016-03-16
Test data file for glove2word2vec.py

manasRK:
This is a 50*50 embedding file for a test for glove2word2vec.py


piskvorky:
Add that to the existing PR please; no need for an extra PR.

piskvorky:
Also, it's better if the matrix dimension are asymmetrical (not 50x50), to catch possible errors in header ordering.

manasRK:
Okay, adding the test file with updated program now.

PR #636, opened 2016-03-24
[MRG] Minor pep8 fixes in docsim.py

dsquareindia:
Minor changes in docsim.py to make it more pep8 compliant. Left some places where I felt readability would suffer.

### DIFF
--- a/gensim/similarities/docsim.py
+++ b/gensim/similarities/docsim.py
@@ -8,43 +8,50 @@
 This module contains functions and classes for computing similarities across
 a collection of documents in the Vector Space Model.
 
-The main class is `Similarity`, which builds an index for a given set of documents.
-Once the index is built, you can perform efficient queries like "Tell me how similar
-is this query document to each document in the index?". The result is a vector
-of numbers as large as the size of the initial set of documents, that is, one float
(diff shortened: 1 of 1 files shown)


gensim/similarities/docsim.py
```diff
@@ -104,7 +113,8 @@ def __getstate__(self):
         return result
 
     def __str__(self):
-        return ("%s Shard(%i documents in %s)" % (self.cls.__name__, len(self), self.fullname()))
+        return ("%s Shard(%i documents in %s)" % (self.cls.__name__, len(self),
```
piskvorky: We don't use hanging indent. It makes code harder to read (too cramped).

piskvorky:
Thanks. We don't follow the 80-char guideline religiously, so I'm not sure we need this. But I guess it doesn't hurt either.

The vertically aligned indent changes we definitely don't want (use normal indent, no args on first line).

gensim/similarities/docsim.py
dsquareindia: Oh alright sorry for that! I'll make the changes.

dsquareindia:
I'll close this. Thanks for informing me about the conventions!

PR #648, opened 2016-03-29
LGPL v2.1 License clarification

tmylk:


setup.py
```diff
@@ -142,7 +142,6 @@ def readfile(fname):
         'Hierarchical Dirichlet Process, HDP, Random Projections, '
         'TFIDF, word2vec',
 
-    license='LGPL',
```
piskvorky: Why this deletion?

tmylk: Our license is standard so has a Trove Classifier specified later.

 From the [setup.py](https://docs.python.org/2/distutils/setupscript.html) guide
> The license field is a text indicating the license covering the package where the license is not a selection from the “License” Trove classifiers. See the Classifier field. Notice that there’s a licence distribution option which is deprecated but still acts as an alias for license.

piskvorky: OK. Do we want to support v2 though? I've never heard of that version, it's another potential can of worms. I'd prefer 2.1+.

piskvorky: The COPYING file also refers to 2.1.

tmylk: I was trying to put v2.1 everywhere. Please let me know if there is v2 that has been accidentally added anywhere.

Do you mean that the PyPi classifier refers to v2?

Unfortunately there is no LGPL v2.1  [PyPi classifier](https://pypi.python.org/pypi?%3Aaction=list_classifiers) .
LGPL v2.1 falls under the LGPLv2+ that we use 'License :: OSI Approved :: GNU Lesser General Public License v2 or later (LGPLv2+)'. We are explicit in COPYING that it is specifically v2.1 so it should be ok.

piskvorky: Yes, that's the one I meant, that v2+ line in `setup.py`. Hopefully it's OK like this... I'm not gonna hire a lawyer for this :)

We've had exactly one version clarification request in 5 years, so probably not a major issue for most users.

Issue #69, opened 2011-11-29
make normalization a transformation


piskvorky:
Right now the similarity computations assume *cosine similarity* and transform all vectors to unit length implicitly, inside the `*Similarity` classes.

Instead, make normalization an explicit transformation and leave it up to the user to choose which (or whether) normalization to use. Example  `vector = norm_l2[lsi_model[norm_l2[tfidf_model[bow_vector]]]]`.

The most common cases (L2 norm, L1 norm, identity) should be pre-defined, probably in `gensim.matutils`.
 
Also connected is issue #64 (allow custom similarity metrics).

dsquareindia:
@piskvorky can I take this up? Should I add another parameter to the Similarity constructor to include a `norm` parameter with default as `norm_l2` and other o? As you suggested, I'll implement l2, l1 and identity in matutils.

dsquareindia:
can this be closed?

piskvorky:
Thanks a lot @dsquareindia !

Successfully closing a ticket from 2011, that's really rare :D Great job.

PR #649, opened 2016-03-30
[MRG] ENH: Make normalization an explicit transformation.

dsquareindia:
Addresses issue #69. What this PR includes:
* Added support for ''l1' norm.
* Added norm parameter to Similarity constructor.
* Added norm parameter to matutils.unitvec.
* Created file models/normmodel.py to make normalization an explicit transformation.

TODO:
- [X] Check for input validation
- [X] Make test file test/test_norms.py
- [X] Add tests for input validation using assertRaises
- [X] Add documentation
- [X] Add more tests to testTransform to cover all input type validations

@piskvorky @tmylk Could you please tell me if I'm going on the right track?

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -319,12 +319,17 @@ def veclen(vec):
     assert length > 0.0, "sparse documents must not contain any explicit zero entries"
     return length
 
+def ret_normalized_vec(vec, length):
+    if length != 1.0:
+        return [(termid, val / length) for termid, val in vec]
+    else:
+        return list(vec)
 
 blas_nrm2 = blas('nrm2', numpy.array([], dtype=float))
 blas_scal = blas('scal', numpy.array([], dtype=float))
 
 
-def unitvec(vec):
+def unitvec(vec, norm='l2'):
     """
     Scale a vector to unit length. The only exception is the zero vector, which
     is returned back unchanged.
@@ -332,9 +337,14 @@ def unitvec(vec):
     Output will be in the same format as input (i.e., gensim vector=>gensim vector,
     or numpy array=>numpy array, scipy.sparse=>scipy.sparse).
     """
-    if scipy.sparse.issparse(vec): # convert scipy.sparse to standard numpy array
+    if norm not in ('l1', 'l2'):
+        raise ValueError("'%s' is not a supported norm. Currently supported norms are 'l1' and 'l2'." % norm)
+    if scipy.sparse.issparse(vec):
         vec = vec.tocsr()
-        veclen = numpy.sqrt(numpy.sum(vec.data ** 2))
+        if norm == 'l1':
+            veclen = numpy.sum(numpy.abs(vec.data))
+        if norm == 'l2':
+            veclen = numpy.sqrt(numpy.sum(vec.data ** 2))
         if veclen > 0.0:
             return vec / veclen
         else:
@@ -342,7 +352,10 @@ def unitvec(vec):
 
     if isinstance(vec, numpy.ndarray):
         vec = numpy.asarray(vec, dtype=float)
-        veclen = blas_nrm2(vec)
+        if norm == 'l1':
+            veclen = numpy.sum(numpy.abs(vec))
+        if norm == 'l2':
+            veclen = blas_nrm2(vec)
         if veclen > 0.0:
             return blas_scal(1.0 / veclen, vec)
         else:
@@ -353,13 +366,13 @@ def unitvec(vec):
     except:
         return vec
 
-    if isinstance(first, (tuple, list)) and len(first) == 2: # gensim sparse format?
-        length = 1.0 * math.sqrt(sum(val ** 2 for _, val in vec))
+    if isinstance(first, (tuple, list)) and len(first) == 2: # gensim sparse format
+        if norm == 'l1':
+            length = float(sum(abs(val) for _, val in vec))
+        if norm == 'l2':
+            length = 1.0 * math.sqrt(sum(val ** 2 for _, val in vec))
         assert length > 0.0, "sparse documents must not contain any explicit zero entries"
-        if length != 1.0:
-            return [(termid, val / length) for termid, val in vec]
-        else:
-            return list(vec)
+        return ret_normalized_vec(vec, length)
     else:
         raise ValueError("unknown input type")
 
--- a/gensim/models/normmodel.py
+++ b/gensim/models/normmodel.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2012 Radim Rehurek <radimrehurek@seznam.cz>
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+
+import logging
+
+from gensim import interfaces, matutils
+
+logger = logging.getLogger(__name__)
+
+
+class NormModel(interfaces.TransformationABC):
+    """
+    Objects of this class realize the explicit normalization of
+    vectors. Supported norms are l1' and 'l2' with 'l2' being
+    default.
+
+    The main methods are:
+
+    1. Constructor which normalizes the terms in the given corpus document-wise.
+    2. The normalize() method which normalizes a simple count representation.
+    3. The [] transformation which internally calls the self.normalize() method.
+
+    >>> norm_l2 = NormModel(corpus)
+    >>> print(norm_l2[some_doc])
+    >>> norm_l2.save('/tmp/foo.tfidf_model')
+
+    Model persistency is achieved via its load/save methods
+    """
+    def __init__(self, corpus=None, norm='l2'):
+        """
+        Compute the 'l1' or 'l2' normalization by normalizing separately
+        for each doc in a corpus.
+        Formula for 'l1' norm for term 'i' in document 'j' in a corpus of 'D' documents is::
+
+          norml1_{i, j} = (i / sum(absolute(values in j)))
+
+        Formula for 'l2' norm for term 'i' in document 'j' in a corpus of 'D' documents is::
+
+          norml2_{i, j} = (i / sqrt(sum(square(values in j))))
+        """
+        self.norm = norm
+        if corpus is not None:
+            self.calc_norm(corpus)
+        else:
+            pass
+
+    def __str__(self):
+        return "NormModel(num_docs=%s, num_nnz=%s, norm=%s)" % (self.num_docs, self.num_nnz, self.norm)
+
+    def calc_norm(self, corpus):
+        """
+        Calculates the norm by calling matutils.unitvec with the norm parameter.
+        """
+        logger.info("Performing %s normalization..." % (self.norm))
+        norms = []
+        numnnz = 0
+        docno = 0
+        for bow in corpus:
+            docno += 1
+            numnnz += len(bow)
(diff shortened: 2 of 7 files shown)


dsquareindia:
@piskvorky @tmylk I've created `models/normmodel.py` in order to make normalization an explicit transformation. I hope I'm not going off on a tangent with this pr :P

dsquareindia:
I've added initial tests for just checking the logic. Input validation testing still remains.

gensim/matutils.py
```diff
@@ -332,9 +343,15 @@ def unitvec(vec):
     Output will be in the same format as input (i.e., gensim vector=>gensim vector,
     or numpy array=>numpy array, scipy.sparse=>scipy.sparse).
     """
+    norm = isnorm(norm)
     if scipy.sparse.issparse(vec): # convert scipy.sparse to standard numpy array
         vec = vec.tocsr()
-        veclen = numpy.sqrt(numpy.sum(vec.data ** 2))
+        if norm == 'l0':
+            veclen = len(vec.data)
```
tmylk: L_0 norm is not homogeneous so it is impossible to scale.

gensim/models/normmodel.py
```diff
@@ -0,0 +1,75 @@
...
+
+    >>> norm_l2 = Normmodel(corpus)
+    >>> print(norm_l2[some_doc])
+    >>> norm_l2.save('/tmp/foo.tfidf_model')
+
+    Model persistency is achieved via its load/save methods
+    """
+    def __init__(self, corpus=None, norm='l2'):
+        """
+        Compute the 'l0', 'l1' or 'l2' normalization by normalizing seperately
```
tmylk: separately

gensim/models/normmodel.py
```diff
@@ -0,0 +1,75 @@
...
+        Formula for 'l1' norm for term 'i' in document 'j' in a corpus of 'D' documents is::
+
+          norml1_{i, j} = (i / sum(absolute(values in j)))
+
+        Formula for 'l2' norm for term 'i' in document 'j' in a corpus of 'D' documents is::
+
+          norml2_{i, j} = (i / sqrt(sum(square(values in j))))
+        """
+        if norm not in ('l0', 'l1', 'l2'):
+            raise ValueError("%s is not a supported norm" % norm)
```
tmylk: Code duplication.

gensim/matutils.py
dsquareindia: Yes true.... Could you please elaborate more on the scaling part? Will it be affecting efficiency since as the corpus size increases we cannot use the fact that `f(ax) = (a^k)f(x)`? Should I remove the support for zero "norm"?

tmylk: Sorry to confuse you, meant scaling in mathematical not IT sense.

What number do we need to divide vector `a=(5,5)` by in order to have `l_0_norm(a)=1`?

dsquareindia: Thanks for the explanation! I think it would be best to remove L0 norm then and keep the normalization just for l1 and l2. Should I proceed with it?

tmylk: Yep. Thanks.

dsquareindia:
@tmylk I've removed L0 norm and added some more tests to `test_normmodel`.

dsquareindia:
@tmylk @piskvorky I've added further tests to cover all input types. I think this pr is ready for review.

dsquareindia:
Removed merge conflicts.

tmylk:
@dsquareindia Hey, why did you close this pull request? It is a useful feature.

dsquareindia:
Sorry for that; I'll open it again.

dsquareindia:
Removed redundant use of corpus, rebased and squashed. Could someone please review this?

gensim/matutils.py
```diff
@@ -354,12 +372,14 @@ def unitvec(vec):
...
     if isinstance(first, (tuple, list)) and len(first) == 2: # gensim sparse format?
-        length = 1.0 * math.sqrt(sum(val ** 2 for _, val in vec))
-        assert length > 0.0, "sparse documents must not contain any explicit zero entries"
-        if length != 1.0:
-            return [(termid, val / length) for termid, val in vec]
-        else:
-            return list(vec)
+        if norm == 'l1':
+            length = float(sum(abs(val) for _, val in vec))
+            assert length > 0.0, "Document contains all zero entries"
```
tmylk: What is the reason for this check?

gensim/models/normmodel.py
```diff
@@ -0,0 +1,69 @@
...
+        norms = {}
+        numnnz, docno = 0, -1
+        for docno, bow in enumerate(corpus):
+            numnnz += len(bow)
+            norms[docno] = matutils.unitvec(bow, self.norm)
+        self.num_docs = docno + 1
+        self.num_nnz = numnnz
+        self.norms = norms
+
+    def __getitem__(self, bow):
```
dsquareindia: call it normalize. getitem a bit confusing.

gensim/test/test_normmodel.py
```diff
@@ -0,0 +1,134 @@
...
+        col = numpy.array([0, 2, 2, 0, 1, 2])
+        data = numpy.array([1, 2, 3, 4, 5, 6])
+        sparse_matrix = csr_matrix((data, (row, col)), shape=(3, 3))
+        normalized = model_l1[sparse_matrix]
+
+        # Check if output is of same type
+        self.assertTrue(issparse(normalized))
+
+        # Check if output is correct
+        expected = numpy.array([[0.04761905, 0., 0.0952381],
```
tmylk: Test description as doc string is best
h

gensim/matutils.py
dsquareindia: Sorry I had just added that to make it symmetric with the L2 case. I should be returning the vector unchanged in case of 0 length right? Should I proceed with the change?

gensim/matutils.py
```diff
@@ -354,12 +372,14 @@ def unitvec(vec):
...
+            return ret_normalized_vec(vec, length)
+        if norm == 'l2':
+            length = 1.0 * math.sqrt(sum(val ** 2 for _, val in vec))
+            assert length > 0.0, "sparse documents must not contain any explicit zero entries"
```
dsquareindia: As done with other input types, I should be returning the vector unchanged in case of 0 length. Should I change this?

gensim/matutils.py
```diff
@@ -353,13 +371,14 @@ def unitvec(vec):
...
-            return [(termid, val / length) for termid, val in vec]
-        else:
-            return list(vec)
+    if isinstance(first, (tuple, list)) and len(first) == 2: # gensim sparse format
+        if norm == 'l1':
+            length = float(sum(abs(val) for _, val in vec))
+            return ret_normalized_vec(vec, length)
+        if norm == 'l2':
+            length = 1.0 * math.sqrt(sum(val ** 2 for _, val in vec))
+            assert length > 0.0, "sparse documents must not contain any explicit zero entries"
```
dsquareindia: Should I change this to returning the vector unchanged in case of zero length as we are doing with other input types?

dsquareindia:
@tmylk I have split the test suite into multiple tests to ease error checking. I have also removed the `__getitem__` transformation and have made it a function called `normalize()`. I hope now the pr is a bit better! :)

dsquareindia:
@piskvorky could you please review this? I think it's ready for merge.

gensim/models/normmodel.py
```diff
@@ -0,0 +1,69 @@
...
+            pass
+
+    def __str__(self):
+        return "Normmodel(num_docs=%s, num_nnz=%s)" % (self.num_docs, self.num_nnz)
+
+    def calc_norm(self, corpus):
+        """
+        Calculates the norm by calling matutils.unitvec with the norm parameter.
+        """
+        logger.info("Normalizing...")
```
piskvorky: This log message is not very illuminating to users... try being more specific.

gensim/models/normmodel.py
```diff
@@ -0,0 +1,69 @@
...
+        norms = {}
+        numnnz, docno = 0, -1
+        for docno, bow in enumerate(corpus):
+            numnnz += len(bow)
+            norms[docno] = matutils.unitvec(bow, self.norm)
```
piskvorky: how about dropping `docno` and using `len(norms)` instead? Seems to simplify the logic a bit.

piskvorky:
Only did a quick scan, looks nicely done, but I'll leave the thorough review to @tmylk.

My only nitpick would be `Normmodel` => `NormModel`, for readability. WDYT?

gensim/models/normmodel.py
```diff
@@ -0,0 +1,69 @@
...
+        self.num_docs = docno + 1
+        self.num_nnz = numnnz
+        self.norms = norms
+
+    def normalize(self, bow):
```
dsquareindia: @piskvorky should this rather be a `__getitem__()` since a transformation [here](https://radimrehurek.com/gensim/interfaces.html#gensim.interfaces.TransformationABC) is defined as something which acts like `transformation[doc]`.

piskvorky: Sure. We can also let users use either -- a named function or `[ ]`.

gensim/models/normmodel.py
dsquareindia: Sorry I didn't understand this. The norms dictionary itself is being constructed by iterating through all the docs in the corpus in this loop. Did you mean rather than using `enumerate(corpus)` I can simply iterate through the corpus and use `len(norms)` for indexing? While writing this, I personally felt that explicitly mentioning `docno` would enhance the readability of the code.....

tmylk: We could just return a list, same as corpus.

gensim/matutils.py
```diff
@@ -319,30 +319,48 @@ def veclen(vec):
...
+    if length != 1.0:
+        return [(termid, val / length) for termid, val in vec]
+    else:
+        return list(vec)
+
+def isnorm(norm):
+    if norm not in ('l1', 'l2'):
+        raise ValueError("'%s' is not a supported norm" % norm)
+    else:
+        return norm
```
tmylk: just return `(lambda x: unitvec(x,norm)`

gensim/matutils.py
```diff
@@ -319,30 +319,48 @@ def veclen(vec):
...
 
+def ret_normalized_vec(vec, length):
+    if length != 1.0:
+        return [(termid, val / length) for termid, val in vec]
+    else:
+        return list(vec)
+
+def isnorm(norm):
+    if norm not in ('l1', 'l2'):
+        raise ValueError("'%s' is not a supported norm" % norm)
```
tmylk: This check should be inside unitvec, at the end of if-else. Then will be no need for this `isnorm` function.

dsquareindia: I had written it here to avoid code duplication in other modules which need to validate the input norm from user.

dsquareindia:
@tmylk @piskvorky I've addressed all the comments. Hopefully it's much better now! Could you please review :)

dsquareindia:
@tmylk @piskvorky thanks a ton for the reviews! :beers:

dsquareindia:
@piskvorky I was thinking of writing a notebook tutorial for this feature however I can't think of a practical use case for it. What would be the best way to go about this?

piskvorky:
Not sure... @tmylk ? Maybe indexing some corpus with/without normalization, showing how the returned "most similar" docs (or words, or embeddings... basically any vectors) differ?

Issue #381, opened 2015-07-02
Get word from vector


sheavalentine:
It would be useful to have a method which returns the closest word when the model is given a vector.

piskvorky:
What model?

sheavalentine:
Sorry... word2vec.  Eg

```python
vectors =  Word2Vec.load_word2vec_format(file)
vectors.word_from_vec(vectors['cat']) == 'cat'
```

gojomo:
You can already do this by passing a single positive example to `most_similar()`, like so:

    vectors.most_similar(positive=[vectors['cat']],topn=1) 

I suppose as a convenience, it could detect a single ndarray (like it detects a single string) and just assume this is what the caller wants...

sheavalentine:
! Oh... that's not entirely obvious.  The expectation was that `positive` needed to be a list of strings.

gojomo:
Yes, the doc-comment should be more clear... and any future/revised tutorials should include a few examples of the multiple styles of arguments it takes.

sheavalentine:
Thank you - also thank you for making this tool.

piskvorky:
Good point.

We'll probably split this API into two distinct methods (`similar_by_word`, `similar_by_vector`) if we want to promote it. Overloading arguments ultimately brings too much trouble and confusion, this was originally supposed to be just an internal quick hack.

sheavalentine:
I can do it, but it'll take me a few days to get the bandwidth.  :)

piskvorky:
No problem.

We plan to make a new release this weekend, so this change will be in the release after that, then.

tmylk:
@sheavalentine Is this still a needed feature? Do you think we could include your Pull Request into the January release?

tmylk:
Closing as abandoned

PR #651, opened 2016-03-31
allow int/string type doc2vec tag

isohyt:
Allow calls like TaggedDocument(words=words,tags='SENT_99'), as a shorthand for TaggedDocument(words=words,tags=['SENT_99'])

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,7 +1,7 @@
 Changes
 =======
 
-* Allow calls like TaggedDocument(words=words,tags='SENT1') or TaggedDocument(words=words,tags=1), as a shorthand for TaggedDocument(words=words,tags=['SENT1'])or TaggedDocument(words=words,tags=["1"]) (Hayate ISO, #651)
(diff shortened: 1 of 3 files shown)


tmylk:
@isohyt Thanks for the PR

Is it a fix for https://github.com/piskvorky/gensim/issues/381 ?

Please add new tests and a note in CHANGELOG.

gojomo:
Mixed feelings about this sort of special-casing: may hide functionality (multiple tags), saves just two characters. But also may turn a common-error into something that does-what-was-expected. If done for strings, should be done for ints too.

isohyt:
@tmylk, thank you for your comment.
I'm writing doc2vec documentation and I bumped into this problem.
So I fixed this doc2vec tag issue.

I also have a plan to fix for issue #381.

isohyt:
@gojomo, thank you for your comment.
I added for int type tags feature.

isohyt:
I overlooked some problems. I'm fixing now.

isohyt:
Hmm...
It goes will in my local environment, but I have no idea to test this proposed code.
Should I rewrite TaggedDocument from scratch without namedtuple to add this feature??

isohyt:
I found the essential mistake, so I close this PR.
I want to try to add this feature from a different approach in the later.

Issue #64, opened 2011-11-04
Add other distance measures to Similarity


piskvorky:
Currently, Similarity works purely over cosine similarity (~the angle between query and indexed document).

Make this more general, using e.g. Hellinger distance for models that represent the documents as probability distributions. 

At the same time, try to still keep things computationally efficient (using BLAS & mmap etc.).

cmooony:
I am trying to add another similarity function to gensim.docsim, but when I search for cossim in the source files , I got only two results: cossim in matutils.py and matutils.cossim in test_lee.py.
So I am wandering how exactly the gensim caculate the similarity of the documents? And How can I add my own simi function for gensim?
Thanks and looking forward to your advice!

piskvorky:
The classes in docsim only support dot product + cossim.

You can add another similarity function by simply writing it and using it; or what sort of API do you need? 

Or did you expect the `Similarity` class to accept arbitrary sim functions as input?

cmooony:
Ahh, thanks. I got it. It is computed as dot product, in MatrixSimilarity ```result = numpy.dot(self.index, query.T).T```, and In SparseMatrixSimilarity, it is ```result = self.index * query.tocsc()```
As you mentioned in #69 , I think it is a better way for 'humans' to choose when we need another method predifined in ```gensim.matutils```.
Thanks again for your nice work!

piskvorky:
Well we support other ways too, it was a honest question.

I'm just not sure what kind of functionality people expect, how to structure the API. I'm not a fan engineering new functionality without clear use cases :)

methodds:
Is it  a reasonable approach to use cosine similarity for gensim lda models? Or should Hellinger's distance be highly prefered? If so, I would love to see support for it. :)

piskvorky:
@methodds looking forward to the PR!

The code for Hellinger distance is really simple; see for example here:
http://stackoverflow.com/questions/22433884/python-gensim-how-to-calculate-document-similarity-using-the-lda-model#answer-22756647

bhargavvader:
Hello, is there any particular reason the `cossim` method is not being used in `docsim`, and a simple dot product is used instead? Is it because the `cossim` method in `mathutils` is only for sparse vectors?

I looked at your stackoverflow answer and wrote a method for Hellinger- but it's very simple and I'm not sure if I'm going in the right direction. 
``` python
def hellinger(vec1, vec2, num_topics):
    dense1 = sparse2full(vec1,num_topics)
[...]

PR #656, opened 2016-04-04
[MRG] Adding Distance metrics to matutils + Tutorial

bhargavvader:
@tmylk , @piskvorky , I've created 3 different metric functions: `Hellinger` and `Kullback-Leibler` for probability distributions, and `Jaccard` Similarity for Bag of Words representations. 

I have a few questions about possible changes, particularly:

1) The Hellinger and Kullback-Leibler methods right now only takes **2-tuple** representations (in numpy array, scipy sparse, or gensim vector representations). Is that ok? Most probability distributions passed should look like that. I can improve it to include dense numpy arrays on input as well as well, if it is expected that people will pass that into the function.

2) The Kullback-Leibler uses the [scipy.sparse.entropy ](http://scipy.github.io/devdocs/generated/scipy.stats.entropy.html) method, which fails if the probability distributions of any of the tuples is 0. For most cases we deal with, this should not be a problem, but if we want to allow 0s to be present in the input, I will have to rewrite the kullback-leibler calculation myself.

Jaccard similarity is primarily only used for bag of words, so I've not bothered to include other kinds of formats on input.

Is this approach for adding similarity functions ok? If it is, I can make the changes you'd suggest, add test cases, and add more of these functions if needed.

This is related to issue #64 .


tmylk:
Thanks for the PR. Please add tests and a note in CHANGELOG, then will proceed with review.

piskvorky:
@bhargavvader I don't think the semantic method (LDA, BOW) should have anything to do with the metrics.

In other words, the metrics should operate over vectors (scipy.sparse / numpy.array / gensim sequence of 2-tuples). It doesn't matter what methods were used to create these vectors.

bhargavvader:
@piskvorky , I've changed my functions to keep what you said in mind. Could you please have a look again before I write any tests? The functions take in a LDA model object only to check the number of topics, now. If it is in a different format without the object passed it will still work though, as long as the lengths of the two passed vectors are the same.

piskvorky:
No models (transformations) should be part of API. Only the end result (vector) matters, not the transformations how we got there.

If a function needs the number of features (~vector length), then let users pass that in as `num_features` (for ex: `num_features=lda.num_topics`). No need for `lda` itself.

bhargavvader:
Noted. I've changed Hellinger such that it doesn't need to accept any inputs apart from two vectors (implementation similar to `cossim` function). It works fine for all the kinds of inputs I've thrown at it, and I can write more comprehensive tests if it's looking ok.
As for Kullback-Leibler, I removed the need to pass anything but `num_of_features`. Is it looking better?
Jaccard similarity also works for the basic tests I ran on it.

I apologize for the constant revisions and mistakes!

piskvorky:
Yeah, the interface looks cleaner, thanks.

I think it's reviewable now; I'll add my comments soon.

gensim/matutils.py
```diff
@@ -382,6 +383,93 @@ def cossim(vec1, vec2):
...
 
 
+def kullback_leibler(vec1,vec2,num_of_features=None):
+    """
+    A similarity metric between two probability distributions.
+    Returns a similarity in range <0,1> where values closer to 0 mean a higher similarity.
+    Uses the scipy.stats.entropy method to identify kullback_leibler convergence value.
+    If the distribution draws from a certain number of docs, that value must be passed.
+    """
+    if scipy.sparse.issparse(vec1) and scipy.sparse.issparse(vec2):
```
piskvorky: What happens when one input is sparse, the other dense?

gensim/matutils.py
```diff
@@ -382,6 +383,93 @@ def cossim(vec1, vec2):
     return result
 
 
+def kullback_leibler(vec1,vec2,num_of_features=None):
```
piskvorky: PEP8: space after comma. Also `num_of_features` => `num_features`.

gensim/matutils.py
```diff
@@ -382,6 +383,93 @@ def cossim(vec1, vec2):
...
+        if vec1.shape[1] == 2 and vec2.shape[1] == 2:
+            vec1 = vec1.tolist()
+            vec2 = vec2.tolist()
+            max_len = max(len(vec1),len(vec2),num_of_features)
+            dense1 = sparse2full(vec1,max_len)
+            dense2 = sparse2full(vec2,max_len)
+            return scipy.stats.entropy(dense1,dense2)
+        else:
+            return scipy.stats.entropy(vec1,vec2)
+    elif type(vec1) is list and type(vec2) is list:
```
piskvorky: Just use `else` (the "sequence of 2-tuples" is not necessarily a `list`; it can be an iterator, tuple...).

gensim/matutils.py
```diff
@@ -382,6 +383,93 @@ def cossim(vec1, vec2):
...
+        else:
+            return scipy.stats.entropy(vec1,vec2)
+    elif isinstance(vec1,numpy.ndarray) and isinstance(vec2,numpy.ndarray):
+        if vec1.shape[1] == 2 and vec2.shape[1] == 2:
+            vec1 = vec1.tolist()
+            vec2 = vec2.tolist()
+            max_len = max(len(vec1),len(vec2),num_of_features)
+            dense1 = sparse2full(vec1,max_len)
+            dense2 = sparse2full(vec2,max_len)
+            return scipy.stats.entropy(dense1,dense2)
```
piskvorky: PEP8: space after comma (here and elsewhere).

gensim/matutils.py
```diff
@@ -382,6 +383,93 @@ def cossim(vec1, vec2):
...
+
+
+def jaccard(vec1,vec2):
+    """
+    A similarity metric between bags of words representation.
+    Return the intersection divided by union, where union is the sum of the size of the two bags.
+    Highest value jaccard similarity of two bags is 1/2, which indicates the highest similarity.
+    """
+    intersection = 0
+    union = sum(row[1] for row in vec1) + sum(row[1] for row in vec2)
```
piskvorky: This is confusing; what is a `row`? Why does iterating over a vector produce "rows"?

bhargavvader: `row` is probably a poor choice of word, what I mean to do is to get the value of the `weight` of the corresponding `id` in a bag of words representation which Jaccard makes sense for. What should I change this variable to to make more sense?

The union is meant to be the sum of the sizes of the two bags you are comparing. This is the definition of jaccard similarity according to [this](http://infolab.stanford.edu/~ullman/mmds/ch3.pdf) text. For e.g, The bag-similarity of bags `{a, a, a, b}` and `{a, a, b, b, c}` is `1/3`. The size of the bags is `9` and the size of intersection is `3`. I'd expect these bags to be represented as `[(0,3),(1,1)]`  and `[(0,2),(1,2),(2,1)]`. 
Though now that I think about this more, I think I have to re-write my jaccard method to accept other kinds of inputs and be more generic... Will work on it.

piskvorky: Best be explicit: if the elements are `(id, weight)` pairs, then `for id, weight in vector` is the clearest expression of that.

gensim/matutils.py
bhargavvader: As of now, it'll go straight down to the final else condition - which throws an error because `matutils.sparse2full` doesn't work well with some scipy.sparse matrices. 

Would you suggest checking for cases when either are sparse, then convert to dense, and go ahead?

gensim/matutils.py
bhargavvader: Changed, thanks, I had forgotten.

gensim/matutils.py
piskvorky: Yes, if the performance is good (=not much slowdown, let's say not more than 20%), we can do that. Another option: more code paths for the various input type combinations.

bhargavvader: I've made it such that all inputted vectors are made dense and then to lists before working on them, if they aren't already dense/list. This solves the problem of input vectors being in different formats. As for the performance, since the vectors were already eventually being converted to lists, it doesn't really hurt it. It also makes the code look a lot cleaner.

bhargavvader:
@tmylk , @piskvorky - could you have a look at the methods once?
I've changed it to take into account the previous comments, and removed quite a bit of previous code duplication, the methods are much cleaner now.

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
     return result
 
 
+def isbow(vec):
+    """
+    Checks if vector passed is in bag of words representation or not.
+    """
+    if scipy.sparse.issparse(vec):
+        vec = vec.todense().tolist()
+    for item in vec:
```
piskvorky: Enough to check one item (the first one). Linear scan too slow.

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+        if type(item) is int or type(item) is long or type(item) is float:
```
piskvorky: This check seems unnecessary: `len(item)` will never equal 2 for these types, so the condition below catches all these cases.

Also, as a note, in Python `isinstance` is preferred to `type`.

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+            return False
+        if len(item) != 2:
```
piskvorky: See `utils.is_corpus` for a better 2-tuple check: https://github.com/piskvorky/gensim/blob/develop/gensim/utils.py#L591

It also includes a little heuristic for the two tuple types inside (=the two elements must be int-able and float-able).

tmylk:
@bhargavvader Ping. Are you planning to incorporate the latest suggestions?

bhargavvader:
@tmylk , yeah, I'll try and wrap up this along with the tests by next week.

bhargavvader:
@piskvorky , could you look at my `isbow`? It works for all the cases I have, and if it is fine in theory, I'll include more tests.

piskvorky:
Sure, let me check.

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
     return result
 
 
+def isbow(vec):
+    """
+    Checks if vector passed is in bag of words representation or not.
+    """
+    if scipy.sparse.issparse(vec):
```
piskvorky: This doesn't sound right -- sparse bag-of-words is not the same format as scipy.sparse.
If scipy.sparse returns True, then so should dense numpy.array... or where's the conceptual line between the formats then?

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+        vec = vec.todense().tolist()
+    try:
+        id_, val_ = vec[0]
```
piskvorky: An empty bag-of-words will throw an `IndexError` here, and `isbow` will return False as a result.
I'm not sure if that's a problem or not, but either way, empty input is clearly a corner case we have to be careful about.

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+    Returns a similarity in range <0,1> where values closer to 0 mean a higher similarity.
+    Uses the scipy.stats.entropy method to identify kullback_leibler convergence value.
+    If the distribution draws from a certain number of docs, that value must be passed.
+    """
+    if scipy.sparse.issparse(vec1):
+        vec1 = numpy.asarray(vec1.todense())
+    if scipy.sparse.issparse(vec2):
+        vec2 = numpy.asarray(vec2.todense())
+    if isinstance(vec1, numpy.ndarray):
+        vec1 = vec1.tolist()
```
piskvorky: Why are we converting arrays to lists here?

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+        dense1 = sparse2full(vec1, max_len)
+        dense2 = sparse2full(vec2, max_len)
+        return scipy.stats.entropy(dense1, dense2)
+    else:
+        return scipy.stats.entropy(vec1, vec2)
+
+
+def hellinger(vec1, vec2):
+    """
+    Hellinger distance is a distance metric to quanitfy the similarity between two probability distributions.
```
piskvorky: Typo: quantify.

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+        vec1 = numpy.asarray(vec1.todense())
+    if scipy.sparse.issparse(vec2):
+        vec2 = numpy.asarray(vec2.todense())
+    if isinstance(vec1, numpy.ndarray):
+        vec1 = vec1.tolist()
+    if isinstance(vec2, numpy.ndarray):
+        vec2 = vec2.tolist()
+    if isbow(vec1) and isbow(vec2):
+        vec1, vec2 = dict(vec1), dict(vec2)
+        if len(vec2) < len(vec1):
```
piskvorky: The reason for this `if` is not obvious and deserves a comment.

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+    """
+    if scipy.sparse.issparse(vec1):
+        vec1 = numpy.asarray(vec1.todense())
+    if scipy.sparse.issparse(vec2):
+        vec2 = numpy.asarray(vec2.todense())
+    if isinstance(vec1, numpy.ndarray):
+        vec1 = vec1.tolist()
+    if isinstance(vec2, numpy.ndarray):
+        vec2 = vec2.tolist()
+    if isbow(vec1) and isbow(vec2):
```
piskvorky: What happens if `vec1` is bag-of-words and `vec2` is a numpy array?

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+            vec1 = numpy.asarray(vec1.todense())
+        if scipy.sparse.issparse(vec2):
+            vec2 = numpy.asarray(vec2.todense())
+        intersection = 0
+        union = sum(weight for id_,weight in vec1) + sum(weight for id_,weight in vec2)
+        vec1, vec2 = dict(vec1), dict(vec2)
+        for item in vec1:
+            if item in vec2:
+                intersection = min(vec2[item],vec1[item]) + intersection
+        return(float(intersection)/float(union))
```
piskvorky: No need for brackets around `return` in Python.

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+            vec2 = numpy.asarray(vec2.todense())
+        if isinstance(vec1, numpy.ndarray) and len(vec1) == 1:
+            vec1 = vec1.tolist()[0]
+        if isinstance(vec2, numpy.ndarray) and len(vec2) == 1:
+            vec2 = vec2.tolist()[0]
+        intersection = 0
+        union = []
+        for item in vec1:
+            if item in vec2:
+                intersection = intersection + 1
```
piskvorky: `intersection += 1` in Python.

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+        for item in vec1:
+            if item in vec2:
+                intersection = min(vec2[item],vec1[item]) + intersection
+        return(float(intersection)/float(union))
+    else:
+        if scipy.sparse.issparse(vec1):
+            vec1 = numpy.asarray(vec1.todense())
+        if scipy.sparse.issparse(vec2):
+            vec2 = numpy.asarray(vec2.todense())
+        if isinstance(vec1, numpy.ndarray) and len(vec1) == 1:
```
piskvorky: The logic for this `if` is not obvious and deserves a comment. Why is length==1 a special case?

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+            vec1 = vec1.tolist()[0]
+        if isinstance(vec2, numpy.ndarray) and len(vec2) == 1:
+            vec2 = vec2.tolist()[0]
+        intersection = 0
+        union = []
+        for item in vec1:
+            if item in vec2:
+                intersection = intersection + 1
+                vec2.remove(item)
```
piskvorky: I'm not sure I follow the logic fully (use more comments), but Python has built-in types, `set` and `frozenset`. These allow efficient set operations like union, intersection or difference.
Simulating sets using lists seems suboptimal and unnecessarily complicated.

gensim/matutils.py
bhargavvader: I only check if it's in sparse format because I want to convert it to a list - because it's easier working with lists - especially because I was iterating over lists in the previous implementation. 
If it is actually sparse or not shouldn't make a difference to checking if it's a bow - right?

gensim/matutils.py
bhargavvader: In this use-case of similarity functions an empty input does not make any sense, but yes, I'll handle that case. What would be the ideal thing to do if it is indeed empty?

gensim/matutils.py
bhargavvader: Will change.

gensim/matutils.py
bhargavvader: I wrote this bit a while ago and I just can't seem to remember why - I've removed it and tried and it's working fine, though I suspect I converted to lists then to catch a corner case somewhere. 
I'm sorry, I should've looked into that before asking you to review, will address in next commit.

gensim/matutils.py
bhargavvader: I check if both are in bow format because then I proceed in a way such that different length inputs don't matter (using dicts). It otherwise just goes ahead with the Hellinger distance normally, which covers the input cases of cases like LDA probability distribution vectors. Is that approach ok?
At any rate, I'll add more comments explaining what is going on.

gensim/matutils.py
bhargavvader: Ah, I didn't think of this. Would I have to convert one to another and proceed?

gensim/matutils.py
bhargavvader: Noted.

gensim/matutils.py
bhargavvader: Noted.

gensim/matutils.py
bhargavvader: If it is not a bow and an `ndarray` then it will look like this, for e.g - `numpy.array([[6, 1, 1, 9]])`, and to use it as a list later I need to get only `[6,1,1,9]`, so I check length == 1. If it is any other length and also not a BOW that input sequence can't be used for `jaccard`. 
The reason I convert it to a list is to check for intersection and union, but that seems redundant now considering `set` and `frozenset` like you mentioned below.

gensim/matutils.py
bhargavvader: I'll check out both and see if it can be used - the thing with jaccard is it doesn't use the standard definitions of union and intersetion which is normally used with sets, so I need to check it out once. 
I'll use more comments for my code in my next commit and see if `set` and `frozenset` fit in.

gensim/matutils.py
piskvorky: I think a good course of action is: make the simplest possible version, make sure it's correct (unittests). For example, this could be converting the inputs (no matter its format) into two numpy arrays and then performing the calculation on that.

Once this version is debugged and working, profile it for different types of input, and perhaps optimize for different code-paths (such as when both inputs are scipy.sparse, or both are bow, etc).

gensim/matutils.py
bhargavvader: @piskvorky , I do the sparse check only to make my checking if it's a bow or not easier - so would it be ok for me to keep this? I now do:

`vec = numpy.array(vec.todense())`

before going ahead, so that the rest of the code only deals with either numpy arrays or lists.

piskvorky: I don't think `isbow` should return `True` for scipy.sparse inputs. It's a conceptual mismatch.

There already exists a check for scipy.sparse inputs: `scipy.sparse.issparse()`.

Or what is the purpose of "isbow()", conceptually?

gensim/matutils.py
piskvorky: In [utils.is_corpus()](https://github.com/piskvorky/gensim/blob/develop/gensim/utils.py#L589), we interpret empty input to mean "yes, this is bag-of-words (return True)".

For consistency, I'd do the same thing here.

gensim/matutils.py
bhargavvader: The purpose of `isbow()` is just to check if it is bag of words in the sense that if it is in a 2-tuple format. Jaccard has a different definition for bow kind of inputs, and I can also speed up the other distance functions if I have this info.

I understand that a sparse bag-of-words is not the same format as `scipy.sparse`, but I am not really checking if it is a sparse matrix or not (conceptually) - I am only checking if it happens to be in `scipy.sparse` form so I can convert it to a form (`ndarray`) I can work with.

I can also remove the `scipy.sparse` check here in the `isbow()` and only check for the same in the distance functions itself; I can then change it to dense and then call `isbow()`. That should clear up some confusion, I think.

piskvorky: Agreed on `isbow`'s purpose. Maybe I just misread the code then.

bhargavvader:
@piskvorky , I'm failing the build at `test_keywords`, but I don't think I have made any changes which would affect that. Could you tell me what I'm doing wrong? 

This is the error message:

`FAIL: test_text_keywords_words (gensim.test.test_keywords.TestKeywordsTest)
  File "/home/travis/build/piskvorky/gensim/gensim/test/test_keywords.py", line 50, in test_text_keywords_words
    self.assertEqual(len(generated_keywords), 16)
AssertionError: 11 != 16 `

piskvorky:
The keyword extractor is a relatively new addition to gensim, @tmylk knows it better than I do -- he should be able to help you with that.

bhargavvader:
@piskvorky , I've incorporated the changes you suggested in the last review. Could you please have another look? 

I also wanted some suggestions on how to write tests for this if this is ok - do I profile each metric with different kinds of inputs (sparse, list, ndarray, etc.), and then make sure it runs on a few standard inputs by asserting?

tmylk:
@bhargavvader The test strategy is correct.

bhargavvader:
@piskvorky I've also added tests for the functions. Could you have a look and see if they are comprehensive enough? They all pass fine on my local environment.

bhargavvader:
@tmylk , I've added the LDA tests as discussed previously. Those tests pass fine (on Travis as well as my system), but `test_doc2vec` is failing. I haven't made any changes to affect that. Could you have a look?

tmylk:
@bhargavvader The tests are green now.

bhargavvader:
@tmylk Great. If it's fine then you can merge - unless more tests are required.
@piskvorky , have another look if needed?

gensim/matutils.py
```diff
@@ -382,6 +383,112 @@ def cossim(vec1, vec2):
...
+    Uses the scipy.stats.entropy method to identify kullback_leibler convergence value.
+    If the distribution draws from a certain number of docs, that value must be passed.
+    """
+    if scipy.sparse.issparse(vec1):
+        vec1 = numpy.asarray(vec1.todense())
+    if scipy.sparse.issparse(vec2):
+        vec2 = numpy.asarray(vec2.todense()) # converted both the vectors to dense in case they were in sparse matrix 
+    if isbow(vec1) and isbow(vec2): # if they are in bag of words format we make it dense
+        if num_features != None: # if not None, make as large as the documents drawing from
+            max_len = max(len(vec1), len(vec2), num_features)
```
piskvorky: Why this `max`? If the number of features is given, it should just be used.

Also, if `num_features` really is larger than `len(vec1)` and `len(vec2)`, it will have no effect on the result anyway. Adding extra zeros to the end of both vectors shouldn't change the outcome of `entropy`.

gensim/matutils.py
```diff
@@ -382,6 +383,112 @@ def cossim(vec1, vec2):
...
+            dense1 = sparse2full(vec1, max_len)
+            dense2 = sparse2full(vec2, max_len)
+            return scipy.stats.entropy(dense1, dense2)
```
piskvorky: This was imported as just `entropy` earlier.

gensim/matutils.py
```diff
@@ -382,6 +383,112 @@ def cossim(vec1, vec2):
...
+        else:
+            max_len = max(len(vec1), len(vec2))
+            dense1 = sparse2full(vec1, max_len)
+            dense2 = sparse2full(vec2, max_len)
+            return scipy.stats.entropy(dense1, dense2)
```
piskvorky: dtto

gensim/matutils.py
```diff
@@ -382,6 +383,112 @@ def cossim(vec1, vec2):
...
+    else:
+        if len(vec1) == 1:
```
piskvorky: Not sure what this branch is for. Why would we accept "list within a list"?

gensim/matutils.py
```diff
@@ -382,6 +383,112 @@ def cossim(vec1, vec2):
...
+
+def kullback_leibler(vec1, vec2, num_features=None):
+    """
+    A similarity metric between two probability distributions.
+    Returns a similarity in range <0,1> where values closer to 0 mean a higher similarity.
+    Uses the scipy.stats.entropy method to identify kullback_leibler convergence value.
+    If the distribution draws from a certain number of docs, that value must be passed.
+    """
+    if scipy.sparse.issparse(vec1):
+        vec1 = numpy.asarray(vec1.todense())
```
piskvorky: Isn't simply `sparse.toarray` better? (`todense` returns a matrix, not sure why we'd want that).

gensim/matutils.py
```diff
@@ -382,6 +383,112 @@ def cossim(vec1, vec2):
...
+    # converting from sparse for easier manipulation    
+    if isbow(vec1) and isbow(vec2): 
+        # if it's in bow format, we use the following definitions:
+        # union = sum of the 'weights' of both the bags
+        # intersection = lowest weight for a particular id; basically the number of common words or items 
+        intersection = 0
+        union = sum(weight for id_, weight in vec1) + sum(weight for id_, weight in vec2)
+        vec1, vec2 = dict(vec1), dict(vec2)
+        for item in vec1:
+            if item in vec2:
```
piskvorky: This nested loop looks like a bug -- the outer `item` is replaced by the inner `item`??

The unit tests should definitely catch this.

gensim/matutils.py
bhargavvader: Makes sense. Will get rid of this.

gensim/matutils.py
bhargavvader: Will change.

gensim/matutils.py
bhargavvader: It isn't a nested loop, it's an `if` statement. There isn't any replacing here right?

gensim/matutils.py
bhargavvader: Wasn't aware of this! Will change.

gensim/matutils.py
bhargavvader: That's because after converting a bow to array using `sparse.toarray()` , it ends up becoming a list within a list.

For e.g, if we have a probability distribution in sparse from like this - 
`sparse = csr_matrix([0.2, 0, 0.3, 0.4, 0.1])`, and we do sparse.toarray(), it becomes `[[0.2, 0, 0.3, 0.4, 0.1]]`, which won't work. This problem is only if it was in `csr_matrix` form and then converted, and isn't a bow.

gensim/matutils.py
piskvorky: Ah yes, sorry. Too late for me :(

Faster & simpler way:

```python
for feature_id, feature_weight in vec1.itermitems():
    intersection += min(feature_weight, vec2.get(feature_id, 0.0))
```

bhargavvader: Nice idea, made the change :)

bhargavvader:
@piskvorky , made the changes just suggested. Could you look at the tests too?

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+        vec1, vec2 = dict(vec1), dict(vec2)
+        if len(vec2) < len(vec1): 
+            vec1, vec2 = vec2, vec1 # swap references so that we iterate over the shorter vector
+        sim = numpy.sqrt(0.5*sum((numpy.sqrt(value) - numpy.sqrt(vec2.get(index, 0.0)))**2 for index, value in iteritems(vec1)))
+        return sim
+    else:
+        sim = numpy.sqrt(0.5 * ((numpy.sqrt(vec1) - numpy.sqrt(vec2))**2).sum())
+        return sim
+
+def jaccard(vec1,vec2):
```
dsquareindia: pep8 here

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+    if isbow(vec1) and isbow(vec2): 
+        # if it's in bow format, we use the following definitions:
+        # union = sum of the 'weights' of both the bags
+        # intersection = lowest weight for a particular id; basically the number of common words or items 
+        intersection = 0
+        union = sum(weight for id_, weight in vec1) + sum(weight for id_, weight in vec2)
+        vec1, vec2 = dict(vec1), dict(vec2)
+        for feature_id, feature_weight in iteritems(vec1):
+            intersection += min(feature_weight, vec2.get(feature_id, 0.0))
+        return float(intersection)/float(union)
```
dsquareindia: here too

dsquareindia:
@bhargavvader sorry for some of the pep8 nitpicks! You might want to invest some time in setting up a flake8 checks plugin in your editor. That can really help eliminate these rather irritating "errors" during coding itself!

bhargavvader:
@dsquareindia , thanks for pointing those out! Yeah, I'm getting myself a PEP8 check for my editor, thanks for the advice - should save time.

bhargavvader:
@piskvorky , could you check it out?

gensim/matutils.py
```diff
@@ -382,6 +383,110 @@ def cossim(vec1, vec2):
...
+        # if it is a bag of words format, instead of converting to dense we use dictionaries to calculate appropriate distance
+        vec1, vec2 = dict(vec1), dict(vec2)
+        if len(vec2) < len(vec1): 
+            vec1, vec2 = vec2, vec1 # swap references so that we iterate over the shorter vector
+        sim = numpy.sqrt(0.5*sum((numpy.sqrt(value) - numpy.sqrt(vec2.get(index, 0.0)))**2 for index, value in iteritems(vec1)))
+        return sim
+    else:
+        sim = numpy.sqrt(0.5 * ((numpy.sqrt(vec1) - numpy.sqrt(vec2))**2).sum())
+        return sim
+
```
piskvorky: PEP8: two blanks between module-level methods.

gensim/matutils.py
```diff
@@ -471,9 +471,8 @@ def jaccard(vec1,vec2):
         intersection = 0
         union = sum(weight for id_, weight in vec1) + sum(weight for id_, weight in vec2)
         vec1, vec2 = dict(vec1), dict(vec2)
-        for item in vec1:
-            if item in vec2:
-                intersection = min(vec2[item], vec1[item]) + intersection
+        for feature_id, feature_weight in iteritems(vec1):
```
piskvorky: Put closer to the `intersection = 0.0` initialization. They belong together conceptually, so should be close together visually too.

piskvorky:
If you addressed my previous comments then it's good to go! Thanks @bhargavvader .

@tmylk please check the tests are thorough though, plus maybe do a few manual sanity checks (outside of gensim) to verify no bug in "math business logic" slipped through.

bhargavvader:
@piskvorky , made PEP8 changes and comments before code block, should be neater now. :)

bhargavvader:
@tmylk , @piskvorky , I've added a python notebook to accompany the methods. Sometimes functionalities like this get missed out on, so I've added some examples for the same so users know where they can use this.

Made changes to changelog too, so if the math business logic is alright, it's ready to merge from my end. :)

bhargavvader:
Made a few documentation changes - while Jaccard is a Similarity measure, Hellinger and KL are distance metrics. `1 - distance` = similarity in these cases.

Will make changes in the notebook to reflect the same, and add a small bit on similarities/distances between topic distributions as well document-topic distributions.

tmylk:
It is best if they are all either distance or similarities.
It is best to use
similarity = [1/(1+distance)]

piskvorky:
What's best depends on the range of `distance` -- is it `<0, inf>`, or `<0, 1>`, or something else?

I think most people expect similarity to be `<0, 1>` (though cossim is `<-1, 1>`, for example).

bhargavvader:
@tmylk , @piskvorky , will shift to using distances for all 3, in the range `<0,1>` where less distance (i.e, value closer to 0) means it has higher similarity.

bhargavvader:
@tmylk , @piskvorky , made changes to the notebook to reflect they are distance metrics and not otherwise. Added an example to find the distance between two topic's word distributions and a few document's topic distributions. 

Could you have a review to see if there's any ambiguity in the docstring, comments, and math business logic?

piskvorky:
Are these general distance functions, or truly metrics in the technical sense (satisfy triangle inequality)?

bhargavvader:
Hellinger and Jaccard are metrics in the technical sense, but KL is not.

tmylk:
More work on the ipynb is needed but merging in the code to include it in 0.13.0 release.

bhargavvader:
Cool, will update notebook in different PR.

PR #657, opened 2016-04-05
add similar by word and vector

isohyt:
Related to Issue #381 
I added `similar_by_word()` and `similar_by_vector()`
We can use these functions as follows.

trained_model.similar_by_word('graph')
      [('user', 0.9999163150787354), ...]

trained_model.similar_by_vector([1,2])
      [('survey', 0.9942699074745178), ...]

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,6 +2,7 @@ Changes
 =======
 
 * build_vocab takes progress_per parameter for smaller output (@zer0n, #624)
+* Add `similar_by_word()` and `similar_by_vector()` to word2vec (@isohyt, #381)
 
 0.12.4, 29/01/2016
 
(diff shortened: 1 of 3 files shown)


piskvorky:
-1, too much code duplication. These would be better served as simple wrappers, calling `most_similar` internally.

isohyt:
I see.
I re-wrote simple wrappers, but both are almost the same.

tmylk:
The two methods are exactly the same. Do the tests pass on your machine? 

A specific test can be run on your local machine with ``python setup.py test -m your.package.tests.test_module ``

isohyt:
Yes, and also passed Travis-ci (https://travis-ci.org/isohyt/gensim/builds/120849620).
But it is not reflected in this PR.

Anyway, I cannot figure out what should I do...

tmylk:
Thanks for the PR!
Merged in https://github.com/piskvorky/gensim/commit/b99e8528594b0898aa629f01be0919a18f25f190

PR #658, opened 2016-04-05
added documentation to save_word2vec_format

hellrich:
Added:
       `fname` is the file used to save the vectors in
         `fvocab` is an optional file used to save the vocabulary
         `binary` is an optional boolean indicating whether the data is to be saved
         in binary word2vec format (default: False)


tmylk:
Thanks! We really need more documentation improvements.

PR #663, opened 2016-04-08
Test that output directory exists.

jMonteroMunoz:
I added a small piece of code to check that the output directory exists. Before this, I waited the whole 8 hours time just to see that I had misspelled the output directory and had to restart the script.

### DIFF
--- a/gensim/scripts/make_wikicorpus.py
+++ b/gensim/scripts/make_wikicorpus.py
@@ -62,11 +62,9 @@ if __name__ == '__main__':
         print(globals()['__doc__'] % locals())
         sys.exit(1)
     inp, outp = sys.argv[1:3]
-    print "Checking if output directory exists: "+'/'.join(outp.split("/")[:-1])
-    if os.path.isdir('/'.join(outp.split("/")[:-1])) == False:
+
+    if not os.path.isdir(os.path.dirname(outp)):
         raise SystemExit("Error: The output directory does not exist. Create the directory and try again.")
-    else:
-        print "OK"
 
     if len(sys.argv) > 3:
         keep_words = int(sys.argv[3])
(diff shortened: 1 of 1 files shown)


gensim/scripts/make_wikicorpus.py
```diff
@@ -62,6 +62,12 @@
         print(globals()['__doc__'] % locals())
         sys.exit(1)
     inp, outp = sys.argv[1:3]
+    print "Checking if output directory exists: "+'/'.join(outp.split("/")[:-1])
```
piskvorky: No need for this `print` -- just raise the exception.
Also, we'd be using `logger`, not printing.

gensim/scripts/make_wikicorpus.py
```diff
@@ -62,6 +62,12 @@
...
+    if os.path.isdir('/'.join(outp.split("/")[:-1])) == False:
```
piskvorky: Check out `os.path.dirname ` and `os.path.sep` -- this line is not portable.

Also, `== False` is not Pythonic (use `not` operator instead).

jMonteroMunoz: I did the changes you suggested. Thank you for explaining. First time contributing to a project!

piskvorky:
Looks good, thanks @jMonteroMunoz!

@tmylk please take over.

tmylk:
Tested manually and merged in.

Thanks for the PR!

piskvorky:
I suppose we could be doing even more checking (is the outdir writeable? enough free space?)... but if someone needs that, they can open a PR. Probably as a general-purpose utils function in `utils.py`, so we can re-use it in other parts of gensim, in other scripts.

jMonteroMunoz:
I will try to do that whenever I get some time.

Issue #599, opened 2016-02-02
gensim==0.12.4 mallet wrapper is not compatible with mallet-2.0.7


janrygl:
Following error is raised for mallet-2.0.7. In mallet-2.0.8RC3, everything passes without any problems.

```python
for doc_no, (topics, stats) in enumerate(zip(model.load_document_topics(), docs.doc_stats)):
  File "/software/p27/local/lib/python2.7/site-packages/gensim/models/wrappers/ldamallet.py", line 271, in read_doctopics
    raise RuntimeError("invalid doc topics format at line %i in %s" % (lineno + 1, fname))
RuntimeError: invalid doc topics format at line 3 in /tmp/4303cb_doctopics.txt
```

```bash
$ head -n 5 /tmp/4303cb_doctopics.txt
#doc name topic proportion ...
0	0	0	0.9378669927812985	9	0.006903667468744601	8	0.006903667468744601	7	0.006903667468744601	6	0.006903667468744601	5	0.006903667468744601	40.006903667468744601	3	0.006903667468744601	2	0.006903667468744601	1	0.006903667468744601	
1	1	0	1.0780612802674239	9	0.005575655428533364	8	0.005575655428533364	7	0.005575655428533364	6	0.005575655428533364	5	0.005575655428533364	30.005575655428533364	2	0.005575655428533364	1	0.005575655428533364	
2	2	0	0.9184413079632608	9	0.009062076892971008	8	0.009062076892971008	7	0.009062076892971008	6	0.009062076892971008	5	0.009062076892971008	40.009062076892971008	3	0.009062076892971008	2	0.009062076892971008	1	0.009062076892971008	
3	3	0	0.9184413079632608	9	0.009062076892971008	8	0.009062076892971008	7	0.009062076892971008	6	0.009062076892971008	5	0.009062076892971008	40.009062076892971008	3	0.009062076892971008	2	0.009062076892971008	1	0.009062076892971008	
```

piskvorky:
Looks like a bug introduced in e454e8c82767e776d97fbd1dafba39f14d048598 CC @jeiranj .

My guess is, in the old (2.0.7) sparse mallet format, not all topics are necessarily listed. Also seems to be true in @janrygl 's example `head` output.

The [original code for 2.0.7](
[...]

PR #664, opened 2016-04-11
error due to mallet 2.0.7 solved

RishabGoel:
purposing a solution for :  gensim==0.12.4 mallet wrapper is not compatible with mallet-2.0.7 #599

### DIFF
--- a/gensim/models/wrappers/ldamallet.py
+++ b/gensim/models/wrappers/ldamallet.py
@@ -55,7 +55,7 @@ class LdaMallet(utils.SaveLoad):
 
     """
     def __init__(self, mallet_path, corpus=None, num_topics=100, alpha=50, id2word=None, workers=4, prefix=None,
-                 optimize_interval=0, iterations=1000):
+                 optimize_interval=0, iterations=1000, topic_threshold=0.0):
         """
         `mallet_path` is path to the mallet executable, e.g. `/home/kofola/mallet-2.0.7/bin/mallet`.
 
@@ -71,6 +71,8 @@ class LdaMallet(utils.SaveLoad):
 
         `iterations` is the number of sampling iterations.
 
+        `topic_threshold` is the threshold of the probability above which we consider a topic. This is basically for sparse topic distribution.
+
         """
         self.mallet_path = mallet_path
         self.id2word = id2word
@@ -83,6 +85,7 @@ class LdaMallet(utils.SaveLoad):
         if self.num_terms == 0:
             raise ValueError("cannot compute LDA over an empty collection (no terms)")
         self.num_topics = num_topics
+        self.topic_threshold=topic_threshold
         self.alpha = alpha
         if prefix is None:
             rand_prefix = hex(random.randint(0, 0xffffff))[2:] + '_'
@@ -154,10 +157,10 @@ class LdaMallet(utils.SaveLoad):
         self.convert_input(corpus, infer=False)
         cmd = self.mallet_path + " train-topics --input %s --num-topics %s  --alpha %s --optimize-interval %s "\
             "--num-threads %s --output-state %s --output-doc-topics %s --output-topic-keys %s "\
-            "--num-iterations %s --inferencer-filename %s"
+            "--num-iterations %s --inferencer-filename %s --doc-topics-threshold %s"
         cmd = cmd % (
             self.fcorpusmallet(), self.num_topics, self.alpha, self.optimize_interval, self.workers,
-            self.fstate(), self.fdoctopics(), self.ftopickeys(), self.iterations, self.finferencer())
+            self.fstate(), self.fdoctopics(), self.ftopickeys(), self.iterations, self.finferencer(), self.topic_threshold)
         # NOTE "--keep-sequence-bigrams" / "--use-ngrams true" poorer results + runs out of memory
         logger.info("training MALLET LDA with %s", cmd)
         check_output(cmd, shell=True)
@@ -170,8 +173,8 @@ class LdaMallet(utils.SaveLoad):
             bow = [bow]
 
         self.convert_input(bow, infer=True)
-        cmd = self.mallet_path + " infer-topics --input %s --inferencer %s --output-doc-topics %s --num-iterations %s"
-        cmd = cmd % (self.fcorpusmallet() + '.infer', self.finferencer(), self.fdoctopics() + '.infer', iterations)
+        cmd = self.mallet_path + " infer-topics --input %s --inferencer %s --output-doc-topics %s --num-iterations %s --doc-topics-threshold %s"
+        cmd = cmd % (self.fcorpusmallet() + '.infer', self.finferencer(), self.fdoctopics() + '.infer', iterations, self.topic_threshold)
         logger.info("inferring topics with MALLET LDA '%s'", cmd)
         check_output(cmd, shell=True)
         result = list(self.read_doctopics(self.fdoctopics() + '.infer'))
--- a/gensim/models/wrappers/ldamallet.py
+++ b/gensim/models/wrappers/ldamallet.py
@@ -36,7 +36,7 @@ import os
 
 import numpy
 
-from bs4 import BeautifulSoup
+import xml.etree.ElementTree as et
 
 from six import iteritems
 from smart_open import smart_open
@@ -250,11 +250,9 @@ class LdaMallet(utils.SaveLoad):
     #########  function to check the version of mallet can be generalised to all the problematic version  ######
     def check_version(self, direc_path):
         xml_path = direc_path.split("bin")[0]
-        soup=BeautifulSoup(open(xml_path+"pom.xml").read())
-        if soup.find("version").text == "2.0.7-SNAPSHOT":
-            return True
-        else:
-            return False
+        doc = et.parse(xml_path+"pom.xml").getroot()
+        namespace = doc.tag[:doc.tag.index('}')+1]
+        return doc.find(namespace+'version').text
 
 
     def read_doctopics(self, fname, eps=1e-6, renorm=True):
@@ -262,7 +260,7 @@ class LdaMallet(utils.SaveLoad):
         Yield document topic vectors from MALLET's "doc-topics" format, as sparse gensim vectors.
 
         """
-        bool_v7 = self.check_version(self.mallet_path)
+        mallet_version = self.check_version(self.mallet_path)
         with utils.smart_open(fname) as fin:
             for lineno, line in enumerate(fin):
                 if lineno == 0 and line.startswith("#doc "):
@@ -282,7 +280,7 @@ class LdaMallet(utils.SaveLoad):
                            for id_, weight in enumerate(map(float, parts))
                            if abs(weight) > eps]
                 else:
-                    if bool_v7:
+                    if mallet_version == "2.0.7-SNAPSHOT":
                         # 1   1   0   1.0780612802674239  9   0.005575655428533364    8   0.005575655428533364    7   0.005575655428533364    6   0.005575655428533364    5   0.005575655428533364    30.005575655428533364   2   0.005575655428533364    1   0.005575655428533364    
                         # 2   2   0   0.9184413079632608  9   0.009062076892971008    8   0.009062076892971008    7   0.009062076892971008    6   0.009062076892971008    5   0.009062076892971008    40.009062076892971008   3   0.009062076892971008    2   0.009062076892971008    1   0.009062076892971008
                         # In the above example there is a mix of the above if and elif statement. There are neither 2*num_topics nor num_topics elements.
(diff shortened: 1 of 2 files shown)


gensim/models/wrappers/ldamallet.py
```diff
@@ -36,6 +36,8 @@
 
 import numpy
 
+from bs4 import BeautifulSoup
```
tmylk: Using Beautiful soup for parsing correct XML is an overkill. Please use `from xml.etree import ElementTree`

gensim/models/wrappers/ldamallet.py
```diff
@@ -245,11 +247,22 @@ def show_topic(self, topicid, topn=10):
     def print_topic(self, topicid, topn=10):
         return ' + '.join(['%.3f*%s' % v for v in self.show_topic(topicid, topn)])
 
+    #########  function to check the version of mallet can be generalised to all the problematic version  ######
+    def check_version(self, direc_path):
+        xml_path = direc_path.split("bin")[0]
+        soup=BeautifulSoup(open(xml_path+"pom.xml").read())
+        if soup.find("version").text == "2.0.7-SNAPSHOT":
+            return True
```
tmylk: Please return version number. There is no problematic version. Both versions are valid.

gensim/models/wrappers/ldamallet.py
```diff
@@ -245,11 +247,20 @@ def show_topic(self, topicid, topn=10):
     def print_topic(self, topicid, topn=10):
         return ' + '.join(['%.3f*%s' % v for v in self.show_topic(topicid, topn)])
 
+    #########  function to return the version of mallet   ######
```
tmylk: Please use Gensim style of doc-strings

gensim/models/wrappers/ldamallet.py
```diff
@@ -245,11 +247,20 @@ def show_topic(self, topicid, topn=10):
...
+    def get_version(self, direc_path):
+        xml_path = direc_path.split("bin")[0]
+        doc = et.parse(xml_path+"pom.xml").getroot()
```
tmylk: The file might not exist. It is possilbe to run Mallet with just a jar. 
Please account for this situation.

gensim/models/wrappers/ldamallet.py
```diff
@@ -269,7 +280,28 @@ def read_doctopics(self, fname, eps=1e-6, renorm=True):
                            for id_, weight in enumerate(map(float, parts))
                            if abs(weight) > eps]
                 else:
-                    raise RuntimeError("invalid doc topics format at line %i in %s" % (lineno + 1, fname))
+                    if mallet_version == "2.0.7-SNAPSHOT":
```
tmylk: The version just needs to be 2.0.7 . No need for SNAPSHOT requirement

gensim/models/wrappers/ldamallet.py
```diff
@@ -269,7 +280,28 @@ def read_doctopics(self, fname, eps=1e-6, renorm=True):
...
+                        # 1   1   0   1.0780612802674239  9   0.005575655428533364    8   0.005575655428533364    7   0.005575655428533364    6   0.005575655428533364    5   0.005575655428533364    30.005575655428533364   2   0.005575655428533364    1   0.005575655428533364    
+                        # 2   2   0   0.9184413079632608  9   0.009062076892971008    8   0.009062076892971008    7   0.009062076892971008    6   0.009062076892971008    5   0.009062076892971008    40.009062076892971008   3   0.009062076892971008    2   0.009062076892971008    1   0.009062076892971008
```
tmylk: Thank you for giving an explanation for this behaviour.
Can you give a shorter example of these 2 formats? Say with num_topics = 3. It will be easier to understand the explanation.

gensim/models/wrappers/ldamallet.py
RishabGoel: So, in case of a jar file should I still check for version from its manifest file, by first decomposing it or not?

tmylk: Decompressing sounds great

RishabGoel: I have a doubt in case of jar, Can you tell me how the path to mallet dir would look like.

gensim/models/wrappers/ldamallet.py
```diff
@@ -269,7 +289,30 @@ def read_doctopics(self, fname, eps=1e-6, renorm=True):
...
+                            It has 2 formats 40.009062076892971008 and 0   1.0780612802674239 which cannot be handled by above if elif.
+                            Also, there are some topics are missing(meaning that the topic is not there) which is another reason why the above if elif
+                            fails even when the `mallet` produces the right results
+
+                        """
+                        pointer = 0
+                        doc = []
+                        while pointer < len(parts):
+                            if float(parts[pointer]) == int(parts[pointer]):
+                                if float(parts[pointer+1]) > eps:
```
piskvorky: Looking at the `while` condition, this `[pointer + 1]` access could result in out-of-bounds `IndexError`.

Also, the logic around these `if`s deserves some comments, it's not clear what's happening and why.

gensim/models/wrappers/ldamallet.py
```diff
@@ -269,7 +289,30 @@ def read_doctopics(self, fname, eps=1e-6, renorm=True):
...
+                                    doc.append((int(parts[pointer]), float(parts[pointer+1])))
+                                pointer+=2
+                            else:
+                                if float(parts[pointer])-int(parts[pointer]) > eps:
+                                    doc.append((int(parts[pointer])%10, float(parts[pointer])-int(parts[pointer])))
+                                pointer+=1
+                    else:
+                        raise RuntimeError("invalid doc topics format at line %i in %s" % (lineno + 1, fname))
```
piskvorky: We count from 0, Python-style.

gensim/models/wrappers/ldamallet.py
```diff
@@ -269,7 +289,30 @@ def read_doctopics(self, fname, eps=1e-6, renorm=True):
...
+                        doc = []
+                        while pointer < len(parts):
+                            if float(parts[pointer]) == int(parts[pointer]):
+                                if float(parts[pointer+1]) > eps:
+                                    doc.append((int(parts[pointer]), float(parts[pointer+1])))
+                                pointer+=2
+                            else:
+                                if float(parts[pointer])-int(parts[pointer]) > eps:
+                                    doc.append((int(parts[pointer])%10, float(parts[pointer])-int(parts[pointer])))
+                                pointer+=1
```
piskvorky: Does this `if` branch logic mean that some parts of one line can be parsed one way, and another parts another way? I don't think that can happen -- the same format is surely used throughout the entire file?

gensim/models/wrappers/ldamallet.py
```diff
@@ -245,11 +247,29 @@ def show_topic(self, topicid, topn=10):
...
+
+    def get_version(self, direc_path):
+        """"
+
+        function to return the version of `mallet`
+
+        """
+        xml_path = direc_path.split("bin")[0]
+        try:
+            doc = et.parse(xml_path+"pom.xml").getroot()
```
piskvorky: Code style: put a space around binary operators, here and elsewhere in this PR (PEP8).

gensim/models/wrappers/ldamallet.py
```diff
@@ -245,11 +247,29 @@ def show_topic(self, topicid, topn=10):
...
+            namespace = doc.tag[:doc.tag.index('}')+1]
+            return doc.find(namespace+'version').text.split("-")[0]
+        except Exception, e:
```
piskvorky: No need to specify `, e` if we're not using it.

gensim/models/wrappers/ldamallet.py
RishabGoel: Yeah, if see the results of the head -n 5 ... on https://github.com/piskvorky/gensim/issues/599. In first and second line there are some topics missing as well as the format is different.

gensim/models/wrappers/ldamallet.py
RishabGoel: ok, on it

gensim/models/wrappers/ldamallet.py
RishabGoel: Correct me if I am wrong. I came up with this logic after seeing that only. I will add comments to make it  more clear.

gensim/models/wrappers/ldamallet.py
RishabGoel: I dont think so. I am adding comments for this if logic. Except when there are no topics for a document, it will work fine.

piskvorky: If I read this branch correctly, then if the last number on a line is an int, this line will raise an `IndexError`.

RishabGoel: I have added the boundary condition as well.

gensim/models/wrappers/ldamallet.py
RishabGoel: Done

gensim/models/wrappers/ldamallet.py
RishabGoel: Yes, it will. But consider the following:
There are only 2 formats possible namely 20.2345 or 2 0.2345, right
in the former case the while block will go to the else condition and in latter case it will go to if section. The format type produced by mallet is a kind of security for this error.

gensim/models/wrappers/ldamallet.py
tmylk: Folder called 'regression' only exists in 2.0.8

RishabGoel:
@janrgyl : Could you post the parameters you used that lead to issue #599.
Thanks

RishabGoel:
@cscorley Please review the fix in this PR. I have also sent you an email about the same with other details.

gensim/models/wrappers/ldamallet.py
```diff
@@ -245,11 +248,40 @@ def show_topic(self, topicid, topn=10):
...
+        """"
+
+        function to return the version of `mallet`
+
+        """
+        try:
+            """
+            Check version of mallet via jar file
+            """
+            archive = zipfile.ZipFile(direc_path, 'r')
```
tmylk: Better to use smart_open

gensim/models/wrappers/ldamallet.py
```diff
@@ -245,11 +248,40 @@ def show_topic(self, topicid, topn=10):
...
+                return '2.0.8RC3'
+        except Exception:
+            
+            xml_path = direc_path.split("bin")[0]
+            try:
+                doc = et.parse(xml_path + "pom.xml").getroot()
+                namespace = doc.tag[:doc.tag.index('}') + 1]
+                return doc.find(namespace + 'version').text.split("-")[0]
+            except Exception:
+                return "No version file to detect"
```
tmylk: Correct message would be "Can't parse pom.xml version file"

gensim/models/wrappers/ldamallet.py
```diff
@@ -245,11 +248,40 @@ def show_topic(self, topicid, topn=10):
...
+        
+
+
     def read_doctopics(self, fname, eps=1e-6, renorm=True):
         """
         Yield document topic vectors from MALLET's "doc-topics" format, as sparse gensim vectors.
 
         """
+        mallet_version = self.get_version(self.mallet_path)
```
tmylk: Now that we know the version we can remove the version guessing based on pattern from this old if
https://github.com/piskvorky/gensim/blob/develop/gensim/models/wrappers/ldamallet.py#L260

RishabGoel: Thanks for pointing them out. I will do all of these changes and update the PR.

RishabGoel: Also should I remove the checking in if elif logic as below with the version check?
`if len(parts) == 2*self.num_topics:`
`elif len(parts) == self.num_topics:`

tmylk: If you can, then yes. 
But both versions can give dense output not sure how it will help with `if len(parts) == 2*self.num_topics:`

gensim/models/wrappers/ldamallet.py
piskvorky: `smart_open` doesn't handle zip files (=multiple files in a single archive). It only handles file compression, not multi-file archives.

RishabGoel: Yes, according to the following eg in the smart_open repo's read me:
for line in smart_open.smart_open('./foo.txt.gz'):
...    print line

We can only  unzip a file (say a txt file).
I think zipfile is OK. If there not then I will look for another library to handle jar file.

syting:
Testing this with Mallet 2.0.8RC3 it seems that when topic_threshold is set the output reverts back to topic/proportion format from the dense format.

RishabGoel:
That seems to be correct. Thanks for pointing out @syting . I have taken care of that.

PR #677, opened 2016-04-20
Added Document-Topic Proportion in DTM tutorial.

bhargavvader:
@tmylk  Added a few lines on how to print Document-Topic proportions in the tutorial, something which would commonly be used but wasn't mentioned how to do before.

It's also not an explicit function and you have to do some digging in the wrapper code comments to figure out how to do so, so I thought this would be useful.

PR #679, opened 2016-04-21
[WIP] DTM tutorial additions

bhargavvader:
Made PEP8 style space changes to DTM tutorial notebook, wrt comments on PR #677.


bhargavvader:
@tmylk , I also wanted to add a bit on using DIM in this tutorial, so please keep this PR open.

bhargavvader:
@tmylk , could you please have a look? I've added instructions for finding document influence for any particular time-slice, document and topic . I'm not sure if much else could be done here. 

I am a little unsure if I should run the model twice, as I've done in the tutorial, instead of just running it once with `model='fixed'` in the beginning. I thought doing it twice would be useful in describing the differences in DTM, DIM though.

tmylk:
Running twice is ok if you can explain the difference.

Could you add more explanation of the advantages of DIM? Runtime of DIM vs DTM, different output, link to potential uses.

bhargavvader:
@tmylk , I've added more information about DIM and it's use-case. About run-time of DIM, how exactly do you want it explained? Should I use `time.time()` to check it - and if so, for how big a corpus? It certainly takes longer than DTM, but I'm unsure how exactly to quantify it.

bhargavvader:
@tmylk , can you have a look and suggest any more changes?

Mowjo:
Hi bhargavvader,

I've been learning to use DTM through your tutorial. The 'dtm' model of the D.Blei implementation  wrapper works perfectly. But when it comes to DIM (model =  'fixed') I have an error like this after 2minutes running  : 

CalledProcessError: Command '['dtm-win64.exe', '--ntopics=10', '--model=fixed', '--mode=fit', '--initialize_lda=true', '--corpus_prefix=c:\\users\\carol\\appdata\\local\\temp\\d30a57_train', '--outname=c:\\users\\carol\\appdata\\local\\temp\\d30a57_train_out', '--alpha=0.01', '--lda_max_em_iter=10', '--lda_sequence_min_iter=6', '--lda_sequence_max_iter=20', '--top_chain_var=0.005', '--rng_seed=0']' returned non-zero exit status 255

Do you have any idea how to overcome this error ? By the way I'm using the DTM binaries files for Windows. 

Kind Regards.

bhargavvader:
@Mowjo , do you happen to have a document with no words in your corpus? While DTM works fine with empty documents, DIM fails. 

This [link](https://lists.cs.princeton.edu/pipermail/topic-models/2014-May/002786.html) talks about this kind of error a little more when using the original Blei code.

Mowjo:
@bhargavvader  Oh I see. Well I use the .dopna() function to remove empty documents in my Dataframe but it might be an error just like what you said. I'll have a look closer on it. 
Thank you a lot !

PR #681, opened 2016-04-24
Stops considering repeated words as a potential combined keywords

### LINKED ISSUE #667: KeyError: Popping a non-existing value from `_keywords`
OS X 10.11.1
Python 2.7.10
gensim 0.12.4

To reproduce this in Python shell:
```
>>> import gensim.summarization
>>> t = "Victor S. Sage Compare Sage 50c Editions Find accounting software that's right for your business Every product comes with anytime, anywhere online access; automatic updates; access to unlimited support; access to built-in credit card processing and payroll; and advanced reporting. Three solutions for your business 1 user From $249/year Buy now Free Trial 1-5 users From $299/year Buy now Free Trial 3-40 users From $1,199/year Buy now Free Trial Essential Accounting Accounts payable, accounts receivable, cash management check check check open check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check check Advanced Accounting Automated tasks, audit trail, budgeting, change order processing check check open check check check check check check check check check check check check check check check check check check check check check check check check check check check check In-depth Accounting Fast processing, industry-specific features, workflow management check open check check check check check check check check check check check Disclaimers open * This product is backed by a no-risk guarantee for first-time Sage 50 customers. If, within 60 days of purchase, you are not convinced that Sage 50 is the best accounting program for your business, we will refund your money (less and rebate you have received for this purchase). Dated proof of purchase and return of product is required. For details, call 877-481-0341."
>>> import gensim.summarization
>>> keywords = gensim.summarization.keywords(t, pos_filter=[], ratio=0.2, lemmatize=True, scores=True)
  File "/usr/local/lib/python2.7/site-packages/gensim/summarization/keywords.py", line 229, in keywords
    combined_keywords = _get_combined_keywords(keywords, text.split())
  File "/usr/local/lib/python2.7/site-packages/gensim/summarization/keywords.py", line 171, in _get_combined_keywords
    _keywords.pop(keyword)
KeyError: u'check'
```

jhedin:
fixes #667, with the assumption that it isn't meaningful to have two of the same word in a combined keyword.


tmylk:
@jhedin Thanks. Looks good. 
Please add a new test with the example from #667 to https://github.com/piskvorky/gensim/blob/develop/gensim/test/test_keywords.py

jhedin:
I added it as a simple test to see if it runs; should I try and figure out what the expected output of it is as well, and check that?

phdowling:
What about actual repeated tokens that are real concepts? E.g. "bora bora" or "piri piri", etc.

gensim/test/test_summarization.py
```diff
@@ -130,6 +130,21 @@ def test_corpus_summarization_ratio(self):
 
             self.assertEqual(len(selected_docs), expected_summary_length)
 
+    def test_repeated_keywords(self):
+        pre_path = os.path.join(os.path.dirname(__file__), 'test_data')
+
+        with utils.smart_open(os.path.join(pre_path, "testrepeatedkeywords.txt")) as f:
+            text = f.read()
```
tmylk: Thanks. Could you please add the txt file into the test data folder?

jhedin: Sorry! I forgot to track it!

jhedin:
@phdowling: those are great counter examples! As implemented, it would also miss good combined keywords like "Bora Bora Beaches" and instead go for "Bora", " Beaches"

I had considered cases where each token has a different meaning like Buffalo buffalo, but not ones where the combined keywords have a meaning different from its components(I couldn't think of any)

Can TextRank actually discern those cases? With the case of true homonyms, I figure it can't decide that one meaning is more important than the other, or that both meanings are important. In the buffalo case, the text might be referring to the animal for most of it, but has "Buffalo buffalo" once, referring to bison from Buffalo. "Buffalo buffalo" is not as descriptive a keyword as "buffalo" in that case.

Maybe a whitelist would be appropriate? 

This function isn't very robust on the whole; I am imagining that it should be redone in a hierarchical manner, and choose the combination with the best rank, rather than combining the first keywords that it sees with good ranks.

piskvorky:
Let's loop in the module's authors, @fbarrios @fedelopez77 :)

tmylk:
Thanks for the PR!

PR #689, opened 2016-05-03
Replace num_topics with self.num_topics

himkt:
I found self.num_topics is casted to Integer in [here](https://github.com/piskvorky/gensim/blob/develop/gensim/models/ldamodel.py#L275).

So I think it's better to use self.num_topics instead of num_topics.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -285,13 +285,13 @@ class LdaModel(interfaces.TransformationABC):
 
         self.alpha, self.optimize_alpha = self.init_dir_prior(alpha, 'alpha')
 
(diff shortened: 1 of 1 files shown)


himkt:
In print_topics, num_topics and self.num_topics are clearly different.
But I think these are the same in __init__, so I made this PR.

If I have a misunderstanding about above, I'm sorry but please reject this PR.

tmylk:
Thanks for the PR. Looks good!

piskvorky:
Also, congrats on being the 100th gensim contributor, @himkt! :)

himkt:
Oh that's true, thanks to celebrate 😄

PR #699, opened 2016-05-12
[WIP] Implement save_word2vec_format can for Doc2Vec 

cedias:
Hi,
I recently had to export my Doc2Vec model to W2V base format and the function save_word2vec_format wasn't implemented for D2V class and was simply calling W2V one.

Therefore I quickly made this implementation.

If you believe it's worthwhile I'll go ahead and implement the load function to properly test the save/load pipeline, let me know.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -722,6 +722,33 @@ class Doc2Vec(Word2Vec):
         report['doctag_lookup'] = self.docvecs.estimated_lookup_memory()
         report['doctag_syn0'] = self.docvecs.count * self.vector_size * dtype(REAL).itemsize
         return super(Doc2Vec, self).estimate_memory(vocab_size, report=report)
+    
+    def save_word2vec_format(self, fname, dfname=None, fvocab=None, binary=False):
+        """
+        Store the input-hidden weight matrix in the same format used by the original
+        C word2vec-tool, for compatibility.
+
+         `fname` is the file used to save the words vectors in
+         `dfname` is an optional file used to save the documents vectors in
+         `fvocab` is an optional file used to save the vocabulary
+         `binary` is an optional boolean indicating whether the data is to be saved
+         in binary word2vec format (default: False)
+
+        """
+        super(Doc2Vec,self).save_word2vec_format(fname, fvocab, binary)
+
+        if dfname is not None:
+            logger.info("storing %sx%s projection weights into %s" % (len(self.docvecs), self.vector_size, dfname))
+            assert (len(self.docvecs), self.vector_size) == self.docvecs.doctag_syn0.shape
+            with utils.smart_open(dfname, 'wb') as fout:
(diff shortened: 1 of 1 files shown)


tmylk:
@cedias Apologies for late reply. What will be the advantage of saving in this format? What is your use case?

cedias:
Well, since `Doc2Vec` extends `Word2Vec`, the `save_word2vec_format` function can already be used to output the word vectors. I believe it's logical to be able to output the document vector as well (At least that is what I was expecting before noticing it was only writing word vectors).

For the `save_word2vec_format` function the main (and probably only) advantage of this format is legacy. Personnally, I had some code which took classic w2v binary output as input and I wanted to try and input document vectors.

As for the load function i'm not really sure about the use case, besides API consistency, which is why I wanted an opinion about it :)

gojomo:
I can see this being useful for others as well. Some thoughts:

* should remain possible to also save word-vecs – perhaps even to the same file? – so superclass functionality shouldn't be completely hidden by the override
* should also support case where user's doctags are just plain int indexes rather than strings (and thus `model.docvecs.doctags` is empty), or a mixture of ints and strings
* load... makes sense as well, subject to the same above points

Achieving those may require some new conventions in the method API and on-disk format to indicate the word-vec/doc-vec distinction... if at all possible those conventions should put a minimal burden on people using older files, other tools, or just one set (word-vecs or doc-vecs) of vectors.

cedias:
Ok, i'll work on it later this month then. Thanks for the tips.

tmylk:
Hi @cedias Would you have time to work on this for our release this month?

cedias:
I believe this feature will be removed/realized in #1107

tmylk:
@cedias Thanks for following the github development. 
The feature that you are proposing is a new feature and is not included in #1107.
That PR doesn't touch on `model.docvecs` and existing behaviour of simply calling `Word2Vec.save_word2vec_format` is preserved.

gojomo:
If Doc2Vec's `DocvecsArray` is fully adapted to use KeyedVectors, it might then inherit a useful `save_word2vec_format()` implementation. 

Plausibly as per my [earlier comment](https://github.com/RaRe-Technologies/gensim/pull/699#issuecomment-224943854), there could be key-munging conventions for mixing word & doc vectors into the same flattened file on save, or even disentangling them on load. (Some downstream applications might like them mixed-together.)

parulsethi:
Another use case - I wanted to visualize docvecs in Tensorboard which require vectors to be in text file format, and this functionality would be useful for that.

menshikh-iv:
Ping @cedias, what status of this PR? Will you finish it soon?

parulsethi:
@menshikh-iv This feature was added in https://github.com/RaRe-Technologies/gensim/pull/1256.

menshikh-iv:
Then I close this PR, I hope the author agrees with @parulsethi (If not, you can reopen it)

Issue #661, opened 2016-04-06
TypeError: startswith first arg must be bytes or a tuple of bytes, not str


GSingh2014:
OS X 10.10.5
Python 3.5
gensim 0.12.4

Running the saved LDAMallet model on new documents is giving the error
2016-04-06 11:33:47,410 : INFO : **loading Dictionary object from /gensim_Mallet/gensim_Mallet_LDA.dict**
2016-04-06 11:33:47,411 : **INFO : loading LdaMallet object from /gensim_Mallet/gensim_Mallet_LDA.model**
2016-04-06 11:33:47,411 : INFO : loading id2word recursively from /gensim_Mallet/gensim_Mallet_LDA.model.id2word.* with mmap=None
bow created
[(0, 5), (3, 8), (7, 99), (9, 6), (14, 1), (22, 1), (60, 3), (79, 1), (80, 1), (121, 3), (163, 2), (189, 1), (204, 6), (210, 5), (214, 6), (215, 5793)]
2016-04-06 11:33:47,424 : INFO : serializing temporary corpus to /var/folders/50/dx53qks96x54d05lq4l5mqxr0000gn/T/134f2a_corpus.txt
2016-04-06 11:33:47,425 : INFO : converting temporary corpus to MALLET format with /usr/local/Cellar/mallet-2.0.6/bin/mallet import-file --preserve-case --keep-sequence --remove-stopwords --token-regex '\S+' --input /var/folders/50/dx53qks96x54d05lq4l5mqxr0000gn/T/134f2a_corpus.txt --output /var/folders/50/dx53qks96x54d05lq4l5mqxr0000gn/T/134f2a_corpus.mallet.infer --use-pipe-from /var/folders/50/dx53qks96x54d05lq4l5mqxr0000gn/T/134f2a_corpus.mallet
2016-04-06 11:33:47,894 : INFO : inferring topics with MALLET LDA '/usr/local/Cellar/mallet-2.0.6/bin/mallet infer-topics --input /var/folders/50/dx53qks96x54d05lq4l5mqxr0000gn/T/134f2a_corpus.mallet.infer --inferencer /var/folders/50/dx53qks96x54d05lq4l5mqxr0000gn/T/134f2a_inferencer.mallet --output-doc-topics /var/folders/50/dx53qks96x54d05lq4l5mqxr0000gn/T/134f2a_doctopics.txt.infer --num-iterations 100'
Traceback (most recent call last):
[...]

PR #702, opened 2016-05-18
#661 bugfix

jagun:
fixed literal from string to binary string


tmylk:
Thank you for the bug-fix!

Issue #683, opened 2016-04-25
API call for Topic Distribution of words.


bhargavvader:
As discussed [here](https://groups.google.com/forum/#!topic/gensim/hgjQPqP-0JU) and [here](https://groups.google.com/forum/#!searchin/gensim/how$20to$20print$20topic$20distribution%7Csort:relevance/gensim/Vv7SSC8HR8k/b4knKmArX0MJ) in the mailing list, there is no current way to find the a `topic index` for a word - i.e the topic a word belongs to - basically the distribution over all topics for a single word. 

As @piskvorky mentioned in the second link, the ideal way now is:

"With variational inference LDA (such as the implementation in gensim), you can get the per-word topic distribution from `phi`, one of the variational parameters. When you call `lda.inference()` with `collect_sstats=True`, it will return a 2-tuple of (gamma, sufficient statistics). In your single-word case, these sufficient statistics directly correspond to (already normalized) phi, so you can read the word topics directly off of that."

[This](https://groups.google.com/forum/#!topic/gensim/Qoj7Agkx3qE) also talks about the same, and describes the problem/what is needed in simpler words.

I'll open a PR soon for this, and would like to take suggestions- would it be as a method in the `ldamodel` class, with just takes a word as a parameter, which uses VB internal parameters and gives back the topic and probability, or is there a different approach?

piskvorky:
I haven't given this API enhancement a deep thought, but I think a good way may be to add a new, optional parameter: `get_document_topics(per_word_topics=True)`.

- The default would be `per_word_topics=False`, so nothing changes unless the user specifically asks for those per-word topic assignments;
- but if `per_word_topics=True`, the method returns a 2-tuple: `(per-topic-probability-for-this-document, per-wordtype-best-topic-for-this-document)` (instead of just the normal `per-topic-probability-for-this-document` output).

The second link in your description seems unrelated -- it asks for static `topic x vocab` assignments, not `topic x document words` assignment (the same word can be assigned to different topics for different documents, depending on other words in the document -- the context matters).

bhargavvader:
I'm still a little confused, because the initial question says:

" want to assign a topic index to each word in the corpus."

Can that not mean the vocabulary? Or does it mean every word of document in the corpus? If it is certainly the latter and it is like you said the dynamic `topic x document` words assignment, then your suggestion does seem like the way to go.

piskvorky:
[...]

PR #704, opened 2016-05-22
[MRG] Per Word Topic in a Document + Notebook Tutorial 

bhargavvader:
With reference to #683 .
@piskvorky , @tmylk , as you suggested, if `per_word_topics=True`, it returns a 2-tuple with `(per-topic-probability-for-this-document, per-wordtype-best-topic-for-this-document)`. 

It runs fine for the test cases on my local machine, and if the logic and approach looks fine I can write tests for it.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -883,7 +883,7 @@ class LdaModel(interfaces.TransformationABC):
         top_topics = sorted(coherence_scores, key=lambda t: t[1], reverse=True)
         return top_topics
 
-    def get_document_topics(self, bow, minimum_probability=None):
+    def get_document_topics(self, bow, minimum_probability=None, per_word_topics=False):
         """
         Return topic distribution for the given document `bow`, as a list of
         (topic_id, topic_probability) 2-tuples.
@@ -899,11 +899,24 @@ class LdaModel(interfaces.TransformationABC):
         is_corpus, corpus = utils.is_corpus(bow)
         if is_corpus:
             return self._apply(corpus)
+        if per_word_topics is False:
+            gamma, _ = self.inference([bow])
+            topic_dist = gamma[0] / sum(gamma[0])  # normalize distribution
+            return [(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
+                    if topicvalue >= minimum_probability]
 
-        gamma, _ = self.inference([bow])
-        topic_dist = gamma[0] / sum(gamma[0])  # normalize distribution
-        return [(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
-                if topicvalue >= minimum_probability]
+        if per_word_topics is True:
+            gamma, _ = self.inference([bow])
+            topic_dist = gamma[0] / sum(gamma[0])  # normalize distribution
+
+            word_phi = []
+            for word, weight in bow:
+                phi_values = []
+                for i in range(0, self.num_topics):
+                    phi_values.append(self.inference([bow], collect_sstats=True)[1][i][word])
+                word_phi.append((word, phi_values.index(max(phi_values))))
+            return ([(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
+                    if topicvalue >= minimum_probability], word_phi)               
 
     def __getitem__(self, bow, eps=None):
         """
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -899,21 +899,20 @@ class LdaModel(interfaces.TransformationABC):
         is_corpus, corpus = utils.is_corpus(bow)
         if is_corpus:
             return self._apply(corpus)
+
+        gamma, phis = self.inference([bow], collect_sstats=True)
+        topic_dist = gamma[0] / sum(gamma[0])  # normalize distribution
+
         if per_word_topics is False:
-            gamma, _ = self.inference([bow])
-            topic_dist = gamma[0] / sum(gamma[0])  # normalize distribution
             return [(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
                     if topicvalue >= minimum_probability]
 
-        if per_word_topics is True:
-            gamma, _ = self.inference([bow])
-            topic_dist = gamma[0] / sum(gamma[0])  # normalize distribution
-
+        elif per_word_topics is True:
             word_phi = [] # contains word and corresponding topic
             for word, weight in bow:
                 phi_values = [] # contains phi values for each topic
                 for i in range(0, self.num_topics):
-                    phi_values.append(self.inference([bow], collect_sstats=True)[1][i][word]) 
+                    phi_values.append(phis[i][word]) 
                     # appends phi values for each topic for that word
                 word_phi.append((word, phi_values.index(max(phi_values))))
                 # appends the word and the highest probability topic
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -907,10 +907,8 @@ class LdaModel(interfaces.TransformationABC):
 
         document_topics = [(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
                     if topicvalue >= minimum_probability]
-
         if not per_word_topics:
             return document_topics
-
         else:
             word_phi = [] # contains word and corresponding topic
             for word_type, weight in bow:
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -909,13 +909,15 @@ class LdaModel(interfaces.TransformationABC):
 
         elif per_word_topics is True:
             word_phi = [] # contains word and corresponding topic
-            for word, weight in bow:
+            for word_type, weight in bow:
                 phi_values = [] # contains phi values for each topic
-                for i in range(0, self.num_topics):
-                    phi_values.append(phis[i][word]) 
+                for topic_id in range(0, self.num_topics):
+                    if phis[topic_id][word_type] >= minimum_probability:
+                        phi_values.append((topic_id, phis[topic_id][word_type]))
                     # appends phi values for each topic for that word
-                word_phi.append((word, phi_values.index(max(phi_values))))
-                # appends the word and the highest probability topic
+                word_phi.append((word_type, phi_values))
+                # appends the word and the corresponding values for each topic.
+                # If we choose the highest value for a word, we get the corresponding most probable topic.
             return ([(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
                     if topicvalue >= minimum_probability], word_phi) # returns 2-tuple
 
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -890,6 +890,8 @@ class LdaModel(interfaces.TransformationABC):
 
         Ignore topics with very low probability (below `minimum_probability`).
 
+        If per_word_topics is True, it also returns a list of topics, sorted in descending order of most likely topics for that word. 
+
         """
         if minimum_probability is None:
             minimum_probability = self.minimum_probability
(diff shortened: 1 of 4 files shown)


gensim/test/test_ldamodel.py
```diff
@@ -251,20 +251,31 @@ def testShowTopics(self):
...
         for k, v in doc_topics:
             self.assertTrue(isinstance(k, int))
             self.assertTrue(isinstance(v, float))
 
+        for w, t in word_topics:
+            self.assertTrue(isinstance(w, int))
+            self.assertTrue(isinstance(t, int))
+
+        # first word in the doc belongs to topic 2
+        self.assertEqual(word_topics[0], (0, 1))
```
tmylk: Why is it a tuple `(0,1)`? Expected `1`

bhargavvader: Tuple `(0,1)` basically means 0th word corresponds to 1st topic. 
Should I change it to `word_topics[0][1]` equals to 1?

bhargavvader: @tmylk changed it. Does it make more sense?

bhargavvader:
@tmylk , @piskvorky , could you please review?

gensim/models/ldamodel.py
```diff
@@ -900,10 +900,26 @@ def get_document_topics(self, bow, minimum_probability=None):
         if is_corpus:
             return self._apply(corpus)
 
-        gamma, _ = self.inference([bow])
+        gamma, phis = self.inference([bow], collect_sstats=True)
         topic_dist = gamma[0] / sum(gamma[0])  # normalize distribution
-        return [(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
-                if topicvalue >= minimum_probability]
+
+        if per_word_topics is False:
```
piskvorky: `if not per_word_topics` more Pythonic.

gensim/models/ldamodel.py
```diff
@@ -900,10 +900,26 @@ def get_document_topics(self, bow, minimum_probability=None):
...
+            return [(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
+                    if topicvalue >= minimum_probability]
+
+        elif per_word_topics is True:
```
piskvorky: `else:` better.

gensim/models/ldamodel.py
```diff
@@ -900,10 +900,26 @@ def get_document_topics(self, bow, minimum_probability=None):
...
+            word_phi = [] # contains word and corresponding topic
+            for word_type, weight in bow:
+                phi_values = [] # contains phi values for each topic
+                for topic_id in range(0, self.num_topics):
+                    if phis[topic_id][word_type] >= minimum_probability:
+                        phi_values.append((topic_id, phis[topic_id][word_type]))
+                    # appends phi values for each topic for that word
```
piskvorky: Comment before code block, not after.

gensim/models/ldamodel.py
```diff
@@ -900,10 +900,26 @@ def get_document_topics(self, bow, minimum_probability=None):
...
+                word_phi.append((word_type, phi_values))
+                # appends the word and the corresponding values for each topic.
+                # If we choose the highest value for a word, we get the corresponding most probable topic.
+            return ([(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
```
piskvorky: Better to keep the logic for topic distribution in one place (DRY), and re-use in both `if` branches.

gensim/models/ldamodel.py
```diff
@@ -883,7 +883,7 @@ def top_topics(self, corpus, num_words=20):
         top_topics = sorted(coherence_scores, key=lambda t: t[1], reverse=True)
         return top_topics
 
-    def get_document_topics(self, bow, minimum_probability=None):
+    def get_document_topics(self, bow, minimum_probability=None, per_word_topics=False):
```
piskvorky: Missing docs for `per_word_topics` parameter.

gensim/models/ldamodel.py
```diff
@@ -900,10 +900,26 @@ def get_document_topics(self, bow, minimum_probability=None):
...
+                phi_values = [] # contains phi values for each topic
+                for topic_id in range(0, self.num_topics):
+                    if phis[topic_id][word_type] >= minimum_probability:
+                        phi_values.append((topic_id, phis[topic_id][word_type]))
+                    # appends phi values for each topic for that word
+                word_phi.append((word_type, phi_values))
+                # appends the word and the corresponding values for each topic.
+                # If we choose the highest value for a word, we get the corresponding most probable topic.
+            return ([(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
+                    if topicvalue >= minimum_probability], word_phi) # returns 2-tuple
```
piskvorky: The return format of `word_phi` is too obscure, should be more user-friendly.

Either return only the top-topic for each word (`{word_id => topic_id}` map), or maybe return all, but sorted by probability (`{word_id => [topic_id_most_probable, topic_id_second_most_probable, ...]`).

gensim/models/ldamodel.py
```diff
@@ -900,10 +900,26 @@ def get_document_topics(self, bow, minimum_probability=None):
...
+            return [(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
+                    if topicvalue >= minimum_probability]
+
+        elif per_word_topics is True:
+            word_phi = [] # contains word and corresponding topic
+            for word_type, weight in bow:
+                phi_values = [] # contains phi values for each topic
+                for topic_id in range(0, self.num_topics):
+                    if phis[topic_id][word_type] >= minimum_probability:
+                        phi_values.append((topic_id, phis[topic_id][word_type]))
```
piskvorky: @tmylk this needs a deeper check -- are the phis directly comparable like this? Aren't they scaled by word count / feature weight?

gensim/models/ldamodel.py
bhargavvader: @tmylk and I were discussing this too... What would you think would be the better way to go ahead? In my first version of this PR I was only returning the most probable topic.

gensim/models/ldamodel.py
bhargavvader: Yup, will do.

gensim/models/ldamodel.py
piskvorky: What I think is not terribly important... it'd be more valuable to ask the users who asked for this feature.

When I needed this functionality, I needed to highlight words from different topics in different colours, for visualization. So an API that facilitates that goal would have helped; a word_id=>topic_id mapping for a particular document would have been helpful.

So that's one data point :)

bhargavvader: I thought of the same use-case too. I think a middle-ground would be a sorted list. I'll wait for more discussion about this before making anything final...

gensim/models/ldamodel.py
bhargavvader: If we are only doing comparisons for a word within topics and not comparing between words, would this matter?

bhargavvader:
@piskvorky , taken into account all your suggestions and implemented it. The current return format is ` ({word_id => [topic_id_most_probable, topic_id_second_most_probable, ...])`, which seems like a fair middle-ground. 
Am still unsure of whether the phis are directly comparable, like you said...
@tmylk , could you clear that up and whether the new output format is ok?

gensim/models/ldamodel.py
bhargavvader: Added docs.

bhargavvader:
@tmylk , I think this is ready for a final review and merge. Could you have a look?

tmylk:
Looks good. Just a changelog and an ipython notebook with colored words would be good. An example of static word assignment would help too.

piskvorky:
Oh yes, good idea. A notebook to demo this new functionality (coloured document words; topics for static terms) would be great.

bhargavvader:
Yup, I have a nice idea for a notebook to accompany this. :)
@piskvorky , I've spoken to @tmylk about it and it seems fine - since the comparisons are between topics and not words, the feature length doesn't change the most probable topic(s). I've also tried it out on a few corpus/docs and it gives consistent results.

bhargavvader:
@piskvorky , @tmylk , have a look at the notebook and suggest any changes? 
I'll make a blog post to accompany this as well if it's fine.

bhargavvader:
@tmylk , @piskvorky , updated the notebook to give a little more illustrations of Document Word coloring. I briefly explain the use case and new method functionalities before jumping into it. 

Made changes to changelog as well, so if the logic and notebook are fine, it's ready to merge from my end.

piskvorky:
Awesome, thanks! This is a much awaited functionality :)

@tmylk how do we get all these nice little notebooks to users, how do they find them?

bhargavvader:
@piskvorky , could you have a brief look at the conversation on #683 towards the end and let me know if the idea of adding a `per_word_phi_topics` option to the user is viable?
 So, just to confirm, we'll have `get_document_topics(per_word_topics=False, per_word_phi_values=False)` as the API, where if both are true then we'll return both a `word_id => [most_probable_topic, ...]`, and also a `word_id => [(topic_0, phi_value),....(topic_n, phi_value)]`. If only one of them is true, we obviously return only that one. 

The more popular one which will be used for practical purposes is the list of most probable `topic_ids` (using which I have demonstrated document word coloring), but if the user really wishes to have a look at the `phi_values`, I think that option should be there.

piskvorky:
I see. It makes sense to me -- since we're implementing this, we might as well offer the "full power" to users.

Not sure we need an extra parameter though. Can't we always return the same thing (both mappings), for consistency?

gensim/models/ldamodel.py
```diff
@@ -900,10 +902,27 @@ def get_document_topics(self, bow, minimum_probability=None):
...
+        document_topics = [(topicid, topicvalue) for topicid, topicvalue in enumerate(topic_dist)
+                    if topicvalue >= minimum_probability]
+        if not per_word_topics:
+            return document_topics
+        else:
+            word_phi = [] # contains word and corresponding topic
+            for word_type, weight in bow:
+                phi_values = [] # contains phi values for each topic
+                for topic_id in range(0, self.num_topics):
+                    if phis[topic_id][word_type] >= minimum_probability:
```
tmylk: should it be a different parameter `minimium_phi_probability` ?

gensim/models/ldamodel.py
```diff
@@ -883,13 +883,15 @@ def top_topics(self, corpus, num_words=20):
...
 
-    def get_document_topics(self, bow, minimum_probability=None):
+    def get_document_topics(self, bow, minimum_probability=None, per_word_topics=False):
         """
         Return topic distribution for the given document `bow`, as a list of
         (topic_id, topic_probability) 2-tuples.
 
         Ignore topics with very low probability (below `minimum_probability`).
 
+        If per_word_topics is True, it also returns a list of topics, sorted in descending order of most likely topics for that word.
```
tmylk: Explain what `phi`s are and that they are multiplied by word count

gensim/models/ldamodel.py
bhargavvader: Hmm... Wouldn't hurt to add it I guess.

gensim/models/ldamodel.py
bhargavvader: Will do.

bhargavvader:
@tmylk , @piskvorky 
Just an update: I've made changes in the documentation to reflect the phi_values being scaled, and the returning of both sorted topics and the phi_vales. 
I've made changes to the tests to reflect this as well.

I have also added a small portion in the notebook to show phi_values and it's scaling - this should make it clear for people if they go through the notebook, they don't have to hunt for the scaling in the doc strings.

Just to clarify, right now if `per_word_topics` is true, it returns **both** sorted list of likely word-topics and the `phi_value`s.

gensim/models/ldamodel.py
bhargavvader: Just a note: if the user does not enter a minimum_phi_probability, it ends up taking `self.minimum_probability`'s value.

graychan:
@bhargavvader , thanks a lot for making this revision for uses like me and answering my questions raised in the comment to issue #683. 

I have a minor comment and hope that this is not a nuisance. I understand that you cannot really scale phi_values back using `cts` if you have multiple documents in `bow`. However, in the [code](https://github.com/bhargavvader/gensim/blob/04012b847ac5030736c0e6dd2ea5aac7fdb3e0e5/gensim/models/ldamodel.py#L886), I wonder whether minimum_phi_probability should be changed to minimum_phi_value since the phi_values that are compared with minimum_phi_probability are not probabilities.

tmylk:
@bhargavvader Did you test it on your Mac?

The test fails in all pythons in our OS X environment with xcod7.0 and xcode 7.3 too

https://travis-ci.org/MacPython/gensim-wheels/jobs/136597581

Can the test be made to be less sensitive to multihreading?

> 
> ======================================================================
> FAIL: testGetDocumentTopics (gensim.test.test_ldamodel.TestLdaMulticore)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/Users/travis/build/MacPython/gensim-wheels/venv/lib/python3.3/site-packages/gensim/test/test_ldamodel.py", line 287, in testGetDocumentTopics
>     self.assertEqual(word_topics[0][1], expected_topiclist)
> nose.proxy.AssertionError: Lists differ: [0, 1] != [1, 0]
> First differing element 0:
> 0
> 1
> - [0, 1]
> + [1, 0]

bhargavvader:
@tmylk , that's very strange, it passes fine on my machine. And it should ideally pass on a multithreaded version as well. At any rate I'll make changes and open a PR.

@graychan I'm opening a PR to address the test problems, I'll change `probabilities` to `values` while I'm at it. :)

Issue #692, opened 2016-05-05
Warning if vocabulary is entirely single character elements in doc2vec


tmylk:
[Many users](https://groups.google.com/d/msg/gensim/YO6hRhFoB9c/dDVG427FCAAJ) on the mailing list pass strings to doc2vec instead of lists. 
It results in poor results as vocab is single chars.

A warning will help detect that.

dsquareindia:
I'll take this up :+1:

gojomo:
Yes! The test could be `words` is a true string type, and/or that the `vocab` keys are all single-characters, and/or that the `vocab` is "very small" (threshold TBD). 

Other sanity checks based on frequent issues are also possible. A few ideas:

* model very large compared to single-pass training-data size (risks overfitting)
* `alpha` raised again after descending to `min_alpha` (may indicate confusion about training cycles)
* calling `load()` on an instance rather than the class (often indicates not saving reference to return value)
* mismatch in expected count for `train()` and actual (may indicate skipped steps when doing `build_vocab()` and `train()` separately)

bloody76:
What about just checking the type of the argument passed ? I can imagine at the beginning of `train_document_*` functions something like:

```
if isinstance(doc_words, str) or isinstance(doc_words, unicode):
    logger.warn("You are passing strings instead of list of ids for the training of your doc2vec model. Expect to have a vocabulary composed of one-char words.")
```

dsquareindia:
Just had a small doubt here: Should I be making an __init__(self, *args, **kwargs) for TaggedDocument to raise those warnings?

tmylk:
Left to implement:

- [ ] Raise warning if calling load() on an instance rather than the class.
- [ ] Raise warning if mismatch in expected count for train() and actual.

aayush96:
> calling `load()` on an instance rather than the class

Working on fixing this but I'm not fully certain of how one might go about "calling `load()` by instance". Is there a previously known issue/error-report that I could refer to?

tmylk:
@dust0x Class methonds and instance methods are different in python, so you can raise warning on one of them.

dust0x:
@tmylk yep, I see that. Could something go wrong when calling `load()` on instance?

What kind of warning should one be looking for here? A call to load by instance seems to be working fine for me.

tmylk:
Do you see the difference in behaviour of the two methods?

gojomo:
[...]

PR #705, opened 2016-05-24
[MRG] ENH: Raise warnings if vocab in single character elements in word2vec/doc2vec

dsquareindia:
Addresses #692. TODO:

- [X]  Raise warning if alpha is raised again after descending to minimum value. (Where should this be ideally done @gojomo?)
- [ ] Raise warning if calling `load()` on an instance rather than the class.
- [ ] Raise warning if mismatch in expected count for `train()` and actual.
- [X] Add test suit to test_doc2vec.py, test_word2vec.py

@tmylk @gojomo am I going on the right track with this? Also how should the warnings be dealt with in `doc2vec_inner.c`?

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -645,8 +645,14 @@ class Doc2Vec(Word2Vec):
         min_reduce = 1
         interval_start = default_timer() - 0.00001  # guard against next sample being identical
         interval_count = 0
+        checked_string_types = 0
         vocab = defaultdict(int)
         for document_no, document in enumerate(documents):
+            if not checked_string_types:
+                if isinstance(document.words, string_types):
+                    logger.warn("Each 'words' should be a list of words (usually unicode strings)."
+                                "First 'words' here is instead plain %s." % type(document.words))
+                checked_string_types += 1
             if document_no % progress_per == 0:
                 interval_rate = (total_words - interval_count) / (default_timer() - interval_start)
                 logger.info("PROGRESS: at example #%i, processed %i words (%i/s), %i word types, %i tags",
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -423,6 +423,7 @@ class Word2Vec(utils.SaveLoad):
         if size % 4 != 0:
             logger.warning("consider setting layer size to a multiple of 4 for greater performance")
         self.alpha = float(alpha)
+        self.min_alpha_yet_reached = float(alpha) # To warn user if alpha increases
         self.window = int(window)
         self.max_vocab_size = max_vocab_size
         self.seed = seed
@@ -520,7 +521,13 @@ class Word2Vec(utils.SaveLoad):
         total_words = 0
         min_reduce = 1
         vocab = defaultdict(int)
+        checked_string_types = 0
         for sentence_no, sentence in enumerate(sentences):
+            if not checked_string_types:
+                if isinstance(sentence, string_types):
+                    logger.warn("Each 'sentences' item should be a list of words (usually unicode strings)."
+                                "First item here is instead plain %s.", type(sentence))
+                checked_string_types += 1
             if sentence_no % progress_per == 0:
                 logger.info("PROGRESS: at sentence #%i, processed %i words, keeping %i word types",
                             sentence_no, sum(itervalues(vocab)) + total_words, len(vocab))
@@ -746,6 +753,9 @@ class Word2Vec(utils.SaveLoad):
             job_batch, batch_size = [], 0
             pushed_words, pushed_examples = 0, 0
             next_alpha = self.alpha
+            if next_alpha > self.min_alpha_yet_reached:
+                logger.warn("Effective 'alpha' higher than previous training cycles")
+            self.min_alpha_yet_reached = next_alpha
             job_no = 0
 
             for sent_idx, sentence in enumerate(sentences):
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,6 @@ before_install:
 install:
   - conda create --yes -n gensim-test python=$TRAVIS_PYTHON_VERSION pip atlas numpy scipy
   - source activate gensim-test
+  - pip install testfixtures
   - python setup.py install
 script: python setup.py test
--- a/gensim/test/test_doc2vec.py
+++ b/gensim/test/test_doc2vec.py
@@ -17,6 +17,7 @@ import tempfile
 
 from six.moves import zip as izip
 from collections import namedtuple
+from testfixtures import log_capture
 
 import numpy as np
 
@@ -279,6 +280,32 @@ class TestDoc2VecModel(unittest.TestCase):
         self.assertEqual(len(model.docvecs.offset2doctag), len(model2.docvecs.offset2doctag))
         self.assertTrue(np.allclose(model.docvecs.doctag_syn0, model2.docvecs.doctag_syn0))
 
+    @log_capture()
+    def testBuildVocabWarning(self, l):
+        """Test if logger warning is raised on non-ideal input to a doc2vec model"""
+        raw_sentences = ['human', 'machine']
+        sentences = [doc2vec.TaggedDocument(words, [i]) for i, words in enumerate(raw_sentences)]
+        model = doc2vec.Doc2Vec()
+        model.build_vocab(sentences)
+        warning = "Each 'words' should be a list of words (usually unicode strings)."
+        self.assertTrue(warning in str(l))
+
+    @log_capture()
+    def testTrainWarning(self, l):
+        """Test if warning is raised if alpha rises during subsequent calls to train()"""
+        raw_sentences = [['human'],
+                         ['graph', 'trees']]
+        sentences = [doc2vec.TaggedDocument(words, [i]) for i, words in enumerate(raw_sentences)]
+        model = doc2vec.Doc2Vec(alpha=0.025, min_alpha=0.025, min_count=1, workers=8, size=5)
+        model.build_vocab(sentences)
+        for epoch in range(10):
+            model.train(sentences)
+            model.alpha -= 0.002
+            model.min_alpha = model.alpha
+            if epoch == 5:
+                model.alpha += 0.05
+        warning = "Effective 'alpha' higher than previous training cycles"
+        self.assertTrue(warning in str(l))
 #endclass TestDoc2VecModel
 
 
--- a/gensim/test/test_word2vec.py
+++ b/gensim/test/test_word2vec.py
@@ -22,6 +22,7 @@ from gensim import utils, matutils
 from gensim.utils import check_output
 from subprocess import PIPE
 from gensim.models import word2vec
+from testfixtures import log_capture
 
 module_path = os.path.dirname(__file__)  # needed because sample data files are located in the same folder
 datapath = lambda fname: os.path.join(module_path, 'test_data', fname)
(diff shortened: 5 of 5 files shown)


gojomo:
For each of these warnings, the goal should be to make note of the problem exactly once, at the earliest possible time. We don't want the overhead of redundant checks; the user doesn't need X million warnings for their X million text examples. So avoid doing checks for every document, during training, and instead do as soon as the problem parameter/input is seen for the 1st time. 

The alpha is managed by the `train()` method, so the check would have to go there. 

Most of these issues apply to Word2Vec as well, and so in most cases the checks will be there (and inherited by Doc2Vec). 

You won't likely have to be concerned about either the `.c` files (because they're auto-generated from the cython `.pyx` files), or the `.pyx` files (as those are optimized implementations for which any problems should have been discovered earlier, in pure-python code).

dsquareindia:
@gojomo thanks for the help! I've updated the PR. Hope this is better. What should be a good threshold value for a "small" vocab? Also, `alpha` monotonically decreases right? I'll change the code if that's not the case.

gensim/models/word2vec.py
```diff
@@ -509,9 +510,16 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
         Each sentence must be a list of unicode strings.
 
         """
+        for sentence in sentences:
```
gojomo: This adds the overhead of a full pass over the corpus in the case where the user has already done things right; that's excessive. It'd be better to add the check to a pass that's already happening (eg `scan_vocab()`). (And, it may even be a desirable optimization to just check the first example... this mistake is likely to be all-or-nothing, and we don't want to burden a user doing the right thing, with a tens-of-millions-of-examples corpus, with a check of every example.)

Also, this code appears to only work for the Doc2Vec case, when 'sentence' is something TaggedDocument-shaped with a 'words' field. In Word2Vec, there's no such field, the example is just the list-of-tokens. Are you testing in a Word2Vec triggering scenario?

gensim/models/word2vec.py
```diff
@@ -509,9 +510,16 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
...
+            if hasattr(sentence, "words"):
+                if isinstance(sentence.words, string_types):
+                    warnings.warn("Sentences should be a list of unicode strings. True string provided.")
+                    break
         self.scan_vocab(sentences, progress_per=progress_per, trim_rule=trim_rule)  # initial survey
         self.scale_vocab(keep_raw_vocab=keep_raw_vocab, trim_rule=trim_rule)  # trim by min_count & precalculate downsampling
         self.finalize_vocab()  # build tables & arrays
+        self.check_vocab() # raise warning if vocab consists of all single characters
```
gojomo: This check seems redundant given the earlier check: the main way someone would wind up with a single-char vocab is if they'd passed simple strings. (If they've already been warned about that, no need to do this check.) 

But also, some will want to do the scan_vocab/scale_vocab/finalize_vocab steps themselves (in lieu of build_vocab), to have more control – and a user doing that should still get whatever checks we deem helpful. So a check like this (if it survives) would better happen within one of those methods, at the earliest possible time the anomaly can be detected.

gensim/models/word2vec.py
```diff
@@ -776,6 +795,9 @@ def job_producer():
                             progress = 1.0 * pushed_words / total_words
                         next_alpha = self.alpha - (self.alpha - self.min_alpha) * progress
                         next_alpha = max(self.min_alpha, next_alpha)
+                        # raise warning in case alpha rises again
+                        if next_alpha > current_alpha:
+                            warnings.warn("'alpha' risen again after descending.")
```
gojomo: I don't see any way this warning could ever be triggered, given `current_alpha` is a local set above. (Have you been able to trigger it in your testing?) The user potential-error we're trying to warn about is when someone, in repeated `train()`s, somehow starts again at a higher `alpha`. To catch that will require remembering a sort of lowest-alpha-yet-reached, across multiple calls to `train()`.

gensim/models/word2vec.py
```diff
@@ -71,6 +71,7 @@
 import logging
 import sys
 import os
+import warnings
```
gojomo: Rather than using the `warnings` module, these advisories should work fine as more simple `logger.warn()` messages.

gojomo:
I don't have any confident idea of a good "your vocabulary is suspiciously small" threshold, so we should probably just skip that idea unless/until there's more evidence it's needed. (It was just one of a bunch of 'and/or' ideas for catching common errors.)

gensim/models/word2vec.py
dsquareindia: So would it be better to just do away with this test? In any case the flow of scan->scale->finalize will have to be followed right? The check will already happen in scan then....

gensim/models/word2vec.py
dsquareindia: Sorry for that I misunderstood the problem. So once alpha rises again, would it be better to keep track of the new alpha or keep the lowest value yet reached? I'm just keeping a variable called `min_alpha_yet_reached` which is initialized as `alpha`. I keep checking in every training cycle if this is greater than `alpha` and set it to `alpha` if so. Whereas if it is lesser, I was thinking a warning should be raised and the `min_alpha_yet_reached` should be set to the raised `alpha` so as to eliminate the noise in the log. Is this approach fine?

gensim/models/word2vec.py
dsquareindia: Oh alright I didn't know about the all-or-nothing scenario. I've added checks in the `scan_vocab` methods of `Word2Vec` and `Doc2Vec` wherein I'm only checking if the `sentence` or `document.words` is an instance of `string_types` for sentence (or document) 0. So I don't we need the check for all single characters in `vocab` since, as you said, this check already covers it. Is this approach fine?

gensim/models/word2vec.py
dsquareindia: `logger.warn()` messages don't appear in the console window right? Does the user have to check into the log file to see the warning?

dsquareindia:
@gojomo I've made the changes to the code. Could you please check? Also regarding raising the warning for a mismatch in the expected count for train() and actual, what is exactly count over here and what should it be compared against to raise the warning?

gensim/models/doc2vec.py
```diff
@@ -636,6 +636,12 @@ def scan_vocab(self, documents, progress_per=10000, trim_rule=None):
         interval_start = default_timer() - 0.00001  # guard against next sample being identical
         interval_count = 0
         vocab = defaultdict(int)
+        try:
+            if isinstance(list(documents)[0].words, string_types):
```
gojomo: Many users won't be passing an indexable/all-in-memory corpus – so this will break for them.

gensim/models/word2vec.py
```diff
@@ -746,6 +753,9 @@ def job_producer():
             job_batch, batch_size = [], 0
             pushed_words, pushed_examples = 0, 0
             next_alpha = self.alpha
+            if next_alpha > self.min_alpha_yet_reached:
```
gojomo: This looks like a good minimal catch of the problem situation! Are you able to trigger this, for example by calling `train()` a second time?

The point of this error message is just to make the user think about whether their re-training (without the usual descending-alpha of typical SGD) makes sense. So let's word it: "Effective 'alpha' higher than previous training cycles. Is this your intent?"

dsquareindia: yes I was able to trigger it by changing the alpha manually like:
```python
model = Doc2Vec(alpha=0.025, min_alpha=0.025, min_count=1, workers=8, size=5)

model.build_vocab(tagged_docs)

for epoch in range(10):

    model.train(tagged_docs)

    model.alpha -= 0.002

    model.min_alpha = model.alpha

    if epoch == 5:
      model.alpha += 0.05
```
The user won't exactly do this but I just wanted to highlight that manual updation can trigger this.
Yes my intent was to make the user aware that the alpha, which should be monotonically decreasing over training cycles, has increased. Yep rewording it like that would be better.

gensim/models/doc2vec.py
dsquareindia: sorry for bothering you so much :(
What would be a better way to check then?

gojomo: On the existing iteration, check the current sentence. *But*, once the first is checked, set a flag to prevent further checks – based on our assumption that the user either gets this "all right" or "all wrong" and that (when wrong) they don't want redundant warnings for every sentence.

dsquareindia:
@tmylk @gojomo I've added unit tests for the `build_vocab()` warning and the `train()` warning. Sorry for still using the warnings module. The warnings weren't displaying on the console window with `logger.warn` and also the testing with `warnings.warn` seemed more convenient. It it fine?

gensim/models/doc2vec.py
```diff
@@ -635,8 +635,13 @@ def scan_vocab(self, documents, progress_per=10000, trim_rule=None):
         min_reduce = 1
         interval_start = default_timer() - 0.00001  # guard against next sample being identical
         interval_count = 0
+        checked_string_types = 0
         vocab = defaultdict(int)
         for document_no, document in enumerate(documents):
+            if not checked_string_types:
+                if isinstance(document.words, string_types):
+                    warnings.warn("'words' should be a list of unicode strings. True string type provided.")
```
piskvorky: Instead of "True string type", why not output the actual type of `document.words`?

Plus maybe even output the entire `document` example, for explicitness and clarity.

gensim/models/doc2vec.py
```diff
@@ -635,8 +635,13 @@ def scan_vocab(self, documents, progress_per=10000, trim_rule=None):
         min_reduce = 1
         interval_start = default_timer() - 0.00001  # guard against next sample being identical
         interval_count = 0
+        checked_string_types = 0
         vocab = defaultdict(int)
         for document_no, document in enumerate(documents):
+            if not checked_string_types:
```
piskvorky: How fast / slow is `warnings.warn`?

If the check is fast (before argument formatting), it may be easier to just issue `warnings.warn` and let the warnings module handle print-never/print-once/print-every-time internally.

piskvorky: On second thoughts, we definitely want these warnings to appear in the log, so we have to check anyway (and don't log if already logged before).

gojomo:
Changed title to reflect that this can be an issue in Word2Vec as well (and should be checked there too).

tmylk:
`testfixtures` package allows to test for logging output so we can use `logger.warn` instead of just warnings.

dsquareindia:
Thanks for the suggestion @tmylk however I'm getting an `ImportError` on `import testfixtures` while running nosetests. I don't get this error outside of nosetests. I've checked an answer [here](http://stackoverflow.com/questions/3073259/python-nose-import-error) but I don't think this answer will work in our case since I can't remove the `__init__.py` nor can I specify a particular target directory since it won't work for other users. Also would this dependency be a problem for other users while running nosetests?

dsquareindia:
@tmylk the build still seems to be failing.....

dsquareindia:
@piskvorky @gojomo @tmylk I think travis is happy. I think the test failing is unrelated to my PR. Could you please review once?

gensim/models/word2vec.py
```diff
@@ -520,7 +521,12 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
         total_words = 0
         min_reduce = 1
         vocab = defaultdict(int)
+        checked_string_types = 0
         for sentence_no, sentence in enumerate(sentences):
+            if not checked_string_types:
+                if isinstance(sentence, string_types):
+                    logger.warn("'sentences' should be a list of list of unicode strings. %s provided.", type(sentence))
```
gojomo: To be pedantically accurate, `sentences` need not be a list – just an iterable. But the important thing being checked here is whether the first item is the expected list-of-strings. So I would word this warning as: "Each `sentences` item should be a list of words (usually unicode strings). First item here is instead plain %s." The warning in doc2vec could be similarly: "Each `words` should be a list of words (usually unicode strings). First 'words' here is instead plain %s".) (I find that clearer than just "%s provided" which may at first glance seem like no contradiction with preceding advice.)

gojomo:
Other than my slight suggestion regarding warning wording, this looks good-to-go. 

The test failure is in gensim.test.test_glove2word2vec.TestGlove2Word2Vec and thus seems unrelated to these changes. @tmylk I suggest that test should be disabled here and in 'develop' until it can be independently made reliable.

gensim/models/word2vec.py
dsquareindia: Agreed! Made the changes.

dsquareindia:
I've made the changes in the log message, rebased and squashed.

dsquareindia:
@tmylk can this be merged?

tmylk:
@dsquareindia Thanks a lot for this very needed PR!

dsquareindia:
No problem! Thanks a ton for the help and reviews!

PR #707, opened 2016-05-25
Adding similarity for two out of training sentences to doc2vec

ellolo:
I added a method to the DocvecsAarray to compute similarity between two sentences that are out of training. Not sure if the method should be better placed under Doc2Vec.
Please bear with me, this is my first contribution to an open source project.
I included a simple test.


tmylk:
Thanks. Looks good to me. Pinging @gojomo for review.

ellolo:
@gojomo could you please review this feature?

ellolo:
@tmylk i got some errors in the test but it seems unrelated to my changes

gojomo:
I can see people needing to do this, but as it's just a one-liner, not sure it needs an API convenience method. 

As an API method name, the "oot" abbreviation is somewhat cryptic. (I see what it means from this PR title, but not even the method comment uses the same language. It's generally good to avoid abbreviations unless they are very, very pervasively used.) 

Since the heart of the method is inference, and our existing inference steps/alpha defaults are somewhat wild-guesses that many people change, allowing the specification of non-default steps/alpha to be used for both inferences would make sense.

gensim/test/test_doc2vec.py
```diff
@@ -123,6 +123,18 @@ def test_empty_errors(self):
...
+    def test_oot_similarity(self):
+        """Test similarity of out of training sentences"""
+        empty_str = ''
+        rome_str = 'rome italy'
+        car_str = 'car'
+        corpus = list(DocsLeeCorpus(True))
+
+        model = doc2vec.Doc2Vec(min_count=1)
+        model.build_vocab(corpus)
+        self.assertEqual(int(model.docvecs.oot_similarity(model, '', '')), 1)
```
gojomo: This is not testing what you think it is. Repeated calls to inference with the exact same tokens don't necessarily result in identical vectors, because of randomness in the Doc2Vec algorithm. But it works here because these empty-strings result in no inference at all, which means each vector stays its initial untrained randomized value (which we have tried to make start in the same place for the same inputs). With actual tokens, you might not get a 1.0 similarity.

gensim/test/test_doc2vec.py
```diff
@@ -123,6 +123,18 @@ def test_empty_errors(self):
...
+        self.assertTrue(model.docvecs.oot_similarity(model, rome_str, rome_str) > model.docvecs.oot_similarity(model, rome_str, car_str))
```
gojomo: This also is not likely testing what's expected, as strings are being passed rather than lists-of-tokens. I'd guess 'rome italy' is being interpreted as ['r', 'o', 'm', 'e', ' ', 'i', 't', 'a', 'l', 'y'], which becomes something like just ['i', 'a'] when reduced to known-tokens. Then 'car' becomes ['a'] via the same process.

ellolo:
@gojomo I incorporated your comments. I hope the name for the method is clear enough. As for test, I interpreted incorrectly the infer_vector method, it should be fine now.

tmylk:
Maybe `similarity_of_unseen_docs` is better?

ellolo:
@tmylk i changed the name according to your suggestion

gensim/models/doc2vec.py
```diff
@@ -494,7 +494,17 @@ def n_similarity(self, ds1, ds2):
         v2 = [self[doc] for doc in ds2]
         return dot(matutils.unitvec(array(v1).mean(axis=0)), matutils.unitvec(array(v2).mean(axis=0)))
 
+    def similarity_unseen_docs(self, model, doc_words1, doc_words2, alpha=0.1, min_alpha=0.0001, steps=5):
+        """
+        Compute cosine similarity between two post-bulk out of training documents.
 
+        Document should be a list of (word) tokens.
+        """
+        d1 = model.infer_vector(doc_words1, alpha, min_alpha, steps)
```
piskvorky: Use named parameters (rather than positional).

It's safer in case the argument order inside `infer_vector` changes in the future.

tmylk:
Thanks for the PR!

gojomo:
Each name suggestion has been better, but I'd make it even more literally descriptive, perhaps: `inferred_similarity`. (There's no need to qualify as 'unseen'.)

PR #720, opened 2016-06-02
Translated doc2vec tutorial script from bash to Python

andrewjlm:
The bash script at the beginning of the doc2vec tutorial notebook wasn't working for me (and several other people) so I rewrote it in Python. Now it works and the data processing is a bit less opaque to users who aren't familiar with sed/awk.


tmylk:
Thanks!

piskvorky:
Awesome, thanks Andrew!

PR #722, opened 2016-06-02
make deepir py3 compatible

jonmcoe:
main changes:
print statements
convert bytes to str before json loads (behind conditional)
x.next --> next(x)

smaller changes:
requirements at top
attempt import matlib before the jupyter directive


tmylk:
Thanks a lot for the fix!

PR #724, opened 2016-06-03
add ipynb checkpoints into gitignore

jonmcoe:
otherwise your repo is dirty basically as soon as you load one


tmylk:
Thanks for the PR!

PR #726, opened 2016-06-03
model_sanity assertion was using wrong value

seanlaw:
PyCon2016 Sprint - Bugfix


tmylk:
Thanks for the fix!

piskvorky:
Hah, was the reason for the flaky doc2vec tests @gojomo ?

gojomo:
In fact this error made that line's assertion guaranteed to succeed.

PR #729, opened 2016-06-03
Tutorial1 bugfix

peterg17:
I changed /gensim/summarization/keywords.py by making an if-statement to check for python2 and import queue/Queue differently for each version so it doesn't fail, also imported sys
Aka I added this code:
[http://stackoverflow.com/questions/29687837/queue-importerror-in-python-3](url)

### DIFF
--- a/gensim/summarization/keywords.py
+++ b/gensim/summarization/keywords.py
@@ -10,7 +10,12 @@ from gensim.summarization.commons import build_graph as _build_graph
 from gensim.summarization.commons import remove_unreachable_nodes as _remove_unreachable_nodes
 from gensim.utils import to_unicode
 from itertools import combinations as _combinations
-from six.moves.queue import Queue as _Queue
+import sys
+is_py2 = sys.version[0] == '2'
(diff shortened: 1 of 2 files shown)


gensim/summarization/keywords.py
```diff
@@ -10,7 +10,12 @@
 from gensim.summarization.commons import remove_unreachable_nodes as _remove_unreachable_nodes
 from gensim.utils import to_unicode
 from itertools import combinations as _combinations
-from six.moves.queue import Queue as _Queue
+import sys
+is_py2 = sys.version[0] == '2'
+if is_py2:
+    import Queue as queue
```
tmylk: should it be _Queue ?

peterg17: Hi tmylk, that's actually on purpose. There's a line below that says "queue=_Queue", but that's the only place where _Queue is used so i figured I would just delete that line and use queue, I can change it back to _Queue and resubmit this pull request if that would be better received

Issue #51, opened 2011-09-02
Approximate similarity search


piskvorky:
Google published a whitepaper http://www.google.com/trends/correlate on approximate kNN queries. See how it could apply to gensim & semantic similarity searches.

Background:
* currently, gensim does a linear scan (compare query against every indexed vector, by means of a matrix multiplication)
* I tried fancier indexing techniques, but they degenerate into linearly checking each datum anyway, for the high-dimensional vectors
* even worse, they access objects *out-of-order* (thrashing caches & HW buffers), so much **much** slower than plain linear scan in reality (matrix multiplication is linear in index size, but the constant factors are super low)
* Google claims this new technique works well for high-dim data as well => maybe finally something faster than a linear scan?

jtmcmc:
Just curious considering http://radimrehurek.com/2013/12/performance-shootout-of-nearest-neighbours-contestants/#survivors if you feel like there is a best library to integrate now if this was to be done (thinking of tackling this issue)

piskvorky:
Hmm, I wonder if it makes sense to integrate the algo fully (as in, implement it directly in Python/C/Cython). Not difficult, but not trivial either.

Another option is to rely on Annoy as a 3rd party lib, no deep integration, and just make it easier to use the Annoy API from gensim (or vice versa). I remember it was picky about the type of input it accepts etc, it was a bit unintuitive.

Tackling this issue will be welcome :)

jtmcmc:
Yes I see how annoy is a bit hard to implement. I've also found https://github.com/ryanrhymes/panns which maybe could be a better fit. I'm going to get annoy installed and try and do some comparisons. Alternatively the google correlate algorithm doesn't seem that complicated to implement so that could be promising as well.

piskvorky:
I don't think Annoy is hard to implement. It's a pretty straightforward algo.

I mean, Erik's C++ implementation is involved, because it's heavily optimized, goes for memory-mapping etc etc. But the algo itself is rather clean and simple.

Anyway, let me know how you progress. Would be great to finally have something efficient in gensim :)

jodaiber:
This would be incredibly useful! Is there any update on approximate sim. search in gensim (i.e. is anyone working on it)?

piskvorky:
The only update is, @erikbern (author of Annoy) left Spotify... but he still works on Annoy, somehow :)

[...]

PR #732, opened 2016-06-03
[WIP] Add support for using Annoy as an external similarity index

karlhigley:
(Related to #617, #527, and #51.)

This is a partial implementation of an external similarity index based on Annoy for approximate nearest neighbor search. It's based on #617 and the related comments, using a test-first approach. I'm running out of time to pursue it during the PyCon sprints, but wanted to share the pieces that may be useful for pushing #617 forward (e.g. class method index construction, conditional Annoy dependency in the tests).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ install:
   - conda create --yes -n gensim-test python=$TRAVIS_PYTHON_VERSION pip atlas numpy scipy
   - source activate gensim-test
   - pip install pyemd
+  - pip install annoy
   - pip install testfixtures
   - python setup.py install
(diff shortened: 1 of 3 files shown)


karlhigley:
@tmylk: I had a bit of trouble with Doc2Vec during the sprints, but I think I figured out a way to do it.

karlhigley:
It looks like the tests don't run on Travis, since Annoy is not installed there. Not sure how to fix the test failure in Python 2.6 either.

karlhigley:
(Nevermind, #758 added annoy.)

tmylk:
Implemented in #784 and others

Issue #740, opened 2016-06-10
testGetDocumentTopics and testTermTopics fail in Appveyor and OSX in MultiCoreLDA


tmylk:
See comments in https://github.com/piskvorky/gensim/pull/656 for details.

it is blocking the 0.13.0rc1 from becoming a release.

bhargavvader:
Do you mean #704 ? 
It's weird that this is happening - it worked fine before. PR #741 addresses (hopefully!) `testGetDocumentTopics`, I'll open a PR to deal with `testTermTopics`.

bhargavvader:
@tmylk , any luck with this?

bhargavvader:
@tmylk , has this been resolved?

tmylk:
Fixed in #741 and #742

PR #742, opened 2016-06-10
[MRG] Static Word Tests

bhargavvader:
@tmylk wrt #740 . Redid the tests so that it only checks the presence of certain topics.

Edit: python 2.7 fails because of a time-out.

### DIFF
--- a/gensim/test/test_ldamodel.py
+++ b/gensim/test/test_ldamodel.py
@@ -293,19 +293,15 @@ class TestLdaModel(unittest.TestCase):
         numpy.random.seed(0)
         model = self.class_(self.corpus, id2word=dictionary, num_topics=2, passes=100)
 
+        # checks if topic '1' is in the result list
         # check with id 
         result = model.get_term_topics(2)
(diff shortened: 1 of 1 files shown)


gensim/test/test_ldamodel.py
```diff
@@ -293,19 +293,15 @@ def testTermTopics(self):
...
         model = self.class_(self.corpus, id2word=dictionary, num_topics=2, passes=100)
 
+        # checks if topic '1' is in the result list
         # check with id 
         result = model.get_term_topics(2)
-        expected = [(1, 0.1066)]
-        # FIXME: fails on win and osx 
-        # self.assertEqual(result[0][0], expected[0][0])
-        # self.assertAlmostEqual(result[0][1], expected[0][1], places=2)
+        self.assertTrue(1 in result[0])
```
tmylk: what are we testing with this test?

bhargavvader: Whether `1` (which is `topic_1`)  beats the probability cutoff and is returned along with a probability value.

Issue #113, opened 2013-02-25
Relying on numpy.random makes results hard to reproduce


larsmans:
... and it also [confuses users of LDA](http://stackoverflow.com/q/15067734/166749).

In scikit-learn, we've solved this by [requiring](http://scikit-learn.org/dev/developers/index.html#random-numbers) a `random_state` argument to anything that wants random numbers; this may either be an `np.random.RandomState` object, or the seed for one. You might want to borrow this habit.

adammenges:
:+1:

piskvorky:
Here's a link to a pull request that did the same for GloVe; looks simple enough:
https://github.com/maciejkula/glove-python/pull/32

cvint13:
When I run the model I still get different results every time despite using a fixed seed:

```
lda = LdaMulticore(corpus, num_topics=100, workers=4,id2word=dictionary,
                   random_state=np.random.RandomState(seed=10101010),
                  alpha=np.array([.0075] * 100))
```

gojomo:
@cvint13 Anything multithreaded will also be subject to scheduling jitter from the OS – operations won't happen in the same order, and thus results will vary. For full reproduceability, you'd have to move to a (much-slower) single-threaded calculation (`workers=1`).

cvint13:
Ahhh ok, that sucks.

piskvorky:
For completeness -- `LdaMulticore` uses processes, not threads.

@menshikh-iv is it expected that @cvint13 would get different results, despite a fixed seed? Are the individual jobs processed in arbitrary (non-deterministic) order?

menshikh-iv:
@piskvorky yes (for multicore implementation only).

PR #748, opened 2016-06-20
Using numpy.random.RandomState instead of numpy.random in LDA

droudy:
In reference to #113 , a random_state parameter has been added to the LdaState initializer and a check_random_seed method as well to check the seed. The check_random_state method was taken from https://github.com/maciejkula/glove-python/pull/32/commits/95a04af9ef76c1b13949dc92c4a1bb1ecaa5cf3c , should I rewrite the method?


tmylk:
Please add tests for this particular function(checking both initializaitons work).

Also need to update existing tests that fix random seed to use this new API, for example https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_ldamodel.py#L253

tmylk:
Just need to add `random_state` to MultiCore

gensim/models/hdpmodel.py
```diff
@@ -586,6 +586,7 @@ def show_topics(self, num_topics=10, num_words=10, log=False, formatted=True):
         if num_topics < 0:
             num_topics = len(self.data)
 
+
```
tmylk: remove the line

gensim/models/ldamodel.py
```diff
@@ -91,6 +91,19 @@ def update_dir_prior(prior, N, logphat, rho):
 
     return prior
 
+def check_random_state(seed):
```
tmylk: `get_random_state` is a better name

gensim/test/test_ldamodel.py
```diff
@@ -48,6 +48,11 @@ def testfile():
     return os.path.join(tempfile.gettempdir(), 'gensim_models.tst')
 
 
+def testRandomState():
+    testcases = [numpy.random.seed(0), None, numpy.random.RandomState(0), 0]
+    for testcase in testcases:
+        assert(isinstance(ldamodel.check_random_state(testcase), numpy.random.RandomState))
```
tmylk: add a test `assertEqual(ldamodel.check_random_state(testcase), numpy.random.RandomState(0))`

gensim/test/test_ldamodel.py
```diff
@@ -97,7 +98,7 @@ def testAlpha(self):
         expected_shape = (2,)
 
         # should not raise anything
-        self.class_(**kwargs)
+        self.class_(**kwarghs)
```
piskvorky: @tmylk what is this change?

droudy:
@piskvorky it was an accident, it was fixed in https://github.com/RaRe-Technologies/gensim/pull/748/commits/8ea4d5852163b0c0178994336a3778aecbe30bf1

PR #753, opened 2016-06-24
Parameter name change in HdpModel and dtmmodel to be consistent with LdaModel

droudy:
Parameters topics and topn have been changed to num_topics and num_words in the methods show_topics() and print_topics() to be consistent with LdaModel. The calls to those methods have been changed as well in the tests.


gensim/test/test_ldamodel.py
```diff
@@ -47,8 +47,6 @@ def testfile():
     # temporary data will be stored to this file
     return os.path.join(tempfile.gettempdir(), 'gensim_models.tst')
 
-
-
```
tmylk: what is the reason for this change?

gensim/test/test_hdpmodel.py
```diff
@@ -65,7 +64,6 @@ def testShowTopics(self):
                 self.assertTrue(isinstance(v, float))
 
 
-
```
tmylk: why remove an empty line?

PR #754, opened 2016-06-24
Gensim adopters

tmylk:
Are you using gensim in production?

Please comment in this PR - we are compiling a list of Gensim adopters in the industry.


piskvorky:
@tmylk how about the other adopters?

tmylk:
For example?

piskvorky:
Sports Authority and the other I showed you.

PR #755, opened 2016-06-24
Parameter name change in HdpModel and dtmmodel to be consistent with LdaModel

droudy:
Parameters topics and topn have been changed to num_topics and num_words in the methods show_topics() and print_topics() to be consistent with LdaModel. The calls to those methods have been changed as well in the tests.


tmylk:
Thanks for the PR!

PR #763, opened 2016-06-29
LsiModel.docs_processed attribute

hobson:
unittests in test_lsimodel.py and test_corpora.py
works for test MmCorpus in unittests
also works for large custom corpora in production app


gensim/models/lsimodel.py
```diff
@@ -395,6 +396,7 @@ def add_documents(self, corpus, chunksize=None, decay=None):
                 if self.dispatcher:
                     logger.info("reached the end of input; now waiting for all remaining jobs to finish")
                     self.projection = self.dispatcher.getstate()
+                self.docs_processed += len(corpus) if hasattr(corpus, '__len__') else doc_no
```
tmylk: can we always use doc_no?

hobson: Yea, I guess so, for this line.

gensim/models/lsimodel.py
```diff
@@ -361,6 +361,7 @@ def add_documents(self, corpus, chunksize=None, decay=None):
                     num_terms=self.num_terms, chunksize=chunksize,
                     extra_dims=self.extra_samples, power_iters=self.power_iters)
                 self.projection.merge(update, decay=decay)
+                self.docs_processed += len(corpus) if hasattr(corpus, '__len__') else 0
```
piskvorky: you can pass a default value to `getattr()`

gensim/test/test_corpora.py
```diff
@@ -16,14 +16,17 @@
 
 import numpy
 
-from gensim.utils import to_unicode, smart_extension
+from gensim.utils import to_unicode  # , smart_extension
```
piskvorky: Remove import if no longer needed.

piskvorky:
Nice! These type of fixes are very valuable.

What was your motivation for this PR @hobson ?

hobson:
`docs_processed += getattr(corpus, '__len__', int)()`

Seems less explicit, less pythonic to me, but happy to do it if you like.


I was debugging my training on an iterable QuerySetCorpus class for a
variable-length database query result. The existing attribute was always
zero, confusing. Need to know exactly how many records trained my model.

gensim/test/test_corpora.py
hobson: Done.

Side Note: There are many more unused imports throughout gensim. They can be dangerous to remove, though, for someone like me unfamiliar with the internals of those packages being imported. For example `import seaborn` has side-effects, and obviously `from future import division` does too.

tmylk:
Agree. 
Thanks for the PR!

tmylk:
Just a line in the CHANGELOG and then will merge.

PR #766, opened 2016-06-30
Added malletmodel2ldamodel transformation function to mallet wrapper

dsquareindia:
This is concerning [this](https://groups.google.com/d/msg/gensim/rVHkzkFRNdI/p9uTDJ3ABAAJ) thread in the mailing list. @piskvorky @tmylk could you please check?

### DIFF
--- a/gensim/models/wrappers/ldamallet.py
+++ b/gensim/models/wrappers/ldamallet.py
@@ -44,6 +44,7 @@ from smart_open import smart_open
 
 from gensim import utils, matutils
 from gensim.utils import check_output
+from gensim.models.ldamodel import LdaModel
 
 logger = logging.getLogger(__name__)
 
@@ -346,3 +347,22 @@ class LdaMallet(utils.SaveLoad):
                     if total_weight:
                         doc = [(id_, float(weight) / total_weight) for id_, weight in doc]
                 yield doc
+
+
+def malletmodel2ldamodel(mallet_model):
+    """
+    Function to convert mallet model to gensim LdaModel. This works by copying the
+    training model weights (alpha, beta...) from a trained mallet model into the
+    gensim model.
+
+    Args:
+    ----
+    mallet_model : Trained mallet model
+
+    Returns:
+    -------
+    model_gensim : LdaModel instance; copied gensim LdaModel
+    """
+    model_gensim = LdaModel(id2word=mallet_model.id2word, num_topics=mallet_model.num_topics, alpha=mallet_model.alpha, iterations=100)
+    model_gensim.expElogbeta[:] = mallet_model.wordtopics
+    return model_gensim
--- a/gensim/models/wrappers/ldamallet.py
+++ b/gensim/models/wrappers/ldamallet.py
@@ -349,7 +349,7 @@ class LdaMallet(utils.SaveLoad):
                 yield doc
 
 
-def malletmodel2ldamodel(mallet_model):
+def malletmodel2ldamodel(mallet_model, gamma_threshold=0.001, iterations=50):
     """
     Function to convert mallet model to gensim LdaModel. This works by copying the
     training model weights (alpha, beta...) from a trained mallet model into the
@@ -358,12 +358,16 @@ def malletmodel2ldamodel(mallet_model):
     Args:
     ----
     mallet_model : Trained mallet model
+    gamma_threshold : To be used for LdaModel training.
+    iterations : number of iterations to be used for LdaModel training.
 
     Returns:
     -------
     model_gensim : LdaModel instance; copied gensim LdaModel
     """
-    model_gensim = LdaModel(id2word=mallet_model.id2word, num_topics=mallet_model.num_topics,
-                            alpha=mallet_model.alpha, iterations=mallet_model.iterations)
+    model_gensim = LdaModel(
+                id2word=mallet_model.id2word, num_topics=mallet_model.num_topics,
+                alpha=mallet_model.alpha, iterations=iterations,
+                gamma_threshold=gamma_threshold)
     model_gensim.expElogbeta[:] = mallet_model.wordtopics
     return model_gensim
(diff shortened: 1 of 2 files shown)


gensim/models/wrappers/ldamallet.py
```diff
@@ -346,3 +347,22 @@ def read_doctopics(self, fname, eps=1e-6, renorm=True):
...
+
+    Args:
+    ----
+    mallet_model : Trained mallet model
+
+    Returns:
+    -------
+    model_gensim : LdaModel instance; copied gensim LdaModel
+    """
+    model_gensim = LdaModel(id2word=mallet_model.id2word, num_topics=mallet_model.num_topics, alpha=mallet_model.alpha, iterations=100)
```
bhargavvader: Just curious, is there no way to pick up the number of iterations from the mallet model? I notice you've kept it as the default `100` right now.

dsquareindia: Yes even I thought of doing that but I instead just followed what @piskvorky wrote in the mailing list. Both are giving similar results though. I guess I'll switch this to `mallet_model.iterations` for better resemblance to the mallet model.

gensim/test/test_ldamallet_wrapper.py
```diff
@@ -95,6 +95,18 @@ def testSparseTransform(self):
...
 
+    def testMallet2Model(self):
+        if not self.mallet_path:
+            return
+        passed = False
+        tm1 = ldamallet.LdaMallet(self.mallet_path, corpus=corpus, num_topics=2, id2word=dictionary)
+        tm2 = ldamallet.malletmodel2ldamodel(tm1)
+        for document in corpus:
+            self.assertAlmostEqual(tm1[document][0][1], tm2[document][0][1], places=1)
+            self.assertAlmostEqual(tm1[document][1][1], tm2[document][1][1])
```
dsquareindia: @piskvorky I'm not too sure about this test. Values sometimes are differing at the first place too.

gensim/models/wrappers/ldamallet.py
```diff
@@ -346,3 +347,23 @@ def read_doctopics(self, fname, eps=1e-6, renorm=True):
...
+    Args:
+    ----
+    mallet_model : Trained mallet model
+
+    Returns:
+    -------
+    model_gensim : LdaModel instance; copied gensim LdaModel
+    """
+    model_gensim = LdaModel(id2word=mallet_model.id2word, num_topics=mallet_model.num_topics,
+                            alpha=mallet_model.alpha, iterations=mallet_model.iterations)
```
piskvorky: `iterations` means something very different between Variational Bayes (gensim) and Gibbs sampling (Mallet). There is no 1:1 correspondence, the semantics are different.

gensim/test/test_ldamallet_wrapper.py
piskvorky: Try more accurate `gamma_threshold` and `iterations`. But I don't think the results can really be made identical -- the inference algorithms are completely different.

I think the best we can hope for is a sanity check -- topics in same order, on some (model, query_document) pair where it's clear what the topic order should be, and where any deviation in topic order is clearly an error.

gensim/models/wrappers/ldamallet.py
dsquareindia: Should I let it remain 100 then? Equating both is not correct then....

piskvorky: Yes, probably leave the default. Or give an option (fnc args) to the user to set `gamma_threshold`/`iterations` explicitly.

gensim/models/wrappers/ldamallet.py
```diff
@@ -346,3 +347,23 @@ def read_doctopics(self, fname, eps=1e-6, renorm=True):
...
+
+    Args:
+    ----
+    mallet_model : Trained mallet model
+
+    Returns:
+    -------
+    model_gensim : LdaModel instance; copied gensim LdaModel
+    """
+    model_gensim = LdaModel(id2word=mallet_model.id2word, num_topics=mallet_model.num_topics,
```
piskvorky: We don't use this type of indent in gensim -- it's hard to maintain when editing & also visually ugly.

Please use normal "hanging indent" with 4 spaces: https://www.python.org/dev/peps/pep-0008/#indentation

dsquareindia:
@piskvorky I've addressed the comments. Could you please check?

gensim/models/wrappers/ldamallet.py
```diff
@@ -358,12 +358,16 @@ def malletmodel2ldamodel(mallet_model):
...
+    iterations : number of iterations to be used for LdaModel training.
 
     Returns:
     -------
     model_gensim : LdaModel instance; copied gensim LdaModel
     """
-    model_gensim = LdaModel(id2word=mallet_model.id2word, num_topics=mallet_model.num_topics,
-                            alpha=mallet_model.alpha, iterations=mallet_model.iterations)
+    model_gensim = LdaModel(
+                id2word=mallet_model.id2word, num_topics=mallet_model.num_topics,
```
piskvorky: PEP8: Hanging indent of 4 spaces.

gensim/models/wrappers/ldamallet.py
```diff
@@ -358,12 +358,16 @@ def malletmodel2ldamodel(mallet_model):
     Args:
     ----
     mallet_model : Trained mallet model
+    gamma_threshold : To be used for LdaModel training.
```
piskvorky: Misleading comment: there is no "training", the model is transferred from Mallet. These parameters only affect inference, model is unchanged.

gensim/models/wrappers/ldamallet.py
```diff
@@ -358,12 +358,16 @@ def malletmodel2ldamodel(mallet_model):
...
+    iterations : number of iterations to be used for LdaModel training.
```
piskvorky: dtto

dsquareindia:
@piskvorky I've made the changes. Could you please check?

Issue #764, opened 2016-06-29
MalletWrapper load_word_topics()  to return value


tmylk:
load_word_topics() doesn't return anything but stores the word topics internally in self.wordtopics.

It shouldn't set internal vars, but rather return the result to the caller (train), who should then assign it to internal object attributes.

https://github.com/RaRe-Technologies/gensim/commit/db70d04c650633cf2d1c0735c3cbd60cc3b36d77#commitcomment-18054588

bhargavvader:
I'll take this up.

bhargavvader:
We can close this now.

PR #767, opened 2016-06-30
load_word_topics() returns value

bhargavvader:
@tmylk, @piskvorky , with reference to #764 .

`load_word_topics()` now returns `word topics` which is assigned by `train` to `self.wordtopics`.

Before, there were two variables; `word_topics` and a `wordtopics`, and only the `wordtopics` variable was being used elsewhere; so I removed `word_topics`.

Is that ok?

### DIFF
--- a/gensim/models/wrappers/ldamallet.py
+++ b/gensim/models/wrappers/ldamallet.py
@@ -201,8 +201,8 @@ class LdaMallet(utils.SaveLoad):
                 tokenid = word2id[token]
                 wordtopics[int(topic), tokenid] += 1.0
         logger.info("loaded assigned topics for %i tokens", wordtopics.sum())
-        self.wordtopics = wordtopics
         self.print_topics(15)
+        return wordtopics
 
     def print_topics(self, num_topics=10, num_words=10):
         return self.show_topics(num_topics, num_words, log=True)
--- a/gensim/models/wrappers/ldamallet.py
+++ b/gensim/models/wrappers/ldamallet.py
@@ -164,7 +164,7 @@ class LdaMallet(utils.SaveLoad):
         # NOTE "--keep-sequence-bigrams" / "--use-ngrams true" poorer results + runs out of memory
         logger.info("training MALLET LDA with %s", cmd)
         check_output(cmd, shell=True)
-        self.wordtopics = self.load_word_topics()
+        self.word_topics = self.load_word_topics()
 
     def __getitem__(self, bow, iterations=100):
         is_corpus, corpus = utils.is_corpus(bow)
@@ -182,7 +182,7 @@ class LdaMallet(utils.SaveLoad):
 
     def load_word_topics(self):
         logger.info("loading assigned topics from %s", self.fstate())
-        wordtopics = numpy.zeros((self.num_topics, self.num_terms), dtype=numpy.float32)
+        word_topics = numpy.zeros((self.num_topics, self.num_terms), dtype=numpy.float32)
         if hasattr(self.id2word, 'token2id'):
             word2id = self.id2word.token2id
         else:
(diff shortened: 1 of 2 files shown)


tmylk:
Rename wordtopics to word_topics.

Also should there be a meaningful error like 'run train or load_word_topcs before showing topics' in https://github.com/bhargavvader/gensim/blob/6fd1ecbfe41d7ab2adc17d20179d467e986f477a/gensim/models/wrappers/ldamallet.py#L244

bhargavvader:
Will do. Also, this PR doesn't seem to have a travis build happening, any reason?

dsquareindia:
oh renaming to `word_topics` would break some other things actually....

tmylk:
@bhargavvader Well spotted about Travis - other PRs are working ok. Let's see what happens on your next commit.

bhargavvader:
@dsquareindia , exactly what else would break? Not if we change all the `wordtopics` to `word_topics` right?

dsquareindia:
@bhargavvader the `mallet2ldamodel` would fail and also the support for `ldamallet` in topic coherence would fail. Tests pass because tests for one of them hasn't been added and for the other one hasn't been merged. I'll change the spelling in my PR but could you please change it in `coherencemodel.py`? Also a quick `git grep` would help.

bhargavvader:
Cool, on it.

dsquareindia:
yeah @bhargavvader I'll make both the changes in my pr once this gets merged. Thanks!

bhargavvader:
@tmylk , @piskvorky , can you have a look?
Also, about the warning, I've just used `logger.info` now. Is that ok?

gensim/models/wrappers/ldamallet.py
```diff
@@ -242,7 +242,9 @@ def show_topics(self, num_topics=10, num_words=10, log=False, formatted=True):
         return shown
 
     def show_topic(self, topicid, topn=10):
-        topic = self.wordtopics[topicid]
+        if self.word_topics is None:
+            logger.info("Run train or load_word_topics before showing topics.")
```
tmylk: warn instead of info

bhargavvader: Yup, did.

tmylk:
Thanks for the PR. It's a useful fix!

piskvorky:
-1 on renaming `wordtopics` to `word_topics`. This breaks backward compatibility. At least make an alias.

gensim/models/wrappers/ldamallet.py
```diff
@@ -199,10 +199,10 @@ def load_word_topics(self):
                 if token not in word2id:
                     continue
                 tokenid = word2id[token]
-                wordtopics[int(topic), tokenid] += 1.0
-        logger.info("loaded assigned topics for %i tokens", wordtopics.sum())
-        self.wordtopics = wordtopics
+                word_topics[int(topic), tokenid] += 1.0
+        logger.info("loaded assigned topics for %i tokens", word_topics.sum())
         self.print_topics(15)
```
piskvorky: This print makes no sense without the `self.wordtopics` assignment.

bhargavvader: You're right - the print was there in the code earlier and I removed the `self.wordtopics` assignment so now it doesn't make sense.

bhargavvader:
@piskvorky , how exactly would one make an alias?
And what exactly do you mean by breaking the older code? I had checked fir occurrences of `wordtopics` in the repo and couldn't find anything I didn't change. Or do you mean tutorials and the like?

I'll make another PR for that and for the unnecessary print statement. Is there anything else you think should change?

piskvorky:
I mean, we use it in our code, in commercial projects :)

I think it's safe to assume some other people depend on it too, in the same way. Unless it's unavoidable, we try to be backward-compatible with our changes.

Alias = two variables referring to the same object, i.e. `x = y = some_var`.

bhargavvader:
Ok, makes sense. 

As for making an alias in this case, would you mean adding a line such as `self.wordtopics = self.word_topics = word_topics` or such whenever an assignment is being made?

piskvorky:
Yes, assign `self.wordtopics = self.word_topics`, with a big fat comment explaining why this alias is there.

Issue #698, opened 2016-05-12
TypeError: a bytes-like object is required, not 'str'


Tesfamariam:
I am trying to implement dynamic topic modeling with python Anaconda 3.4 distribution on Linux OS.However, I am having the following error:
TypeError: a bytes-like object is required, not 'str'
Any idea how I could solve this problem?

gojomo:
The issue tracker is for bugs/feature-requests, not support questions – those are better handled at the project discussion list: https://groups.google.com/forum/#!forum/gensim

And, you'd have to provide a lot more context/code/logging-info for us to have any idea what line of your code is triggering that error. So if you ask on the list, please better describe what you're trying to accomplish, and how.

piskvorky:
Sounds like a bug report... but very incomplete.

@Tesfamariam, please review the [contributing guide](https://github.com/piskvorky/gensim/blob/develop/CONTRIBUTING.md). Add relevant information so we know what you're talking about.

Tesfamariam:
Sorry for the incomplete information! Sample dataset:
['lecture', 'notes', 'edited', 'goos', 'hartmanis', 'van', 'leeuwen', 'berlin', 'heidelberg', 'york', 'barcelona', 'hong', 'kong', 'london', 'milan', 'paris', 'singapore', 'tokyo', 'vassil', 'alexandrov', 'jack', 'dongarra', 'benjoe', 'juliano', 'renner', 'kenneth', 'tan', 'eds', 'san', 'francisco', 'usa', 'proceedings', 'volume', 'editors', 'vassil', 'alexandrov', 'university', 'reading', 'school', 'cybernetics', 'electronic', 'engineering', 'whiteknights', 'box', 'reading', 'mail', 'alexandrov', 'rdg', 'jack', 'dongarra']
Then I feed the whole dataset to:
class DTMcorpus(corpora.textcorpus.TextCorpus):

    def get_texts(self):
        return self.input

    def __len__(self):
        return len(self.input)

corpus = DTMcorpus(texts)
Then determined the time slices:
my_timeslices = [1, 1, 1, 1, 1, 1, 1, 1, 1, 1,1,1, 1, 1, 1]
model = gensim.models.wrappers.DtmModel('/media/tesfish/data/Topic Modeling/dtm-master/bin/dtm-linux64', corpus, my_timeslices, num_topics=15, id2word=dictionary_text, initialize_lda=True)
finally I got the following error:
TypeError                                 Traceback (most recent call last)
[...]

PR #768, opened 2016-06-30
Str to byte fix for DTM wrapper.

bhargavvader:
@piskvorky , this is with respect to #698.

You were right, all that had to be done was use `utils.to_utf8` before writing to file. 
Works fine with both python2 and 3.

### DIFF
--- a/gensim/models/wrappers/dtmmodel.py
+++ b/gensim/models/wrappers/dtmmodel.py
@@ -171,6 +171,7 @@ class DtmModel(utils.SaveLoad):
         logger.info("serializing temporary corpus to %s" % self.fcorpustxt())
         # write out the corpus in a file format that DTM understands:
         corpora.BleiCorpus.save_corpus(self.fcorpustxt(), corpus)
+
         with utils.smart_open(self.ftimeslices(), 'wb') as fout:
             fout.write(six.u(utils.to_utf8(str(len(self.time_slices)) + "\n")))
             for sl in time_slices:
--- a/gensim/models/wrappers/dtmmodel.py
+++ b/gensim/models/wrappers/dtmmodel.py
@@ -173,9 +173,9 @@ class DtmModel(utils.SaveLoad):
         corpora.BleiCorpus.save_corpus(self.fcorpustxt(), corpus)
 
         with utils.smart_open(self.ftimeslices(), 'wb') as fout:
-            fout.write(six.u(str(len(self.time_slices)) + "\n"))
+            fout.write(six.u(str(len(self.time_slices)) + "\n").encode('utf-8'))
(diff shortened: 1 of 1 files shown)


bhargavvader:
Note: would ideally want to convert to bytes from the list itself instead of converting to string and then to byte.

bhargavvader:
@tmylk , I can't directly convert from integer to bytes while writing to file, it has to be to string first.
For e.g, `bytes(4)` is `b'\x00\x00\x00\x00'`, but `str(4).encode('utf-8')` is `b'4'` as expected. 

Is there any idea for a workaround to this?

gensim/models/wrappers/dtmmodel.py
```diff
@@ -173,9 +173,9 @@ def convert_input(self, corpus, time_slices):
         corpora.BleiCorpus.save_corpus(self.fcorpustxt(), corpus)
 
         with utils.smart_open(self.ftimeslices(), 'wb') as fout:
-            fout.write(six.u(str(len(self.time_slices)) + "\n"))
+            fout.write(six.u(utils.to_utf8(str(len(self.time_slices)) + "\n")))
```
piskvorky: I don't get this. Doesn't `six.u` return unicode? How does this work?

gensim/models/wrappers/dtmmodel.py
```diff
@@ -173,9 +173,9 @@ def convert_input(self, corpus, time_slices):
...
             for sl in time_slices:
-                fout.write(six.u(str(sl) + "\n"))
+                fout.write(six.u(utils.to_utf8(str(sl) + "\n")))
```
piskvorky: dtto

piskvorky:
@bhargavvader we want to store binary strings, not "bytes" as such. We don't want to be storing numbers as raw bytes here, as in `b'\x00\x00\x00\x04'`. We want to store their (binary) string representation, as in `b'4'`. we just don't want to be storing these strings as unicode, as in `u'4'`.

Integers are always representable in ASCII encoding (subset of utf8), so the bytes/unicode conversion is no problem.

bhargavvader:
@piskvorky , [here](https://pythonhosted.org/six/) it says that `six.u` returns unicode in python 2 and string in python 3.

I understand we'd want to store it in binary strings, yeah.
So can the `six.u` be removed? Can it just be `utils.to_utf8(str())` in all places?

piskvorky:
I don't understand how this version with storing unicode to binary files even worked. It means our unit tests must be faulty / incomplete.

klvbdmh:
Were the changes pushed to PyPI? Pip installs version 0.13.1. and the bug still persists.

Issue #769, opened 2016-07-02
Usage of topn in dtmmodel breaks print_topics and show_topics.


bhargavvader:
After #755 was merged, `topn` was changed to be `num_words`, but because this was not changed everywhere (line [283](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/wrappers/dtmmodel.py#L283), line [300](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/wrappers/dtmmodel.py#L300)) it would throw up `NameError: name 'topn' is not defined` whenever `self.print_topics` or `self.show_topics` was called (because of line [274](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/wrappers/dtmmodel.py#L274)).

This pretty much means that dtm wrapper is unusable wrt displaying any topics, putting in a PR to fix this.

piskvorky:
@tmylk how do you review these PRs before merging?

There are too many errors, we cannot merge code so carelessly.

bhargavvader:
No, this was after that in 0.13.2. 
I noticed it because when I was testing the #768 solution, `print_topics` was failing.

PR #770, opened 2016-07-02
DTM wrapper bug fixes. 

bhargavvader:
Changed `topn` for `num_words`, wrt #769 .

While the same problem as in LdaMallet wrapper (PR #767 ) for it not being backward compatible is there, the PR #755 already made the non-backward compatible changes so I don't think this should be a problem. (@piskvorky , please have a look at this)

At any rate, either only `topn` or `num_words` should be there (not half and half), the `print_topics` function will not work otherwise.

I've also removed `six.u` with respect to #768 as well.

Edit: Made changes to the DTM notebook to reflect this as well.

### DIFF
--- a/docs/notebooks/dtm_example.ipynb
+++ b/docs/notebooks/dtm_example.ipynb
@@ -41,15 +41,7 @@
    "metadata": {
     "collapsed": false
    },
-   "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "DEBUG:root:test\n"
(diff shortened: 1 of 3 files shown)


bhargavvader:
@piskvorky , @tmylk , could you review?

piskvorky:
I like `topn` (or perhaps `top_n`) a little better, because it's not dependent on what the features represent (words, phrases, entities, characters...).

But it's not a big deal either way. We should just be consistent throughout, whichever terminology we use.

PR #771, opened 2016-07-02
LdaMallet Fixes

bhargavvader:
This is with respect to #767 .

I removed the unnecessary log mentioning the variable assignment, and tried to add an alias by doing `self.word_topics = self.wordtopics = self.load_word_topics()`. 
Is that correct?

I also changed occurrences of `topn` to `num_words` to keep it consistent throughout. #755 already did this, but left `print_topic` and `show_topic` untouched. It doesn't cause any problems but I thought I'd make it the same throughout.

@tmylk , @piskvorky , please review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Changes
 =======
 0.13.2
 
+* wordtopics has changed to word_topics in ldamallet, and fixed issue #764. (@bhargavvader, #771) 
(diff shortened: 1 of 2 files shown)


piskvorky:
Looks good to me... except still needs a comment explaining why the alias is there.

And maybe a mention in the changelog, so we can deprecate the old name, gradually.

bhargavvader:
Added comment, made change in changelog.

PR #786, opened 2016-07-13
Description added for America's Next Topic Model slides

droudy:


droudy:
@tmylk  test_dbow_neg (gensim.test.test_doc2vec.TestDoc2VecModel) failed for python 3.4

PR #788, opened 2016-07-13
Fixing the docstrings in WmdSimilarity and wmdistance

olavurmortensen:
@tmylk @piskvorky 

I noticed that I had formatted the docstrings of `WmdSimilarity` and `wmdistance` incorrectly so that their decriptions on the API reference looked hideous (see [here](https://radimrehurek.com/gensim/models/word2vec.html#gensim.models.word2vec.Word2Vec.wmdistance) and [here](https://radimrehurek.com/gensim/similarities/docsim.html#gensim.similarities.docsim.WmdSimilarity)).

So I took the liberty of fixing these (I can't test that they actually will look correct on the Gensim website though). Do the docstrings look correct?

I also noticed two small errors while fixing this (two unused parameters in wmdistance and an error in one of the docstrings), so I fixed that too.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1244,7 +1244,7 @@ class Word2Vec(utils.SaveLoad):
         result = [(self.index2word[sim], float(dists[sim])) for sim in best if sim not in all_words]
         return result[:topn]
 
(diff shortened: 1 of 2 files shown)


tmylk:
Thanks!

PR #792, opened 2016-07-15
Changed Annoy implementation 

droudy:
Added more complete functionality. `Word2Vec.most_similar()` and `Doc2Vec.most_similar()` can now take an indexer as a parameter(annoy indexer is the only one supported as of now). Ex:
<pre><code>
indexer = gensim.similarities.index.AnnoyIndexer(example_model, 100)
example_model.most_similar([example_vector], topn=5, indexer=indexer)
</pre></code>

The annoytutorial notebook has been updated to use this API


droudy:
Thanks to @karlhigley for his contributions

gensim/models/word2vec.py
```diff
@@ -1235,6 +1235,9 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None):
             raise ValueError("cannot compute similarity with no input")
         mean = matutils.unitvec(array(mean).mean(axis=0)).astype(REAL)
 
+        if indexer is not None:
+            return indexer.most_similar(mean, topn)
+
```
gojomo: Just a thought for potential future refactorings: this callback is suggestive that the lookup/positive-negative/mean vector-combination steps may deserve being factored out to its own method, say `combo()`. 

Then instead of `model.most_similar(positive, negative)`, a more-explicit idiom would be `model.most_similar(model.combo(positive, negative))`. And rather than `model.most_similar(positive, negative, indexer=indexer)` it'd be `indexer.most_similar(model.combo(positive, negative)`.

Issue #765, opened 2016-06-29
LDA u_mass coherence calculation triggers numpy deprecation warning


masscognition:
Using Python 2.7.11, latest gensim , latest numpy.

Line 41 of segmentation.py:

for w_star in top_words[:w_prime_index]:

triggers this warning from numpy:

DeprecationWarning: converting an array with ndim > 0 to an index will result in an error in the future

fix is to cast w_prime_index to int 

Also via Devashish: 

"the comment right next to

w_prime_index = np.where(top_words == w_prime)[0] should be # To get index of w_prime in top_words (apologies for my mistake)."

dsquareindia:
Thanks for raising this issue! Would you mind raising a PR fix for this? You can follow a workflow similar to that given here: http://scikit-learn.org/stable/developers/contributing.html#contributing-code. Would be happy to help if you encounter any problems.

dsquareindia:
Sorry @masscognition I addressed this issue in #793 since I happened to be working on the same thing. Thanks for raising the issue though!

masscognition:
No problem :-)

PR #750, opened 2016-06-23
Added vowpalwabbit wrapper to notebook.

dsquareindia:
Adds wrapper example to `topic_coherence_notebook`

PR #793, opened 2016-07-16
[MRG] Topic coherence update 3

dsquareindia:
Changes:
- Added backtracking dictionary for storing the context vector for (w_prime or w_star, w) pair. Each such tuple is mapped one to one to a context vector. Earlier each context vector was calculated every time. Changed window size to 110.
- Added `c_uci` coherence measure.
- Added `c_npmi` coherence measure.
- Added `window_size` parameter to CoherenceModel init.

### DIFF
--- a/gensim/models/coherencemodel.py
+++ b/gensim/models/coherencemodel.py
@@ -166,6 +166,6 @@ class CoherenceModel(interfaces.TransformationABC):
         elif self.coherence == 'c_v':
             segmented_topics = self.seg(self.topics)
             per_topic_postings, num_windows = self.prob(texts=self.texts, segmented_topics=segmented_topics,
-                                                        dictionary=self.dictionary, window_size=2)  # FIXME : Change window size to 110 finally.
+                                                        dictionary=self.dictionary, window_size=110)
             confirmed_measures = self.conf(self.topics, segmented_topics, per_topic_postings, 'nlr', 1, num_windows)
             return self.aggr(confirmed_measures)
--- a/gensim/models/coherencemodel.py
+++ b/gensim/models/coherencemodel.py
@@ -31,10 +31,38 @@ from gensim.models.wrappers import LdaVowpalWabbit, LdaMallet
 
 import numpy as np
 
+from collections import namedtuple
+
 logger = logging.getLogger(__name__)
 
 boolean_document_based = ['u_mass']
 sliding_window_based = ['c_v', 'c_uci', 'c_npmi']
+make_pipeline = namedtuple('Coherence_Measure', 'seg, prob, conf, aggr')
+
+coherence_dict = {
+    'u_mass': make_pipeline(segmentation.s_one_pre,
+                            probability_estimation.p_boolean_document,
+                            direct_confirmation_measure.log_conditional_probability,
+                            aggregation.arithmetic_mean),
+    'c_v': make_pipeline(segmentation.s_one_set,
+                         probability_estimation.p_boolean_sliding_window,
+                         indirect_confirmation_measure.cosine_similarity,
+                         aggregation.arithmetic_mean),
+    'c_uci': make_pipeline(segmentation.s_one_one,
+                           probability_estimation.p_boolean_sliding_window,
+                           direct_confirmation_measure.log_ratio_measure,
+                           aggregation.arithmetic_mean),
+    'c_npmi': make_pipeline(segmentation.s_one_one,
+                            probability_estimation.p_boolean_sliding_window,
+                            direct_confirmation_measure.log_ratio_measure,
+                            aggregation.arithmetic_mean),
+}
+
+sliding_windows_dict = {
+    'c_v': 110,
+    'c_uci': 10,
+    'c_npmi': 10
+}
 
 class CoherenceModel(interfaces.TransformationABC):
     """
@@ -59,7 +87,7 @@ class CoherenceModel(interfaces.TransformationABC):
 
     Model persistency is achieved via its load/save methods.
     """
-    def __init__(self, model=None, topics=None, texts=None, corpus=None, dictionary=None, window_size=None, coherence='c_v'):
+    def __init__(self, model=None, topics=None, texts=None, corpus=None, dictionary=None, window_size=None, coherence='c_v', topn=10):
         """
         Args:
         ----
@@ -84,6 +112,7 @@ class CoherenceModel(interfaces.TransformationABC):
                     'c_npmi'
                     For 'u_mass' corpus should be provided. If texts is provided, it will be converted to corpus using the dictionary.
                     For 'c_v', 'c_uci' and 'c_npmi' texts should be provided. Corpus is not needed.
+        topn : Integer corresponding to the number of top words to be extracted from each topic.
         """
         if model is None and topics is None:
             raise ValueError("One of model or topics has to be provided.")
@@ -118,7 +147,7 @@ class CoherenceModel(interfaces.TransformationABC):
                 self.texts = texts
         else:
             raise ValueError("%s coherence is not currently supported." % coherence)
-
+        self.topn = topn
         self.model = model
         if model is not None:
             self.topics = self._get_topics()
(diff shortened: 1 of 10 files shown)


dsquareindia:
@tmylk should I add the benchmark testing notebooks for 20NG and Movies dataset too?

dsquareindia:
Correcting the tests.

dsquareindia:
@tmylk I've added `c_uci`, `c_npmi` and `window_size` parameter to `CoherenceModel` init. This pr also addresses issue #765.

dsquareindia:
@tmylk I've added the benchmark testing notebook on movies dataset.

gensim/models/coherencemodel.py
```diff
@@ -91,7 +101,7 @@ def __init__(self, model=None, topics=None, texts=None, corpus=None, dictionary=
         else:
             self.dictionary = dictionary
```
tmylk: what is the point of checking `if isinstance(model.id2word, FakeDict):` above? why is it note enough to check for None?

gensim/models/coherencemodel.py
```diff
@@ -132,6 +143,18 @@ def __init__(self, model=None, topics=None, texts=None, corpus=None, dictionary=
             self.conf = indirect_confirmation_measure.cosine_similarity
             self.aggr = aggregation.arithmetic_mean
 
+        elif self.coherence == 'c_uci':
```
tmylk: Above use `enumerate` instead of `for t in range(len(topic)):`

gensim/models/coherencemodel.py
```diff
@@ -132,6 +143,18 @@ def __init__(self, model=None, topics=None, texts=None, corpus=None, dictionary=
...
+            self.seg = segmentation.s_one_one
+            self.prob = probability_estimation.p_boolean_sliding_window
+            self.conf = direct_confirmation_measure.log_ratio_measure
+            self.aggr = aggregation.arithmetic_mean
+
+        elif self.coherence == 'c_npmi':
```
tmylk: Better to create a data structure(dictionary?) of coherence measures and look up by name there. It would make it easier to see the difference between them.

gensim/models/coherencemodel.py
```diff
@@ -132,6 +143,18 @@ def __init__(self, model=None, topics=None, texts=None, corpus=None, dictionary=
...
+            self.seg = segmentation.s_one_one
+            self.prob = probability_estimation.p_boolean_sliding_window
+            self.conf = direct_confirmation_measure.normalized_log_ratio_measure
+            self.aggr = aggregation.arithmetic_mean
+
     def __str__(self):
         return "CoherenceModel(segmentation=%s, probability estimation=%s, confirmation measure=%s, aggregation=%s)" % (
             self.seg, self.prob, self.conf, self.aggr)
```
tmylk: Below in `_get_topics`, make `topn` a configurable parameter. Why do you use 10?

tmylk: Also how does `_get_topics` below deal with @bhargavvader 's DTM code? He had to make some adjustments to fit it in.

gensim/models/coherencemodel.py
```diff
@@ -164,8 +187,28 @@ def get_coherence(self):
             return self.aggr(confirmed_measures)
 
         elif self.coherence == 'c_v':
+            if self.window_size is None:
```
tmylk: can we have just two ifs - whether it is window-based or not? It seems some code duplication here.

gensim/test/test_coherencemodel.py
```diff
@@ -77,6 +77,18 @@ def testCv(self):
         cm2 = CoherenceModel(topics=self.topics2, texts=texts, dictionary=dictionary, coherence='c_v')
         self.assertTrue(cm1.get_coherence() > cm2.get_coherence())
 
+    def testCuci(self):
```
tmylk: code duplication. just parametrise by coherence measure name.

gensim/test/test_coherencemodel.py
```diff
@@ -77,6 +77,18 @@ def testCv(self):
...
+
+    def testCnpmi(self):
+        """Test C_npmi topic coherence algorithm on given topics"""
+        cm1 = CoherenceModel(topics=self.topics1, texts=texts, dictionary=dictionary, coherence='c_npmi')
+        cm2 = CoherenceModel(topics=self.topics2, texts=texts, dictionary=dictionary, coherence='c_npmi')
+        self.assertTrue(cm1.get_coherence() > cm2.get_coherence())
+
     def testUMassLdaModel(self):
         """Perform sanity check to see if u_mass coherence works with LDA Model"""
         try:
```
tmylk: this test should be more than just checking for exception raised. can you do `self.assertTrue(lda1.get_coherence() > lda2.get_coherence())` here?

gensim/topic_coherence/direct_confirmation_measure.py
```diff
@@ -96,7 +96,7 @@ def normalized_log_ratio_measure(segmented_topics, per_topic_postings, num_docs)
             w_star_docs = per_topic_postings[w_star]
             co_docs = w_prime_docs.intersection(w_star_docs)
             co_doc_prob = len(co_docs) / float(num_docs)
-            m_nlr_i = numerator / np.log(co_doc_prob + EPSILON)
+            m_nlr_i = numerator / (-np.log(co_doc_prob + EPSILON))
             m_nlr.append(m_nlr_i)
```
tmylk: looks like code duplicated between normalized and not-normalized `log-ratio` measure. could you just add an `if` statement on `normalize=True`?

gensim/topic_coherence/indirect_confirmation_measure.py
```diff
@@ -26,6 +26,23 @@
 
 logger = logging.getLogger(__name__)
 
+
```
tmylk: Please make valid latex by putting { and } in the formulas
 `m_sim(m, gamma)(W', W*) = s_sim(V_{m,gamma}(W'), V_{m,gamma}(W*))`

tmylk: add description for  `sigma` in ` {sigma(w' belonging to W')`

gensim/models/coherencemodel.py
dsquareindia: `None` doesn't work here actually. When no word->id mapping is provided while creating an LdaModel [this](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ldamodel.py#L279) line is called which returns a `FakeDict` from [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/utils.py#L558). So this code:
```python
tm1 = LdaModel(corpus=corpus, num_topics=2, iterations=0, passes=0, random_state=10)
if tm1.id2word is None:
    print 'aye'
else:
    print 'naye'
```
actually prints `naye` but if I change it to `isinstance(tm1.id2word, FakeDict)` it outputs correctly.
Am I correct here?

dsquareindia:
@tmylk I've addressed your initial comments. I hope I've addressed them correctly.

gensim/topic_coherence/indirect_confirmation_measure.py
dsquareindia: Done.

dsquareindia:
@tmylk should I change all `Args` to `Parameters`?

gensim/topic_coherence/indirect_confirmation_measure.py
```diff
@@ -8,12 +8,12 @@
 This module contains functions to compute confirmation on a pair of words or word subsets.
```
tmylk: Explain how the indirect confirmation measures work and why they are useful similar to the competing car brands explanation in the paper.

gensim/topic_coherence/indirect_confirmation_measure.py
```diff
@@ -26,6 +26,23 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _present(w_prime_star, w, w_backtrack):
+    """
+    Internal helper function to return index of (w_prime_star, w) in w_backtrack.
+    Return -1 if not present.
+    """
+    index = -1
```
tmylk: isn't this just a dictionary lookup?

dsquareindia: yes but I have to maintain `w_prime_star` and `w` tuple in a list which is `w_backtrack` since numpy arrays are unhashable. So this is basically a lookup of a numpy tuple inside the backtracking list. What could be a better way to do this?

gensim/topic_coherence/probability_estimation.py
```diff
@@ -95,8 +95,6 @@ def add_topic_posting():
         it = iter(document)
         window = tuple(islice(it, window_size))
         add_topic_posting()
-        if len(window) <= window_size:
-            pass  # FIXME : Handle case when window size is bigger than length of document
         for elem in it:
             window = window[1:] + (elem,)
             add_topic_posting()
```
tmylk: dont' use 'global' variables in `add_topic_posting` - pass and return them.

gensim/topic_coherence/segmentation.py
```diff
@@ -37,13 +37,48 @@ def s_one_pre(topics):
     for top_words in topics:
         s_one_pre_t = []
         for w_prime in top_words[1:]:
-            w_prime_index = np.where(top_words == w_prime)[0] # To get index of w_star in top_words
+            w_prime_index = int(np.where(top_words == w_prime)[0])  # To get index of w_prime in top_words
```
tmylk: can you just use enumerat to get both item and index?

gensim/topic_coherence/segmentation.py
```diff
@@ -37,13 +37,48 @@ def s_one_pre(topics):
...
+    Returns:
+    -------
+    s_one_one : list of list of (W', W*) tuples for all unique topic ids
+    """
+    s_one_one = []
+
+    for top_words in topics:
+        s_one_one_t = []
+        for w_prime in top_words:
+            w_prime_index = int(np.where(top_words == int(w_prime))[0])  # To get index of w_prime in top_words
```
tmylk: same as above

dsquareindia:
@tmylk I've addressed your comments.

tmylk:
Merged in https://github.com/RaRe-Technologies/gensim/commit/6f53b31b6f04b2c059c29a7744ecad5e293ab579

dsquareindia:
Linking to #750.

Issue #714, opened 2016-06-01
Change word2vec default to avoid lowercasing


tmylk:
Make default `model.accuracy("file", use_lowercase=False)`.
to fix the [mailing list question](https://groups.google.com/d/msg/gensim/Zq-ZAutQdF8/ORbw1KKBBAAJ)

See related PR https://github.com/piskvorky/gensim/pull/607

gojomo:
Even though this fixes things for the case of loading the GoogleNews vectors, I'm not sure this will improve things other common cases – people often lowercase their training data.

piskvorky:
What does the C tool do, these days? Let's emulate that.

gojomo:
A quick glance suggests compute-accuracy.c in the original package force-uppercases for all string-comparisons, with the effect that they count the analogy as 'correct' if the closest-word is any of the potentially case-varied versions in the vocabulary. Unfortunately, the `use_lowercase` parameter forced in either way doesn't quite emulate that behavior – the comparison of `predicted` & `expected` would need to be changed instead. Still pretty easy – but the parameter should probably be named 'case_insensitive' instead, and not just transform the question-words.

piskvorky:
OK, let's do that.

I'll leave this issue open with the `easy` label.

jayantj:
Came across this issue too when I noticed the gensim word2vec model using a lower number of analogies for computing accuracy, as compared to some other implementations. It might not be as easy as making the comparison of `predicted` and `expected` case-insensitive. 
```
try:
    if use_lowercase:
        a, b, c, expected = [word.lower() for word in line.split()]  # assumes vocabulary preprocessing uses lowercase, too...
    else:
        a, b, c, expected = [word for word in line.split()]
except:
    logger.info("skipping invalid line #%i in %s" % (line_no, questions))
if a not in ok_vocab or b not in ok_vocab or c not in ok_vocab or expected not in ok_vocab:
    logger.debug("skipping line #%i with OOV words: %s" % (line_no, line.strip()))
    continue
```
[Source](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec.py#L1602-L1626)

The key check of the analogy word existing in the vocab would return false and the analogy would be skipped.

The original C code works around this by creating an all-uppercase vocabulary from the model files (specifically for `compute-accuracy.c`), and uppercasing all analogy words too.

[...]

PR #804, opened 2016-07-25
Case insensitive option for word2vec accuracy 

jayantj:
Related to #714 

Quick PR to get some feedback. Also, should we have tests for this? There aren't any currently, which seems fine to me since `accuracy` is an informal evaluation metric rather than basic functionality

A couple of concerns mentioned as comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Changes
 * Implemented LsiModel.docs_processed attribute
 * Added LdaMallet support. Added LdaVowpalWabbit, LdaMallet example to notebook. Added test suite for coherencemodel and aggregation.
   Added `topics` parameter to coherencemodel. Can now provide tokenized topics to calculate coherence value (@dsquareindia, #750)
+* Changed `use_lowercase` option in word2vec accuracy to `case_insensitive` to account for case variations in training vocabulary (@jayantj, #714)
 
 0.13.1, 2016-06-22
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1576,8 +1576,8 @@ class Word2Vec(utils.SaveLoad):
         Use `restrict_vocab` to ignore all questions containing a word whose frequency
         is not in the top-N most frequent words (default top 30,000).
 
-        Use `use_lowercase` to convert all words in questions to thier lowercase form before evaluating
-        the accuracy. It's useful when assuming the text preprocessing also uses lowercase.  (default True).
+        Use `case_insensitive` to convert all words in questions and vocab to their lowercase form before evaluating
+        the accuracy. Useful in case of case-mismatch between training tokens and question words.  (default True).
 
         This method corresponds to the `compute-accuracy` script of the original C word2vec.
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1573,8 +1573,10 @@ class Word2Vec(utils.SaveLoad):
         The accuracy is reported (=printed to log and returned as a list) for each
         section separately, plus there's one aggregate summary at the end.
 
-        Use `restrict_vocab` to ignore all questions containing a word whose frequency
-        is not in the top-N most frequent words (default top 30,000).
+        `restrict_vocab` is an optional integer which limits the vocab to be used
+        for answering questions. For example, restrict_vocab=10000 would only check
+        the first 10000 word vectors in the vocabulary order. (This may be meaningful
+        if you've sorted the vocabulary by descending frequency.)
 
         Use `case_insensitive` to convert all words in questions and vocab to their lowercase form before evaluating
         the accuracy. Useful in case of case-mismatch between training tokens and question words.  (default True).
@@ -1582,11 +1584,9 @@ class Word2Vec(utils.SaveLoad):
         This method corresponds to the `compute-accuracy` script of the original C word2vec.
 
         """
-        ok_vocab = dict(sorted(iteritems(self.vocab),
-                               key=lambda item: -item[1].count)[:restrict_vocab])
+        ok_vocab = [(w, self.vocab[w]) for w in self.index2word[:restrict_vocab]]
+        ok_vocab = dict((w.lower(), v) for w, v in reversed(ok_vocab)) if case_insensitive else dict(ok_vocab)
         ok_index = set(v.index for v in itervalues(ok_vocab))
-        if case_insensitive:
-            ok_vocab = dict((w.lower(), v) for w, v in iteritems(ok_vocab))
 
         sections, section = [], None
         for line_no, line in enumerate(utils.smart_open(questions)):
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1573,13 +1573,15 @@ class Word2Vec(utils.SaveLoad):
         The accuracy is reported (=printed to log and returned as a list) for each
         section separately, plus there's one aggregate summary at the end.
 
-        `restrict_vocab` is an optional integer which limits the vocab to be used
-        for answering questions. For example, restrict_vocab=10000 would only check
-        the first 10000 word vectors in the vocabulary order. (This may be meaningful
-        if you've sorted the vocabulary by descending frequency.)
+        Use `restrict_vocab` to ignore all questions containing a word not in the first `restrict_vocab`
+        words (default 30,000). This may be meaningful if you've sorted the vocabulary by descending frequency.
+        In case `case_insensitive` is True, the first `restrict_vocab` words are taken first, and then
+        case normalization is performed.
 
-        Use `case_insensitive` to convert all words in questions and vocab to their uppercase form before evaluating
-        the accuracy. Useful in case of case-mismatch between training tokens and question words.  (default True).
+        Use `case_insensitive` to convert all words in questions and vocab to their uppercase form before 
+        evaluating the accuracy (default True). Useful in case of case-mismatch between training tokens 
+        and question words. In case of multiple case variants of a single word, the vector for the first
+        occurrence (also the most frequent if vocabulary is sorted) is taken.
 
         This method corresponds to the `compute-accuracy` script of the original C word2vec.
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1608,6 +1608,7 @@ class Word2Vec(utils.SaveLoad):
                         a, b, c, expected = [word for word in line.split()]
                 except:
                     logger.info("skipping invalid line #%i in %s" % (line_no, questions))
+                    continue
                 if a not in ok_vocab or b not in ok_vocab or c not in ok_vocab or expected not in ok_vocab:
                     logger.debug("skipping line #%i with OOV words: %s" % (line_no, line.strip()))
                     continue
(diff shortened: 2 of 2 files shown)


gensim/models/word2vec.py
```diff
@@ -1610,13 +1612,16 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=most_similar, u
                     logger.debug("skipping line #%i with OOV words: %s" % (line_no, line.strip()))
                     continue
 
+                original_vocab = self.vocab
+                self.vocab = ok_vocab
                 ignore = set(self.vocab[v].index for v in [a, b, c])  # indexes of words to ignore
                 predicted = None
                 # find the most likely prediction, ignoring OOV words and input words
                 sims = most_similar(self, positive=[b, c], negative=[a], topn=False, restrict_vocab=restrict_vocab)
```
jayantj: The behaviour for `restrict_vocab` in `most_similar` is a little different from its behaviour in `accuracy`
In accuracy, it simply takes the first `restrict_vocab` vectors for calculating similarity, and in `accuracy`, `ok_vocab` is created by sorting by frequency and taking first `restrict_vocab` words.

This could cause some unintended behaviour, especially with low values for `restrict_vocab` since it is passed to `most_similar` from accuracy. The two sets of vocab used in the two methods could be significantly different, and possibly even completely disjoint.

Setting `self.vocab = ok_vocab` before the `most_similar` call should handle this, I think?

gensim/models/word2vec.py
```diff
@@ -1610,13 +1612,16 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=most_similar, u
...
+                self.vocab = original_vocab
                 for index in matutils.argsort(sims, reverse=True):
```
jayantj: The exact same functionality is present as part of `most_similar` in case `topn` is true valued. Would it be better to simply pass `topn` as 1 and do a string comparison? Can get rid of `ok_index` and `ignore` in that case too

gensim/models/word2vec.py
```diff
@@ -1600,7 +1602,7 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=most_similar, u
                 if not section:
                     raise ValueError("missing section header before line #%i in %s" % (line_no, questions))
                 try:
-                    if use_lowercase:
+                    if case_insensitive:
```
jayantj: Missing `continue` after line 1610? (Sorry, I wasn't able to comment at that line)

gensim/models/word2vec.py
```diff
@@ -1576,15 +1576,17 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=most_similar, u
         Use `restrict_vocab` to ignore all questions containing a word whose frequency
         is not in the top-N most frequent words (default top 30,000).
 
-        Use `use_lowercase` to convert all words in questions to thier lowercase form before evaluating
-        the accuracy. It's useful when assuming the text preprocessing also uses lowercase.  (default True).
+        Use `case_insensitive` to convert all words in questions and vocab to their lowercase form before evaluating
```
gojomo: It shouldn't make a difference in results, but in the spirit of mimicking `compute-accuracy.c` unless there's a good reason not to, perhaps we should case-normalize to uppercase?

gensim/models/word2vec.py
gojomo: Yes, I think the "skipping invalid line" log-message should be followed with a continue to definitely skip.

gensim/models/word2vec.py
gojomo: Yes, I agree it's currently a subtle bug that this call to `most_similar()` passes the `restrict_vocab` argument, since it may not mean the exact same thing. (It's not always the case that the word-vectors have already been frequency-sorted.) It's only truly correct to pass the count in if the interpretation is the same. 

But, it's likely a big speedup, in cases where the vocabulary is large compared to the `restrict_vocab`, as it prevents lots of unnecessary distance-calculations in the `dot(...)` line of `most_similar()`. (Executing `most_similar()` with a filtered `vocab` doesn't provide the same optimization.) 

Perhaps we just want to document that for `restrict_vocab` to work as expected here, the vocabulary must be sorted? (Or even redefine the argument here to mean limiting searches to the 1st indexes, whether they're the most-frequent or not. So no extra sort-by-frequency here.)

gensim/models/word2vec.py
gojomo: I think there's still a subtle difference with the `most_similar(topn=1)` behavior, especially in the case-flattening scenario: all words that are case-flattened equivalent to the `ignore` words should plausibly be ignored. We don't do this now... but `compute-accuracy.c` does. (Unsure how best to do this. Maybe there needs to be a case-flattened temporary `index2word` list, too.)

gensim/models/word2vec.py
jayantj: Hmm, yes that's a good catch, I hadn't noticed a filtered `vocab` doesn't prevent all those vector multiplications.

I think the latter makes most sense, especially seeing that `sort_vocab` is true by default in the word2vec initialization. Doesn't seem to make sense to perform an operation like sorting on the entire vocab for temporary use.

gensim/models/word2vec.py
jayantj: All right

gensim/models/word2vec.py
jayantj: How about instead of checking if `index` exists in `ok_index` and not in `ignore`, we check if `self.index2word[index]` exists in ok_vocab and not in `ignore` (of course, changing `ignore` to a set of words instead of indices?

gojomo: Makes sense! 

But, also makes me realize another thorny weirdness of the case-flattened approach. Not just the target words, but the 3 supplied-analogy words may have multiple-case versions. And the one version of those words that survives case-normalization into `ok_vocab` may in fact (due to the arbitrariness of the dict-iteration order) be a less-frequent variant. (For example, a vector index for 'Usa' might be the one that lands in `ok_vocab` over potentially more frequent variants 'USA' or even 'usa'.)

I *think* a principled solution could be to build the case-flattened `ok_vocab` from `index2word`, clipped to `restrict_vocab`, in reverse-order (so last-seen/most-frequent wins). I think this will even match `compute-accuracy.c`'s behavior, because all of its lookups are linear-probes from the beginning of the loaded word-vectors. Does that seem correct to you?

jayantj: Yes, that seems correct. 
Just making sure I understand - in case `sort_vocab` is `True` (which it is, by default), using `index2word` to build `ok_vocab` would imitate `compute_accuracy.c`'s behaviour, since the most frequent version of the analogy words would be taken.

In case `sort_vocab` is `False`, the last-seen version of the analogy words would be taken, therefore not quite imitating `compute_accuracy`'s behaviour.

jayantj:
@gojomo Made some changes, please have a look?

gensim/models/word2vec.py
```diff
@@ -1573,18 +1573,19 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=most_similar, u
         The accuracy is reported (=printed to log and returned as a list) for each
         section separately, plus there's one aggregate summary at the end.
 
-        Use `restrict_vocab` to ignore all questions containing a word whose frequency
-        is not in the top-N most frequent words (default top 30,000).
+        `restrict_vocab` is an optional integer which limits the vocab to be used
+        for answering questions. For example, restrict_vocab=10000 would only check
+        the first 10000 word vectors in the vocabulary order. (This may be meaningful
+        if you've sorted the vocabulary by descending frequency.)
```
tmylk: Please add a comment to what happens for `case_insensitive` and `restrict_vocab` together.

tmylk:
If you add changelog and tests than it would look good to me.
Waiting for @gojomo to review too.

gensim/models/word2vec.py
gojomo: Current comment no longer mentions default of 30,000 in effect.

gojomo:
I think latest changes match compute-accuracy.c behavior, but would be good to test against it, using same sets of word2vec.c-format-vectors with mixed-cases. If we've matched behavior exactly, then should get exact-same correct-counts for many values of threshold/restrict_vocab, or both case_insensitive options.

gensim/models/word2vec.py
jayantj: Done

jayantj:
Yeah, that makes sense. What do you think the best approach would be, adding pretrained word2vec models to the tests and checking their accuracy against hardcoded values (obtained from running `compute-accuracy.c` locally), or adding the compute-accuracy binary to the tests and automating everything?

piskvorky:
@jayantj Definitely *not* the latter.

gojomo:
I was thinking the need could be met by running a few tests manually – for example using the GoogleNews vectors, which are too big to bundle (or even download/load) during automated unit testing. 

I suppose a toy-sized set of word-vectors (say just enough to get a few analogies right/wrong) could also be bundled for testing purposes, but it would have to be hand-crafted to trigger problems for any deviation from the `compute-accuracy.c` behaviors with regard to threshold/case.

tmylk:
Waiting for manual verification results on GoogleNews vectors and a small toy mixed case dataset.

jayantj:
Manually verified on a midsize mixed case corpus (brown) for multiple values of `restrict_vocab`
I'm running into memory issues while verifying on GoogleNews vectors
Verified for text8 though

Are we skipping on automated tests then?

tmylk:
Could you link to a gist or notebook  with your manual results?

jayantj:
Here's the [ipython notebook](https://gist.github.com/jayantj/17ff697ca984bf6517975f1225b5edc4)
I've compared gensim trained models for text8 and gutenberg (I've modified the compute accuracy printf statements a little, the output is still not very readable though)

Found a new error that'd been missed previously while training on gutenberg - 

Suppose the results of `most_similar` for `boy girl he she` return `He` followed by `She`, with the previous logic and `case_insensitive=True`, `He` gets ignored as one of the analogy words itself (as the matching is performed on the value and not the index), and the analogy gets marked as correct.

`compute-accuracy.c` works with indices though. This was how it had been implemented previously in gensim too, the change to value-based matching was in one my commits itself :/

You could actually argue `He` should be ignored, and the analogy be marked as correct, but since we're following the original implementation exactly, this is definitely not desirable.

Let me know if you find any other issues with the PR

tmylk:
@jayantj Could you please pull master and resolve the merge conflicts?

tmylk:
Thanks! Merged in https://github.com/RaRe-Technologies/gensim/commit/3c4731908d56f46c9c9d39fe00e6e986c1bd9ea0

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Changes
 * Implemented LsiModel.docs_processed attribute
 * Added LdaMallet support. Added LdaVowpalWabbit, LdaMallet example to notebook. Added test suite for coherencemodel and aggregation.
   Added `topics` parameter to coherencemodel. Can now provide tokenized topics to calculate coherence value (@dsquareindia, #750)
+* Changed `use_lowercase` option in word2vec accuracy to `case_insensitive` to account for case variations in training vocabulary (@jayantj, #714)
```
piskvorky: Please link directly to the github pullrequest/issue page, so people can check the changes and discussions easily.

The markdown syntax for links is `[link text](link url)`.

(@tmylk please start doing this consistently in CHANGELOG starting this release)

PR #806, opened 2016-07-25
Update dtmmodel.py

Eickho:
Adds a check if any documents in the supplied corpus are empty, which breaks the DIM model without providing a usable error description to the user. It should be noted that this _only_ seems to cause errors with the DIM model and not the DTM model. However, I don't think there is any upside to having empty documents in the DTM anyway?

### DIFF
--- a/gensim/models/wrappers/dtmmodel.py
+++ b/gensim/models/wrappers/dtmmodel.py
@@ -93,7 +93,7 @@ class DtmModel(utils.SaveLoad):
             lencorpus = sum(1 for _ in corpus)
         if lencorpus == 0:
             raise ValueError("cannot compute DTM over an empty corpus")
-        if any([i == 0 for i in [len(text) for text in corpus.get_texts()]]):
+        if any([i == 0 for i in [len(text) for text in corpus.get_texts()]]) and model == "fixed:
             raise ValueError("""There is a text without words in the input corpus.
                     This breaks method='fixed' (The DIM model).""")
         if lencorpus != sum(time_slices):
--- a/gensim/models/wrappers/dtmmodel.py
+++ b/gensim/models/wrappers/dtmmodel.py
@@ -93,7 +93,7 @@ class DtmModel(utils.SaveLoad):
             lencorpus = sum(1 for _ in corpus)
         if lencorpus == 0:
             raise ValueError("cannot compute DTM over an empty corpus")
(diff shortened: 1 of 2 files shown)


bhargavvader:
Yeah, this is a documented problem with the DIM code. This should be useful.

gensim/models/wrappers/dtmmodel.py
```diff
@@ -93,6 +93,9 @@ def __init__(
             lencorpus = sum(1 for _ in corpus)
         if lencorpus == 0:
             raise ValueError("cannot compute DTM over an empty corpus")
+        if any([i == 0 for i in [len(text) for text in corpus.get_texts()]]):
```
tmylk: should it be `and model=='fixed'` as well?

tmylk:
@Eickho Thanks for the PR. Please add a changelog and a check to only affect `model=='fixed'`

bhargavvader:
@Eickho  will you be taking this up?

gensim/models/wrappers/dtmmodel.py
```diff
@@ -93,6 +93,9 @@ def __init__(
             lencorpus = sum(1 for _ in corpus)
         if lencorpus == 0:
             raise ValueError("cannot compute DTM over an empty corpus")
+        if any([i == 0 for i in [len(text) for text in corpus.get_texts()]]) and model == "fixed:
```
bhargavvader: Wouldn't it be better to do `if model == "fixed" and any([i == 0 for i in [len(text) for text in corpus.get_texts()]]):`
so that it won't bother checking the condition if it isn't in `fixed` mode?

Eickho: Yes that would be better, updating.

gensim/models/wrappers/dtmmodel.py
```diff
@@ -93,6 +93,9 @@ def __init__(
             lencorpus = sum(1 for _ in corpus)
         if lencorpus == 0:
             raise ValueError("cannot compute DTM over an empty corpus")
+        if model == "fixed" and any([i == 0 for i in [len(text) for text in corpus.get_texts()]]):
```
piskvorky: Replace with `any(not text for text in corpus.get_texts())` (more Pythonic). This looks unnecessarily complicated.

Issue #549, opened 2015-11-26
Potential refactor: a 'NamedVectors' class for reuse by Word2Vec, Doc2Vec, etc


gojomo:
Some wishlist features for Word2Vec/Doc2Vec vector sets, like approximate-neighbor-indexing (#51, #527) or post-training transformations ('retrofit' #547 or [translation](https://github.com/piskvorky/gensim/wiki/Word2Vec-&-Doc2Vec-Wishlist#implement-translation-matrix-of-exploiting-similarities-among-languages-for-machine-translation)), really just need a set-of-vectors-of-which-some-are-string-named. That is, the bundle of state now mixed-into Word2Vec/DocvecsArray as `syn0` array, `index2word` list, and `vocab` dict. 

It may ease those features and improve the code-organization logic to refactor that state out into a new class, tentatively called 'NamedVectors', to be reused via composition into these other classes, but also available for separate use when you don't need the training-model-wrapper. 

Many operations – loading from a plain word2vec.c model, similarity-lookups, (future) approximate-indexing, (future) learned-projections, etc – would live inside or operate on these NamedVectors objects, which could be loaded/saved separately from trainable models. For example, if you only want the vectors and the similarity-operations, why wrap that state in a broken full training model (or keep around the expansive extra state that's not needed)?

piskvorky:
The challenge here is designing a flexible enough API, i.e. identifying relevant use cases.

From the linked tickets, it looks like a good-enough API may be just plain Python mapping (e.g., `dict`). Word2vec already conforms -- `model['some_word']` returns a vector.  A new class (`NamedVectors`) seems superfluous.

We could also offer a method to extract this dict from word2vec explicitly (`model.to_dict()`), for advanced users.

gojomo:
Yes, the API would definitely support `__getitem__()` lookup; the benefit would be disentangling that state from other state/behavior that isn't necessarily attached. So it'd be easier to process/talk-about the *output* of a Word2Vec/Doc2Vec process, as something separate from its parameters/algorithms/internal-state. 

For example:

* Say you want just a set of vectors for similarity-searches? You wouldn't have to 'strip down' a larger Word2Vec model, leaving it in some halfway state where some operations give cryptic errors. If you only want vectors-keyed-by-strings, you'd just pull out (or in fact only ever instantiate) one of these new objects. 

[...]

PR #809, opened 2016-07-28
[WIP] Add ability to use Tensorflow to train a word2vec model 

droudy:
The flexible architecture of tensorflow allows you to deploy computation to one or more CPUs or GPUs  with a single API. The benefit of using tensorflow for training in w2v is that it can distribute computations across GPUs. This PR adds the ability for a user to easily create a w2v model that is trained using tensorflow but still allows gensim w2v methods such as `most_similar()` and `doesnt_match()` to be called on the model that is trained with tensorflow. 

It works around an existing tensorflow module, `tensorflow.models.embedding.word2vec_optimized`
<br><pre><code>
# Training data can be a gensim style corpus or a text file

model = TfWord2Vec(example_corpus)
model.most_similar("army")

### DIFF
--- a/gensim/models/tfword2vec.py
+++ b/gensim/models/tfword2vec.py
@@ -86,8 +86,8 @@ class Options(object):
                    will be randomly down-sampled. Set to 0 to disable.
         """
 
-        if train_data is None or save_path is None:
-            raise ValueError("train_data and save_path must be specified.")
+        if train_data is None:
+            raise ValueError("train_data must be specified.")
 
         # Model options.
 
@@ -141,7 +141,8 @@ def modified_tfw2v_init(self, options, session):
     self._id2word = []
     self.build_graph()
     self.build_eval_graph()
-    self.save_vocab()
+    if options.save_path is not None:
+        self.save_vocab()
     if options.eval_data is not None:
         self._read_analogies()
 
--- a/gensim/models/tfword2vec.py
+++ b/gensim/models/tfword2vec.py
@@ -14,37 +14,37 @@ class GensimWord2VecNoTraining(GensimWord2Vec):
     """
 
     def make_cum_table(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def create_binary_tree(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def build_vocab(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def scan_vocab(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def scale_vocab(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def finalize_vocab(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def sort_vocab(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def _do_train_job(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def train(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def score(self, *args, **kwargs):
-        pass
+        raise Exception("Cannot call a gensim training method on a tf trained model")
 
     def save_word2vec_format(self, *args, **kwargs):
-        pass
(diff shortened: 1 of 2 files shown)


tmylk:
Ping @gojomo

piskvorky:
What is the status on tensor flow -- is there a notebook comparing the performance / pros / cons?

Also, what is @gojomo 's role here?

gensim/models/tfword2vec.py
```diff
@@ -0,0 +1,206 @@
+import tensorflow as tf
+from tensorflow.models.embedding.word2vec_optimized import Word2Vec
+from gensim.models.word2vec import Word2Vec as GensimWord2Vec, Vocab
+from gensim import utils
+from six import string_types
+import logging
+logging.basicConfig(format='%(asctime)s : %(levelname)s : %(message)s', level=logging.INFO)
```
piskvorky: This doesn't belong in module scope -- libraries do not set up logging. That's up to applications that use them.

gensim/models/tfword2vec.py
```diff
@@ -0,0 +1,206 @@
...
+    self._word2id = {}
+    self._id2word = []
+    self.build_graph()
+    self.build_eval_graph()
+    if options.save_path is not None:
+        self.save_vocab()
+    if options.eval_data is not None:
+        self._read_analogies()
+
+Word2Vec.__init__ = modified_tfw2v_init
```
piskvorky: What is this meant to do?

Such magic -- monkey-patching outside modules -- has to be very carefully motivated and documented.

gensim/models/tfword2vec.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        Converts gensim corpus to a file that can be used by tf word2vec
+
+        """
+        #assumes that the string represents a file extension
+        if not isinstance(corpus, str):
+            with utils.smart_open('/tmp/converted_corpus', 'w+') as fout:
+                for line in corpus:
+                    for word in line:
+                        fout.write(utils.to_utf8(str(word) + " "))
```
piskvorky: Does this do what you intended? Looks like this will fail on unicode strings.

gensim/models/tfword2vec.py
```diff
@@ -0,0 +1,206 @@
...
+            self.vocab[word] = Vocab(index=self.model._word2id[word])
+
+    def convert_input(self, corpus):
+        """
+        Converts gensim corpus to a file that can be used by tf word2vec
+
+        """
+        #assumes that the string represents a file extension
+        if not isinstance(corpus, str):
+            with utils.smart_open('/tmp/converted_corpus', 'w+') as fout:
```
piskvorky: This is not portable -- use `os.path.join` and `tempfile`.

More importantly --- what is this supposed to do? Hardwiring some paths into code looks like a bad idea.

gensim/models/tfword2vec.py
```diff
@@ -0,0 +1,206 @@
...
+        for word in self.options.vocab_words:
+            self.vocab[word] = Vocab(index=self.model._word2id[word])
+
+    def convert_input(self, corpus):
+        """
+        Converts gensim corpus to a file that can be used by tf word2vec
+
+        """
+        #assumes that the string represents a file extension
+        if not isinstance(corpus, str):
```
piskvorky: Filenames can be unicode too (check against `basestring`).

gensim/models/tfword2vec.py
```diff
@@ -0,0 +1,206 @@
...
+        self.save_path = save_path
+
+        # Eval options.
+
+        # The text file for eval.
+        self.eval_data = eval_data
+
+
+def modified_tfw2v_init(self, options, session):
+    self._options = options
```
piskvorky: Is there a conflict with another variable? Or why are these attributes internal (~underscored, `_`)?

gojomo:
I think the idea of being able to use TF training or import vectors from a TF session is good, but this structuring seems fragile/confusing - especially mixed-overriding & renaming of parameters. 

I suspect the better approach would involve some combination of: (1) a new common superclass for what is shared in implementation or interface, and having the TF implementation being a sibling class, rather than patchwork-subclass, of the traditional implementations; (2) moving the vectors-and-vocab entity out of the algorithmic entity, as proposed in #549.  Of course such refactoring is a kind of big and disruptive project...

tmylk:
Current status: Blocked by #549

tmylk:
@parthoiiitm This PR can now be updated to use KeyedVecs from  https://github.com/RaRe-Technologies/gensim/pull/847

markroxor:
Extending this PR here https://github.com/RaRe-Technologies/gensim/pull/1033

gensim/models/tfword2vec.py
markroxor: These are the [standard parameters used by TF](https://github.com/tensorflow/tensorflow/blob/r0.7/tensorflow/models/embedding/word2vec_optimized.py#L140) which are being overrided.

PR #816, opened 2016-08-07
Small error on Doc2vec notebook tutorial 

charlessutton:
I have just corrected a small error,
In the current version of the notebook, in the last box it prints  document from train corpus instead of test corpus,
Now the document that should appears when printing is the one from the test corpus,
Best regards,


tmylk:
Thanks a lot for the PR!

PR #817, opened 2016-08-09
new options for load/intersect_word2vec_format

gojomo:
load_word2vec_format:
* `limit` to only read 1st N vectors from file 
* `datatype` to force smaller datatype (risky/slow but can save memory)

intersect_word2vec_format:
* `lockf` argument to set whether imported vectors are locked-against-changes (0.0) or not (1.0)

also: tiny fix in Text8Corpus, ensuring final chunk fragment gets same unicode-treatment as all other created-lines.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1054,7 +1054,8 @@ class Word2Vec(utils.SaveLoad):
                     fout.write(utils.to_utf8("%s %s\n" % (word, ' '.join("%f" % val for val in row))))
 
     @classmethod
-    def load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8', unicode_errors='strict'):
+    def load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8', unicode_errors='strict',
+                             limit=None, datatype=REAL):
         """
         Load the input-hidden weight matrix from the original C word2vec-tool format.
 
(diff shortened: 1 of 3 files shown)


gojomo:
@piskvorky @tmylk Some small improvements for loading word2vec.c-format vectors, and a tiny fix for issue I hit when comparing w/ fastText. Please ack for merge before drifts-into-conflict?

gensim/models/word2vec.py
```diff
@@ -1120,7 +1135,8 @@ def add_word(word, weights):
                     weights = fromstring(fin.read(binary_len), dtype=REAL)
                     add_word(word, weights)
             else:
-                for line_no, line in enumerate(fin):
+                for line_no in xrange(vocab_size):
```
piskvorky: Why this change?

gensim/models/word2vec.py
```diff
@@ -1170,7 +1190,7 @@ def intersect_word2vec_format(self, fname, binary=False, encoding='utf8', unicod
                     if word in self.vocab:
                         overlap_count += 1
                         self.syn0[self.vocab[word].index] = weights
-                        self.syn0_lockf[self.vocab[word].index] = 0.0  # lock it
+                        self.syn0_lockf[self.vocab[word].index] = lockf  # lock-factor: 0.0 stopss further changes
```
piskvorky: Little typo: `stopss`.

piskvorky:
The change is desirable and code looks fine to me. Final decision and merge up to @tmylk .

gensim/models/word2vec.py
gojomo: Without it, any user-specified `limit` cap on `vocab_size` won't be respected for `binary=False`-format reads. 

(Without a `limit`, the only way this would prevent a line from being read is if the top-of-file declared count is smaller than the following number of lines... and in such case, reading exactly the declared number of vectors is still arguably a reasonable thing to do, and would also match the behavior of the `binary=True` branch.)

piskvorky: Yes; I meant why the line split.

`islice` seems more idiomatic and also safer (in case `vocab_size > len(fin)`).

gojomo: In the moment, the `xrange` approach was to mirror the way the binary-branch worked, where changing `vocab_size` to `limit` already had the desired effect. Considering it against `islice`, it's probably better for a mismatch to generate an error, than just silently accept fewer-than-the-declared count of vectors. 

But, thinking of this made me realize the code wasn't yet handling a `limit` > the available count (so that's now been fixed), and the `binary` path could reach a busy-hang if reaching EOF while trying to read its token char-by-char (so both it and the text-path now recognize when a read returns nothing, indicating file-end-before-declared-count-reached, and raise an EOFError).

tmylk:
Resolved conflicts, squashed trivial commits and merged in https://github.com/RaRe-Technologies/gensim/commit/15ee57fed9d02661d11996dd14cde24fc981c195

PR #821, opened 2016-08-10
Fix failing testRandomState()

droudy:
`numpy.random.RandomState(0) ==  numpy.random.RandomState(0)` returns `False` so the test was failing.


piskvorky:
What was that test for, originally? I don't remember that code, or its motivation.

droudy:
@piskvorky it was originally from pr #748 motivated by issue #113

PR #823, opened 2016-08-11
Add a new Doc2Vec example

544895340:
I add a new Doc2Vec example in tutorial. It is about sentiment analysis on airline tweets dataset


tmylk:
Thanks!

PR #833, opened 2016-08-19
KeyedVectors refactor for word2vec

droudy:
Addresses #549. Refactored syn0, syn0norm, vocab, and index2word into their own class, as well as methods not involved in training that require vectors such as `most_similar()`. Maintains backwards compatibility so that calls such as `trained_model.most_similar("the")` and `trained_model.syn0` don't break. 

If you are done with training, you can retrieve the keyedvectors of your trained model and save them to disk like so:
<pre> <code>
# kv --> KeyedVectors

retrieved_kv = trained_model.kv 
retrieved_kv.save("/dir/saved_vecs") 
# Later on

loaded_kv = KeyedVectors.load("/dir/saved_vecs")
loaded_kv.most_similar("the")
</pre></code>
@gojomo please review 
The changes to `utils.SaveLoad` in particular are a bit strange due to the need to maintain backwards compatibility, I can elaborate on them if the comments aren't comprehensive enough.


gojomo:
This separate-file, take-as-much-generic-functionality-as-possible approach is what I think is the right approach. 

As before, I'd expect a role-based abbreviation (`wv`) to be clearer than a type-based abbreviation (`kv`). There's the potential this class could also replace the DocvecsArray in Doc2Vec, in which case a model will have two KeyedVector properties. There's even the potential this could replace the `syn1` layer (perhaps moreso in negative-sampling) and offer 'OUT' vector access-by-word-key. (See the 'Dual Embedding Space Model' papers from Mitra et al at MSFT for why that might be interesting.)

Whether KeyedVectors should get the full Vocab functionality, including frequencies and (in `hs` mode) huffman-coding-info), still seems a messy question to me. Not all KV users will have or need such info; on the other hand it makes sense to keep it close to the key-to-index lookup, for those activities that do need it. I wonder if KV should have an expandable property-set, perhaps as a dict of single-type arrays that are int-indexed in parallel with the vectors. (And if so, `syn0_lockf`-like per-slot values could also move into KV.) Unsure of right approach – just see this as a current area of unclean-separation-of-roles.

The use of `__getattr__` patching to maintain property-access backward-compatibility makes me uncomfortable; it makes functionality more mysterious/magical, and adds confusing failure modes. I wouldn't say for sure that such direct-property access needs to be held compatible or should be considered part of the public API. (Perhaps code making such direct accesses *should* update itself when upgrading to a new gensim. Though alternatively, if we *do* consider such properties to be part of the API, and we strictly adhere to 'semantic versioning', such an API change would necessitate a major-version increment.)

The main backward-compatibility I'd be concerned about would be the ability to load older saved models. (But. I wouldn't expect newer models, or older models once loaded, to be savable in a format that older code could load.) It'd be good to have a unit test of that, perhaps by bundling tiny saved-models in the test data. (Even though tiny models don't usually save their arrays as separate files, we'd want to test that mode, too.) 

I don't see the corresponding updates to the cython files, but maybe that's because the `__getattr__` patching is keeping it working?

piskvorky:
Thanks! Like @gojomo says, we have to be careful about maintaining backward compatibility, so we'll need tests for all appropriate combinations of {load/save, old/new model, single file/multiple mmap files}.

Re. testing on smaller models -- there's a threshold that says how large internal arrays need to be before they're saved as separate, mmap-able files. This threshold is configurable: `model.save(sep_limit=0)` (default `sep_limit` is 10MB).

gensim/models/word2vec.py
```diff
@@ -654,27 +654,27 @@ def finalize_vocab(self):
...
+            v.index = len(self.kv.vocab)
+            self.kv.index2word.append(word)
+            self.kv.vocab[word] = v
         # set initial input/projection and hidden weights
         self.reset_weights()
 
     def sort_vocab(self):
         """Sort the vocabulary so the most frequent words have the lowest indexes."""
-        if hasattr(self, 'syn0'):
+        if self.kv.syn0:
```
jayantj: I'm not completely sure this'll work since right now, `self.syn0` is initialized in `reset_weights`, which is called after `sort_vocab`. With `kv` being assigned in `__init__`, wouldn't this always raise an exception?
Possibly have a `reset_weights` equivalent in `KeyedVectors`?

gensim/models/word2vec.py
```diff
@@ -1191,383 +1191,78 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
...
+            return self.kv.syn0norm
+        if item == "vocab":
+            return self.kv.vocab
+        if item == "index2word":
+            return self.kv.index2word
+        raise AttributeError("'{0}' object has no attribute '{1}'".format(type(self).__name__, item))
 
     def __contains__(self, word):
-        return word in self.vocab
+        return self.kv.__contains__(word)
```
jayantj: Explicitly calling `__contains__` seems a little awkward. `return word in self.kv`?

gensim/models/word2vec.py
```diff
@@ -1191,383 +1191,78 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
...
-            if replace:
-                for i in xrange(self.syn0.shape[0]):
-                    self.syn0[i, :] /= sqrt((self.syn0[i, :] ** 2).sum(-1))
-                self.syn0norm = self.syn0
-                if hasattr(self, 'syn1'):
-                    del self.syn1
-            else:
-                self.syn0norm = (self.syn0 / sqrt((self.syn0 ** 2).sum(-1))[..., newaxis]).astype(REAL)
+        if hasattr(self, 'syn1'):
+            del self.syn1
```
jayantj: `replace` still needs to be checked. Otherwise, a call to `init_sims` even with `replace=False` will cause the model to become incapable of continuing training.

gensim/models/word2vec.py
droudy: @jayantj `self.kv.syn0` gets initialized in `__init__` but isn't populated, so if the vectors/weights have not been initialized it will remain an empty list and an empty list as a conditional is `False

jayantj: Ah, yes, hadn't thought of that, sorry

gensim/models/word2vec.py
droudy: @jayantj That's right, I'll make the according changes, thank you

gensim/models/word2vec.py
```diff
@@ -79,6 +79,7 @@
 import itertools
 
 from gensim.utils import keep_vocab_item
+from keyedvectors import KeyedVectors
```
jayantj: Relative imports don't play well with python3, and gensim seems to use absolute imports everywhere

gensim/models/keyedvectors.py
```diff
@@ -0,0 +1,479 @@
...
+                if a not in ok_vocab or b not in ok_vocab or c not in ok_vocab or expected not in ok_vocab:
+                    logger.debug("skipping line #%i with OOV words: %s" % (line_no, line.strip()))
+                    continue
+
+                original_vocab = self.vocab
+                self.vocab = ok_vocab
+                ignore = set([a, b, c])  # input words to be ignored
+                predicted = None
+                # find the most likely prediction, ignoring OOV words and input words
+                sims = most_similar(self, positive=[b, c], negative=[a], topn=False, restrict_vocab=restrict_vocab)
```
jayantj: This causes some rather peculiar behaviour when `word2vec_object.accuracy` is called. The `most_similar` from `Word2Vec` gets passed as the default argument, and it gets called from here with `self` set to the KeyedVectors instance. Hence, the `self.kv` statements in Word2Vec's `most_similar` raise an exception

@piskvorky Slightly curious about this choice of having `most_similar` as a parameter, is there a niche use case for it? If so, we should probably add documentation for it

piskvorky: @jayantj No idea, this is @gojomo 's code. 

I haven't reviewed yet, but obviously we want to keep the API as un-convoluted and simple as possible (but not any simpler).

jayantj: @gojomo any idea why `most_similar` is a parameter? Is there a certain use case for it?

gensim/utils.py
```diff
@@ -376,11 +376,26 @@ def _save_specials(self, fname, separately, sep_limit, ignore, pickle_protocol,
...
         for attrib in separately + list(ignore):
-            if hasattr(self, attrib):
-                asides[attrib] = getattr(self, attrib)
-                delattr(self, attrib)
+            # As a result of maintaining backwards compatibility through __getattr__ after refactoring
+            # "syn0", "syn0norm", "vocab", and "index2word" out of Word2Vec to KeyedVectors,
+            # hasattr(self, "syn0") will return True but delattr(self, "syn0") will fail because it's
+            # not a true attribute of Word2Vec, Word2Vec.__getattr__ just reroutes it to KeyedVectors' attributes
+            if attrib in ["syn0", "syn0norm", "vocab", "index2word"]:
+                if type(self) == "Word2Vec":  # if saving a Word2Vec model, syn0, etc. are stored in self.kv
```
jayantj: This currently doesn't work as intended because `type(self)` returns a class, not a string. But even if changed to `type(self).__name__`, this would call `delete_attribute` with the Word2Vec object and syn0/syn0norm, and so when properties are being restored, syn0/syn0norm would be restored to the word2vec object instead of the KeyedVecs object. 

The correct way to handle it seems via the recursive_saveloads that is already implemented, but we'll still need a special condition check to ensure delete_attribute isn't called on the word2vec object. 

All this need for special condition checks is resulting from the getattr patching to handle syn0, vocab and syn0norm backwards compatibility. 

Even if we decide not to maintain syn0 and syn0norm backwards compatibility, vocab is definitely part of the API, and there doesn't seem to be any way to avoid special patching.

jayantj: Or we could use property descriptors and add setters and deleters for syn0norm, vocab etc, and perform those operations on `self.kv`. Wouldn't require any special condition checks anywhere, and we could add warnings that direct access to syn0/syn0norm would be removed in a future release.

gensim/models/keyedvectors.py
gojomo: It was to allow similarity-measures other than plain cosine-similarity, such as the multiplicative-cosine-similarity ("3CosMul") proposed by Goldberg/Levy that does a little better on analogy questions.

gojomo:
I'd said before that use of `__getattr__` patching to maintain property-access backward-compatibility made me uncomfortable; I would like to upgrade that to "I think it's a really bad idea". Similarly, the special-casing of things by class/attribute name inside `utils.py` for save/load seems very fragile/delocalized. 

Overall this is a big enough change that maintaining full API compatibility, down to direct-access of properties, adds complexity and risk compared to consciously breaking things. As a result, I'd suggest this either (1) hold to happen on a major-version rev where it's expected we might say, "your code must be updated"; or (2) happen in a new parallel class (like `NewWord2Vec` or `future.Word2Vec`. In case (2), pre- and post-refactor code can be tested and verified as behaving identically within the same build, and advanced users can opt to use to the new API knowing some code tweaks will be necessary. Later, when there's a major-version rev, the old class can swapped out for the new class, forcing users' code to adjust only with that major-version rev.

In any case, the main form of backward-compatibility to be maintained would be the ability to load/convert older saved models into the new format - ideally as a specific named function with all special-casing local, rather than as a series of conditionals/attribute-interceptions spread across many places.

jayantj:
I've handled the special patching inside `utils.py` and `__getattr__` inside an overriden `_load_specials` in word2vec, and with the use of property descriptors for `syn0`, `syn0norm`, `vocab` and `index2word`.

This way, backwards compatibility can be maintained for now without having a series of confusing conditionals distributed in the codebase. Also, it allows us to add warnings in case of direct syn0/syn0norm lookup (along the lines, "syn0 access will no longer be supported in a future gensim release, please use model.kv.syn0 instead")

All these changes are in #852 
@droudy if you can give me access to this branch, I can push all my code to this PR and keep all code and discussion here

droudy:
@jayantj I sent an invite to collaborate so you can push to it

jayantj:
Thanks @droudy, pushed.
Also, the build seems to be failing due to #853

tmylk:
@anmol01gulati has kindly expressed interest in finishing this

anmol01gulati:
@jayantj I worked on this. So, your tests fail as there is a break issue between Python 3.4 and 3.5 as well. See [here](https://groups.google.com/d/msg/gensim/wA2l4EjlqtA/dUERYYHuJAAJ)
So, you would want to add another model file explicitly for Python 3.4 and 3.5 and it would work.

gensim/test/test_word2vec.py
```diff
@@ -186,13 +187,19 @@ def testSyn0NormNotSaved(self):
...
         """Test loading pre-KeyedVectors word2vec model"""
 
-        # Model stored in one file
-        model = word2vec.Word2Vec.load(datapath('word2vec_pre_kv'))
-        self.assertTrue(model.wv.syn0.shape == (len(model.wv.vocab), model.vector_size))
-        self.assertTrue(model.syn1neg.shape == (len(model.wv.vocab), model.vector_size))
+        if sys.version_info < (3,):
+            model_file_suffix = '_py2'
+            # Model stored in one file
+            model_file = 'word2vec_pre_kv%s' % model_file_suffix
```
anmol01gulati: This block of checking the model should come after the if/else part.

jayantj:
@anmol01gulati Thanks a lot for looking into it. That sounds right, should work. The test also seems to be failing for Python 2.7.

Also this PR is from @droudy's fork, so I can't give you push access. @droudy could you please grant @anmol01gulati access?

Regarding break between Python 3.4 and 3.5, a fairly clean workaround for the pickle bug is doable, even though it isn't a bug with gensim.  @tmylk do you think it'd be a good idea to have a fix for it, or is this too obscure?

tmylk:
@anmol01gulati The 2.7 tests fail because of an unrelated known(though not investigated) wikitest glitch.

Also feel free to create a new PR with this code as @droudy is busy till mid-november.

It would be good to add the Python 3.4 fix inside the `if` statement to make sure it only affects the Python 3.4 version.

tmylk:
Merged in #980

PR #837, opened 2016-08-25
[WIP] Phrases optimizations

gojomo:
* avoids an unnecessary duplicate giant dictionary in common case of a single `add_vocab()` call
* adds new `Phraser` helper class & supporting options on existing methods

`Phraser` takes a `Phrases` and does a single (time-consuming) pass to discover all the Phrases that it would want to create, saving those into a much much smaller (and somewhat faster) helper object. This can be saved & used separately.

Needs more testing.

### DIFF
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -350,7 +350,7 @@ class Phraser(interfaces.TransformationABC):
         delimiter = self.delimiter
         for word_a, word_b in zip(s, s[1:]):
             bigram_tuple = (word_a, word_b)
-            if bigram_tuple in phrasegrams and not last_bigram:
+            if phrasegrams.get(bigram_tuple, (-1, -1))[1] > self.threshold and not last_bigram:
                 bigram_word = delimiter.join((word_a, word_b))
                 new_s.append(bigram_word)
                 last_bigram = True
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -79,7 +79,7 @@ class Phrases(interfaces.TransformationABC):
 
     """
     def __init__(self, sentences=None, min_count=5, threshold=10.0,
-            max_vocab_size=40000000, delimiter=b'_'):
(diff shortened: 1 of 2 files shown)


piskvorky:
Note to self: use the faster detection loop in Python.

@lev good student project: implement the inner loops as C (via Cython) **optional** extension (optional ala word2vec extension).

piskvorky:
@gojomo @lev ready to merge? What's missing?

tmylk:
Created #918 asking for volunteers to create tests

tmylk:
Merged in #954

tmylk:
@gojomo Should `Phrases` be deprecated? The existing warning about its slowness [doesn't](https://groups.google.com/d/msg/gensim/VmCai9ciUz0/oRvbWKjsEgAJ) seem to deter users.

gojomo:
It's necessary to use a `Phrases` to construct a `Phraser`. `Phraser` takes a bunch of extra time to create, but then is *slightly* faster but *much more compact*. (The main benefit is memory, not speed, and certainly not speed-to-first-use, which is worse.) 

Also, a `Phraser` essentially locks in the effective `min_count` and `threshold` values at the time of its creation – so if you want to try other values, you need to go back to a `Phrases` with the full frequency data. 

So `Phrases` can't quite be deprecated or hidden... perhaps some renaming/refactoring and updated docs/examples would guide people to the best options for them. There's also the potential for using  more memory-efficient approximate counts, to use less memory during initial `Phrases` analysis. I believe one or two folks have started down that road before, but their PRs never reached mergeable state. If there's someone with the interest/skills to tackle that, they might want to, at the same time, try to rationalize the interfaces/names for better clarity.

piskvorky:
@tmylk a fast C/Cython implementation, ideally multicore, would be a great project.

The entire "training" is essentially incrementing a counter, no reason it shouldn't be as fast as your input iterator provides.

Plus, phrases (collocations) are not going anywhere, so it's a stable module to invest more time into.

PR #843, opened 2016-08-25
Feature/842/word2vec words dropped better logging

loisaidasam:


gensim/models/word2vec.py
```diff
@@ -578,24 +578,26 @@ def scale_vocab(self, min_count=None, sample=None, dry_run=False, keep_raw_vocab
...
-        logger.info("min_count=%d retains %i unique words (drops %i)",
-                    min_count, len(retain_words), drop_unique)
-        logger.info("min_count leaves %i word corpus (%i%% of original %i)",
-                    retain_total, retain_total * 100 / max(original_total, 1), original_total)
+        original_unique_total = len(retain_words) + drop_unique
+        retain_unique_pct = len(retain_words) * 100 / max(original_unique_total, 1)
+        logger.info("min_count=%d retains %i unique words (%i%% of original %i, drops %i)",
+                    min_count, len(retain_words), retain_unique_pct, original_unique_total, drop_unique)
+        original_total = retain_total + drop_total
+        retain_pct = retain_total * 100 / max(original_total, 1)
```
tmylk: Should it be 100.0 to force float?

loisaidasam: Sure, good idea.

I was simply using the same style as already existed for calculating percentage.

gojomo: The earlier `from __future__ import division` already ensures a float result (rather than int) when appropriate.

tmylk:
Looks good. Thanks for the PR
```python
sentences = [['first', 'sentence'], ['second', 'sentence']]
model = gensim.models.Word2Vec(sentences, min_count=2)
```

```
2016-09-27 10:45:13,881 : INFO : min_count=2 retains 1 unique words (33% of original 3, drops 2)
2016-09-27 10:45:13,881 : INFO : min_count=2 leaves 2 word corpus (50% of original 4, drops 2)
```

PR #845, opened 2016-08-30
Added save/load functionality to AnnoyIndexer

fortiema:
Having Annoy integrated directly into gensim is really great, but one feature that I was personally missing is the ability to save/load indexes. I am working with indexes in the 10s of GB and having to recreate them every time I run my code is a waste of time.

So I added a simple save/load interface that is similar to Annoy.

For example this code:
```python
fname = 'index'
if os.path.exists(fname):
    self.index_en = AnnoyIndexer()
    self.index_en.load(fname)
    self.index_en.model = model
else:
    self.index_en = AnnoyIndexer(model, 1)
    self.index_en.save(fname)
```

Will create 2 files, _index_ and _index.d_. Both files must be present when using the `load` function, otherwise nothing happens.

For this to work, I also added a if case in the constructor to allow for object creation without passing _model_ and _num_trees_.

try/except on import and using pickle protocol v2 to stay 2-3 compatible.

All comments and suggestions are welcome.

### DIFF
--- a/gensim/similarities/index.py
+++ b/gensim/similarities/index.py
@@ -45,8 +45,8 @@ class AnnoyIndexer(object):
     def load(self, fname):
         fname_dict = fname+'.d'
         if not (os.path.exists(fname) and os.path.exists(fname_dict)):
-            raise IOError("Can't find index files '{}' and '{}' - Unable to restore AnnoyIndexer state."
-                          .format(fname, fname_dict))
+            raise IOError(
+                "Can't find index files '%s' and '%s' - Unable to restore AnnoyIndexer state." % (fname, fname_dict))
         else:
             with smart_open(fname_dict) as f:
                 d = _pickle.loads(f.read())
--- a/gensim/similarities/index.py
+++ b/gensim/similarities/index.py
@@ -4,10 +4,12 @@
 # Copyright (C) 2013 Radim Rehurek <me@radimrehurek.com>
 # Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
 import os
+
+from smart_open import smart_open
 try:
-    import cPickle as pickle
+    import cPickle as _pickle
 except ImportError:
-    import pickle
+    import pickle as _pickle
 
 from gensim.models.doc2vec import Doc2Vec
 from gensim.models.word2vec import Word2Vec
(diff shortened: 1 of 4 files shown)


fortiema:
Added basic test cases to test_similarities for both Word2Vec and Doc2Vec

tmylk:
Thanks for the PR!

Could you please add a line to CHANGELOG and update the annoy notebook tutorial with the new functionality?

fortiema:
Sure! Will push when done.

fortiema:
Any other suggestions to make this interface more robust?

gensim/similarities/index.py
```diff
@@ -15,16 +19,32 @@
...
+                self.build_from_doc2vec()
+            elif isinstance(self.model, Word2Vec):
+                self.build_from_word2vec()
+            else:
+                raise ValueError("Only a Word2Vec or Doc2Vec instance can be used")
+
+    def save(self, fname):
+        self.index.save(fname)
+        d = {'f': self.model.vector_size, 'num_trees': self.num_trees, 'labels': self.labels}
+        pickle.dump(d, open(fname+'.d', 'wb'), 2)
```
tmylk: Please use smart_open as in https://github.com/RaRe-Technologies/gensim/blob/6a289fefd72f038c8cc14826f63624950f5de1f8/gensim/utils.py#L896

gensim/similarities/index.py
```diff
@@ -15,16 +19,32 @@
...
+
+    def load(self, fname):
+        if os.path.exists(fname) and os.path.exists(fname+'.d'):
+            d = pickle.load(open(fname+'.d', 'rb'))
```
tmylk: Please use smart_open as in https://github.com/RaRe-Technologies/gensim/blob/6a289fefd72f038c8cc14826f63624950f5de1f8/gensim/utils.py#L907

gensim/test/test_similarities.py
```diff
@@ -464,6 +464,32 @@ def testApproxNeighborsMatchExact(self):
...
+    def testSave(self):
+        self.index.save('index')
+        self.assertTrue(os.path.exists('index'))
+        self.assertTrue(os.path.exists('index.d'))
+
+    def testLoadNotExist(self):
+        from gensim.similarities.index import AnnoyIndexer
+        self.test_index = AnnoyIndexer()
+        self.test_index.load('test-index')
+
```
tmylk: It has to raise IOError

tmylk:
Thanks for the quick fix. Once 2.6 tests runs we could merge.

Also it would be interesting to see a test where 2 parallel processes load the same model from disk and mmap the same index file?

fortiema:
Great suggestion, let me add this as well!

PR #847, opened 2016-09-01
[MRG] Wrapper for FastText

jayantj:
I've branched this off @droudy's `KeyedVectors` PR (#833). The changes I've made are all in `gensim/models/wrappers/fasttext.py`

Just wondering if this is the correct approach to go with, if we can decide that, I'll properly document the code, write tests, refactor it, right now it's mostly a POC. The main things I'm concerned about are - 
1. Portability (using `structs` to read off the binary file generated by FastText, I'm not completely sure about how portable this is across different compiler versions and architectures)
2. FastText changes to binary file - If FastText makes any changes to the binary file, we'll have to make those changes too, and since the binary file seems to be mostly for internal use, it could very well see changes?
3. Memory usage - FastText takes up a huge amount of memory - with a tiny vocabulary of around 4k, my memory usage goes to ~750 MB. Regular word2vec takes up ~70 MB for the same. This seems to be due a lot of unused vectors assigned for the weight matrix, so it should be possible to reduce it significantly

Steps to reproduce (you'll need to setup [FastText](https://github.com/facebookresearch/fastText) first - 

```
from gensim.models.wrappers import FastText
ft_model = FastText.train(<fast_text_path>, <corpus_file>)
ft_model[<in_vocab_word>]
ft_model[<oov_word>]
```

I've manually verified on a small set that it produces the same results as FastText

@gojomo @tmylk


piskvorky:
Awesome! We want to finalize and merge the `KeyedVector` functionality ASAP, so things don't diverge too much.

jayantj:
What do you think would be the best way to handle OOV words with `KeyedVectors`?
Operations like `most_similar`, `similarity` etc. all need to be modified to allow for out of vocab words. Subclassing `KeyedVectors` and using that subclass for the `FastText` class?

tmylk:
@parthoiiitm This PR can now be updated to use KeyedVecs and merged in  https://github.com/RaRe-Technologies/gensim/pull/847

shashankg7:
@tmylk, I want to take up this issue. Can you please tell me how to start?

tmylk:
@shashankg7 All that is left to do here: merge in the latest changes from develop(KeyedVectors especially) and make it run.

Only 3 files here actually relate to FastText. Other files should not be changed
```
gensim/models/wrappers/__init__.py
gensim/models/wrappers/fasttext.py
test_fasttext_wrapper.py
```
Our [FastText tutorial](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/Word2Vec_FastText_Comparison.ipynb) should also be updated to include a new section on the wrapper.

jayantj:
@tmylk If it's okay, I'll be taking up this issue.

Things that remain - 
 - [ ] merge latest `develop` and ensure it works
 - [ ] ensure word2vec similarity methods which take input words as arguments make use of FastText-like logic for out-of-vocab words
 - [ ] update tutorial
 - [ ] more asserts while loading from the FastText binary file to ensure errors are thrown in case there are breaking changes in the FastText binary format
 - [ ] some more tests

jayantj:
@gojomo From your comment on the original  `KeyedVectors` issue #549 ,
```
Many operations – loading from a plain word2vec.c model, similarity-lookups, (future) approximate-indexing, (future) learned-projections, etc – would live inside or operate on these NamedVectors objects, which could be loaded/saved separately from trainable models. For example, if you only want the vectors and the similarity-operations, why wrap that state in a broken full training model (or keep around the expansive extra state that's not needed)?
```

Do you think it'd be appropriate to have methods like `Word2Vec.load_word2vec_format` and `FastText.load_fasttext_format` return a `KeyedVectors` instance rather than a `Word2Vec` instance, since the objects returned by these methods aren't fully trainable models, simply sets-of-vectors that support the similarity operations?

tmylk:
We should return keyedvectors rather than w2v here. Similar to the WordRank and TensorFlow wrappers.

gensim/models/wrappers/fasttext.py
```diff
@@ -0,0 +1,231 @@
...
+
+This module allows training a word embedding from a training corpus with the additional ability
+to obtain word vectors for out-of-vocabulary words, using the fastText C implementation.
+
+The wrapped model can NOT be updated with new documents for online training -- use gensim's
+`Word2Vec` for that.
+
+Example:
+
+>>> model = gensim.models.wrappers.LdaMallet('/Users/kofola/fastText/fasttext', corpus_file='text8')
```
anmol01gulati: Shouldn't this be gensim.models.wrappers.FastText(..)

jayantj: Yep, fixed. Thanks

jayantj:
All right. Completed most of the pending work. Further work - 

 - [ ] update tutorial
 - [ ] `Word2Vec.load_word2vec_format` and `FastText.train` should return `KeyedVector` instances
 - [ ] check with multiple versions of FastText (if any)

One interesting thing I discovered -

FastText creates a giant matrix (2mil rows by default), initialized to a uniform distribution between `(-1.0 / vector_dim, 1.0 / vector_dim)`
All ngrams are hashed and mapped to this matrix. So whenever an out-of-vocab word is encountered -
  - all ngrams for the word are computed
  - hashes of ngrams are used as index into this giant matrix
  - vectors corresponding to the hashes are summed up to obtain the vector for the word

The interesting point is that vectors for _all_ ngrams are looked up (even ngrams that weren't present in training data and contain the initially randomized values) during this process. I've changed this to only use ngrams which were present in training data. Two advantages - 
  - A large amount of vectors can be discarded (~1.6mil out of 2 mil when training on text8), saving a large amount of memory
  - Theoretically, the vectors for oov words are more accurate. Not significant practically though, since the initially randomized values are very small compared to the final vectors.

Just noting this since this can cause some of the vectors for oov words to be slightly different from the original FastText vectors.

jayantj:
Build is passing now.
Added a tutorial for the [FastText wrapper](https://github.com/jayantj/gensim/blob/2f37b0415d97606e64ef23ffcb4a4a4297cc6a38/docs/notebooks/FastText_Tutorial.ipynb). 
Let me know if there's any changes you think should be made.

Regarding `KeyedVecs` - the WordRank wrapper seems to return a `Word2Vec` instance right now.
I agree it makes sense to have FastText return a `KeyedVecs` instance, but it requires the `Word2Vec.load_word2vec_format` to return a `KeyedVecs` instance too. That seems like a fairly major change, as the `load_word2vec_format` is (probably) a widely used method. It'd also break quite a few tests in the current `test_word2vec.py`.
I think it'd make more sense to have a separate PR for that. Along with that PR, I'll also change the FastText wrapper to return a `KeyedVecs` instance.

jayantj:
On second thoughts, I'm unsure if the FastText wrapper should return a `KeyedVec` instance. The loaded FastText model isn't simply a set of vectors, it contains complete data about the model (hyperparameters and weight matrices). Theoretically, the model can be further trained, we are unable to, currently, because of limitations in the FastText CLI.
Maybe a `FastText` instance makes most sense. The `load_word2vec_format` method in `Word2Vec` definitely should be changed to return a `KeyedVecs` instance.

jayantj:
Pushed a new branch and created a PR #1078 which contains the update to `load_word2vec_format`, tests and the FastText wrapper.

tmylk:
We don't test wrappers for DTM, Mallet etc in Travis as it takes a long time to download the binaries for them.

jayantj:
@tmylk thanks, that sounds good.

@piskvorky I'm unable to request a review via Github, the option doesn't seem to show up under "Reviewers".
Would appreciate if you could look over this once.
Link to the [tutorial for the FastText wrapper](https://github.com/jayantj/gensim/blob/2f37b0415d97606e64ef23ffcb4a4a4297cc6a38/docs/notebooks/FastText_Tutorial.ipynb).

gensim/models/keyedvectors.py
```diff
@@ -46,6 +46,25 @@ def save(self, *args, **kwargs):
...
         super(KeyedVectors, self).save(*args, **kwargs)
 
+    def word_vec(self, word, use_norm=False):
+        """
+        Accept a single word as input.
+        Returns the word's representations in vector space, as a 1D numpy array.
+
+        Example::
+
+          >>> trained_model['office']
```
piskvorky: Misleading example (doesn't use this method at all).

gensim/models/keyedvectors.py
```diff
@@ -90,11 +109,10 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
         for word, weight in positive + negative:
             if isinstance(word, ndarray):
                 mean.append(weight * word)
-            elif word in self.vocab:
-                mean.append(weight * self.syn0norm[self.vocab[word].index])
-                all_words.add(self.vocab[word].index)
             else:
-                raise KeyError("word '%s' not in vocabulary" % word)
+                mean.append(weight * self.word_vec(word, use_norm=True))
+                if word in self.vocab:
```
piskvorky: Dead code, can never reach here (above line would throw a KeyError).

gensim/models/keyedvectors.py
```diff
@@ -230,22 +248,14 @@ def most_similar_cosmul(self, positive=[], negative=[], topn=10):
...
-                raise KeyError("word '%s' not in vocabulary" % word)
-
-        positive = [word_vec(word) for word in positive]
-        negative = [word_vec(word) for word in negative]
+        positive = [self.word_vec(word, use_norm=True) for word in positive]
+        negative = [self.word_vec(word, use_norm=True) for word in negative]
         if not positive:
             raise ValueError("cannot compute similarity with no input")
 
+        all_words = set([self.vocab[word].index for word in positive+negative if word in self.vocab])
```
piskvorky: What is the `all_words` created above for?

gensim/models/keyedvectors.py
```diff
@@ -311,11 +321,18 @@ def doesnt_match(self, words):
         """
         self.init_sims()
 
-        words = [word for word in words if word in self.vocab]  # filter out OOV words
-        logger.debug("using words %s" % words)
         if not words:
             raise ValueError("cannot select a word from an empty list")
-        vectors = vstack(self.syn0norm[self.vocab[word].index] for word in words).astype(REAL)
+        logger.debug("using words %s" % words)
```
piskvorky: Use lazy log formatting (params passed via comma, not directly formatted, which is mostly wasteful because `debug` is not output).

gensim/models/keyedvectors.py
```diff
@@ -311,11 +321,18 @@ def doesnt_match(self, words):
...
+        vectors = []
+        for word in words:
```
piskvorky: The line that initialized `words` was removed above, so how does this work?

The previous version seemed shorter and cleaner, what is the purpose of this explicit loop?

gensim/models/keyedvectors.py
```diff
@@ -311,11 +321,18 @@ def doesnt_match(self, words):
...
+            try:
+                vectors.append(self.word_vec(word))
+            except KeyError:
+                logger.debug("vector for word %s not present, ignoring the word", word)
+        if not vectors:
+            raise ValueError("vector for all given words absent")
```
piskvorky: Confusing message, please rephrase. How about `"cannot compute similarity with no input"` like above, for consistency?

Also wondering whether we should throw an exception in case of missing words, rather than silently ignoring them (DEBUG level message is almost ignore).

gensim/models/word2vec.py
```diff
@@ -469,6 +469,9 @@ def __init__(
             self.build_vocab(sentences, trim_rule=trim_rule)
             self.train(sentences)
 
+    def initialize_word_vectors(self):
+        self.wv = KeyedVectors()  # wv --> word vectors
```
piskvorky: Remove comment, adds nothing.

gensim/models/wrappers/fasttext.py
```diff
@@ -0,0 +1,355 @@
...
+        try:
+            os.remove('%s.vec' % model_file)
+            os.remove('%s.bin' % model_file)
+        except FileNotFoundError:
+            logger.debug('Training files %s not found when attempting to delete', model_file)
+            pass
+
+    def load_binary_data(self, model_binary_file):
+        """Loads data from the output binary file created by FastText training"""
+        with open(model_binary_file, 'rb') as f:
```
piskvorky: Use `smart_open` (everywhere).

gensim/models/wrappers/fasttext.py
```diff
@@ -0,0 +1,355 @@
...
+        for i, ngram in enumerate(all_ngrams):
+            ngram_hash = self.ft_hash(ngram)
+            ngram_indices.append((len(self.wv.vocab) + ngram_hash) % self.bucket)
+            self.wv.ngrams[ngram] = i
+        self.wv.syn0_all = self.wv.syn0_all.take(ngram_indices, axis=0)
+
+    @staticmethod
+    def compute_ngrams(word, min_n, max_n):
+        ngram_indices = []
+        BOW, EOW = ('<','>')  # Used by FastText to attach to all words as prefix and suffix
```
piskvorky: PEP8: space after comma.

piskvorky:
@jayantj left a few comments in the review.

Looking at the notebook:

* What does `Training time for FastText is significantly higher than the Gensim version of FastText` mean? How can a wrapper be faster than the original?
* `atleast` => `at least`

Otherwise looks good 👍

gensim/models/keyedvectors.py
jayantj: The `KeyError` has been removed.

gensim/models/keyedvectors.py
jayantj: To remove the input words from the returned `most_similar` words.

gensim/models/word2vec.py
jayantj: Done

gensim/models/wrappers/fasttext.py
jayantj: Done

gensim/models/wrappers/fasttext.py
jayantj: Done

gensim/models/keyedvectors.py
jayantj: Done

gensim/models/keyedvectors.py
jayantj: `words` was also the name of one of the params of the method, and the removed line was filtering out oov words.
Renamed/reworked things to make things a little clearer now.

gensim/models/keyedvectors.py
jayantj: Changed to a warning for now, that should be noticeable.
Not sure if an exception would be ideal, since it's a change in behaviour, and in case the intention is only to make things more transparent, a warning probably serves the purpose best.

jayantj:
Thanks for the review, much appreciated.
The line `Training time for FastText is significantly higher than the Gensim version of FastText` was meant to be `Training time for FastText is significantly higher than the Gensim version of Word2Vec` - fixed.

gensim/models/keyedvectors.py
piskvorky: Eh, never mind, the review snippet showed me code for `all_words` from `most_similar`. They are unrelated.

Square brackets `[ ]` not needed.

gensim/models/keyedvectors.py
piskvorky: No, it's still there, on line 66.

jayantj: That line raises a `KeyError` in case `word in self.vocab` is `False`. So in case it's `True`, line 115 would be executed.
Also, `word_vec` has been overriden in the `KeyedVectors` subclass for `FastText`.

piskvorky: Yes, my point is -- isn't it always `True`? How could it be `False`, when that would raise an exception at the line above?

gensim/models/keyedvectors.py
piskvorky: OK, thanks.

PR #854, opened 2016-09-12
Process cross index queries in parallel

shubhamchaudhary:
This reduced the processing time from 10hrs to less than 2hrs on a machine with 24 cores (100Gig RAM) and a dataset of size ~635,000. 

screen shot 2017-03-07 at 1 38 29 am

### DIFF
--- a/gensim/similarities/docsim.py
+++ b/gensim/similarities/docsim.py
@@ -394,11 +394,12 @@ class Similarity(interfaces.SimilarityABC):
 
         import multiprocessing
         import functools
-        with multiprocessing.Pool() as pool:
-            worker = functools.partial(_query_chunk_worker, index=self)
-            for result in pool.map(worker, self.iter_chunks()):
-                for sim in result:
-                    yield sim
+        pool = multiprocessing.Pool()
+        worker = functools.partial(_query_chunk_worker, index=self)
+        for result in pool.map(worker, self.iter_chunks()):
+            for sim in result:
+                yield sim
+        pool.terminate()
 
         self.norm = norm  # restore normalization
 
--- a/gensim/similarities/docsim.py
+++ b/gensim/similarities/docsim.py
@@ -50,8 +50,6 @@ uses the faster, batch queries internally and **is ideal for all-vs-all pairwise
 """
 
 
-import concurrent.futures
-import functools
 import logging
 import itertools
 import os
@@ -394,17 +392,12 @@ class Similarity(interfaces.SimilarityABC):
         # turn off query normalization (vectors in the index are already normalized, save some CPU)
         norm, self.norm = self.norm, False
 
-        with concurrent.futures.ProcessPoolExecutor() as executor:
-            futures = []
-            process_chunk = functools.partial(_query_chunk, index=self)
-
-            chunks = list(self.iter_chunks())
-            for chunk in chunks:
-                futures.append(executor.submit(process_chunk, chunk))
-            del chunks
-
-            for future in concurrent.futures.as_completed(futures):
-                for sim in future.result():
+        import multiprocessing
+        import functools
+        with multiprocessing.Pool() as pool:
+            worker = functools.partial(_query_chunk_worker, index=self)
+            for result in pool.map(worker, self.iter_chunks()):
+                for sim in result:
                     yield sim
 
         self.norm = norm  # restore normalization
@@ -466,17 +459,16 @@ class Similarity(interfaces.SimilarityABC):
             os.remove(fname)
 
 
-def _query_chunk(chunk, index):
-    """ Allow pickling """
-    return list(_query_chunk_gen(chunk, index))
+def _query_chunk_worker(chunk, index):
+    """ To allow pickling of method for multiprocessing """
+    def _query_chunk_work():
+        if chunk.shape[0] > 1:
+            for sim in index[chunk]:
+                yield sim
+        else:
+            yield index[chunk]
 
-
-def _query_chunk_gen(chunk, index):
-    if chunk.shape[0] > 1:
-        for sim in index[chunk]:
-            yield sim
-    else:
-        yield index[chunk]
+    return list(_query_chunk_work())
 #endclass Similarity
 
 
--- a/gensim/similarities/docsim.py
+++ b/gensim/similarities/docsim.py
@@ -50,6 +50,8 @@ uses the faster, batch queries internally and **is ideal for all-vs-all pairwise
 """
 
 
+import concurrent.futures
+import functools
 import logging
 import itertools
 import os
@@ -392,12 +394,18 @@ class Similarity(interfaces.SimilarityABC):
         # turn off query normalization (vectors in the index are already normalized, save some CPU)
         norm, self.norm = self.norm, False
 
-        for chunk in self.iter_chunks():
-            if chunk.shape[0] > 1:
-                for sim in self[chunk]:
+        with concurrent.futures.ProcessPoolExecutor() as executor:
+            futures = []
+            process_chunk = functools.partial(_query_chunk, index=self)
+
+            chunks = list(self.iter_chunks())
+            for chunk in chunks:
+                futures.append(executor.submit(process_chunk, chunk))
+            del chunks
+
+            for future in concurrent.futures.as_completed(futures):
+                for sim in future.result():
                     yield sim
-            else:
-                yield self[chunk]
 
         self.norm = norm  # restore normalization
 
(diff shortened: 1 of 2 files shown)


gensim/similarities/docsim.py
```diff
@@ -392,12 +394,18 @@ def __iter__(self):
...
         norm, self.norm = self.norm, False
 
-        for chunk in self.iter_chunks():
-            if chunk.shape[0] > 1:
-                for sim in self[chunk]:
+        with concurrent.futures.ProcessPoolExecutor() as executor:
+            futures = []
+            process_chunk = functools.partial(_query_chunk, index=self)
+
+            chunks = list(self.iter_chunks())
```
tmylk: Better avoid loading entire index into memory by loading the chunk inside the executor.

gensim/similarities/docsim.py
```diff
@@ -456,6 +464,19 @@ def destroy(self):
         for fname in glob.glob(self.output_prefix + '*'):
             logger.info("deleting %s", fname)
             os.remove(fname)
+
+
+def _query_chunk(chunk, index):
+    """ Allow pickling """
+    return list(_query_chunk_gen(chunk, index))
```
tmylk: what is an example use of this method?

gensim/similarities/docsim.py
```diff
@@ -392,12 +394,18 @@ def __iter__(self):
         # turn off query normalization (vectors in the index are already normalized, save some CPU)
         norm, self.norm = self.norm, False
 
-        for chunk in self.iter_chunks():
-            if chunk.shape[0] > 1:
-                for sim in self[chunk]:
+        with concurrent.futures.ProcessPoolExecutor() as executor:
```
tmylk: how big is the speed up from this parallelizatoin?

shubhamchaudhary: Depends on the size of machine, for me it took down a task that took 10hrs down to ~2hrs. 
What is the standard way to measure perf?

gensim/similarities/docsim.py
shubhamchaudhary: workers can't return a generators, since generator can't be pickled. This method's sole purpose in life is to allow returning results from a worker thread/process.

gensim/similarities/docsim.py
tmylk: that sounds like a great speed-up. Could you add description of your machine, data size and speed up into the pr description at the top?

piskvorky:
For `MatrixSimilarity`, each chunk is already processed using multiple threads (if BLAS is configured to use threads). So more parallelization likely won't help, and may hurt (more context switching, memory bus contention).

Could be useful for other types of indexes though (`SparseMatrixSimilarity`, `Annoy` etc).

We cannot use py3k-only features, but that shouldn't be hard to replace/backport.

shubhamchaudhary:
@tmylk Made the changes.

tmylk:
Thanks for the changes. Before merging will need to see the `simspeed2.py` benchmark before and after the parallelisation.

Ideally this parallelisation should be in `SimilarityABC` so that `SparseMatrixSimilarity` and `WMDSimilarity` could take advantage of it as well, but not `MatrixSimilarity` as it uses `numpy.dot` which is parallelised for dense matrices.  I will create another improvement issue for this, unless you wish to make it as a part of this PR. It requires changes to `MatrixSimilarity` to not use parallelisation and to `Similarity` to use `iter` code from `SimilarityABC`.

gensim/similarities/docsim.py
```diff
@@ -392,12 +392,14 @@ def __iter__(self):
         # turn off query normalization (vectors in the index are already normalized, save some CPU)
         norm, self.norm = self.norm, False
 
-        for chunk in self.iter_chunks():
-            if chunk.shape[0] > 1:
-                for sim in self[chunk]:
-                    yield sim
-            else:
-                yield self[chunk]
+        import multiprocessing
```
piskvorky: Imports at the top of the file please.

gensim/similarities/docsim.py
```diff
@@ -392,12 +392,14 @@ def __iter__(self):
...
+        import functools
+        pool = multiprocessing.Pool()
```
piskvorky: How many workers / processes?

gensim/similarities/docsim.py
```diff
@@ -392,12 +392,14 @@ def __iter__(self):
...
+        worker = functools.partial(_query_chunk_worker, index=self)
+        for result in pool.map(worker, self.iter_chunks()):
```
piskvorky: This will be problematic: multiprocessing has no queue limits, it keeps feeding the input queue constantly. This will blow up the memory (queue up all chunks), for large enough index, slow enough workers or slow enough result consumption after `yield`.

piskvorky:
The feature is nice in principle, but the implementation has to be carefully tested on large data. A naive `multiprocessing.map` will take up all memory, I'm afraid we'll need proper input/output queueing.

ludakas:
I am examining the requested changes and I would like to ask if I understand few things correctly and an advice on expected behaviour.  

I suppose that the benchmark is the simspeed2.py in gensim/test directory. It fails for me on the second test. Is it supposed to fail, is this the bug I should correct? (it fails in "correct" place where the blocking should be implemented..)
Here is the simspeed2.py benchmark output for reference:
https://gist.github.com/ludakas/31e78c9ed8adeacfdf6228052e5a2b7a

blocking - if I understand it right the "pool.map(worker, self.iter_chunks())" in "\_\_iter\_\_" function (docsim.py) loads all data which are slowly processed and it fills the memory. I was reading about Queue class in multiprocessing. Intuitively the iter_chunks should be entering the queue and only when needed passed to a worker which computes the result. Is the intuition correct?
I am not sure how to translate this into code with pool.map. I found examples which were using Process..


number of workers - this seems easy using the PARALLEL_SHARDS which are declared at the beginning and passing it to the Pool constructor. However I do not understand why it is commented? How should the pool behave if the multiprocessing is disabled?

reference from the beginning:
```
try:
    import multiprocessing
    # by default, don't parallelize queries. uncomment the following line if you want that.
    # PARALLEL_SHARDS = multiprocessing.cpu_count() # use #parallel processes = #CPus
except ImportError:
    pass
```

tmylk:
@ludakas does `simspeed2.py ` fail even in the `develop` branch outside of this PR?

For queue limits see [example in MulticoreLDA](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ldamulticore.py#L200)

The `num_workers` should be a parameter passed in constructor.

ludakas:
@tmylk yes `simspeed2.py` fails even in the `develop` branch on the second test, same as in this PR. 

Thanks for the queue limits example, however I encountered another problem even before I got to the queues.

At the top of the `docsim.py` file is the import of multiprocessing in try except with an option to uncomment setting the number of PARALLEL_SHARDS to the number of cpus.  When I uncomment it the code fails with the following error:
`
AssertionError: daemonic processes are not allowed to have children`
-->multiprocessing does not allow for the nested pools, which is the case in this situation. I tried a dirty version of creating sub-class of multiprocesing.pool.Pool described in the stackoverflow post below, but the code never finished, maybe my mistake..(it run more than 30 minutes, normally it takes 4 minutes for simspeed2.py).
http://stackoverflow.com/questions/6974695/python-process-pool-non-daemonic

I made another observation - when I keep the PARALLEL_SHARDS commented out (implying `query_shards` being sequential) as is default in the branch and I remove the parallelism  from `__iter__` (here should be implemented the queue) and only keeps there standard for loop shown below, the `simspeed2.py` runs in only 40 seconds, passing all the tests while with the parallel pool  takes 4 minutes to run and it fails the second test. I am not sure what I am doing wrong or whether there is some bug already. 
```
# modified, sequential, runs in 40 seconds
for result in self.iter_chunks():
    for sim in result:
        yield sim

# original, parallel, runs in 4 minutes
pool = multiprocessing.Pool()
worker = functools.partial(_query_chunk_worker, index=self)
for result in pool.map(worker, self.iter_chunks()):
    for sim in result:
        yield sim
pool.terminate()
```

tmylk:
@ludakas Thanks for running the benchmark. A `line_profiler` run is required to find out in more detal, but it  is not too surprising that adding parallelism speeds up some cases and slows down others.

piskvorky:
My 2 cents: that `PARALLEL_SHARDS` code that is commented out is ancient and should probably be just removed. Simply uncommenting it is certainly not expected to work, or do anything meaningful.

tmylk:
It seems that the parallelisation doesn't pass the benchmark with `PARALLEL_SHARDS` commented out:40 seconds old vs 240 seconds new.

PR #864, opened 2016-09-20
CoherenceModel documentation addition

dsquareindia:
`CoherenceModel` documentation should mention which topic models can currently be plugged in as they are without using the `topics` parameter.


dsquareindia:
@tmylk this is just a delta addition. Can this be merged?

PR #872, opened 2016-09-23
Quick Start improvement: print dictionary before print bow to explain what indices are in bag of words

Hitechverma:
The arrangement of In[4] and In[5] was wrong. Rearranged them to desired positions.


tmylk:
Thanks, please change the wording as well.

tmylk:
@Hitechverma Thanks for the PR!

Issue #865, opened 2016-09-20
A numerical bug in word2vec make_cum_table 


pderkowski:
A simple way of recreating:

``` python
from numpy import zeros, uint32

class Dummy(object):
    count = 2;

class Issue(object):
    def __init__(self):
        reps = 10000000
        self.index2word = [i for i in xrange(reps)]
        self.vocab = { i: Dummy() for i in xrange(reps) }

    def make_cum_table(self, power=0.75, domain=2**31 - 1):
        vocab_size = len(self.index2word)
        self.cum_table = zeros(vocab_size, dtype=uint32)
        train_words_pow = float(sum([self.vocab[word].count**power for word in self.vocab]))
        cumulative = 0.0
        for word_index in range(vocab_size):
            cumulative += self.vocab[self.index2word[word_index]].count**power / train_words_pow
            self.cum_table[word_index] = round(cumulative * domain)
        if len(self.cum_table) > 0:
            assert self.cum_table[-1] == domain

def main():
    issue = Issue()
    issue.make_cum_table()

if __name__=="__main__":
    main()
```

On my computer assertion in the last line of make_cum_table fails because self.cum_table[-1] is 2147483646 instead of expected 2147483647. The reason for this is the accumulated numerical error in the computation of `cumulative`.

A simple way to fix this is to replace:

``` python
        for word_index in range(vocab_size):
            cumulative += self.vocab[self.index2word[word_index]].count**power / train_words_pow
            self.cum_table[word_index] = round(cumulative * domain)
```

with:

``` python
        for word_index in range(vocab_size):
            cumulative += self.vocab[self.index2word[word_index]].count**power
            self.cum_table[word_index] = round(cumulative / train_words_pow * domain)
```

tmylk:
Thanks for reporting it and suggesting the fix. Confirm reproduced and fix works. 
Leaving PR creation until tomorrow - it is a good beginner exercise for our sprint at PyCon India tomorrow.

hswick:
Hi I ran into this bug using most recent gensim. Probably not doing this correctly, new user to gensim, but wanted to report that I got this issue with latest version.

python version: Python 2.7.10 :: Anaconda 2.3.0 (64-bit)
gensim version: gensim==0.13.4.1

I am using a large dataset of xml files and not preprocessing them so it made the word count huge. 

Here is the last log I got:

[...]

PR #876, opened 2016-09-24
fix: Numerical bug in word2vec make_cum_table

rhnvrm:
Resolves: #865
Done during PyConIndia 16


tmylk:
@rhnvrm Thanks for the PR!

rhnvrm:
@tmylk Great :+1:

PR #888, opened 2016-09-26
Raise warning if calling load() on an instance rather than the class

dust0x:
Addresses one out of four issues in [#692](https://github.com/RaRe-Technologies/gensim/issues/692)

Makes use of bound functions to differentiate between calls made through `self` and those made through class.


dust0x:
Some errors from my side. Re-opened a new PR.

PR #889, opened 2016-09-26
Throw exception if load() is called on instance rather than the class

dust0x:
Addresses one out of four issues in [#692](https://github.com/RaRe-Technologies/gensim/issues/692)


dust0x:
I'll fix the `global name 'methodize' is not defined` error

gojomo:
I believe a more compact fix, with less meta-method-juggling in multiple places, would be possible and desirable here. Can the load/load_word2vec_format methods just detect is their first parameter (`cls`) isn't the expected class-object?

dust0x:
@gojomo I'd have liked that too, but it seems that `load` does not seem to differentiate between class or self once the call is made.

Here is a digest version of what I'm trying to do here instead:

```
import logging
from types import MethodType

class notWord2Vec:
    def __init__(self):
        self.notLoad = methodize(notLoad, self)
    
    @classmethod
    def notLoad(cls, var):
        print "o%s"%var
        #do something

class notDoc2Vec(notWord2Vec):
    def __init__(self):
        self.notLoad = methodize(notLoad, self)
        pass

def methodize(func, instance):
    return MethodType(func, instance.__class__)

def notLoad(self, var):
    logger.warn("<warning message for %s goes here>"%var)
    print "x%s"%var
    notWord2Vec.notLoad(var)

notWord2Vec.notLoad(0)

w = notWord2Vec()
w.notLoad(1)

notDoc2Vec.notLoad(2)

d = notDoc2Vec()
d.notLoad(3)
```

Still not fully certain what kind of errors does calling by instance cause. Perhaps you could help me understand the nature of this problem better since you raised this issue.

Thanks.

gojomo:
When it's called on an instance, isn't the `cls` parameter an instance rather than the class-object?

gensim/models/word2vec.py
```diff
@@ -1869,3 +1873,33 @@ def __iter__(self):
...
 
     logger.info("finished running %s", program)
+    
+def methodize(func, instance):
+    return MethodType(func, instance, instance.__class__)
+
+
+def load(self, *args, **kwargs):
+    logger.warn('Load was called on instance. Calling on class instead')
+    Word2Vec.load(*args, **kwargs)
```
tmylk: This is not expected behaviour. 
Expected behaviour is to print warning but still run the instance method.

gojomo: I'd suggest that best would be to error – this method doesn't modify an instance in-place, so the "one and only one right way to do it" is to call it is via the class.

dust0x:
@gojomo when calling on instance, `classmethod` takes the calling instance's class as the hidden parameter.

`cls` is of type class-object: as a call to `type (cls)` will show.

gensim/models/word2vec.py
```diff
@@ -452,6 +453,9 @@ def __init__(
         self.total_train_time = 0
         self.sorted_vocab = sorted_vocab
         self.batch_words = batch_words
+        
+        self.load = methodize(load, self)
```
tmylk: Did you mean `self.load = methodize(self.load, self)` ?
it is very confusing to assign `self.load` and then define `def load`

gensim/models/word2vec.py
```diff
@@ -1869,3 +1873,15 @@ def __iter__(self):
         model.accuracy(args.accuracy)
 
     logger.info("finished running %s", program)
+    
+def methodize(func, instance):
```
tmylk: These methods are defined outside of the class. Maybe that is why you see no difference in calling these methods and class methods? Can you link to a gist of the code you use to test the first parameter?

dust0x: This was done on purpose. Please see [here](http://stackoverflow.com/a/861109) for reference.

`MethodType` binds the method to the calling instance. What I'm doing here is something like overloading `load` based on whether it is called on instance or class.

So the argument `load` to `methodize` is the name of the function that is to be bound.

dust0x: My [latest build](https://travis-ci.org/dust0x/gensim/builds/163828265) is failing on Python 2 for some reason, I'm working on that right now. Once that's done, I'll write the tests and merge it back with develop.

gensim/models/word2vec.py
dust0x: This was done on purpose. Please see [here](http://stackoverflow.com/a/861109) for reference.

`MethodType` binds the method to the calling instance. What I'm doing here is something like overloading `load` based on whether it is called on instance or class.

So the argument `load` to `methodize` is the name of the function that is to be bound.

gojomo:
Aha, sorry, I thought an `obj.func_name()` invocation always put `obj` in as the 1st argument. 

This sort of warning would be a nice convenience for explaining a common user error that's been asked-about up a bunch of times (and probably many more times where the user eventually figured it out for themselves)... but it wouldn't justify much extra code or meta-method complexity. 

How about if instead, there's a generic utility function (taking any number of positional/keyword params) that just throws an error to the effect of, "this method name should only be called on the class object" – and that function is installed on instances in the relevant method names? That could be a smaller and clearer way to catch the error.

dust0x:
@gojomo you're right, this is a very inefficient way of doing things.

What if I use `inspect` to get the calling frame and then create a utility that basically parses through the string to figure out if the function was called using `Word2Vec.load` etc.?

gojomo:
This approach seems a bit more focused. But my concern is still complicating the code with advanced meta-features for such a minor convenience API warning.

(Separate observations: as `SaveLoad` might even now be used elsewhere, having a list of acceptable-known-subclasses coded inside its method is suboptimal – it shouldn't have to know all its subclass names. Also, at the moment post-upstream-merge, this PR diff is showing other unrelated changes as well – maybe a Github issue or your catchup-merge wasn't to current state of `develop`?)

dust0x:
@gojomo any solution we find for this *will* require complicating of the code. We've exhausted a lot of possible options till now.

Clearly, though, this approach has much lesser meta-features as compared to others.

> as `SaveLoad` might even now be used elsewhere, having a list of acceptable-known-subclasses coded inside its method is suboptimal – it shouldn't have to know all its subclass names.

This is a valid concern, do you have any recommendations?

>  maybe a Github issue or your catchup-merge wasn't to current state of `develop`?

It's the latter. I'd forgotten about that.

gojomo:
The suggestion in my [20161002 comment](https://github.com/RaRe-Technologies/gensim/pull/889#issuecomment-251011773) might be minimal in terms of line-count and magic. Roughly (not tested – I think this would work):

    def call_on_class_only(*args, **kwargs):
        logger.warn('this method name should only be called on class objects')

(in Word2Vec/etc)

    def __init__(...):
        ...
        self.load = call_on_class_only
        ...

Though perhaps also, if iterating over the `@classmethods` is easy, there'd be a utility function to auto-perform this replacement, so the impact in `Word2Vec.__init__` (etc) would just be:

    ...
    hide_class_only_methods(self)
    ...

(Essentially, this would be a slightly-slimmer refinement of your first approach.)

dust0x:
@gojomo [it works](https://travis-ci.org/dust0x/gensim/builds/170707512). I'll sync-up with master and then commit the changes.

Any idea why there's a write access conflict?

dust0x:
This is odd. It says that the job errored. It never actually started. Is this an error on Travis' side?

dust0x:
@gojomo @tmylk is this ready to be merged?

gensim/utils.py
```diff
@@ -217,6 +217,10 @@ def any2unicode(text, encoding='utf8', errors='strict'):
     return unicode(text, encoding, errors=errors)
 to_unicode = any2unicode
 
+def call_on_class_only(*args, **kwargs):
+    """Warns when load methods are called on instance"""
+    logger.warn('this method name should only be called on class objects')
```
tmylk: Please correct to 'This method should be called on a class object.'

tmylk:
Tests are needed as [here](https://github.com/RaRe-Technologies/gensim/blob/b6e9570c3d5b6b2cf80e175e0c705405b045ca44/gensim/test/test_word2vec.py#L492)

gensim/utils.py
dust0x: Done.

Will the snippet below do for test purpose?

```
@log_capture()
def test_call_on_class_only_for_w2v_load(self, l):
    model.save(testfile())
    w = Word2Vec()
    model = w.load()
    warning = "This method should be called on a class object."
    self.assertTrue(warning in str(l))
```

dust0x:
I see an unrelated error while running the [CI](https://travis-ci.org/RaRe-Technologies/gensim/builds/172051844) in `test_cbow_hs` in Python 3.4 ??

tmylk:
Nevermind that error.

gojomo:
Changes look good to me, as soon as tests are sensible and pass. Thanks for patiently sticking with this!

gensim/utils.py
```diff
@@ -217,6 +217,10 @@ def any2unicode(text, encoding='utf8', errors='strict'):
     return unicode(text, encoding, errors=errors)
 to_unicode = any2unicode
 
+def call_on_class_only(*args, **kwargs):
+    """Warns when load methods are called on instance"""
+    logger.warn('This method should be called on a class object.')
```
piskvorky: We don't want to be failing quietly (or, if logging not configured, completely in silence), ignoring the user's explicit request.

Exception is more appropriate.

gensim/models/word2vec.py
```diff
@@ -432,6 +432,8 @@ def __init__(
 
         """
 
+        self.load = call_on_class_only
```
piskvorky: This will leave the variable unbound, leading to potential confusion / serialization issues.

Proper way would be to bind it properly with `types.MethodType` or somesuch.

gojomo: If the serialization tests pass, and the exception message is understandable if it's ever called (by mistake), I don't see what other clarity would come from binding the method. That'd just make it look more like an instance-specific operation... which it specifically is not.

gensim/test/test_word2vec.py
```diff
@@ -506,7 +506,28 @@ def testTrainWarning(self, l):
...
+        """Test if logger warning is raised when loading doc2vec model on instance"""
+        model = word2vec.Word2Vec(sentences, min_count=1)
+        model.save(testfile())
+        load_on_instance = word2vec.Word2Vec()
+        model = load_on_instance.load(testfile())
+        warning = "Load methods should be called on a class object."
+        self.assertTrue(warning in str(l))
+
+    @log_capture()
+    def testLoadWord2VecOnClassWarning(self):
```
tmylk: why is this second test necessary? The first test above is enough.

gensim/models/word2vec.py
dust0x: @piskvorky, that is what I did at the beginning. Apart from the points @gojomo raises, there are some issues with the way `types.MethodTypes` is handled in Python versions 2 and 3.

gensim/test/test_word2vec.py
dust0x: Yes, you're right. I'll remove that.

gensim/test/test_doc2vec.py
```diff
@@ -306,6 +306,16 @@ def testTrainWarning(self, l):
...
+
+    @log_capture()
+    def testLoadOnClassWarning(self):
+        """Test if logger warning is raised when loading doc2vec model on instance"""
+        model = doc2vec.Doc2Vec(DocsLeeCorpus(), min_count=1)
+        model.save(testfile())
+        load_on_instance = doc2vec.Doc2Vec()
+        model = load_on_instance.load(testfile())
+        warning = "Load methods should be called on a class object."
+        self.assertTrue(warning in str(l))
```
piskvorky: What is this `l`, where does it come from?

Not a good name for a variable.

gensim/test/test_word2vec.py
```diff
@@ -506,7 +506,28 @@ def testTrainWarning(self, l):
...
+        """Test if logger warning is raised when loading doc2vec model on instance"""
+        model = word2vec.Word2Vec(sentences, min_count=1)
+        model.init_sims()
+        model.save_word2vec_format(testfile(), binary=True)
+        binary_model = word2vec.Word2Vec()
+        load_on_instance = binary_model.load_word2vec_format(testfile(), binary=True)
+        warning = "Load methods should be called on a class object."
+        self.assertTrue(warning in str(l))
```
piskvorky: Dtto -- what is `l`?

gensim/utils.py
piskvorky: @dust0x please fix.

gensim/test/test_doc2vec.py
gojomo: Unfortunately it seems to be the 'usual' name for the LogCapture objects provided by the testfixtures package `@log_capture` decorator. (See for example: https://testfixtures.readthedocs.io/en/latest/logging.html#the-decorator) But, that would seem to require an `l` parameter be added to the method, which I don't see here... so a bit baffled that this could work.

dust0x: No of course it wouldn't work. Sorry, I was trying something you weren't supposed to see this.

dust0x:
@gojomo `call_on_class_only` now raises an exception. Two questions:

1. Is `AttributeError` the right error to raise?

2. Are unittests written for exceptions?

dust0x:
@tmylk @gojomo `call_on_class_only` now raises an exception. Two questions:

1. Is `AttributeError` the right error to raise?

2. I couldn't find any unit-test cases for exceptions. If it needs to be done, can you please point me to something related? I'll quickly write it up.

dust0x:
- Method `call_on_class_only` now raises `AttributeError` as requested.

- Tests for load on instance completed.

- CI passed.

Sorry for the delay. Request to review.

dust0x:
@tmylk @gojomo please review.

Issue #853, opened 2016-09-11
Unpickling models across python3 and python2


jayantj:
Training and saving the model with python3 -

```
from gensim.test.test_word2vec import LeeCorpus
from gensim.models.word2vec import Word2Vec
model = Word2Vec(LeeCorpus(), size=10)
model.save('temp-py3')
```

Loading it with python2.6 or 2.7 -

```
from gensim.models.word2vec import Word2Vec
model = Word2Vec.load('temp-py3')
```

Raises this error - 

```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    model = Word2Vec.load('temp-py3')
  File "gensim/models/word2vec.py", line 1684, in load
    model = super(Word2Vec, cls).load(*args, **kwargs)
  File "gensim/utils.py", line 248, in load
    obj = unpickle(fname)
  File "gensim/utils.py", line 911, in unpickle
    return _pickle.loads(f.read())
AttributeError: 'module' object has no attribute 'defaultdict'
```

This goes the other way around too (loading models saved with python2.7 in python3), although with a different error - 

```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    m = Word2Vec.load('w2v_py2.7')
  File "/home/jayant/Projects/gensim/gensim/models/word2vec.py", line 1684, in load
    model = super(Word2Vec, cls).load(*args, **kwargs)
  File "/home/jayant/Projects/gensim/gensim/utils.py", line 252, in load
    obj = unpickle(fname)
  File "/home/jayant/Projects/gensim/gensim/utils.py", line 915, in unpickle
    return _pickle.loads(f.read())
UnicodeDecodeError: 'ascii' codec can't decode byte 0xfb in position 0: ordinal not in range(128)
```

jayantj:
The issue with unpickling a model pickled in python3 with python2.7 results from a bug in pickle which incorrectly maps `collections` to `UserList/UserString`
Here are some details - https://bugs.python.org/issue18473
The bug has been fixed in later versions of pickle, but python3.4 users are likely to run into issues.

The issue with unpickling a python2.7 model in python3 is because of encoding issues with numpy arrays - http://stackoverflow.com/questions/28218466/unpickling-a-python-2-object-with-python-3
Decoding with 'latin1' seems to work, though it could cause issues in case the pickle contains both numpy arrays and non-ISO-8859 characters

tmylk:
[...]

PR #913, opened 2016-10-03
Loading and Saving LDA Models across Python 2 and 3.

anmol01gulati:
Modified load and save methods in ldamodel.py to manage compatibility issues when loading and saving models across Python 2 and 3.
This PR tackles Isssue #853


tmylk:
Thanks. Could you please add tests for these methods? Maybe we need to add 2 pickled models to the test_data folder, one for Python 2 and one for Python 3?

anmol01gulati:
Yes sure, I was gonna ask for suggestions on testing this actually.

anmol01gulati:
So, I added two saved LDA models, one in Python 2.7 and other in Python 3.5 environments in test_data folder. The method I used to create these models is in test_ldamodel.py.
Also added a test to load these models and compare(both must be the same actually).
But right now, the test Fails in Python 2.7 environment and Gives an error (same as referred in [FAQ](https://github.com/RaRe-Technologies/gensim/wiki/Recipes-&-FAQ#q9-how-do-i-load-a-model-in-python-3-that-was-trained-and-saved-using-python-2)).
@tmylk Any suggestions?

anmol01gulati:
So I got this working. The LDA model loads fine now across both in Python 2 and 3.
The test fails in checking expElogbeta array equality in the two saved models. But this is just due to precision differences, they are different when the model was created itself. It's not a problem of persistence. We should probably remove this check, or add other checks for equality of models.
Otherwise, the tests pass.

tmylk:
How different are they? Can you add a comparison with an epsilon of each other?

anmol01gulati:
Well, the epsilon is big for some indexes, I'm adding the exact arrays I printed.
(Python 2.7 expElogbeta: [[ 0.00685637,  0.01305742,  0.00683859,  0.09222717,  0.09222721,
 0.09235907,  0.00893736,  0.07625897,  0.00682716,  0.14222405, 0.14247683,  0.09542658],
[ 0.10647068,  0.09757566,  0.10649878,  0.01035072,  0.01035068, 0.01023777,  0.2084948 ,  0.07635718,  0.10651685,  0.00796197, 0.00776243,  0.00771203]])

(Python 3.5 expElogbeta: [[ 0.00683436,  0.00685224,  0.01325946,  0.00895329,  0.09220773,  0.0764139, 0.09233712,  0.09220773,  0.006823, 0.14213993,  0.14239321,  0.09537042]
[ 0.10656816,  0.10653988,  0.09735885, 0.20859865,  0.01032817,  0.07618502,
  0.01021732,  0.01032817,  0.10658615,  0.00796839,  0.00776821,  0.00771765]])

Also, the id2word dictionary saved in both Python formats is not in the same order.

gensim/models/ldamodel.py
```diff
@@ -1023,7 +1024,41 @@ def save(self, fname, ignore=['state', 'dispatcher'], *args, **kwargs):
...
+        # someone sets the separately list themselves.
+        separately_explicit = ['expElogbeta', 'sstats']
+        # Also add 'alpha' and 'eta' to separately list if they are set 'auto' or some 
+        # array manually.
+        if (isinstance(self.alpha, six.string_types) and self.alpha == 'auto') or len(self.alpha.shape) != 1:
+            separately_explicit.append('alpha')
+        if (isinstance(self.eta, six.string_types) and self.eta == 'auto') or len(self.eta.shape) != 1:
+            separately_explicit.append('eta')
+        # Merge separately_explicit with separately.
+        if separately is not None and separately:
```
tmylk: `if separately` is enough

gensim/models/ldamodel.py
```diff
@@ -1037,6 +1072,18 @@ def load(cls, fname, *args, **kwargs):
         """
         kwargs['mmap'] = kwargs.get('mmap', None)
         result = super(LdaModel, cls).load(fname, *args, **kwargs)
+        # Load the separately stored id2word dictionary saved in json.
+        id2word_fname = utils.smart_extension(fname, '.json')
```
tmylk: please make all files for one model in a special folder, so it is easy to keep track

tmylk:
This solution looks really complicated. Let's try something simpler. Let's save the Python version with the pickle. If the versions differ on load than raise an exception to use Dill. (assuming dill works across versions)

gensim/models/ldamodel.py
anmol01gulati: Made changes as suggested.

anmol01gulati:
@tmylk Actually, I think no more changes need to be made for compatibility in word2vec or doc2vec models. The load/save methods work fine now for them. We could probably use the present pickle methods itself.
I've added tests now to Check compatibility for Word2Vec Models. The models load without errors. But the tests fail though in testing equality of models across python versions. I also added the code I used to create the word2vec models in test_word2vec.py (commented code). Please review.

tmylk:
Add an epsilon for equality comparison. 
Please `git fetch, git merge develop`, so the PR becomes reviewable (there are unrelated commits).

gensim/test/test_word2vec.py
```diff
@@ -474,11 +487,11 @@ def testRNG(self):
 
     def models_equal(self, model, model2):
         self.assertEqual(len(model.vocab), len(model2.vocab))
-        self.assertTrue(numpy.allclose(model.syn0, model2.syn0))
+        self.assertTrue(numpy.allclose(model.syn0, model2.syn0, atol=1e-4))
```
tmylk: how big is the difference on your machine?

gensim/test/test_word2vec.py
```diff
@@ -233,6 +233,19 @@ def testPersistenceWord2VecFormatCombinationWithStandardPersistence(self):
...
         self.assertEqual(model.vocab['human'].count, binary_model_with_vocab.vocab['human'].count)
+    
+    # def testSaveModelsForPythonVersion(self):
+    #     fname = os.path.join(os.path.dirname(__file__), 'word2vecmodel_python_3_5')
+    #     model = word2vec.Word2Vec(sentences, size=10, min_count=0, seed=42, hs=1, negative=0)
+    #     model.save(fname)
+    #     logging.warning("Word2Vec model saved")
+    
+    def testModelCompatibilityWithPythonVersions(self):
+        fname_model_2_7 = os.path.join(os.path.dirname(__file__), 'word2vecmodel_python_2_7')
```
piskvorky: Use `module_path` and `datapath` defined above.

gensim/models/ldamodel.py
```diff
@@ -1032,6 +1067,18 @@ def load(cls, fname, *args, **kwargs):
...
+        try:
+            with utils.smart_open(id2word_fname, 'r') as fin:
```
piskvorky: Open file as binary, decode as necessary (if necessary).

gensim/models/ldamodel.py
```diff
@@ -1018,7 +1019,41 @@ def save(self, fname, ignore=['state', 'dispatcher'], *args, **kwargs):
...
+            separately = [e for e in separately if e] # make sure None and '' are not in the list
+            separately = list(set(separately_explicit) | set(separately))
+        else:
+            separately = separately_explicit
+        
+        # id2word needs to saved separately. 
+        # If id2word is not already in ignore, then saving it separately in json.
+        id2word = None
+        if self.id2word is not None and 'id2word' not in ignore:
+            id2word = dict((k,v) for k,v in self.id2word.iteritems())
```
piskvorky: PEP8: space after comma.

gensim/models/ldamodel.py
```diff
@@ -1018,7 +1019,41 @@ def save(self, fname, ignore=['state', 'dispatcher'], *args, **kwargs):
...
+        self.id2word = None # remove the dictionary from model
+        super(LdaModel, self).save(fname, ignore=ignore, separately = separately, *args, **kwargs)
+        self.id2word = id2word # restore the dictionary.
+
+        # Save the dictionary separately in json.
+        id2word_fname = utils.smart_extension(fname, '.json')   
+        try:
+            with utils.smart_open(id2word_fname, 'w', encoding='utf-8') as fout:
+                json.dump(id2word, fout)
```
piskvorky: Better open the output as binary and write encoded utf8 to it.

Actually, the `json` module already produces binary strings in `dump` AFAIK, so what is this even for?

gensim/utils.py
```diff
@@ -907,10 +907,12 @@ def pickle(obj, fname, protocol=2):
 
 def unpickle(fname):
     """Load pickled object from `fname`"""
-    with smart_open(fname) as f:
+    with smart_open(fname, 'rb') as f:
         # Because of loading from S3 load can't be used (missing readline in smart_open)
-        return _pickle.loads(f.read())
-
+        if sys.version_info > (3,0):
```
piskvorky: PEP8: space after comma.

tmylk:
Merged in #1039

Issue #857, opened 2016-09-19
DTM ldaseqmodel tutorial notebook - Corpus and cleaning instructions


einon:
In code cell 2, neither of these files are available, with no instructions on how to obtain them:
# loading our corpus and dictionary

dictionary = Dictionary.load('Corpus/news_dictionary')
corpus = bleicorpus.BleiCorpus('Corpus/news_corpus')

bhargavvader:
Hello @einon , I didn't include these test files and they were just an example. You can find the data set here: [link](https://github.com/derekgreene/dynamic-nmf/tree/master/data).

Thanks for pointing this out, I'll add this information to the notebook as well.

tmylk:
Do you have the pre-processing code somewhere? It would be great to add it to the tutorial.

tmylk:
Ping @bhargavvader

bhargavvader:
@tmylk I was wondering if I should just upload the corpus and dictionary to load - it's easier, because my cleaning and pre-processing wasn't very pretty.

tmylk:
Sounds good if it's not too big

piskvorky:
@bhargavvader @tmylk status here? I see some data files have been merged, but have the notebook instructions been updated?

bhargavvader:
I've included the necessary data files and mentioned it requires basic pre-processing, and linked to basic [gensim pre-processing](https://radimrehurek.com/gensim/tut1.html#from-strings-to-vectors). I haven't included detailed instructions on the pre-processing.

einon:
This fixes the original issue, so happy to close. Thanks!

bhargavvader:
:)

PR #921, opened 2016-10-06
Added datasets

bhargavvader:
@tmylk , this is with regard to #857.

I've added the datasets - it is around 2 mb, which I do not think is too big. 
Would this be fine?


tmylk:
Just make sure they are excluded from the package in manifest.in but still run in travis ipynb checks...

bhargavvader:
How exactly do I check for that?

tmylk:
You need to exclude them in manifest.in but still see the tests of ipynb being run in travis logs

bhargavvader:
Cool, done!

tmylk:
Does this ipynb now run in travis? Can it be seen in travis log?

Issue #901, opened 2016-09-29
hdp_to_lda should return LdaModel instance


dsquareindia:
I was recently using this feature and found that the documentation doesn't mention anything about what [this](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/hdpmodel.py#L510) function returns. It would also help if a short code snippet can be inserted telling that for using this in an LDA model, `alpha` can be set in the `LdaModel` init and `ldam.expElogbeta[:]` can be set as the `beta` in the new LdaModel.

dsquareindia:
@tmylk I was wondering that wouldn't it be better to return a proper LdaModel instance itself as do in say `malletmodel2ldamodel`?

tmylk:
Agree that returning an actual `LDAModel` is more explicit. That would be a good improvement.

dsquareindia:
Alright! I'll wait for a while for someone to take this up.

bhargavvader:
I don't mind giving this a shot! Though I can try some other issue if this would be a good `beginner` issue for someone else.

dsquareindia:
@bhargavvader are you taking this up? I'll take this up this weekend otherwise :)

bhargavvader:
Yup, on it!

bhargavvader:
Also, @dsquareindia , the reason it returns `alpha` and `beta` and not a model is because in it's inference it uses that method to get the updated `alpha` and `beta` values. You can see line [246](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/hdpmodel.py#L246) for an example of this.

So what I'm going to do is add an extra parameter, which instead returns an `ldamodel` instead of the alpha and beta values, but let the default behaviour be to return these values, and something in the docs about how this needs to be done.

@tmylk is that ok?

tmylk:
A new method would be better.

bhargavvader:
Yeah, #996 implements a new method.

bhargavvader:
@dsquareindia can this be closed? #1055 fixes this.

PR #922, opened 2016-10-06
HDP to LDA model

bhargavvader:
With respect to #901 , the PR returns an LDA model if the user requests, and also adds some documentation and what is going on.

Also included a test for the same.

### DIFF
--- a/gensim/models/hdpmodel.py
+++ b/gensim/models/hdpmodel.py
@@ -38,6 +38,7 @@ import numpy as np
 import scipy.special as sp
 
 from gensim import interfaces, utils, matutils
+from gensim.models import ldamodel
 from six.moves import xrange
 
 logger = logging.getLogger(__name__)
@@ -507,9 +508,13 @@ class HdpModel(interfaces.TransformationABC):
             fout.write('eta: %s\n' % str(self.m_eta))
             fout.write('gamma: %s\n' % str(self.m_gamma))
 
-    def hdp_to_lda(self):
+    def hdp_to_lda(self, model=False):
         """
         Compute the LDA almost equivalent HDP.
+        By default returns the corresponding alpha and beta values, where alpha can be set as the ldamodel init,
+        and ldam.expElogbeta[:] can be set as the beta in the new LdaModel.
+        This is the default value to be returned as this method is used in the HDP update methods.
+        If explicitly mentioned that a model is required to be returned, it returns the corresponding model.
         """
         # alpha
         sticks = self.m_var_sticks[0] / (self.m_var_sticks[0] + self.m_var_sticks[1])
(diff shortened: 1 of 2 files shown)


gensim/models/hdpmodel.py
```diff
@@ -507,9 +508,13 @@ def save_options(self):
             fout.write('eta: %s\n' % str(self.m_eta))
             fout.write('gamma: %s\n' % str(self.m_gamma))
 
-    def hdp_to_lda(self):
+    def hdp_to_lda(self, model=False):
```
dsquareindia: I would rather deprecate the older behaviour of returning `alpha` and `beta` and only have the feature of returning a model. There should also be parameters on number of iterations and gamma_threshold to be used for the LdaModel. See [this](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/wrappers/ldamallet.py#L354) for an example.

gensim/models/hdpmodel.py
```diff
@@ -525,6 +530,12 @@ def hdp_to_lda(self):
         beta = (self.m_lambda + self.m_eta) / (self.m_W * self.m_eta + \
                 self.m_lambda_sum[:, np.newaxis])
 
+        # if the user requests to return an ldamodel instead, it is returned with the appropriate alpha and beta values
+        if model:
+            ldam = ldamodel.LdaModel(num_topics=150, alpha=alpha, id2word=self.id2word)
```
dsquareindia: The num_topics should be what HdpModel determines as the ideal number of topics isn't it?

gensim/models/hdpmodel.py
bhargavvader: You cannot do this. Please refer to comment on your issue.

gensim/models/hdpmodel.py
bhargavvader: No, because the alpha returned here is for 150 topics. If we wish to do that then we have to change the shape of alpha.

gensim/models/hdpmodel.py
bhargavvader: I do agree about the parameters of number of iterations and gamma, but I couldn't find an equivalent of that in `hdp`. Any ideas or suggestions?

bhargavvader:
ping @tmylk

gensim/test/test_hdpmodel.py
```diff
@@ -64,6 +64,9 @@ def testShowTopics(self):
                 self.assertTrue(isinstance(k, six.string_types))
                 self.assertTrue(isinstance(v, float))
 
+    def testHDP_LDA(self):
+        ldam = self.model.hdp_to_lda(model=True)
+        self.assertEqual(ldam.expElogbeta.shape, self.model.lda_beta.shape)
```
tmylk: please add more assertions

gensim/models/hdpmodel.py
tmylk: It would be better to have a new method called "suggested_lda_model" that returns the model with reshaped alpha

bhargavvader: Agreed!

tmylk:
Ping @bhargavvader

bhargavvader:
Sorry, last weeks have been busy! Will get to this.

bhargavvader:
@tmylk I'm closing this PR and opening a new HDP PR to address all the HDP related issues. The current PR needs lots of changes anyway!

PR #925, opened 2016-10-08
Removes "No handlers could be found for logger "gensim.models.word2vec", fix #910

### LINKED ISSUE #910: Remove "No handlers could be found for logger "gensim.models.word2vec" message during import
Instead print the actual message that word2vec is trying to print

markroxor:
Creates a default handler for logger.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -96,7 +96,7 @@ from six.moves import xrange
 from types import GeneratorType
 
 logger = logging.getLogger(__name__)
-logging.basicConfig(format='%(asctime)s : %(levelname)s : %(message)s') is better
+logging.basicConfig(format='%(asctime)s : %(levelname)s : %(message)s')
 
 try:
     from gensim.models.word2vec_inner import train_batch_sg, train_batch_cbow
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -96,7 +96,7 @@ from six.moves import xrange
 from types import GeneratorType
 
 logger = logging.getLogger(__name__)
-logging.basicConfig()
+logging.basicConfig(format='%(asctime)s : %(levelname)s : %(message)s') is better
 
 try:
     from gensim.models.word2vec_inner import train_batch_sg, train_batch_cbow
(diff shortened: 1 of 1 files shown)


markroxor:
ping @tmylk

gensim/models/word2vec.py
```diff
@@ -96,6 +96,7 @@
 from types import GeneratorType
 
 logger = logging.getLogger(__name__)
+logging.basicConfig()
```
tmylk: logging.basicConfig(format='%(asctime)s : %(levelname)s : %(message)s') is better

markroxor:
Acknowledged.

tmylk:
Thanks for the PR!

piskvorky:
Absolutely not, please undo the merge.

It is no business for a library to be configuring *application* logging.

tmylk:
basicConfig takes effect only if nothing is configured in the application. If the application has configured logging then there will be no changes from this line. It's just a default fallback.

piskvorky:
Regardless -- we don't want to be configuring logging at all, anywhere (except in main).

This is bad practice and harmful (application's own logging will be ignored if configured after imports... which is common).

tmylk:
Agreed. It would be nice to tell the user they need to configure logging by calling basicConfig ? The Python logging error message 'No handlers could be found' is too confusing. We could create a handler, log to console "Please run BasicConfig" plus word2vec message, and then remove the handler.

tmylk:
@markroxor. Let's undo this and move the message to __init__ instead of import as logging is configured in an application after imports.

markroxor:
On it.

piskvorky:
I already forgot, please help me: is "No handlers could be found" output once per call, once per module, once per process?

@gojomo thoughts on a good, general way to let users know logs are being suppressed (because they didn't configure logging)?

markroxor:
image

Once the module is imported the warning doesn't appear again.

tmylk:
what is the issue with logging in the `__init__` call? By that time the application will have logger configured.

Logger is configured after import so there will always be a problem if we log during `import`.

markroxor:
@piskvorky You were right. Executing the word2vec.py file after the above merge results in suppressed logger output. 
image

If I instead move the line [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec.py#L108) the logger ouput is visible.
image (but that is possibly because I am using the 'slow version')

tmylk:
We should not be logging anything in the import part of the code. If no handler exists, then it creates a default handler on OSX. That stops the user from running `basicConfig` later. 
On Linux we get the error. Both outcomes are bad. Please move the message to `__init__`

Issue #795, opened 2016-07-17
Doc2Vec defaults shouldn't override Word2Vec defaults without good reason


gojomo:
Following #534, Word2Vec matches the word2vec.c defaults (as of the time Google stopped updating/maintaining that code). However, where our Doc2Vec class takes its own parameters – `hs`, `negative`, `sample`, `window`, and `dm_mean` – it's still using the older defaults. There's not really any case that these older defaults are better – and in some cases they're probably worse. 

Unless there's a reason for different defaults, Doc2Vec should use the same values as Word2Vec... ideally by not specifying them at all (to avoid duplication). This probably means removing those params from the `__init__()` signature – but still documenting them in the doc-comment – so that the extra keywords just pass through to `Word2Vec.__init__()` if/when present.

SebastinSanty:
Hi! I would like to work on this bug.

PR #927, opened 2016-10-09
Doc2Vec defaults shouldn't override Word2Vec defaults without good reason (#795)

SebastinSanty:


gensim/models/doc2vec.py
```diff
@@ -529,10 +529,9 @@ def repeat(self, word_count):
 
 class Doc2Vec(Word2Vec):
     """Class for training, using and evaluating neural networks described in http://arxiv.org/pdf/1405.4053v2.pdf"""
-    def __init__(self, documents=None, size=300, alpha=0.025, window=8, min_count=5,
-                 max_vocab_size=None, sample=0, seed=1, workers=1, min_alpha=0.0001,
-                 dm=1, hs=1, negative=0, dbow_words=0, dm_mean=0, dm_concat=0, dm_tag_count=1,
-                 docvecs=None, docvecs_mapfile=None, comment=None, trim_rule=None, **kwargs):
+    def __init__(self, documents=None, window=8, sample=0, dm=1, hs=1, 
+                negative=0, dbow_words=0, dm_mean=0, dm_concat=0, dm_tag_count=1,
+                 docvecs=None, docvecs_mapfile=None, comment=None, **kwargs):
```
gojomo: Since the Word2Vec defaults now include `window=5, sample=1e-3, hs=0, negative=5, cbow_mean=1` some of these values are still overriding those. (`cbow_mean` in Word2Vec is essentially analogous to `dm_mean` in Doc2Vec.) Also, the indention is a little off.

gensim/models/doc2vec.py
```diff
@@ -643,7 +642,7 @@ def reset_from(self, other_model):
         self.docvecs.borrow_from(other_model.docvecs)
         super(Doc2Vec, self).reset_from(other_model)
 
-    def scan_vocab(self, documents, progress_per=10000, trim_rule=None, update=False):
+    def scan_vocab(self, documents, progress_per=10000, trim_rule=None):
```
gojomo: Looks like the `update` parameter may have been inadvertently removed?

SebastinSanty:
Thanks for the review @gojomo

gojomo:
Looking better, but (1) looks like there's still an automated-tests failure; (2) seems #929 is tackling the same issue – are you both at the same sprint?

SebastinSanty:
I guess @markroxor is pursuing his PR actively. I think, this PR can be closed.

PR #929, opened 2016-10-09
Removing Doc2Vec defaults so that it won't override Word2Vec defaults. fix #795

### LINKED ISSUE #795: Doc2Vec defaults shouldn't override Word2Vec defaults without good reason

markroxor:
Removing Doc2Vec defaults so that it won't override Word2Vec defaults.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -529,7 +529,7 @@ class Doctag(namedtuple('Doctag', 'offset, word_count, doc_count')):
 
 class Doc2Vec(Word2Vec):
     """Class for training, using and evaluating neural networks described in http://arxiv.org/pdf/1405.4053v2.pdf"""
-    def __init__(self, documents=None,
+    def __init__(self, documents=None,dm_mean=None,
                  dm=1, dbow_words=0, dm_concat=0, dm_tag_count=1,
                  docvecs=None, docvecs_mapfile=None, comment=None, trim_rule=None, **kwargs):
         """
@@ -600,9 +600,8 @@ class Doc2Vec(Word2Vec):
 
         """
 
-
         super(Doc2Vec, self).__init__(
-            sg=(1 + dm) % 2,
+            sg=(1 + dm) % 2, dm_mean=dm_mean,
             null_word=dm_concat, **kwargs)
 
         self.dbow_words = dbow_words
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -529,7 +529,7 @@ class Doctag(namedtuple('Doctag', 'offset, word_count, doc_count')):
 
 class Doc2Vec(Word2Vec):
     """Class for training, using and evaluating neural networks described in http://arxiv.org/pdf/1405.4053v2.pdf"""
-    def __init__(self, documents=None,dm_mean=None,
+    def __init__(self, documents=None, dm_mean=None,
                  dm=1, dbow_words=0, dm_concat=0, dm_tag_count=1,
                  docvecs=None, docvecs_mapfile=None, comment=None, trim_rule=None, **kwargs):
         """
@@ -601,8 +601,11 @@ class Doc2Vec(Word2Vec):
         """
 
         super(Doc2Vec, self).__init__(
-            sg=(1 + dm) % 2, dm_mean=dm_mean,
+            sg=(1 + dm) % 2,
             null_word=dm_concat, **kwargs)
+
+        if dm_mean is not None:
+            self.cbow_mean = dm_mean
 
         self.dbow_words = dbow_words
         self.dm_concat = dm_concat
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -529,9 +529,8 @@ class Doctag(namedtuple('Doctag', 'offset, word_count, doc_count')):
 
 class Doc2Vec(Word2Vec):
     """Class for training, using and evaluating neural networks described in http://arxiv.org/pdf/1405.4053v2.pdf"""
-    def __init__(self, documents=None, size=300, alpha=0.025, window=8, min_count=5,
-                 max_vocab_size=None, sample=0, seed=1, workers=1, min_alpha=0.0001,
-                 dm=1, hs=1, negative=0, dbow_words=0, dm_mean=0, dm_concat=0, dm_tag_count=1,
+    def __init__(self, documents=None,
+                 dm=1, dbow_words=0, dm_concat=0, dm_tag_count=1,
                  docvecs=None, docvecs_mapfile=None, comment=None, trim_rule=None, **kwargs):
         """
         Initialize the model from an iterable of `documents`. Each document is a
@@ -601,9 +600,8 @@ class Doc2Vec(Word2Vec):
 
         """
 
+
         super(Doc2Vec, self).__init__(
-            size=size, window=window, min_count=min_count,
-            sample=sample, workers=workers,
             sg=(1 + dm) % 2,
             null_word=dm_concat, **kwargs)
 
@@ -612,8 +610,7 @@ class Doc2Vec(Word2Vec):
         self.dm_tag_count = dm_tag_count
         if self.dm and self.dm_concat:
             self.layer1_size = (self.dm_tag_count + (2 * self.window)) * self.vector_size
-        else:
-            self.layer1_size = self.size
+
         self.docvecs = docvecs or DocvecsArray(docvecs_mapfile)
         self.comment = comment
         if documents is not None:
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -529,9 +529,9 @@ class Doctag(namedtuple('Doctag', 'offset, word_count, doc_count')):
 
 class Doc2Vec(Word2Vec):
     """Class for training, using and evaluating neural networks described in http://arxiv.org/pdf/1405.4053v2.pdf"""
-    def __init__(self, documents=None, size=300, alpha=0.025, window=8, min_count=5,
-                 max_vocab_size=None, sample=0, seed=1, workers=1, min_alpha=0.0001,
-                 dm=1, hs=1, negative=0, dbow_words=0, dm_mean=0, dm_concat=0, dm_tag_count=1,
+    def __init__(self, documents=None, size=300, alpha=0.025, min_count=5,
+                 max_vocab_size=None, seed=1, workers=1, min_alpha=0.0001,
+                 dm=1, dbow_words=0, dm_concat=0, dm_tag_count=1,
                  docvecs=None, docvecs_mapfile=None, comment=None, trim_rule=None, **kwargs):
         """
         Initialize the model from an iterable of `documents`. Each document is a
@@ -601,9 +601,9 @@ class Doc2Vec(Word2Vec):
 
         """
         super(Doc2Vec, self).__init__(
-            size=size, alpha=alpha, window=window, min_count=min_count, max_vocab_size=max_vocab_size,
-            sample=sample, seed=seed, workers=workers, min_alpha=min_alpha,
-            sg=(1+dm) % 2, hs=hs, negative=negative, cbow_mean=dm_mean,
+            size=size, alpha=alpha, min_count=min_count, max_vocab_size=max_vocab_size,
+            seed=seed, workers=workers, min_alpha=min_alpha,
+            sg=(1+dm) % 2,
             null_word=dm_concat, **kwargs)
         self.dbow_words = dbow_words
         self.dm_concat = dm_concat
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -529,7 +529,9 @@ class Doctag(namedtuple('Doctag', 'offset, word_count, doc_count')):
 
 class Doc2Vec(Word2Vec):
     """Class for training, using and evaluating neural networks described in http://arxiv.org/pdf/1405.4053v2.pdf"""
-    def __init__(self, documents=None, dm=1, dbow_words=0, dm_concat=0, dm_tag_count=1,
+    def __init__(self, documents=None, size=300, alpha=0.025, window=8, min_count=5,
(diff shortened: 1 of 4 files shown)


markroxor:
@gojomo please review.

gojomo:
Some thoughts: 

* even though I can click through to #795, it helps to have a more-descriptive PR title to establish context
* this is a good start – but looks to be still failing auto testing, and it's best for that to be resolved before giving close human review
* it looks like PR #927 is tacking the same issue – are you both at the same sprint?

markroxor:
@gojomo 
I've added a relevant PR description, sorry about that.
I just need a little help, the Travis CI traceback suggets that -
`Traceback (most recent call last):
  File "/home/travis/build/RaRe-Technologies/gensim/gensim/test/test_doc2vec.py", line 190, in test_dmm_hs
    alpha=0.05, min_count=2, iter=20)
  File "/home/travis/build/RaRe-Technologies/gensim/gensim/models/doc2vec.py", line 603, in __init__
    null_word=dm_concat, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'dm_mean'`

I don't get the error. There is no "dm_mean" variable in either word2vec.py or doc2vec.py. Am I missing something? 
thanks

gojomo:
There *was* a `dm_mean` before your edits, and the tests passed then – so look at the effects of that change!

markroxor:
In #795 it was said that 

>  ideally by not specifying them at all (to avoid duplication). This probably means removing those params from the __init__() signature

but according to the CI test files, in this [line](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_doc2vec.py#L189) we have to initialize params like dm_mean, window etc. So to implement this either we can change the test files or hard-code the default params from `word2vec.py` file in the `__init__` method of `doc2vec.py`.

gojomo:
The added description is helpful – but a descriptive title (as edited atop the PR page) is even better, as it then appears in the Github list/notification views. 

`dm_mean` is the trickiest parameter, as it is essentially a synonym for `cbow_mean` but still (like the others) should only override the default if explicitly specified by the caller. 

We don't want to change the tests – holding them stable through any changes helps assure us that we're not totally breaking user code. (We're still altering the effective default, which may affect user results.)

Hard-coding `dm_mean` to have the same value as `Word2Vec.__init__(cbow_mean)` kind-of works, but means the same default is redundantly declared in two places, and any future changes would require two edits. (The best fix would keep it in a single place.)

markroxor:
Okay things are more easier then. My approach is to leave the params of `__init__` of `doc2vec.py` as it is and remove only the overriding params from [this method](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/doc2vec.py#L603). Such that all the common params are now inherited from `word2vec.py` now checks are failing with this new error
`RuntimeError: you must first build vocabulary before training the model`. 
Strange.
I guess there is something else wrong with code.

gensim/models/doc2vec.py
```diff
@@ -600,18 +600,20 @@ def __init__(self, documents=None, size=300, alpha=0.025, window=8, min_count=5,
           of the model.
 
         """
+
         super(Doc2Vec, self).__init__(
-            size=size, alpha=alpha, window=window, min_count=min_count, max_vocab_size=max_vocab_size,
-            sample=sample, seed=seed, workers=workers, min_alpha=min_alpha,
-            sg=(1+dm) % 2, hs=hs, negative=negative, cbow_mean=dm_mean,
+            size=size, window=window, min_count=min_count,
+            sample=sample, workers=workers,
```
gojomo: If `size`, `window`, `min_count`, `sample`, and `workers` all use the same defaults as Word2Vec, there's no need to re-specify them in this method signature nor pass to the call of superclass `__init__()`.

gensim/models/word2vec.py
```diff
@@ -459,6 +459,7 @@ def __init__(
         self.total_train_time = 0
         self.sorted_vocab = sorted_vocab
         self.batch_words = batch_words
+        self.size = size
```
gojomo: This new instance variable seems unnecessary; better to consult existing `layer1_size`.

markroxor: okay I will remove [this line](https://github.com/markroxor/gensim/blob/4b27d3687971ea45bc950f877c1bb0fb208e4ca4/gensim/models/doc2vec.py#L616) instead.

gensim/models/doc2vec.py
gojomo: The goal of issue #795 is to harmonize these defaults, unless there is a specific reason for the Doc2Vec defaults to differ. (Is there a reason to prefer `size=300', 'window=8', `sample=0', or 'workers=1` as being more-sensible defaults in Doc2Vec?) And something like `min_count=5` already matches Word2Vec – so no need to redundantly specify it again.

markroxor: No that is definitely not the case, in the previous commit that I submitted I removed these variables too but the Travis build seem to fail. Maybe we need to change the tests as well. What do you advice?

gojomo: It would depend specifically on what tests fail and why. (Looking at the failures at that time, it looks the problem was you retained a reference to `size` that was no longer valid: you'd taken our too little, not too much.) If at all possible, the tests shouldn't change: they help confirm that the refactoring didn't break anything else. But it might be the case that a test was dependent on the old defaults, and if that's the case, it could be appropriate to adjust the test as a last resort.

markroxor: For initialising the default params of doc2vec if I remove all the params passed on to super(Doc2Vec, self).init(....). In that case even if you pass some parameter to doc2vec. Suppose "workers", even in that case self.workers for both word2vec and doc2vec will be defaulted to 3(the default param of word2vec ). which implies that not passing the parameters at all to super(Doc2Vec, self).init(....) results in taking the default params of word2vec irrespective of whatever params you pass.
In the case when you pass all the params to super(Doc2Vec, self).init(size=size,alpha=alpha...) the default params of doc2vec will always be considered.
and defaults of word2vec will never be used.
I did some digging, but could not find a way (and possibly no way exist) to know whether the param is the default one or passed explicitly by the user.
I want to know your opinion about this.

gojomo: The use of `**kwargs` handles this. If you pass `workers=5` to Doc2Vec, if it's not explicitly declared in the `__init__()` signature, it stays in the `kwargs` dict. That's then passed to the superclass `__init__()` – overriding the default of 3. So the desired effect is achieved: shared defaults exist in one place (Word2Vec), but calls to either can override with an explicit valiue. 

(As noted in a prior comment, only `dm_mean` is especially tricky, needing a different approach, because it's a de facto synonym for `cbow_mean`. The goal is to have it so that if `dm_mean` is unspecified, the default Word2Vec `cbow_mean` value is operative, but if specified, it sets `cbow_mean` to the specified value. It's definitely also possible, and a solution might rely on the fact `kwargs` is a normal dict.)

markroxor:
@gojomo there is an issue with the travis build. My builds pass perfectly [here](https://travis-ci.org/markroxor/gensim/builds/169148331). Please have a look at #957 and travis issue [#6751](https://github.com/travis-ci/travis-ci/issues/6751)

gensim/test/test_wikicorpus.py
```diff
@@ -30,10 +30,11 @@ def setUp(self):
         wc = WikiCorpus(datapath(FILENAME))
 
     def test_get_texts_returns_generator_of_lists(self):
+
```
gojomo: No real reason to have these whitespace changes in this unrelated patch.

gensim/test/test_doc2vec.py
```diff
@@ -168,15 +171,15 @@ def model_sanity(self, model):
     def test_training(self):
         """Test doc2vec training."""
         corpus = DocsLeeCorpus()
-        model = doc2vec.Doc2Vec(size=100, min_count=2, iter=20)
+        model = doc2vec.Doc2Vec(size=100, min_count=2, iter=20, window=8, sample=0.01, workers=1)
```
gojomo: Working with a tiny corpus, `workers=1` is a good idea (and I believe what most of the Word2Vec tests do. Were the other two parameter changes also necessary for the test to pass? (If not, better to test the new defaults.)

gensim/test/test_doc2vec.py
```diff
@@ -168,15 +171,15 @@ def model_sanity(self, model):
...
         model.build_vocab(corpus)
         self.assertEqual(model.docvecs.doctag_syn0.shape, (300, 100))
         model.train(corpus)
 
         self.model_sanity(model)
 
         # build vocab and train in one step; must be the same as above
-        model2 = doc2vec.Doc2Vec(corpus, size=100, min_count=2, iter=20)
+        model2 = doc2vec.Doc2Vec(corpus, size=100, min_count=2, iter=20, window=8, sample=0.01, workers=1)
```
gojomo: As above, `workers=1` is good but the other parameters should only be added if definitely necessary.

gensim/test/test_wikicorpus.py
markroxor: It was suggested by Lev to leave a blank line to pass the Travis build. I am not sure why.

gensim/test/test_doc2vec.py
```diff
@@ -84,10 +84,11 @@ def test_int_doctags(self):
         """Test doc2vec doctag alternatives"""
         corpus = DocsLeeCorpus()
 
-        model = doc2vec.Doc2Vec(min_count=1)
+        size = 300
+        model = doc2vec.Doc2Vec(min_count=1, size=size)
```
gojomo: If these tests would succeed with the new (inherited) default size of 100, it'd be better to run them that way – and then test the internal array sizes for either matching '100' or (perhaps simply to verify internal self-consistency) 'model.vector_size'.

gensim/test/test_doc2vec.py
```diff
@@ -106,11 +107,13 @@ def test_string_doctags(self):
         # force duplicated tags
         corpus = corpus[0:10] + corpus
 
-        model = doc2vec.Doc2Vec(min_count=1)
+        size = 300
+        model = doc2vec.Doc2Vec(size=size, min_count=1)
```
gojomo: As above, though holding test params constant for at least one cycle was useful, if new-default value would also work as suitable test, best to have it be the value going forward.

gensim/test/test_doc2vec.py
markroxor: Seem to fail with the previous value.

gensim/test/test_doc2vec.py
markroxor: This too, seem to fail with the size = 100.

gensim/test/test_doc2vec.py
markroxor: Tests work fine without them so removed the other two params.

gensim/models/word2vec.py
```diff
@@ -460,6 +459,11 @@ def __init__(
         self.sorted_vocab = sorted_vocab
         self.batch_words = batch_words
 
+        if "dm_mean" in kwargs:
+            self.cbow_mean = int(kwargs["dm_mean"])
+        else:
+            self.cbow_mean = int(cbow_mean)
+
```
markroxor: okay

gensim/test/test_wikicorpus.py
tmylk: It should be removed. it was just a travis ping.

markroxor:
The build pass perfectly in my [forked repo's Travis](https://travis-ci.org/markroxor/gensim/builds/169185272).

gensim/test/test_doc2vec.py
gojomo: There's both a 300 (count of text examples, length of `syn0`) that needs to be tested, and a 100 (dimensionality of individual vectors) - you've combined them both into `size`. These tests will pass at default (100) dimensionality, as long as you're not asserting the example-count is also 100.

gojomo:
Yes, there's a test that's often hanging on Python 2.7 that's unrelated to your changes (TestWikiCorpus.test_first_element). I've made a separate issue for that (#971). (It may be that forcing re-tries will cause it to eventually succeed; it may be that something unique on the test machine is now causing it to reliably fail; in either case, when you see a failure there, it's not the fault of your changes.)

Regarding the handling of `dm_mean`, your latest changes still have the Word2Vec class interpreting a `dm_mean` parameter, and accepting incorrect/misspelled parameters without error. The best fix would keep all knowledge of `dm_mean` inside Doc2Vec, and still error for completely-bad parameters.

markroxor:
Removed `kwargs` argument and any trace of `dm_mean` from `word2vec`.

markroxor:
Ping @gojomo

tmylk:
Looks good to me. Waiting for @gojomo review.

piskvorky:
@tmylk:  @gojomo may be overloaded with github notifications (I know I am!), so probably best to ping him at Slack `#opensource` for anything urgent.

Same if my input is urgently needed anyway... though I'll munch through the notifications on my own, eventually.

gojomo:
Chnages all look good! The unrelated test timeout (wiki_corpus-related) means I'm not sure all Word2Vec/Doc2Vec tests have passed for the change, though.

markroxor:
Can you try restarting the build manually. The tests passed on my forked repo though.

tmylk:
@markroxor Please rebase to resolve merge conflicts.
Then ready to merge.

markroxor:
ping @tmylk

tmylk:
Thanks for the PR!

PR #944, opened 2016-10-12
Fix CellExecutionError in doc2vec-lee.ipynb. Solves issue #943

### LINKED ISSUE #943: Doc2Vec tutorial on the Lee Dataset does not execute correctly with python2.7
While testing the doc2vec-lee.ipynb notebook using python-2.7, the following line causes a problem:
`
with open(fname, encoding="iso-8859-1") as f:
`
This is because, in python 3, open() and io.open() are the same and allow specification of the encoding as a parameter. This is not the case in python 2.6/2.7. 

So, we need to explicitly specify io.open() for it to work correctly.

mon95:
Changed open() to explicitly specify io.open() to ensure correct execution of cells while using python 2.6/2.7. Fixes #943.

### DIFF
--- a/docs/notebooks/doc2vec-lee.ipynb
+++ b/docs/notebooks/doc2vec-lee.ipynb
@@ -18,7 +18,7 @@
     "import gensim\n",
     "import os\n",
     "import collections\n",
-    "import io\n",
+    "import smart_open\n",
     "import random"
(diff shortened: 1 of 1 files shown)


tmylk:
Thanks for the PR!

PR #948, opened 2016-10-13
Move log_normalize from hdpmodel to matutils and add feature normalization capability

ArifAhmed1995:
- Log normalizing may be used for other models too, hence it's better to shift it to matutils
- log normalize : Currently for an input 2D array , independently normalizing each sample is done.An optional axis parameter is introduced to give support for normalizing along feature.

See axis parameter in : 
http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.normalize.html


tmylk:
Please remove the unnecessary commits by running 'git fetch' and then 'git merge develop'

ArifAhmed1995:
@tmylk #950

Issue #936, opened 2016-10-12
The automated tests for notebooks from #874 do not run during CI 


mon95:
I'm working on the same.

bhargavvader:
Awesome, we certainly need this done.

dsquareindia:
However they should all run on "pseudo toy-datasets" just to ensure that the notebooks execute from start to finish. We don't want `wgets` inside the notebooks as some might take a long long time to execute on the actual datasets! `sklearn` does that pretty well and we can look into their tests.

bhargavvader:
+1 to what Devashish said. @mon95 , for the notebooks which fail, could you notify on this issue? We could accordingly modify the notebooks to accommodate this.

mon95:
@dsquareindia The prescribed method of testing ipython notebooks is to execute cell by cell and check for errors in the output. So, in order to use 'toy datasets', my guess is that we'll need to modify the tutorials. 

@bhargavvader I just ran into this: 

`Testing ./docs/notebooks/Similarity_Queries.ipynb
ERROR

======================================================================
ERROR: test_notebooks (gensim.test.test_notebooks.TestNotebooks)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/media/ssreehari10/gdrive/gensim/gensim/test/test_notebooks.py", line 66, in test_notebooks
    nb, errors = _notebook_run(notebook)
  File "/media/ssreehari10/gdrive/gensim/gensim/test/test_notebooks.py", line 43, in _notebook_run
    raise e
CellExecutionError: An error occurred while executing the following cell:
------------------
from gensim import corpora, models, similarities

dictionary = corpora.Dictionary.load('/tmp/deerwester.dict')
corpus = corpora.MmCorpus('/tmp/deerwester.mm') # comes from the first tutorial, "From strings to vectors"
print(corpus)
------------------

IOError: [Errno 2] No such file or directory: '/tmp/deerwester.dict'
`
This is because the line throwing the error depends on cell number 5 here https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/Corpora_and_Vector_Spaces.ipynb 

This needs to be changed

tmylk:
Let's unite them into one ipynb then as there is a dependency.

tmylk:
If the dataset is not available then the ipynb is skipped. See #874 for 
`raise ValueError(\"SKIP: Please download the Corpus/news_dictionary dataset.\")`

mon95:
[...]

PR #953, opened 2016-10-16
Add tests for ipython notebooks to run during Travis CI

mon95:
Added tests for ipython notebooks to run during CI and check each cell of the notebooks for correct execution. Errors due to timeout resulting from wget commands are handled. If the dataset is not available then the ipynb is skipped. 
Solves https://github.com/RaRe-Technologies/gensim/issues/936


tmylk:
The `wget` or other timeouts should not be silent. Add raising of an explicit 'SKIP' to the ipynb if it happens.

tmylk:
Plus there are CI failures in Travis.

gensim/test/test_notebooks.py
```diff
@@ -0,0 +1,82 @@
...
+def _notebook_run(path):
+    """
+    Execute a notebook via nbconvert and collect output.
+       :returns (parsed nb object, execution errors)
+    """
+    kernel_name = 'python%d' % sys.version_info[0]
+    this_file_directory = os.path.dirname(__file__)
+    errors = []
+
+    fl = tempfile.NamedTemporaryFile(suffix=".ipynb", mode='wt', delete=False)
```
mon95: There is no need for it while testing. Will remove it

gensim/test/test_notebooks.py
```diff
@@ -0,0 +1,82 @@
...
+
+import logging
+import unittest
+import os
+import os.path
+import tempfile
+import numbers
+import codecs   
+
+class TimeoutError(Exception):
```
mon95: Removing it gives me this error on timeout: 
```NameError: global name 'TimeoutError' is not defined```

Is this a python2 vs python3 issue? I understand that TimeoutError is a subclass of OSError but I'm not sure how to use it in code.

**Comment by mon95** [comment]
Will modify the ipython notebooks to include something like this:
```raise ValueError(\"SKIP: please download the questions-word.txt```

Would this be alright? This would get detected as a CellExecutionError.

File: gensim/test/test_notebooks.py (hunk 2 above)
Comment (tmylk) [inline gensim/test/test_notebooks.py]: It is a Python 3 error. Why does it need special treatment?

**Comment by tmylk** [comment]
Adding SKIP instead of download is ok. It would be best to make sure it only raises it during testing.

**Comment by mon95** [comment]
Sure, I'll make the necessary changes to the ipython notebooks.

File: gensim/test/test_notebooks.py (hunk 2 above)
Comment (tmylk) [inline gensim/test/test_notebooks.py]: why should we catch any of them at all?

Comment (mon95) [inline gensim/test/test_notebooks.py]: Changed this. Thank you!

**Comment by gojomo** [comment]
Won't this change the CI tests from taking about 5 minutes to run, to many hours? 

Also, notebooks sometimes include code that's intended to error (as illustration) or examples of variants (not intended to be always run)... so this automatically-run-everything approach seems constraining to me.

File: docs/notebooks/Similarity_Queries.ipynb [hunk 3]
```diff
@@ -60,8 +60,11 @@
    "source": [
     "from gensim import corpora, models, similarities\n",
     "\n",
-    "dictionary = corpora.Dictionary.load('/tmp/deerwester.dict')\n",
-    "corpus = corpora.MmCorpus('/tmp/deerwester.mm') # comes from the first tutorial, \"From strings to vectors\"\n",
+    "try:\n",
+    "    dictionary = corpora.Dictionary.load('/tmp/deerwester.dict')\n",
+    "    corpus = corpora.MmCorpus('/tmp/deerwester.mm') # comes from the first tutorial, \"From strings to vectors\"\n",
+    "except:\n",
+    "    raise ValueError(\"SKIP: Dependency - Corpora_and_Vector_Spaces.ipynb file\")\n",
```
tmylk: actually these 2 notebooks should be united into one as they are dependent.

docs/notebooks/online_w2v_tutorial.ipynb
```diff
@@ -78,7 +79,10 @@
    },
    "outputs": [],
    "source": [
-    "old, new = [WikiCorpus('enwiki-{}-pages-articles.xml.bz2'.format(ymd)) for ymd in ['20101011', '20160820']]"
+    "try:\n",
+    "    old, new = [WikiCorpus('enwiki-{}-pages-articles.xml.bz2'.format(ymd)) for ymd in ['20101011', '20160820']]\n",
+    "except:\n",
+    "    raise ValueError(\"SKIP: please download the wiki dump files.\")\n"
```
tmylk: Please download the wiki dump files by uncommenting the wget commands above.

docs/notebooks/Similarity_Queries.ipynb
mon95: Moved the Similarity queries tutorial to the Corpora_and_Vector_Spaces.ipynb file. 
Will change links within tutorials.md if this is alright.

mon95:
@gojomo The tests shouldn't take too long to execute as we're setting each cell's execution time limit to be about 10 seconds(this can be changed).

gojomo:
@mon95 So when a cell times out – or the next cell depends on the results of timed-out cell – the notebook registers as a test failure? (While I haven't looked at all the notebooks, the ones I'm familiar with tend to have multiple cells that take more than 10 seconds to execute.)

mon95:
As of now, some of the ipython notebooks have been modified to include something like this:
```raise ValueError(\"SKIP: please download the questions-word.txt``` 

And these CellExecutionErrors do not lead to a test failure.

gojomo:
@mon95 That seems different than timeout failures, and I don't see similar changes in notebooks like doc2vec-IMDB.ipynb, where many cells need tens-of-minutes to complete. Is the intent that this notebooks-as-tests only runs on some subset of `docs/notebooks`, those that are intended to run straight-through quickly?

mon95:
Haven't run into errors or timeouts on those notebooks yet because other errors seem to crop up while testing. Will make necessary changes to the notebooks if such timeouts are encountered. 

Notebooks that require large datasets that are skipped. The idea is to get the tests to run in CI. https://github.com/RaRe-Technologies/gensim/issues/936

tmylk:
@mon95 do the tests run on your local machine?

mon95:
Yes! The tests run for both python2.7 and python3.5

tmylk:
Please add nbconvert to [test dependencies](https://github.com/RaRe-Technologies/gensim/blob/develop/setup.py#L183). It will probably solve the import problems

mon95:
Added nbconvert to the test dependencies.

tmylk:
@mon95 Tests fail because nbconvert can't be installed in travis for some reason. Can you see Travis logs?

mon95:
Yes! I'm able to see the logs now. I'll try the solution that's mentioned in the last comment here [https://github.com/jupyter/jupyter/issues/150](url)

tmylk:
@olavurmortensen Your new LDA tutorial doesn't pass the CI on notebooks - please add data if it is small, or add a SKIP option.

tmylk:
The goal of this PR is to make CI tests pass in at least one version of Python and skip on others. It will already be a great improvement.
Next stage will be to make them pass in other Python versions.

tmylk:
Need to add pyldavis to [test dependencies](https://github.com/RaRe-Technologies/gensim/blob/9a3329cfccfcf98e9b88f5f52437f1049f4e3c44/setup.py#L183)

mon95:
@tmylk Could you please take a look at this? 
The tests for the notebooks do seem to run fine and they fail elsewhere (TestLdaMulticore).

markroxor:
@mon95 please resolve the conflicts.
@tmylk Are you available to review this? or should I?

tmylk:
Looking for volunteer to resolve the conflicts, then should be good to merge after a quick review.

tmylk:
Suggest using nbdime package https://github.com/jupyter/nbdime

PR #973, opened 2016-10-21
[WIP] Add pep8 to Travis 

### LINKED ISSUE #965: Add automatic PEP8 checking to Travis
A good example of this is sklearn

souravsingh:
Fixes #965

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,5 +24,5 @@ install:
 
 script:
   - python setup.py test
-  - coala-ci
+  - coala --non-interactive
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,5 @@ install:
 
 script:
   - python setup.py test
+  - coala-ci
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_install:
   - conda update --yes conda
   - pip install --upgrade pip
   - pip3 install coala
+  - pip3 install coala-bears
 install:
   - conda create --yes -n gensim-test python=$TRAVIS_PYTHON_VERSION pip atlas numpy scipy
   - source activate gensim-test
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,5 +24,5 @@ install:
 
 script:
   - python setup.py test
-  - find . -name \*.py -exec pep8 --ignore=E402 {} + | wc -l
+  - find . -name \*.py -exec pep8 --ignore=E501 {} + | wc -l
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
   - export PATH=/home/travis/miniconda2/bin:$PATH
   - conda update --yes conda
   - pip install --upgrade pip
-  - pip install pep8
+  - pip3 install coala
 install:
   - conda create --yes -n gensim-test python=$TRAVIS_PYTHON_VERSION pip atlas numpy scipy
   - source activate gensim-test
@@ -24,5 +24,4 @@ install:
 
 script:
   - python setup.py test
-  - pep8 --ignore=E501,E241,E701 .
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,11 @@ before_install:
   - ./miniconda.sh -b
   - export PATH=/home/travis/miniconda2/bin:$PATH
   - conda update --yes conda
+  - sudo apt-get update
+  - sudo apt-get -y install python-pip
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -12,11 +12,17 @@ before_install:
...
   - source activate gensim-test
   - pip install pyemd
   - pip install annoy
   - pip install testfixtures
   - python setup.py install
-script: python setup.py test
+
+script:
+  - python setup.py test
+  - find . -name \*.py -exec pep8 --ignore=E402 {} + | wc -l
```
markroxor: change ignore to E501 for suppressing "long line" errors.

tmylk:
There are definitely pep8 violations so I expected pep8 check to fail. How come it passes?

.travis.yml
```diff
@@ -12,11 +12,17 @@ before_install:
...
   - source activate gensim-test
   - pip install pyemd
   - pip install annoy
   - pip install testfixtures
   - python setup.py install
-script: python setup.py test
+
+script:
+  - python setup.py test
+  - find . -name \*.py -exec pep8 --ignore=E501 {} + | wc -l
```
souravsingh: Alright!! done!

souravsingh:
Coala installation is creating a problem for Python 2.x systems.

tmylk:
It is ok that coala is only in Python 3. It is sufficient to have pep8 checks in Python 3 only.

souravsingh:
I have added the coala config file. Is there any directories that needs excemption from PEP8 check?( I am thinking of adding /models to ignore list)

tmylk:
the pep8bear is not running according to the travis log. Please add it to run in Travis CI

souravsingh:
coala bears is now up and running

souravsingh:
@tmylk We can configure the coala to run the Bears by making modifications to the config file accordingly.

nishnik:
@tmylk 
I have cleaned most of the code.
You can see it [here](https://github.com/nishnik/gensim/tree/pep).
I have used a different method, though I would like to mention the pros of the method that I have incorporated (Check [.travis.yml](https://github.com/nishnik/gensim/blob/pep/.travis.yml)):
- Travis will run for a smaller time. The script checks all the errors only for Python 2.7 (Though we should change it to python 3, I was currently on a system where python 3 was not supported. Apologies)
- There wont be any coala ci. I dont know whether this is desirable or not.
- I personally think that using pep8 as it is will help the user spot the errors easily on a local machine.

souravsingh:
@nishnik First thanks for the contribution. I will give my two cents on thw issue
1) Coala CI is quite flexible in the sense that you can configure the bears which you need( GitCommitBear, PyCodeStyleBear and so on..) by just a simple commit to the coala config file.

2) PEP8Bear is essentially PEP8 optimized to run with coala. There is no change in the list of errors, arguments or anything. Only the name of the tools has changed.

3) Coala requires Python 3 to run nicely, so an extra line for check for Python 3 won't be necessary.

nishnik:
@souravsingh 
I completely agree with you on the first two points.
1. **Coala** is very flexible, I had attended a workshop where it was demonstrated.
2. That was a personal statement I made for pep8, I don't like extra dependencies when you can get your work done in a decent manner with the available resources (though I will prefer Coala for other languages (C++, C) but why use a wrapper over pep8 when you can get your work done with pep8) 

Sorry that I was not able to get my intentions clearly to you regarding your 3rd point. I have contributed to [SymEngine](https://github.com/symengine/symengine) where while I was working on lint checking, it was made sure that we should check for lint errors only once in travis instances, because it is redundant to check for lint errors in every instance (this increases the wait time for getting the ci results).

tmylk:
@nishnik Agree that just `pep8` is better than Coala as it is easier to use. Also that running linter once per CI is good.

Also thanks a lot for the cleaned code! May I use it as a PR?

May I also ask about the choice of the [errors](http://pep8.readthedocs.io/en/release-1.7.x/intro.html#configuration) to ignore: E501,E241,W,E731,E402,E265. Are they based on some specific experience in Gensim code?

E501 for line length is clear. Curious about others.

nishnik:
@tmylk 
Sure! I will send a PR after a few desired changes.
I had ignored warnings(`W`) and other errors because they couldn't be fixed by `autopep8`. As of now, I have manually removed (`E241, W, E265, E402`) will look into `E731` and update you.

souravsingh:
@tmylk Closing the PR since the work is finished.

tmylk:
@souravsingh Finished by whom?

Issue #979, opened 2016-10-26
WMD Earth mover Distance for sentences that contain a single word return inf


rbahumi:
`import gensim
from reuse import load_model

if **name** == "**main**":
    print "gensim version: %s" % gensim.**version**

```
path = "models_no_short_text_20.10"
model = load_model(path)

sentence_obama = 'Obama speaks to the media in Illinois'.lower().split()
sentence_president = 'The president greets the press in Chicago'.lower().split()
print model.wmdistance(sentence_president, sentence_obama)

obama = "Obama".lower().split()
print 'Result for sentence "%s": %s' % (" ".join(obama), model.wmdistance(obama, obama))


obama = "Obama".lower().split() * 5
print 'Result for sentence "%s": %s' % (" ".join(obama), model.wmdistance(obama, obama))`
```

Will   ##output:

> gensim version: 0.13.2
> 3.69295766061
> Result for sentence "obama": inf
> Result for sentence "obama obama obama obama obama": inf

rbahumi:
This is caused by the np_sum(distance_matrix) is equal to 0 when the distance matrix is a 1*1 matrix

[https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec.py#L1415](url)

tmylk:
Thanks for highlighting this. What would be a better value to return?

rbahumi:
@tmylk, thanks for your replay. I suggest it should return float 0.0. If you agree, I have attached a pull request that deals with it. [Please take a look.](https://github.com/RaRe-Technologies/gensim/pull/986)

tmylk:
How d

PR #986, opened 2016-10-30
Fix for issue #979

rbahumi:
Issue: WMD Earth mover Distance for sentences that contain a single word return inf
Fix: Return float 0.0 when the dictionary size is 1 (both sentences are composed with the same token)


rbahumi:
This is the relevant open issue:
https://github.com/RaRe-Technologies/gensim/issues/979

tmylk:
Thanks for the PR!

PR #895, opened 2016-09-27
Fixed issue #838

markroxor:
Fixed RuntimeWarning: overflow encountered in exp.

PR #989, opened 2016-11-01
optimizing sigmoid function

markroxor:
Optimising sigmoid function created in PR #895.


gensim/models/word2vec.py
```diff
@@ -300,7 +295,7 @@ def train_cbow_pair(model, word, input_word_indices, l1, alpha, learn_vectors=Tr
             if w != word.index:
                 word_indices.append(w)
         l2b = model.syn1neg[word_indices]  # 2d matrix, k+1 x layer1_size
-        fb = sigmoid(dot(l1, l2b.T))  # propagate hidden -> output
+        fb = expit(dot(l1, l2b.T))  # propagate hidden -> output
```
piskvorky: Is this the only place where inverse logit is used? Probably worth updating all occurrence, consistently.

Issue #838, opened 2016-08-25
word2vec overflow when running


jackyjwwang:
/gensim-0.13.1/gensim/models/word2vec.py:296: RuntimeWarning: overflow encountered in exp
  fb = 1. / (1. + exp(-dot(l1, l2b.T)))  # propagate hidden -> output

how can fix it, thank you

tmylk:
Have you tried using the compiled C extension? It is much faster than Python.

Could you please give more instructions on how to reproduce? A sample corpus would help.

jackyjwwang:
I'm sorry, it's long time to check the mail.

I can't use the C version, Because it depends the C complier lib\


best regards

thanks

2016-08-29 16:31 GMT+08:00 Lev Konstantinovskiy <notifications@github.com>:

gojomo:
The necessary C compilation tools should be freely available on all platforms where gensim runs – and they can offer a 100X or more speedup. For example, they can turn runtimes that would be over 4 days (without) into under-an-hour. So, we highly recommend ensuring the optimized code is running.

Still, there shouldn't be any fatal warnings in the pure-python code. Does the warning actually cause execution to stop, or does your training still complete and result in usable word-vectors?

(I think the `MAX_EXP` conditional check and early-loop-continue in the cython code – eg https://github.com/RaRe-Technologies/gensim/blob/9a02527ab315d00dae30088855d2ca466cc3e436/gensim/models/word2vec_inner.pyx#L83 and elsewhere  – may be what prevents a similar overflow message there.)

markroxor:
Hello @gojomo ,
I think limiting the value of 
`fb = 1. / (1. + exp(-dot(l1, l2b.T))) # propagate hidden -> output` 
[here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec.py#L296)
will fix the issue, I am willing to send a PR but am quite not sure about the values to which I should limit it to. Any ideas?

tmylk:
If the values aren't too big then this [solution](http://stackoverflow.com/questions/9478663/logit-and-inverse-logit-functions-for-extreme-values) will help:
```python
def inv_logit(p):
    if p > 0:
        return 1. / (1. + np.exp(-p))
    elif p <= 0:
        np.exp(p) / (1 + np.exp(p))
    else:
        raise ValueError
```

markroxor:
That should do the trick as well. Sending a PR, merge if it helps.

[...]

PR #990, opened 2016-11-01
Optimising sigmoid function everywhere

markroxor:


gensim/models/word2vec.py
```diff
@@ -320,14 +315,14 @@ def score_sg_pair(model, word, word2):
...
-    lprob = -log(1.0 + exp(-sgn*dot(l1, l2a.T)))
+    lprob = log(expit(sgn * dot(l1, l2a.T)))
     return sum(lprob)
 
 
 def score_cbow_pair(model, word, word2_indices, l1):
     l2a = model.syn1[word.point]  # 2d matrix, codelen x layer1_size
     sgn = (-1.0)**word.code  # ch function, 0-> 1, 1 -> -1
-    lprob = -log(1.0 + exp(-sgn*dot(l1, l2a.T)))
+    lprob = log(expit(sgn * dot(l1, l2a.T)))
```
piskvorky: Not directly related to this PR, but since we're on the topic, there's a function in numpy for safely adding up logs of exps: `numpy.logaddexp(0, sgn * dot(l1, l2a.T))`.

piskvorky:
These are all welcome @markroxor , but I wonder why such overflows/underflows happen in the first place.

That's probably the root cause of the trouble here, and making `exp(1000)` or `exp(-1000)` "work" is probably just masking the real underlying problem.

markroxor:
I've never encountered a case when such a warning might occur but it was reported in issue #838 (it depends on the corpus I suppose). AFAIK we are just handling the huge exponents in a better way so it should not cause any problem.

gojomo:
Using a scipy function seems like a good idea. However, note these pure-Python paths are *not* tested by our automated testing (because the Cythonized routines are available at that point), so such changes need careful testing by other more manual-means. @markroxor, have you run the Word2Vec/Doc2Vec test suites locally with the optimized-code disabled?

piskvorky:
> AFAIK we are just handling the huge exponents in a better way so it should not cause any problem.

Well, that's the part I doubt. The fact that we now return 0.0 or inf or -inf correctly, without an overflow/underflow warning, doesn't really qualify as "no problem" IMO. It's possible it just pushes the problem further upstream.

markroxor:
@gojomo "with the optimized-code disabled". Do you mean that I should test word2vec/doc2vec without incorporating these changes?

gojomo:
@markroxor - the change you've made is in code that's only run if the Cython-optimized versions (word2vec_inner.pyx, doc2vec_inner.pyx) are not available. In a normal installation, or in the Travis-CI tests, those versions *are* available. So the changed code won't be run unless you do something to disable the optimized code - like changing the load-attempt at the top of doc2vec.py or word2vec.py. 

At times in the past to test the pure-Python paths, I've commented-out that load, then run a full test suite locally. It might be nice if the classes allowed a user-switch without editing code. But, without the optimization tests that normally take seconds take many minutes, so as a practical matter the pure-Python paths may never be part of the automated test suite.

markroxor:
@gojomo I executed the tests locally and they passed perfectly. Indeed the execution was very slow.

PR #993, opened 2016-11-05
The end index in list slicing is exclusive

Renata1995:


piskvorky:
Can you provide context in your PR description? What fails, how, why this change?

According to the comments in the code, your suggested change is incorrect (loop iterates over `m` numbers instead of `m-1` numbers).

Renata1995:
The second index in the slicing syntax is exclusive.
i.e. list = [1,2,3,4]   list[:2] = [1,2]
Therefore, to iterate from index 0 to m-1, list[:m] should be used instead of list[:m-1].

Suppose there is a topic which contains 4 words {a,b,c,d}
The old code will count {ca, da, db}, but forget {ba, cb, dc}.

piskvorky:
No, the old code is correct. It lists items 1...m-1 (not 0...m-1).

Unless there is a clear bug report or explanation associated with this, I suggest closing.

dsquareindia:
@piskvorky I think @Renata1995 is right. In `u_mass` coherence `s_one_pre` segmentation is used as explained [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/topic_coherence/segmentation.py#L17). I just wrote a small function to validate both algos:
```python
import numpy as np

def test_seg(topics):
    seg_1 = []
    seg_2 = []
    for t, top_words in enumerate(topics):
        seg_t1 = []
        seg_t2 = []
        for m in top_words[1:]:
            m_index = np.where(top_words == m)[0]
            for l in top_words[:m_index]:
                seg_t1.append((m, l))

            for n in top_words[:m_index - 1]:
                seg_t2.append((m, n))

        seg_1.append(seg_t1)
        seg_2.append(seg_t2)
    return seg_1, seg_2


topics = [np.array([1, 2, 3]), np.array([4, 5, 6])]
seg_1, seg_2 = test_seg(topics)
print seg_1
print seg_2
```
Here's the output
```
test.py:11: VisibleDeprecationWarning: converting an array with ndim > 0 to an index will result in an error in the future
  for l in top_words[:m_index]:
test.py:14: VisibleDeprecationWarning: converting an array with ndim > 0 to an index will result in an error in the future
  for n in top_words[:m_index - 1]:
[[(2, 1), (3, 1), (3, 2)], [(5, 4), (6, 4), (6, 5)]]
[[(3, 1)], [(6, 4)]]
```
The first output which corresponds to @Renata1995's PR is correct.

tmylk:
@dsquareindia @Renata1995  Should we replace the code from the `top_topics` with a call to the new `topic_coherence` code? It is easier to maintain the code in one place.

piskvorky:
OK, then the original comments in the code are incorrect too.

@dsquareindia since you understand what's going on, can you fix the ticket name & add context and description to the PR? In particular, why is one of the outputs wrong and the other right. Cheers.

dsquareindia:
@piskvorky a more accurate ticket name would be

"**`s_one_pre` segmentation in existing `u_mass` code in `ldamodel.py`  is incorrect**"

_Description_: `u_mass` segmentation uses `s_one_pre` segmentation. This segmentation is described [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/topic_coherence/segmentation.py#L16). For the current code the segmentation is incorrect because:

if `topics = [np.array([1, 2, 3]), np.array([4, 5, 6])]` then the correct output should be `[[(2, 1), (3, 1), (3, 2)], [(5, 4), (6, 4), (6, 5)]]` however currently the output is `[[(3, 1)], [(6, 4)]]`. This is because the end index slicing in the current code is incorrect.

@tmylk I am for your suggestion that we should replace the current code with a call to the `u_mass` code in `topic_coherence`. Would be easier to manage later on.

tmylk:
Ping @dsquareindia @Renata1995 Can this PR be changed to use `u_mass` in `topic_coherence`?

tmylk:
Opened #1128 to remove duplicate code.

PR #997, opened 2016-11-08
Remove hanging test_get_texts_returns_generator_of_lists from TestWikiCorpus

tmylk:


gensim/test/test_ldamallet_wrapper.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-#
+#h
```
piskvorky: Typo?

PR #1000, opened 2016-11-09
numpy as np

bhargavvader:
Wanted to make import abbreviations standard across the package, so changed all imports from `import numpy` to `import numpy as np`, and made all `numpy` as `np`.
Just a cosmetic change, but it is the abbreviation used in almost all other scientific computing packages.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -32,7 +32,7 @@ The algorithm:
 
 
 import logging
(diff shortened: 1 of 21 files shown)


piskvorky:
Congratulations on PR #1000!! 🎉 🎉

bhargavvader:
:dancing_men:

bhargavvader:
@tmylk , I'm done with this. If it's ok feel free to merge.

Issue #1009, opened 2016-11-13
Move "Pattern not installed" warning from import time to when lemmatization is used


tmylk:
It confuses many first time users as the warning is the first thing they see when importing gensim.

souravsingh:
@tmylk I would like to work on the issue. How do I start?

shashankg7:
@tmylk , @souravsingh . I have fixed this issue. I moved 'has_pattern()' method's pattern installation check to lemmatize method. The warning will only occur when user calls the lemmatization method and if the pattern library is not installed. I also modified the call to "has_pattern" in corpora/wikicorpus's constructor.

Please refer to #1018

PR #1018, opened 2016-11-14
Fix for issue #1009 

shashankg7:
This PR is to address the issue #1009. Removed has_pattern method from utils and added pattern library's installation check to lemmatize method. Warning will only appear if the user is using the lemmatize method and pattern library is not installed. 

Also removed has_pattern method's call from corpora/wikicorpus's constructor.

### DIFF
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -258,7 +258,7 @@ class WikiCorpus(TextCorpus):
     >>> MmCorpus.serialize('wiki_en_vocab200k.mm', wiki) # another 8h, creates a file in MatrixMarket format plus file with id->word
 
     """
-    def __init__(self, fname, processes=None, lemmatize=False, dictionary=None, filter_namespaces=('0',)):
+    def __init__(self, fname, processes=None, lemmatize=True, dictionary=None, filter_namespaces=('0',)):
         """
         Initialize the corpus. Unless a dictionary is provided, this scans the
         corpus once, to determine its vocabulary.
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -258,7 +258,7 @@ class WikiCorpus(TextCorpus):
     >>> MmCorpus.serialize('wiki_en_vocab200k.mm', wiki) # another 8h, creates a file in MatrixMarket format plus file with id->word
 
     """
-    def __init__(self, fname, processes=None, lemmatize=True, dictionary=None, filter_namespaces=('0',)):
+    def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), dictionary=None, filter_namespaces=('0',)):
         """
         Initialize the corpus. Unless a dictionary is provided, this scans the
         corpus once, to determine its vocabulary.
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -258,7 +258,7 @@ class WikiCorpus(TextCorpus):
     >>> MmCorpus.serialize('wiki_en_vocab200k.mm', wiki) # another 8h, creates a file in MatrixMarket format plus file with id->word
 
     """
-    def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), dictionary=None, filter_namespaces=('0',)):
+    def __init__(self, fname, processes=None, lemmatize=False, dictionary=None, filter_namespaces=('0',)):
         """
         Initialize the corpus. Unless a dictionary is provided, this scans the
         corpus once, to determine its vocabulary.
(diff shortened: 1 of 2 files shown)


gensim/corpora/wikicorpus.py
```diff
@@ -258,7 +258,7 @@ class WikiCorpus(TextCorpus):
     >>> MmCorpus.serialize('wiki_en_vocab200k.mm', wiki) # another 8h, creates a file in MatrixMarket format plus file with id->word
 
     """
-    def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), dictionary=None, filter_namespaces=('0',)):
+    def __init__(self, fname, processes=None, lemmatize=False, dictionary=None, filter_namespaces=('0',)):
```
tmylk: Is there a reason that default can't be True?

gensim/corpora/wikicorpus.py
```diff
@@ -274,7 +274,16 @@ def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), diction
         if processes is None:
             processes = max(1, multiprocessing.cpu_count() - 1)
         self.processes = processes
-        self.lemmatize = lemmatize
+        # if the user has set lemmatize flag as True, check if installed
+        if lemmatize == True:
```
tmylk:
```
if lemmatize
```
is enough in python

gensim/corpora/wikicorpus.py
```diff
@@ -274,7 +274,16 @@ def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), diction
...
+            try:
```
tmylk: is there a reason for this code to be duplicated here and in `lemmatize`?
Just leaving it in `lemmatize` is enough

shashankg7: Sure, that's correct. So the lemmatize in the constructor should be set to default value True then, and pattern related checks will be handled in lemmatize method, is that right?

tmylk: Yep

shashankg7: Removed the duplicate pattern library check code.

gensim/utils.py
```diff
@@ -1037,10 +1023,10 @@ def lemmatize(content, allowed_tags=re.compile('(NN|VB|JJ|RB)'), light=False,
...
 
     """
-    if not has_pattern():
-        raise ImportError("Pattern library is not installed. Pattern library is needed in order  \
-         to use lemmatize function")
-    from pattern.en import parse
+    try:
+        from pattern.en import parse
+    except ImportError:
+        warnings.warn("Pattern library is not installed, lemmatization won't be available.")
```
tmylk: does it cause an unhandled exception later on if `parse` not imported?

shashankg7: yeah, it will create an unhandled expection later in the method where parse is used, since only a warning is created. I'll change that.

gensim/utils.py
```diff
@@ -1005,15 +1004,13 @@ def pyro_daemon(name, obj, random_suffix=False, ip=None, port=None, ns_conf={}):
...
-    pattern = False
     try:
-        from pattern.en import parse
-        pattern = True
-    except ImportError:
-        warnings.warn("Pattern library is not installed, lemmatization won't be available.")
-    return pattern
+        import pattern
+        return True
+    except:
```
tmylk: just catch `ImportError`

shashankg7: This method is just for testing if pattern is present or not, later to be called in wikicorpus method. Will adding ImportError make any difference? 

Sorry I am not very good with exception handling. We hardly use them in our academic setting.

gensim/utils.py
```diff
@@ -1037,10 +1034,10 @@ def lemmatize(content, allowed_tags=re.compile('(NN|VB|JJ|RB)'), light=False,
     ['rank/NN', 'study/VB', 'hard/RB']
 
     """
-    if not has_pattern():
-        raise ImportError("Pattern library is not installed. Pattern library is needed in order  \
-         to use lemmatize function")
-    from pattern.en import parse
+    try:
```
tmylk: Please only check import in one place

shashankg7: Done. Please review.

gensim/utils.py
```diff
@@ -1005,15 +1004,13 @@ def pyro_daemon(name, obj, random_suffix=False, ip=None, port=None, ns_conf={}):
...
 def has_pattern():
     """
-    Function to check if there is installed pattern library
+    Function which returns a flag indicating whether pattern is installed or not
     """
-    pattern = False
     try:
-        from pattern.en import parse
-        pattern = True
+        import pattern
```
shashankg7: Sure, will change it. Are there any changes required in the exception handling part?

tmylk: Nope

shashankg7: I have fixed it. During testing there was some bug in Travis's python2.7 and python3.3 initialization. Can you please check that?

tmylk:
Thanks for the PR!

PR #1024, opened 2016-11-18
LDA hyperparameter fix: eta dimensionality and optimization

olavurmortensen:
Working on a problem with the parameter `eta` and automatic learning of `eta`. For now, the unchanged version is in the `ldamodelold.py` file, so that I can compare both versions. There is a notebook in `docs/notebooks/lda_tests.ipynb` with the tests I have been doing.

The changes I've made improve convergence (increase in bound over iterations), when learning `eta`.

## The problem

Basically, `eta` is supposed to be either
* a scalar
* a V vector

where V is the size of the vocabulary. In `ldamodel`, we instead have the options that `eta` is
* a scalar
* a K vector
* a K x V matrix

where K is the number of topics. As we will see in a little bit, this causes some problems.

## Results so far

The figure below shows the bound (`perwordbound`) as a function of the number of passes over data. It shows the bound both before and after the changes I made, and for `eta='symmetric'` and for `eta='auto'`.

The figure above shows that using automatically learned `eta` actually decreased the bound before this fix.

I also noticed that if we set `iterations=1`, and `eta='auto'`, the algorithm diverges. This is no longer a problem with this commit.

## Asymmetric priors

I added a check that raises a `ValueError` if `eta` is equal to 'asymmetric'. While a explicit asymmetric prior supplied by the user makes sense (e.g. if the user wants to boost certain words, or has estimated the prior from data), there is no reason for using an arbitrarily initialized asymmetric prior. Furthermore, with the current method of initializing asymmetric priors, and when eta is the same size as the vocabulary, I noticed some convergence problems.

As a side note on the topic of asymmetric priors, I'm not sure the current method is ideal. One could, for example, initialize from a gamma distribution the same way lambda and gamma are initialized instead.


olavurmortensen:
There is one problem that remains. The symmetric priors are initialized as
```init_prior = np.asarray([1.0 / self.num_topics for i in xrange(prior_shape)])```

instead of 
```init_prior = np.asarray([1.0 / prior_shape for i in xrange(prior_shape)])```

Because if `1.0 / prior_shape` is used, the problem does not converge. This happens because when the prior is a small number (e.g. eta = 0.001), the problem diverges. This does not happen if e.g. alpha=0.001, however.

While this works fine, I'm not sure if this way of initializing the priors is sound in general.

**Comment by olavurmortensen** [comment]
PR ready for review.

@tmylk @piskvorky

**Comment by orezpraw** [comment]
I agree that this should be changed because when I wrote the original auto_eta code, I didn't realize that gensim was using the smoothed LDA model, and thus I misinterpreted the variables and my code doesn't make sense in the context of the smoothed LDA model.

**Comment by olavurmortensen** [comment]
The failing unit tests are not from my code. There is more on that on issue https://github.com/RaRe-Technologies/gensim/issues/971.

File: gensim/models/ldamodel.py [hunk 1]
```diff
@@ -84,7 +84,7 @@ def update_dir_prior(prior, N, logphat, rho):
 
     dprior = -(gradf - b) / q
 
-    if all(rho * dprior + prior > 0):
+    if (rho * dprior + prior > 0).all():
```
olavurmortensen: The old way has never worked for me. Whenever I install Gensim from source I have to change this line. Can we just keep it this way?

docs/notebooks/lda_tests.ipynb
```diff
@@ -0,0 +1,762 @@
+{
```
olavurmortensen: Once we are ready to merge, I'll remove this file.

gensim/models/ldamodelold.py
```diff
@@ -0,0 +1,1049 @@
+#!/usr/bin/env python
```
olavurmortensen: Once we are ready to merge, I'll remove this file.

gensim/models/ldamodel.py
```diff
@@ -301,17 +303,19 @@ def __init__(self, corpus=None, num_topics=100, id2word=None,
         self.minimum_phi_value = minimum_phi_value
         self.per_word_topics = per_word_topics
 
+        self.random_state = get_random_state(random_state)
```
olavurmortensen: Oh, there was a reason, but it's not relevant any more. I'll just move it back to its old spot.

gensim/models/ldamodel.py
```diff
@@ -301,17 +303,19 @@ def __init__(self, corpus=None, num_topics=100, id2word=None,
...
+            if eta == 'asymmetric':
+                raise ValueError("The 'asymmetric' option cannot be used for eta")
 
-        self.random_state = get_random_state(random_state)
+        self.eta, self.optimize_eta = self.init_dir_prior(eta, 'eta')
 
-        assert (self.eta.shape == (self.num_topics, 1) or self.eta.shape == (self.num_topics, self.num_terms)), (
-            "Invalid eta shape. Got shape %s, but expected (%d, 1) or (%d, %d)" %
-            (str(self.eta.shape), self.num_topics, self.num_topics, self.num_terms))
+        assert self.eta.shape == (self.num_terms,), "Invalid alpha shape. Got shape %s, but expected (%d, )" % (str(self.eta.shape), self.num_terms)
```
olavurmortensen: Ok, re-re-formatted :)

gensim/test/test_ldamodel.py
```diff
@@ -159,57 +159,40 @@ def testEta(self):
...
-        testeta = np.array([[0.5] * len(dictionary)] * 2)
-        kwargs['eta'] = testeta
-        self.class_(**kwargs)
+        self.assertTrue(all(model.eta == np.array([0.3] * num_terms)))
 
         # all should raise an exception for being wrong shape
-        kwargs['eta'] = testeta.reshape(tuple(reversed(testeta.shape)))
-        self.assertRaises(AssertionError, self.class_, **kwargs)
-
         kwargs['eta'] = [0.3, 0.3, 0.3]
```
olavurmortensen: Done. Although just num_terms, not num_terms+1.

gensim/test/test_ldamodel.py
```diff
@@ -159,57 +159,40 @@ def testEta(self):
...
         kwargs['eta'] = 'symmetric'
         model = self.class_(**kwargs)
         self.assertEqual(model.eta.shape, expected_shape)
-        self.assertTrue(all(model.eta == np.array([[0.5], [0.5]])))
-
-        kwargs['eta'] = 'asymmetric'
-        model = self.class_(**kwargs)
-        self.assertEqual(model.eta.shape, expected_shape)
-        self.assertTrue(np.allclose(model.eta, [[0.630602], [0.369398]]))
+        self.assertTrue(all(model.eta == np.array([0.5] * num_terms)))
```
olavurmortensen: The "asymmetric" option should not be used for eta. It makes arbitrary words more likely apriori, which does not make sense. I have added an exception in the ldamodel.py file to make sure that `eta` is not equal to "asymmetric".

olavurmortensen:
I re-introduced the K x V asymmetric prior. While a K vector eta is not correct, there is a special place for the K x V prior.

As said in the docstring for eta, using K x V asymmetric prior the user may impose specific priors for each topic (e.g. to coerce a specific topic, or to get a "garbage collection" topic).

@tmylk can you read the new docstring, just to check that it's ok?

gensim/models/ldamodel.py
tmylk: what is the error that you get? which python version?

gensim/test/test_ldamodel.py
tmylk: where do you test that exception happens?
Please add `assertRaises` test case

gensim/test/test_ldamodel.py
tmylk: where is this change? line 213/196 is still the same

gensim/test/test_ldamodel.py
olavurmortensen: It is tested [here](https://github.com/olavurmortensen/gensim/blob/lda_hyperparam_fix/gensim/models/ldamodel.py#L310). I added an assertRaise in unit tests.

gensim/test/test_ldamodel.py
olavurmortensen: https://github.com/olavurmortensen/gensim/blob/lda_hyperparam_fix/gensim/test/test_ldamodel.py#L196

gensim/test/test_ldamodel.py
```diff
@@ -159,58 +159,41 @@ def testEta(self):
...
-        kwargs['eta'] = testeta
-        self.class_(**kwargs)
+        self.assertTrue(all(model.eta == np.array([0.3] * num_terms)))
 
         # all should raise an exception for being wrong shape
-        kwargs['eta'] = testeta.reshape(tuple(reversed(testeta.shape)))
-        self.assertRaises(AssertionError, self.class_, **kwargs)
-
-        kwargs['eta'] = [0.3, 0.3, 0.3]
+        kwargs['eta'] = [0.3] * num_terms
```
tmylk: this is same as line 188 above

gensim/test/test_ldamodel.py
tmylk: that is not a test, it is a check :) just kidding.

gensim/test/test_ldamodel.py
olavurmortensen: Rmoved it. Don't know why I put that there.

gensim/test/test_ldamodel.py
```diff
@@ -210,15 +201,15 @@ def testEta(self):
         kwargs['eta'] = testeta.reshape(tuple(reversed(testeta.shape)))
         self.assertRaises(AssertionError, self.class_, **kwargs)
 
-        kwargs['eta'] = [0.3, 0.3, 0.3]
-        self.assertRaises(AssertionError, self.class_, **kwargs)
-
         kwargs['eta'] = [0.3]
         self.assertRaises(AssertionError, self.class_, **kwargs)
```
tmylk: where did assertraises for a longer eta shape go?

tmylk:
Let's clean it up and then ready to merge. 
Please add a good explanation to CHANGELOG.md. I will publish your explanation of this breaking change together with release notes.

gensim/test/test_ldamodel.py
olavurmortensen: I'll add an assertRaises where shape of eta is num_terms+1.

olavurmortensen:
@tmylk should the explanation be under "0.13.5, 2016-11-12" in the changelog? Or where should I put it?

olavurmortensen:
@tmylk you said:
> Please re-do the merge with develop to make sure keyedvectors get properly merged in.

What do you mean? What happened?

tmylk:
The explanation should be on the top of CHANGELOG.md

olavurmortensen:
@tmylk [CHANGELOG.md updated](https://github.com/olavurmortensen/gensim/blob/013b5bb1b4f80db21babef7a38fcf03ff28c194f/CHANGELOG.md).

PR #1026, opened 2016-11-21
Open the vector file as utf-8 under Python 3 on Windows.

notnami:
Currently, this script crashes under Python 3 on Windows: The vector embedding files are opened with cp-1251 encoding by default, and so if the file contains any Unicode (which the GloVe embedding files do), the script will crash.

This PR passes 'utf-'8 as an explicit encoding to the smart_open library, which therefore handles the files correctly.

This PR comes in conjunction with https://github.com/RaRe-Technologies/smart_open/pull/97

### DIFF
--- a/gensim/scripts/glove2word2vec.py
+++ b/gensim/scripts/glove2word2vec.py
@@ -49,9 +49,9 @@ def glove2word2vec(glove_input_file, word2vec_output_file):
                     fout.write(line)
         return num_lines, num_dims
     else:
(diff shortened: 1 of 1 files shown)


piskvorky:
I'm not sure why this is dealing with text at all -- @tmylk shouldn't the strings be binary, and files opened as binary for read/write?

menshikh-iv:
@piskvorky you are right, the default mode is binary.

Issue #1006, opened 2016-11-11
Optimise LdaModel.show_topics performance to only call get_lambda once


nicoxx:
This function calls show_topic for each topic, which calls get_topic_terms which calls self.state.get_lambda().

self.state.get_lambda() call is very expensive, it is not dependent on a topic and can be called once for all.
On my setup (500 topics, 100k words, learned with 500k documents), calling show_topics takes 30sec, with 28sec inside the 500 get_lambda() calls

souravsingh:
I would like to work on the issue. How do I start?

tmylk:
Let's finish your other PRs first @souravsingh :)

bhargavvader:
@tmylk I tried a fair amount to try and fix this but the only solutions which I could think of are not very pretty.

We could maybe have an option to pass the `lambda` values, so instead of calling it within the function each time, the matrix is just passed around and we can pick up the row we need. 

The other option is to completely avoid calling `show_topic` in `show_topics`, and just write up code to do what is happening and keep `self.state.get_lambda()` only once in the beginning of the function. This also works, but is not really pretty, and there is code duplication.

Both the methods I tried work and pass tests without having to call `get_lambda` multiple times, but don't seem like the best solution, so I did not open a PR. If one of them sounds okay, I'll open a PR for it.

Any other suggestions?

tmylk:
The optimised show_topics should not call the show_topic. That is a good solution

nicoxx:
Thanks a lot ! I'll check this

PR #1028, opened 2016-11-23
Optimised show_topics

bhargavvader:
Possible solution to #1006 .
@tmylk could you review?

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -810,7 +810,7 @@ class LdaModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
                 topic_ = [(self.id2word[id], topic_[id]) for id in bestn]
                 topic_ = ' + '.join(['%.3f*"%s"' % (v, k) for k, v in topic_])
             else:
-                topic_ = topic_[i]
+                topic_ = topic[i]
                 topic_ = topic_ / topic_.sum()  # normalize to probability distribution
                 bestn = matutils.argsort(topic_, num_words, reverse=True)
                 topic_ = [(self.id2word[id], topic_[id]) for id in bestn]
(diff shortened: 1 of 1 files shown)


gensim/models/ldamodel.py
```diff
@@ -800,15 +800,24 @@ def show_topics(self, num_topics=10, num_words=10, log=False, formatted=True):
             chosen_topics = sorted_topics[:num_topics // 2] + sorted_topics[-num_topics // 2:]
 
         shown = []
+
+        topic = self.state.get_lambda()
         for i in chosen_topics:
             if formatted:
```
tmylk: Too much code duplication
Just add an extra `if formatted` block in the end that changes formatting

bhargavvader: Oh yes, that was quite stupid of me :shit:
Will fix!

bhargavvader:
@tmylk , removed duplicate code.

tmylk:
Is there a performance improvement with `%%timeit`?

bhargavvader:
Yeah, with the old method and 5 topics for the `news_corpus` dataset it was `100 loops, best of 3: 6.89 ms per loop`, with the optimised one it was `100 loops, best of 3: 4.28 ms per loop`

tmylk:
hmm, that doesn't agree with the original report of 

> On my setup (500 topics, 100k words, learned with 500k documents), calling show_topics takes 30sec, with 28sec inside the 500 get_lambda() calls

bhargavvader:
Could it maybe be because of the number of topics? Will increase number of topics and see what the difference is like. I only have 5 topics.

tmylk:
Yes, let's test the reported case

bhargavvader:
I got a significant improvement (20 times faster) with 500 topics (vocab is smaller though, len is 44143).

For the old method:

`1 loop, best of 3: 2.05 s per loop`

For the optimised method:

`10 loops, best of 3: 107 ms per loop`

tmylk:
Thanks a lot for the improvement!

bhargavvader:
😄

PR #1039, opened 2016-12-07
Lda models load/save backward compatibility across Python versions

anmol01gulati:
I've branched this off PR #913 . This PR is specifically for loading LDA models. There was some still some issue in loading(across python vecrsion) word2vec models in PR #913, so thought to separately tackle that issue. This PR looks ready to be merged for now.
@tmylk @piskvorky Please review.


gensim/models/ldamodel.py
```diff
@@ -995,7 +996,7 @@ def __getitem__(self, bow, eps=None):
         """
         return self.get_document_topics(bow, eps, self.minimum_phi_value, self.per_word_topics)
 
-    def save(self, fname, ignore=['state', 'dispatcher'], *args, **kwargs):
+    def save(self, fname, ignore=['state', 'dispatcher'], separately = None, *args, **kwargs):
```
piskvorky: PEP8: no `separately=None` (no spaces in argument params).

gensim/models/ldamodel.py
```diff
@@ -1034,7 +1035,41 @@ def save(self, fname, ignore=['state', 'dispatcher'], *args, **kwargs):
...
+        if self.id2word is not None and 'id2word' not in ignore:
+            id2word = dict((k,v) for k, v in self.id2word.iteritems())
+        self.id2word = None # remove the dictionary from model
+        super(LdaModel, self).save(fname, ignore=ignore, separately = separately, *args, **kwargs)
+        self.id2word = id2word # restore the dictionary.
+
+        # Save the dictionary separately in json.
+        id2word_fname = utils.smart_extension(fname, '.bin')   
+        try:
+            with utils.smart_open(id2word_fname, 'w', encoding='utf-8') as fout:
```
piskvorky: I'd prefer to open the file as binary and explicitly write binary (utf8) strings into it.

gensim/models/ldamodel.py
```diff
@@ -1034,7 +1035,41 @@ def save(self, fname, ignore=['state', 'dispatcher'], *args, **kwargs):
...
+                json.dump(id2word, fout)
+        except Exception as e:
+            logging.warning("failed to save id2words dictionary in %s: %s", id2word_fname, e)
```
piskvorky: Should this be a warning, or exception? What's the user contract on storing this "id2words" dictionary?

gensim/test/test_data/ldamodel_python_2_7.bin
```diff
@@ -0,0 +1 @@
+{"0": "interface", "1": "computer", "2": "human", "3": "response", "4": "time", "5": "survey", "6": "system", "7": "user", "8": "eps", "9": "trees", "10": "graph", "11": "minors"}
```
piskvorky: This doesn't look like a binary (extension `.bin`?) file.

gensim/utils.py
```diff
@@ -906,10 +906,12 @@ def pickle(obj, fname, protocol=2):
 
 def unpickle(fname):
     """Load pickled object from `fname`"""
-    with smart_open(fname) as f:
+    with smart_open(fname, 'rb') as f:
         # Because of loading from S3 load can't be used (missing readline in smart_open)
-        return _pickle.loads(f.read())
-
+        if sys.version_info > (3, 0):
+            return _pickle.load(f, encoding='latin1')
```
piskvorky: Absolutely not! What is this `latin1`?

gensim/models/ldamodel.py
anmol01gulati: Done.

gensim/models/ldamodel.py
anmol01gulati: I was saving the id2word dictionary  separately in json format(following [this](https://github.com/RaRe-Technologies/gensim/wiki/Recipes-&-FAQ#q9-how-do-i-load-a-model-in-python-3-that-was-trained-and-saved-using-python-2). I've changed it now, and added id2word in the separately list, while saving the model.

gensim/models/ldamodel.py
anmol01gulati: Yup, it would be better to have it as an exception. Removed this now. It should return an exception if it fails to save the model.

anmol01gulati:
I've made the necessary changes now. The id2word dictionary is saved in binary format. 
We only create a new file and save the id2word dictionary when it's not marked as 'ignored'. So, when loading thedictionary from the file(in ldamodel), we check if the file exists, and only then load the dictionary.
To accomodate this, I had to change the testfile() function in test_ldamodel, so thus, a different file is created for each test. Does this sound good? 
@piskvorky @tmylk We could merge it then, if it looks fine.

gensim/utils.py
anmol01gulati: This works as a fix for when loading objects in Python 3 which were pickled in Python 2, which gives an exception.
Basically, Python 3 attempts to convert the pickled py2 object into a str object, when we need it to be bytes and gives an exception. I used the latin1 encoding for as a work around for that. [Asked on Stackoverflow](http://webcache.googleusercontent.com/search?q=cache:lHZB47tGMV4J:stackoverflow.duapp.com/questions/33596082/load-gensim-word2vec-computed-in-python-2-in-python-3+&cd=5&hl=en&ct=clnk&gl=in)

tmylk: Can you add a code comment to explain this?

tmylk:
A code comment on `latin`` is needed in a separate PR.

gensim/utils.py
piskvorky: Yes, this hack needs to be marked and explained thoroughly in a comment.

I'm not familiar with such py2/py3 pickle work arounds, but isn't there a cleaner way to achieve this? @tmylk @anmol01gulati

anmol01gulati: @piskvorky Umm, I had actually searched quite a lot, and tried various things on my system. This is the only way(a hack actually), through which it works. By the way, I felt, we would not want to have this functionality in the future and could do away with the backward compatibility, if majority of the users shift to one Python 3 later (it's not the case right now though).
I'll open up a new PR to add a comment in the code though.

SwedishBotMafia: I am coding something entirely different and this solution is the only thing that worked for loading python2 pickles in python3... The creators claim that pickle is backwards compatible but apparently only if I pass latin1... Any other way just breaks and burns.

PR #1053, opened 2016-12-22
Ignore DocvecsArray.doctag_syn0norm in save

### LINKED ISSUE #789: Doc2Vec.DocvecsArray.doctag_syn0norm not ignored in save()
Unlike with Word2Vec.syn0norm, oversized `doctag_syn0norm` arrays aren't ignored during native `save()`. As a result, even if they're a shared-reference to `doctag_syn0` after an `init_sims(replace=True)`, both `doctag_syn0` and `doctag_syn0norm` will be written and then read back in, using more storage/RAM than necessary. (See https://groups.google.com/d/msg/gensim/fCcHoZCyV3M/djEZgNWEAAAJ )

accraze:
This PR contains an update to `DocvecsArray`:

* Ignores normalized vectors on `save()`

Fixes #789

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -353,6 +353,11 @@ class DocvecsArray(utils.SaveLoad):
         else:
             return index in self.doctags
 
+    def save(self, *args, **kwargs):
+        # don't bother storing the cached normalized vectors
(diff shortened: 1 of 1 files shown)


tmylk:
Thanks for the fix!

PR #1054, opened 2016-12-22
Making Default hs error explicit

### LINKED ISSUE #673: default-value of 'hs' (hierarchical-softmax) parameter changed w/o explicit warning in 0.12.4
In vs. 0.12.4, the new default of 'hs=0' stays in accordance with the online documenation, but is in contrast to the default-behavior of versions <= 0.12.2. 
If not set explicitely set to 'hs=1', the w2vec.score() function will fail with the error message "we have only implemented score for hs". In a productive environment, this error message might appear as not trivial, and is assumed to be more productive when pointing directly to the change-need of the model-parameters.

accraze:
This PR contains an update to `Word2Vec`:

* make default-hs error explicit

Fixes #673


tmylk:
Thanks for the PR!

PR #1061, opened 2016-12-27
Tests for the evaluate_word_pairs function

akutuzov:
Test for evaluating model against semantic similarity datasets  (#1047).
Also fixes an error in the function call.


akutuzov:
@tmylk the tests are ready.

gensim/test/test_word2vec.py
```diff
@@ -378,6 +378,16 @@ def testAccuracy(self):
...
 
+    def testEvaluateWordPairs(self):
+        """Test Spearman and Pearson correlation coefficients give sane results on similarity datasets"""
+        corpus = word2vec.LineSentence(datapath('head500.noblanks.cor.bz2'))
+        model = word2vec.Word2Vec(corpus, min_count=3, iter=10)
+        correlation = model.evaluate_word_pairs(datapath('wordsim353.tsv'))
+        pearson = correlation[0][0]
+        spearman = correlation[1][0]
+        self.assertTrue(0.1 < pearson < 1.0)
+        self.assertTrue(0.1 < spearman < 1.0)
```
tmylk: could we please test for `oov_ratio` in `correlation[2]` too?

akutuzov:
Sure, done.

tmylk:
Thanks for the improvement!

tmylk:
By the way, how is it better than using https://github.com/mfaruqui/eval-word-vectors ? CC @anmol01gulati

akutuzov:
It's better in that this code works directly from Gensim :)
In fact, my code is simpler as it uses _Scipy_ functions for Pearson and Spearman coefficients ([eval-word-vectors](https://github.com/mfaruqui/eval-word-vectors) implements Spearman from scratch). Also, it features some useful options, like case-(in)sensitivity and smart handling of OOV pairs.

anmol01gulati:
I agree with @akutuzov. The code currently in gensim for Pearson and Spearman coefficients is shorter. But I feel, we could also include the whole dataset for evaluating word vectors, given in [https://github.com/mfaruqui/eval-word-vectors](https://github.com/mfaruqui/eval-word-vectors). It's just 205 KB, and contains all the major gold standards, it'd be good to integrate them into gensim itself, and have one method to directly evaluate word2vec models, right inside gensim. What do you think?

The script I used to convert word2vec into the format for evaluating word vectors is quite small actually: 
```
import gensim

model = gensim.models.Word2Vec.load_word2vec_format(
    'GoogleNews-vectors-negative300.bin', binary=True)

words = [line.split()[0] for line in open(
    "eval-word-vectors/vocab.txt", 'r')]

with open('output_vecs.txt', 'wb') as f:
    for word in words:
        if word in model:
            word_vector = model[word]
            f.write("%s " % word)
            f.write(" ".join(str(x) for x in word_vector))
            f.write("\n")
```

akutuzov:
I am not sure it's a good idea to overload Gensim with various semantic similarity datasets included in the distribution.
Most people would use their own gold datasets anyway, either because they deal with non-English data or because their text preprocessing differs from the preprocessing in SimLex999 or WS353 (lemmatization/stemming, POS-tagging, etc).
So I think it's better to leave WS353 as an example (and for testing), and may be put a couple of links to other datasets in the documentation.

anmol01gulati:
Yeah you are right. Sounds Good.

PR #1062, opened 2016-12-27
Utils and Matutils changes

bhargavvader:
1. Moved `dirichlet_expectation` to `matutils` as the same code was being used in both `ldamodel` and `hdpmodel`.
2. Changed all `numpy` in `utils` and `matutils` to `np`, keeping it uniform across the package.
3. Fixed hanging indent.
4. Changed import conventions in `hdpmodel` to make it consistent with `ldamodel`

### DIFF
--- a/gensim/models/hdpmodel.py
+++ b/gensim/models/hdpmodel.py
@@ -35,7 +35,7 @@ from __future__ import with_statement
 
 import logging, time
 import numpy as np
-import scipy.special as sp
+from scipy.special import gammaln, psi  # gamma function utils
 
 from gensim import interfaces, utils, matutils
 from gensim.models import basemodel, ldamodel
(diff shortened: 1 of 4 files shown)


gensim/utils.py
```diff
@@ -608,7 +607,7 @@ def is_corpus(obj):
             doc1 = next(iter(obj))  # empty corpus is resolved to False here
         if len(doc1) == 0:  # sparse documents must have a __len__ function (list, tuple...)
             return True, obj  # the first document is empty=>assume this is a corpus
-        id1, val1 = next(iter(doc1))  # if obj is a numpy array, it resolves to False here
+        id1, val1 = next(iter(doc1))  # if obj is a np array, it resolves to False here
```
piskvorky: This replacement (and the one above) look unnecessary, the original was clearer.

bhargavvader: Cool, will fix it

gensim/models/hdpmodel.py
```diff
@@ -74,15 +64,15 @@ def expect_log_sticks(sticks):
 
 def lda_e_step(doc_word_ids, doc_word_counts, alpha, beta, max_iter=100):
     gamma = np.ones(len(alpha))
-    expElogtheta = np.exp(dirichlet_expectation(gamma))
+    expElogtheta = np.exp(matutils.dirichlet_expectation(gamma))
```
tmylk: better to add `from matutils import dirichlet_expectation`

bhargavvader: Okay, changed

tmylk:
Thanks for the improvement!

gensim/utils.py
```diff
@@ -829,15 +828,15 @@ def __init__(self, q, corpus, chunksize, maxsize, as_numpy):
 
     def run(self):
         if self.as_numpy:
-            import numpy # don't clutter the global namespace with a dependency on numpy
+            import np # don't clutter the global namespace with a dependency on numpy
```
piskvorky: Looks like the global module namespace already has numpy, so this line is unnecessary.

gensim/utils.py
```diff
@@ -801,13 +800,13 @@ def chunkize_serial(iterable, chunksize, as_numpy=False):
     [[0, 1, 2], [3, 4, 5], [6, 7, 8], [9]]
 
     """
-    import numpy
+    import numpy as np
```
piskvorky: Already imported at module level, this line no longer unnecessary.

PR #1065, opened 2016-12-29
Remove unnecessary numpy imports in Utils

bhargavvader:
With respect to the comments by @piskvorky in #1062 .


piskvorky:
Thanks for the swift resolution :)

PR #1074, opened 2017-01-05
Converting notebooks to html for further publishing

markroxor:
Raw generation of htmls from the notebook in the folder docs/notebooks (for now).


markroxor:
Extended at #1091

PR #1081, opened 2017-01-07
Pep8 fixes

nishnik:
@tmylk 
I have ignored `E731`. I think that it is better to ignore it else it forbids us to define lambda expressions like:
`datapath = lambda fname: os.path.join(module_path, 'test_data', fname)`


gensim/corpora/lowcorpus.py
```diff
@@ -166,8 +167,8 @@ def save_corpus(fname, corpus, id2word=None, metadata=False):
 
         if truncated:
             logger.warning("List-of-words format can only save vectors with "
-                            "integer elements; %i float entries were truncated to integer value" %
-                            truncated)
+                           "integer elements; %i float entries were truncated to integer value" %
```
piskvorky: We want no vertical indent in gensim. I don't know if you're using some tool or making the changes manually in this PR, but can you fix those cases too (use hanging indent)?

gensim/corpora/malletcorpus.py
```diff
@@ -107,8 +108,8 @@ def save_corpus(fname, corpus, id2word=None, metadata=False):
 
         if truncated:
             logger.warning("Mallet format can only save vectors with "
-                            "integer elements; %i float entries were truncated to integer value" %
-                            truncated)
+                           "integer elements; %i float entries were truncated to integer value" %
```
piskvorky: Dtto (no vertical indent).

gensim/corpora/sharded_corpus.py
```diff
@@ -233,10 +233,10 @@ def __init__(self, output_prefix, corpus, dim=None,
         self.current_shard = None    # The current shard itself (numpy ndarray)
         self.current_shard_n = None  # Current shard is the current_shard_n-th
         self.current_offset = None   # The index into the dataset which
-                                     # corresponds to index 0 of current shard
+        # corresponds to index 0 of current shard
```
piskvorky: Don't split the comment: all in one line (either same line as var assignment, or the preceding line, but not after).

gensim/corpora/sharded_corpus.py
```diff
@@ -233,10 +233,10 @@ def __init__(self, output_prefix, corpus, dim=None,
...
 
         logger.info('Initializing sharded corpus with prefix '
-                     '{0}'.format(output_prefix))
+                    '{0}'.format(output_prefix))
```
piskvorky: No vertical indent.

gensim/corpora/sharded_corpus.py
```diff
@@ -245,7 +245,7 @@ def __init__(self, output_prefix, corpus, dim=None,
             # and retain information about how the corpus
             # was serialized.
             logger.info('Saving ShardedCorpus object to '
-                         '{0}'.format(self.output_prefix))
+                        '{0}'.format(self.output_prefix))
```
piskvorky: No vertical indent.

gensim/corpora/sharded_corpus.py
```diff
@@ -262,12 +262,12 @@ def init_shards(self, output_prefix, corpus, shardsize=4096, dtype=_default_dtyp
...
             if self.dim is None:
                 logger.info('Deriving dataset dimension from corpus: '
-                             '{0}'.format(proposed_dim))
+                            '{0}'.format(proposed_dim))
             else:
                 logger.warn('Dataset dimension derived from input corpus diffe'
-                             'rs from initialization argument, using corpus.'
-                             '(corpus {0}, init arg {1})'.format(proposed_dim,
-                                                                 self.dim))
+                            'rs from initialization argument, using corpus.'
```
piskvorky: Vertical indent (last such comment here, but there are more cases).

gensim/corpora/sharded_corpus.py
```diff
@@ -647,15 +647,15 @@ def __getitem__(self, offset):
             #    to (stop - current_offset)
             shard_start = start - self.current_offset
             shard_stop = self.offsets[self.current_shard_n + 1] - \
-                         self.current_offset
+                self.current_offset
```
piskvorky: Should be on previous line (no line splitting).

gensim/corpora/sharded_corpus.py
```diff
@@ -789,8 +789,8 @@ def save(self, *args, **kwargs):
         if 'ignore' not in kwargs:
             kwargs['ignore'] = frozenset(attrs_to_ignore)
         else:
-            kwargs['ignore'] = frozenset([v for v in kwargs['ignore']]
-                                         + attrs_to_ignore)
+            kwargs['ignore'] = frozenset([v for v in kwargs['ignore']] +
+                                         attrs_to_ignore)
```
piskvorky: Should be on previous line (no line splitting).

piskvorky:
Wow, massive PR! Thanks, much needed.

I would ask for two changes if possible:
* get rid of all cases of "vertical indent" (hard to maintain, gets out of sync quickly, less screen real estate) -- use normal hanging indent, consistently
* some lines are split nonsensically (hurts readability & less screen estate) -- merge them back into a single line.

I pointed out some examples of these issues in review comments.

gensim/corpora/sharded_corpus.py
nishnik: @piskvorky 
These lines were already split. Here `autopep8` corrected the vertical indentation.
Though we can write a script to join these kind of lines, but that would result into many counter-intuitive fixings where lines are split for documentation purposes.

gensim/corpora/sharded_corpus.py
nishnik: Here `autopep8` changed the position of binary operator `+`. It is to correct the warning `W503`, i.e. *line break occurred before a binary operator*.

nishnik:
@piskvorky 
Thanks for the review.
I would run the script again to get rid of "Vertical indent" fixes. But for the second case, none of the lines were divided by the script, they are on their original lines with fixes for indentation or binary operators.

piskvorky:
I don't think your script did the vertical indents either. It's something that has crept into gensim in previous merges, and we need to get rid of it and clean up now (and be careful that new PRs don't re-introduce it).

nishnik:
@piskvorky
I think I misunderstood your comments and removed all the indentation related changes.
I will apply them back.
And yes the vertical indents are not generated by the script, but one can never tell whether split lines are desired or not (there are some places where lines are split explicitly for documentation purposes), so I would advocate that through this PR, we can solve all the lint related issues which can be detected by `pep8`. And the Travis check will make sure that they are not introduced in any future code.

piskvorky:
To be clear, this is what I meant:

```python
# NO
result = some_function(long_blabla,
                                       long_blublu)

# YES: easier to maintain & more compact
result = some_function(
    long_blabla,
    long_blublu)
```

I'm OK if this is fixed separately (though this PR looks as good place as any). But we better do it sooner rather than later, before it creeps even further into the gensim code base.

tmylk:
@nishnik / @souravsingh Is there a way to error on vertical indent and enforce hanging indent with pep8/coala?

souravsingh:
@tmylk There is a coala tool called IndentationBear, which I suppose can help enforce  hanging indents.

nishnik:
It would be nice if @souravsingh can use Coala to do this, I think `pep8` doesn't support "enforcing hanging indents" though I am looking of ways to hack into its code and make it work as required.

.travis.yml
```diff
@@ -8,6 +8,9 @@ python:
   - "3.5"
   - "3.6"
 before_install:
+  - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
+        pip install pep8;
```
OrkoHunter: Hi @nishnik,
It has been renamed to [pycodestyle](https://github.com/PyCQA/pycodestyle) ! :)

nishnik:
@tmylk @piskvorky I have made some changes in two files in the [recent commit](https://github.com/RaRe-Technologies/gensim/pull/1081/commits/036afd5fac0be254b139f20f40b36b64b2785d18).
Have a look, I tweaked with `pep8` (now `pycodestyle`) to catch these errors.

.travis.yml
nishnik: Thanks @OrkoHunter . Their [discussion](https://github.com/PyCQA/pycodestyle/issues/466) is awesome too.

tmylk:
@nishnik Thanks. So is there a way to enforce hanging indent over vertical in `pycodestyle`? Sadly the `indentbear` in Coala doesn't support Python

nishnik:
@tmylk I tweaked with `pycodestyle` source code. I think it won't be merged in the original repo of `pycodestyle` If we need to raise errors for vertical indent then we will need the whole script in our repo (*or a safe place so that we could fetch it in travis*). I would run it on all files, make the suggested changes and then push it on my repo. There are many errors, so it will take some time.

It would be better if you look into the two files changed in the previous commit and comment if we need more changes in them.

gensim/models/lda_dispatcher.py
```diff
@@ -181,24 +181,24 @@ def exit(self):
...
     parser.add_argument('-v', '--verbose', help='Verbose flag', action='store_const', dest="loglevel",
-                        const=logging.INFO, default=logging.WARNING)
+        const=logging.INFO, default=logging.WARNING)
     args = parser.parse_args()
 
     logging.basicConfig(format='%(asctime)s : %(levelname)s : %(message)s', level=args.loglevel)
     logger.info("running %s", " ".join(sys.argv))
 
     ns_conf = {"broadcast": args.no_broadcast,
-               "host": args.host,
```
piskvorky: Hanging indent has the first line empty (content starts at the first indented line). See the examples in [PEP8](https://www.python.org/dev/peps/pep-0008/#indentation).

(comment applies here and elsewhere)

souravsingh:
@nishnik Could you close the PR and make separate PR, one for each module in Gensim?

It would help in "review and rectification" process.

PR #1084, opened 2017-01-10
Lazy formatting in evaluate_word_pairs

akutuzov:
As requested in #1079.


tmylk:
Thanks for changing to hanging indent.

PR #1099, opened 2017-01-20
Phrases and Phraser allow a generator corpus

ELind77:
Allow Phrases and Phraser models to take a generator
function/expression as input to the transformation method. Previously,
only indexable iterables could be used which is problematic for large
corpora.

Add additional tests to test_phrases using a generator as input.

I was trying to build a Phrases model with a corpus bigger than 1GB.  My corpora are always generators and usually that's fine with gensim but it didn't work this time so I fixed it.  The fix is basically just adding a function in [phrases.py](https://github.com/ELind77/gensim/blob/phrases_with_generator_corpus/gensim/models/phrases.py#L75) to do some more checking as to the type of the input.  It also accommodates a generator of generators in case each document/token is undergoing some kind of transformation as it's pulled through.

This PR does still need a small modification to the docstings in Phraser and Phrases in order to indicate that the input type can be a generator and not just a list but my rst isn't all that great.  Maybe if someone reviews this I could get a pointer on that?

### DIFF
--- a/gensim/test/test_phrases.py
+++ b/gensim/test/test_phrases.py
@@ -50,6 +50,19 @@ class TestPhrasesCommon(unittest.TestCase):
         self.bigram_utf8 = Phrases(sentences, min_count=1, threshold=1)
         self.bigram_unicode = Phrases(unicode_sentences, min_count=1, threshold=1)
 
+    def testEmptyInputsOnBigramConstruction(self):
+        """Test that empty inputs don't throw errors and return the expected result."""
+        # Empty list -> empty list
+        self.assertEqual(list(self.bigram_default[[]]), [])
+        # Empty iterator -> empty list
+        self.assertEqual(list(self.bigram_default[iter(())]), [])
+        # List of empty list -> list of empty list
+        self.assertEqual(list(self.bigram_default[[[], []]]), [[], []])
+        # Iterator of empty list -> list of empty list
+        self.assertEqual(list(self.bigram_default[iter([[], []])]), [[], []])
+        # Iterator of empty iterator -> list of empty list
+        self.assertEqual(list(self.bigram_default[(iter(()) for i in range(2))]), [[], []])
+
     def testSentenceGeneration(self):
         """Test basic bigram using a dummy corpus."""
         # test that we generate the same amount of sentences as the input
--- a/gensim/test/test_phrases.py
+++ b/gensim/test/test_phrases.py
@@ -38,6 +38,10 @@ sentences = [
 unicode_sentences = [[utils.to_unicode(w) for w in sentence] for sentence in sentences]
 
 
+def gen_sentences():
+    return ((w for w in sentence) for sentence in sentences)
+
+
 class TestPhrasesCommon(unittest.TestCase):
     """ Tests that need to be run for both Prases and Phraser classes."""
     def setUp(self):
@@ -50,6 +54,11 @@ class TestPhrasesCommon(unittest.TestCase):
         """Test basic bigram using a dummy corpus."""
         # test that we generate the same amount of sentences as the input
         self.assertEqual(len(sentences), len(list(self.bigram_default[sentences])))
+
+    def testSentenceGenerationWithGenerator(self):
+        """Test basic bigram production when corpus is a generator."""
+        self.assertEqual(len(list(gen_sentences())),
+                         len(list(self.bigram_default[gen_sentences()])))
 
     def testBigramConstruction(self):
         """Test Phrases bigram construction building."""
@@ -74,6 +83,20 @@ class TestPhrasesCommon(unittest.TestCase):
         self.assertTrue(u'graph_minors' in self.bigram[sentences[-2]])
         self.assertTrue(u'graph_minors' in self.bigram[sentences[-1]])
         self.assertTrue(u'human_interface' in self.bigram[sentences[-1]])
+
+    def testBigramConstructionFromGenerator(self):
+        """Test Phrases bigram construction building when corpus is a generator"""
+        bigram1_seen = False
+        bigram2_seen = False
+
+        for s in self.bigram[gen_sentences()]:
+            if not bigram1_seen and 'response_time' in s:
+                bigram1_seen = True
+            if not bigram2_seen and 'graph_minors' in s:
+                bigram2_seen = True
+            if bigram1_seen and bigram2_seen:
+                break
+        self.assertTrue(bigram1_seen and bigram2_seen)
 
     def testEncoding(self):
         """Test that both utf8 and unicode input work; output must be unicode."""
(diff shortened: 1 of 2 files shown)


gensim/test/test_phrases.py
```diff
@@ -75,6 +84,20 @@ def testBigramConstruction(self):
         self.assertTrue(u'graph_minors' in self.bigram[sentences[-1]])
         self.assertTrue(u'human_interface' in self.bigram[sentences[-1]])
 
+    def testBigramConstructionFromGenerator(self):
+        """Test Phrases bigram construction building when corpus is a generator"""
+        bigram1_seen = False
+        bigram2_seen = False
+
+        for s in self.bigram[gen_sentences()]:
+            if not bigram1_seen and 'response_time' in s:
```
tmylk: what would be lost if just `assert 'response_time'  in self.bigram[gen_sentences()]`

tmylk:
Would it be easier to modify `_apply` so that other methods can benefit from this as well?

gojomo:
I like the goal and thank you for including tests! But I'm concerned this seems a bit of a convoluted approach. It's got twisty conditionals, a catchall exception handler, and is named-like a predicate-function (`is_…`) but returns a tuple. 

Maybe it has to be, to continue to offer Phrase-ification via `__getitem__()` index-lookup, for multiply-shaped input types. The existing `is_single` test to achieve that is already kind of clunky, and the idea of treating phrase-ification as if it were a []-keyed-lookup has always been non-intuitive to me, though I know some other parts of the gensim API work similarly.

In the interest of explicitness/simplicity, what if `phrasify_one(sentence)` and `phrasify_many(sentences)` were explicit parts of the API, so the user of the code indicates what they expect done, rather than leaving it to a bunch of type-checking? 

If a `__getitem__()` accept-and-adapt-to-whatever entry point remains. I believe that peeking at the first item from the argument (which should always be iterable of some sort) should be enough. If it's a string, treat it as a single-text, and if it's anything else (generator or not), consider it an iterable-of-iterables. For example:
~~~
sent_iter = iter(sentences)
try:
    peek = sent_iter.next()
    sentences = it.chain([peek], sent_iter)  # restore sentences to include peeked item
except StopIteration:
    return []
# …switch behavior just on whether peek is string or anything-else…
~~~

gensim/test/test_phrases.py
ELind77: This is a cop-out, but I just copied the format of the [existing test](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_phrases.py#L54) :P

To stream-line that section you could do:
```python
assert len(set(['response_time', 'graph_minors']).intersection(set(it.chain.from_iterable(self.bigram[gen_sentences()])))) == 2
```

However, the current formulation in both my test and [testBigramConsturction](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_phrases.py#L54) short-circuits so that it doesn't have to go through the entire input.  This would allow longer test corpora in the future.  

In either case, I didn't want to question or rethink the work of whoever designed the tests originally, I just wanted to make sure my changes were non-breaking.  Do you want me to change the tests?

ELind77:
@tmylk I don't think there's any problem with `_apply`.  The issue is [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/phrases.py#L250), when Phrases/Phraser checks if the input is a single document or a corpus.

@gojomo I think you're right, it's convoluted, I was more concerned with changing as little of the existing code as possible than with making the new approach clear.  I think your proposed `peek` with `phrasiphy_one` and `phraseiphy_many` is the better approach.  Is there any argument for tying to unify `Phrases` and `Phraser`?  The `phrasiphy_one` code would be sooo close but not quite close enough to actually be easy to do well.

Also, should the docstings for `__getitem__` be modified to indicate more versatile input types?  E.g. here: https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/phrases.py#L231

Last question, should I merge my branch back in to my fork's develop branch?

tmylk:
@ELind77 This PR won't be merged as it is. Waiting for an improvement with `phrasify_one/many` .

piskvorky:
My suggestion would be to only have one `transform()` function, which accepts a stream of multiple items (here, sentences). Calling it on one sentence is just `transform([one_sentence])`, no need to clutter the API.

And yes, we could offer a similar function (called `transform` to keep in sync with sklearn terminology) in other modules too, in addition to the existing `[ ]` notation.

ELind77:
After going back and looking at the code again I don't think it makes sense to change the public API.  I think it's better and more consistent to keep the API exactly as it is.  

I think the main issue @gojomo pointed out is the complexity of `_is_single`.  In order to improve that I took a look at how `TfidfModel` does it and was pleasantly surprised when I discovered that [utils.is_corpus](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/utils.py#L581) behaves in exactly the same way as my `_is_single` function.  I think we can resolve the problems pointed out above by simplifying `_is_single` (I can make it public too if that's preferable) and giving it a contract and documentation like `is_corpus` has in the utilities.

A simple modification to @gojomo's code with some documentation inspired by `is_corpus` give this:

```python
def _is_single(obj):
    """
    Check whether `obj` is a single document or an entire corpus.
    Returns (is_single, new) 2-tuple, where `new` yields the same
    sequence as `obj`.

    `obj` is a single document if it is an iterable of strings.  It
    is a corpus if it is an iterable of documents.
    """
    obj_iter = iter(obj)
    try:
        peek = obj_iter.next()
        obj_iter = it.chain([peek], obj_iter)
    except StopIteration:
        # An empty object is a single document
        return True, obj
    if isinstance(peek, string_types):
        # It's a document, return
        return True, obj_iter
    else:
        return False, obj_iter
```

What do you guys think?  I think the naming could use some tweaking, but it is exactly the same problem and solution that `is_corpus` solves for raw inputs instead of transformed corpora.

tmylk:
@ELind77 Looks concise. Would you like to update the PR with that version?
The change return order to `is_single, sentence` is also good idea to match `is_corpus`.

Adding a test for `self.bigram[[]]` would be good too

ELind77:
@tmylk I've pushed the changes to `_is_single`.  What were the changes you wanted me to make to the tests?  My original PR adds two tests that repeat some of the existing test functions with a generator input instead of lists.

ELind77:
I took a look at the error from travis and it looks like one of my tests is failing for python 3.x but I'm not sure why.  Something different in the way 3.x handles generator expressions perhaps?

tmylk:
`next` is different between Python 2 and 3
```
from six import next
next(g)
```

And an explicit test that no ouput and no exception is received for an empty input `[]` is needed: `self.bigram[[]]`

ELind77:
Ok, changes made and test added.  

I also removed the TODO comments in `phrases.py` concerning the docstrings as I think the existing module documentation already implies that the phrase models are consistent with iterators.

tmylk:
@ELind77 Thanks for the PR!

Issue #1108, opened 2017-01-26
Comment error in utils.py


greninja:
In comment of  [Line 610](https://github.com/RaRe-Technologies/gensim/blob/3cb735b278995fc5d4e2e05964cb162cbb6a8327/gensim/utils.py#L610) of utils.py , even if  `obj` is a numpy array it shouldn't resolve to `False`. 

I ran all the tests on it from `test_utils.py`  , except those from `test_invalid_formats` , by explicitly converting `potentialCorpus` to a numpy array and it gave 'True'.

I dont get it why it should resolve to False?
PS : Using numpy version '1.11.3'.

piskvorky:
`obj` being a numpy array is fine and the function returns `True`, if the content of the array is correct. Its elements have to be sequences of 2-tuples (or something convertable to 2-tuple, such as a 1d array of length 2).

If its first element, `doc1` in code, is a 1d numpy array instead of a sequence of 2-tuple, that line raises an exception. Then `is_corpus == False` => `obj` isn't a corpus.

For example, the following are all numpy arrays that pass as corpora (return `True`):

```python
obj = np.random.rand(10, 5, 2)  # 10 documents, each 5 non-zero features
obj = np.array([[(1, 0.2), (10, 0.3)], [(0, 3.8)]])  # 2 documents: 1st has two features, 2nd 1 feature
```

The following are *not* corpora:

```python
obj = np.random.rand(10, 5, 3)  # document elements 3-tuples instead of 2-tuples
obj = np.random.rand(10, 5)  # documents elements 1-tuples (scalars) instead of 2-tuples
obj = np.random.rand(10)  # documents scalars instead of a sequence of 2-tuples
```

Makes sense?

greninja:
Absolutely.

Then I think it should be mentioned explicitly that it should be a 2 tuple
sequence(in that particular comment) otherwise it's a bit misleading.

piskvorky:
You're right. Can you send a PR improving/clarifying that comment?

greninja:
Umm. Actually even if the obj is a 1D numpy array of length 2, it would still give an error => obj is not a corpus.

Consider : 
`obj  = [1,2]` or even `obj = [(1,2)]` for that matter
The first `next(iter(obj)` would give the first element of the iterable itself which is an `int` and hence the second `next(iter(doc1))` would raise exception since `int` object is not iterable.

piskvorky:
[...]

PR #1109, opened 2017-01-27
Clarifying comment in is_corpus func in utils.py

greninja:
Discussed in https://github.com/RaRe-Technologies/gensim/issues/1108#issuecomment-275392051


tmylk:
Thanks for the clarification!

piskvorky:
That comment is not coherent:

```
if obj is a 1D numpy array(scalars) instead of 2-tuples, it resolves to False here
```

I'd propose `if obj is a 1D numpy array instead of a sequence of 2-tuples, doc1 will be a scalar => resolves to False here`.

Issue #1105, opened 2017-01-24
Annoy tutorial inconsistency


piskvorky:
The [Annoy tutorial](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/annoytutorial.ipynb) states that Annoy is much faster than brute force.

But then cells 108 and 109 show the opposite -- Annoy is 3x *slower* than brute force.

This is confusing to the reader.

The discrepancy could be to any number of reasons (cold caches etc). For a slightly better "micro benchmark", measure a loop of e.g. 1000 random queries.

In case either index is memory-mapped, make sure it's fully in RAM before starting the benchmarking (go over it once with a "dry run", to generate all the page faults). If either index doesn't fit fully in RAM, don't run the benchmark at all.

greninja:
I ran a benchmark test over a random sample of 1000 words from `model.vocab` for similarity queries ( havent tried retrieving indices from disk as of yet) and the wall time taken by Annoy still comes out to be roughly 3 times more.

Also , in cell 112, is the word "army" or "science"?

tmylk:
@greninja This is very different from the [first commit](https://github.com/RaRe-Technologies/gensim/blob/4ce3ce143dcc97bf0a917dbf3a7b828f1890668b/docs/notebooks/annoytutorial.ipynb) where it is 12ms vs 678us.

It needs further investigation:
Does annoy use multiple cores on your machines? Did something take change in the implementation?

greninja:
No change in implementation.

I ran this code : 

`import random
vocab_list = random.sample(model.vocab, 1000)  # to get random queries

# Traditional method
%%time
    ...: for word in vocab_list:
    ...:     vector = model[word]
    ...:     model.most_similar([vector],topn=5)
    ...: 
CPU times: user 1.52 s, sys: 1.6 s, total: 3.12 s
Wall time: 914 ms

# Using Annoy
In [24]: %%time
    ...: for word in vocab_list:
    ...:     vector = model[word]
    ...:     model.most_similar([vector],topn=5,indexer=annoy_index)
    ...: 
CPU times: user 320 ms, sys: 0 ns, total: 320 ms
Wall time: 317 ms
`

It showed results what the notebook doc claims.

droudy:
I put in a PR to address the issue: #1113

lgmoneda:
@tmylk Annoy isn't running in mutiple cores, while traditional implementation does.

Running to 10k words:
**Traditional**
[...]

PR #1113, opened 2017-01-29
Address benchmark inconsistencies in Annoy tutorial  #1105

droudy:
Issue #1105 

Uses average query time of 1000 random queries as opposed to only a single query. Includes a "dry run" before running queries. Also fixes a discrepancy where a comment says that the vector for "army" is being retrieved when the word is actually "science".  Benchmarks were ran on a 2.4GHz 4 core i7 processor.


docs/notebooks/annoytutorial.ipynb
```diff
@@ -145,22 +141,34 @@
...
-      "('in', 0.9971976953092963)\n",
-      "('an', 0.9971216244157404)\n",
-      "('which', 0.9964874279685318)\n",
-      "('has', 0.996418114984408)\n",
-      "CPU times: user 4 ms, sys: 0 ns, total: 4 ms\n",
-      "Wall time: 3.22 ms\n"
+      "Gensim: 0.007451029\n",
+      "Annoy: 0.002149934\n",
+      "\n",
+      "Annoy is 3.46570127269 times faster on average over 1000 random queries\n"
```
piskvorky: The focus and emphasis on such a level of precision is misleading (and unnecessary).

Also, please mention the other factors that affect this number, like index size etc. So people don't go away thinking "annoy is ~3.5x faster than gensim", whereas in reality this is anything between 1x-infinity.

docs/notebooks/annoytutorial.ipynb
```diff
@@ -269,16 +277,16 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "('orbits', 0.6323515474796295)\n",
-      "('changes,', 0.6309329569339752)\n",
-      "('dr.', 0.6305853128433228)\n",
-      "('terrorism,', 0.6300898194313049)\n",
-      "('creditors', 0.6264415979385376)\n"
+      "('signature', 0.5921074748039246)\n",
+      "('\"dangerously', 0.5920691192150116)\n",
```
piskvorky: This looks like bad preprocessing. Any reason not to simply use `simple_preprocess`?

docs/notebooks/annoytutorial.ipynb
droudy: @piskvorky Should I round to a smaller decimal place or leave the exact figure out completely?

piskvorky:
This doesn't look right -- I still see `"dangerously` in the notebook as a token, which should never happen with `simple_preprocess`.

PR #1115, opened 2017-01-29
Update Transformation and Topics link from quick start notebook

mariana393:
"Transformation and Topics" link was changed to redirect to the github page


tmylk:
Thanks for the fix!

PR #1116, opened 2017-01-29
Corpora_and_Vector_Spaces tutorial text fix

lgmoneda:
It's just a text fix to make the tutorial more friendly. It's about the words id given by the built dictionary.

Some places assume it'll be the same and explain the output using fixed ids, that might be confusing. like in:
> The sparse vector [(0, 1), (1, 1)] therefore reads: in the document “Human computer interaction”, the words computer (id 0) and human (id 1) appear once; the other ten dictionary words appear (implicitly) zero times.

I'm suggesting a change to:
> The sparse vector `[(word_id, 1), (word_id, 1)]` therefore reads: in the document *“Human computer interaction”*, the words "computer" and "human", identified by an integer id given by the built dictionary, appear once; the other ten dictionary words appear (implicitly) zero times. Check their id at the dictionary displayed in the previous cell and see that they match.

Since the user can check what was the id given to a word in the previous cell's output, where it prints the dictionary.token2id.

And it has too a little warning about the ids being different from runned notebooks, or the link provided for comparison ([Quick Example](https://radimrehurek.com/gensim/tutorial.html#first-example)), so beginners won't find it strange.

### DIFF
--- a/docs/notebooks/Corpora_and_Vector_Spaces.ipynb
+++ b/docs/notebooks/Corpora_and_Vector_Spaces.ipynb
@@ -213,7 +213,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The function `doc2bow()` simply counts the number of occurrences of each distinct word, converts the word to its integer word id and returns the result as a sparse vector. The sparse vector `[(0, 1), (1, 1)]` therefore reads: in the document *“Human computer interaction”*, the words computer (id 0) and human (id 1) appear once; the other ten dictionary words appear (implicitly) zero times."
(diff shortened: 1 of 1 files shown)


tmylk:
Maybe add a link to HashDictionary for persistent ids?

lgmoneda:
I've read the documentation for hash dictionary, but i think it's not a case of knowing that it's possible to make ids persistent, but warning about the differences a beginner doing the tutorial may experience. 

Looking to a text saying "your output is  [(0, 1), (1, 1)] and that means..." and having a different output is confusing.

Do you say about putting its link in addition to the modifications suggested? I would just worry about too much information, since i believe the main point is saying "results may change, but that's ok".

tmylk:
Agree, thx for the fix.

PR #1118, opened 2017-01-29
Remove Pattern lib dependency in News Classification tutorial

luizcavalcanti:
Even thought NLTK is old and all, Pattern does not currently work on Python3. So changing this tutorial to use NLTK for lemmatization makes it possible to work on both python 2.5+ and 3.x. We are also making sure the user has its NLTK packages downloaded and up-to-date.


tmylk:
Thanks for supporting Python 3.
CC @dsquareindia

PR #1124, opened 2017-01-30
Improve topics and transformations

onilton:
* Improve Available transformations section indentation 

The previous one, bullet point based, looked really bad on python notebook. 
It was hard to follow and understand

* Add relative to link to next tutorial.


tmylk:
Thanks for the improvement!

Issue #1126, opened 2017-01-31
Can't run word2vec2tensor in Python 3


lucianosb:
Hi, I'm trying to run the recently added script for generating tensorboard embeddings. However I get an error all the time

    $ python3 -m gensim.scripts.word2vec2tensor --input ./ptwiki-vectors --output ptwiki4tensor

    2017-01-31 14:16:55,051 : MainThread : INFO : running /usr/local/lib/python3.5/dist-packages/gensim/scripts/word2vec2tensor.py --input ./ptwiki-vectors --output ptwiki4tensor
    2017-01-31 14:16:55,052 : MainThread : INFO : loading projection weights from ./ptwiki-vectors
    2017-01-31 14:19:01,690 : MainThread : INFO : loaded (748533, 300) matrix from ./ptwiki-vectors
    2017-01-31 14:19:01,690 : MainThread : WARNING : direct access to index2word will not be supported in future gensim releases, please use model.wv.index2word
    Traceback (most recent call last):
     File "/usr/lib/python3.5/runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
     File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
     File "/usr/local/lib/python3.5/dist-packages/gensim/scripts/word2vec2tensor.py", line 84, in <module>
    word2vec2tensor(args.input, args.output, args.binary)
     File "/usr/local/lib/python3.5/dist-packages/gensim/scripts/word2vec2tensor.py", line 54, in word2vec2tensor
    file_metadata.write(word.encode('utf-8') + '\n')
    TypeError: can't concat bytes to str

I'm running gensim in Ubuntu 16.04 with Python 3.5. The vectors are from wikipedia in portuguese generated with the help of https://github.com/idio/wiki2vec, which gave me a binary file that I later converted to text with 

    model.save_word2vec_format('ptwiki-vectors')

Have I done anything wrong? Or is it an issue with the script?

tmylk:
Hi, the issue is due to str.encode() working differently in Python 2(in which the script was written) and Python 3.

Can you try `gensim.utils.to_unicode(word)` instead of `word.encode('utf-8')` ?

Also, what does `no attribute __path__` in the title refer to?

lucianosb:
Changed the title. It was a previous error of my own due to a typo that I fixed.

I'll try the suggestion. Will report back shortly.

lucianosb:
[...]

PR #1127, opened 2017-01-31
word2vec2tensor with python 3

lucianosb:
As discussed with @tmylk in issue [1126](https://github.com/RaRe-Technologies/gensim/issues/1126), python 3 does not support `word.encode('utf-8')`. Changing to `gensim.utils.to_unicode(word)` worked. Tested with Python 3.5 in Ubuntu 16.04.


tmylk:
Thank you for the quick fix!

piskvorky:
That's not a good idea, please revert.

Never write unicode strings to output (or concatenate unicode and bytestrings with `+` like this). Encode explicitly to utf8.

tmylk:
Reverted in #1144

PR #1139, opened 2017-02-09
[WIP][DNM] error-resistant train(). Fix #1052

### LINKED ISSUE #1052: Make Word2vec multiple passes more explicit
As discussed on [the mailing list](https://groups.google.com/forum/#!searchin/gensim/fat$20warning%7Csort:relevance/gensim/0GVxA055yOU/5m2jwfdBBAAJ)

```
Perhaps `train()` should require an explicit (non-default) `passes` parameter. In the 1-liner/initialization-trains case, it'd be called with the 'iter' value. For anyone calling it themselves, they'd have to make an explicit choice of 'passes'. And after the change, any old code without the parameter would (by design) break, forcing a change to explicit specification. 

Perhaps even `alpha` and `min_alpha` should be explicit required parameters to `train()`, to ensure those calling it directly aren't see-sawing the values each call. Calling 'train()' directly is kind of an advanced approach, so requiring this level of choice may be appropriate. 

```

gojomo:
Changes `train()` to require explicit `epochs` argument, and explicit size of corpus (either `total_examples` or `total_words`) – thus forcing users to be conscious of `train()`s behavior in these regards, and (intentionally) breaking older code that might be operating under misconceptions. 

Tests updated to be as explicit as needed. 

To do:

- [ ] Update any notebooks using `train()`

Also a potential related change: add an optional `epoch_callback` argument to `train()`, that could be called after each training epoch. That could completely obviate the need for users/tutorials running `train()` in their own loop to do extra steps (such as evaluation) before training is done.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -448,7 +448,8 @@ class Word2Vec(utils.SaveLoad):
             if isinstance(sentences, GeneratorType):
                 raise TypeError("You can't pass a generator as the sentences argument. Try an iterator.")
             self.build_vocab(sentences, trim_rule=trim_rule)
-            self.train(sentences)
+            self.train(sentences, total_examples=self.corpus_count, epochs=self.iter,
+                       start_alpha=self.alpha, end_alpha=self.min_alpha)
 
     def initialize_word_vectors(self):
         self.wv = KeyedVectors()
@@ -729,16 +730,23 @@ class Word2Vec(utils.SaveLoad):
         """Return the number of words in a given job."""
         return sum(len(sentence) for sentence in job)
 
-    def train(self, sentences, total_words=None, word_count=0,
-              total_examples=None, queue_factor=2, report_delay=1.0):
+    def train(self, sentences, total_examples=None, total_words=None,
+              epochs=None, start_alpha=None, end_alpha=None,
+              word_count=0,
+              queue_factor=2, report_delay=1.0):
         """
         Update the model's neural weights from a sequence of sentences (can be a once-only generator stream).
         For Word2Vec, each sentence must be a list of unicode strings. (Subclasses may accept other examples.)
 
-        To support linear learning-rate decay from (initial) alpha to min_alpha, either total_examples
-        (count of sentences) or total_words (count of raw words in sentences) should be provided, unless the
-        sentences are the same as those that were used to initially build the vocabulary.
+        To support linear learning-rate decay from (initial) alpha to min_alpha, and accurate
+        progres-percentage logging, either total_examples (count of sentences) or total_words (count of
+        raw words in sentences) MUST be provided. (If the corpus is the same as was provided to
+        `build_vocab()`, the count of examples in that corpus will be available in the model's
+        `corpus_count` property.)
 
+        To avoid common mistakes around the model's ability to do multiple training passes itself, an
+        explicit `epochs` argument MUST be provided. In the common and recommended case, where `train()`
+        is only called once, the model's cached `iter` value should be supplied as `epochs` value.
         """
         if (self.model_trimmed_post_training):
             raise RuntimeError("Parameters for training were discarded using model_trimmed_post_training method")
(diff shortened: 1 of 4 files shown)


robotcator:
@gojomo  Hello, when I fix the notebooks/docs to reflect the change of "train",  I got an error in the notebooks/doc2vec-IMDB.ipynb:

```
<ipython-input-2-4bb56643bf06> in <module>()
     53 
     54         for txt in txt_files:
---> 55             with open(txt, 'r', encoding='utf-8') as t:
     56                 control_chars = [chr(0x85)]
     57                 t_clean = t.read()

TypeError: 'encoding' is an invalid keyword argument for this function
```
I use Ubuntu 14.04, python 2.7.6 and gensim version 1.0.1, the error was caused by 'encoding' is an invalid keyword argument for the open function in python2.7. Dose it need to change the code for support the python2?

tmylk:
The notebooks should support both Python 2 and 3, same as our source code. Unfortunately we have been lax about it before - fixing it as a part of this release would be a great contribution.

prakhar2b:
@tmylk will python 2 and 3 compatibility library [six](https://pypi.python.org/pypi/six) be a good choice here ?

tmylk:
Yes, gensim already uses six

robotcator:
@prakhar2b would you mind guide me to work through this? I think this is a good opportunity for me to get family with the process of contribution.

prakhar2b:
@robotcator Definitely, that would be a pleasure. I suggest you post your doubts on [gensim google group](https://groups.google.com/forum/#!forum/gensim), the community is very active in responding. You can also mail me directly if you want - er.prakhar2b@gmail.com

tmylk:
Ping @robotcator for a status update

robotcator:
@tmylk sorry for late status update. I have fixed the compatibility between Python 3 and Python 2 by using codecs library. But the code runs in Python 2 seems inefficiently, I am stuck in finding a solution to optimizing python2 code.

tmylk:
Could you please update the pr with the new code?

robotcator:
@tmylk sorry for late reply. I have make an silly mistake on the git. I create an pull request[#1220](https://github.com/RaRe-Technologies/gensim/pull/1220). Since I am not familiar with git and I am curious about why there are so many commits and comments on my pull request.

tmylk:
@robotcator Could you please add tests to make sure that ValueError is indeed thrown when params are not supplied.
notebooks that need to be changed:
https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/doc2vec-lee.ipynb
https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/doc2vec-wikipedia.ipynb
https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/online_w2v_tutorial.ipynb
https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/word2vec.ipynb

robotcator:
@tmylk Of course.

robotcator:
@tmylk I have create [pr 1237](https://github.com/RaRe-Technologies/gensim/pull/1237) to to make sure that ValueError is thrown when params are not supplied. I don't know why there has conflicts with the base branch.

tmylk:
@gojomo Could you please write a note for users on how to upgrade their code for this change? It will be the breaking change to warrant a major release to `gensim 2.0.0`

gensim/models/word2vec.py
```diff
@@ -448,7 +448,8 @@ def __init__(
             if isinstance(sentences, GeneratorType):
                 raise TypeError("You can't pass a generator as the sentences argument. Try an iterator.")
             self.build_vocab(sentences, trim_rule=trim_rule)
-            self.train(sentences)
+            self.train(sentences, total_examples=self.corpus_count, epochs=self.iter,
```
tmylk: Hanging indent please

gojomo:
The note could be:

> Any direct calls to method `train()` of Word2Vec/Doc2Vec now require an explicit `epochs` parameter and explicit estimate of corpus size (usually as `total_examples`). See the [method documentation] for more information. 

I'm in favor of generous version-incrments, but if considering a "2.0.0" release to signal a break in backward compatibility, so soon after the "1.0.0" release, we may want to look at any other prposed/pending/desirable changes in names/defaults/parameters, to include as well.

Some things that come to mind:

* `infer_vector()` defaults & parameter names
* `KeyedVector` property/method names still retain lots of word2vec-specific details – like `word` where the vectors might be more general, or `syn0` when it's no longer necessarily the sort of NN-layer that has that name
* Doc2Vec's DocvecsArray may be mergeable with KeyedVectors, eliminating some code duplication, adding some other flexibility (at cost of complexity) to KeyedVectors, maybe making `model.wv` and `model.dv` perfectly-parallel ways to access word-vectors-by-key and doc-vectors-by-key, respectively.

gojomo:
We could use a new label to mark issues/PRs where breaks with backward-compatibility are happening (or proposed), requiring sync w/ version numbering. I've created one for that purpose, and applied it here to get things started.

tmylk:
Thanks for suggesting more ways to make the API and code more streamlined.

Unfortunately SemVer adoption means that (contrary to topic modelling) ["Version numbers are not for humans"](https://youtu.be/tc2UgG5L7WM?t=697)

Small changes in releases are better than big releases, so if we get to version 10.1.0 by the end of the year then it is ok.

gojomo:
The idea of batching together sets of breaking changes, to help minimize user upgrade efforts, seems orthogonal to SemVer adoption to me. But if you'd like to move to a more rapid tempo, sure. 

That style of numbering/rapid-incremental-release might also justify eventually splitting gensim into narrow single-algorithm subprojects (with some shared core dependencies). Then small changes, not affecting most users, would only trigger a release/version-increment/need-to-review-release-notes among the users of that one module.

tmylk:
Merged in #1237

piskvorky:
@gojomo any suggestions where those "subproject split lines" ought to be?

gojomo:
@piskvorky In the very-small, very-focused packaging style of (say) node/npm, almost every module would be a different package. There'd be one or a few core or util packages, but then nearly every 'model' would be its own package. In the extreme, even Word2Vec and Doc2Vec would be different packages (with Doc2Vec depending on Word2Vec).

Still mulling over whether I'd advocate this – it's less common in the Python world than the JS world, and may result in a bewildering variety of projects each with their own version-number – but it does mesh with rigorous SemVer and rapid, high-resolution release cycles.

PR #1144, opened 2017-02-14
Write UTF-8 byte strings in tensorboard conversion

tmylk:
In reply to https://github.com/RaRe-Technologies/gensim/pull/1127#issuecomment-276879071


tmylk:
The build errors are unrelated to the change and are caused by numpy 1.12. Working on a fix.

PR #980, opened 2016-10-26
KeyedVecs refactoring for word2vec

anmol01gulati:
Adresses #833.

PR #1107, opened 2017-01-24
Move load and save word2vec_format out of word2vec class to KeyedVectors 

tmylk:
Move `save_word2vec_format` and `load_word_vec_format` out of word2vec class and into KeyedVectors.

Benefits of refactoring:
- support of word2vec C format in FastText, WordRank and other word embedding wrappers

Deprecation warnings in word2vec class on save_word2vec_format and load_word_vec_format

`word2vec.load_word2vec_format` still returns a `word2vec` class. This method will soon be removed.

Continuing #1078.

Issue #1069, opened 2017-01-03
Save and load methods generate KeyedVector warnings word2vec and doc2vec


tmylk:
Introduced in https://github.com/RaRe-Technologies/gensim/pull/980
CC @anmol01gulati 
```
model.save(temp_path)
WARNING:gensim.models.word2vec:direct access to syn0norm will not be supported in future gensim releases, please use model.wv.syn0norm
WARNING:gensim.models.word2vec:direct access to syn0norm will not be supported in future gensim releases, please use model.wv.syn0norm
WARNING:gensim.models.word2vec:direct access to syn0norm will not be supported in future gensim releases, please use model.wv.syn0norm
WARNING:gensim.models.word2vec:direct access to syn0norm will not be supported in future gensim releases, please use model.wv.syn0norm

new_model = gensim.models.Word2Vec.load(temp_path)  # open the model
WARNING:gensim.models.word2vec:direct access to syn0norm will not be supported in future gensim releases, please use model.wv.syn0norm
WARNING:gensim.models.word2vec:direct access to index2word will not be supported in future gensim releases, please use model.wv.index2word
```

tmylk:
The issue is partially fixed by #1072
There is still a problem that after introducing KeyedVectors now syn0, vocab, id2word are saved **TWO** times. Once in word2vec and once in keyedvectors. After direct access is deprecated it will be only once.

In the mean time we can add an `ignore_sub` parameter to `utils.save_specials` where it handles recursive saves. That way `syn0norm` and `syn0` will be ignored from word2vec pickle but saved in KeyedVectors pickle.

gojomo:
Rather than silencing the warnings, via a new warning-enablement state/API, shouldn't the save/load code just not do the warned-against actions? (An API where a SaveLoad object self-reports its eligible properties would make more sense than a hide-warnings workaround.)

(Such complications are why my preference, as per comment at https://github.com/RaRe-Technologies/gensim/pull/833#issuecomment-245982648 , was to avoid backward-compatibility tricks by keeping the refactor in a parallel class and/or holding it for an API-breaking major release.)

tmylk:
Current behaviour of SaveLoad API is to save everything except ignored fields. To change it to save what is explicitly given will take a while. For example  even calling 'hasattr' on 'model.vocab' raises a warning. The silencing will be removed in the next API breaking release

gojomo:
[...]

PR #1147, opened 2017-02-14
Remove direct access to properties moved to KeyedVectors

tmylk:
Following up on deprecation warnings added in #980

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -330,9 +330,6 @@ class Word2Vec(utils.SaveLoad):
 
     """
 
-    # TODO: delete this flag after direct access to syn0norm, syn0, vocab is removed
-    keyed_vector_warnings = True
-
     def __init__(
             self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
             max_vocab_size=None, sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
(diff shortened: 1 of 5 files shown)


gojomo:
@tmylk - the class comments still reference `load_word2vec_format` (etc), with now-invalid examples-of-use, and there's still the `call_on_class_only` shim that's now unneeded.

tmylk:
Agree, also the method should throw a "Deprecated, please use KeyedVectors" exception because it confuses a lot of user and it is not clear where did this function go.

piskvorky:
I'm getting lost too. Why not just return what the user asked for, instead of raising an exception?

tmylk:
After removing the direct access now there is only one way left to access the attribute. It makes things simpler and uniform across various embeddings. Do you agree?
There used to be two ways. We raised deprecation warnings for 2 months on one of the ways to access it. Now it became a deprecation exception.

piskvorky:
It's fine if the "recommended way" changes to something else, no problem with that.

But raising exceptions, breaking backward compatibility, where we could actually serve the request no problem, sounds strange. I don't remember the reasoning any more, but it looks plain malicious to me. Why would we do that?

tmylk:
Sorry, I don't understand the question. One way to access an attribute is better than two ways. The reason we added KeyedVectors class is described in the release notes. After adding KeyedVectors there were two ways to do things so we needed to deprecate one of them.

gojomo:
As I recall the reasoning:

There would be somewhat of a surprise/violating-expectations problem if `Word2Vec.load_word2vec_format()` still worked, but now returned a new, different type (`KeyedVectors`). Maybe that'd still be OK.

If it still returned a `Word2Vec` instance that was only-good-for-lookups (and thus really only functionally-equivalent to `KeyedVectors`), that'd have maintained expectations, but also continued the somewhat messy conflation of 'improted vectors' and 'trainable model' that the refactoring was meant to address. So that path was appropriate for a while – and maybe the deprecation period should have been longer – but at some point the eventual change would still be "stop calling this through Word2Vec, because what you're loading isn't necessarily logically linked with the internals of this class".

piskvorky:
If I understand correctly, continuing to support attribute access like `model.syn0` costs us nothing (we just alias a few variables, we have all the necessary information).

On the other hand, not supporting it costs a lot to the users, who are confused about this (seemingly) useless obfuscation of `model.wv.syn0`.

Weighing the costs/benefits, I don't think removing the support is a good choice.

tmylk:
The cost of cleanly supporting both `model.syn0` and `model.wv.syn0` would have been a change to Save/Load API to make `ignore` a class variable as [suggested](https://github.com/RaRe-Technologies/gensim/issues/1069#issuecomment-271503551) by @jayantj . 

Reasons why this change is required:

1) In 0.13.4.1 `syn0` and others were saved [twice](https://github.com/RaRe-Technologies/gensim/issues/1069#issuecomment-270271397) - once in `word2vec` `save()` and another time in `KeyedVectors` `save_specials`. Ignoring it in `word2vec` and not `keyedvectors` is impossible because `ignore` is not a class variable.

2) The warnings in `save` for `syn0norm` direct access had to be silenced in a not so pretty way in #1072 

Another minor cost is that bugs like #1173 would be harder to catch where "the vectors are not loaded into the syn0 attribute of the KeyedVector instance, and are instead loaded into the syn0 attribute of the Word2Vec instance".

I avoided the change to `Save/Load` API because it affects *all the models* and thus a high risk. 

The deprecation seemed a better choice because it only affects word2vec and the users were warned with Warnings for 2 months. It was a learning experience in how much attention users pay to warnings. And they pay less attention than I wish they did. :) In hindsight, in this release instead of a complete code path deprecation, the warning should have become an Exception suggesting to use `wv.syn0`,  like  #1107 for `load_word2vec_format`. Removing the Exception code path in the next release would not have required a major version bump because the working API remained the same.

There are also advantages of deprecation as @gojomo wrote above: better API for a minimal `read-only` model and uniform code between word2vec for other embeddings.

PR #1152, opened 2017-02-17
Fix hdpmodel constructor docstring for print_topics

toliwa:
Fix hdpmodel constructor docstring for the
print_topics parameter. It should be 'num_topics'
to specify the number of topics.


tmylk:
Thanks for the fix!

PR #1154, opened 2017-02-18
Fix lda transform prefomance

menshikh-iv:
I founded one terrible performance issue connected with apply LDA model

Typical example 

```python
# *_PATH - path to file with model
# tokens - list with tokens

from gensim.models import LdaModel

dct_model = Dictionary.load(DICT_PATH)
lda_model = LdaModel.load(LDA_PATH)

d2b_vector = dct_model.doc2bow(tokens)

print(lda_model[d2b_vector])
```
This triggers a chain of calls

[\_\_getitem__(self, bow, eps=None)](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ldamodel.py#L966)

[self.get_document_topics(bow, eps, self.minimum_phi_value, self.per_word_topics)](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ldamodel.py#L974)

[gamma, phis = self.inference([bow], collect_sstats=True)](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ldamodel.py#L915)

`collect_sstats=True` initiates heavy computation for `sstats`, but `sstats` (`phis`) can't used if `per_word_topics=False` [(Proof block)](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ldamodel.py#L915-L922)

I replaced `True` flag to `per_word_topics` for significant speedup `get_document_topics` for case `per_word_topics=False` (this effect is clearly visible if you use LdaModel with ~180 topics and 700k+ dictionary, i.e. huge topic matrix)


tmylk:
Thanks for the fix. The docstrings in `__init__` and `get_item` need to be updated to include the `per_word_topics` parameter.

Issue #1082, opened 2017-01-08
RandomState (Fix to issue #113) breaks backwards compatibility with old LDA models 


mattilyra:
LDA models saved before version 0.13.2 can not be used in version 0.13.2 and up because they do not contain `random_state` variable. This should be a pretty trivial fix in `load` functionality though.

The breaking PR is this one
https://github.com/RaRe-Technologies/gensim/commit/2e0ed26c03bfa997c8748d03fafe313a05cd9741

ibrahimsharaf:
I am interested in working on this bug, how can I start, what does "barking pull request" mean?

tmylk:
@ibrahimsharaf The pull request that introduced this error is https://github.com/RaRe-Technologies/gensim/commit/2e0ed26c03bfa997c8748d03fafe313a05cd9741

ibrahimsharaf:
should we use `        state_fname = utils.smart_extension(fname, '.random_state')` instead of `        state_fname = utils.smart_extension(fname, '.state')` in the load function?

tmylk:
@ibrahimsharaf That line is used to load [saved state](https://github.com/RaRe-Technologies/gensim/blob/cebee95b19ce8950585dba8e9338472d06f1698e/gensim/models/ldamodel.py#L1004). It is different from random_state in this issue

ibrahimsharaf:
@tmylk can you review my pull request? https://github.com/RaRe-Technologies/gensim/pull/1155

tmylk:
@ibrahimsharaf It is strange that it can't be reproduced - trying to use an unpickled object that doesn't have a required field should cause an error. what is your code to reproduce?

ibrahimsharaf:
here's it (http://codepad.org/EvBkXSgm), I am using PyCharm, changing Gensim versions via project interpreters (I created a simple LDA model in 0.12.0 version then saved it, loaded it in 0.13.3 version at it worked fine!).

tmylk:
Current status: awaiting a code fix

chinmayapancholi13:
@tmylk @menshikh-iv On running the code below :

```python
from gensim import corpora, models

#save with version <=0.13.1
# texts = [['human', 'interface', 'computer'],
#  ['survey', 'user', 'computer', 'system', 'response', 'time'],
#  ['eps', 'user', 'interface', 'system'],
#  ['system', 'human', 'system', 'eps'],
#  ['user', 'response', 'time'],
#  ['trees'],
[...]

PR #1155, opened 2017-02-18
 RandomState (Fix to issue #113) breaks backwards compatibility with old LDA models

ibrahimsharaf:
LDA models saved before version 0.13.2 can not be used in version 0.13.2 and up because they do not contain random_state variable. (https://github.com/RaRe-Technologies/gensim/issues/1082).

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -1049,11 +1049,10 @@ class LdaModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
         result = super(LdaModel, cls).load(fname, *args, **kwargs)
 
         random_state_fname = utils.smart_extension(fname, '.random_state')
-        random_state_val = utils.get_random_state(random_state_fname)
         try:
-            result.random_state = super(LdaModel, cls).load(random_state_val, *args, **kwargs)
+            result.random_state = super(LdaModel, cls).load(random_state_fname, *args, **kwargs)
         except Exception as e:
-            logging.warning("failed to load random_state from %s: %s", random_state_val, e)
+            logging.warning("failed to load random_state from %s: %s", random_state_fname, e)
         state_fname = utils.smart_extension(fname, '.state')
         try:
             result.state = super(LdaModel, cls).load(state_fname, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


gensim/models/ldamodel.py
```diff
@@ -1047,6 +1047,12 @@ def load(cls, fname, *args, **kwargs):
         """
         kwargs['mmap'] = kwargs.get('mmap', None)
         result = super(LdaModel, cls).load(fname, *args, **kwargs)
+
+        random_state_fname = utils.smart_extension(fname, '.random_state')
+        try:
+            result.random_state = super(LdaModel, cls).load(random_state_fname, *args, **kwargs)
+        except Exception as e:
+            logging.warning("failed to load random_state from %s: %s", random_state_fname, e)
```
tmylk: Have you tested it with 0.13.1 version? 
the fix is to assign `get_random_state(None)` if it is missing on disk

piskvorky:
Why should there be an extra file on disk? I don't understand the solution. Please explain what is happening and why (in the PR description).

tmylk:
Ping @ibrahimsharaf

ibrahimsharaf:
@tmylk I added this snippet
 ```
random_state_fname = utils.smart_extension(fname, '.random_state')

if (os.path.isfile(random_state_fname)):
             result.random_state = super(LdaModel, cls).load(random_state_fname, *args, **kwargs)
else:
            result.random_state = utils.get_random_state(None)
```

I tried reverting back to an older version of Gensim for testing but didn't see how to in the installation docs.

tmylk:
`pip install gensim==0.12.0` to install an earlier version

ibrahimsharaf:
I created a simple LDA model in 0.12.0 version then saved it, loaded it in 0.13.3 version at it worked fine!, I couldn't reproduce the issue, here's the script I used (http://codepad.org/EvBkXSgm), I am using PyCharm, changing Gensim versions via project interpreters.

tmylk:
We need to reproduce the issue before making a fix.
Are you suggesting that the issue should be closed as not reproduced?

ibrahimsharaf:
I tried with the mentioned above versions and I couldn't reproduce the issue, maybe the issue reporter can give us some thoughts, before closing the issue?

piskvorky:
-1: the file shenanigans are not explained. How/where are these files created? I don't think we ever want that.

tmylk:
Closing as this code is not related to the issue

PR #1164, opened 2017-02-23
Dictionary Coloring

bhargavvader:
This is an update for the word-colouring notebook - I've had some requests about how to statically color a dictionary, and have demonstrated the same in the notebook.


tmylk:
Could you please clarify  what is the reason for the incorrect coloring of 'mud'? Lack of context in `get_term_topics` or in the corpus?

bhargavvader:
I reckon it's because of the small vocabulary the lda model isn't trained perfectly; will investigate and see the exact reason.

tmylk:
could you please remove/correct the text about it? will merge after it's fixed

bhargavvader:
Sorry, been busy last couple of days. Would you just like the text to be removed? I thought I could spend some time in fixing it by fine-tuning the model.

tmylk:
Just need to be explicit in that the coloring is correct according to the model. But the model is wrong because of small corpus.

bhargavvader:
Sorry about the delay - I've added a small explanation now.

bhargavvader:
@tmylk unrelated tests failing.

tmylk:
Travis tests re-ran after smart_open update

tmylk:
Thx for the update!

PR #1165, opened 2017-02-24
Generate Deprecated exception when using Word2Vec.load_word2vec_format

tmylk:
Users have been thrown off by the Word2Vec.load_word2vec_format method disappearing without an obvious alternative. An Exception is now thrown directing to KeyedVectors.

Also docstrings and ipynbs updated with KeyedVectors changes.


gensim/models/doc2vec.py
```diff
@@ -21,17 +21,21 @@
 >>> model.save(fname)
 >>> model = Doc2Vec.load(fname)  # you can continue training with the loaded model!
 
-The word vectors in the model can also be instantiated from an existing file on disk in the word2vec C format:
-NOTE: document vectors are not loaded/saved with .load/save_word2vec_format(). Use .save()/.load() instead.
+If you're finished training a model (=no more updates, only querying), you can do
+
+  >>> model.delete_temporary_training_data(keep_doctags_vectors=True, keep_inference=True):
```
gojomo: I believe this will also break inference, so comment should mention that too.

gensim/models/doc2vec.py
```diff
@@ -21,17 +21,21 @@
...
 
-  >>> model = Doc2Vec.load_word2vec_format('/tmp/vectors.txt', binary=False)  # C text format
-  >>> model = Doc2Vec.load_word2vec_format('/tmp/vectors.bin', binary=True)  # C binary format
 
 .. [1] Quoc Le and Tomas Mikolov. Distributed Representations of Sentences and Documents. http://arxiv.org/pdf/1405.4053v2.pdf
 .. [2] Tomas Mikolov, Kai Chen, Greg Corrado, and Jeffrey Dean. Efficient Estimation of Word Representations in Vector Space. In Proceedings of Workshop at ICLR, 2013.
 .. [3] Tomas Mikolov, Ilya Sutskever, Kai Chen, Greg Corrado, and Jeffrey Dean. Distributed Representations of Words and Phrases and their Compositionality.
        In Proceedings of NIPS, 2013.
 .. [blog] Optimizing word2vec in gensim, http://radimrehurek.com/2013/09/word2vec-in-python-part-two-optimizing/
+.. [tutorial] Doc2vec in gensim tutorial, http://radimrehurek.com/2013/09/word2vec-in-python-part-two-optimizing/
```
gojomo: Wrong link.

gensim/models/word2vec.py
```diff
@@ -26,39 +28,64 @@
...
 
-The model can also be instantiated from an existing file on disk in the word2vec C format::
+The word vectors are stored in a KeyedVectors instance in model.wv. This separates the read-only word vector lookup operations in KeyedVectors from the training code in Word2Vec.
+
+  >>> model.wv['computer']  # numpy vector of a word
+  array([-0.00449447, -0.00310097,  0.02421786, ...], dtype=float32)
+
+The word vectors can also be instantiated from an existing file on disk in the word2vec C format as a KeyedVectors instance::
+
+NOTE: It is impossible to continue training the vectors loaded from the C format because the binary tree is missing.
```
gojomo: Not just the binary tree (which is only used in `hs` mode), but the hidden-weights and vocabulary-frequency information are missing.

gensim/models/word2vec.py
```diff
@@ -26,39 +28,64 @@
...
+
+  >>> model.wv.accuracy(os.path.join(module_path, 'test_data', 'questions-words.txt'))
 
 and so on.
 
+
 If you're finished training a model (=no more updates, only querying), you can do
 
-  >>> model.init_sims(replace=True)
+  >>> model.delete_temporary_training_data(replace_word_vectors_with_normalized=True)
```
gojomo: With KeyedVectors now the recommended form for read-only access, perhaps the proper recommendation for "if you're sure you're done training" is to discard the Word2Vec model instance entirely, and just retain the KeyedVectors.

gensim/models/word2vec.py
```diff
@@ -26,39 +28,64 @@
...
 
 to trim unneeded model memory = use (much) less RAM.
 
 Note that there is a :mod:`gensim.models.phrases` module which lets you automatically
 detect phrases longer than one word. Using phrases, you can learn a word2vec model
 where "words" are actually multiword expressions, such as `new_york_times` or `financial_crisis`:
 
->>> bigram_transformer = gensim.models.Phrases(sentences)
+>>> bigram_transformer = gensim.models.Phraser(gensim.models.Phrases(sentences))
```
gojomo: Personally I might not recommend all users prefer `Phraser` without understanding the extra steps it requires, because of the extra time of the reduction-pass, and the fact it throws out some info (in `Phrases`) that was expensive to collect and allow experimentation with different count/threshold values.

gensim/models/word2vec.py
```diff
@@ -1272,6 +1299,17 @@ def _load_specials(self, *args, **kwargs):
             self.wv = wv
         super(Word2Vec, self)._load_specials(*args, **kwargs)
 
+    @classmethod
+    def load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8', unicode_errors='strict',
```
gojomo: This now makes the `call_on_class_only` reference in `__init__()` superfluous/wrong.

gensim/models/word2vec.py
piskvorky: Of course. This is some weird mix, "the worst of both world", complicating the API and confusing people.

gensim/models/doc2vec.py
piskvorky: Isn't that what the `keep_inference=True` is for?

gensim/models/word2vec.py
tmylk: The difference between KeyedVectors and a word2vec model post delete_temporary_training_data is in the `score` function from DeepIR. Added that to docstring.

gensim/models/doc2vec.py
tmylk: Inference is preserved. It is tested in https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_doc2vec.py#L319

gojomo: I see, though this is still kind of odd. As called in this prominent example, this method hardly gets rid of anything – just the relatively-tiny `doctag_syn0_lockf`. Someone who just needs that tiny benefit could be coached to execute `del model.docvecs.doctag_syn0_lockf`. (I fear here, and to some extend on Word2Vec too, this method is attractive to novices but likely to cause headaches for them and then support/maintenance issues down the road.)

Issue #814, opened 2016-08-05
Loading fastText binary output to gensim like word2vec


phunterlau:
Facebook's recent open sourced `fasttext` https://github.com/facebookresearch/fastText improves the `word2vec` SkipGram model. It follows a similar output format for `word` - `vector` key value pairs, and the similarity calculation is about the same too, but their binary output format is kind of different from that of the C version of `word2vec` binary format. Do we want to support loading `fastText` model output in `gensim`? Thanks.

gojomo:
Definitely! Reading/writing their word-vector format (the `.vec` and perhaps part of the `.bin`) would be an obvious 1st step. 

(As a later step, the `.bin` might include enough extra info for models to continue training… though their classification modes might not map directly to the existing gensim output-layer models.)

gojomo:
It appears the `.vec` output of fastText is already compatible with the original word2vec.c text format, and readable in gensim by `load_word2vec_format(filename, binary=False)`. 

The `.bin` output, written in parallel (rather than as an alternative format like in word2vec.c), seems to have extra info  – such as the vectors for char-ngrams – that wouldn't map directly into gensim models unless/until they're extended with new features. So supporting the load of such info isn't a mere matter of format understanding/translation.

phunterlau:
@gojomo thanks. I can confirm the vec format is compatible with gensim

tmylk:
See FastText comparison notebook in https://github.com/RaRe-Technologies/gensim/pull/815

gojomo:
@tmylk - we may want to keep this open for the larger issue of doing something with the `.bin` output. We might be able to map its weights (and word-frequency info) into gensim's objects, to support continued training, as a small translation-of-values patch. Loading the buckets-of-subword-vectors, and making them usable for OOV prediction, would require a bit more actual functionality... but would still be practical. Maybe at first, the subword-buckets wind up in a different class – even perhaps a KeyedVectors variant/sibling - which would offer both subword vector lookup, by hashed key, and word-vector-reconstruction (incl. OOV words), by composition of subword vectors. (cc @droudy)

tmylk:
Implemented in https://github.com/RaRe-Technologies/gensim/blob/2a70e3a726404cd4230542a35cfd2dc4d63da6f1/gensim/models/wrappers/fasttext.py#L246

AritzBi:
[...]

PR #1176, opened 2017-03-01
[MRG] Fix for #814

jayantj:
The current code attempts to decode all characters for vocab words individually as ascii.
Instead, the code is changed to do the following - characters are read as raw bytes first, then decoded as utf-8 once the entire word has been read.

Also includes a test fasttext model with non-ascii vocab words, and checks vector lookup for the same.

Haven't been able to test the newly released FastText vectors for wiki in various languages since the models are large-ish (~10 GB). 

Related issue reported [here](https://github.com/RaRe-Technologies/gensim/issues/814#issuecomment-283381412)


gensim/test/test_data/non_ascii_fasttext.vec
```diff
@@ -0,0 +1,172 @@
...
+byl -0.89449 1.4131 
+tu -1.1255 1.5505 
+tedy -1.1693 1.6446 
+teto -1.2134 1.546 
+bylo -0.86091 1.3805 
+kde -1.3468 1.7507 
+ke -1.0699 1.6688 
+pravé -0.9391 1.5172 
+nad -1.3404 1.7661 
+nejsou -0.85023 1.5033
```
anmol01gulati: Shouldn't this file be in a binary format in the package and later converted to text? This also works, but just feel that would be better.

anmol01gulati:
@jayantj This looks good, but should we wait for to test this on the new pre-trained models first and then ask Lev to merge? Though I feel, it shouldn't create any difference, as you have already added a test for non-ascii words.

tmylk:
Agree with @anmol01gulati. Does it work on https://github.com/facebookresearch/fastText/blob/master/pretrained-vectors.md, @jayantj  ?

piskvorky:
Does fastText always output utf8? Is the encoding really fixed?

piskvorky:
@jayantj @tmylk  the fastText file header is incorrect (states I wrote the code in 2013). Please fix.

piskvorky:
@tmylk use better issue/PR titles, so search engines can index them and users find them.

piskvorky:
@jayantj status on the questions above?

jayantj:
The fastText readme says that the training file should contain utf-8 encoded text. That was the rationale behind choosing `utf8` as the encoding while deserializing the model.

Looking at the fastText code and experimenting with different encodings though, fastText doesn't really decode the bytes in the training file and simply treats the input file as a stream of bytes.

The only place it does make use of the values inside the bytes is while tokenizing the text, where it checks each byte against whitespace characters (`' ', '\n', '\r', '\t', '\v', '\f'`)

So, encodings which have the same values for these characters as ascii would produce sane output. (`utf16` and `utf32` don't, lots of others do). Original encoding information isn't stored anywhere though.

I can think of two ways to handle this - 

1. Assume words in input file are utf8 encoded (since FastText readme mentions it, and it might be enforced in the future), and raise a more friendly error when the input isn't valid utf8
2. Provide an parameter `encoding` allowing the user to specify the input file encoding (defaulting to `utf8`, of course)

jayantj:
Re: file header, I'll create a single PR with the updated header and the fix for the above issue, whichever one we go with.

tmylk:
Option 2 sounds as the most flexible

PR #1177, opened 2017-03-02
most_similar_cosmul bug fix

dkim010:
remove `unhashable type: 'numpy.ndarray'` error

Signed-off-by: Dongwon Kim <dongwon.kim@navercorp.com>

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -455,13 +455,12 @@ class KeyedVectors(utils.SaveLoad):
             # allow calls like most_similar_cosmul('dog'), as a shorthand for most_similar_cosmul(['dog'])
             positive = [positive]
 
+        all_words = set([self.vocab[word].index for word in positive+negative if word in self.vocab])
 
         positive = [self.word_vec(word, use_norm=True) for word in positive]
         negative = [self.word_vec(word, use_norm=True) for word in negative]
         if not positive:
             raise ValueError("cannot compute similarity with no input")
-
-        all_words = set([self.vocab[word].index for word in positive+negative if word in self.vocab])
 
         # equation (4) of Levy & Goldberg "Linguistic Regularities...",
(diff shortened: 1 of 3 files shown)


tmylk:
Thank you for the quick fix!

`model.most_similar_cosmul('graph', topn=10)` works but `model.most_similar_cosmul(positive=[graph_vector], topn=11)` still produces results `TypeError: unhashable type: 'numpy.ndarray'`.

Please add this unit test to `test_word2vec.py`
```
def test_cosmul():
        model = word2vec.Word2Vec(sentences, size=2, min_count=1, hs=1, negative=0)
        sims = model.most_similar_cosmul('graph', topn=10)
        # self.assertTrue(sims[0][0] == 'trees', sims)  # most similar

        # test querying for "most similar" by vector
        graph_vector = model.wv.syn0norm[model.wv.vocab['graph'].index]
        sims2 = model.most_similar_cosmul(positive=[graph_vector], topn=11)
        sims2 = [(w, sim) for w, sim in sims2 if w != 'graph']  # ignore 'graph' itself
        self.assertEqual(sims, sims2)
```

dkim010:
Thank you for the quick reply!

I think that `most_similar_cosmul()` method allows only the string (or a string list) as the positive/negative parameters.
The method is related to the `most_similar()` method and both methods cannot pass the vectors of the words.
Is my idea wrong?

tmylk:
`most_similar` works with vectors and `most_similar_cosmul` is expected to work too https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_word2vec.py#L360

dkim010:
Okay, I'll fix it.

dkim010:
I fixed it. `most_similar_cosmul` works with vectors!
I refered `most_similar`.

tmylk:
Thanks for the quick fix!

Issue #703, opened 2016-05-18
Error details in wrappers.


bhargavvader:
Since all of the wrappers use `check_output` to run on command line, if an error does occur, there is no information apart from `subprocess.CalledProcessError` error message. The user must manually run the command on shell again to figure out what's wrong in these cases.

There should ideally be a way for the user to know what the exact nature of the error is, or at least be given instructions on how to figure out what the error is.

piskvorky:
Good idea!

bhargavvader:
Would like to add here that having a way to print console log/messages while running wrappers would be very useful.

piskvorky:
Yes... especially if we can capture that output and direct it to log.

We definitely don't want to clobber our stdout with the subprocess's stdout, unless the user explicitly requests this. IMO logging is the right place for this output.

kirit93:
To get a verbose error on the command line as you would had the command been run on the shell, the `subprocess.check_output` function does the trick if used as follows : 

```
try:
	subprocess.check_output(cmd, shell=True)
except subprocess.CalledProcessError as outerror:
	print(outerror)
```

Any particular reason a gensim `check_output` is being used instead of the `subprocess` method?

bhargavvader:
@piskvorky , what's the reason we don't just use the `subprocess` `check_output`? Or if there was a way we could pass the `shell` flag in the current function, this would be a very useful fix.

tmylk:
The reason for custom check_output was support for Python 2.6 which is now removed. 
We can use the  `subprocess.check_output` now

menshikh-iv:
Will be fixed in #1501

PR #1182, opened 2017-03-04
Added function to replace gensim check_output with subprocess.check_output

kirit93:
@tmylk I've added a function above the existing ```check_output``` function in an attempt to solve this issue. Please let me know if there's anything I need to change. This is with reference to issue #703


tmylk:
@kirit93 Thanks for the PR. 
How do we know that the output is printed to the log? Please add a test

Could you please replace the actual `check_output` and run the tests for mallet, dtm, fasttext and wordrank wrappers. 

There was some reason I added extra `KeyboardInterrupt` handling to the code - need to look more in order to find what was the issue now. Please check that behaviour is preserved.

kirit93:
@tmylk I ran the tests for mallet, dtm, fasttext and wordrank wrappers. All pass except for fasttext, but the error I get with my ```check_output``` function is the same I get when I use the existing gensim ```check_output``` function. So I'm not sure why that is failing, any idea?
This is what is shown on the log -
```
FAIL: testSimilarity (__main__.TestFastText)
Test similarity for in-vocab and out-of-vocab words
  File "test_fasttext_wrapper.py", line 139, in testSimilarity
    self.assertEqual(self.test_model.similarity('night', 'nights'), self.test_model.similarity('nights', 'night'))
AssertionError: 0.3508754735782928 != 0.35087547357829274

Ran 15 tests in 9.282s

FAILED (failures=1)
```

I'll get started with the rest of the things you mentioned.

piskvorky:
Looks like a bad test -- floats should never be compared for bit equality. Instead there's `np.allclose()` etc.

kirit93:
Should I fix those tests?
Currently the function uses : 
```
self.assertEqual(self.test_model.similarity('night', 'nights'), self.test_model.similarity('nights', 'night'))
```
I'll change it to : 
```
self.assertTrue(numpy.allclose(self.test_model.similarity('night', 'nights'), self.test_model.similarity('nights', 'night')))
```
I made the change and ran the test and there are no failures, the assertEqual is used in a couple of other places in this test. If this fix is okay, I'll make the changes everywhere.

tmylk:
@kirit93 Confirm that using `allclose` is the right way to go. 

Does the error output get printed to the log with this new change?

kirit93:
@tmylk ```numpy.testing.assert_allclose()``` seems to be a better option for float comparisons. It prints a verbose error message to the log unlike ```numpy.allclose```. It uses a default relative tolerance of 1e-7 which can be set to 0 if desired.

gensim/utils.py
```diff
@@ -1146,27 +1146,31 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
-                cmd = popenargs[0]
-            error = subprocess.CalledProcessError(retcode, cmd)
-            error.output = output
-            raise error
-        return output
+        res = subprocess.check_output(args, shell=True)
+        return res
     except KeyboardInterrupt:
-        process.terminate()
+        raise
```
piskvorky: It makes sure the catch-all `except:` below doesn't apply to `KeyboardInterrupt`.

But it's not a good practice anyway, because there's also `SystemExit` etc. At least limit the catch-all to `Exception`.

gensim/utils.py
```diff
@@ -1176,4 +1180,4 @@ def sample_dict(d, n=10, use_random=True):
      according to natural dict iteration.
      """
      selected_keys = random.sample(list(d), min(len(d), n)) if use_random else itertools.islice(iterkeys(d), n)
-     return [(key, d[key]) for key in selected_keys]
+     return [(key, d[key]) for key in selected_keys]
```
piskvorky: Unwanted change.

gensim/utils.py
```diff
@@ -1146,27 +1146,31 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
-            raise error
-        return output
+        res = subprocess.check_output(args, shell=True)
+        return res
     except KeyboardInterrupt:
-        process.terminate()
+        raise
+    except:
+        error = "Error in check_output while trying to execute: \n ' " + str(args) + " '\nthis command was not found"
+        print(error)
```
piskvorky: The original exception was more useful, because it included stdout/stderr, useful for extra debugging.

Is there any way to do the same here?

gensim/utils.py
```diff
@@ -1146,27 +1146,31 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
+    custom error message in the except block. This allows the user to receive an accurate error message if subprocess
+    fails to execute the specified command.
+    If flag is set to true, subprocess.check_output takes 'args' as a string instead of a list. To abstract the user from this,
+    this function will convert the argument list to a string if needed.
+    >>> test_checkoutput(args=['/usr/bin/python', '--version'])
+    Python 2.7.10
+    In case args generates an error
+    >>> test_checkoutput(args=['/usr/bin/pythons', '-ve']) #Incorrect argument
+    /bin/sh: /usr/bin/pythons: No such file or directory
+    *Error in args : /usr/bin/pythons -ve
```
piskvorky: Is this something that is printed? Or logged?

In general, we don't want to pollute stdout at all, unless the user explicitly requested printing.

gensim/utils.py
kirit93: Should I change it to catch all ```Exception``` and then raise the error?

gensim/utils.py
```diff
@@ -1146,27 +1146,31 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
-            error.output = output
-            raise error
-        return output
+        res = subprocess.check_output(args, shell=True)
+        return res
     except KeyboardInterrupt:
-        process.terminate()
+        raise
+    except:
+        error = "Error in check_output while trying to execute: \n ' " + str(args) + " '\nthis command was not found"
```
kirit93: Yes, I overlooked that detail. Should I change the error message to something more generic like " xyz command could not be executed "?

gensim/utils.py
kirit93: Right now the error message gets printed but I will change that so it outputs to log instead.

gensim/utils.py
kirit93: I will make this change.

kirit93: I'm sorry, I didn't fully understand what you mean here. Which original exception are you talking about?

gensim/utils.py
kirit93: Would it be a good idea to catch the ```subprocess.CalledProcessError``` and output an error message that is easy for the user to understand? This is the error raised when ```check_output``` fails.
For any other error including ```KeyboardInterrupt``` I can use a ```catch-all Exception``` and raise it.

gensim/utils.py
piskvorky: I meant `subprocess.CalledProcessError`, a few lines above.

I don't have any strong opinion on what this function should do on failure, or how to capture the subprocess stdout/stderr. Except for sure we don't want to be printing anything -- logging preferred.

gensim/utils.py
```diff
@@ -1146,27 +1146,38 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
-        process.terminate()
+        res = subprocess.check_output(args, shell=flag)
+        return res
+    except subprocess.CalledProcessError:
+        """
+        If this error is raised, it is because check_output could not execute the command.
+        Instead of raising the error, output a more specific error message
+        """
+        error = "subprocess.check_output could not execute command ' " + str(args) + " '"
+        print(error, file=sys.stderr)
```
tmylk: Please log instead of printing, see examples in the code

gensim/utils.py
```diff
@@ -1146,27 +1146,38 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
+        return error
+    except Exception:
```
tmylk: this code is redundant. It will be raised anyway

gensim/test/test_utils.py
```diff
@@ -90,9 +90,12 @@ def test_sample_dict(self):
         self.assertEqual(sampled_dict,expected_dict)
         sampled_dict_random = utils.sample_dict(d,2)
         if sampled_dict_random in expected_dict_random:
-            self.assertTrue(True)    
-
+            self.assertTrue(True)
 
+class TestOutput(unittest.TestCase):
+    def test_check_output(self):
```
tmylk: Please add a test that exception is raised

gensim/test/test_utils.py
```diff
@@ -97,6 +97,10 @@ def test_check_output(self):
         res = utils.check_output(args=["echo", "hello"])
         self.assertEqual(res, b'hello\n')
 
+    def test_check_output_exception(self):
+        error = utils.check_output(args=['ldfs'])
+        self.assertEqual(error, "subprocess.check_output could not execute command ' ldfs '")
```
kirit93: ```test_utils.pyt``` imports ```utils.py``` from the installed ```gensim```. So this test case is still running the old version of ```check_output``` and therefore the exception test fails. However, if you change the import to use the current version of ```utils.check_output```, all tests pass.

tmylk: you can install your new version of gensim with `python setup.py install`

tmylk:
Please add unit test that exception is indeed raised.

gensim/utils.py
tmylk: Please output to log instead of printing.

gensim/utils.py
```diff
@@ -1146,28 +1146,34 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
+        res = subprocess.check_output(args, shell=flag)
+        return res
+    except subprocess.CalledProcessError:
+        """
+        If this error is raised, it is because check_output could not execute the command.
+        Instead of raising the error, output a more specific error message
+        """
+        error = "subprocess.check_output could not execute command ' " + str(args) + " '"
+        logger.error(error)
+        return error
```
tmylk: what is the reason to `return` instead of previous `raise`?

kirit93: The reason I've chosen to simple ```return``` the error message is because ```raise``` results in the following large and not user friendly message. 
```
  File "test.py", line 51, in test_check_out
    res = subprocess.check_output(args, shell=flag)
  File "/Users/kirit/anaconda/lib/python3.5/subprocess.py", line 626, in check_output
    **kwargs).stdout
  File "/Users/kirit/anaconda/lib/python3.5/subprocess.py", line 708, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command 'ldfs' returned non-zero exit status 127

  File "test.py", line 76, in <module>
    test_check_out(args=["ldfs"])
  File "test.py", line 60, in test_check_out
    raise error
TypeError: exceptions must derive from BaseException
```

piskvorky: Raising a string (instead of some exception) is definitely not a good idea.

kirit93: How would we benefit from raising an exception here? Once the ```subprocess.CalledProcessError``` occurs and is caught, an appropriate message is returned to the user indicating why the ```check_output``` command failed. Wouldn't that be enough for the user to proceed correctly?

piskvorky: `raise error` already raises an exception. It's just that proper exceptions inherit from `Exception` (or appropriate subclasses), rather than being just strings.

kirit93: I could do a ```raise BaseException(error)```. 
Would that be okay?

piskvorky: I haven't looked into the actual logic. Just commenting on the discussion above, where you say `The reason I've chosen to simple return the error message is because raise results in the following large and not user friendly message.`, but that message is only due to raising a string :)

piskvorky: My 2c would be to use the same logic that the function did previously, before your changes. The least surprise to users.

kirit93:
@tmylk @piskvorky I've modified the function to handle errors just as the earlier function would. I hope this is okay. Please let me know if there are any further changes to be made so that the PR can be merged. Thanks!

gensim/utils.py
```diff
@@ -1146,27 +1146,33 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
+        res = subprocess.check_output(args, shell=flag)
+        return res
+    except subprocess.CalledProcessError:
+        """
+        If this error is raised, it is because check_output could not execute the command.
+        Instead of raising the error, output a more specific error message
+        """
+        error = "subprocess.check_output could not execute command ' " + str(args) + " '"
+        logger.error(error)
         raise
```
tmylk: Please raise the CalledProcessError exception as before.

tmylk:
It seems that tests didn't run for the last commit, `test_check_output_exception` is expected to fail now. Please remove that test.

Is it correct that the only difference between old code and this improvement is that the error is printed to log here:
```
error = "subprocess.check_output could not execute command ' " + str(args) + " '"
logger.error(error)
```

However this doesn't answer the original purpose of the issue in #703 : 

> There should ideally be a way for the user to know what the exact nature of the error is, 

Please add code and tests that actually do that.

gensim/utils.py
kirit93: I may be missing something, but why exactly do we need to raise the ```CalledProcessError``` at all. As long as we tell the user why ```check_output``` failed, shouldn't that be fine? Using ```shell=true``` with ```subprocess.check_output```, in case of erroneous input, the same error message gets displayed as it would if the user were to enter the command on the terminal. This serves the purpose of telling the user why ```check_output``` failed.

If I pass ```ldfs``` to ```check_output```, stderr will have ```/bin/sh: ldfs: command not found```. Similarly, if you were to type the same command on the terminal, the output would be ```bash: ldfs: command not found```.

In the earlier code stderr would have had a huge error message indicating a ```FileNotFound```. If required ```CalledProcessError``` can be captured and a similar error message can be logged.

In case I've missed something and there is some reason that an implementation similar to the earlier one is preferred, I will look into that.

tmylk: First of all we need the actual output to be written to log, not to stderr.
Secondly, the tests should check that the return message is indeed for `ldfs: command not found`. The current message they test for is not informative

kirit93: Would it be okay for the log to have the error message as follows 
```/bin/sh: ldfs: command not found
Command '['ldfs']' returned non-zero exit status 127.``` 
along with a ```subprocess.CalledProcessError``` being raised? 

In the test case, I'm unable to figure out how to use ```assertRaises``` to verify that it was in fact a ```subprocess.CalledProcessError``` that was raised.

**Comment by kirit93** [comment]
@tmylk, could you check out my latest push and let me know if that works for us?

**Comment by kirit93** [comment]
@menshikh-iv, could you let me know if this PR is okay to merge?

File: gensim/utils.py [hunk 12]
```diff
@@ -1154,29 +1154,29 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
-def check_output(stdout=subprocess.PIPE, *popenargs, **kwargs):
-    """
-    Run command with arguments and return its output as a byte string.
-    Backported from Python 2.7 as it's implemented as pure python on stdlib.
-
-    >>> check_output(args=['/usr/bin/python', '--version'])
-    Python 2.6.2
-    Added extra KeyboardInterrupt handling
+def check_output(args, flag=True):
+    r"""
```
menshikh-iv: Not required `r`

gensim/utils.py
```diff
@@ -1154,29 +1154,29 @@ def keep_vocab_item(word, count, min_count, trim_rule=None):
...
 
-def check_output(stdout=subprocess.PIPE, *popenargs, **kwargs):
-    """
-    Run command with arguments and return its output as a byte string.
-    Backported from Python 2.7 as it's implemented as pure python on stdlib.
-
-    >>> check_output(args=['/usr/bin/python', '--version'])
-    Python 2.6.2
-    Added extra KeyboardInterrupt handling
+def check_output(args, flag=True):
```
menshikh-iv: A new interface breaks down many wrappers from gensim. First, please merge develop to your branch (it needs to correct test run)

Second, run tests locally with all wrappers like this 
`FT_HOME=/home/ivan/release/test/fastText WR_HOME=/home/ivan/release/test/wordrank  VOWPAL_WABBIT_PATH=/home/ivan/release/test/vowpal_wabbit/vowpalwabbit/vw DTM_PATH=/home/ivan/release/test/dtm/dtm/main MALLET_HOME=/home/ivan/release/test/Mallet python setup.py test`

Another variant is [PR#1368 docker image](https://github.com/RaRe-Technologies/gensim/pull/1368) for this purposes

kirit93: My branch is upto date with develop. I ran ```python setup.py test``` with the appropriate env variable. Two tests failed but these are tests unrelated to what I am working on. 

What should I do now?

gensim/utils.py
kirit93: Done.

gensim/utils.py
menshikh-iv: @kirit93 hm, strange, I'll run all the tests and write the results here.

kirit93: It seems like the travis build succeeded after develop was merged into issue. Is the PR good to go?

menshikh-iv:
I run all tests and get 39 errors, but they are all similar to each other:

First type
```
ERROR: testEnsemble (gensim.test.test_wordrank_wrapper.TestWordrank)
Test ensemble of two embeddings
  File "/home/ivan/release/test/gensim/gensim/test/test_wordrank_wrapper.py", line 37, in setUp
    self.test_model = wordrank.Wordrank.train(self.wr_path, self.corpus_file, self.out_name, iter=6, dump_period=5, period=5)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/wordrank.py", line 87, in train
    os.makedirs(meta_dir)
  File "/home/ivan/.virtualenvs/clean2/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/home/ivan/release/test/wordrank/testmodel/meta'
```

Second type
```
ERROR: testLargeMmap (gensim.test.test_ldamallet_wrapper.TestLdaMallet)
  File "/home/ivan/release/test/gensim/gensim/test/test_ldamallet_wrapper.py", line 61, in setUp
    self.model = ldamallet.LdaMallet(self.mallet_path, corpus, id2word=dictionary, num_topics=2, iterations=1)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/ldamallet.py", line 100, in __init__
    self.train(corpus)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/ldamallet.py", line 158, in train
    self.convert_input(corpus, infer=False)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/ldamallet.py", line 155, in convert_input
    check_output(args=cmd, shell=True)
TypeError: check_output() got an unexpected keyword argument 'shell'
```

Third type
```
ERROR: testCalledProcessError (gensim.test.test_dtm.TestDtmModel)
  File "/home/ivan/release/test/gensim/gensim/test/test_dtm.py", line 70, in testCalledProcessError
    rng_seed=1)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/dtmmodel.py", line 129, in __init__
    self.train(corpus, time_slices, mode, model)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/dtmmodel.py", line 202, in train
    check_output(args=cmd, stderr=PIPE)
TypeError: check_output() got an unexpected keyword argument 'stderr'
```

For this reason, I can't merge your PR.

kirit93:
@menshikh-iv for the second two error types I can fix it by changing the definition of my ```check_output``` function to be like this ```def check_output(args, flag=True, stderr=None)```.

This way the way the function is called will not have to be changed in the other files in the code. 
Another option is that I could change the calls to ```check_output``` so that they fit in with the new implementation.

And the first type of error is not related to ```check_output``` right?

menshikh-iv:
@kirit93 First type error is strange, I will investigate it later. Could you please do a few things:
- Return old interface for `check_output` function
- Re-raise exception (if subprocess call breaks down)
- Logging (with debug level) full command string with args, that user pass to `check_output`
- Check case with `KeyboardInterrupt` (it should be possible to stop this call with `Ctrl + C`)

After this fixes, I will run all tests again and write a result to this PR.

kirit93:
@menshikh-iv I made the changes you requested and pushed them. Travis built successfully. Please let me know if the PR is okay now.

menshikh-iv:
Sorry for late response,

Now, I get a new kind of error, look at command
```
ERROR: testLargeMmap (gensim.test.test_ldamallet_wrapper.TestLdaMallet)
  File "/home/ivan/release/test/gensim/gensim/test/test_ldamallet_wrapper.py", line 61, in setUp
    self.model = ldamallet.LdaMallet(self.mallet_path, corpus, id2word=dictionary, num_topics=2, iterations=1)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/ldamallet.py", line 100, in __init__
    self.train(corpus)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/ldamallet.py", line 158, in train
    self.convert_input(corpus, infer=False)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/ldamallet.py", line 155, in convert_input
    check_output(args=cmd, shell=True)
  File "/home/ivan/release/test/gensim/gensim/utils.py", line 1171, in check_output
    res = subprocess.check_output(args, shell=True)
  File "/usr/lib/python2.7/subprocess.py", line 219, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
CalledProcessError: Command '/ h o m e / i v a n / r e l e a s e / t e s t / M a l l e t / b i n / m a l l e t   i m p o r t - f i l e   - - p r e s e r v e - c a s e   - - k e e p - s e q u e n c e   - - r e m o v e - s t o p w o r d s   - - t o k e n - r e g e x   " \ S + "   - - i n p u t   / t m p / 1 5 e 4 d b _ c o r p u s . t x t   - - o u t p u t   / t m p / 1 5 e 4 d b _ c o r p u s . m a l l e t' returned non-zero exit status 126
```

kirit93:
@menshikh-iv, I reverted to the old implementation of ```check_output``` and I ran the tests again. I still get an error like the one you posted.

```
ERROR: testCnpmiVWModel (gensim.test.test_coherencemodel.TestCoherenceModel)
Perform sanity check to see if c_npmi coherence works with LDA VW gensim wrapper
  File "/Users/kirit/OpenSource/gensim/gensim/test/test_coherencemodel.py", line 70, in setUp
    self.malletmodel = LdaMallet(mallet_path=self.mallet_path, corpus=corpus, id2word=dictionary, num_topics=2, iterations=0)
  File "/Users/kirit/OpenSource/gensim/gensim/models/wrappers/ldamallet.py", line 100, in __init__
    self.train(corpus)
  File "/Users/kirit/OpenSource/gensim/gensim/models/wrappers/ldamallet.py", line 158, in train
    self.convert_input(corpus, infer=False)
  File "/Users/kirit/OpenSource/gensim/gensim/models/wrappers/ldamallet.py", line 155, in convert_input
    check_output(args=cmd, shell=True)
  File "/Users/kirit/OpenSource/gensim/gensim/utils.py", line 1206, in check_output
    raise error
subprocess.CalledProcessError: Command '/home/ivan/release/test/Mallet/bin/mallet import-file --preserve-case --keep-sequence --remove-stopwords --token-regex "\S+" --input /var/folders/gq/gqtxhd8j2cl0v9chmcjpd3mr0000gn/T/9de601_corpus.txt --output /var/folders/gq/gqtxhd8j2cl0v9chmcjpd3mr0000gn/T/9de601_corpus.mallet' returned non-zero exit status 127.
```

Could the cause of the error be something unrelated to the ```check_output``` code in this PR?

piskvorky:
@kirit93 unless your system user is `ivan`, that error log looks extremely fishy.

kirit93:
@piskvorky, I ran the following command ```FT_HOME=/home/ivan/release/test/fastText WR_HOME=/home/ivan/release/test/wordrank VOWPAL_WABBIT_PATH=/home/ivan/release/test/vowpal_wabbit/vowpalwabbit/vw DTM_PATH=/home/ivan/release/test/dtm/dtm/main MALLET_HOME=/home/ivan/release/test/Mallet python setup.py test```

piskvorky:
@kirit93 Where did you get that command from, do those paths really exist?

kirit93:
No they don't exist on my system, but @menshikh-iv asked me to run that command to test whether the PR works or not.

menshikh-iv:
@kirit93 It's a paths in my filesystem.

For the full test run, you should:
1. Install all external dependencies for wrappers (Fasttext, WordRank, WovpalVabbit, DTM, MALLET).
2. Create a clean virtual environment.
3. Install all test dependencies `pip install .[test]` (in gensim folder).
4. Substitute the necessary paths to environment variables and run a command from your message with correct paths for your filesystem.

menshikh-iv:
@kirit93 I just ran all the tests on the development branch and made sure that everything works with the current implementation (from `origin/develop`)

kirit93:
@menshikh-iv, any idea what could be the cause of this error?

menshikh-iv:
In my opinion, it is now easier to recreate PR and slightly change the code from `develop` branch.
Before that, please learn to run all tests for wrappers ([instruction from previous post](https://github.com/RaRe-Technologies/gensim/pull/1182#issuecomment-308337613))

PR #1183, opened 2017-03-05
Make Word2vec multiple passes more explicit. Fix #1052

### LINKED ISSUE #1052: Make Word2vec multiple passes more explicit

robotcator:
I modify the method 'train()' by adding some parameters --iteration=None, epochs=1, start_alpha=None, end_alpha=None, cur_epoch=None. And the default epochs is one, others are None. If the user trains the model with a loop, the epochs and cur_epoch(the current epoch) must be specified. So if the user don't specify the start_alpha and end_alpha(the usage is the same as alpha and min_alpha), we can recaculate the start_alpha and end_alpha according to the current epoch to support linear learning-rate decay from (initial) alpha to min_alpha.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1271,7 +1271,7 @@ class Word2Vec(utils.SaveLoad):
         # update older models
         if hasattr(model, 'table'):
             delattr(model, 'table')  # discard in favor of cum_table
-        if model.negative and hasattr(model, 'index2word'):
+        if model.negative and hasattr(model.wv, 'index2word'):
             model.make_cum_table()  # rebuild cum_table from vocabulary
         if not hasattr(model, 'corpus_count'):
             model.corpus_count = None
--- a/gensim/test/test_doc2vec.py
+++ b/gensim/test/test_doc2vec.py
@@ -142,7 +142,7 @@ class TestDoc2VecModel(unittest.TestCase):
         model.build_vocab(corpus)
         self.assertTrue(model.docvecs.similarity_unseen_docs(model, rome_str, rome_str) > model.docvecs.similarity_unseen_docs(model, rome_str, car_str))
 
-    def model_sanity(self, model):
+    def model_sanity(self, model, keep_training=True):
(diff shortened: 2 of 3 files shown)


robotcator:
I also have another idea that we can force the user to pass those parameters(iter, alpha, min_alpha) explicitly. If the user train the model using a loop, they can carefully pass the start_alpha and end_alpha to support one long smooth descent from alpha to min_alpha.

gojomo:
Another pending PR with same general idea is #1139 - forcing users to be aware of `train()`'s parameters by making them required.

tmylk:
@robotcator it is best to close this PR and finish #1139 in particular add tests for `assertRaises` if the new `epochs` param is ommiteed; update the ipynbs that use train; implement `epoch_callback`,

robotcator:
@tmylk Thank you for your suggestion. But can you give more details about implementing epoch_callback ?
@gojomo  Would you mind let me to finish the [#1139](https://github.com/RaRe-Technologies/gensim/pull/1139) to do list.

gojomo:
@robotcator - Please feel free to continue/finish #1139. In my opinion the most important thing is to ensure notebooks/docs reflect the change & current best-practices, and resolve any conflicts with the main `develop` branch. 

Implementing `epoch_callback` would require some careful refactoring of the existing loops and multi-threading; I'm not sure of the best way to proceed yet, which is one of the reasons I put #1139 aside.

PR #1189, opened 2017-03-07
[MRG] Load FastText models with specified encoding

jayantj:
Related to discussion in [#1176](https://github.com/RaRe-Technologies/gensim/pull/1176#issuecomment-284623428)

Allows user to specify a custom encoding themselves, since the FastText model file does not contain any encoding information, and FastText can create files with non-utf8 encodings in practice.

In the future, in case FastText decides to enforce utf8 (as mentioned in the readme), we may want to remove this parameter.

Also updates the header in the fasttext wrapper file.


tmylk:
Awesome, even with non-utf8 test!

PR #1017, opened 2016-11-13
PEP8 Fixes for Summarization.

souravsingh:

Issue #632, opened 2016-03-23
Error: from scipy.sparse import sparsetools


ghost:
Hello,

After installation into Python 2.7, WinPython version, I got this error:

import gensim

  File "D:\_devs\Python01\WinPython-64-2710\python-2.7.10.amd64\lib\site-packages\gensim\models\lsimodel.py", line 59, in <module>
    from scipy.sparse import sparsetools

ImportError: cannot import name sparsetools

How can I resolve the issue (it seems sparsetools is deprecated in Scipy)

Thanks so much

piskvorky:
Related to #462 and #557 .

tmylk:
@arita291 Did you use anaconda, windows wheel to install or did you build from source?

Downgrading to scipy 0.16 is a temporary workaround.
A permanent fix would be https://github.com/scipy/scipy/issues/5348

arita291:
Hi,
Winpython and install with  !pip install gensim  in Python 2.7

Best is to donwgrade to scipy 0.16 ?

tmylk:
It imports in scipy 0.18 on linux ok.
Closing as can't reproduce

PR #1202, opened 2017-03-10
pep8/pycodestyle fixes for hanging indents in Summarization module 

SamriddhiJain:
Updated the code according to `pycodestyle`. Fixed occurrences of vertical indent to hanging indent.

PR in pycodestyle for detecting hanging indents: [#632](https://github.com/PyCQA/pycodestyle/pull/632) 
References: [#1081](https://github.com/RaRe-Technologies/gensim/pull/1081), [#1017](https://github.com/RaRe-Technologies/gensim/pull/1017)


tmylk:
Thanks. When it is ready in due time, could you please link to a PR in pycodestyle repo that  highlights error when indent is not a hanging indent?

gensim/summarization/bm25.py
```diff
@@ -40,16 +40,17 @@ def initialize(self):
...
         score = 0
         for word in document:
             if word not in self.f[index]:
                 continue
             idf = self.idf[word] if self.idf[word] >= 0 else EPSILON * average_idf
-            score += (idf*self.f[index][word]*(PARAM_K1+1)
-                      / (self.f[index][word] + PARAM_K1*(1 - PARAM_B+PARAM_B*self.corpus_size / self.avgdl)))
+            score += (
+                idf * self.f[index][word] * (PARAM_K1 + 1) /
+                (self.f[index][word] + PARAM_K1 * (1 - PARAM_B + PARAM_B * self.corpus_size / self.avgdl)))
```
tmylk: line break should be [before binary operator](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)

SamriddhiJain: will update that, should I also change line 51 as in this [diff](https://github.com/RaRe-Technologies/gensim/pull/1081/files#diff-5f346637ecf4de2ea08be73e33dd6e71), or is this fine (no arguments on first line).

SamriddhiJain:
As per the implementation of check for non hanging indents, you can see the diff [here](https://github.com/PyCQA/pycodestyle/compare/master...SamriddhiJain:b1).
For hanging indents there are two conditions, no argument in first line and continuation line check. The latter is already implemented in pycodestyle [1](https://github.com/PyCQA/pycodestyle/blob/ca5350195031c8e29c50292404eaec848bbb073b/pycodestyle.py#L445). So I have implemented the first check (which occurs in the case of visual indents), when arguments are present on the first line.
Although I have submitted the pull request, but the tests failed because most of the tests and code itself are in vertical indents format, that is why I think its difficult to merge these changes in their branch, unless the tests are also carefully changed.

gensim/summarization/bm25.py
tmylk: yes, It's not a function call so no need for hanging indent.

tmylk:
For `pycodestyle` PR it should be an optional new check which is off by default. That will cause no issues with existing tests.

The current implementation has caused a fase positive on bm25.py line 51 so more work is needed.

gensim/summarization/bm25.py
SamriddhiJain: updating this part to something like this [diff](https://github.com/RaRe-Technologies/gensim/pull/1081/files#diff-5f346637ecf4de2ea08be73e33dd6e71) raises this error `E128 continuation line under-indented for visual indent` and using visual indent is not what we want. So should I keep this as it is?

SamriddhiJain: And just to confirm the hanging indents are applicable to multi-line return statements too?

tmylk: there should be no E128 error and hanging indent is applicable to all multi-line statements

tmylk:
Could you please link to the PR in the pycodestyle repo?

SamriddhiJain:
@tmylk Link for PR in pycodestyle: [1](https://github.com/PyCQA/pycodestyle/pull/632).

I tried to keep this test in default ignore by editing in `pycodestyle.py, setup.cfg, test_api.py`, but it still runs. Is there any documentation which can be used to implement this?

tmylk:
Thanks for the fixes and improving pycodestyle. We are now closer to achieving #965.

piskvorky:
@tmylk that's cool, but in the meanwhile, please inspect & fix cody style manually, before merging PRs.

tmylk:
@piskvorky The hanging indent requirement has made it harder to automate the style check.

1. Gensim is the only Python open source project that enforces the hanging indent. It's not a trivial common practice. A [specific extension](https://github.com/PyCQA/pycodestyle/pull/632) is in the process of being added to pycodestyle as described in this PR.

2. Even in gensim codebase there is code from years ago that doesn't comply with hanging indent. ([237 lines](https://github.com/RaRe-Technologies/gensim/issues/965#issuecomment-289554028))

The next steps to automation are merging to pycodestyle and fixing the existing violations. 

To address your point above - without automation we have to resort to the manual process that is bound to let mistakes slip through as shown in point 2 above.

piskvorky:
Gensim didn't have any hanging indent requirement until recently. Its style was more casual, because there weren't as many contributors. We could rely more on common sense.

Fixing existing violations sounds great -- let's start with a consistent, clean slate.

PR #1207, opened 2017-03-12
Allows training if model is not modified by "_minimize_model". Adds deprecation warning.

chinmayapancholi13:
This PR modifies the function `_minimize_model` in word2vec.py by checking if model was modified at all before setting `model_trimmed_post_training`. Also, a deprecation warning has been added for the function `_minimize_model`. The function would be deprecated in the future as using KeyedVectors is the way to go for word2vec.

Maling list [discussion](https://groups.google.com/d/msg/gensim/PxttBVSLixc/SlLehCJrEgAJ).

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -106,6 +106,7 @@ from copy import deepcopy
 from collections import defaultdict
 import threading
 import itertools
+import warnings
 
 from gensim.utils import keep_vocab_item, call_on_class_only
 from gensim.utils import keep_vocab_item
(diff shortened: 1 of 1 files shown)


gensim/models/word2vec.py
```diff
@@ -1251,7 +1251,9 @@ def _minimize_model(self, save_syn1 = False, save_syn1neg = False, save_syn0_loc
             del self.syn1neg
         if hasattr(self, 'syn0_lockf') and not save_syn0_lockf:
             del self.syn0_lockf
-        self.model_trimmed_post_training = True
+        if not save_syn1 or not save_syn1neg or not save_syn0_lockf:
+            self.model_trimmed_post_training = True
+        logger.warning("This method would be deprecated in the future. Use KeyedVectors to retain the previously trained word vectors instead.")
```
piskvorky: `warning.warn` better?

chinmayapancholi13: Yes, I agree. Thanks for your suggestion. I am making this change in the PR.

gensim/models/word2vec.py
```diff
@@ -1251,7 +1251,10 @@ def _minimize_model(self, save_syn1 = False, save_syn1neg = False, save_syn0_loc
             del self.syn1neg
         if hasattr(self, 'syn0_lockf') and not save_syn0_lockf:
             del self.syn0_lockf
-        self.model_trimmed_post_training = True
+        if not save_syn1 or not save_syn1neg or not save_syn0_lockf:
```
tmylk: Do you mean when all 3 are true? Then use `if (save_syn1 and save_syn1neg and save_syn0_lockf)` and just return. Please move this check to the top of the function.

gensim/models/word2vec.py
```diff
@@ -1251,7 +1251,10 @@ def _minimize_model(self, save_syn1 = False, save_syn1neg = False, save_syn0_loc
...
+            self.model_trimmed_post_training = True
+        import warnings
```
tmylk: Please move imports to the file header

gensim/models/word2vec.py
```diff
@@ -1251,7 +1251,10 @@ def _minimize_model(self, save_syn1 = False, save_syn1neg = False, save_syn0_loc
...
+        warnings.warn("This method would be deprecated in the future. Use KeyedVectors to retain the previously trained word vectors instead.")
```
tmylk: Please move to the top of the function. Better text is "This method would be deprecated in the future. Keep `just_word_vectors = model.wv`  to retain just the KeyedVectors instance for read-only querying of word vectors."

chinmayapancholi13:
@tmylk Thanks for the suggestions. I have updated the code to reflect these changes.

tmylk:
Thanks for the pr

Issue #1171, opened 2017-02-24
BleiCorpus after serialize cannot be loaded


vincentmajor:
Python version 2.7.12
gensim version 0.13.2

I'm serializing my corpus into Blei LDA-C format using `corpora.BleiCorpus.serialize(filename, corpus)` which is then later used in a dynamic topic model, and not in python. (I know I can use the DTMModel wrapper, unrelated.)

If I need to come back and load the corpus back into Python I tried `corpora.BleiCorpus.load(filename)`, I get an unpickling error:

     ---------------------------------------------------------------------------
     UnpicklingError                           Traceback (most recent call last)
    <ipython-input-28-0dd15393a941> in <module>()
    ----> 1 test = corpora.BleiCorpus.load('corpus-mult.dat')

    /Applications/anaconda/envs/py27/lib/python2.7/site-packages/gensim/utils.pyc in load(cls, fname, mmap)
        246         compress, subname = SaveLoad._adapt_by_suffix(fname)
        247 
    --> 248         obj = unpickle(fname)
        249         obj._load_specials(fname, mmap, compress, subname)
        250         return obj

    /Applications/anaconda/envs/py27/lib/python2.7/site-packages/gensim/utils.pyc in unpickle(fname)
        909     with smart_open(fname) as f:
        910         # Because of loading from S3 load can't be used (missing readline in smart_open)
    --> 911         return _pickle.loads(f.read())
        912 
        913 

    UnpicklingError: invalid load key, '7'.

The only other argument to `load()` is `mmap` but I don't believe the arrays were stored separately and using `mmap='r'` doesn't help anyway.

tmylk:
@vincentmajor Thanks for reporting it. Are you trying to load the model with 0.13.3 or older version? There is a know issue that is currently being worked on https://github.com/RaRe-Technologies/gensim/issues/1082

vincentmajor:
@tmylk Yes, gensim version 0.13.2 on Python 2.7.12.
But, that issue referenced LDA models whereas I am serializing and loading a corpus without a model. I don't know if the same absence of a `random_state` variable influences loading of a corpus which should be random state invariant.

tmylk:
Agree, it is not related to BleiCorpus. There haven't been any changes. Are you serializing and loading on the same platform, say Linux?

vincentmajor:
Yep, OS X. It doesn't even work loading straight after serializing.

tmylk:
[...]

PR #1215, opened 2017-03-15
Fixing #1171: BleiCorpus after serialize cannot be loaded

pranaydeep-af:
@tmylk. Apologies for the delay. Had to deal with some college course work.
I went through the save() and load() methods and did a few experiments. The save() method stores a pickle dump of the particular object. So in the case of a corpora.dictionary object that is rendered useless due to the presence of a superior method ie. save_corpus(), however the function may be handy when used for saving pickles of some other objects.
So I have added a check to see if the function was called from a corpora object and if it was, I have raised an exception. Otherwise the function works as usual.
For the load function I have added a warning to not use it for loading corpora objects.
Do tell me if some naive mistake has been made.


tmylk:
Travis tests re-ran after smart_open update

gensim/utils.py
```diff
@@ -830,7 +844,7 @@ def run(self):
         it = iter(self.corpus)
         while True:
             chunk = itertools.islice(it, self.chunksize)
-            if self.as_numpy:
+            if self.as_numpy:
```
piskvorky: PEP8: no trailing whitespace.

gensim/utils.py
```diff
@@ -490,6 +497,13 @@ def save(self, fname_or_handle, separately=None, sep_limit=10 * 1024**2,
         in both Python 2 and 3.
 
         """
+        import os
```
piskvorky: Imports belong to the top of file (separated into builtin - external - internal blocks).

(here and elsewhere)

gensim/utils.py
```diff
@@ -260,6 +260,9 @@ def load(cls, fname, mmap=None):
         is encountered.
 
         """
+        import warnings
+        warnings.warn("This method loads pickle dumps. If a corpora was saved with the save_corpus() method," +
```
piskvorky: The message, when emitted, will leave the user confused -- what method?

`corpora` is plural of `corpus`. Do not split messages in the middle with `\n`, leave formatting to the terminal. End sentences in full stop.

gensim/utils.py
```diff
@@ -358,6 +361,10 @@ def _smart_save(self, fname, separately=None, sep_limit=10 * 1024**2,
         in both Python 2 and 3.
 
         """
+        import os
+        directory = os.path.dirname(fname)
+        if not os.path.exists(directory):
```
piskvorky: Definitely not -- no creating of directories please. Non-existent path should end in an exception (as it was before).

gensim/utils.py
```diff
@@ -490,6 +497,13 @@ def save(self, fname_or_handle, separately=None, sep_limit=10 * 1024**2,
         in both Python 2 and 3.
 
         """
+        import os
+        directory = os.path.dirname(fname_or_handle)
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+        import corpora.dictionary
+        if isinstance(self, corpora.dictionary.Dictionary):
+            raise TypeError('The method is not built to handle corpora. Use the save_corpus() method')
```
piskvorky: The `save` method can handle corpora fine, it's just probably not what the user wants. Plus, you're testing here for `Dictionary` anyway, not corpora.

The recommended way for serializing corpora is `SomeCorpusClass.serialize()`, not `save_corpus()`.

tmylk:
Closing as abandoned.
Also this needs to be specced out better as to what exactly the fix should be.

Issue #1129, opened 2017-02-01
gensim/utils.py occur TypeError: file must have a 'write' attribute


AaronKim-CN:
Hello, I am new to gensim library. 

When I run a code like bellow, the type error will occur.

```python
    model = models.Doc2Vec(alpha=0.025, min_alpha=0.025)  # use fixed learning rate
    model.build_vocab(sentences)
    # sentences is a labeled text data. using models.doc2vec.LabeledSentence()

    # store the model to mmap-able files
    model.save(model_basename+'.d2v')
```

the method of save will cause the error

```
Traceback (most recent call last):
  File "/root/.pyenv/versions/3.6.0/lib/python3.6/site-packages/gensim/utils.py", line 493, in save
    _pickle.dump(self, fname_or_handle, protocol=pickle_protocol)
TypeError: file must have a 'write' attribute
```

Could you help me?

tmylk:
Hi,
Could you please paste code defining `model_basename` and complete stack trace?

This is strange because the exception should be [caught](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/utils.py#L495) by the code. The Python 3.6 tests pass in Continuous Integration where filename is given as a string.

AaronKim-CN:
Hi, tmylk

Thank you for your response.

the defination of model_basename is follwing.

 model_basename = './D2V/doc2vec_result/model'

I have confirmed the type of [ model_basename+'.d2v' ] is str.

And following is the stack trace.

ERROR:root:Traceback (most recent call last):
  File "/root/.pyenv/versions/3.6.0/lib/python3.6/site-packages/gensim/utils.py", line 493, in save
    _pickle.dump(self, fname_or_handle, protocol=pickle_protocol)
TypeError: file must have a 'write' attribute

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "doc2vecKmeans.py", line 114, in <module>
    model.save(model_basename+'.d2v')
  File "/root/.pyenv/versions/3.6.0/lib/python3.6/site-packages/gensim/models/word2vec.py", line 1462, in save
    super(Word2Vec, self).save(*args, **kwargs)
  File "/root/.pyenv/versions/3.6.0/lib/python3.6/site-packages/gensim/utils.py", line 497, in save
    pickle_protocol=pickle_protocol)
  File "/root/.pyenv/versions/3.6.0/lib/python3.6/site-packages/gensim/utils.py", line 369, in _smart_save
[...]

PR #1225, opened 2017-03-21
Fixing #1129

pranaydeep-af:
Added Code to check if directory exists, if it doesn't then created. @tmylk


tmylk:
Re-reading the comment in https://github.com/RaRe-Technologies/gensim/issues/1129#issuecomment-278040036 it seems that this will not fix the issue. Please reproduce the issue as reported and suggest a fix to the save/smart_save methods.

PR #1226, opened 2017-03-21
[WIP] index object pickle forced to be always under `index.output_prefix` in Similarity

### LINKED ISSUE #691: Force the pickle to be always under `index.output_prefix`
Force the pickle to be always under `index.output_prefix`

See discussion in 
https://groups.google.com/d/msg/gensim/W4iJRVoeiO8/EWkyuR1oKgAJ

prakhar2b:
@piskvorky @tmylk Fix #691 

ref- https://github.com/RaRe-Technologies/gensim/pull/1226#discussion_r125277354


prakhar2b:
Test for similarity needs to be updated as well now

gensim/utils.py
```diff
@@ -266,6 +266,10 @@ def load(cls, fname, mmap=None):
         obj = unpickle(fname)
         obj._load_specials(fname, mmap, compress, subname)
         logger.info("loaded %s", fname)
+        if cls.__name__ is 'Similarity':
+            if obj.output_prefix is not fname:
+                obj.output_prefix = fname[: fname.find("/shard")]
```
prakhar2b: **Correction** -
> -rfind should be used instead of find

gensim/similarities/docsim.py
```diff
@@ -191,6 +191,11 @@ def __init__(self, output_prefix, corpus, num_features, num_best=None, chunksize
             self.output_prefix = utils.randfname(prefix='simserver')
         else:
             self.output_prefix = output_prefix
+            
+        self.shard_dir = self.output_prefix + 'shard/'
```
piskvorky: Not portable -- `/` is UNIX-only.

gensim/similarities/docsim.py
```diff
@@ -429,9 +434,9 @@ def check_moved(self):
         """
         Update shard locations, in case the server directory has moved on filesystem.
         """
-        dirname = os.path.dirname(self.output_prefix)
+        dirname = self.output_prefix
         for shard in self.shards:
-            shard.dirname = dirname
+            shard.dirname = dirname + '/shard'
```
piskvorky: Not portable -- `/` is UNIX-only.

gensim/utils.py
piskvorky: Not portable -- `/` is UNIX-only.

gensim/utils.py
```diff
@@ -266,6 +266,10 @@ def load(cls, fname, mmap=None):
         obj = unpickle(fname)
         obj._load_specials(fname, mmap, compress, subname)
         logger.info("loaded %s", fname)
+        if cls.__name__ is 'Similarity':
```
piskvorky: `issubclass` better?

Also, comparing strings for object equality with `is`, instead of content equality with `==`, is undesirable. Python string interning is a non-trivial subject.

prakhar2b: @piskvorky  should issubclass be called from inside a ```classmethod``` function, when we are not calling it from the class we intend to compare it with 

```isinstance(cls, Similarity)``` throws error global name 'Similarity' is not defined

**Comment by prakhar2b** [comment]
@piskvorky I made changes accordingly. Please review and look into this [comment above](https://github.com/RaRe-Technologies/gensim/pull/1226#issuecomment-288078063)

**Comment by prakhar2b** [comment]
@piskvorky I made changes accordingly. Please review and look into this [comment above](https://github.com/RaRe-Technologies/gensim/pull/1226#issuecomment-288078063)

File: gensim/utils.py (hunk 4 above)
Comment (piskvorky) [inline gensim/utils.py]: Hm, I haven't though of that.

So there's some sort of circular dependency? The current solution may be the simplest then (just with `==` instead of `is`).

File: gensim/similarities/docsim.py [hunk 5]
```diff
@@ -190,7 +190,11 @@ def __init__(self, output_prefix, corpus, num_features, num_best=None, chunksize
             # undocumented feature: set output_prefix=None to create the server in temp
             self.output_prefix = utils.randfname(prefix='simserver')
         else:
-            self.output_prefix = output_prefix
+            self.output_prefix = os.path.join(os.path.dirname(output_prefix), 'shard\\' if os.name == 'nt' else 'shard/')
```
piskvorky: try `os.sep`.

prakhar2b: ``` os.path.join ( os.path.dirname('x/y/z/abc'),'shard', '') ``` will give 
```x/y/z/shard/ ```

we want everything to happen inside shard directory. For example, we want shard index to be saved as ```/x/y/z/shard/.0``` not as ```x/y/z/shard.0``` which is the case in absence of leading slash.

Earlier, if user had passed path as self.output_index to be ```x/y/z/abc``` the shard would be ```x/y/z/abc.0``` and later user may provide some other path for index, so the idea is to have both in same directory.

File: gensim/similarities/docsim.py [hunk 6]
```diff
@@ -190,7 +190,7 @@ def __init__(self, output_prefix, corpus, num_features, num_best=None, chunksize
             # undocumented feature: set output_prefix=None to create the server in temp
             self.output_prefix = utils.randfname(prefix='simserver')
         else:
-            self.output_prefix = os.path.join(os.path.dirname(output_prefix), 'shard\\' if os.name == 'nt' else 'shard/')
+            self.output_prefix = os.path.join(os.path.dirname(output_prefix), 'shard', '') # '' to get the leading slash
```
piskvorky: Do you mean trailing slash?

prakhar2b: oh yes, pardon me, it's trailing slash only. Updated the PR @piskvorky . Thanks

tmylk:
Please add a test where pickle and shards are moved.

prakhar2b:
**Note : ** The`shard` directory is created only when `output_prefix` parameter provided is not `None` , as there might be permission issues while creating a directory in `/tmp` folder.

For reference --  [IPython](https://github.com/prakhar2b/Weekend-Projects/blob/master/gensim/%23691.ipynb) demonstrating the changes.

prakhar2b:
@tmylk I think I should also update tutorial notebook for similarity.

As of now, ```similarities.MatrixSimilarity``` is demonstrated, but ```similarities.Similarity``` is not demonstrated in the [Ipython](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/Similarity_Queries.ipynb). After finalizing the api for proposed changes in this PR, I will update the tutorial accordingly

gensim/test/test_similarities.py
```diff
@@ -427,6 +427,17 @@ def testReopen(self):
...
+    def testShardDir(self):
+        """test where shard and pickles are moved """
+        
+        homedir = os.path.join(os.path.expanduser('~'),'')
+        index = similarities.Similarity(homedir, corpus[:5], num_features=len(dictionary), shardsize=9)
+        index.save('test_index')
+        self.assertTrue(os.path.exists(os.path.join(os.path.dirname(homedir), 'shard', '')))
+        self.assertTrue(os.path.exists(os.path.join(os.path.dirname(homedir), 'shard', '.0')))
+        self.assertTrue(os.path.exists(os.path.join(os.path.dirname(homedir), 'shard', 'test_index')))
+        index.destroy()
```
tmylk: please add a test that actually moves the files with `os.rename`

prakhar2b: yes, things to do 
- [x] create shard directory, and test for pickle and shard
- [ ] move the files with os.rename
- [ ] load the files from new location, and test for check_moved()
- [ ] delete shard directory (destroy workspace thus created)

I'll do it asap.Thanks @tmylk

prakhar2b:
@tmylk  updated the code . ```os.rename``` can't be used ih this case as we want the users to move the entire ```shard``` folder to new location. I've used ```shutil``` python inbuilt library.

Apologies for delay in fixing the issue. I'll be done with my examinations by next week. Thanks

gensim/utils.py
```diff
@@ -266,6 +266,10 @@ def load(cls, fname, mmap=None):
         obj = unpickle(fname)
         obj._load_specials(fname, mmap, compress, subname)
         logger.info("loaded %s", fname)
+        if cls.__name__ == 'Similarity':
```
piskvorky: This deserves a comment (the reason we're not checking for a subclass instead).

gensim/utils.py
```diff
@@ -266,6 +266,10 @@ def load(cls, fname, mmap=None):
...
+            if obj.output_prefix is not fname:
+                obj.output_prefix = os.path.join(fname[: fname.rfind("shard")] , 'shard' ,'')   # '' to get trailing slash
```
piskvorky: This deserves a comment -- why are we removing / adding the string `shard`?

Also nitpick: space after comma, not before.

gensim/similarities/docsim.py
```diff
@@ -190,7 +190,10 @@ def __init__(self, output_prefix, corpus, num_features, num_best=None, chunksize
             # undocumented feature: set output_prefix=None to create the server in temp
             self.output_prefix = utils.randfname(prefix='simserver')
         else:
-            self.output_prefix = output_prefix
+            self.output_prefix = os.path.join(os.path.dirname(output_prefix), 'shard', '')   # '' to get the trailing slash
+            if not os.path.exists(self.output_prefix):
+                os.makedirs(self.output_prefix)
```
prakhar2b: yes, there should an except statement for when it fails to create ```shard``` directory. I'll update the pr asap, thanks.

prakhar2b:
@tmylk @piskvorky Updated the PR according to review. One more thing to do -

- [ ] Change required in test_similarities.py -- Now, with `shard` directory saving both index and pickle file for class `Similarity' , we don't need to provide file location to be saved, we only need index file name. This is the reason for check fail as of now.

tmylk:
Apart from making the tests pass, please also take care of the style warning in `docsim.py` in the Travis logs

menshikh-iv:
@prakhar2b what's a status of this PR?

prakhar2b:
**Update** - updated the description [above](https://github.com/RaRe-Technologies/gensim/pull/1226#issue-215736126). Pushing updated codes asap

gensim/similarities/docsim.py
```diff
@@ -186,11 +186,27 @@ def __init__(self, output_prefix, corpus, num_features, num_best=None, chunksize
...
+                    os.makedirs(self.output_prefix)
+                    logger.info("created %s directory . Both shard and index files are stored at this location. Please use \
+                    this entire folder for portability across different systems. ", self.output_prefix)
+                    self.is_shard = True  # helps while saving the index file as we don't want to create directories in temp
+                else:
+                    logger.warning("It seems there is already a file named shard in location %s . Please ensure  \
+                    that there is no other files named shard in the provided location. Starting similarity index under temp folder", os.path.join(os.path.dirname(output_prefix), ''))
+                    self.output_prefix = utils.randfname(prefix='simserver')
+            except:
+                logger.warning("Failed to create shard folder in %s. Please ensure that you have required permission to create  \
```
piskvorky: Huge -1: what is this?

User asked used us to save to location X, but we choose to save to location Y willy-nilly?

gensim/similarities/docsim.py
```diff
@@ -443,7 +459,14 @@ def save(self, fname=None, *args, **kwargs):
         """
         self.close_shard()
         if fname is None:
-            fname = self.output_prefix
+            if(issubclass(self.__class__, Similarity)):
```
piskvorky: Python doesn't use brackets to enclose `if` expressions (see PEP8 for Python code style).

gensim/similarities/docsim.py
prakhar2b: yes, this API concern is causing delay for this PR. As mentioned in the description, as of now, we ask user for `output_prefix` location where we save shards as `out_prefix.shard_number`, and while saving we ask for location for `index object pickle`. This causes problem while loading, if user has moved pickle to a different system where they don't have shards, or if user has changed the location of `shards`.

I understand this approach of creating a directory is quite ambiguous, so do you think it will be better to ask the user for a location like `/home/prakhar/index` and create a directory `index` in `/home/peakhar/` and inside that directory create shards as `index.0` etc. And while saving too, we only ask for pickle name and not entire location, which we will save inside that directory only. The catch here, however, is what if user provides something like `/home/prakhar/` only. As of now, we would create shards `.0`, `.1` etc, but how would we create a directory then ?

After all, we have to save as `pickle_path = output_prefix + index.pkl`

piskvorky: I do not understand the PR description, the phrasing and formatting is hard to parse.

Either way, if user says "store under directory X", we cannot store under directory Y silently. That's just wrong.

If we wish to only support storing to directory Y (and nowhere else), then we shouldn't give the user even the option of saying "store to directory X". The API should be clear.

The problem is not failing if the user asks for some impossible action -- we should fail, explicitly and clearly. The problem is if the API is confusing or does unintuitive clandestine stuff.

piskvorky: To be clear, why not something simple (just pseudo code, I don't remember the exact API, please correct me if this is nonsense):

```python
x = Similarity(prefix=some_partial_path) 
x.add_documents(generator)  # shards stored under some_partial_path.0, .1 etc
x.save()  # automatically stored under some_partial_path.pkl

# user moves all files elsewhere
y = Similarity.load(another_partial_path)  # updates the prefix to the new location
y = Similarity.load(another_partial_path.pkl)  # same thing -- the .pkl is optional
```

prakhar2b: yes, we discussed similar thing on hangouts with @tmylk . We don't need to create a directory.

My implementation was because of one observation - that if input is something like `\home\prakhar\` then we will be creating `.0` , `.1` which will be a hidden file, little difficult for user to locate and move (especially when we have different `name` for shards and pickle). Therefore, I tried for folder approach. (but that would be bad API, I agree)

I think, it's a better approach to ask user for only one input `prefix` and not again while saving. We can raise error if user tries to input a location which is a sub-directory only.
For example-

```python
dir, file_name = os.path.split(input_prefix)
# raise error if file_name is empty
```

piskvorky: That's the idea, yes.

Although that particular test you posted wouldn't work -- we want to test if the given prefix is a directory or not, not whether it ends in a slash.

prakhar2b: yes, that was naive and old approach, will push the updated code asap

gensim/similarities/docsim.py
```diff
@@ -191,6 +191,10 @@ def __init__(self, output_prefix, corpus, num_features, num_best=None, chunksize
             self.output_prefix = utils.randfname(prefix='simserver')
         else:
             self.output_prefix = output_prefix
+            _, file_name = os.path.split(self.output_prefix)
```
piskvorky: This is what we discussed in the PR comments that wouldn't work.

gensim/similarities/docsim.py
```diff
@@ -444,8 +448,20 @@ def save(self, fname=None, *args, **kwargs):
...
             fname = self.output_prefix
+
         super(Similarity, self).save(fname, *args, **kwargs)
 
+    @classmethod
+    def load(self, fname, mmap=None):
+
+        obj = super(Similarity, self).load(fname, mmap)
+        if isinstance(obj,Similarity):
+            if obj.output_prefix is not fname:
```
piskvorky: `is` isn't a good operator for comparing strings.

gensim/similarities/docsim.py
```diff
@@ -444,8 +448,20 @@ def save(self, fname=None, *args, **kwargs):
...
         if fname is None:
             fname = self.output_prefix
+
         super(Similarity, self).save(fname, *args, **kwargs)
 
+    @classmethod
+    def load(self, fname, mmap=None):
+
+        obj = super(Similarity, self).load(fname, mmap)
+        if isinstance(obj,Similarity):
```
piskvorky: PEP8: space after comma.

gensim/similarities/docsim.py
```diff
@@ -444,8 +448,20 @@ def save(self, fname=None, *args, **kwargs):
         self.close_shard()
         if fname is None:
             fname = self.output_prefix
+
         super(Similarity, self).save(fname, *args, **kwargs)
 
+    @classmethod
+    def load(self, fname, mmap=None):
```
piskvorky: Class methods don't have `self`. Instead it's a class, called `cls` by convention.

gensim/similarities/docsim.py
```diff
@@ -444,8 +448,20 @@ def save(self, fname=None, *args, **kwargs):
...
+
+        obj = super(Similarity, self).load(fname, mmap)
+        if isinstance(obj,Similarity):
+            if obj.output_prefix is not fname:
+                obj.output_prefix = fname
```
piskvorky: Doesn't this assume the pickle path is the same as the prefix path? Do we want that?

prakhar2b: yes, I think we discussed this only [here](https://github.com/RaRe-Technologies/gensim/pull/1226#discussion_r125277354). Correct me if I'm wrong

gensim/similarities/docsim.py
prakhar2b: ok, I looked more into it and I think I should better use `os.path.isdir()`

prakhar2b:
still need to make some changes, please don't bother reviewing yet

gensim/similarities/docsim.py
piskvorky: The suggestion in that discussion was to save the pickle under `output_prefix.pkl`.

gensim/similarities/docsim.py
```diff
@@ -191,6 +191,9 @@ def __init__(self, output_prefix, corpus, num_features, num_best=None, chunksize
             self.output_prefix = utils.randfname(prefix='simserver')
         else:
             self.output_prefix = output_prefix
+            if os.path.isdir(self.output_prefix):
+                raise ValueError("output_prefix should not be a directory")
```
piskvorky: Let's include the actual path in the message, for easier debugging.

gensim/similarities/docsim.py
```diff
@@ -444,8 +447,20 @@ def save(self, fname=None, *args, **kwargs):
         self.close_shard()
         if fname is None:
             fname = self.output_prefix
+
         super(Similarity, self).save(fname, *args, **kwargs)
 
+    @classmethod
+    def load(cls, fname, mmap=None):
+
+        obj = super(Similarity, cls).load(fname, mmap)
```
piskvorky: Not safe -- use named argument for `mmap`.

gensim/similarities/docsim.py
```diff
@@ -444,8 +447,20 @@ def save(self, fname=None, *args, **kwargs):
...
+        if isinstance(obj, Similarity):
```
piskvorky: What is this check for?

gensim/similarities/docsim.py
```diff
@@ -444,8 +447,20 @@ def save(self, fname=None, *args, **kwargs):
...
+            if obj.output_prefix != fname:
```
piskvorky: Is this condition needed? IIRC `check_moved` is idempotent if there's no change in the prefix.

jayantj:
Hi @prakhar2b thanks for the PR. From the PR description/our discussion, I'm slightly unclear about what the goal of the PR is. There are two situations involved here - 
1. `Similarity` instance + shards stored as `/original_dir/original_index` and `/original_dir/original_index.0` and then the parent dir being renamed - so the files are now `/renamed_dir/original_index` and `/renamed_dir/original_index.0`. This is the case described in the linked google groups issue. 

2. `Similarity` instance + shards stored as `/original_dir/original_index` and `/original_dir/original_index.0`  and then the index file and shards being renamed to `/original_dir/renamed_index` and `/original_dir/renamed_index.0`. This is the case handled in the unit test.

IMO, for (1) all we need to be doing is to check if the argument passed to `load` matches `output_prefix` or not, and if not, update it accordingly before loading the shards. This is partially done in the PR, but it seems to have a few unnecessary checks along with it which I don't understand (mentioned by @piskvorky too). It also needs to be tested in the unit test.

For (2), I don't think it's easily doable - the test written tries to test for (2), but it does so incorrectly. The shard isn't even loaded when the `Similarity` instance is loaded. It is loaded on `__getitem__` calls, and those fail with a `IOError: [Errno 2] No such file or directory` exception message. I guess we don't need to be handling this case anyway, if the user has changed the actual filenames (and not just the parent directory/path), he is responsible for modifying `output_prefix` himself.

Issue #1224, opened 2017-03-20
PySpaRNN indexer


tmylk:
Add and bechmark new Approximate nearest neighbours library from Facebook [PySpaRNN](https://github.com/facebookresearch/pysparnn)

The code needed is the same as for [Annoy integration](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/annoytutorial.ipynb).

markus-beuckelmann:
I can work on this. Just to make sure we're on the same page: This would involve implementing a `PySpaRNNIndexer` class (or similar) in `gensim/similarities/index.py` that essentially implements the same interface as `AnnoyIndexer`, querying `PySpaRNN` under the hood?

Additionally, we should probably add a unit test in `gensim/test/test_similarities.py` I guess.

tmylk:
Confirm. And a benchmark ipynb is needed too

known-ai:
Hi - the way I did it was to modify the PySpaRNN itself to create a method most_similar. See 
https://github.com/known-ai/KeyedVectorsANN
 But what I should have really done is create a subclass that adds this method. 

Will have to get around to this unless you guys get to it first.

What motivated this was that I couldn't get Annoy to work on my windows machine running Anaconda so I had to find an alternative!

markus-beuckelmann:
I already wrote most of the code this morning. Will submit a PR either tonight or sometime tomorrow.

known-ai:
Nice, I would love to use it!

markus-beuckelmann:
I submitted a PR on this that is ready for review ([#1227](http://github.com/RaRe-Technologies/gensim/pull/1227)). A benchmark notebook will follow...

known-ai:
I should mention that I had submitted some bug fixes to make PySpaRNN compatible with Python3/Anaconda to their developers and they said the fixes were integrated. I haven't checked this.

PR #1227, opened 2017-03-21
[WIP] Add support for PySpaRNN (#1224)

markus-beuckelmann:
This wraps Facebook's [PySpaRNN library](http://github.com/facebookresearch/pysparnn) for approximate NN search on sparse data. The interface allows the `PySpaRNNIndexer` object to be used as indexer within gensim, e.g. something like
```
pysparnn_index = PySpaRNNIndexer(model)
model.most_similar([vector], topn = 3, indexer = pysparnn_index)
```

with some arbitrary `vector`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,14 @@ before_install:
   - ./miniconda.sh -b
   - export PATH=/home/travis/miniconda2/bin:$PATH
   - conda update --yes conda
+  - git clone https://github.com/facebookresearch/pysparnn pysparnn
 install:
   - conda create --yes -n gensim-test python=$TRAVIS_PYTHON_VERSION pip atlas numpy scipy
   - source activate gensim-test
   - pip install pyemd
   - pip install annoy
+  - pip install -r pysparnn/requirements.txt
+  - cd pysparnn; python setup.py install; cd ..
   - pip install testfixtures
   - pip install unittest2
   - pip install Morfessor==2.0.2a4
--- a/docs/notebooks/pysparnn-benchmark.ipynb
+++ b/docs/notebooks/pysparnn-benchmark.ipynb
@@ -1,0 +1,498 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Approximate NN with `PySpaRNN` (for sparse data)\n",
+    "This tutorial will show you how to use the `PySpaRNNIndexer` object for approximate nearest neighbour search."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false,
+    "deletable": true,
+    "editable": true
+   },
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "import numpy as np\n",
+    "\n",
+    "import gensim\n",
+    "from gensim.similarities.index import PySpaRNNIndexer\n",
+    "from gensim.models.word2vec import Word2Vec"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# How to use `PySpaRNNIndexer`\n",
+    "Let's prepare some data first: For this we will load text data from the Lee Corpus."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# This is adapted from /docs/notebooks/annoytutorial.ipynb\n",
+    "training_file = os.path.join(gensim.__path__[0], 'test', 'test_data', 'lee_background.cor')\n",
+    "\n",
+    "class MyText(object):\n",
+    "    def __iter__(self):\n",
+    "        for line in open(training_file):\n",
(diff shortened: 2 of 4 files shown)


tmylk:
Thanks. Please add unit tests for this new feature and a benchmark tutorial

tmylk:
Thanks for the unit tests. Do you think they can be refactored to avoid duplicate code with Annoy tests? The only difference is the indexer

markus-beuckelmann:
True. I refactored the unit tests, which indeed removes quite a bit of redundant code.

markus-beuckelmann:
Two more things:
1. We should probably move the `import` statements for both indexers, because right now I would get an `ImportError` if I ran `from gensim.similarities.index import PySpaRNNIndexer` when `annoy` (!) package is not installed and vice versa. In other words: You can not import either of the two indexers without having both backends installed.
2. How can we make sure that `pysparnn` gets installed in the Travis CI environment?

tmylk:
1) Agree about imports. New file pysparnn_indexer.py is needed in `gensim/similarities` folder just for the new indexer. 

2) To install `pysparnn`  please add it to `.travis.yml` file alongside annoy.

markus-beuckelmann:
Just to give you a heads up, I'm still working on the benchmark notebook. Coming soon!

markus-beuckelmann:
Here is what I get when benchmarking `PySpaRNN` vs. `annoy` on artificial data:
* Benchmarking annoy:
`1 loop, best of 3: 5.08 s per loop`
* Benchmarking PySpaRNN:
`1 loop, best of 3: 817 ms per loop`

So it seems that `PySpaRNN` is indeed significantly faster when the data is sufficiently sparse. The code that produced the above:

```python
import numpy as np
from scipy.sparse import csr_matrix

from annoy import AnnoyIndex
from pysparnn.cluster_index import MultiClusterIndex

# Number of vectors
N = 1000
# Number of features
K = 50000
# Sparsity parameters
f = 0.0005
M = int(np.ceil(f * K))

# Hallucinate a corpus
corpus = [[(a, b) for a, b in zip(np.random.randint(0, K, size = M), np.random.normal(size = M))] for i in range(N)]
# Build the feature matrix
vectors, labels = np.zeros((N, K)), np.arange(0, N, 1)
for i, document in enumerate(corpus):
    for j, x in document:
        vectors[i, j] = x

# Train the Annoy model
a = AnnoyIndex(K)
for i, vector in enumerate(vectors):
    a.add_item(i, vector)
n_trees = 300
a.build(n_trees)

# Train the PySpaRNN model
b = MultiClusterIndex(csr_matrix(vectors), labels)

def test_annoy(model, indices, k = 3):

    for i in indices:
        dense = vectors[i]
        sparse = csr_matrix(dense)
        model.get_nns_by_vector(dense, k)

def test_pysparnn(model, indices, k = 3):

    for i in indices:
        dense = vectors[i]
        sparse = csr_matrix(dense)
        model.search(sparse, k = k, return_distance = False)

# Randomly draw some sample vectors to predict on
indices = np.random.randint(N, size = 100)

print('Benchmarking annoy:')
%timeit test_annoy(a, indices)
print('Benchmarking PySpaRNN:')
%timeit test_pysparnn(b, indices)```

markus-beuckelmann:
I will try it on real–world text data now. Any ideas for sufficiently *sparse* data sets that might be particularly suitable for this?

tmylk:
Thanks. Could you please compare it to a brute-force exact NN with BLAS?

tmylk:
Topic distributions per document are sparse. See an example dataset in #1243

markus-beuckelmann:
* Benchmarking annoy:
` 1 loop, best of 3: 5.01 s per loop`
* Benchmarking PySpaRNN:
`1 loop, best of 3: 759 ms per loop`
* Benchmarking brute–force exact NN:
`1 loop, best of 3: 17.2 s per loop`

tmylk:
Thanks! could you please add the bechmarking ipynb with a tutorial and performance graphs to gensim/docs/notebooks?

Similar to https://github.com/RaRe-Technologies/gensim/blob/ed49ea63fb0e49450e9219423d19b929e0402ec8/docs/notebooks/annoytutorial.ipynb

tmylk:
Ping @markus-beuckelmann

markus-beuckelmann:
> Thanks! Please specify BLAS setup and number of cores.

@tmylk: This is Intel MKL (11.3.3) on an i5 with 2 cores. On full load BLAS uses 4 threads for this. The exact NN mentioned above is the brute–force scikit-learn `NearestNeighbors` [implementation](http://scikit-learn.org/stable/modules/neighbors.html#brute-force).

Busy days, notebook is coming up! I will at least add a work in progress version, tomorrow.

menshikh-iv:
This OL looks almost ready, @markus-beuckelmann can you finish this PR?

menshikh-iv:
I close it, because it is abandoned

piskvorky:
@menshikh-iv It's an awesome feature, and @markus-beuckelmann did a good chunk of work.

I'm -1 on abandoning this. What is actually missing here?

markus-beuckelmann:
@menshikh-iv, @piskvorky: A little late here, sorry for that. Support for PySpaRNN is complete (including unit test) and could be merged. Essentially, the only thing that is missing is the tutorial. I can't finish this right now, but maybe other people are willing to pick it up...

PR #1229, opened 2017-03-22
[WIP] Add new restrict_vocab functionality, most_similar_among

shubhvachher:
Fix #481 .

I have looked at the suggestions there and implemented this fix in a new function most_similar_among. @gojomo. I have tested the results and they work well.

The topn variable there has a new functionality that became necessary due to the restricted vocabulary.

Also added being able to get vocabulary as a list from a trained word2vec model.

After the first review, will work on making other _among functions.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -377,28 +377,28 @@ class KeyedVectors(utils.SaveLoad):
         """
 
         if isinstance(words_list, int):
-            raise ValueError("words_list must be a set/list of words. \
-                                Maybe you wanted the most_similar function.")
+            raise ValueError("words_list must be a set/list of words. " \
+                                "Maybe you wanted the most_similar function.")
         elif isinstance(words_list, list) or isinstance(words_list, set):
             pass
         else:  # This is triggered for empty words_list parameter
-            raise ValueError("words_list must be set/list of words. \
-                                Maybe you wanted the most_similar function. \
-                                Please read doc string")
+            raise ValueError("words_list must be set/list of words. " \
+                                "Maybe you wanted the most_similar function. " \
+                                "Please read doc string")
 
         if type(topn) is not int:
             if topn is False:
                 pass
             else:
                 if suppress_warnings is False:
-                    logger.warning("topn needs to either be a number or False. \
-                                    Please read docstring. \
-                                    Displaying all similarities!")
+                    logger.warning("topn needs to either be a number or False. " \
+                                    "Please read docstring. " \
+                                    "Displaying all similarities!")
             topn = len(self.index2word)
 
         self.init_sims()
 
-        if isinstance(positive, string_types) and negative is False:
+        if isinstance(positive, string_types) and not negative:
             # allow calls like most_similar('dog'),
             # as a shorthand for most_similar(['dog'])
             positive = [positive]
@@ -436,21 +436,21 @@ class KeyedVectors(utils.SaveLoad):
         words_to_use = vocabulary_words.intersection(words_list)
 
         if not words_to_use:
-            raise ValueError("None of the words in words_list \
-                                exist in current vocabulary")
+            raise ValueError("None of the words in words_list " \
+                                "exist in current vocabulary")
 
         if suppress_warnings is False:
             missing_words = words_list.difference(vocabulary_words)
             if not missing_words:  # missing_words is empty
                 pass
             else:
-                logger.warning("The following words are not in \
-                                trained vocabulary : %s", str(missing_words))
-                logger.info("This warning is expensive to calculate, \
-                                especially for largs words_list. \
-                                If you would rather not remove the missing_words \
-                                from words_list please set the \
-                                suppress_warnings flag.")
+                logger.warning("The following words are not in " \
+                                "trained vocabulary : %s", str(missing_words))
+                logger.info("This warning is expensive to calculate, " \
+                                "especially for large words_list. " \
+                                "If you would rather not remove the missing_words " \
+                                "from words_list please set the " \
+                                "suppress_warnings flag.")
 
         words_list_indices = [self.vocab[word].index for word in words_to_use]
         # limited = self.syn0norm[words_list_indices]
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -345,39 +345,46 @@ class KeyedVectors(utils.SaveLoad):
         result = [(self.index2word[sim], float(dists[sim])) for sim in best if sim not in all_words]
         return result[:topn]
 
-    def most_similar_among(self, positive=[], negative=[], topn=10, words_list=None, indexer=None,
+    def most_similar_among(self, positive=[], negative=[],
+                            topn=10, words_list=None, indexer=None,
                             suppress_warnings=False):
         """
-        Find the top-N most similar words among words_list to given words. Positive words
-        contribute positively towards the similarity, negative words negatively.
+        Find the top-N most similar words among words_list to given words.
+
+        Positive words contribute positively towards the similarity,
+        negative words negatively.
 
         Please refer to docs of most_similar function.
 
-        If topn is False, most_similar returns the vector of similarity scores for all words
-        in vocabulary of model, restriced by the supplied words_list.
+        If topn is False, most_similar returns the vector of similarity scores
+        for all words in vocabulary of model, restriced by the supplied words_list.
 
-        'words_list' should be a list/set of words. The returned word similarities will only
-        contain similarity scores for those words that are in words_list (and in trained vocabulary).
+        'words_list' should be a list/set of words. The returned word similarities
+        will only contain similarity scores for those words that are in words_list
+        (and in trained vocabulary).
 
-        If some words in words_list are not in vocabulary then a warning is issued to the user.
+        If some words in words_list are not in vocabulary then a warning is
+        issued to the user.
 
         Warnings can be supressed by setting the suppress_warnings flag.
 
         Example::
 
-          >>> trained_model.most_similar_among(positive=['man'], topn=1, words_list=['woman','random_word'])
+          >>> trained_model.most_similar_among(positive=['man'], topn=1,
+                                                words_list=['woman','random_word'])
           [('woman', 0.75882536)]
 
         """
 
         if isinstance(words_list, int):
-            raise ValueError("words_list must be a set/list of words. Maybe you wanted the \
-                                most_similar function.")
+            raise ValueError("words_list must be a set/list of words. \
+                                Maybe you wanted the most_similar function.")
         elif isinstance(words_list, list) or isinstance(words_list, set):
             pass
         else:  # This is triggered for empty words_list parameter
-            raise ValueError("words_list must be set/list of words. Maybe you wanted the \
-                                most_similar function. Please read doc string")
+            raise ValueError("words_list must be set/list of words. \
+                                Maybe you wanted the most_similar function. \
+                                Please read doc string")
 
         if type(topn) is not int:
             if topn is False:
@@ -385,16 +392,19 @@ class KeyedVectors(utils.SaveLoad):
             else:
                 if suppress_warnings is False:
                     logger.warning("topn needs to either be a number or False. \
-                                    Please read docstring. Displaying all similarities!")
+                                    Please read docstring. \
+                                    Displaying all similarities!")
             topn = len(self.index2word)
 
         self.init_sims()
 
         if isinstance(positive, string_types) and negative is False:
-            # allow calls like most_similar('dog'), as a shorthand for most_similar(['dog'])
+            # allow calls like most_similar('dog'),
+            # as a shorthand for most_similar(['dog'])
             positive = [positive]
 
-        # add weights for each word, if not already present; default to 1.0 for positive and -1.0 for negative words
+        # add weights for each word, if not already present;
+        # default to 1.0 for positive and -1.0 for negative words
         positive = [
             (word, 1.0) if isinstance(word, string_types + (ndarray,)) else word
             for word in positive
@@ -426,20 +436,24 @@ class KeyedVectors(utils.SaveLoad):
         words_to_use = vocabulary_words.intersection(words_list)
 
         if not words_to_use:
-            raise ValueError("None of the words in words_list exist in current vocabulary")
+            raise ValueError("None of the words in words_list \
+                                exist in current vocabulary")
 
         if suppress_warnings is False:
             missing_words = words_list.difference(vocabulary_words)
             if not missing_words:  # missing_words is empty
                 pass
             else:
-                logger.warning("The following words are not in trained vocabulary : %s", str(missing_words))
-                logger.info("This warning is expensive to calculate, especially for largs words_list. \
-                                If you would rather not remove the missing_words from words_list \
-                                please set the suppress_warnings flag.")
+                logger.warning("The following words are not in \
+                                trained vocabulary : %s", str(missing_words))
+                logger.info("This warning is expensive to calculate, \
+                                especially for largs words_list. \
+                                If you would rather not remove the missing_words \
+                                from words_list please set the \
+                                suppress_warnings flag.")
 
         words_list_indices = [self.vocab[word].index for word in words_to_use]
-        # limited = self.syn0norm[words_list_indices] #syn0norm is an ndarray so this indexing works
+        # limited = self.syn0norm[words_list_indices]
         # Storing 'limited' might add a huge memory overhead so we avoid doing that
 
         dists = dot(self.syn0norm[words_list_indices], mean)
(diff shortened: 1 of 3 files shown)


gensim/models/keyedvectors.py
```diff
@@ -336,6 +346,95 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
...
+        elif isinstance(restrict_vocab, list) or isinstance(restrict_vocab, set):
+            restrict_vocab = set(restrict_vocab)
+            vocabulary_words = set(self.get_words_from_vocab())
+
+            words_to_use = vocabulary_words.intersection(restrict_vocab)
+
+            if(not suppress_warning):
+                toWarn = "The following words are not in trained vocabulary : "
+                toWarn += str(restrict_vocab.difference(vocabulary_words))
+                warnings.warn(toWarn, UserWarning)
```
gojomo: Seems this will log an attention-grabbing WARNING even if all words are present. Preferable to only log when problems occur.

gensim/models/keyedvectors.py
```diff
@@ -336,6 +346,95 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
...
+            if(not words_to_use):
+                raise ValueError("None of the words in restrict_vocab exist in current vocabulary")
+
+            restrict_vocab_indices = [self.vocab[word].index for word in words_to_use]
+            limited = self.syn0norm[restrict_vocab_indices] #syn0norm is an ndarray
```
gojomo: I think this creates a new array (rather than just a view), which might be a significant memory cost, for large models and when the words-of-interest are a large subset of all words. Maybe this is unavoidable, but might there be any way to do the calculation against the indexes in the original  `syn0norm`?

gensim/models/word2vec.py
```diff
@@ -1180,9 +1180,16 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
                         self.wv.syn0[self.wv.vocab[word].index] = weights
         logger.info("merged %d vectors into %s matrix from %s" % (overlap_count, self.wv.syn0.shape, fname))
 
+    def get_words_from_vocab(self):
```
gojomo: For new KeyedVectors functionality, I don't think we need to add forwarding methods to Word2Vec – those exist to ease the way for older code; any new code can access the KeyedVectors methods directly. (This goes for this method and `most_similar_among()`.)

gensim/models/keyedvectors.py
```diff
@@ -272,6 +273,15 @@ def word_vec(self, word, use_norm=False):
         else:
             raise KeyError("word '%s' not in vocabulary" % word)
 
+    def get_words_from_vocab(self):
```
shubhvachher: Oh, yes! Will change! I'm still keeping the get_words_from_vocab() because when I started out with gensim, it was confusing for me initially what the w2v.wv.vocab keys and items were for!
```python
    def get_words_from_vocab(self):
        """
        returns the words in the current vocabulary as a list.
        """

        if(not self.index2word):
            raise ValueError("Vocabulary needs to be built before calling this function")
        return self.index2word
```

gensim/models/keyedvectors.py
```diff
@@ -336,6 +346,95 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
         result = [(self.index2word[sim], float(dists[sim])) for sim in best if sim not in all_words]
         return result[:topn]
 
+    def most_similar_among(self, positive=[], negative=[], topn=10, restrict_vocab=None, indexer=None,
```
shubhvachher: For lack of imagination : <words_list> !

gensim/models/keyedvectors.py
shubhvachher: Ok, done. thanks for the catch.

gensim/models/keyedvectors.py
shubhvachher: Sure, done. Not storing `limited` anymore...
```python
    words_list_indices = [self.vocab[word].index for word in words_to_use]
            # limited = self.syn0norm[words_list_indices] #syn0norm is an ndarray; storing limited might add a huge memory overhead
        else:
            raise ValueError("words_list must be set/list of words. Please read doc string")

        dists = dot(self.syn0norm[words_list_indices], mean)
        result = []

```

gensim/models/word2vec.py
shubhvachher: hmm, I was confused at first myself but then I just did it to maintain coding style of word2vec.

Do you think it might be confusing for newcomers if we don't have the forwarding methods. It took me quite a while to figure KeyedVectors as having all after training functionality as word2vec!

Also, most of the word2vec functions don't have `docstring` and one has to read the `docstring` of w2v.wv to figure things out. That might throw people off. Would you like me to add docstrings for forwarding functions?

tmylk:
Thanks for the feedback! Yes, having more docstrings is very useful. Explaining what functions do and forwarding to KeyedVectors for more

shubhvachher:
My god. Unit testing is going to be comprehensive(CBOW, SG, cosmul?). I'll committed the changes until now. Adding unit tests soon.

shubhvachher:
@tmylk Will do. Also, what do you think about @gojomo's last review suggestion? Should I add forwarding functions to the Word2Vec file for new functions in KeyedVectors or are we going to ask people to use the KeyedVectors class always?

tmylk:
Please avoid adding new forwarding functions. KeyedVectors is the way to go.

shubhvachher:
Those were my first unit tests! Please do review!

shubhvachher:
I've only added docstrings in the first commit to fail tests. Have any idea what broke? Travis says : 
```

FAIL: testPipeline (gensim.test.test_sklearn_integration.TestSklearnLDAWrapper)

  File "/home/travis/build/RaRe-Technologies/gensim/gensim/test/test_sklearn_integration.py", line 108, in testPipeline

    self.assertGreater(score, 0.50)

AssertionError: 0.5 not greater than 0.5

```

I haven't changed anything though but I found this within the test_word2vec.py file **(not the test that broke)** : 
```python

if not hasattr(TestWord2VecModel, 'assertLess'):
    # workaround for python 2.6
    def assertLess(self, a, b, msg=None):
        self.assertTrue(a < b, msg="%s is not less than %s" % (a, b))

    setattr(TestWord2VecModel, 'assertLess', assertLess)

```

Do you think something similar is needed?

tmylk:
Thanks for reporting this. Please change the test in `testPipeline` to be   `self.assertGreater(score, 0.40)`. It is a new test.

shubhvachher:
Ok, could you please also check #1233. If proper Ill add that fix here as well

shubhvachher:
added fix for #1233

shubhvachher:
This is done, when free could you please look this over @gojomo ? Thanks!

gensim/test/test_word2vec.py
```diff
@@ -461,6 +462,141 @@ def test_cbow_neg(self):
                                   min_count=5, iter=10, workers=2, sample=0)
         self.model_sanity(model)
 
+    def test_most_similar_among_CBOW(self):
```
gojomo: Because `most_similar_among()` functionality is merely on a set of vectors, it's not dependent on the mode-of-training – so there's no need for multiple models built different ways. Testing against one vector set, trained if necessary in the most simple/default way, would be sufficient. (In fact, it'd ultimately make even more sense to be inside a set of KeyedVectors tests, rather than Word2Vec tests, but since its sibling methods are tested here it's not necessary to make that big change yet.)

gensim/test/test_word2vec.py
```diff
@@ -461,6 +462,141 @@ def test_cbow_neg(self):
...
+        """
+        Testing most_similar_among method for CBOW model
+        """
+
+        #Testing Error
+        model = word2vec.Word2Vec(sentences, size=2, sg=0, min_count=1, hs=1, negative=0)
+        self.assertRaises(ValueError, model.wv.most_similar_among, positive=['graph'])
+
+        res_voc = model.wv.index2word[:5] #Gives first 5 words in model vocab
+        #Testing Warnings
```
gojomo: As noted previously, I think the `suppress_warnings` toggle adds unnecessary complication. Also, that logging may be more appropriate than warnings.

gensim/test/test_word2vec.py
```diff
@@ -461,6 +462,141 @@ def test_cbow_neg(self):
...
+    def test_most_similar_among_CBOW(self):
+        """
+        Testing most_similar_among method for CBOW model
+        """
+
+        #Testing Error
+        model = word2vec.Word2Vec(sentences, size=2, sg=0, min_count=1, hs=1, negative=0)
+        self.assertRaises(ValueError, model.wv.most_similar_among, positive=['graph'])
+
+        res_voc = model.wv.index2word[:5] #Gives first 5 words in model vocab
```
gojomo: Usual python style is two-spaces before an end-of-line `#`-comment, and a space after the `#` before text. (Applies many other places as well.)

shubhvachher:
Thank you for the comprehensive review @gojomo ! I will work on this asap.

tmylk:
@shubhvachher Please continue this pr instead of opening a new one in #1255.

shubhvachher:
Ah, sorry. I am working on develop here; will remember not to do that in the future!

So, changes made : Moved drive by fixes into separate PRs and replied to comments above. rebased uptil now.

gensim/models/keyedvectors.py
```diff
@@ -336,6 +346,106 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
         result = [(self.index2word[sim], float(dists[sim])) for sim in best if sim not in all_words]
         return result[:topn]
 
+    def most_similar_among(self, positive=[], negative=[], topn=10, words_list=None, indexer=None,
+                            suppress_warnings=False):
```
shubhvachher: Actually, calculating difference in sets (which is needed to generate this warning) can be an unnecessary overhead for people using this code in a production environment. I have added a `logging.info` call to inform users of the same.
Sometimes it is not possible to remove all non-vocabulary words from the `words-list`, for example, if `words-list` is generated during runtime. Here using `suppress_warnings` would be ideal.

gensim/models/keyedvectors.py
```diff
@@ -336,6 +346,106 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
...
+            raise ValueError("words_list must be a set/list of words. Maybe you wanted the most_similar function.")
+        elif isinstance(words_list, list) or isinstance(words_list, set):
+            words_list = set(words_list)
+            vocabulary_words = set(self.index2word)
+
+            words_to_use = vocabulary_words.intersection(words_list)
+
+            if not suppress_warnings:
+                missing_words = words_list.difference(vocabulary_words)
+                if(not missing_words): # missing_words is empty
```
shubhvachher: Changed for every occourance

gensim/models/word2vec.py
```diff
@@ -355,6 +355,8 @@ class Word2Vec(utils.SaveLoad):
     """
     Class for training, using and evaluating neural networks described in https://code.google.com/p/word2vec/
 
+    If you're finished training a model (=no more updates, only querying), then switch to the :mod:`gensim.models.KeyedVectors` instance in wv
```
shubhvachher: Tried following pep8, please do check now

gensim/models/word2vec.py
```diff
@@ -1181,33 +1187,83 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
         logger.info("merged %d vectors into %s matrix from %s" % (overlap_count, self.wv.syn0.shape, fname))
 
     def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, indexer=None):
+        """
+        Please refer to the documentation for `gensim.models.KeyedVectors.most_similar`
+        In the future please try and use the `gensim.models.KeyedVectors` instance in wv
```
shubhvachher: Changed this and split into a separate PR!

gensim/test/test_sklearn_integration.py
```diff
@@ -105,7 +105,7 @@ def testPipeline(self):
         text_lda = Pipeline((('features', model,), ('classifier', clf)))
         text_lda.fit(corpus, data.target)
         score = text_lda.score(corpus, data.target)
-        self.assertGreater(score, 0.50)
+        self.assertGreater(score, 0.40)
```
shubhvachher: Split into a separate PR! Thank you so much. Taught me a lot about git and got me my first commit :)

gensim/test/test_word2vec.py
shubhvachher: Sure, Removed extra tests; using CBOW method to check functionality.

That second part might be important given @tmylk's comment
> KeyedVectors is the way to go.

What do you think @tmylk ?

gensim/test/test_word2vec.py
shubhvachher: logging is used. Suppress warnings becomes useful in case `words-list` is generated on the go and the user does not want to know about missing words.

gensim/models/word2vec.py
shubhvachher: I have added only this method to class `Word2Vec` in order to solve the original "people might not find index2word" problem.

gensim/models/keyedvectors.py
```diff
@@ -338,6 +345,129 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
...
+        else:  # This is triggered for empty words_list parameter
+            raise ValueError("words_list must be set/list of words. " \
+                                "Maybe you wanted the most_similar function. " \
+                                "Please read doc string")
+
+        if type(topn) is not int:
+            if topn is False:
+                pass
+            else:
+                if suppress_warnings is False:
```
tmylk: This must be an exception, not a warning. Incorrect input can't be surpressed.

gensim/models/keyedvectors.py
```diff
@@ -338,6 +345,129 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
...
+            missing_words = words_list.difference(vocabulary_words)
+            if not missing_words:  # missing_words is empty
+                pass
+            else:
+                logger.warning("The following words are not in " \
+                                "trained vocabulary : %s", str(missing_words))
+                logger.info("This warning is expensive to calculate, " \
+                                "especially for large words_list. " \
+                                "If you would rather not remove the missing_words " \
+                                "from words_list please set the " \
```
tmylk: Better message is "Please intersect with vocabulary words_to_use = vocabulary_words.intersection(words_list) prior to calling the most_similar_among". 
Please remove the `suppress_warnings` flag.

gensim/models/keyedvectors.py
```diff
@@ -338,6 +345,129 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
...
+                                "suppress_warnings flag.")
+
+        words_list_indices = [self.vocab[word].index for word in words_to_use]
+        # limited = self.syn0norm[words_list_indices]
+        # Storing 'limited' might add a huge memory overhead so we avoid doing that
```
tmylk: Please memory profile this code to provide foundation for this statement.
Please remove commented out code.

gensim/models/keyedvectors.py
```diff
@@ -338,6 +345,129 @@ def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, i
...
+
+        Example::
+
+          >>> trained_model.most_similar_among(positive=['man'], topn=1,
+                                                words_list=['woman','random_word'])
+          [('woman', 0.75882536)]
+
+        """
+
+        if isinstance(words_list, int):
```
tmylk: Please use a single check and a single  `raise ValueError`. 
The `most_similar` function doesn't take a list of `int`s so it should not be mentioned here

tmylk:
Ping @shubhvachher . It is a useful functionality, would be good to finish this PR

shubhvachher:
Yup, sorry! Finishing up exams here. On this asap.

menshikh-iv:
Ping @shubhvachher, what status of this PR? Will you finish it soon?

shubhvachher:
Yes @menshikh-iv; my bad. Been travelling and working. I do have access to my laptop though. I'd get on this coming weekend if thats ok!

menshikh-iv:
Ok @shubhvachher, we will wait.

menshikh-iv:
ping @shubhvachher

menshikh-iv:
I close current PR because this looks abandoned.

shubhvachher:
This is surely my fault for lack of time from my summer work and travels; Will open this again as soon as I get the chance and if uncleared. Apologies,
Shubh

piskvorky:
@shubhvachher please do. The (semi-automated) closing of stale tickets is a maintenance thing from our side, not an indication we're not interested in the PR!

PR #1234, opened 2017-03-22
Fix #1230. Fix word2vec reset_from bug in v1.0.1 and added unittest

### LINKED ISSUE #1230: 1.0.1 reset_from bug
code should be:

def reset_from(self, other_model):
        """
        Borrow shareable pre-built structures (like vocab) from the other_model. Useful
        if testing multiple models in parallel on the same corpus.
        """
        self.wv.vocab = other_model.wv.vocab
        self.wv.index2word = other_model.wv.index2word

Kreiswolke:
Fixed reset_from attribute bug in gensim 1.0.1 version.

### DIFF
--- a/gensim/test/test_word2vec.py
+++ b/gensim/test/test_word2vec.py
@@ -678,6 +678,13 @@ class TestWord2VecModel(unittest.TestCase):
     def testLoadOnClassError(self):
         """Test if exception is raised when loading word2vec model on instance"""
         self.assertRaises(AttributeError, load_on_instance)
+
+    def test_reset_from(self):
+        """Test if exception is raised when reset_from is used"""
+        model = word2vec.Word2Vec(sentences, min_count=1)
+        model2 = word2vec.Word2Vec(new_sentences, min_count=1)
+        self.assertRaises(AttributeError, model.reset_from(model2))
+
 #endclass TestWord2VecModel
 
 class TestWMD(unittest.TestCase):
--- a/gensim/test/test_word2vec.py
+++ b/gensim/test/test_word2vec.py
@@ -683,7 +683,10 @@ class TestWord2VecModel(unittest.TestCase):
         """Test if exception is raised when reset_from is used"""
         model = word2vec.Word2Vec(sentences, min_count=1)
         model2 = word2vec.Word2Vec(new_sentences, min_count=1)
-        self.assertRaises(AttributeError, model.reset_from(model2))
+        try:
+            model.reset_from(model2)
+        except AttributeError:
+            self.fail('AttributeError in reset_from()')
 
 #endclass TestWord2VecModel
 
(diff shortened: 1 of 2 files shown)


tmylk:
Please fix the typo to make unit tests pass.

Kreiswolke:
Typo fixed, tests passed.

gensim/test/test_word2vec.py
```diff
@@ -678,6 +678,13 @@ def test_sentences_should_not_be_a_generator(self):
     def testLoadOnClassError(self):
         """Test if exception is raised when loading word2vec model on instance"""
         self.assertRaises(AttributeError, load_on_instance)
+
+    def test_reset_from(self):
+        """Test if exception is raised when reset_from is used"""
+        model = word2vec.Word2Vec(sentences, min_count=1)
+        model2 = word2vec.Word2Vec(new_sentences, min_count=1)
+        self.assertRaises(AttributeError, model.reset_from(model2))
```
tmylk: why does it raise an error? A positive outcome is expected. Please change this test to a positive test.

gensim/test/test_word2vec.py
```diff
@@ -678,6 +678,16 @@ def test_sentences_should_not_be_a_generator(self):
...
         self.assertRaises(AttributeError, load_on_instance)
+
+    def test_reset_from(self):
+        """Test if exception is raised when reset_from is used"""
+        model = word2vec.Word2Vec(sentences, min_count=1)
+        model2 = word2vec.Word2Vec(new_sentences, min_count=1)
+        try:
+            model.reset_from(model2)
+        except AttributeError:
+            self.fail('AttributeError in reset_from()')
```
tmylk: what is the purpose of this line? what would be different if it was removed?

gensim/test/test_word2vec.py
```diff
@@ -678,6 +678,16 @@ def test_sentences_should_not_be_a_generator(self):
     def testLoadOnClassError(self):
         """Test if exception is raised when loading word2vec model on instance"""
         self.assertRaises(AttributeError, load_on_instance)
+
+    def test_reset_from(self):
+        """Test if exception is raised when reset_from is used"""
```
tmylk: is there some reason that exception should be raised when reset_from is used? correct behavious is no exception. please add checks that the model is actually reset.

gensim/test/test_word2vec.py
```diff
@@ -683,7 +683,10 @@ def test_reset_from(self):
         """Test if exception is raised when reset_from is used"""
         model = word2vec.Word2Vec(sentences, min_count=1)
         model2 = word2vec.Word2Vec(new_sentences, min_count=1)
-        self.assertRaises(AttributeError, model.reset_from(model2))
+        try:
+            model.reset_from(model2)
+        except AttributeError:
+            self.fail('AttributeError in reset_from()')
```
gojomo: Because an uncaught error will result in a test failure, this catch-then-fail seems superfluous - it's enough to try the `reset_from()`, and do some afterwards checking that it's had the intended effect.

Kreiswolke: I agree, it is not needed and I just updated the test.

tmylk:
Thanks for the fix!

Kreiswolke:
Happy I could help ;)

Issue #1019, opened 2016-11-14
Segmentation fault when performing online learning for Doc2Vec


danoneata:
Hello!

I'm performing online learning for `Doc2Vec`, that is, I learn an initial model on a set of tagged documents and try to update the model on a new set of tagged documents. If the second set contains new tags (tags that were not present in the initial set of documents), then I usually get segmentation fault (this behavior is not deterministic, but it happens most of time).

Below you can find a toy example that reproduces the issue; and [here](http://pastebin.com/dNUyAnAD) is the output of that code. I'm using Python 3.4.3 and Gensim 0.13.3.

I've debugged with gdb and I've got the following output:

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff9a4f8700 (LWP 29422)]
__pyx_f_6gensim_6models_13doc2vec_inner_fast_document_dm_hs (__pyx_v_learn_hidden=1, __pyx_v_size=300, __pyx_v_work=0x7fff80001480, __pyx_v_alpha=0.0250000004, __pyx_v_syn1=0x1693ce0, __pyx_v_neu1=0x7fff80001a00, __pyx_v_word_code_len=6,
    __pyx_v_word_code=<optimized out>, __pyx_v_word_point=0x13fe410) at ./gensim/models/doc2vec_inner.c:2078
```

I'm willing to help fixing this issue if someone can provide me some guidance. Thanks!

Sample code that reproduces the issue:

```python
import logging

from gensim.models.doc2vec import (
    Doc2Vec,
    TaggedDocument,
)

logging.basicConfig(
    format='%(asctime)s : %(threadName)s : %(levelname)s : %(message)s',
    level=logging.DEBUG,
)


def to_str(d):
    return ", ".join(d.keys())


SENTS = [
    "anecdotal using a personal experience or an isolated example instead of a sound argument or compelling evidence",
    "plausible thinking that just because something is plausible means that it is true",
    "occam razor is used as a heuristic technique discovery tool to guide scientists in the development of theoretical models rather than as an arbiter between published models",
    "karl popper argues that a preference for simple theories need not appeal to practical or aesthetic considerations",
    "the successful prediction of a stock future price could yield significant profit",
]

SENTS = [s.split() for s in SENTS]


def main():
    sentences_1 = [
        TaggedDocument(SENTS[0], tags=['SENT_0']),
        TaggedDocument(SENTS[1], tags=['SENT_0']),
[...]

PR #1239, opened 2017-03-24
Hotfix made #1019

Arcady27:
With parameter _hs_ of model set to 0 there are no segmentation faults (Gensim 0.13.3, both Python 2. and 3. , proved with ddd). So, proposed hotfix is to turn off _hs_ mode when model is upgraded.

For #1019


tmylk:
The segfault only happens in doc2vec codepath so this restriction is too strict.

gojomo:
Also if someone intends a hierarchical-softmax model, and indeed has `negative=0` so no negative-sampling happens, just turning off `hs`, making the model completely non-functional, is just breaking things a different way.

PR #1241, opened 2017-03-25
fix the word2vec's reset_from()

robotcator:


gensim/test/test_word2vec.py
```diff
@@ -644,7 +644,15 @@ def testTrainWarning(self, l):
...
         self.assertTrue(warning in str(l))
-    
+
+    def test_reset_from(self):
+        models = [
+            word2vec.Word2Vec(size=2, min_count=1, hs=1, negative=0),
+            word2vec.Word2Vec(size=4, min_count=1, hs=1, negative=0),
+        ]
+        models[0].build_vocab(sentences)
+        models[1].reset_from(models[0])
```
tmylk: missing assert that models are indeed the same

tmylk:
Duplicate of #1241

PR #1244, opened 2017-03-27
[WIP] scikit_learn wrapper for LSI Model in Gensim

chinmayapancholi13:
This PR adds a scikit-learn wrapper for LSI Model in Gensim.

### DIFF
--- a/gensim/sklearn_integration/sklearn_wrapper_gensim_lsimodel.py
+++ b/gensim/sklearn_integration/sklearn_wrapper_gensim_lsimodel.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2011 Radim Rehurek <radimrehurek@seznam.cz>
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+#
+"""
+Scikit learn interface for gensim for easy use of gensim with scikit-learn
+Follows scikit-learn API conventions
+"""
+import numpy as np
+
+from gensim import models
+from gensim import matutils
+from scipy import sparse
+from sklearn.base import TransformerMixin, BaseEstimator
+
+# accuracy defaults for the multi-pass stochastic algo
+P2_EXTRA_DIMS = 100  # set to `None` for dynamic P2_EXTRA_DIMS=k
+P2_EXTRA_ITERS = 2
+
+class SklearnWrapperLsiModel(models.LsiModel, TransformerMixin, BaseEstimator):
+    """
+    Base LSI module
+    """
+
+    def __init__(self, corpus=None, num_topics=200, id2word=None, chunksize=20000,
+                 decay=1.0, onepass=True, power_iters=P2_EXTRA_ITERS, extra_samples=P2_EXTRA_DIMS):
+        """
+        Sklearn wrapper for LSI model. Class derived from gensim.model.LsiModel.
+        """
+        self.corpus = corpus
+        self.num_topics = num_topics
+        self.id2word = id2word
+        self.chunksize = chunksize
+        self.decay = decay
+        self.onepass = onepass
+        self.extra_samples = extra_samples
+        self.power_iters = power_iters
+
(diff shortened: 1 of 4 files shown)


gensim/sklearn_integration/sklearn_wrapper_gensim_lsimodel.py
```diff
@@ -0,0 +1,103 @@
...
+P2_EXTRA_DIMS = 100  # set to `None` for dynamic P2_EXTRA_DIMS=k
+P2_EXTRA_ITERS = 2
+
+class SklearnWrapperLsiModel(models.LsiModel, TransformerMixin, BaseEstimator):
+    """
+    Base LSI module
+    """
+
+    def __init__(self, corpus=None, num_topics=200, id2word=None, chunksize=20000,
+                 decay=1.0, onepass=True, power_iters=P2_EXTRA_ITERS, extra_samples=P2_EXTRA_DIMS):
```
tmylk: please make defaults explicits. what is the reasons for using constant variables?

chinmayapancholi13: @tmylk Thanks for pointing this out. I'll make the default values explicit here.

gensim/test/test_sklearn_integration.py
```diff
@@ -107,5 +108,55 @@ def testPipeline(self):
         score = text_lda.score(corpus, data.target)
         self.assertGreater(score, 0.50)
 
+class TestSklearnLSIWrapper(unittest.TestCase):
+    def setUp(self):
+        self.model = SklearnWrapperLsiModel(id2word=dictionary, num_topics=2)
+        self.model.fit(corpus)
+
+    def testPrintTopic(self):
+        topic = self.model.print_topics(2)
```
chinmayapancholi13: @tmylk My rationale for having this test was basically to serve as a sanity check after `setUp()` prepares the test fixture. Since such a test was present among the unit tests for the sklearn wrapper written for LDA model as well, I decided to include it here too. Please let me know if we should drop this test.

tmylk: please rename as testModelSanity then

tmylk:
Great, please add this example to https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/sklearn_wrapper.ipynb

chinmayapancholi13:
@tmylk Sure. Adding LsiModel wrapper example to the IPython notebook.

chinmayapancholi13:
@tmylk For LsiModel, I have added an example which uses `pipeline.Pipeline`. I was initially planning to add examples like the isolated usage (without scikit_learn) of the wrapper and usage with only `CountVectorizer`. But such cases would be exactly congruent to the examples given for LdaModel and would lead to unnecessary repetition. Please let me know if we should add such examples in the notebook.

gensim/test/test_sklearn_integration.py
```diff
@@ -107,5 +108,55 @@ def testPipeline(self):
...
+
+    def testPipeline(self):
+        model = SklearnWrapperLsiModel(num_topics=2)
+        with open(datapath('mini_newsgroup'),'rb') as f:
+            compressed_content = f.read()
+            uncompressed_content = codecs.decode(compressed_content, 'zlib_codec')
+            cache = pickle.loads(uncompressed_content)
+        data = cache
+        id2word=Dictionary(map(lambda x : x.split(), data.data))
+        corpus = [id2word.doc2bow(i.split()) for i in data.data]
```
tmylk: Please respond to this comment.

tmylk:
Thanks for the new wrapper!

gensim/test/test_sklearn_integration.py
chinmayapancholi13: @tmylk My sincerest apologies for the delayed response. I'd be happy to submit a PR to resolve this issue. I did not add code for this in the current PR because some of the code for the wrapper for LdaModel might also have to be refactored. So maintaining modularity and proposing a solution in a different PR seemed to be a better choice in this case.

PR #1247, opened 2017-03-29
glove2word2vec improvements

parulsethi:
This PR combines the 'file read/write' code for both python 2 and 3, in glove2word2vec conversion.

Also, removed tests from end of the conversion script and added it to the previously added [test_glove2word2vec.py](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_glove2word2vec.py)


tmylk:
Thank you for the improvement. 
For better tests please add some non-ascii characters to https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_data/test_glove.txt

parulsethi:
Sure, added few non-ascii characters in the beginning of file.

PR #1248, opened 2017-03-30
[MRG] Keras wrapper for Word2Vec model in Gensim

chinmayapancholi13:
This PR adds a Keras wrapper for Word2Vec Model in Gensim.


chinmayapancholi13:
@tmylk I have tried to use the wrapper for a smaller version of the [20NewsGroup](http://qwone.com/~jason/20Newsgroups/) task. The code used is as follows. (It is based on the code used [here](https://blog.keras.io/using-pre-trained-word-embeddings-in-a-keras-model.html))

```python

from __future__ import print_function

import os
import sys
import numpy as np
from gensim.sklearn_integration.keras_wrapper_gensim_word2vec import KerasWrapperWord2VecModel
from keras.engine import Input
from keras.layers import merge
from keras.preprocessing.text import Tokenizer
from keras.preprocessing.sequence import pad_sequences
from keras.utils.np_utils import to_categorical
from keras.layers import Dense, Input, Flatten
from keras.layers import Conv1D, MaxPooling1D, Embedding
from keras.models import Model

BASE_DIR = ''
TEXT_DATA_DIR = BASE_DIR + './path/to/text/data/dir'
MAX_SEQUENCE_LENGTH = 1000
EMBEDDING_DIM = 100
VALIDATION_SPLIT = 0.2
BATCH_SIZE = 128

# prepare text samples and their labels

texts = []  # list of text samples
labels_index = {}  # dictionary mapping label name to numeric id
labels = []  # list of label ids

for name in sorted(os.listdir(TEXT_DATA_DIR)):
    path = os.path.join(TEXT_DATA_DIR, name)
    if os.path.isdir(path):
        label_id = len(labels_index)
        labels_index[name] = label_id
        for fname in sorted(os.listdir(path)):
            if fname.isdigit():
                fpath = os.path.join(path, fname)
                if sys.version_info < (3,):
                    f = open(fpath)
                else:
                    f = open(fpath, encoding='latin-1')
                t = f.read()
                i = t.find('\n\n')  # skip header
                if 0 < i:
                    t = t[i:]
                texts.append(t)
                f.close()
                labels.append(label_id)

# vectorize the text samples into a 2D integer tensor
tokenizer = Tokenizer()
tokenizer.fit_on_texts(texts)
sequences = tokenizer.texts_to_sequences(texts)

word_index = tokenizer.word_index
data = pad_sequences(sequences, maxlen=MAX_SEQUENCE_LENGTH)
labels = to_categorical(np.asarray(labels))

# split the data into a training set and a validation set
indices = np.arange(data.shape[0])
np.random.shuffle(indices)
data = data[indices]
labels = labels[indices]
num_validation_samples = int(VALIDATION_SPLIT * data.shape[0])

x_train = data[:-num_validation_samples]
y_train = labels[:-num_validation_samples]
x_val = data[-num_validation_samples:]
y_val = labels[-num_validation_samples:]

# train the embedding matrix
data1 = word2vec.LineSentence('./path/to/input/data')
Keras_w2v = KerasWrapperWord2VecModel(data1, min_count=1)
embedding_layer = Keras_w2v.get_embedding_layer()

sequence_input = Input(shape=(MAX_SEQUENCE_LENGTH,), dtype='int32')
embedded_sequences = embedding_layer(sequence_input)

x = Conv1D(128, 5, activation='relu')(embedded_sequences)
x = MaxPooling1D(5)(x)
x = Conv1D(128, 5, activation='relu')(x)
x = MaxPooling1D(5)(x)
x = Conv1D(128, 5, activation='relu')(x)
x = MaxPooling1D(35)(x)  # global max pooling
x = Flatten()(x)
x = Dense(128, activation='relu')(x)
preds = Dense(len(labels_index), activation='softmax')(x)

model = Model(sequence_input, preds)
model.compile(loss='categorical_crossentropy',
              optimizer='rmsprop',
              metrics=['acc'])

model.fit(x_train, y_train, validation_data=(x_val, y_val), batch_size=BATCH_SIZE)

```

tmylk:
Please add unit tests and an ipynb

chinmayapancholi13:
@tmylk Sure.

chinmayapancholi13:
@tmylk I have added unit tests for word similarity task (using cosine distance) as well as a smaller version of the [20NewsGroups](http://qwone.com/~jason/20Newsgroups/) classification task. I have also created an IPython notebook for the wrapper explaining both these examples.

docs/notebooks/datasets/20_newsgroup_keras/alt.atheism/49960
```diff
@@ -0,0 +1,58 @@
+Atheist Resources
+Addresses of Atheist Organizations
+USA
+FREEDOM FROM RELIGION FOUNDATION
+Darwin fish bumper stickers and assorted other atheist paraphernalia are available from the Freedom From Religion Foundation in the US.
```
tmylk: Please put data into `test/test_data`

gensim/keras_integration/__init__.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2011 Radim Rehurek <radimrehurek@seznam.cz>
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+"""Keras wrappers for gensim.
```
tmylk: "Wrapper to allow gensim word2vec as input into Keras." is more clear. And in other docstrings and ipynbs too

tmylk:
Thanks for the feature. The change seems to be so small - just 5 lines in `get_embedding_layer()`

Instead of creating a new class, could you please just add this one method to `KeyedVectors`?

chinmayapancholi13:
@tmylk Sure. I'll add the function in `KeyedVectors` instead of creating a new class.

docs/notebooks/datasets/20_newsgroup_keras/alt.atheism/49960
chinmayapancholi13: @tmylk The data used in the unit tests is present in `test/test_data` folder already. The data used in the IPython notebooks is present in `docs/notebooks/datasets` folder. We are using the same data at both places so to avoid unnecessary duplication, should I use the data in `test/test_data` (i.e. set the path accordingly in the ipynb) in the ipynb notebooks as well?

gensim/keras_integration/__init__.py
chinmayapancholi13: @tmylk The file `__init__.py` is common for the entire folder `gensim/keras_integration` (which in turn would have the files for integration of the various models with Keras). So shouldn't the docstring here be more generic (like it already is)?

docs/notebooks/datasets/20_newsgroup_keras/alt.atheism/49960
tmylk: yes, `test/test_data` is a better location for data used in both

gensim/keras_integration/__init__.py
tmylk: It can be more general, just pointing out the difference in meaning between "keras wrapper for gensim" vs "gensim wrapper for keras". Which one is it?
I think it is a "Wrapper to allow gensim models as input into Keras."

tmylk:
@chinmayapancholi13 this is more appropriate to have in KeyedVectors class.
And to have an example integration with Classification  in https://github.com/stephenhky/PyShortTextCategorization/blob/db246e3ade2fcdea58953ff807d259464765a661/shorttext/classifiers/embed/nnlib/VarNNEmbedVecClassification.py

docs/notebooks/datasets/20_newsgroup_keras/alt.atheism/49960
chinmayapancholi13: Got it. So, I'll change the path set in the ipynb for this functionality.

gensim/keras_integration/__init__.py
chinmayapancholi13: Okay. Understood it now. Thanks for pointing this out. So, I'll change "Keras wrappers for gensim" to "Wrappers to allow gensim models as input into Keras".

chinmayapancholi13:
@tmylk I have moved the function `get_embedding_layer` to `keyedvectors.py`. I am adding the classification example now.
Also, the tests are failing because of PEP8 checking being done in the **test data**. Is this what we expect?

chinmayapancholi13:
@tmylk I have added the classification example as discussed. As I had mentioned above, the tests are failing because of PEP8 checks for the **test data**.

tmylk:
you can add `txt` to flake8 ignore list like in https://github.com/RaRe-Technologies/gensim/pull/1309

chinmayapancholi13:
@tmylk Thanks! :) I have added `.txt` and `.csv` to flake8 ignore list.

gensim/models/keyedvectors.py
```diff
@@ -78,6 +78,7 @@
 from six import string_types, iteritems
 from six.moves import xrange
 from scipy import stats
+from keras.layers import Embedding
```
tmylk: this is  an optional dependency. please change into a flag

gensim/test/test_keras_integration.py
```diff
@@ -0,0 +1,147 @@
...
+    ['graph', 'minors', 'survey']
+]
+
+module_path = os.path.dirname(__file__)  # needed because sample data files are located in the same folder
+datapath = lambda fname: os.path.join(module_path, 'test_data', fname)
+
+
+class TestKerasWord2VecWrapper(unittest.TestCase):
+    def setUp(self):
+        self.model_cos_sim = word2vec.Word2Vec(sentences, size=100, min_count=1, hs=1)
```
tmylk: Please skip tests if keras not installed

gensim/test/test_keras_integration.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # test querying for "most similar" by vector
+        graph_vector = model.wv.syn0norm[model.wv.vocab['graph'].index]
+        sims2 = model.most_similar(positive=[graph_vector], topn=11)
+        sims2 = [(w, sim) for w, sim in sims2 if w != 'graph']  # ignore 'graph' itself
+        self.assertEqual(sims, sims2)
+
+    def testEmbeddingLayerCosineSim(self):
+        """
+        Test Keras 'Embedding' layer returned by 'get_embedding_layer' function for a simple word similarity task.
```
tmylk: How much memory does it take? Is it doubling up? Please run memory profiler and add a cautionary note

gensim/models/keyedvectors.py
chinmayapancholi13: Sure. Added a flag to check if Keras is installed or not. If not installed and function `get_embedding_layer()` called, then it raises an `ImportError`.

gensim/test/test_keras_integration.py
chinmayapancholi13: Done. Added a check to ensure that if Keras is not installed then the tests are skipped with a suitable error mesage.

.travis.yml
```diff
@@ -19,6 +19,9 @@ install:
   - pip install testfixtures
   - pip install unittest2
   - pip install scikit-learn
+  - pip install theano
+  - pip install tensorflow
+  - pip install keras
```
tmylk: please add them to test_requires in setup.py

gensim/test/test_keras_integration.py
tmylk: Any update?

tmylk:
comments about ipynb: 

rename `model_wv` as `wv`

cosine of 2 vectors is not the probability of them occuring together.

Please fix `The `Merge` layer is deprecated and will be removed after 08/2017`

Please compare the result of keras cos model with a simple `wv[word_b].dot(wv[word_a])`. it should be the same.

In the final cell the score of `{'mathematics': 0.97023982,` is very good. Please call it a good result!

gensim/test/test_keras_integration.py
chinmayapancholi13: Sorry for the delay in doing this. I was first trying to finish the coding part of the PR. I'll do the memory profiling on this now and post the updates very soon.

.travis.yml
chinmayapancholi13: The discussion [here](https://github.com/RaRe-Technologies/gensim/pull/1336/files/f0ba459269460fe24d527550f223a90924688faa#r117319253) seems to suggest that we actually use the same list for `test_env` and `tests_require` in `setup.py`. So, I have added these 3 requirements to `test_env`. 
Please let me know if I have misunderstood this.

chinmayapancholi13:
@tmylk Thanks a lot for your feedback. I have incorporated your suggestions as follows : 
- **rename `model_wv` as `wv`** : Done

- **cosine of 2 vectors is not the probability of them occuring together** : I have replaced the comment with `output is the cosine distance between the two words (as a similarity measure)`

- **Please fix `The Merge layer is deprecated and will be removed after 08/2017`** : I have replaced `merge` with `dot` function (with param `normalize` set to `True` for taking the cosine distance). The particular warning is no longer there.

- **Please compare the result of keras cos model with a simple `wv[word_b].dot(wv[word_a])`. it should be the same** : The values are the same when we set `normalize` as `False` in the `dot` function. This is because in such a case the output value is only the dot product (and not the cosine distance) of the two vectors. 
Would we want this check to be explicitly added somewhere in the unit-tests or the ipynb notebook? Or was this checking only for our verification of the behavior of the Keras function?

- **In the final cell the score of {'mathematics': 0.97023982, is very good. Please call it a good result!** : I have updated the comment in the last cell.

gensim/test/test_keras_integration.py
chinmayapancholi13: @tmylk Since we are not using any new extra array/matrix in `get_embedding_layer()`, no extra memory is being used because of the function. Thus, as expected, memory profiler shows no extra memory usage because of the function.

.travis.yml
menshikh-iv: You are right @chinmayapancholi13, add this to [this list](https://github.com/RaRe-Technologies/gensim/blob/develop/setup.py#L229)

chinmayapancholi13: @menshikh-iv Thanks! Done.

gensim/keras_integration/__init__.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
tmylk: This file and folder are no longer needed

gensim/test/test_keras_integration.py
tmylk: There is a new keras object `Embedding` - when is it initialised? Expect it to take at least some memory

chinmayapancholi13: The new Keras object `Embedding` is initialised within the `get_embedding_layer()` function as : 

```python
weights = self.syn0
layer = Embedding(input_dim=weights.shape[0], output_dim=weights.shape[1], weights=[weights])
```

But it shouldn't be taking significant memory. On running the `memory_profiler` with 100 MB input text file (text8 example), the particular line (or the function overall) doesn't show any extra memory used (probably because the memory used in the particular line is less than 1 MB). Should I try profiling the function with higher size inputs?

gensim/keras_integration/__init__.py
chinmayapancholi13: Right! Removing this folder right away.

gensim/test/test_keras_integration.py
tmylk: it's probably lazy evaluation

setup.py
```diff
@@ -233,6 +233,9 @@ def finalize_options(self):
     'scikit-learn',
     'pyemd',
     'annoy',
+    'theano',
```
tmylk: would only one of them be enough? tf is preferred

gensim/test/test_keras_integration.py
chinmayapancholi13: Here is the relevant code's link for later reference as well : https://github.com/fchollet/keras/blob/master/keras/engine/topology.py#L310

tmylk:
Let's add a note about shorttext to ipynb.

And then LGTM

menshikh-iv:
Need to fix `.travis` file (add keras dependencies in another branch, merge it and re-run Travis tests in this PR)

chinmayapancholi13:
@menshikh-iv This is because Keras also needs Tensorflow to be installed (when Tensorflow is the default backend). And we are not installing Tensorflow in `.travis.yml` because if we do `pip install tensorflow`, the 'only CPU-support' version gets installed by default. So if a user already has GPU-supported version of tensorflow (using `pip install tensorflow-gpu`) installed, it would get overwritten. This was also the reason behind removing automatic installation of TF while installation of Keras (see https://github.com/fchollet/keras/issues/5776#issuecomment-286637929).

Thus, we(similar to Keras) expect the users to install Tensorflow by themselves.

menshikh-iv:
@chinmayapancholi13 thank for your clarification, but what will we do with Travis?

chinmayapancholi13:
@menshikh-iv One solution can be to include TF in the installation and add a note that users would have to re-install the TF of their choice again (i.e. in case it gets overwritten). For the tests to get passed in travis, I believe we must install TF somehow so this can be a solution in my opinion.

menshikh-iv:
@chinmayapancholi13 Sounds good for me, let's do it.

menshikh-iv:
@chinmayapancholi13 Great :1st_place_medal:

docs/notebooks/keras_wrapper.ipynb
```diff
@@ -0,0 +1,800 @@
...
+    "    ['graph', 'minors', 'trees'],\n",
+    "    ['graph', 'minors', 'survey']\n",
+    "]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Then, we call the wrapper and pass appropriate parameters."
```
piskvorky: What wrapper?

Here and below, the text refers some "wrapper", but I see no wrapper.

docs/notebooks/keras_wrapper.ipynb
```diff
@@ -0,0 +1,800 @@
...
+     "output_type": "stream",
+     "text": [
+      "[[[ 0.00596689]]]\n"
+     ]
+    }
+   ],
+   "source": [
+    "word_a = 'graph'\n",
+    "word_b = 'trees'\n",
+    "output = keras_model.predict([np.asarray([model.wv.vocab[word_a].index]), np.asarray([model.wv.vocab[word_b].index])])  # output is the cosine distance between the two words (as a similarity measure)\n",
```
piskvorky: The comment would be better moved after the code, on a separate line, to improve readability.

docs/notebooks/keras_wrapper.ipynb
```diff
@@ -0,0 +1,800 @@
...
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# global variables\n",
+    "\n",
+    "nb_filters=1200  # number of filters\n",
```
piskvorky: PEP8: space around the assignment operator (`x = 1`).

Here and elsewhere in the notebook.

docs/notebooks/keras_wrapper.ipynb
```diff
@@ -0,0 +1,800 @@
...
+    "    \"\"\"\n",
+    "    Retrieve the training data in a CSV file, with the first column being the\n",
+    "    class labels, and second column the text data. It returns a dictionary with\n",
+    "    the class labels as keys, and a list of short texts as the value for each key.\n",
+    "    \"\"\"\n",
+    "    df = pd.read_csv(filepath)\n",
+    "    category_col, descp_col = df.columns.values.tolist()\n",
+    "    shorttextdict = defaultdict(lambda : [])\n",
+    "    for category, descp in zip(df[category_col], df[descp_col]):\n",
+    "        if type(descp)==str:\n",
```
piskvorky: Will this work across Python 2 / Python 3?

docs/notebooks/keras_wrapper.ipynb
```diff
@@ -0,0 +1,800 @@
...
+    "            shorttextdict[category] += [descp]\n",
```
piskvorky: `append` simpler, faster and more readable?

docs/notebooks/keras_wrapper.ipynb
```diff
@@ -0,0 +1,800 @@
...
+    "\n",
+    "def retrieve_csvdata_as_dict(filepath):\n",
+    "    \"\"\"\n",
+    "    Retrieve the training data in a CSV file, with the first column being the\n",
+    "    class labels, and second column the text data. It returns a dictionary with\n",
+    "    the class labels as keys, and a list of short texts as the value for each key.\n",
+    "    \"\"\"\n",
+    "    df = pd.read_csv(filepath)\n",
+    "    category_col, descp_col = df.columns.values.tolist()\n",
+    "    shorttextdict = defaultdict(lambda : [])\n",
```
piskvorky: `defaultdict(list)`

docs/notebooks/keras_wrapper.ipynb
```diff
@@ -0,0 +1,800 @@
...
+    "        if type(descp)==str:\n",
+    "            shorttextdict[category] += [descp]\n",
+    "    return dict(shorttextdict)\n",
+    "\n",
+    "def subjectkeywords():\n",
+    "    \"\"\"\n",
+    "    Return an example data set, with three subjects and corresponding keywords.\n",
+    "    This is in the format of the training input.\n",
+    "    \"\"\"\n",
+    "    data_path = './datasets/keras_classifier_training_data.csv'\n",
```
piskvorky: `is.path.join` better (will work on Windows too).

Here and elsewhere.

docs/notebooks/keras_wrapper.ipynb
```diff
@@ -0,0 +1,800 @@
...
+    "    scoredict[classlabel] = predictions[0][idx]\n",
+    "\n",
+    "print scoredict"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The result above clearly suggests (~ 98% probability!) that the input `artificial intellegence` should belong to the category `mathematics`, which conforms very well with the expected output in this case.\n",
```
piskvorky: `intellegence` => `intelligence`

chinmayapancholi13:
@piskvorky Thanks a lot for your comprehensive feedback! :) I'd be happy to make these changes in a new PR. I'll also try to keep in mind these code-style issues in the future.

docs/notebooks/keras_wrapper.ipynb
chinmayapancholi13: Yes. It is working fine for both Python 2 and 3.

PR #1251, opened 2017-03-30
Docs for word2vec.py forwarding functions and one more

shubhvachher:
Do you think we should add depreciated warnings to these forwarding functions as well? @gojomo.

Quoting @tmylk for `Word2Vec` class.
> Please avoid adding new forwarding functions. KeyedVectors is the way to go.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1190,7 +1190,8 @@ class Word2Vec(utils.SaveLoad):
 
     def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, indexer=None):
         """
-        Please refer to the documentation for `gensim.models.KeyedVectors.most_similar`
+        Please refer to the documentation for
+        `gensim.models.KeyedVectors.most_similar`
         This is just a forwarding function.
         In the future please use the `gensim.models.KeyedVectors` instance in wv
         """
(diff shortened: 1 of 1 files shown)


shubhvachher:
A lot of new people consider the documentation for gensim to be incomplete or done poorly. This should help fix that for the Word2Vec class.

gensim/models/word2vec.py
```diff
@@ -1181,33 +1189,103 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
         logger.info("merged %d vectors into %s matrix from %s" % (overlap_count, self.wv.syn0.shape, fname))
 
     def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, indexer=None):
+        """
+        Please refer to the documentation for
+        `gensim.models.KeyedVectors.most_similar`
+        This is just a forwarding function.
+        In the future please use the `gensim.models.KeyedVectors` instance in wv
```
piskvorky: Strange line breaks + docstring a bit obscure. I'm not sure reading this would help me if I didn't already know what it's talking about.

How about simply `Deprecated; use self.ww.most_similar() instead.`?

Same with all the other "forwarding" functions.

shubhvachher: Hmm, The strange line breaks were added by me in the [fix pep8](https://github.com/RaRe-Technologies/gensim/pull/1251/commits/73396652986654e19e1ad6f415f00ce5324f5c73) commit! I'm new to following pep8 all the time, do you think I overdid it there?

Is a depreciated warning to be added to these methods? Will they become unavailable in a later version? If yes, then this should surely be changed asap and I'll send a new PR with the same.

piskvorky: It's a question for @tmylk , but I think we want to nudge users toward using the new API, yes.

The line breaks definitely not good, hurts readability.

shubhvachher: Ok, changes made as suggested. #1274

Thanks for the line breaks tip :+1: . Will line break only if comments go excessively above PEP8 standards!

Issue #1233, opened 2017-03-22
testTrainingSgNegative is possibly incorrect


shubhvachher:
**File :** gensim/test/test_word2vec.py
**Method :** testTrainingSgNegative
```python
    def testTrainingSgNegative(self):
        """Test skip-gram (negative sampling) word2vec training."""
        # to test training, make the corpus larger by repeating its sentences over and over
        # build vocabulary, don't train yet
        model = word2vec.Word2Vec(size=2, min_count=1, hs=0, negative=2)
        model.build_vocab(sentences)
        self.assertTrue(model.wv.syn0.shape == (len(model.wv.vocab), 2))
        self.assertTrue(model.syn1neg.shape == (len(model.wv.vocab), 2))

        model.train(sentences)
        sims = model.most_similar('graph', topn=10)
        # self.assertTrue(sims[0][0] == 'trees', sims)  # most similar

        # test querying for "most similar" by vector
        graph_vector = model.wv.syn0norm[model.wv.vocab['graph'].index]
        sims2 = model.most_similar(positive=[graph_vector], topn=11)
        sims2 = [(w, sim) for w, sim in sims2 if w != 'graph']  # ignore 'graph' itself
        self.assertEqual(sims, sims2)

        # build vocab and train in one step; must be the same as above
        model2 = word2vec.Word2Vec(sentences, size=2, min_count=1, hs=0, negative=2)
        self.models_equal(model, model2)
```

**Lines that read :**  model = word2vec.Word2Vec(size=2, min_count=1, hs=0, negative=2)
**Should read :** model = word2vec.Word2Vec(size=2, min_count=1, sg=1, hs=0, negative=2)

New testMethod does pass unittests.

Is my assumption correct? If so, should I go ahead correcting this line and submitting?

gojomo:
Yes - it should be teting `sg=1`, both for `model` and `model2`, thanks! Do you see any similar mismatches between intent-of-method, and what's actually tested?

shubhvachher:
Well, I'm trying to familiarize myself with your codebase for GSoC! I'm going through it all anyway :D 
Meanwhile, adding fix to PR #1229.

gojomo:
Thanks! Note both sets of changes can be easier to review in separate PRs.

tmylk:
@shubhvachher Thanks for your contribution. Please email me/chat on gitter about GSOC.

shubhvachher:
@gojomo, Thanks for the advice!
[...]

PR #1252, opened 2017-03-30
SgNegative test fixup

shubhvachher:
Fix issue #1233. No problem running new test.

Sorry for the delay; Issue was hung up in a separate PR.


tmylk:
Thanks for the fix.

Issue #1231, opened 2017-03-22
DocvecsArray getitem int64 bug


knownact:
def __getitem__(self, index):
        """
        Accept a single key (int or string tag) or list of keys as input.

        If a single string or int, return designated tag's vector
        representation, as a 1D numpy array.

        If a list, return designated tags' vector representations as a
        2D numpy array: #tags x #vector_size.
        """
        if isinstance(index, string_types + (int,)):
            return self.doctag_syn0[self._int_index(index)]

        return vstack([self[i] for i in index])

should be 

        if isinstance(index, integer):
            return self.doctag_syn0[self._int_index(index)]

gojomo:
Why? Do you have an example where the chabge works better? Does the change still support indexed-lookup of string doctags?

knownact:
it is because index here may be the numpy.int64 type, and the code here will hurt the indexed-lookup of string doctag, so maybe add a addition compare is ok, like:


from numpy import integer:

    if isinstance(index, string_types + (int,)) or isinstance(index, integer + (int,)) :
        return self.doctag_syn0[self._int_index(index)]

gojomo:
I see. A PR with a test for the new case this enables, & the fix, would be welcome.

PR #1254, opened 2017-03-31
Allow indexing with np.int64 in doc2vec - #1231

bogdanteleaga:
This PR attempts to fix https://github.com/RaRe-Technologies/gensim/issues/1231.
It also includes a drive-by fix for a warning that was thrown during tests that asked the `assertEquals` to be changed into an `assertEqual`.

As far as the fix itself goes, I am not completely sure this is sane and am willing to discard the PR. Indexing with `np.int32` works fine with the previous code and if indexing with `np.int64` is going to be allowed here, it might as well be allowed everywhere else, which is a way bigger change.

However, this does fix the original issue as it has been presented.

### DIFF
--- a/gensim/corpora/indexedcorpus.py
+++ b/gensim/corpora/indexedcorpus.py
@@ -18,7 +18,7 @@ of each document.
 """
 
 import logging
-import shelve
+import six
 
 import numpy
 
@@ -124,7 +124,7 @@ class IndexedCorpus(interfaces.CorpusABC):
 
         if isinstance(docno, (slice, list, numpy.ndarray)):
             return utils.SlicedCorpus(self, docno)
-        elif isinstance(docno, (int, numpy.integer)):
+        elif isinstance(docno, six.integer_types + (numpy.integer,)):
             return self.docbyoffset(self.index[docno])
         else:
             raise ValueError('Unrecognised value for docno, use either a single integer, a slice or a numpy.ndarray')
--- a/gensim/models/atmodel.py
+++ b/gensim/models/atmodel.py
@@ -33,13 +33,8 @@ A tutorial can be found at https://github.com/RaRe-Technologies/gensim/tree/deve
 # are included in the code where this is the case, for example in the log_perplexity
 # and do_estep methods.
 
-import pdb
-from pdb import set_trace as st
-from pprint import pprint
-
 import logging
 import numpy as np  # for arrays, array broadcasting etc.
-import numbers
 from copy import deepcopy
 from shutil import copyfile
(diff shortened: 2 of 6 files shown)


tmylk:
Agree that it should be changed throughout doc2vec.
Some of the doc2vec code uses `isinstance(x, integer_types)` which includes `int64` . So suggest  to change all `isinstance(x, int)` in doc2vec to this check. It is more concise than proposed  `isinstance(x, (int, integer))`

bogdanteleaga:
I meant across the codebase, but after a quick search it seems this is the only place where this actually happens.
There's two other spots that do this: one uses `numbers.Integral` and `np.integer`, the other uses `int` and `np.integer`.
Now, for me `numbers.Integral` actually covers all 3 cases, while just `integer_types` fails for numpy(see below). I think `numbers.Integral` should be safe considering [this](https://github.com/scikit-learn/scikit-learn/issues/5786). 

Does the CI run the tests on multiple python versions?

```python
>>> l=[42, np.int32(42), np.int64(42)]
>>> [isinstance(x, numbers.Integral) for x in l]
[True, True, True]
>>> [isinstance(x, six.integer_types) for x in l]
[True, False, False]
```

bogdanteleaga:
Okay, did some more testing and the last commit makes the most sense I think. `integer_types` takes care of regular python stuff and `integer` takes care of numpy integers. It is a bit ugly but it is the only one that was consistent. `numbers.Integral` was still failing for some numpy types on python2.7.

tmylk:
Thanks for the changes across the broader codebase. New tests of `int64` are needed for every changed class.

bogdanteleaga:
Well, at this point I guess it's not just `np.int64` anymore, it's other numpy integer types as well. I'll take a look later this week.

tmylk:
@bogdanteleaga Are you interested in adding tests to this PR?

bogdanteleaga:
Sorry, it's been a really busy week and this one's not looking that great either. But I should be able to do it at the end of this one/start of next one. If anybody else wants to take it up faster than that that's ok as well.

bogdanteleaga:
Added 2 simple tests for indexed corpus and word2vec. Not really sure how to test lda/at. Also I'm not completely sure on how to test other types(if that's wanted). I could yield all the different types and iterate over them, but it could make tests take longer. Let me know what you think.

tmylk:
Thanks for the fix.

I am not sure what is the point of having ` and not isinstance(doc[0][0], six.integer_types + (np.integer,)):` in LDAModel is either.

PR #1255, opened 2017-04-01
most_similar_among functionality added with tests

### LINKED ISSUE #148: word2vec errors with GoogleNews-vectors-negative300.bin model
Using the pretrained model GoogleNews-vectors-negative300.bin I have problems with terms not being found in the model.

./distance GoogleNews-vectors-negative300.bin
internet

produces the below, but using word2vec.py after loading the model, 'internet' in model produces False. Other models seem to work fine.

---

```
                                      Internet      0.834454
                                        online      0.650548
                        Visit_OMNOVA_Solutions      0.649711
                    via_www.mobilebaybears.com      0.647773
                      Earn_Rs.####/day_working      0.629283
                  http://www.mineralstech.com/      0.618483
                            Earn_Rs.####_daily      0.611737
                                           web      0.596768
                              InternetInternet      0.588310
                                       intenet      0.584393
```

shubhvachher:
Fix #148. After @gojomo's comprehensive review from #1229

I have split PR's for all drive-by bug fixes as suggested. Some follow ups from previous comments :

>Since people might not find index2word, I can see the benefit of an accessor method. However, given that this is now a generic KeyedVectors class, the name shouldn't be 'words'-centric. I'd suggest ordered_keys.

Done. I have also added a method to class `Word2Vec` in order to solve the original "people might not find index2word" problem.

>I think suppress_warnings adds extra complication for little benefit.

Actually, calculating difference in sets (which is needed to generate this warning) can be an unnecessary overhead for people using this code in production environment. I have added a `logging.info` call to inform users of the same.
Sometimes it is not possible to remove all non-vocabulary words from the `words-list`, for example, if `words-list` is generated during runtime. Here using `suppress_warnings` would be ideal. 

>Seems more like something to log than use warnings.

Done. Yes, you're right. The module uses logging throughout, unnecessary import and use of warnings was not a good choice.

>Because most_similar_among() functionality is merely on a set of vectors, it's not dependent on the mode-of-training – so there's no need for multiple models built different ways.

Ok sure, removing extra tests.

>In fact, it'd ultimately make even more sense to be inside a set of KeyedVectors tests, rather than Word2Vec tests, but since its sibling methods are tested here it's not necessary to make that big change yet.

Quoted here to keep track.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -272,6 +272,13 @@ class KeyedVectors(utils.SaveLoad):
         else:
             raise KeyError("word '%s' not in vocabulary" % word)
 
+    def get_ordered_keys(self):
+        """
+        Returns the keys in the current KeyedVectors instant as a list.
+        If the model is not trained yet then an empty list is returned.
+        """
+        return self.index2word
+
     def most_similar(self, positive=[], negative=[], topn=10, restrict_vocab=None, indexer=None):
         """
         Find the top-N most similar words. Positive words contribute positively towards the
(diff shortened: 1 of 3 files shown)


tmylk:
Please continue in the original PR

PR #1256, opened 2017-04-01
Added save method for doc2vec

parulsethi:
This PR adds a method `save_doc2vec_format` to save document vectors in similar format as `save_word2vec_format` saves word vectors. 

Another PR [#699](https://github.com/RaRe-Technologies/gensim/pull/699) was opened with this issue and to address @gojomo comment there
> mixing word & doc vectors into the same flattened file on save

I’ve added a flag parameter to indicate whether to save word vectors along with doc vectors. For now, I append word vectors after doc vectors, let me know if some other criteria is more favorable. 

And if someone would like to keep doc and word vectors disentangled as pointed out by @gojomo, they can separately save them to different files now using `save_doc2vec_format` (with flag `word_vec=False`) and then `save_word2vec_format`.

@tmylk  @gojomo  If this seems ok, I’ll add the corresponding load method


gensim/models/doc2vec.py
```diff
@@ -808,6 +808,45 @@ def delete_temporary_training_data(self, keep_doctags_vectors=True, keep_inferen
...
+        if word_vec:
+            total_vec = total_vec + len(self.wv.vocab)
+        logger.info("storing %sx%s projection weights into %s" % (total_vec, self.vector_size, fname))
+
+        # save document vectors
+        with utils.smart_open(fname, 'wb') as fout:
+            fout.write(utils.to_utf8("%s %s\n" % (total_vec, self.vector_size)))
+            # store as in input order
+            for i in range(len(self.docvecs)):
+                doctag = self.docvecs.index_to_doctag(i)
```
parulsethi: this will work in case user's `model.docvecs.doctags` is empty, and will assign the vector index as doctag

gojomo:
Cool! Some thoughts:

* since the *format* written is still the "word2vec_format" used by word2vec.c, the method name should still reflect that
* potentially this could even be a `save_word2vec_format()` method on the DocvecsArray, rather than on the Doc2Vec model itself – though that leaves open the question of how to get both word-and-doc vectors in the same file
* if I understand correctly, if strings are repeated between the word vocabulary and doctags, the resulting file will have repeating keys? A convention for avoiding those collisions, such as prefixing all doctags in a unique way (perhaps with a default that the user could override) would be helpful.
* a new `load_…` may not be necessary; there's not enough info in the format to rebuilt a trainable model. But, since the format is compatible with `KeyedVectors.load_word2vec_format()`, anything written here can be reloaded, for read-only access, using that.

parulsethi:
@gojomo made changes according to 1st and 3rd point, and agree on 4th point.
And yeah Word2vec instance is used for directly accessing vocab and syn0 to save word vectors in `Doc2vec`'s `save_word2vec_format()`. I'm doubtful for how it could be done in `DocvecsArray`

tmylk:
Could you please add an ipynb with images showing how to visualise doc2vec in Tensorboard?

parulsethi:
@tmylk Sure, I'll prepare a notebook tutorial for that

gensim/models/doc2vec.py
```diff
@@ -808,6 +809,50 @@ def delete_temporary_training_data(self, keep_doctags_vectors=True, keep_inferen
         if self.docvecs and hasattr(self.docvecs, 'doctag_syn0_lockf'):
             del self.docvecs.doctag_syn0_lockf
 
+    def save_word2vec_format(self, fname, doc_vec=True, word_vec=False, prefix='dt_', fvocab=None, binary=False):
```
gojomo: To be consistent with behavior before this change, default should be to just write word-vectors (as when implementation just inherited from Word2Vec). Let the new capability require explicit activation with a new parameter.

gensim/models/doc2vec.py
```diff
@@ -808,6 +809,50 @@ def delete_temporary_training_data(self, keep_doctags_vectors=True, keep_inferen
...
+         in binary word2vec format (default: False)
+
+        """
+        total_vec = len(self.docvecs)
+        if word_vec:
+            total_vec = total_vec + len(self.wv.vocab)
+        # save document vectors
+        if doc_vec:
+            logger.info("storing %sx%s projection weights into %s" % (total_vec, self.vector_size, fname))
+            with utils.smart_open(fname, 'wb') as fout:
```
gojomo: If all writing is done in this method, seems that should be possible, and would be cleaner, to only open the file once, rather once then a second time in append-mode. BUT, see later comment for a way that this method might be able to offload some of the writing, and thus would just choose to re-open append.

gensim/models/doc2vec.py
```diff
@@ -808,6 +809,50 @@ def delete_temporary_training_data(self, keep_doctags_vectors=True, keep_inferen
...
 
+    def save_word2vec_format(self, fname, doc_vec=True, word_vec=False, prefix='dt_', fvocab=None, binary=False):
+        """
+        Store the input-hidden weight matrix.
+
+         `fname` is the file used to save the vectors in
+         `doc_vec` is an optional boolean indicating whether to store document vectors
+         `word_vec` is an optional boolean indicating whether to store word vectors
+         (if both doc_vec and word_vec are True, then both vectors are stored in the same file)
+         `prefix` to uniquely indentify doctags from word vocab, and avoid collision
```
gojomo: Type: 'indentify'

gensim/models/doc2vec.py
```diff
@@ -808,6 +809,50 @@ def delete_temporary_training_data(self, keep_doctags_vectors=True, keep_inferen
...
+         in case of repeated string in doctag and word vocab
+         `fvocab` is an optional file used to save the vocabulary
```
gojomo: The potential to save the vocabulary, with particular index-positions that correspond to the word-vectors only, makes me think that when both word+doc vectors are stored, the word-vectors should go first. Then, at least, any vocab written aligns one-for-one with the word-vectors portion of the save file. (Also: does `fvocab` currently do anything in the save-both case?)

gensim/models/doc2vec.py
```diff
@@ -808,6 +809,50 @@ def delete_temporary_training_data(self, keep_doctags_vectors=True, keep_inferen
...
+            if word_vec:
+                with utils.smart_open(fname, 'ab') as fout:
+                    for word, vocab in sorted(iteritems(self.wv.vocab), key=lambda item: -item[1].count):
+                        row = self.wv.syn0[vocab.index]
+                        if binary:
+                            fout.write(utils.to_utf8(word) + b" " + row.tostring())
+                        else:
+                            fout.write(utils.to_utf8("%s %s\n" % (word, ' '.join("%f" % val for val in row))))
+        else:
+            KeyedVectors.save_word2vec_format(self.wv, fname, fvocab=fvocab, binary=binary)
```
gojomo: A thought, in combination with my other comments: perhaps code duplication can be reduced by, if word-vecs are enabled, just calling KeyedVectors on the word-vectors first, *then* appending doc-vecs if necessary. This would reuse the word-writing (and vocab-writing) code from KeyedVectors, but then re-open the file for append to add doc-vectors (if enabled). To make sure the front-of-file vector-count was correct, the `KeyedVectors.load_word2vec_format()` would need a new parameter telling it to boost the count by some factor the caller knows.

gensim/models/doc2vec.py
gojomo: Thinking more about names: 

* to acknowledge the doc-vectors aren't necessarily one-per-doc, but one-per-doctag, and to follow the convention elsewhere, let's enable with `doctag_vec=True`, rather than `doc_vec=True`.
* let's make the default prefix even weirder and less-at-risk of collision with any real tokens. In Mikolov's example sentence-vectors scripts, he prefixed those vector-keys with `*_`. So let's use `*dt_` as the default prefix.

gensim/models/doc2vec.py
parulsethi: Done

gensim/models/doc2vec.py
parulsethi: Done

gensim/models/doc2vec.py
parulsethi: Done, word-vectors go first now.

> (Also: does fvocab currently do anything in the save-both case?)

It didn't, earlier. But now that only `KeyedVectors.save_word2vec` is used for save-only-wv and save-both, vocab is saved in both the cases.

gensim/models/doc2vec.py
parulsethi: Done

gojomo:
Looks good!

parulsethi:
Sure, and a test too.

gensim/test/test_doc2vec.py
```diff
@@ -76,6 +76,13 @@ def test_persistence(self):
         model.save(testfile())
         self.models_equal(model, doc2vec.Doc2Vec.load(testfile()))
 
+    def testPersistenceWord2VecFormat(self):
+        """Test storing the entire model in word2vec format."""
+        model = doc2vec.Doc2Vec(DocsLeeCorpus(), min_count=1)
+        model.save_word2vec_format(testfile(), doctag_vec=True, binary=True)
+        binary_model_dv = keyedvectors.KeyedVectors.load_word2vec_format(testfile(), binary=True)
+        self.assertEqual(len(model.wv.vocab) + len(model.docvecs), len(binary_model_dv.vocab))
```
tmylk: Please add tests for more combinations of `word_vec`/`doctag_vec` `True`/`False`

PR #1267, opened 2017-04-08
syn0_lockf updated in intersect_word2vec_format()

KiddoZhu:
For the else branch with binary=False, syn0_lockf is not initialized with the parameter. So I add it.


gojomo:
+1

Thanks!

PR #1278, opened 2017-04-12
use smart_open for doc2vec-IMDB

robotcator:
drop the codecs module, use the smart_open module


tmylk:
Thanks for the improvement

PR #1295, opened 2017-04-29
[WIP] Replaced `logger.warn` by `logger.warning`

chinmayapancholi13:
This PR updates `logger.warn` to `logger.warning` in the code since `logging.warn` has been deprecated since Python 3.3 and it is advisable to use `logging.warning` ([Reference](http://stackoverflow.com/questions/15539937/whats-the-difference-between-logging-warn-and-logging-warning-in-python-or-are)).

Relevant Google mailing list discussion [here](https://groups.google.com/forum/#!topic/gensim/xztXxrKix9Q).

### DIFF
--- a/gensim/corpora/sharded_corpus.py
+++ b/gensim/corpora/sharded_corpus.py
@@ -264,10 +264,11 @@ class ShardedCorpus(IndexedCorpus):
                 logger.info('Deriving dataset dimension from corpus: '
                              '{0}'.format(proposed_dim))
             else:
-                logger.warning('Dataset dimension derived from input corpus diffe'
-                             'rs from initialization argument, using corpus.'
-                             '(corpus {0}, init arg {1})'.format(proposed_dim,
-                                                                 self.dim))
+                logger.warning(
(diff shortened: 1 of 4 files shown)


gensim/corpora/sharded_corpus.py
```diff
@@ -264,7 +264,7 @@ def init_shards(self, output_prefix, corpus, shardsize=4096, dtype=_default_dtyp
                 logger.info('Deriving dataset dimension from corpus: '
                              '{0}'.format(proposed_dim))
             else:
-                logger.warn('Dataset dimension derived from input corpus diffe'
+                logger.warning('Dataset dimension derived from input corpus diffe'
```
piskvorky: Unrelated, but could you please replace the vertical indent with hanging indent in these warnings? This is a great example of why we don't want vertical indent -- harder to maintain.

chinmayapancholi13: Sure! I'll replace vertical indent with hanging indent for all the warnings here.

tmylk:
Thanks for the improvement!

PR #1304, opened 2017-05-05
Word2Vec default constants

loisaidasam:
Helpful for reference from outside of this library
(Imagine you're training a model and you'd like to make the input parameters tunable, yet see the current defaults used in Word2Vec. With this change you can say `Word2Vec.DEFAULT_MIN_COUNT`, for example)

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -361,12 +361,37 @@ class Word2Vec(utils.SaveLoad):
     compatible with the original word2vec implementation via `wv.save_word2vec_format()` and `KeyedVectors.load_word2vec_format()`.
 
     """
+    DEFAULT_SENTENCES = None
+    DEFAULT_SIZE = 100
+    DEFAULT_ALPHA = 0.025
+    DEFAULT_WINDOW = 5
+    DEFAULT_MIN_COUNT = 5
+    DEFAULT_MAX_VOCAB_SIZE = None
+    DEFAULT_SAMPLE = 1e-3
+    DEFAULT_SEED = 1
+    DEFAULT_WORKERS = 3
+    DEFAULT_MIN_ALPHA = 0.0001
+    DEFAULT_SG = 0
+    DEFAULT_HS = 0
+    DEFAULT_NEGATIVE = 5
+    DEFAULT_CBOW_MEAN = 1
+    DEFAULT_HASHFXN = hash
+    DEFAULT_ITER = 5
+    DEFAULT_NULL_WORD = 0
+    DEFAULT_TRIM_RULE = None
+    DEFAULT_SORTED_VOCAB = 1
+    DEFAULT_BATCH_WORDS = MAX_WORDS_IN_BATCH
 
     def __init__(
-            self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
-            max_vocab_size=None, sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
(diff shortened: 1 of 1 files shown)


menshikh-iv:
You can always see default parameters in class if you run in ipython
```python
from gensim.models.word2vec import Word2Vec
Word2Vec?
````
or if you use default python interpreter
```python
from gensim.models.word2vec import Word2Vec
help(Word2Vec.__init__)
```
Before start tuning process, you need to see default parameters, but why you need to store this in variables?

loisaidasam:
@menshikh-iv your comments assume that any user of this library is using it in a REPL (ipython, jupyter notebook, etc.), a common use case for data scientists.

But what if you're building out some class-based code that doesn't run interactively, where you can't reference the defaults interactively on-the-fly, and need to reference them programatically? If you don't store default values somewhere, there's no way to do this.

menshikh-iv:
In fact, there is a way to get the standard parameters from function (it's black runtime magic)
```python
from gensim.models.doc2vec import Doc2Vec
import inspect
inspect.getargspec(Doc2Vec.__init__)
```
but your solution more simple and clear, thank you @loisaidasam

@tmylk can you merge this PR?

tmylk:
@loisaidasam 
Personally I haven't seen this way to define defaults in Python before.

Could you please elaborate more on the use case? Does `getargspec` solve it?

menshikh-iv:
@loisaidasam please answer to @tmylk, otherwise, unfortunately, I will be rejected PR

loisaidasam:
@menshikh-iv @tmylk My use case is that I'm referencing the `Word2Vec` library in my own custom framework (with custom command-line tooling) to build out some latent-vector models. In training these models, I offer the end-user the ability to tweak parameters, including suggestions for parameter defaults.

While yes, `getargspec(Word2Vec.__init__)` does get me the data that I'm looking for, it seems hacky/unnatural. The more pythonic way (in my opinion) for my use-case of external default reference is explicit declaration over implicit introspection.

Another added benefit of my proposal is that should these default parameter values change in the future, seeing the differences over time (using Github blame or some other tooling) will be even more straightforward.

menshikh-iv:
@loisaidasam 

> Another added benefit of my proposal is that should these default parameter values change in the future, seeing the differences over time (using Github blame or some other tooling) will be even more straightforward.

Changes can be seen in any case in the constructor line


>  I offer the end-user the ability to tweak parameters, including suggestions for parameter defaults.

You can copy this default params from `Word2Vec.__init__` to your tool and pass it to the constructor. I think it's good variant for you

tmylk:
Having explicit params in your tool sounds like most explicit.
closing as "won't implement"

piskvorky:
@gojomo thoughts on having such global vars with defaults?

gojomo:
I see the potential value, for some downstream re-users, of default parameters being defined prominently, explicitly, in a one-parameter-per-line manner (as opposed to nested in the function definition). But, that's a style far more common in other languages, that I've not seen used in Python. And as Radim notes, it makes reading the function signature require a little extra indirection. 

I wasn't familiar with the `inspect` option @menshikh-iv mentions... now knowing it exists, my feeling is that relying on it would be more Pythonic. In particular, the fact that it also gives you a list of the parameters, with their defaults, enables 'default-aware' external code to potentially even avoid hard-coding assumptions about what DEFAULT_ variables are in use, or to detect things like unexpected new parameters (in new library versions). So it seems both sufficient and (for some potential uses) superior.

PR #1307, opened 2017-05-07
Doc2Vec docstring updated to reflect actual default value of sample parameter. Fix #1302.

### LINKED ISSUE #1302: Documentation error in Doc2Vec
This is super simple, but the documentation says the default value of the `sample` parameter in Doc2Vec is 0:

> sample = threshold for configuring which higher-frequency words are randomly downsampled;
default is 0 (off), useful value is 1e-5.

But the actual default value is 1e-3 (0.001), based on the fact that the DocVec inherits from WordVec, where sample is set to 1e-3. See: https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec.py#L367

I can confirm this by simply initiating a model with all defaults:

```
>>> from gensim.models import Doc2Vec
>>> model = Doc2Vec()
>>> print(model.sample)
0.001
```

Not sure if the desire behavior is what's reflected in the documentation, in which case the default should be changed, or if the documentation should be updated. Simple fix either way, I imagine, but as is it causes some confusion.

datapythonista:


gensim/models/doc2vec.py
```diff
@@ -580,7 +580,7 @@ def __init__(self, documents=None, dm_mean=None,
         need about 1GB of RAM. Set to `None` for no limit (default).
 
         `sample` = threshold for configuring which higher-frequency words are randomly downsampled;
-                default is 0 (off), useful value is 1e-5.
+                default is 1e-3, useful value is 1e-5.
```
piskvorky: Thanks for the fix.

This doc reads very strange now: `default value is X, useful value is Y`. Why not use the useful value? 🤔
 @gojomo

gojomo: I think the 'useful value' remark dates back to something in word2vec.c code, before it had a (modest) non-zero default. I suppose we could stop at just mentioning the default, or hint that more aggressive values are also worth trying with a comment like: "default is 1e-3; values to 1e-5 (or lower) may also be useful". (On large datasets, I've seen ok results with 1e-06 and 1e-07, which each offer additional training speedups.)

piskvorky: Agreed. @menshikh-iv @datapythonista can you do this tiny fix?

menshikh-iv: @piskvorky Yes, of course.

PR #1313, opened 2017-05-10
Code style fixes to the TFIDF module

piskvorky:
That module is ancient, didn't adhere to the current gensim code style.

### DIFF
--- a/gensim/models/tfidfmodel.py
+++ b/gensim/models/tfidfmodel.py
@@ -28,8 +28,7 @@ def precompute_idfs(wglobal, dfs, total_docs):
     """Precompute the inverse document frequency mapping for all terms."""
     # not strictly necessary and could be computed on the fly in TfidfModel__getitem__.
     # this method is here just to speed things up a little.
-    return dict((termid, wglobal(df, total_docs))
-                for termid, df in iteritems(dfs))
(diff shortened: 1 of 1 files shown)


piskvorky:
@tmylk how far did you get, making the gensim code base style consistent?

@menshikh-iv as Lev probably explained to you, we want to keep the code style consistent. We don't want to merge any PRs that break this style, and we want to fix any existing code style issues too (such as with this PR).

menshikh-iv:
@piskvorky Where can I read about code style in gensim project? (in [developer page](https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#code-style) I see information about spaces only) ?

tmylk:
@piskvorky There is a hook at every commit that runs the `flake8` check on the touched file. That way the code style will improve over time. The `hanging indent` check is being reviewed by `pycodestyle`  project https://github.com/PyCQA/pycodestyle/pull/632 . It will become a part of gensim's automated checks when it is merged.

BTW [This](https://github.com/RaRe-Technologies/gensim/pull/1313/files#diff-896942304f9417b1b8842823b19d7431R95) is not a strictly style fix and I would rather omit it. In general having 2 attributes `id2word` and `dictionary` with identical function is confusing, but should stay to support backwards compatibility.

piskvorky:
I don't think there is a `dictionary` attribute, so there no duplicity. Or what do you mean?

PR #1321, opened 2017-05-14
Sparse support when num_best not None in interfaces. Fixes #1294

### LINKED ISSUE #1294: Support sparse input in interfaces_getitem when num_best is not None
Sparse input is ony supported when num_best=None, otherwise errors.
Raised on the mailing list https://groups.google.com/d/msg/gensim/-PhUHorj9-E/N6apxcqnHQAJ

```
from scipy.sparse import random
from gensim.similarities import SparseMatrixSimilarity
from gensim.matutils import Sparse2Corpus

X = random(200e3, 15e3, density=2e-4, format="csc")
index = SparseMatrixSimilarity(Sparse2Corpus(X, documents_columns=False), num_features=X.shape[1], num_terms=X.shape[0], maintain_sparsity=True, num_best=100)

for s in index:
    print s

```
returns
```
ValueError                                Traceback (most recent call last)
<ipython-input-2-04d8dd3277db> in <module>()
----> 1 for s in index:
      2         print s
      3 

/home/lev/miniconda2/lib/python2.7/site-packages/gensim/interfaces.pyc in __iter__(self)
    264                 chunk_end = min(self.index.shape[0], chunk_start + self.chunksize)
    265                 chunk = self.index[chunk_start : chunk_end]
--> 266                 for sim in self[chunk]:
    267                     yield sim
    268         else:

/home/lev/miniconda2/lib/python2.7/site-packages/gensim/interfaces.pyc in __getitem__(self, query)
    226         # most similar for each document in turn
    227         if matutils.ismatrix(result):
--> 228             return [matutils.full2sparse_clipped(v, self.num_best) for v in result]
    229         else:
    230             # otherwise, return top-n of the single input document

/home/lev/miniconda2/lib/python2.7/site-packages/gensim/matutils.pyc in full2sparse_clipped(vec, topn, eps)
    241     if topn <= 0:
    242         return []
--> 243     vec = np.asarray(vec, dtype=float)
    244     nnz = np.nonzero(abs(vec) > eps)[0]
    245     biggest = nnz.take(argsort(abs(vec).take(nnz), topn, reverse=True))

/home/lev/miniconda2/lib/python2.7/site-packages/numpy/core/numeric.pyc in asarray(a, dtype, order)
    529 
    530     """
--> 531     return array(a, dtype, copy=False, order=order)
    532 
    533 

ValueError: setting an array element with a sequence.
```

manneshiva:
Fixes #1294. Added a new function ( any2sparse_clipped() ), in matutils by extending any2sparse() to return the topn elements of greatest magnitude. Also, replaced full2sparse_clipped in interfaces.py to any2sparse_clipped to handle sparse cases when num_best isn't None. 
I am fairly new to contributing to open source and so do not have much idea about testing my commits. I have only tested/verified against the code to reproduce the issue and it seemed to be working. Let me know in case I need to do anything else.

### DIFF
--- a/gensim/interfaces.py
+++ b/gensim/interfaces.py
@@ -224,8 +224,7 @@ class SimilarityABC(utils.SaveLoad):
 
         # if maintain_sparity is True, result is scipy sparse. Sort, clip the
         # topn and return as a scipy sparse matrix.
-        if hasattr(self, 'maintain_sparsity'):
-            if self.maintain_sparsity:
+        if getattr(self, 'maintain_sparsity', False):
                 return matutils.scipy2scipy_clipped(result, self.num_best)
 
         # if the input query was a corpus (=more documents), compute the top-n
--- a/gensim/interfaces.py
+++ b/gensim/interfaces.py
@@ -19,8 +19,6 @@ import itertools
 from gensim import utils, matutils
 from six.moves import xrange
 
-import numpy as np
-import scipy.sparse
 
 logger = logging.getLogger('gensim.interfaces')
 
--- a/gensim/interfaces.py
+++ b/gensim/interfaces.py
@@ -225,10 +225,10 @@ class SimilarityABC(utils.SaveLoad):
         # if the input query was a corpus (=more documents), compute the top-n
         # most similar for each document in turn
         if matutils.ismatrix(result):
-            return [matutils.full2sparse_clipped(v, self.num_best) for v in result]
+            return [matutils.any2sparse_clipped(v, self.num_best) for v in result]
         else:
             # otherwise, return top-n of the single input document
-            return matutils.full2sparse_clipped(result, self.num_best)
+            return matutils.any2sparse_clipped(result, self.num_best)
 
 
     def __iter__(self):
--- a/gensim/interfaces.py
+++ b/gensim/interfaces.py
@@ -19,6 +19,8 @@ import itertools
 from gensim import utils, matutils
 from six.moves import xrange
 
+import numpy as np
+import scipy.sparse
 
 logger = logging.getLogger('gensim.interfaces')
 
@@ -222,13 +224,19 @@ class SimilarityABC(utils.SaveLoad):
         if self.num_best is None:
             return result
 
+        # if maintain_sparity is True, result is scipy sparse. Sort, clip the
+        # topn and return as a scipy sparse matrix.
+        if hasattr(self, 'maintain_sparsity'):
+            if self.maintain_sparsity:
+                return matutils.scipy2scipy_clipped(result, self.num_best)
+
         # if the input query was a corpus (=more documents), compute the top-n
         # most similar for each document in turn
         if matutils.ismatrix(result):
-            return [matutils.any2sparse_clipped(v, self.num_best) for v in result]
+            return [matutils.full2sparse_clipped(v, self.num_best) for v in result]
         else:
             # otherwise, return top-n of the single input document
-            return matutils.any2sparse_clipped(result, self.num_best)
+            return matutils.full2sparse_clipped(result, self.num_best)
 
 
     def __iter__(self):
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -185,9 +185,13 @@ def scipy2scipy_clipped(matrix, topn, eps=1e-9):
         matrix_indices = []
         matrix_data = []
         matrix_indptr = [0]
-        for v in matrix:
+        matrix.sort_indices()  # ensure data is stored in row major, sorts inplace
+        # calling abs() on entire matrix once is faster than calling abs() iteratively for each row
+        matrix_abs = abs(matrix.data)
+        for i in range(matrix.shape[0]):
+            v = matrix.getrow(i)
             # Sort and clip each row vector first.
-            biggest = argsort(abs(v.data), topn, reverse=True)
+            biggest = argsort(matrix_abs[matrix.indptr[i]:matrix.indptr[i + 1]], topn, reverse=True)
             indices, data = v.indices.take(biggest), v.data.take(biggest)
             # Store the topn indices and values of each row vector.
             matrix_data.append(data)
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -175,8 +175,14 @@ def any2sparse_clipped(vec, topn, eps=1e-9):
         return []
     if isinstance(vec, np.ndarray):
         return full2sparse_clipped(vec, topn, eps)
-    vec_sparse = any2sparse(vec, eps)
-    return sorted(vec_sparse, key=lambda x: x[1], reverse=True)[:topn]
+    if scipy.sparse.issparse(vec):
+        biggest = argsort(abs(vec).data, topn, reverse=True)
+        return list(zip(vec.indices.take(biggest), vec.data.take(biggest)))
+    else:
+        vec_csr = scipy.sparse.csr_matrix(vec)
+        biggest = argsort(abs(vec_csr).data, topn, reverse=True)
+        return list(zip(vec_csr.indices.take(biggest), vec_csr.data.take(biggest)))
+
 
 
 def scipy2sparse(vec, eps=1e-9):
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -176,7 +176,7 @@ def any2sparse_clipped(vec, topn, eps=1e-9):
     if isinstance(vec, np.ndarray):
         return full2sparse_clipped(vec, topn, eps)
     vec_sparse = any2sparse(vec, eps)
-    return sorted(vec_sparse,key=lambda x: x[1], reverse=True)
+    return sorted(vec_sparse, key=lambda x: x[1], reverse=True)
 
 
 def scipy2sparse(vec, eps=1e-9):
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -185,13 +185,13 @@ def scipy2scipy_clipped(matrix, topn, eps=1e-9):
         matrix_indices = []
         matrix_data = []
         matrix_indptr = [0]
-        matrix.sort_indices()  # ensure data is stored in row major, sorts inplace
         # calling abs() on entire matrix once is faster than calling abs() iteratively for each row
-        matrix_abs = abs(matrix.data)
+        matrix_abs = abs(matrix)
         for i in range(matrix.shape[0]):
             v = matrix.getrow(i)
+            v_abs = matrix_abs.getrow(i)
             # Sort and clip each row vector first.
-            biggest = argsort(matrix_abs[matrix.indptr[i]:matrix.indptr[i + 1]], topn, reverse=True)
+            biggest = argsort(v_abs.data, topn, reverse=True)
             indices, data = v.indices.take(biggest), v.data.take(biggest)
             # Store the topn indices and values of each row vector.
             matrix_data.append(data)
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -176,7 +176,7 @@ def any2sparse_clipped(vec, topn, eps=1e-9):
     if isinstance(vec, np.ndarray):
         return full2sparse_clipped(vec, topn, eps)
     vec_sparse = any2sparse(vec, eps)
-    return sorted(vec_sparse, key=lambda x: x[1], reverse=True)
+    return sorted(vec_sparse, key=lambda x: x[1], reverse=True)[:topn]
 
 
 def scipy2sparse(vec, eps=1e-9):
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -168,7 +168,7 @@ def any2sparse(vec, eps=1e-9):
 
 def scipy2scipy_clipped(matrix, topn, eps=1e-9):
     """
-    Return a scipy.sparse vector/matrix consisting of 'topn' elements of greatest magnitude(abs).
+    Return a scipy.sparse vector/matrix consisting of 'topn' elements of the greatest magnitude (absolute value).
     """
     if not scipy.sparse.issparse(matrix):
         raise ValueError("'%s' is not a scipy sparse vector." % matrix)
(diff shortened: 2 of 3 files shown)


manneshiva:
oops, forgot to add a whitespace after comma.

menshikh-iv:
Thank you @manneshiva, at first, please write unit-test for check that there is no exception (you can rewrite [this code from issue](https://github.com/RaRe-Technologies/gensim/issues/1294#issue-224641381) as unittest) and write test for `any2sparse_clipped` for checking output (with sparse and non-sparse inputs)

gensim/matutils.py
```diff
@@ -176,7 +176,7 @@ def any2sparse_clipped(vec, topn, eps=1e-9):
     if isinstance(vec, np.ndarray):
         return full2sparse_clipped(vec, topn, eps)
     vec_sparse = any2sparse(vec, eps)
-    return sorted(vec_sparse,key=lambda x: x[1], reverse=True)
+    return sorted(vec_sparse, key=lambda x: x[1], reverse=True)
```
menshikh-iv: Please read [this comment](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/matutils.py#L241) in `full2sparse_clipped` function

manneshiva:
@menshikh-iv Thanks for your suggestions. I will write the unit-test asap. Regarding [this comment](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/matutils.py#L241), it is handled for dense inputs. For other(non-dense) cases, we have two options:
1. Converting the input vector to sparse and then sorting these 2-tuples(implemented in current commit). (or)
2. Creating a full array from input vector and then proceeding like full2sparse_clipped().

I chose method (1), because I feel that in case of sparse vectors, length of the sparse 2-tuples << num_documents and thus sorting these tuples directly will take lesser time(which isn't the case with full inputs and therefore handled separately).  Does this make sense? Would you rather recommend method (2) ?

gensim/matutils.py
```diff
@@ -166,6 +166,19 @@ def any2sparse(vec, eps=1e-9):
...
+    """
+    Like `any2sparse`, but only returns the `topn` elements of greatest magnitude (abs).
+
+    """
+    if topn <= 0:
+        return []
+    if isinstance(vec, np.ndarray):
+        return full2sparse_clipped(vec, topn, eps)
+    vec_sparse = any2sparse(vec, eps)
+    return sorted(vec_sparse, key=lambda x: x[1], reverse=True)
```
tmylk: Please only return `topn`

tmylk:
Could you please benchmark the performance of 1 and 2 for, say, 1% of non-zero elements in a 10k long array?

manneshiva:
@menshikh-iv @tmylk In the process of re-writing the code from [the issue](https://groups.google.com/forum/#!msg/gensim/-PhUHorj9-E/N6apxcqnHQAJ) for the unit test, I realized a fault in **my**/our understanding of the problem. 

```
def testMaintainSparsity(self):
        """Sparsity is correctly maintained when maintain_sparsity=True"""
        num_features = len(dictionary)

        index = self.cls(corpus, num_features=num_features)
        dense_sims = index[corpus]

        index = self.cls(corpus, num_features=num_features, maintain_sparsity=True)
        sparse_sims = index[corpus]

        self.assertFalse(scipy.sparse.issparse(dense_sims))
        self.assertTrue(scipy.sparse.issparse(sparse_sims))
        numpy.testing.assert_array_equal(dense_sims, sparse_sims.todense())
```
As seen from this unit test, 'maintain_sparsity' indicates the type of the output(whether scipy sparse or not) and has nothing to do with the type of the input. In the case when num_best is not None, the output is expected to be a list of reverse sorted tuples(index,similarity). So basically, the problem is not the sparse input but rather asking for a sparse output when num_best=some_n. It does not make sense to return a scipy sparse matrix in the case when num_best is not None and maintain_sparsity is True, which is exactly what was assumed in the present code. The fix for this problem is to simply raise an error when trying to set maintain_sparsity to True and num_best=n. Should have caught this before. Or am I missing something?

tmylk:
The goal of this issue was to support `num_best is not None` and `maintain_sparsity=True` case. You see that there is a problem with this. Could you please explain better what that problem is?

manneshiva:
Suppose we were to support `num_best is not None` and `maintain_sparsity=True` , what should be the output? To maintain sparity, is it sufficient to return a list of scipy.sparce.csr rows(i.e. list of 1xn sparse rows). For clarity, I am posting the output of the code that I have written:
`Output of present gensim code for num_best=3 and maintain_sparsity=False
`
`[[(0, 0.99999994039535522), (2, 0.28867512941360474), (1, 0.2357022613286972)], [(1, 1.0), (4, 0.70710676908493042), (2, 0.40824830532073975)], [(2, 1.0), (3, 0.61237245798110962), (1, 0.40824830532073975)], [(3, 1.0), (2, 0.61237245798110962), (1, 0.3333333432674408)], [(4, 0.99999994039535522), (1, 0.70710676908493042), (2, 0.28867512941360474)], [(5, 1.0), (6, 0.70710676908493042), (7, 0.57735025882720947)], [(6, 0.99999994039535522), (7, 0.81649655103683472), (5, 0.70710676908493042)], [(7, 0.99999994039535522), (6, 0.81649655103683472), (8, 0.66666662693023682)], [(8, 0.99999994039535522), (7, 0.66666662693023682), (6, 0.40824827551841736)]]`
`----------------------------------------------------------------------------------`
`Output of my current code for num_best=3 and maintain_sparsity=True`
```
[<1x3 sparse matrix of type '<type 'numpy.float32'>'
with 3 stored elements in Compressed Sparse Row format>, <1x5 sparse matrix of type '<type 'numpy.float32'>'
with 3 stored elements in Compressed Sparse Row format>, <1x4 sparse matrix of type '<type 'numpy.float32'>'
with 3 stored elements in Compressed Sparse Row format>, <1x4 sparse matrix of type '<type 'numpy.float32'>'
with 3 stored elements in Compressed Sparse Row format>, <1x5 sparse matrix of type '<type 'numpy.float32'>'
with 3 stored elements in Compressed Sparse Row format>, <1x8 sparse matrix of type '<type 'numpy.float32'>'
with 3 stored elements in Compressed Sparse Row format>, <1x8 sparse matrix of type '<type 'numpy.float32'>'
with 3 stored elements in Compressed Sparse Row format>, <1x9 sparse matrix of type '<type 'numpy.float32'>'
with 3 stored elements in Compressed Sparse Row format>, <1x9 sparse matrix of type '<type 'numpy.float32'>'
with 3 stored elements in Compressed Sparse Row format>]
```
`----------------------------------------------------------------------------------`
`Content of this list of sparse rows:`
`[[(0, 0.99999994), (2, 0.28867513), (1, 0.23570226)], [(1, 1.0), (4, 0.70710677), (2, 0.40824831)], [(2, 1.0), (3, 0.61237246), (1, 0.40824831)], [(3, 1.0), (2, 0.61237246), (1, 0.33333334)], [(4, 0.99999994), (1, 0.70710677), (2, 0.28867513)], [(5, 1.0), (6, 0.70710677), (7, 0.57735026)], [(6, 0.99999994), (7, 0.81649655), (5, 0.70710677)], [(7, 0.99999994), (6, 0.81649655), (8, 0.66666663)], [(8, 0.99999994), (7, 0.66666663), (6, 0.40824828)]]`

The sparse output is the same as the dense output as seen above. But it is not possible to convert this list of sparse rows to a sparse matrix since the shape of each row is different(depends on the highest index in topn indices). Is it ok to return a list of sparse rows?

manneshiva:
@tmylk Found a way to return a sparse csr_matrix (possible even with variable shape of each row) instead of list of sparse rows as mentioned. But it loses the ordering of the values in each row. `zip(row[i].indices, row[i].data)` gives a different order rather than a decreasing similarity value order. Once you confirm on the required output, I shall immediately commit my changes. Thanks!

manneshiva:
@tmylk Tried and tested a way to create a sparse scipy matrix which preserves the order. So the current output will be a scipy  sparse order preserving matrix. Will code it and push the changes soon. Thanks for your patience.

manneshiva:
@tmylk I have fixed the issue with the following changes:
1. Replaced any2sparse_clipped with scipy2scipy_clipped(function for clipping topn from a scipy matrix).
2. Added new code path for the case when `maintain_sparse = True` and `num_best is not None`. 
3. Added unit tests for (1) and (2).

The output type (for `maintain_sparse = True` and `num_best is not None`) is now **consistent** with the output when `maintain_sparse = True` and `num_best is None`. I have also tested against the code in the [issue](https://github.com/RaRe-Technologies/gensim/issues/1294) and compared the output of sparse and non sparse when `num_best is not None`. The issue looks resolved. Please review and let me know if any changes are needed. Thanks.

manneshiva:
@tmylk Let me know if any changes are required.

gensim/test/test_similarities.py
```diff
@@ -405,6 +421,21 @@ def testMaintainSparsity(self):
         self.assertTrue(scipy.sparse.issparse(sparse_sims))
         numpy.testing.assert_array_equal(dense_sims, sparse_sims.todense())
 
+    def testMaintainSparsityWithNumBest(self):
+        """Tests that sparsity is correctly maintained when maintain_sparsity=True and num_best is not None"""
+        num_features = len(dictionary)
+
+        index = self.cls(corpus, num_features=num_features, num_best=3)
```
tmylk: Explicit `maintain_sparsity=False` will add clarity.

tmylk:
@manneshiva I am concerned about the call to `matrix.data` as it creates a numpy array in memory and thus the benefits of having sparse format disappear.

Please preserve sparsity when selecting num_best

gensim/interfaces.py
```diff
@@ -222,6 +222,12 @@ def __getitem__(self, query):
         if self.num_best is None:
             return result
 
+        # if maintain_sparity is True, result is scipy sparse. Sort, clip the
+        # topn and return as a scipy sparse matrix.
+        if hasattr(self, 'maintain_sparsity'):
```
piskvorky: `getattr` accepts a default value.

gensim/matutils.py
```diff
@@ -166,6 +166,39 @@ def any2sparse(vec, eps=1e-9):
     return [(int(fid), float(fw)) for fid, fw in vec if np.abs(fw) > eps]
 
 
+def scipy2scipy_clipped(matrix, topn, eps=1e-9):
+    """
+    Return a scipy.sparse vector/matrix consisting of 'topn' elements of greatest magnitude(abs).
```
piskvorky: `of the greatest magnitude (absolute value)`.

gensim/matutils.py
```diff
@@ -166,6 +166,39 @@ def any2sparse(vec, eps=1e-9):
...
+    """
+    if not scipy.sparse.issparse(matrix):
+        raise ValueError("'%s' is not a scipy sparse vector." % matrix)
+    if topn <= 0:
+        return scipy.sparse.csr_matrix([])
+    # Return clipped sparse vector if input is a sparse vector.
+    if matrix.shape[0] == 1:
+        # use np.argpartition/argsort and only form tuples that are actually returned.
+        biggest = argsort(abs(matrix).data, topn, reverse=True)
```
piskvorky: `abs(matrix.data)` should be much faster than `abs(matrix).data` (numpy vs scipy).

gensim/matutils.py
```diff
@@ -166,6 +166,39 @@ def any2sparse(vec, eps=1e-9):
...
+        indices, data = matrix.indices.take(biggest), matrix.data.take(biggest)
+        return scipy.sparse.csr_matrix((data, indices, [0, len(indices)]))
+    # Return clipped sparse matrix if input is a matrix, processing row by row.
+    else:
+        matrix_indices = []
+        matrix_data = []
+        matrix_indptr = [0]
+        for v in matrix:
+            # Sort and clip each row vector first.
+            biggest = argsort(abs(v).data, topn, reverse=True)
```
piskvorky: Dtto.

Even faster to call `abs` just once, on the entire matrix, rather than once per row.

manneshiva:
@tmylk cc @piskvorky  Will push the required changes soon. 

> @manneshiva I am concerned about the call to matrix.data as it creates a numpy array in memory and thus the benefits of having sparse format disappear.
Please preserve sparsity when selecting num_best

I am not sure about what you mean. matrix.data returns all nonzero values and not the full length(#docs) array thus still preserves sparsity. Am I missing something?

menshikh-iv:
Please try run code from issue and fix this problem @manneshiva 

```
IndexError                                Traceback (most recent call last)
<ipython-input-57-3e2df5bae673> in <module>()
      6 index = SparseMatrixSimilarity(Sparse2Corpus(X, documents_columns=False), num_features=X.shape[1], num_terms=X.shape[0], maintain_sparsity=True, num_best=100)
      7 
----> 8 for s in index:
      9     print s, s.data

/home/ivan/.virtualenvs/cl/local/lib/python2.7/site-packages/gensim/interfaces.pyc in __iter__(self)
    269                 chunk_end = min(self.index.shape[0], chunk_start + self.chunksize)
    270                 chunk = self.index[chunk_start : chunk_end]
--> 271                 for sim in self[chunk]:
    272                     yield sim
    273         else:

/home/ivan/.virtualenvs/cl/local/lib/python2.7/site-packages/gensim/interfaces.pyc in __getitem__(self, query)
    226         # topn and return as a scipy sparse matrix.
    227         if getattr(self, 'maintain_sparsity', False):
--> 228                 return matutils.scipy2scipy_clipped(result, self.num_best)
    229 
    230         # if the input query was a corpus (=more documents), compute the top-n

/home/ivan/.virtualenvs/cl/local/lib/python2.7/site-packages/gensim/matutils.pyc in scipy2scipy_clipped(matrix, topn, eps)
    193             # Sort and clip each row vector first.
    194             biggest = argsort(matrix_abs[matrix.indptr[i]:matrix.indptr[i + 1]], topn, reverse=True)
--> 195             indices, data = v.indices.take(biggest), v.data.take(biggest)
    196             # Store the topn indices and values of each row vector.
    197             matrix_data.append(data)

IndexError: cannot do a non-empty take from an empty axes.
```

manneshiva:
@menshikh-iv Made required changes and tested on code from issue (`maintain_sparsity = True `, `num_best=100`). Also checked correctness by comparing output with `maintain_sparsity = False `, `num_best=100`. The issue looks resolved. Please confirm and let me know.
`

menshikh-iv:
Thank you @manneshiva

PR #1327, opened 2017-05-16
[WIP] Fix backward incompatibility due to `random_state`

chinmayapancholi13:
This PR fixes issue https://github.com/RaRe-Technologies/gensim/issues/1082.


tmylk:
Please add unit tests

chinmayapancholi13:
@tmylk Yes. Working on adding tests.

chinmayapancholi13:
@tmylk I have added unit tests for ensuring that backward compatibility is not broken due to `id2word` and `random_state` in the first 2 commits in this PR.

In the following 2 commits, I also had to add checks `isinstance(self.alpha, np.ndarray)` and `isinstance(self.eta, np.ndarray)` otherwise this was leading to an error when `self.alpha` or `self.eta` were of type `float` (since we are checking `len(self.alpha.shape)` and `len(self.eta.shape)` here). Particularly, this was happening when I was trying to again save the pre-0.13.2 model after loading.

And the last commit is failing for Python 3.5 because of the test  `testPasses` in `gensim.test.test_ldamodel.TestLdaMulticore` due to an assertion-error while comparing two floating point numbers. I haven't updated any code related to this in the PR. Should I do something about this error?

gensim/models/ldamodel.py
```diff
@@ -1004,6 +1004,11 @@ def save(self, fname, ignore=['state', 'dispatcher'], separately=None, *args, **
         """
         if self.state is not None:
             self.state.save(utils.smart_extension(fname, '.state'), *args, **kwargs)
+
+        # Save 'random_state' separately
```
tmylk: what is the purpose of saving it separately?

gensim/models/ldamodel.py
```diff
@@ -1054,13 +1059,22 @@ def load(cls, fname, *args, **kwargs):
...
-                result.id2word = utils.unpickle(id2word_fname)
-            except Exception as e:
-                logging.warning("failed to load id2word dictionary from %s: %s", id2word_fname, e)
+
+        random_state_fname = utils.smart_extension(fname, '.random_state')
+        if (os.path.isfile(random_state_fname)):
+            result.random_state = utils.unpickle(random_state_fname)
         else:
-            result.id2word = None
+            logging.warning("random_state not stored on disk so using default value")
```
tmylk: if random_state is not stored that means that it is a new version of the model and it is going to be loaded in the main pickle load. Please change the logic

gensim/test/test_ldamodel.py
```diff
@@ -470,6 +470,29 @@ def testLargeMmapCompressed(self):
...
+        model_pre_0_13_2 = self.class_.load(pre_0_13_2_fname)
+
+        model_topics = model_pre_0_13_2.print_topics(num_topics=3, num_words=3)
+
+        for i in model_topics:
+            self.assertTrue(isinstance(i[0], int))
+            self.assertTrue(isinstance(i[1], six.string_types))
+
+    def testRandomStateBackwardCompatibility(self):
+        # load a model saved using a pre-0.13.2 version of Gensim
```
tmylk: this test is identical to the previous one. just one test is enough that checks all the fields

tmylk:
could you please test that this exact test fails in the version of the code prior to your changes?
Please create a commit with only the test/test_data and zero changes.

gensim/test/test_ldamodel.py
chinmayapancholi13: Sure. I thought since these were two different issues so we'd want to put separate tests to verify both are resolved. I'll make the update according to your suggestion and remove the earlier test here.

chinmayapancholi13:
@tmylk The last commit now shows that all checks have passed. I guess it's sorted for now then.

gensim/models/ldamodel.py
chinmayapancholi13: @tmylk Your suggestion is a bit different from the [earlier discussion](https://github.com/RaRe-Technologies/gensim/issues/1082#issuecomment-301164932) on the issue. Hence, I want to make sure I understand the desired solution before making the changes.

If I understand it correctly, what you are suggesting is :  
- If there is indeed a file with the extension `.random_state` present on disk, this means that the model was saved using a pre-0.13.2 version of Gensim. So we use this file to set `result.random_state` at the time of loading. 
- However, if there is no such file present on disk, then this means that the model was saved using a post-0.13.2 version of Gensim and thus `result.random_state` got set at the time of the main pickle load. So in this case, we don't need to do anything else.

But for models saved using a pre-0.13.2 version of Gensim, there was no `.random_state` file created at the time of saving the model. So while loading such a model from disk, where would the `.random_state` file come from in this case? Is the user responsible for creating this file explicitly in such a case? If this is true, then I believe we don't need to make any changes in the `save` function for LdaModel at all and we just need to change the `load` function. 

Please correct me if I am wrong or missing something here. Otherwise, if this is indeed what we need, I could go ahead and make the appropriate changes.

tmylk: Please add a check that this indeed happened: `thus result.random_state got set at the time of the main pickle load. `

gensim/test/test_ldamodel.py
tmylk: Better to avoid code duplication

tmylk:
To validate this as a fix we need to see what changed. This is how Test Driven Development works
1) write new tests. See them fail. Please create a commit with only the test/test_data and zero code changes.

2) Write new code. see tests pass.

Could you please add these 2 commits to this PR?

gensim/models/ldamodel.py
menshikh-iv: I think It's better for case "file don't exist" (and more flexible)

chinmayapancholi13:
@tmylk Got it. Thanks a lot for the comprehensive feedback. :) 
I am adding these 2 commits to the PR.

gensim/models/ldamodel.py
chinmayapancholi13: @tmylk I believe I have understood the solution that we want. However, I have a minor doubt about where would the `.random_state` file come from? Is it that the user is responsible for creating it explicitly always and we (i.e. from within the `save` function) need not create it ever?
If this is true, then in case we are loading a pre-0.13.2 model and no `.random_state` file exists on disk, then should we set `result.random_state` using a default value like `get_random_state(None)` with a `logger.warning`?

chinmayapancholi13: @tmylk Could you please respond to this query?

menshikh-iv: >  Is it that the user is responsible for creating it explicitly always

I think the user should not think about additional files, he saves the whole model

> If this is true, then in case we are loading a pre-0.13.2 model and no .random_state file exists on disk, then should we set result.random_state using a default value like get_random_state(None) with a logger.warning?

I think it's correct

menshikh-iv: The simple way to do this is check that random_state was loaded, if not - you `set result.random_state using a default value like get_random_state(None) with a logger.warning`

chinmayapancholi13: @menshikh-iv Then there is no `.random_state` file involved at all, correct? To summarize, the solution is : 
- First, load the entire model.
- Check if `result.random_state` was set or not. For the newer (post 0.13.2) models, it would have been set through the main pickle load. For the older (pre 0.13.2) models, `result.random_state` would not be set through the main pickle load so we set `result.random_state` to `get_random_state(None)`.

And in this solution, we don't need to make any changes in the `save` function, just the `load` function.

chinmayapancholi13:
@tmylk @menshikh-iv I have made changes to the code as well. One of the tests is failing for Python 2.7 in the last commit because of `Cannot allocate memory` error.

PR #1332, opened 2017-05-17
Fix issue-1310

parulsethi:
Renamed `out_path`. Also, shifted a log line just above its corresponding command execution.

### DIFF
--- a/gensim/models/wrappers/wordrank.py
+++ b/gensim/models/wrappers/wordrank.py
@@ -98,11 +98,14 @@ class Wordrank(KeyedVectors):
         input_fnames = [corpus_file.split('/')[-1], corpus_file.split('/')[-1], cooccurrence_file]
         output_fnames = [temp_vocab_file, cooccurrence_file, cooccurrence_shuf_file]
 
-        logger.info("Prepare training data using glove code '%s'", commands)
+        logger.info("Prepare training data using glove code")
         for command, input_fname, output_fname in zip(commands, input_fnames, output_fnames):
+            logger.info("Prepare '%s' using '%s'", output_fname, command)
             with smart_open(input_fname, 'rb') as r:
                 with smart_open(output_fname, 'wb') as w:
                     utils.check_output(w, args=command, stdin=r)
+
+        logger.info("Delete frequencies from vocab file")
         with smart_open(vocab_file, 'wb') as w:
             utils.check_output(w, args=cmd_del_vocab_freq)
 
--- a/gensim/models/wrappers/wordrank.py
+++ b/gensim/models/wrappers/wordrank.py
@@ -8,7 +8,7 @@ The wrapped model can NOT be updated with new documents for online training -- u
 `Word2Vec` for that.
 
 Example:
->>> model = gensim.models.wrappers.Wordrank('/Users/dummy/wordrank', corpus_file='text8', out_path='wr_model')
+>>> model = gensim.models.wrappers.Wordrank('/Users/dummy/wordrank', corpus_file='text8', out_name='wr_model')
 >>> print model[word]  # prints vector for given words
 
 .. [1] https://bitbucket.org/shihaoji/wordrank/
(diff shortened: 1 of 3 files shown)


tmylk:
please add logging before every check_output call

parulsethi:
Done

gensim/models/wrappers/wordrank.py
```diff
@@ -95,14 +95,17 @@ def train(cls, wr_path, corpus_file, out_path, size=100, window=15, symmetric=1,
...
         output_fnames = [temp_vocab_file, cooccurrence_file, cooccurrence_shuf_file]
 
+        logger.info("Prepare training data using glove code")
         for command, input_fname, output_fname in zip(commands, input_fnames, output_fnames):
+            logger.info("Prepare '%s' using '%s'", output_fname, command)
             with smart_open(input_fname, 'rb') as r:
                 with smart_open(output_fname, 'wb') as w:
                     utils.check_output(w, args=command, stdin=r)
+
+        logger.info("Delete frequencies from vocab file")
```
menshikh-iv: Please add a logging with `cmd_del_vocab_freq` like line 103

menshikh-iv: It may be necessary to add logging arguments to the function `check_output`, @tmylk what do you think about it?

tmylk: Yes, let's add a utils logger

menshikh-iv:
@parulsethi could you please add logging parameters to `utils.check_output` function

gensim/utils.py
```diff
@@ -1164,11 +1164,12 @@ def check_output(stdout=subprocess.PIPE, *popenargs, **kwargs):
     Added extra KeyboardInterrupt handling
     """
     try:
+        cmd = kwargs.get("args")
+        logger.info("COMMAND: %s", cmd)
```
parulsethi: I think if some specific information needs to be logged for a command, then the log statement can be specified in the code there and this will just make sure to log the command that has to be run for debugging purposes. WDYT?

Or should a parameter for a log string be provided so that the command specific info is also logged from check_output

menshikh-iv: 1. You should use `logger.debug` here (because it's needed for debug purposes only)
2. Need to logging all input parameters (`popenargs` and `kwargs`)
3. I think now is not necessary to logging some "command specific" data (based on troubles from our users in mailing list), logging all arguments will be enough

parulsethi: Made suggested changes, and removed logging command(`cmd`) at `logger.info` level

menshikh-iv:
Looks good, thank you @parulsethi  :+1:

gensim/models/wrappers/wordrank.py
tmylk: would be great

gensim/utils.py
```diff
@@ -1164,6 +1164,7 @@ def check_output(stdout=subprocess.PIPE, *popenargs, **kwargs):
     Added extra KeyboardInterrupt handling
     """
     try:
+        logger.debug("COMMAND: %s %s", str(popenargs), str(kwargs))
```
piskvorky: `%s` = `str`, no need to call that function explicitly.

gensim/models/wrappers/wordrank.py
```diff
@@ -45,14 +45,14 @@ class Wordrank(KeyedVectors):
...
-    def train(cls, wr_path, corpus_file, out_path, size=100, window=15, symmetric=1, min_count=5, max_vocab_size=0,
+    def train(cls, wr_path, corpus_file, out_name, size=100, window=15, symmetric=1, min_count=5, max_vocab_size=0,
               sgd_num=100, lrate=0.001, period=10, iter=90, epsilon=0.75, dump_period=10, reg=0, alpha=100,
               beta=99, loss='hinge', memory=4.0, cleanup_files=True, sorted_vocab=1, ensemble=0):
         """
         `wr_path` is the path to the Wordrank directory.
         `corpus_file` is the filename of the text file to be used for training the Wordrank model.
         Expects file to contain space-separated tokens in a single line
-        `out_path` is the path to directory which will be created to save embeddings and training data.
+        `out_name` is name of the directory which will be created(in wordrank folder) to save embeddings and training data.
```
piskvorky: Space before bracket `(`.

gensim/models/wrappers/wordrank.py
```diff
@@ -95,14 +95,16 @@ def train(cls, wr_path, corpus_file, out_path, size=100, window=15, symmetric=1,
         cmd_del_vocab_freq = ['cut', '-d', " ", '-f', '1', temp_vocab_file]
 
         commands = [cmd_vocab_count, cmd_cooccurence_count, cmd_shuffle_cooccurences]
-        logger.info("Prepare training data using glove code '%s'", commands)
         input_fnames = [corpus_file.split('/')[-1], corpus_file.split('/')[-1], cooccurrence_file]
         output_fnames = [temp_vocab_file, cooccurrence_file, cooccurrence_shuf_file]
 
+        logger.info("Prepare training data using glove code")
```
piskvorky: We try to make `info` messages more informative: what data? How much data?

If it's not meant to be seen by end users (internal opaque debug message), it's better to use `debug`.

gensim/models/wrappers/wordrank.py
```diff
@@ -95,14 +95,16 @@ def train(cls, wr_path, corpus_file, out_path, size=100, window=15, symmetric=1,
...
         for command, input_fname, output_fname in zip(commands, input_fnames, output_fnames):
             with smart_open(input_fname, 'rb') as r:
                 with smart_open(output_fname, 'wb') as w:
                     utils.check_output(w, args=command, stdin=r)
+
+        logger.info("Delete frequencies from vocab file")
```
piskvorky: `Delete` = imperative. Is that intented? Should the user delete something?

Probably meant to be "deleting" or "will delete" instead.

PR #1334, opened 2017-05-18
Lda difference

menshikh-iv:
Only first case from [PR 1243](https://github.com/RaRe-Technologies/gensim/pull/1243) integrated into LdaModel as 'diff' method with basic tests

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -532,6 +532,10 @@ def jaccard(vec1, vec2):
         return 1 - float(len(intersection)) / float(len(union))
 
 
+def jaccard_set(set1, set2):
+    return 1. - float(len(set1 & set2)) / float(len(set1 | set2))
+
+
 def dirichlet_expectation(alpha):
     """
     For a vector `theta~Dir(alpha)`, compute `E[log(theta)]`.
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -33,11 +33,13 @@ The algorithm:
 import logging
 import numpy as np
 import numbers
+from random import sample
 import os
 
 from gensim import interfaces, utils, matutils
 from gensim.matutils import dirichlet_expectation
 from gensim.models import basemodel
+from gensim.matutils import kullback_leibler, hellinger, jaccard_set
 
 from itertools import chain
 from scipy.special import gammaln, psi  # gamma function utils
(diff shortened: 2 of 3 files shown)


gensim/models/ldamodel.py
```diff
@@ -965,6 +967,73 @@ def get_term_topics(self, word_id, minimum_probability=None):
...
+
+        z = np.zeros((t1_size, t2_size))
+
+        for topic1 in range(t1_size):
+            for topic2 in range(t2_size):
+
+                z[topic1][topic2] = distance_func(d1[topic1], d2[topic2])
+
+        if np.abs(np.max(z)) > 1e-8:
+            z /= np.max(z)
```
tmylk: Please comment on why normalisation is important

menshikh-iv:
Need to add a notebook with simple tutorial with usage & it's PR will be complete

gensim/matutils.py
```diff
@@ -532,6 +532,10 @@ def jaccard(vec1, vec2):
         return 1 - float(len(intersection)) / float(len(union))
 
 
+def jaccard_set(set1, set2):
+    return 1. - float(len(set1 & set2)) / float(len(set1 | set2))
```
piskvorky: Will throw an exception if both inputs empty -- is that desired?

Missing docstring.

gensim/models/ldamodel.py
```diff
@@ -965,6 +967,72 @@ def get_term_topics(self, word_id, minimum_probability=None):
...
+            `diff_l` is word from symmetric difference of `topic_i` and `topic_j`
+        `normed` is a flag. If `true`, matrix Z will be normalized
+        Example:
+        >>> m1, m2 = LdaMulticore.load(path_1), LdaMulticore.load(path_2)
+        >>> mdiff, annotation = m1.diff(m2)
+        >>> print(mdiff) # get matrix with difference for each topic pair from `m1` and `m2`
+        >>> print(annotation) # get array with positive/negative words for each topic pair from `m1` and `m2`
+        """
+
+        distances = {"kulback_leibler": kullback_leibler,
```
piskvorky: Hanging indent. @tmylk

gensim/models/ldamodel.py
```diff
@@ -965,6 +967,72 @@ def get_term_topics(self, word_id, minimum_probability=None):
...
+        z = np.zeros((t1_size, t2_size))
+        for topic1 in range(t1_size):
+            for topic2 in range(t2_size):
+                z[topic1][topic2] = distance_func(d1[topic1], d2[topic2])
+
+        if normed:
+            if np.abs(np.max(z)) > 1e-8:
+                z /= np.max(z)
+
+        annotation = [[None for _ in range(t1_size)] for _ in range(t2_size)]
```
piskvorky: You can create lists using `*`: `[None] * t1_size`.

Although I don't see the point of this initialization. Why not just start empty and append, in the loop below? What's with the `None`s?

gensim/test/test_tmdiff.py
```diff
@@ -0,0 +1,54 @@
...
+import unittest
+import numpy as np
+
+from gensim.corpora import Dictionary
+from gensim.models import LdaModel
+
+
+class TestLdaDiff(unittest.TestCase):
+    def setUp(self):
+        texts = [['human', 'interface', 'computer'],
```
piskvorky: Hanging indent.

gensim/models/ldamodel.py
menshikh-iv: > Why not just start empty and append, in the loop below?

Initialization allows writing more readable code (only assignment to cell).

piskvorky: I see. If that's your worry, isn't creating the 2D matrix as a numpy matrix (2D array) simpler/more readable?

menshikh-iv: Numpy matrix with complex object type of element `[str, str]` is not the best choice

PR #1336, opened 2017-05-18
add test_req

menshikh-iv:
Add `tests_require` to setup.py
The reason is `python setup.py test` don't install test dependencies if `test_require` not presented

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -226,6 +226,12 @@ Copyright (c) 2009-now Radim Rehurek
 """
 
 
+test_env = ['testfixtures',
+            'unittest2',
+            'Morfessor==2.0.2a4',
+            'scikit-learn',
+            'pyemd']
+
 setup(
     name='gensim',
     version='2.1.0',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -282,14 +282,22 @@ def finalize_options(self):
...
+        'testfixtures',
+        'unittest2',
+        'Morfessor==2.0.2a4',
+        'scikit-learn',
+        'pyemd'
+    ],
     extras_require={
         'distributed': ['Pyro4 >= 4.27'],
         'wmd': ['pyemd >= 0.2.0'],
         'test': [
```
tmylk: Please remove `[test]` option

menshikh-iv: It's better that I remove the duplication of the list of libraries (use common variable) because the option `[test]` is very convenient for some cases.

setup.py
```diff
@@ -226,6 +226,12 @@ def finalize_options(self):
 """
 
 
+test_env = ['testfixtures',
```
piskvorky: Hanging indent please, not vertical (everywhere).

Issue #1236, opened 2017-03-23
Gensim error when loading French FastText


jplu:
Hello,

I'm trying to use the fasttext wrapper in order to load the French model that one can find [here](https://s3-us-west-1.amazonaws.com/fasttext-vectors/wiki.fr.zip). Unfortunately I get the following error:

```
Traceback (most recent call last):
  File "app.py", line 18, in <module>
    model = FastText.load_fasttext_format(os.path.abspath('./wiki.fr'))
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gensim/models/wrappers/fasttext.py", line 238, in load_fasttext_format
    model.load_binary_data('%s.bin' % model_file)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gensim/models/wrappers/fasttext.py", line 255, in load_binary_data
    self.load_dict(f)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gensim/models/wrappers/fasttext.py", line 277, in load_dict
    assert len(self.wv.vocab) == vocab_size, 'mismatch between vocab sizes'
AssertionError: mismatch between vocab sizes
```
I'm using the following environment:

```
>>> import platform; print(platform.platform())
Darwin-16.4.0-x86_64-i386-64bit
>>> import sys; print("Python", sys.version)
('Python', '2.7.13 (default, Dec 28 2016, 14:29:07) \n[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)]')
>>> import numpy; print("NumPy", numpy.__version__)
('NumPy', '1.12.0')
>>> import scipy; print("SciPy", scipy.__version__)
('SciPy', '0.19.0')
>>> import gensim; print("gensim", gensim.__version__)
('gensim', '1.0.1')
>>> from gensim.models import word2vec;print("FAST_VERSION", word2vec.FAST_VERSION)
('FAST_VERSION', 0)
```

Steps to reproduce the error:

```
wget https://s3-us-west-1.amazonaws.com/fasttext-vectors/wiki.fr.zip
unzip wiki.fr.zip
python -c "import os;from gensim.models.wrappers import FastText;FastText.load_fasttext_format(os.path.abspath('./wiki.fr'))"
```
I don't know if it is a bug from gensim or an issue from the used model. Any help would be appreciated.

[...]

Issue #1261, opened 2017-04-04
Improve FastText loading times


tmylk:
Consider reading just the `bin` file as sugested in https://github.com/RaRe-Technologies/gensim/issues/814#issuecomment-289464725

Compare to C++ code in https://github.com/salestock/fastText.py/blob/77bdf69ee97a7446e314b342b129c5d46e9e4e29/fasttext/fasttext.pyx#L143

prakhar2b:
@tmylk  I would like to work on this

tmylk:
@prakhar2b that would be great

jayantj:
My thoughts about loading the FastText model using only the `.bin` file - 
1. Definitely doable, the `.bin` file contains all information about the model
2. The `.vec` file was used mainly to be able to reuse code from `KeyedVectors`, changing to use simply the `.bin` file would require some architectural changes.
3. The next step should be clearly benchmarking the FastText loading times for gensim and fastText.py for small and large models, and then profiling the gensim method. In case loading the `.vec` file is a significant bottleneck, it would make sense to load using the `.bin` file only.

manneshiva:
I was looking to use fasttext in python and the gensim wrapper seems to be considerably slower than salestock/fasttext.py . The gensim wrapper takes around 35seconds to load a moderately large(~200K vocab) pre trained model (.bin-1Gb, .vec-0.2gb) while the fasttext.py takes just above a second to load the same model. Posting the cProfile stats for both the runs/loads below: 
`Gensim Wrapper`
![image](https://cloud.githubusercontent.com/assets/26998249/26436964/e4dd435c-4137-11e7-88f4-2107ef7680f6.png)
`Fasttext.py`
![image](https://cloud.githubusercontent.com/assets/26998249/26436985/09171f2c-4138-11e7-8294-9074af98e794.png)
As seen from the first figure, loading the .bin file alone(by reading in python) would not bring the load time anywhere close to the salestock/fasttext.py load time. Achieving comparable load times would require reading in C/C++(like salestock).

jayantj:
Hi @manneshiva , thanks for the profiling report.

It could be possible to reduce the time required for loading the .bin file further by optimizing the bottlenecks (we could rewrite them in Cython if required), but we should definitel get rid of the call to `load_word2vec_format`.

Does the profiling report also include what exactly is slow inside the `load_vectors` function? That would be helpful in determining what the next steps should be.

manneshiva:
@jayantj 
[...]

PR #1200, opened 2017-03-09
gensim models show_topic/print_topic parameter num_words changed to topn to match other topic models

prakhar2b:
show_topic parameter num_words changed to topn in order to make it consistent with LdaModel. Fix [#1198](https://github.com/RaRe-Technologies/gensim/issues/1198)

PR #1341, opened 2017-05-22
Loading fastText models using only bin file

prakhar2b:
solve #1236. Mismatch in `bin` and `vec` files while loading wiki.fr pretrained model is  resolved in this PR


prakhar2b:
@tmylk error is about whitespace in the `vec` file that I have uploaded. 
@jayantj  As of now, this code reads additional word (vocab_size is one more than nwords), but skips loading any information into dictionary as it seems undesirable. Should we consider loading those information?

jayantj:
Thanks for the PR @prakhar2b 
As discussed, I believe the ideal fix for this would be loading FastText models from the `.bin` file only. It wouldn't require us to handle special cases in code, as is the case with the current fix. 

Related to issue #1261

prakhar2b:
@jayantj  yes, it was supposed to be only a workaround for now.

gensim/models/wrappers/fasttext.py
```diff
@@ -281,15 +288,22 @@ def load_model_params(self, file_handle):
...
+        if not bin_only:
+            assert len(self.wv.vocab) == nwords, 'mismatch between vocab sizes'
+            if len(self.wv.vocab) != vocab_size:
+                logger.warnings("If you are loading any model other than pretrained vector wiki.fr, ")
+                logger.warnings("Please report to gensim or fastText.")
+        else:
+            self.wv.syn0 = zeros((vocab_size, self.vector_size), dtype=REAL)
+            logger.info("here?")
+            # TO-DO : how to update this
+        ntokens= self.struct_unpack(file_handle, '@1q')
```
jayantj: Code style: Space before `=`

gensim/models/wrappers/fasttext.py
```diff
@@ -281,15 +288,22 @@ def load_model_params(self, file_handle):
...
-        assert len(self.wv.vocab) == nwords, 'mismatch between vocab sizes'
-        assert len(self.wv.vocab) == vocab_size, 'mismatch between vocab sizes'
-        self.struct_unpack(file_handle, '@1q')  # number of tokens
+        if not bin_only:
+            assert len(self.wv.vocab) == nwords, 'mismatch between vocab sizes'
+            if len(self.wv.vocab) != vocab_size:
+                logger.warnings("If you are loading any model other than pretrained vector wiki.fr, ")
+                logger.warnings("Please report to gensim or fastText.")
+        else:
+            self.wv.syn0 = zeros((vocab_size, self.vector_size), dtype=REAL)
```
jayantj: This doesn't belong here, we need to be loading the vectors for the in-vocab words after loading the ngram weight matrix (`syn0_all`)

gensim/models/wrappers/fasttext.py
```diff
@@ -281,15 +288,22 @@ def load_model_params(self, file_handle):
...
 
-    def load_dict(self, file_handle, encoding='utf8'):
+    def load_dict(self, file_handle, bin_only = False, encoding='utf8'):
         vocab_size, nwords, _ = self.struct_unpack(file_handle, '@3i')
         # Vocab stored by [Dictionary::save](https://github.com/facebookresearch/fastText/blob/master/src/dictionary.cc)
-        assert len(self.wv.vocab) == nwords, 'mismatch between vocab sizes'
-        assert len(self.wv.vocab) == vocab_size, 'mismatch between vocab sizes'
-        self.struct_unpack(file_handle, '@1q')  # number of tokens
+        if not bin_only:
+            assert len(self.wv.vocab) == nwords, 'mismatch between vocab sizes'
```
jayantj: Let's also log `vocab_size`/`nwords`/`len(self.wv.vocab)` in case of mismatches.

gensim/models/wrappers/fasttext.py
```diff
@@ -298,8 +312,24 @@ def load_dict(self, file_handle, encoding='utf8'):
                 char_byte = file_handle.read(1)
             word = word_bytes.decode(encoding)
             count, _ = self.struct_unpack(file_handle, '@qb')
-            assert self.wv.vocab[word].index == i, 'mismatch between gensim word index and fastText word index'
-            self.wv.vocab[word].count = count
+            if bin_only and word not in self.wv.vocab:
```
jayantj: This seems unnecessarily convoluted - can the word be in `self.wv.vocab` if we're loading from bin file only?

gensim/models/wrappers/fasttext.py
```diff
@@ -298,8 +312,24 @@ def load_dict(self, file_handle, encoding='utf8'):
...
+                self.wv.vocab[word].count = count
+
+            if bin_only:
+                #self.wv.syn0[i] = weight  # How to get weight vector for each word ?
+                self.wv.index2word.append(word)
+
+        if bin_only:
+            if self.wv.syn0.shape[0] != len(self.wv.vocab):
+                logger.info(
+                    "duplicate words detected, shrinking matrix size from %i to %i",
```
jayantj: Not sure I understand this. Can duplicate words exist?

gensim/test/test_fasttext_wrapper.py
```diff
@@ -171,6 +171,37 @@ def testLoadFastTextNewFormat(self):
         self.assertEquals(self.test_new_model.wv.min_n, 3)
         self.model_sanity(new_model)
 
+    def testLoadBinOnly(self):
+        """ Test model succesfully loaded from fastText (new format) .bin files only """
+        new_model = fasttext.FastText.load_fasttext_format(self.test_new_model_file, bin_only = True)
+        vocab_size, model_size = 1763, 10
+        self.assertEqual(self.test_new_model.wv.syn0.shape, (vocab_size, model_size))
```
jayantj: Shouldn't we be testing `new_model` and not `self.test_new_model`?

gensim/models/wrappers/fasttext.py
```diff
@@ -337,8 +367,26 @@ def init_ngrams(self):
...
+            word_ngrams = self.compute_ngrams(w, self.wv.min_n, self.wv.max_n)
+            all_ngrams += word_ngrams
+
+            
+            if bin_only:
+                #self.wv.syn0 = zeros((len(self.wv.vocab), self.vector_size), dtype=REAL)
+                word_vec = np.zeros(self.wv.syn0.shape[1])
+
+                num_word_ngram_vectors = len(word_ngrams)
+                for word_ngram in word_ngrams:
```
jayantj: This seems redundant and possibly error prone.
Instead you could simply use `self.word_vec(word)` to obtain the vector for the in-vocab words, after the `syn0_all` trimming takes place. Could you please try that, and see if it initializes the correct values for `self.syn0`?

jayantj: Okay, so from looking at the FastText code, it seems this is more complicated than what we originally thought.
The final in-vocab word vectors (aka the syn0 vectors, the vectors in the `.vec` file) are obtained by summing up - 
1. all vectors for its component ngrams
2. the vector for the original word itself 

The large weight matrix in the `.bin` file contains - 
1. vectors for the original words in the first `len(vocab)` rows
2. in the remaining rows, weights for component ngrams

This will require a change in logic then.

Possibly useful reference - https://github.com/facebookresearch/fastText/blob/master/src/dictionary.cc#L71

gensim/models/wrappers/fasttext.py
```diff
@@ -284,12 +285,12 @@ def load_model_params(self, file_handle):
     def load_dict(self, file_handle, encoding='utf8'):
         vocab_size, nwords, _ = self.struct_unpack(file_handle, '@3i')
         # Vocab stored by [Dictionary::save](https://github.com/facebookresearch/fastText/blob/master/src/dictionary.cc)
-        assert len(self.wv.vocab) == nwords, 'mismatch between vocab sizes'
-        assert len(self.wv.vocab) == vocab_size, 'mismatch between vocab sizes'
+        logger.info("loading projection weights")
```
jayantj: This message doesn't seem right, we're loading the vocabulary here, not the weights.

gensim/models/wrappers/fasttext.py
```diff
@@ -298,8 +299,16 @@ def load_dict(self, file_handle, encoding='utf8'):
                 char_byte = file_handle.read(1)
             word = word_bytes.decode(encoding)
             count, _ = self.struct_unpack(file_handle, '@qb')
-            assert self.wv.vocab[word].index == i, 'mismatch between gensim word index and fastText word index'
-            self.wv.vocab[word].count = count
+
+            if word != "__label__":
```
jayantj: Please add a note briefly clarifying this, since it's a hack-ish fix and likely to confuse any future developers.

gensim/models/wrappers/fasttext.py
```diff
@@ -298,8 +299,16 @@ def load_dict(self, file_handle, encoding='utf8'):
...
+                self.wv.vocab[word] = Vocab(index=i, count=count)
```
jayantj: Is this correct? The word `"__label__"` doesn't have a corresponding vector in the weight matrix, if I understand our previous discussion correctly. As a result, any of the words after the `"__label__"` word will have indices shifted.

gensim/models/wrappers/fasttext.py
```diff
@@ -348,6 +361,16 @@ def init_ngrams(self):
...
 
+        ngram_weights = self.wv.syn0_all
+
+        for w, v in self.wv.vocab.items():
+            word_ngrams = self.compute_ngrams(w, self.wv.min_n, self.wv.max_n)
+            for word_ngram in word_ngrams:
+                self.wv.syn0[self.wv.vocab[w].index] += np.array(ngram_weights[self.wv.ngrams[word_ngram]])
+
+            self.wv.syn0[self.wv.vocab[w].index] /= (len(word_ngrams) + 1)
+        logger.info("loaded %s matrix from %s" % (self.wv.syn0.shape))
```
jayantj: This message seems inconsistent with `self.wv.syn0.shape`

gensim/models/wrappers/fasttext.py
```diff
@@ -337,8 +346,12 @@ def init_ngrams(self):
         """
         self.wv.ngrams = {}
         all_ngrams = []
+        self.wv.syn0 = np.zeros((len(self.wv.vocab), self.vector_size), dtype=REAL)
+
         for w, v in self.wv.vocab.items():
             all_ngrams += self.compute_ngrams(w, self.wv.min_n, self.wv.max_n)
+            self.wv.syn0[self.wv.vocab[w].index] += np.array(self.wv.syn0_all[self.wv.vocab[w].index])
```
jayantj: `v` instead of `self.wv.vocab[w]`? Please rename the variable to something more verbose when making this change.

gensim/models/wrappers/fasttext.py
prakhar2b: Actually, you were right about it, this is the last word, but we can't skip reading it otherwise there will be error in further bytes reading.

gensim/models/wrappers/fasttext.py
prakhar2b: ok, I tried to replicate the logging that we were getting earlier.
https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/keyedvectors.py#L204

gensim/models/wrappers/fasttext.py
prakhar2b: ok, `vocab` should be fine ?

gensim/models/wrappers/fasttext.py
prakhar2b: Again, tried to replicate older logging
https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/keyedvectors.py#L266

prakhar2b:
@jayantj Also, we don't need this anymore `load_word2vec_format` I guess -  https://github.com/prakhar2b/gensim/blob/0fc11591f6477ebd67eb783ce242aaa39f078f71/gensim/models/wrappers/fasttext.py#L223

gensim/models/wrappers/fasttext.py
jayantj: Yes, but here we aren't loading exactly the same thing as in the case linked above - in that case, we're loading the vocab words along with their vectors, here we're simply building a dictionary.

gensim/models/wrappers/fasttext.py
jayantj: Yes, I understand we have to ready the bytes, agreed.
In that case, I'd prefer something more explicit and clear, like - 
```python
if i == nwords and i < vocab_size:
    assert word == "__label__"
   continue   # don't add word to vocab

File: gensim/models/wrappers/fasttext.py (hunk 11 above)
Comment (jayantj) [inline gensim/models/wrappers/fasttext.py]: Okay, but you're passing in a tuple, so the message is going to look something like `loaded 30000 matrix from 100`. Please update the message.

**Comment by jayantj** [comment]
Good point, let's remove this and any other unused code.

File: gensim/models/wrappers/fasttext.py [hunk 13]
```diff
@@ -298,8 +299,16 @@ def load_dict(self, file_handle, encoding='utf8'):
...
-            assert self.wv.vocab[word].index == i, 'mismatch between gensim word index and fastText word index'
-            self.wv.vocab[word].count = count
+
+            if word != "__label__":
+                self.wv.vocab[word] = Vocab(index=i, count=count)
+                self.wv.index2word.append(word)
+
+        assert len(self.wv.vocab) == nwords, 'mismatch between vocab sizes'
+        if len(self.wv.vocab) != vocab_size:
+            logger.warning("If you are loading any model other than pretrained vector wiki.fr, ")
```
jayantj: Any particular reason for two separate warning statements? Why not a single one?

gensim/models/wrappers/fasttext.py
piskvorky: Don't use `%` to format the string; pass the arguments as function arguments instead.

gensim/models/wrappers/fasttext.py
```diff
@@ -363,13 +377,15 @@ def init_ngrams(self):
 
         ngram_weights = self.wv.syn0_all
 
-        for w, v in self.wv.vocab.items():
+        logger.info("loading vocabulary weights")
```
piskvorky: Try to make the INFO messages more informative: loading how many weights, loading from where etc.

When there's a problem, it really helps when the log contains concrete numbers and messages. The problem often becomes apparent even without debugging.

gensim/models/wrappers/fasttext.py
```diff
@@ -220,6 +220,10 @@ def save(self, *args, **kwargs):
         super(FastText, self).save(*args, **kwargs)
 
     @classmethod
+    def load_word2vec_format(cls, *args, **kwargs):
+        return FastTextKeyedVectors.load_word2vec_format(*args, **kwargs)
```
gojomo: I believe that a load using this method *only* learns the full-word vectors as in the `.vec` file. If so, isn't it true that the resulting object doesn't have any other capabilities beyond a plain `KeyedVectors`? In that case, using a specialized class like `FastTextKeyedVectors` – that maybe is trying to do more, such as ngram-tracking, but inherently is not because that info was lost in the sequence-of-steps used to load it – seems potentially misleading. So unless I'm misunderstanding, I think this load-technique should use a plain `KeyedVectors`.

prakhar2b: yes, this method is **not used now** for loading using bin only. I removed this unused code, but got a [strange flake8 error](https://travis-ci.org/RaRe-Technologies/gensim/jobs/244678046) for python 3+, therefore re-added this for this PR. I'll try removing these unused codes later maybe in a different PR. @gojomo

gojomo: That is an odd error! I suspect it's not really the presence/absence of that method that triggered it, but something else either random or hidden in the whitespace.

gensim/models/wrappers/fasttext.py
```diff
@@ -233,11 +233,12 @@ def load_fasttext_format(cls, model_file, encoding='utf8'):
...
         FastText outputs two training files - `/path/to/train.vec` and `/path/to/train.bin`
-        Expected value for this example: `/path/to/train`
+        Expected value for this example: `/path/to/train`. However, you only need .bin
+        file to load the entire model.
 
         """
         model = cls()
-        model.wv = cls.load_word2vec_format('%s.vec' % model_file, encoding=encoding)
+        model.file_name = model_file
         model.load_binary_data('%s.bin' % model_file, encoding=encoding)
```
jayantj: I wonder if it would be a good idea to allow taking the whole model filename (including the .bin) as valid input - with the latest changes, we're loading from the bin file only, so it makes intuitive sense for the entire filename to be valid input. 
The only reason IMO we're still allowing the filename without extension as valid input is backward compatibility.

gensim/models/wrappers/fasttext.py
```diff
@@ -284,12 +285,12 @@ def load_model_params(self, file_handle):
     def load_dict(self, file_handle, encoding='utf8'):
         vocab_size, nwords, _ = self.struct_unpack(file_handle, '@3i')
         # Vocab stored by [Dictionary::save](https://github.com/facebookresearch/fastText/blob/master/src/dictionary.cc)
-        assert len(self.wv.vocab) == nwords, 'mismatch between vocab sizes'
-        assert len(self.wv.vocab) == vocab_size, 'mismatch between vocab sizes'
+        logger.info("loading vocabulary words for fastText model from %s.bin", self.file_name)
```
jayantj: Adding the number of words would be helpful in this logging statement.

gensim/models/wrappers/fasttext.py
```diff
@@ -298,8 +299,25 @@ def load_dict(self, file_handle, encoding='utf8'):
                 char_byte = file_handle.read(1)
             word = word_bytes.decode(encoding)
             count, _ = self.struct_unpack(file_handle, '@qb')
-            assert self.wv.vocab[word].index == i, 'mismatch between gensim word index and fastText word index'
-            self.wv.vocab[word].count = count
+
+            if i == nwords and i < vocab_size:
+                """
+                To handle the error in pretrained vector wiki.fr (French).
```
jayantj: This format - 
```
"""
Some comments
"""
```
 - is generally reserved for docstrings. Regular comments would be preferable.

gensim/models/wrappers/fasttext.py
```diff
@@ -298,8 +299,25 @@ def load_dict(self, file_handle, encoding='utf8'):
...
+                continue   # don't add word to vocab
+
+            self.wv.vocab[word] = Vocab(index=i, count=count)
+            self.wv.index2word.append(word)
+
+        assert len(self.wv.vocab) == nwords, 'mismatch between vocab sizes'
+        if len(self.wv.vocab) != vocab_size:
+            logger.warning("mismatch between vocab sizes")
+            logger.warning("If you are loading any model other than pretrained vector wiki.fr, ")
+            logger.warning("Please report to Gensim.")
```
jayantj: Please change the multiple warning statements to a single concatenated statement.

gensim/models/wrappers/fasttext.py
```diff
@@ -348,6 +370,18 @@ def init_ngrams(self):
             self.wv.ngrams[ngram] = i
         self.wv.syn0_all = self.wv.syn0_all.take(ngram_indices, axis=0)
 
+        ngram_weights = self.wv.syn0_all
+
+        logger.info("loading weights for %s vocabulary words for fastText models from %s.bin", len(self.wv.vocab), self.file_name)
```
jayantj: Typo: fastText model (not models)

gensim/test/test_fasttext_wrapper.py
```diff
@@ -115,7 +114,7 @@ def testNormalizedVectorsNotSaved(self):
         self.assertTrue(loaded_kv.syn0_all_norm is None)
 
     def testLoadFastTextFormat(self):
-        """Test model successfully loaded from fastText .vec and .bin files"""
+        """Test model successfully loaded from fastText .bin files"""
```
jayantj: Typo here and below: .bin file (not files)

gensim/models/wrappers/fasttext.py
piskvorky: Also, "Please report to Gensim" is vague, and probably unnecessary.

If people encounter bugs or get exceptions, they'll let us know, don't worry about that.

I'd prefer if the logging message was more concrete instead: what mismatch, what are the vocab sizes? We want the logs as concrete and useful as possible, for our own sanity (remote debugging via mailing list etc).

prakhar2b: @jayantj @piskvorky ok, updating logging info with `mismatch between vocab size, and number of vocab words ` and removing logging info about wiki.fr error.

gensim/models/wrappers/fasttext.py
prakhar2b: @gojomo ok, test passed this time after removing this code :smile:

gensim/models/wrappers/fasttext.py
prakhar2b: @jayantj I remember a few weeks ago, I received a review comment from @piskvorky that concatenated statement would contain white space therefore better to split into multiple statements.  Correct me if I didn't understand [that comment](https://github.com/RaRe-Technologies/gensim/pull/1200#discussion_r116380522)

jayantj: I believe @piskvorky meant to split the string itself across multiple lines, like this - 
```python
logger.warning(
	"mismatch between vocab sizes"
	"If you are loading any model other than pretrained vector wiki.fr, "
	"Please report to Gensim.")
```
He left a comment later in the PR clarifying it too.

prakhar2b: ohh, thanks for clarifying :smile:

jayantj:
Pushed some changes to model filename handling and logging.
Also, a minor fix to the flake8 diff script that was causing the weird utf8 error in python3+ discussed above.

gensim/models/wrappers/fasttext.py
jayantj: This was a bug in the flake8 script, fixed in cebb3fc16c5d588879bc2c7b0b8f84a3deed64e4

PR #1347, opened 2017-05-22
Explain new concepts introduced in corpus streaming example

oonska:
Add paragraph describing dictionary.dfs and dictionary.compactify()

In code snippet 13 there are two new concepts introduced that have not
been explained yet. In addition the workflow to create the dictionary
here is completely different from the workflow described in code
snippets 4 and 5. I've added a paragraph that tries to explain the new
workflow and concepts.


tmylk:
Thanks. Could you please put the explanation before the code? "We are going to ..."

oonska:
@tmylk Explaination moved.

tmylk:
Thanks!

PR #1348, opened 2017-05-22
Add section headings for parameters

MikeTheReader:
I was having a hard time understanding the switches between parameter descriptions, so added level-3 headers. Also made a couple of small grammar fixes.


tmylk:
Thanks for the changes!

PR #1355, opened 2017-05-22
Updated Annoy Tutorial with Text8

pmbaumgartner:
Updates:
- Runs in python 3
- Added required packages (in addition to gensim) to top of notebook
- Cleaned up some descriptive language throughout the notebook.
- Used Text8 Corpus rather than Lee corpus, since I wasn't seeing an improvement using the Lee corpus. My hypothesis is that it's too small to see any benefit from approximating. 
- Added watermark info for benchmarking comparisons
- Reorganized and restructured notebook
  - Added an outline according to the logic of the notebook
  - separated notebook cells for import, function definition, and executing code
  - Added optional logging
  - Combined num_trees performance analysis into one loop & plots into one figure and upped num_trees to 300

@tmylk

### DIFF
--- a/docs/notebooks/annoytutorial-text8.ipynb
+++ b/docs/notebooks/annoytutorial-text8.ipynb
@@ -1,0 +1,743 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Similarity Queries using Annoy Tutorial"
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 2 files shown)


tmylk:
Thanks a lot! Let's keep the old name and replace the Lee corpus ipynb with this time.

pmbaumgartner:
Okay, should be good now, just renamed it.

tmylk:
awesome thx

piskvorky:
Beautiful!

Issue #1323, opened 2017-05-15
Fix wrapper tests


menshikh-iv:
After running all tests I found some problems

1. [ldavw wrapper](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_ldavowpalwabbit_wrapper.py#L223)
(replace assertEqual to assertAlmostEqual for second element)
```
Traceback (most recent call last):
  File "/home/ivan/release/test/gensim/gensim/test/test_ldavowpalwabbit_wrapper.py", line 223, in testvwmodel2ldamodel
    self.assertEqual(tm1[document][0], tm2[document][0])
AssertionError: Tuples differ: (1, 0.99731147) != (1, 0.99731144183873388)

First differing element 1:
0.99731147
0.99731144183873388

- (1, 0.99731147)
+ (1, 0.99731144183873388)
?              +++++ ++++
```

2. [ldamallet wrapper](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_ldamallet_wrapper.py#L110)
(replace assertEqual to assertAlmostEqual for second element with places=1)
```
Traceback (most recent call last):
  File "/home/ivan/release/test/gensim/gensim/test/test_ldamallet_wrapper.py", line 110, in testMallet2Model
    self.assertEqual(tm1[document][0], tm2[document][0])
AssertionError: Tuples differ: (0, 0.5157232704402516) != (0, 0.50981136383330472)

First differing element 1:
0.5157232704402516
0.509811363833

- (0, 0.5157232704402516)
+ (0, 0.50981136383330472)
```

3. [varembed wrapper](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_varembed_wrapper.py#L62)
(remove test because we don't support 2.6 from 1.0.0)

4. [dtm wrapper](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_dtm.py#L52)
```
Traceback (most recent call last):
  File "/home/ivan/release/test/gensim/gensim/test/test_dtm.py", line 55, in testDim
    rng_seed=1)
  File "/home/ivan/release/test/gensim/gensim/models/wrappers/dtmmodel.py", line 96, in __init__
    if model == "fixed" and any([i == 0 for i in [len(text) for text in corpus.get_texts()]]):
AttributeError: 'MmCorpus' object has no attribute 'get_texts'
```

[...]

PR #1359, opened 2017-05-23
Fixed wrapper tests #1323

shubhamjain74:


menshikh-iv:
Thank you @shubhamjain74, It's a nice start :+1: 
I check this PR with full test run on 2.7 and 3.5

setup.py
```diff
@@ -281,6 +281,7 @@ def finalize_options(self):
         'scipy >= 0.7.0',
         'six >= 1.5.0',
         'smart_open >= 1.2.1',
+        'morfessor==2.0.2alpha4',
```
piskvorky: -1: why should this be a hard dependency??

Has this been thoroughly vetted, by @tmylk and @gojomo ? Is `morfessor` really a core, indispensable part of gensim now, with all that it entails (we'll have to support its installation problems etc).

gensim/models/wrappers/ldamallet.py
```diff
@@ -190,7 +190,7 @@ def load_word_topics(self):
         if hasattr(self.id2word, 'token2id'):
             word2id = self.id2word.token2id
         else:
-            word2id = dict((v, k) for k, v in iteritems(self.id2word))
+            word2id = dict((v, k) for k, v in iteritems(dict(self.id2word)))
```
piskvorky: What is this `dict` casting this for?

gensim/models/wrappers/dtmmodel.py
```diff
@@ -93,7 +93,7 @@ def __init__(
             lencorpus = sum(1 for _ in corpus)
         if lencorpus == 0:
             raise ValueError("cannot compute DTM over an empty corpus")
-        if model == "fixed" and any([i == 0 for i in [len(text) for text in corpus.get_texts()]]):
+        if model == "fixed" and any([i == 0 for i in [len(text) for text in corpus]]):
```
piskvorky: No need for the outer brackets `[` and `]` (generator fine).

But the whole construction feels unpythonic and strange -- simple `any(not text for text in corpus)` better?

gensim/models/wrappers/ldamallet.py
menshikh-iv: It's for python3 compatibility (if self.id2word is `FakeDict` object, `iteritems` can't work correctly)

setup.py
menshikh-iv: It's dependency for `VarEmbed` wrapper. Maybe need to create several mode for install.

gensim/models/wrappers/ldamallet.py
piskvorky: Let's update `utils.revdict` then, and use that consistently. There are probably more places like this in the code.

setup.py
piskvorky: Definitely. See the optional `extras_require` setup param.

PR #879, opened 2016-09-24
Fix : Bug in Phrases.export_phrases()

AadityaJ:
Refer to the Issue : [#794](https://github.com/RaRe-Technologies/gensim/issues/794)

Issue #794, opened 2016-07-16
Bug in Phrases.export_phrases()


jeradf:
A small bug in `bigram.export_phrases(sentences)`  causes it to return a maximum of one bigram per sentence.
For example:

```
from gensim.models import Phrases

sentences = [
    ['human', 'interface', 'computer'],
    ['survey', 'user', 'computer', 'system', 'response', 'time'],
    ['eps', 'user', 'interface', 'system'],
    ['system', 'human', 'system', 'eps'],
    ['user', 'response', 'time'],
    ['trees'],
    ['graph', 'trees'],
    ['graph', 'minors', 'trees'],
    ['graph', 'minors', 'survey','human','interface']]

bigram = Phrases(sentences, min_count=1, threshold=2)
for phrase, score in bigram.export_phrases(sentences):
    print(u'{0}   {1}'.format(phrase, score))
```

Returns:

```
>>> human interface   3.44444444444
response time   7.75
response time   7.75
graph minors   5.16666666667
graph minors   5.16666666667
```

The last sentence has two bigrams, but only the first is returned. This happens because this line
https://github.com/RaRe-Technologies/gensim/blob/master/gensim/models/phrases.py#L216
prevents further iteration within a sentence.
If it's commented out, it correctly returns:

```
>>> human interface   3.44444444444
response time   7.75
response time   7.75
graph minors   5.16666666667
graph minors   5.16666666667
human interface   3.44444444444
```

gojomo:
Good find! 

I believe the intent of `last_bigram` is to prevent a word from being in two overlapping bigrams; the bug may be that `last_bigram` is not toggled back to `False` at the end of each non-bigram iteration. 

So perhaps best fix is to add `continue` after the toggle-on, and at bottom of (non-continued/non-bigrammed) loop block, add a `last_bigram = False` line.

AadityaJ:
Hi everyone!
Today I worked on this. I am sending a pull request.

gojomo:
[...]

PR #1362, opened 2017-05-23
fixing a single phrase not returning back multiple bigrams, #794

toumorokoshi:
Here's another try at fixing #794. The previous PR #879 added a unit test, but didn't catch the issue because the bigram "human interface" was already included from the very first entry. Added a more targetted unit test to catch this.

### DIFF
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -250,7 +250,7 @@ class Phrases(interfaces.TransformationABC):
                                 yield (out_delimiter.join((word_a, word_b)), score)
                             last_bigram = True
                             continue
-                        last_bigram = False
+                    last_bigram = False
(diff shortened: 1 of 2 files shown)


tmylk:
Thanks. Let's remove the print and LGTM when tests pass

gojomo:
Thanks! To ensure the same confusion as befell the 1st fix attempt doesn't occur – have you verified the updated test does fail, without the updated fix?

toumorokoshi:
@gojomo yes! Although I didn't commit the test first separately.

A unit test was added in the previous PR, and it did pass. However, it wasn't verifying the behavior, as the expected bigram was already added by a previous sentence.

My unit test isolates the actual behavior.

PR #1228, opened 2017-03-21
Updated word2vec.ipynb to include predict_output_word example

chinmayapancholi13:
Added an example in https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/word2vec.ipynb for the function `predict_output_word` added in PR #1209.

PR #1369, opened 2017-05-27
Minor changes in fresh PRs (pep8, misprints, etc)

menshikh-iv:
I have fixed all recent comments from you @piskvorky + some code style fixes. Now it's okay?

cc @tmylk

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -533,7 +533,17 @@ def jaccard(vec1, vec2):
 
 
 def jaccard_set(set1, set2):
-    return 1. - float(len(set1 & set2)) / float(len(set1 | set2))
+    """
+    A distance metric between set representation.
+    Returns 1 minus the intersection divided by union.
+    Returns a value in range <0, 1> where values closer to 0 mean less distance and thus higher similarity.
+    """
+
+    union_cardinality = len(set1 | set2)
+    if union_cardinality == 0:  # Both sets are empty
+        return 1.
+
+    return 1. - float(len(set1 & set2)) / float(union_cardinality)
 
 
 def dirichlet_expectation(alpha):
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -532,11 +532,10 @@ def jaccard(vec1, vec2):
         return 1 - float(len(intersection)) / float(len(union))
 
 
-def jaccard_set(set1, set2):
+def jaccard_distance(set1, set2):
     """
-    A distance metric between set representation.
-    Returns 1 minus the intersection divided by union.
-    Returns a value in range <0, 1> where values closer to 0 mean less distance and thus higher similarity.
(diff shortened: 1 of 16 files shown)


chinmayapancholi13:
@menshikh-iv Could you please also incorporate a minor PEP8 suggestion made by @piskvorky [here](https://github.com/RaRe-Technologies/gensim/pull/1228#discussion_r110534838). Since you are making all such fixes in this PR, I though it would be a good idea to bring this change to your notice as well. :)

gensim/matutils.py
```diff
@@ -533,7 +533,17 @@ def jaccard(vec1, vec2):
 
 
 def jaccard_set(set1, set2):
-    return 1. - float(len(set1 & set2)) / float(len(set1 | set2))
+    """
+    A distance metric between set representation.
+    Returns 1 minus the intersection divided by union.
```
piskvorky: PEP257: docstrings in imperative mode ("do X, return Y", not "does X, returns Y").

gensim/matutils.py
```diff
@@ -533,7 +533,17 @@ def jaccard(vec1, vec2):
...
+    Returns a value in range <0, 1> where values closer to 0 mean less distance and thus higher similarity.
```
piskvorky: `less` => `smaller`.

I see this is a potential gotcha for our users, because other metrics in gensim are similarity (not distance). How about we put this info directly into the method name -- `jaccard_distance` instead of `jaccard_set`?

menshikh-iv:
@chinmayapancholi13 thank you

menshikh-iv:
@piskvorky Need to walk with an automatic tool throughout the code (especially the old one) and fix code style issues (it will be a very big diff)

menshikh-iv:
But this is very important. New users come and write code according to the code they see (this code is not perfect).

piskvorky:
Agreed! Fixing the existing code has been our (Lev's) priority for a long time now.

Except don't use an automatic tool please (except as rough guidance). Use common sense to make the core readable and easy to understand, and at the same time catch other non-PEP8 lapses too: direct formatting of strings inside `logger.info`, comprehensions (we're now 2.7+, no longer support 2.5 or 2.6, so the syntax is much simpler) etc.

tmylk:
Automatic linters are the solution for code style adopted in most projects, for example see this [presentation](https://www.youtube.com/watch?v=iNG1a--SIlk) from itertools maintainer.

piskvorky:
There was one PR produced by an automatic linter recently, and that's definitely not what we want in gensim.

menshikh-iv:
I think It's a good idea to use linter for highlight the mistakes in codestyle, but fix it manually. Also, now flake8 break the build if some PEP8 problem founded in PR code and a user can fix them by reading the Travis log.

PR #1374, opened 2017-05-29
Lda difference visualization

menshikh-iv:
You can see result with [nbviewer link with interactive graphs](http://nbviewer.jupyter.org/github/menshikh-iv/gensim/blob/ebdbebc56a4dafb583ee54bd94ddeb94dabd016b/docs/notebooks/lda_model_difference.ipynb)

Continuing #1243


tmylk:
Please explain the annotation with '++' and '--' a little bit more. it's explained already, just need to mention the `+` and `-` symbols explicitly.

menshikh-iv:
Done!

tmylk:
It is better than before but please add a very simple viz where it's just two topics in order to explain the idea more clearly.

piskvorky:
There's no way to review a notebook (on github), so I'll comment here instead:

1. add full stops = `.` at the end of sentences (many places)
2. prefer list/set/dict comprehensions in place of `filter`/`map`
3. use hanging indent (!)

parulsethi:
@menshikh-iv An example that I think could be helpful for user: 
After fitting the LDA models, print an example using `show_topics()` and then describe that the diff. we are calculating is basically between the topic's probability distributions that we see above.

Maybe, illustrate the process also by taking an example cell in the matrix - print both topic distribution of that cell and then use distance function on these 2 distribution(vectors) to print the distance which comes out to be of same value as in the matrix cell.

PR #1377, opened 2017-06-01
Fix train error of ConcatenatedDoc2Vec in the notebook of doc2vec-IMDB

robotcator:


docs/notebooks/doc2vec-IMDB.ipynb
```diff
@@ -610,7 +610,10 @@
     "        duration = 'na'\n",
     "        train_model.alpha, train_model.min_alpha = alpha, alpha\n",
     "        with elapsed_timer() as elapsed:\n",
-    "            train_model.train(doc_list, total_examples=train_model.corpus_count, epochs=train_model.iter)\n",
+    "            if not isinstance(train_model, ConcatenatedDoc2Vec):\n",
+    "              train_model.train(doc_list, total_examples=train_model.corpus_count, epochs=train_model.iter)\n",
+    "            else:\n",
+    "              train_model.train(doc_list)\n",
```
gojomo: A simpler and more robust fix would be to change the ConcatenatedDoc2Vec class, in `test_doc2vec.py`, to make its (no-op) `train()` match the new `train()` parameters-signature.

robotcator: Yes, you are right.  If the train() method is modified, the total_examples and epochs should be provided. But the ConcatenatedDoc2Vec class has no attribute 'corpus_count'.

gojomo: The call doesn't have to use `train_model.corpus_count` from inside the model - it can just use `len(doc_list)`. And since the outside loop is handling the multiple passes, the `epochs` argument should be `1`.

menshikh-iv:
@robotcator Please merge `develop` to your branch (missing tensorflow in Travis config).

menshikh-iv:
@robotcator I open this notebook and get error `Notebook validation failed: u'execution_count' is a required property: ...`, can you please re-run all cells and commit it (this should help)

robotcator:
@menshikh-iv all the cells were re-ran and the error was fixed.

docs/notebooks/doc2vec-IMDB.ipynb
```diff
@@ -437,159 +388,166 @@
...
-      "*0.220720 : 1 passes : dbow+dmc 0.0s 1.1s\n",
-      "*0.189200 : 1 passes : dbow+dmc_inferred 0.0s 19.3s\n",
+      "START 2017-06-06 15:19:50.208091\n",
+      "*0.408320 : 1 passes : Doc2Vec(dm/c,d100,n5,w5,mc2,s0.001,t4) 131.9s 33.6s\n",
+      "*0.341600 : 1 passes : Doc2Vec(dm/c,d100,n5,w5,mc2,s0.001,t4)_inferred 131.9s 48.3s\n",
+      "*0.239960 : 1 passes : Doc2Vec(dbow,d100,n5,mc2,s0.001,t4) 35.3s 45.9s\n",
+      "*0.193200 : 1 passes : Doc2Vec(dbow,d100,n5,mc2,s0.001,t4)_inferred 35.3s 48.3s\n",
+      "*0.268640 : 1 passes : Doc2Vec(dm/m,d100,n5,w10,mc2,s0.001,t4) 48.6s 48.5s\n",
+      "*0.208000 : 1 passes : Doc2Vec(dm/m,d100,n5,w10,mc2,s0.001,t4)_inferred 48.6s 47.4s\n",
+      "*0.216160 : 1 passes : dbow+dmm 0.0s 168.9s\n",
```
gojomo: The change in reported elapsed-time for these concatenated-model results, here and below, is so large it deserves a closer look. I'm no aware of any change, in this PR or prior related work, that could account for such changes – from <2.0 seconds to >170 seconds. It might be a spurious report, or something amiss.

robotcator: yes, the concatenated-model train method didn't do anything actually. I will check what' going on this.

menshikh-iv:
@robotcator What's a status of this PR?

menshikh-iv:
Ping @robotcator

robotcator:
@menshikh-iv sorry for late reply, I didn't receive last notification.  This PR seems works fine and all the cell have been rerun. The warning of notebook is gone. It's able to be merged .

gojomo:
There's still the suspiciously-different reported-runtimes on some of the log lines.

menshikh-iv:
@gojomo As I see, we have only two code change [here](https://github.com/RaRe-Technologies/gensim/pull/1377/files#diff-83d57c0db44869cef461bc453d5f13abR571), and [here](https://github.com/RaRe-Technologies/gensim/pull/1377/files#diff-7268265e58ffaa87707f3d358f25ae17R411). Another change from notebook re-run. I don't see any suspicious things in code changes, although elapsed time looks strange.

gojomo:
In fact since the last time the notebook was run and its output committed, there have been many changes (outside this PR). So many things could cause the discrepancy; I just suggest the big runtim difference be understood alongside the updated notebook cells.

piskvorky:
That's indeed very worrying. @menshikh-iv can you find which commit / PR caused this slowdown?

robotcator:
is there any possible that the performance of computer which results in the difference of output. Because I rerun that notebook twice and the second seems alright.

menshikh-iv:
@piskvorky accepted. I will search and fix if needed.

Issue #1357, opened 2017-05-23
loading ensemble embeddings take wordrank .word and context file where these files are located i did'nt get this. Is this embeddings are generated by demo script given in wordrank or by your train function given in wordrank wrapper


gauravsaxenaiiit:
#### Description
TODO: change commented example

tmylk:
CC @parulsethi Let's add more comments in the docstring

parulsethi:
@gauravsaxenaiiit You can set `cleanup_files=False` to keep the .word and .context files generated by wordrank, and then they can be located in wordrank directory. And embeddings are generted by gensim wrapper's train() not demo script. 

@tmylk Sure

menshikh-iv:
Implemented in #1066

PR #1378, opened 2017-06-01
Add docstrings for Wordrank 

parulsethi:
1. Added description for wordrank training 
2. Corrected `output_dir` param (previous one would give error at `.load_wordrank_model()` which loads files from `output_dir`)
3. fixed example doc in wordrank file (fix https://github.com/RaRe-Technologies/gensim/issues/1384)


tmylk:
Just the style failures `W291 trailing whitespace`

gensim/models/wrappers/wordrank.py
```diff
@@ -47,8 +47,12 @@ class Wordrank(KeyedVectors):
     @classmethod
     def train(cls, wr_path, corpus_file, out_name, size=100, window=15, symmetric=1, min_count=5, max_vocab_size=0,
               sgd_num=100, lrate=0.001, period=10, iter=90, epsilon=0.75, dump_period=10, reg=0, alpha=100,
-              beta=99, loss='hinge', memory=4.0, cleanup_files=True, sorted_vocab=1, ensemble=0):
+              beta=99, loss='hinge', memory=4.0, cleanup_files=False, sorted_vocab=1, ensemble=0):
```
menshikh-iv: What is reason for change `cleanup_file` to `False`?

parulsethi: `cleanup_files=False` will not delete the (word/context) embedding files and vocab file generated by wordrank during training, which are saved inside wordrank's directory . Though the `train()` method [loads the final required embedding file](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/wrappers/wordrank.py#L159) before deleting everything that was generated during training but it could be [confusing to users](https://github.com/RaRe-Technologies/gensim/issues/1357) who expect to find it after the training is finished.
So, making the default behavior to not delete them could be better to avoid confusion.

menshikh-iv: Please enumerate output files (filename and what the file contains) in docstring.

parulsethi: I've added the output filenames and content info. in `out_name` param description because it is the directory which contain these files.

menshikh-iv:
Thank you @parulsethi :+1:

PR #1379, opened 2017-06-01
Fix doc2vec __init__ documentation

jstol:
Fixed descriptions of various arguments passed from `Doc2Vec.__init__` method up to the parent `Word2Vec.__init__` method.

- `alpha`: says will drop to 0 as training progressed, but should say `min_alpha`.
- `hs`: is said to be 1 by default, but is 0 by default (and also mentions "hierarchical sampling", rather than "hierarchical softmax").
- `negative`: doesn't mention that the default is 5.


gojomo:
+1

menshikh-iv:
Thank you @jstol  :+1:

PR #1385, opened 2017-06-02
[WIP] Cythonizing phrases module

prakhar2b:
Performance Improvement in Phrases module.
For context - [link to my gsoc live blog](https://rare-technologies.com/google-summer-of-code-2017-live-blog-performance-improvement-in-gensim-and-fasttext/)

[Phrases optimization benchmark](https://github.com/prakhar2b/gensim/blob/any2utf8/docs/notebooks/phrases-optimization-benchmark.ipynb) (**for text8**)

| Optimization       | Python 2.7     | Python 3.6 | PR |
| ------------- |:-------------:| :------------:| :---------------:|
|    original  | ~ 36-38 sec | ~32-35 sec |
| cython (static typing)      | ~30-32 sec    |
| any2utf8 (without cython)| ~20-22 sec     | ~23-26 sec |  #1413  |
| cython (with any2utf8)| ~15-18 sec     |  ~19-21 sec |  This PR |


gensim/models/phrases_inner.pyx
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env cython
+# cython: boundscheck=False
```
piskvorky: Better leave as `True`, during debugging.

gensim/models/phrases.py
```diff
@@ -105,6 +109,44 @@ class Phrases(interfaces.TransformationABC):
     and `phrases[corpus]` syntax.
 
     """
+    #from gensim.models.phrases_inner import learn_vocab
+    try:
+        from gensim.models.phrases_inner import learn_vocab
+        logger.info("Cython file loaded")
+    except ImportError:
+        logger.info("Cython file not loaded")
+        #failed... fall back to plain numpy (20-80x slower training than the above)
```
piskvorky: Is the 20-80x figure correct? If not, better remove the stale comments and start development with a clean slate.

prakhar2b:
@piskvorky Thanks for making comments here. yes, I'll document the progress as I go along. So far, I've been working on `text8` corpus, I'll test it for English wiki too for documentation. 

Also, I've been researching a lot about optimization and currently looking to cythonize and parallelize the code. It would be great to have some advise from you regarding optimizing (specifically, for phrases module) considering your experience with word2vec optimization. :smile:

gensim/models/phrases.py
```diff
@@ -105,6 +109,44 @@ class Phrases(interfaces.TransformationABC):
...
+            sentence_no = -1
+            total_words = 0
+            logger.info("collecting all words and their counts")
+            vocab = defaultdict(int)
+            min_reduce = 1
+            for sentence_no, sentence in enumerate(sentences):
+                if sentence_no % progress_per == 0:
+                    logger.info("PROGRESS: at sentence #%i, processed %i words and %i word types" %
+                                (sentence_no, total_words, len(vocab)))
+                #sentence = [utils.any2utf8(w) for w in sentence]
```
jayantj: @piskvorky was there any particular reason behind creating the vocab for `Phrases` with utf8-encoded bytestrings, rather than unicode strings themselves?
Currently, according to profiling done by Prakhar, the utf8 conversion significantly affects performance due to overhead in the conversion and the fact that the conversion is done for every individual word.

piskvorky: Yes, saving memory.

Up to Python 3.3 (and including all Python 2.x), unicode strings take up 2-4x as much memory, compared to UTF8 byte strings.

Since memory is more critical than speed here, we went with UTF8.

piskvorky:
@prakhar2b sure, I'll be happy to help.

Before any Cythonization or parallelization, decide on the architecture of the system -- what kind of inputs it will accept (probably any iterable of hashable items), what algorithms, what data structures. This is a critical piece of functionality, so we want something fast and memory-efficient, see #508, #556.

Then implement that in pure Python for reference.

Then profile and rewrite hotspots in Cython, for performance.

Finally, see if parallelization makes sense -- probably something more coarse grained, since updating counters is such a tiny operation. Maybe accept multiple iterables on input and update on all in parallel (the use-case being, people will supply readers from multiple files at once, the algo will consume them in parallel).

prakhar2b:
**Update so far - (text8)**


| Optimization       | Time      |
| ------------- |:-------------:| 
|    original  | ~ 36-38 sec |
| cython (static typing)      | ~30-32 sec    | 
| any2utf8 (without cython)| ~20-22 sec     |  
| any2utf8 (with cython)| ~15-18 sec     |

jayantj:
Please add the benchmark notebook we discussed to this PR - along with clear labels about which version of commit (commit hash) was used to create which benchmark.

gensim/models/phrases.py
```diff
@@ -171,17 +175,18 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
...
             raise ValueError("threshold should be positive")
 
+        self.recode_to_utf8 = recode_to_utf8
         self.min_count = min_count
         self.threshold = threshold
         self.max_vocab_size = max_vocab_size
         self.vocab = defaultdict(int)  # mapping between utf8 token => its count
         self.min_reduce = 1  # ignore any tokens with count smaller than this
-        self.delimiter = delimiter
+        self.delimiter = delimiter if recode_to_utf8 else utils.any2unicode(delimiter)
```
piskvorky: How does this work? Why is the delimited unicode if we're **not** recoding (using bytestrings)?

Deserves a comment at least.

gensim/models/phrases.py
```diff
@@ -71,6 +71,45 @@
 
 logger = logging.getLogger(__name__)
 
+try:
+    from gensim.models.phrases_inner import learn_vocab
+except ImportError:
+    logger.info("failed to load cython")
```
piskvorky: But we're not loading cython -- needs a better message (also more descriptive).

menshikh-iv:
I close this PR because this code doesn't give needed performance (expected minimum x10 through cython but give only x2.5 through any2utf hack)

PR #1389, opened 2017-06-03
Update logic to set 'updatetype' variable in LdaModel.py

chinmayapancholi13:
This PR updates the logic to set `updatetype` variable used in `ldamodel.py`. Now, we have 3 modes for training the model : `online (single-pass)`, `online (multi-pass)` and `batch`.


chinmayapancholi13:
@tmylk I have changed the logic for setting the variable `updatetype`. Instead of adding a new variable (like `streamingtype`), I have made changes to the same variable (i.e. `updatetype`) so that no logging changes are needed.

menshikh-iv:
Thanks @chinmayapancholi13

PR #1390, opened 2017-06-03
1307-docfix

menshikh-iv:
Fix docstring in accordance with [this comment](https://github.com/RaRe-Technologies/gensim/pull/1307#discussion_r119010587)
CC: @piskvorky

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -580,7 +580,7 @@ class Doc2Vec(Word2Vec):
         need about 1GB of RAM. Set to `None` for no limit (default).
 
         `sample` = threshold for configuring which higher-frequency words are randomly downsampled;
-                default is 1e-3, useful value is 1e-5.
+                default is 1e-3, values to 1e-5 (or lower) may also be useful.
 
         `workers` = use this many worker threads to train the model (=faster training with multicore machines).
 
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -580,7 +580,7 @@ class Doc2Vec(Word2Vec):
         need about 1GB of RAM. Set to `None` for no limit (default).
 
         `sample` = threshold for configuring which higher-frequency words are randomly downsampled;
-                default is 1e-3, values to 1e-5 (or lower) may also be useful.
+                default is 1e-3, values of 1e-5 (or lower) may also be useful, value 0. disable downsampling.
 
         `workers` = use this many worker threads to train the model (=faster training with multicore machines).
 


gensim/models/doc2vec.py
```diff
@@ -580,7 +580,7 @@ def __init__(self, documents=None, dm_mean=None,
         need about 1GB of RAM. Set to `None` for no limit (default).
 
         `sample` = threshold for configuring which higher-frequency words are randomly downsampled;
-                default is 1e-3, useful value is 1e-5.
+                default is 1e-3, values to 1e-5 (or lower) may also be useful.
```
piskvorky: `values to` => `values of`. Also, mention the corner case option of "no downsampling" value.

menshikh-iv: About "no downsampling" case, see examples

with `sample = 0.999999`
```
INFO:gensim.models.word2vec:sample=0.999999 downsamples 0 most-common words
```

with `sample = 1.`
```
INFO:gensim.models.word2vec:sample=1 downsamples 21143 most-common words
```

The reason is [alternative logic](https://github.com/menshikh-iv/gensim/blob/30a8c8f2fbbc95585b2322992bdd8f09ba29fa12/gensim/models/word2vec.py#L672) when `sample >= 1`

piskvorky: Aha, so it doesn't actually work? I'd consider that a bug. @gojomo thoughts?

menshikh-iv: It's work, but we have two different variants of `sample` usage.

gojomo: Going back to word2vec.c, `sample=0` meant to turn it off. Values 1.0 and higher were redundant (all having the same 'turn-it-off' meaning). So a convenience hack was added that if you wanted to start downsampling not at some hard-to-calculate tiny float, but a specific word count, you could specify an int `sample`, >=1, and the model would adjust the internal formula so that downsampling would start at that word-frequency. I used it for some experiments, and it might be more intuitive to tune than the float-value, but it's not documented beyond the inline comment. 

IIRC FastText then adopted the convention that `sample=1.0` means off (which has a certain consistency with the downsampling algorithm), but creates the potential for confusion at this boundary number. 

Everything "works" for one purpose or another but might not be clear. 

Really this bit of doc-comment should go into word2vec.py - where sample is defined. (Here, it's merely inherited – and for those values maybe the Doc2Vec comment should just say 'see Word2Vec'.) And, any guidance here should be kept in sync with the `add_parameter` doc-text for command-line invocation, defined at the bottom of word2vec.py.

menshikh-iv: Thank you @gojomo for the detailed answer. I hope I will do this in the process of refactoring the documentation.

gensim/models/doc2vec.py
```diff
@@ -580,7 +580,7 @@ def __init__(self, documents=None, dm_mean=None,
         need about 1GB of RAM. Set to `None` for no limit (default).
 
         `sample` = threshold for configuring which higher-frequency words are randomly downsampled;
-                default is 1e-3, useful value is 1e-5.
+                default is 1e-3, values of 1e-5 (or lower) may also be useful, value 0. disable downsampling.
```
piskvorky: `value 0. disable downsampling` => `set to 0.0 to disable downsampling`.

Issue #74, opened 2012-01-29
Add more data sanity checks


piskvorky:
There's been a steady trickle of reports that LSI/LDA misbehave, produce degenerate models, crash Python etc. 

Typically this is a user data problem (bad input data, feature id mismatch, ...), but since gensim targets the wide general public, this is gensim's "fault" anyway.

Create utility functions that perform basic sanity checks on user's input data:
1. check that the all feature ids in a corpus are compatible with the user-provided dictionary (should avoid issues like http://projects.scipy.org/scipy/ticket/1582 )
2. check that the data range is valid -- look for NaNs, Infs, explicit zeros => these are all illegal in gensim input.
3. check that the data is not degenerate => all vectors identical/empty/?/model looks weird
4. ~check corpus type and warn the user if it's plain list (promote the memory-friendly generator interface, shown in tutorials)~ NOT NEEDED

prakhar2b:
@tmylk I would like to work on this. @piskvorky the link above is broken, could you brief me what it was about ?

piskvorky:
@prakhar2b I think it was a scipy crash (segfault) when using sparse arrays and indexing an element out-of-bounds.

I wouldn't say this issues is "easy" -- it will need some careful thinking and planning. We definitely don't want to slow down processing too much, by (for example) requiring an extra data pass just to check for bad values.

rasto2211:
I would like to work on this issue. Could you please give me some pointers to the code where to start?

tmylk:
@rasto2211  Adding a warning to `LdaModel.init` when the input is a list is a good way to start ( item 4 above)

rasto2211:
@menshikh-iv  Do you also want to close this issue since you closed my PR without merging?

menshikh-iv:
@rasto2211 No, because the remaining points are important (see [radim comment](https://github.com/RaRe-Technologies/gensim/pull/1392#issuecomment-308373375))

menshikh-iv:
Be attentive with #1732, I already see exactly same problem twice

PR #1392, opened 2017-06-04
Warn the user if the corpus for LDA/LSI is plain list

rasto2211:
Encourage the user to use the memory-friendly generator interface
presented in the tutorial on corpus streaming. See the issue #74 .

I have decided to move the warning ldamodel.update and lsimodel.add_documents since both of the methods are called by constructors.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -579,6 +579,15 @@ class LdaModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
         if gamma_threshold is None:
             gamma_threshold = self.gamma_threshold
 
+        if isinstance(corpus, list):
+            logger.warning(
+                "taking plain list as a corpus for LDA;"
+                "we strongly recommend you to use corpus streaming instead of lists;"
+                "LDA is processing the corpus in batches and"
+                "it only loads one batch of documents into memory at once;"
+                "please take a look at the tutorial on corpus streaming: "
+                "https://radimrehurek.com/gensim/tut1.html#corpus-streaming-one-document-at-a-time")
+
         try:
             lencorpus = len(corpus)
         except:
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -581,10 +581,10 @@ class LdaModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
 
         if isinstance(corpus, list):
             logger.warning(
-                "Taking plain list as a corpus for LDA."
+                "Taking plain list as a corpus for LDA. "
                 "We strongly recommend you to use corpus streaming instead of lists. "
                 "LDA is processing the corpus in batches and "
-                "it only loads one batch of documents into memory at once."
+                "it only loads one batch of documents into memory at once. "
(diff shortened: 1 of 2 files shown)


gensim/models/ldamodel.py
```diff
@@ -579,6 +579,15 @@ def update(self, corpus, chunksize=None, decay=None, offset=None,
         if gamma_threshold is None:
             gamma_threshold = self.gamma_threshold
 
+        if isinstance(corpus, list):
+            logger.warning(
```
piskvorky: What does this warning message look like when actually emitted?

The spacing around line breaks and punctuation doesn't seem right.

menshikh-iv: It looks like this
```
taking plain list as a corpus for LDA;we strongly recommend you to use corpus streaming instead of lists;LDA is processing the corpus in batches andit only loads one batch of documents into memory at once;please take a look at the tutorial on corpus streaming: https://radimrehurek.com/gensim/tut1.html#corpus-streaming-one-document-at-a-time
```

gensim/models/lsimodel.py
```diff
@@ -348,6 +348,15 @@ def add_documents(self, corpus, chunksize=None, decay=None):
         """
         logger.info("updating model with new documents")
 
+        if isinstance(corpus, list):
+            logger.warning(
+                "taking plain list as a corpus for LSI;"
```
menshikh-iv: Please make a few fixes:
- Use capital letters/dots as a start/end of the sentence.
- Be careful that your words do not stick together, for example
```python
print(
"hello"
"world"
)
```
output is `helloworld`

gensim/models/ldamodel.py
rasto2211: I added space.

gensim/models/lsimodel.py
rasto2211: I also think that the logging messages formatted as proper English sentences look better. I have just seen that the rest of the code in this file is always starting the logging message with lower case letter and ends with semicolon and I thought that it's the "gensim code style".

Fixed the problem with words sticking together.

gensim/models/ldamodel.py
```diff
@@ -579,6 +579,15 @@ def update(self, corpus, chunksize=None, decay=None, offset=None,
...
+                "Taking plain list as a corpus for LDA.\n"
+                "We strongly recommend you to use corpus streaming instead of lists. "
+                "LDA is processing the corpus in batches and "
+                "it only loads one batch of documents into memory at once.\n"
```
piskvorky: -1 on newlines in log lines. Makes the log harder to process and parse (missing timestamps, broken format).

If you must output multiple lines, output them individually.

rasto2211: Ok, I removed the newlines from log messages and added spaces instead of them.

menshikh-iv:
I think that this change will only annoy our users, we discussed this with @piskvorky and he agrees with me. For this reason, I close this PR, sorry @rasto2211.

piskvorky:
Let me add that checking errors and sanity checks are still very desirable (incorrect input, potential issues etc).

PR #1405, opened 2017-06-09
[WIP] Added sklearn wrapper for LDASeq model

chinmayapancholi13:
This PR adds a scikit-learn wrapper for Gensim's LDASeq model.


gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
```diff
@@ -0,0 +1,86 @@
...
+    Base LdaSeq module
+    """
+
+    def __init__(self, time_slice=None, id2word=None, alphas=0.01, num_topics=10,
+                initialize='gensim', sstats=None, lda_model=None, obs_variance=0.5, chain_variance=0.005, passes=10,
+                random_state=None, lda_inference_max_iter=25, em_min_iter=6, em_max_iter=20, chunksize=100):
+        """
+        Sklearn wrapper for LdaSeq model. Class derived from gensim.models.LdaSeqModel
+        """
+        self.corpus = None
```
menshikh-iv: Why you needed a field for a corpus?

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
```diff
@@ -0,0 +1,86 @@
...
+        self.model = None
```
menshikh-iv: Please do this field "private" (start with underscores)

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def fit(self, X, y=None):
+        """
+        Fit the model according to the given training data.
+        Calls gensim.models.LdaSeqModel:
+        >>> gensim.models.LdaSeqModel(corpus=None, time_slice=None, id2word=None, alphas=0.01, num_topics=10,
+                initialize='gensim', sstats=None,  lda_model=None, obs_variance=0.5, chain_variance=0.005, passes=10,
+                random_state=None, lda_inference_max_iter=25, em_min_iter=6, em_max_iter=20, chunksize=100)
+        """
+        self.corpus = X
```
menshikh-iv: Don't need to save X.

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        Set all parameters.
+        """
+        super(SklearnWrapperLDASeqModel, self).set_params(**parameters)
+
+    def fit(self, X, y=None):
+        """
+        Fit the model according to the given training data.
+        Calls gensim.models.LdaSeqModel:
+        >>> gensim.models.LdaSeqModel(corpus=None, time_slice=None, id2word=None, alphas=0.01, num_topics=10,
```
menshikh-iv: Please remove this block `>>> ... `, this example does not help for a new user.

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
chinmayapancholi13: The user might be interested to know about the `corpus` used for training the model (using the `get_params` function).

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
chinmayapancholi13: Should we remove this `>>> ....` statement in all the model wrappers? This line basically tells us how the associated Gensim model is actually called.

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
menshikh-iv: @chinmayapancholi13 No, sklearn does not store X, so we should not

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
menshikh-iv: You just need to specify the class that is used (you have already done above) and write where user can read documentation.

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
chinmayapancholi13: @menshikh-iv Yes, that is true for sklearn. Removing `corpus` attribute from all the wrappers then.

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
```diff
@@ -0,0 +1,95 @@
...
+        Calls gensim.models.LdaSeqModel
+        """
+        self.gensim_model = models.LdaSeqModel(corpus=X, time_slice=self.time_slice, id2word=self.id2word,
+            alphas=self.alphas, num_topics=self.num_topics, initialize=self.initialize, sstats=self.sstats,
+            lda_model=self.lda_model, obs_variance=self.obs_variance, chain_variance=self.chain_variance,
+            passes=self.passes, random_state=self.random_state, lda_inference_max_iter=self.lda_inference_max_iter,
+            em_min_iter=self.em_min_iter, em_max_iter=self.em_max_iter, chunksize=self.chunksize)
+        return self
+
+    def transform(self, docs):
```
menshikh-iv: Chek case, when you create instance and call transform immediately (without fit), you need to raise exception like sklearn

menshikh-iv: Also, please add an example of `docs` param in docstring.

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
```diff
@@ -0,0 +1,95 @@
...
+            transformed_author = self.gensim_model[v]
+            # Everything should be equal in length
+            if len(transformed_author) != self.num_topics:
+                transformed_author.extend([1e-12] * (self.num_topics - len(transformed_author)))
+            X[k] = transformed_author
+
+        return np.reshape(np.array(X), (len(docs), self.num_topics))
+
+    def partial_fit(self, X):
+        raise NotImplementedError("'partial_fit' has not been implemented for the LDA Seq model")
```
menshikh-iv: `LDA Seq model` -> `SklLdaSeqModel`

gensim/test/test_sklearn_integration.py
```diff
@@ -192,5 +231,39 @@ def testSetGetParams(self):
...
+        self.assertEqual(model_params["num_topics"], 3)
+
+        # updating multiple params
+        param_dict = {"passes": 20, "chunksize": 200}
+        self.model.set_params(**param_dict)
+        model_params = self.model.get_params()
+        for key in param_dict.keys():
+            self.assertEqual(model_params[key], param_dict[key])
+
+
```
menshikh-iv: Add persistence test with pickle

menshikh-iv: + Add test with pipeline

gensim/sklearn_integration/sklearn_wrapper_gensim_ldaseqmodel.py
chinmayapancholi13: @menshikh-iv For checking if the model has been fitted, would it be a good idea to check if `self.gensim_model` is `None` or not? This approach would clearly give an error when `fit` hasn't been called before calling `transform` but this also allows the user to set the value of `self.gensim_model` through `set_params` function (or even as `wrapper.gensim_model=...`) and then call `transform` function, which makes sense for us to allow.

menshikh-iv: I completely forgot about `set_param`, so, I think if you disable `gensim_model` in set_param, you can check `model is None` (it does not cover all cases, but covers the most obvious)

chinmayapancholi13: Could you elaborate the meaning of "disabling" `gensim_model` param from the function `set_params`? 
Actually, `gensim_model` is a public attribute of the model so it can be set like `ldaseq_wrapper.gensim_model = some_model`, which is almost the same as using `set_params` function to set this value. So, checking whether `self.gensim_model` is `None` should be enough, right?
This would be like : 

```
    def transform(self, docs):
        """
        Return the topic proportions for the documents passed.
        """
        if self.gensim_model is None:
            raise NotFittedError("This model has not been fitted yet. Call 'fit' with appropriate arguments before using this method.")

        # The input as array of array
        check = lambda x: [x] if isinstance(x[0], tuple) else x
        ..........................................................................
        ..........................................................................
        ..........................................................................
        ..........................................................................

```

menshikh-iv: Ok, as a temporary option.

gensim/test/test_sklearn_integration.py
```diff
@@ -192,5 +232,70 @@ def testSetGetParams(self):
...
+        id2word = Dictionary(map(lambda x: x.split(), test_data))
+        corpus = [id2word.doc2bow(i.split()) for i in test_data]
+        model = SklLdaSeqModel(id2word=id2word, num_topics=2, time_slice=[1, 1, 1], initialize='gensim')
+        clf = linear_model.LogisticRegression(penalty='l2', C=0.1)
+        text_ldaseq = Pipeline((('features', model,), ('classifier', clf)))
+        text_ldaseq.fit(corpus, test_target)
+        score = text_ldaseq.score(corpus, test_target)
+        self.assertGreater(score, 0.50)
+
+    def testPersistence(self):
```
menshikh-iv: It's sanity check only.
For persistence, you need to compare `current` and `loaded` models. For this purpose, you need to compare `current` and `loaded` inner matrices OR get corpus, transform it with both variant and compare results

gensim/test/test_sklearn_integration.py
```diff
@@ -192,5 +232,70 @@ def testSetGetParams(self):
...
+        test_data = data.data[0:2]
+        test_target = data.target[0:2]
+        id2word = Dictionary(map(lambda x: x.split(), test_data))
+        corpus = [id2word.doc2bow(i.split()) for i in test_data]
+        model = SklLdaSeqModel(id2word=id2word, num_topics=2, time_slice=[1, 1, 1], initialize='gensim')
+        clf = linear_model.LogisticRegression(penalty='l2', C=0.1)
+        text_ldaseq = Pipeline((('features', model,), ('classifier', clf)))
+        text_ldaseq.fit(corpus, test_target)
+        score = text_ldaseq.score(corpus, test_target)
+        self.assertGreater(score, 0.50)
```
menshikh-iv: It's will be correct every time? No needed to fix seeds for reproducibility?

gensim/test/test_sklearn_integration.py
chinmayapancholi13: Thanks. I have now added code for comparing the vectors transformed from original and loaded models, in addition to this sanity check. :)

gensim/test/test_sklearn_integration.py
chinmayapancholi13: We now have a fixed seed which is set before the test `testPipeline` to ensure that we get similar values.

PR #1407, opened 2017-06-10
Use relative paths to notebooks in the same folder

vochicong:
Relative paths to notebooks in the same folder (same repository) enable me to follow the links in notebooks when doing experiments in my Jupyter environment, as well as when navigating notebooks in my fork.

I've also fixed the dead link to the Word2Vec and FastText comparison notebook.


menshikh-iv:
Thank you @vochicong, I think it's can be helpful, what do you think about this change @piskvorky?

piskvorky:
Looks good to me! 👍 

As long as the links continue to work when viewed on github (which they seem to do).

menshikh-iv:
Thank you @vochicong  :+1:

Issue #308, opened 2015-03-21
WikiCorpus: Support random sample


adammenges:
Related to #307. 

`random.sample(wiki, n)` will get killed on a reasonably spec'd machine, regardless of the size of `n`, it'd be nice to get a decently sized random sample for creating models quickly.

piskvorky:
Sounds like useful functionality.

There actually exist algorithms that take provably random samples even from a stream (non-repeatable generator).

I actually thought Python's `random.sample` implements such an algo -- why do you think it wouldn't work? Did you try?

vlejd:
`random.sample(population, k)` is (very) silently expecting a list. It calls `len(population)` and `population[i]` so it does not work on streams. To be honest, the implementation of `random.sample` is quiet bad. I will implement the algorithm that does work on streams. I will use reservoir sampling.

menshikh-iv:
Resolved in #1408

gojomo:
I see I'm late but a couple thoughts: 

Finding the length can require an initial extra full-scan of the texts – allowing a knowledgeable caller to optionally pre-specify a length could skip this potentially costly operation. 

Using the global `random` means any one sampling isn't easily reproducible from run-to-run – adding an optional random seed parameter would allow samples to be reproduced.

vlejd:
I believe that global random actually means better reproducibility! If you manually set the random.seed() in your script, it will set the seed for the library as well. I have tested it only on a toy example so maybe I am horribly wrong, but it should work.

Regarding the length. We should at least mention it in documentation. There are two options. Ask user, to manually set the self.length variable or to add a parameter first_x to the random sample function that would be interpreted as stream length (but not saved to self.length). Then the random sample would be only from first_x elements. However I am not sure what should happend if the first_x would be greater than the real length of stream. 
I am not sure how strict gensim wants to be about this type of performance.

gojomo:
If people try pre-seeding the global random to try to achieve reproducibility in this function, it will have side-effects on other consumers of random numbers, and potentially suffer interference from any global-random-consumers in other threads, or in the (interleaved) processing of each `next()`-offered text. Using a local random instance, explicitly seeded (if a seed provided) ensures no side-effects or interleaved interference. 

[...]

PR #1408, opened 2017-06-10
308 random sample on stream

vlejd:
#308

### DIFF
--- a/gensim/test/test_wikicorpus.py
+++ b/gensim/test/test_wikicorpus.py
@@ -63,12 +63,6 @@ class TestWikiCorpus(unittest.TestCase):
         l = wc.get_texts()
         self.assertTrue(u'папа' in next(l))
 
-    def test_sample_text(self):
-        wc = WikiCorpus(datapath(FILENAME_U), processes=1)
-        print(len(wc))
-        for x in wc.sample_texts(1):
-            print(x)
-
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


gensim/test/test_wikicorpus.py
```diff
@@ -62,6 +63,13 @@ def test_unicode_element(self):
         l = wc.get_texts()
         self.assertTrue(u'папа' in next(l))
 
+    def test_sample_text(self):
+        wc = WikiCorpus(datapath(FILENAME_U), processes=1)
+        print(len(wc))
```
menshikh-iv: Why do we need this line?

gensim/test/test_wikicorpus.py
```diff
@@ -62,6 +63,13 @@ def test_unicode_element(self):
...
+        for x in wc.sample_texts(1):
+            print(x)
```
menshikh-iv: No needed to print result in tests. You can use `pass` instead.

gensim/test/test_wikicorpus.py
vlejd: This whole function should not be here. I removed it.

gensim/test/test_wikicorpus.py
vlejd: This whole function should not be here. I remove it.

menshikh-iv:
Thank you @vlejd, сongratulations on your first PR :1st_place_medal:

PR #1410, opened 2017-06-12
Fix wordrank tests

parulsethi:
PR https://github.com/RaRe-Technologies/gensim/pull/1378 changed default `cleanup_files` to False. It results in wordrank test failures, as `output_dir` already exists when training is done again. 

This sets `cleanup_files` to True for training in wordrank tests.


menshikh-iv:
Thank you @parulsethi

PR #1411, opened 2017-06-12
Add napolean sphinx extension to allow parsing docstrings in various formats

rasto2211:
See the discussion on mailinglist:
<https://groups.google.com/forum/#!topic/gensim/M0N4mBgi9TY>

I also updated wiki:
<https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#documentation>

If you want to see what has changed then take a look at the documentation of `gensim.models.atmodel.AuthorTopicModel.update` . The docstring is written in a Google style but not really properly that's the reason why it looks a bit weird. Gensim is full of docstrings in various custom made docstring formats.

### DIFF
--- a/docs/src/Makefile
+++ b/docs/src/Makefile
@@ -1,5 +1,8 @@
 # Makefile for Sphinx documentation
 #
+# In order to compile this documentation you need to have sphinx
+# and sphinxcontrib-napoleon. You can install them using:
+# pip3 install sphinx sphinxcontrib-napoleon
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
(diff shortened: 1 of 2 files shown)


menshikh-iv:
Thank you @rasto2211. Did you hear about automated tools for docstring checking (in google style)? It's very important tool for us.

rasto2211:
To be honest, I don't know of any tool like that. There is a tool called [pydocstyle](https://github.com/pycqa/pydocstyle) which can check if docstrings are in numpy style but I did not manage to find anything like that for google style docstrings. I know that google is using pylint for checking the coding style and their config file is [here](https://github.com/vinitkumar/googlecl/blob/master/googlecl-pylint.rc), but I don't think that it checks docstrings formats. It seems that pylint only warns you if there is no docstring for a function/class.

menshikh-iv:
Thank you @rasto2211, congratulations on your first PR :1st_place_medal:  :+1:

PR #1413, opened 2017-06-14
[WIP] Phrases: make any2utf8  optional

prakhar2b:
**Updated benchmark (for text8)**

| Optimization | Python 2.7 | Python 3.6 | 
| ------------- |:-------------:| :------------:|
|    original  | ~ 36-38 sec | ~32-34 sec |
|recode_to_utf8=False| ~19-21 sec      | ~20-22 sec    | 

Speed improvement in Phrases module.
[Phrases optimization benchmark](https://github.com/prakhar2b/gensim/blob/any2utf8/docs/notebooks/phrases-optimization-benchmark.ipynb) (**for text8**)

------------------------------------------------------------
**Note--** Leaving it for the context of prior discussions

| Optimization       | Python 2.7     | Python 3.6 | PR |
| ------------- |:-------------:| :------------:| :---------------:|
|    original  | ~ 36-38 sec | ~32-35 sec |
| cython (static typing)      | ~30-32 sec    |
| any2utf8 (without cython)| ~20-22 sec     | ~23-26 sec |  This PR  |
| cython (with any2utf8)| ~15-18 sec     |  ~19-21 sec |  #1385  |


piskvorky:
What is the memory impact of this change? The conversion was there for a reason, IIRC.

prakhar2b:
@piskvorky yes, we had a [discussion here](https://github.com/RaRe-Technologies/gensim/pull/1385#discussion_r121989037) regarding this. This PR will be updated accordingly asap.

gensim/models/phrases.py
```diff
@@ -169,7 +169,9 @@ def learn_vocab(sentences, max_vocab_size, delimiter=b'_', progress_per=10000):
             if sentence_no % progress_per == 0:
                 logger.info("PROGRESS: at sentence #%i, processed %i words and %i word types" %
                             (sentence_no, total_words, len(vocab)))
-            sentence = [utils.any2utf8(w) for w in sentence]
+
+            sentence = [w for w in (utils.any2utf8(u'_'.join(sentence)).split('_'))]
```
jayantj: A few issues here - 
1. You are trying to split a bytestring (the result of the `any2utf8` call) by `'_'` - this will not work in python3+ because literal strings are unicode by default. You've faced similar problems previously, so I think it would be helpful to understand character encodings at a conceptual level, and the differences between string handling in python2 and 3.

2. Simply `sentences = utils.any2utf8(u'_'.join(sentence)).split('_'))` would be enough - no need for the extra `[w for w in ...]`

3. We're not accounting for the possibility that a word in the sentence contains `'_'` here - it would be wrong to make implicit assumptions like these about user input, unless there was an explicit constraint in the API. Escaping could be an option - although I'm not sure it is feasible, performance-wise.

gensim/models/phrases.py
```diff
@@ -169,8 +169,8 @@ def learn_vocab(sentences, max_vocab_size, delimiter=b'_', progress_per=10000):
             if sentence_no % progress_per == 0:
                 logger.info("PROGRESS: at sentence #%i, processed %i words and %i word types" %
                             (sentence_no, total_words, len(vocab)))
-            # sentence = [utils.any2utf8(w) for w in sentence]
-            # Unicode tokens in dictionary (not utf8)
+
+            sentence = [w for w in (utils.any2utf8('_'.join(w for w in sentence)).split('_'))]
```
piskvorky: This should all be in the optimized path (C / Cython), so there's no point wasting time playing with Python calls.

The utf8 conversion overhead will be ~0, once this is optimized properly.

You can check with `cython -a` to see how much slow (Python) code is still left on the critical path. There should be none.

piskvorky:
This whole style of optimization is not desirable.

Optimize the code properly, by writing in C/Cython, not by shuffling Python calls around, joining characters. That's not the way to gain a significant speed boost.

tmylk:
@piskvorky Some notes on the wider context of this work. 

Unfortunately, re-writing code in C is outside of the scope of the June evaluation milestone in Prakhar's [GSOC proposal](https://github.com/numfocus/gsoc/blob/master/2017/proposals/prakhar_gsoc_17.md) submitted in March. Another part of this proposal is selecting a multi-thread/multi-process architecture for Phrases - he is running experiments for it in the joblib PR.

Once the proper benchmarks for the `any2utf8` and Cython optimizations are submitted, these will be an improvement for the existing Phrases code. IMHO any improvements in the code are worth merging (of course unless they complicate too much which these changes don't). 

These minor improvements to Phrases have been a good GSOC learning experience for Prakhar in preparation for FastText performance optimisation which is the main focus of his GSOC project.

jayantj:
I agree, even if in the future we follow a different approach, I think the current changes are worthwhile, as they do improve the times for Phrases significantly (of course, we need clearer benchmarks, but from initial results, it certainly seems so).

gensim/models/phrases.py
```diff
@@ -166,10 +170,17 @@ def learn_vocab(sentences, max_vocab_size, delimiter=b'_', progress_per=10000):
         vocab = defaultdict(int)
         min_reduce = 1
         for sentence_no, sentence in enumerate(sentences):
+            s = sentence
             if sentence_no % progress_per == 0:
                 logger.info("PROGRESS: at sentence #%i, processed %i words and %i word types" %
                             (sentence_no, total_words, len(vocab)))
-            sentence = [utils.any2utf8(w) for w in sentence]
+            if isinstance(sentence[0], bytes):
```
piskvorky: What happens if `sentence` is empty?

piskvorky:
I disagree. This may be a good exercise for @prakhar2b , in preparation for fastText (like @tmylk says), but these utf8 changes obfuscate the code and are not the type of changes that the `Phrases` module needs.

Curious to see the benchmarks :)

gensim/models/phrases.py
```diff
@@ -166,10 +170,17 @@ def learn_vocab(sentences, max_vocab_size, delimiter=b'_', progress_per=10000):
...
+                sentence = [w for w in (utils.any2utf8(b';'.join(sentence)).split(b';'))]
```
piskvorky: What happens if the sentence tokens contain `;`?

prakhar2b:
@piskvorky for `recode_to_utf8=False`, do we expect this param to be used only when input is already `bytestring` (as you [mentioned here](https://github.com/RaRe-Technologies/gensim/pull/1454#issue-239064961)), or should user be able to do this even with raw `unicode` corpus. This PR still needs a minor change regarding this.

piskvorky:
I'd say allow unicode too. Or indeed anything that can be joined using the provided `delimiter`.

We should just make sure the delimiter works: unicode for unicode inputs, bytestring for bytestrings... I see that as one potential gotcha: user supplies unicode, says no recode, but code fails on `b'_'.join(unicode)`. How do you think we should handle it?

prakhar2b:
@piskvorky I think something like this should work (for `recode_to_utf8=False`) - 
```python
if not isinstance(sentence[0], bytes):
    delimiter = utils.to_unicode(delimiter)
```
I also think we should rename this param to `tokenize_to_utf8` (recode suggests that input is already bytestrimg, IMO)

gensim/models/phrases.py
```diff
@@ -105,7 +105,7 @@ class Phrases(interfaces.TransformationABC):
     and `phrases[corpus]` syntax.
 
     """
-    def __init__(self, sentences=None, min_count=5, threshold=10.0,
+    def __init__(self, sentences=None, recode_to_utf8=True, min_count=5, threshold=10.0,
                  max_vocab_size=40000000, delimiter=b'_', progress_per=10000):
         """
```
tmylk: Please add new param to the docstring

tmylk:
it is a strange result that  `any2utf8 (without cython)` `23-26 seconds` is slower than in 2.7. I expect the batch conversion to be faster than in 2.7

gensim/test/test_phrases.py
```diff
@@ -184,6 +188,15 @@ def setUp(self):
...
 
+    def testRecodeToUtf8False(self):
+        """Test that Phraser works as expected when `recode_to_utf8 = False` """
+        expected = ['survey', 'user', 'computer', 'system', 'response_time']
+
+        bigram_recode_false = Phrases(sentences, recode_to_utf8=False, min_count=1, threshold=1)
+
+        bigram_phraser = Phraser(bigram_recode_false)
+        self.assertEqual(bigram_phraser[sentences[1]], expected)
+
```
tmylk: Please add tests for input in bytes and utf-8 and errors for a mixture of the two

gensim/models/phrases.py
```diff
@@ -133,19 +133,30 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
         `delimiter` is the glue character used to join collocation tokens, and
         should be a byte string (e.g. b'_').
 
+        `recode_to_utf8` is an optional parameter to tokenize the input words into
```
piskvorky: The docstring is not correct -- there is no tokenization taking place.

gensim/models/phrases.py
```diff
@@ -133,19 +133,30 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
...
+        utf8 encoding. For faster Phrases, you may use `recode_to_utf8=False` if you have
+        larger RAM for this memory-time tradeoff.
```
piskvorky: This is incorrect. If the sentences are already bytestrings, there is no penalty to `recode_to_utf8=False` at all. No "larger RAM" needed.

gensim/models/phrases.py
```diff
@@ -133,19 +133,30 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
...
+
         """
         if min_count <= 0:
-            raise ValueError("min_count should be at least 1")
+            min_count = 1
```
piskvorky: What are these two conditions for? I see no reason for them, so I'd suggest removing these (`min_count`, `threshold`).

gensim/models/phrases.py
```diff
@@ -133,19 +133,30 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
...
+        self.recode_to_utf8 = recode_to_utf8
         self.min_count = min_count
         self.threshold = threshold
         self.max_vocab_size = max_vocab_size
         self.vocab = defaultdict(int)  # mapping between utf8 token => its count
         self.min_reduce = 1  # ignore any tokens with count smaller than this
         self.delimiter = delimiter
+
+        if not recode_to_utf8:
+            if not isinstance(next(iter(next(iter(sentences)))), bytes):
```
piskvorky: What is `sentences` is a non-repeatable stream? You'd silently throw away the first items.

prakhar2b: @piskvorky I tested for list for sentences and for text8 generators, it was working fine. What do you suggest, though ?

piskvorky: How did you test the generator? That does not seem possible.

prakhar2b: bigram = Phrases(Text8Corpus('/home/prakhar/text8'), recode_to_utf8 = False)

piskvorky: That's not a generator.

prakhar2b: @piskvorky ok, my mistake ! I need to get first word of any input corpus to check the encoding, how do you suggest we do that?

piskvorky: Have a look at `utils.is_corpus` for inspiration. The idea is to "peek" and then put the item back if necessary.

menshikh-iv: Short example @prakhar2b 

```
import itertools as it

corpus = ((w for w in ["a" * _, "b" * _, "c" * _]) for _ in range(1, 5))
sentence = list(next(iter(corpus)))
curr_word = sentence[0] # check encoding here

corpus = it.chain([sentence], corpus)
```

prakhar2b: @menshikh-iv Thanks. I think it would be better to add a test for non-repeatable corpus too, wdyt ? As of now, all the tests are passing already. cc @piskvorky

gensim/models/phrases.py
```diff
@@ -133,19 +133,32 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
         `delimiter` is the glue character used to join collocation tokens, and
         should be a byte string (e.g. b'_').
 
+        `recode_to_utf8` is an optional parameter (default True) for any2utf8 conversion of input sentences
```
piskvorky: How about

*By default, the input sentences will be internally encoded to UTF-8. Set `recode_to_utf8=False` to skip this recoding step. Then the Phrases object may use more memory (unless your sentences are already bytestrings), but will be much faster to train (~2x faster).*

gensim/models/phrases.py
```diff
@@ -133,19 +133,32 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
...
+        into bytestrings. You may use `recode_to_utf8=False` for ~1.8-2x speed up.
         """
         if min_count <= 0:
-            raise ValueError("min_count should be at least 1")
+            min_count = 1
```
piskvorky: Remove these two checks.

gensim/models/phrases.py
```diff
@@ -133,19 +133,32 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
...
             raise ValueError("threshold should be positive")
 
+        self.recode_to_utf8 = recode_to_utf8
         self.min_count = min_count
         self.threshold = threshold
         self.max_vocab_size = max_vocab_size
         self.vocab = defaultdict(int)  # mapping between utf8 token => its count
         self.min_reduce = 1  # ignore any tokens with count smaller than this
         self.delimiter = delimiter
+        self.is_bytes = True  # for storing encoding type in vocab for supporting both unicode and bytestring input
```
piskvorky: Comment hard to understand, and I think it's because the logic is not clear. What is it that is bytes in `is_bytes`?

Isn't it better to simply create a flag of whether the input sequences are `bytes` or not?

Then the comment becomes `# do the input sentences consist of bytestrings?` which is clear.

Also, this check seems to belong to `learn_vocab`, not here.

prakhar2b: Adding this comment (and I think it's rightly in init) -
```python
self.is_bytes = True  # do the input sentences consist of bytestrings?

# With default (recode_to_utf8=True) we encode input sentences to utf8 bytestrings, but
# with recode_to_utf8=False, we retain encoding, so need to store this encoding
# information to later convert token inputs accordingly (in __getitem__ and export_phrases)
```

piskvorky: If it's rightly in init, then what happens if `sentences=None` and user is calling `learn_vocab` manually later?

prakhar2b: @piskvorky  updated the PR

gensim/models/phrases.py
```diff
@@ -133,19 +133,24 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
         `delimiter` is the glue character used to join collocation tokens, and
         should be a byte string (e.g. b'_').
 
+        `recode_to_utf8`- By default, the input sentences will be internally encoded to
+        UTF-8 bytestrings, to save memory and ensure valid UTF-8. Set recode_to_utf8=False
```
piskvorky: Best to render code blocks as literal text (put `recode_to_utf8=False` in backticks).

gensim/models/phrases.py
```diff
@@ -133,19 +133,24 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
...
+        to skip this recoding step in case you don't care about memory or if your sentences
+        are already bytestrings. This will result in much faster training (~2x faster)
```
piskvorky: Missing full stop at the end of sentence.

gensim/models/phrases.py
```diff
@@ -227,12 +243,17 @@ def export_phrases(self, sentences, out_delimiter=b' ', as_tuples=False):
...
             last_bigram = False
             vocab = self.vocab
             threshold = self.threshold
             delimiter = self.delimiter  # delimiter used for lookup
             min_count = self.min_count
+
+            if self.recode_to_utf8:
+                s = [utils.any2utf8(w) for w in sentence]
+            else:
+                s = [utils.any2utf8(w) for w in sentence] if self.is_input_bytes else list(sentence)
```
piskvorky: Looks like a bug -- recode is False but recoding happens.

gensim/models/phrases.py
```diff
@@ -279,12 +300,19 @@ def __getitem__(self, sentence):
...
         last_bigram = False
         vocab = self.vocab
         threshold = self.threshold
         delimiter = self.delimiter
         min_count = self.min_count
+
+        if self.recode_to_utf8:
+            s = [utils.any2utf8(w) for w in sentence]
+        else:
+            s = [utils.any2utf8(w) for w in sentence] if self.is_input_bytes else list(sentence)
```
piskvorky: Looks like a bug -- recode is False but recoding happens.

gensim/models/phrases.py
```diff
@@ -368,10 +397,17 @@ def __getitem__(self, sentence):
...
 
-        s, new_s = [utils.any2utf8(w) for w in sentence], []
         last_bigram = False
         phrasegrams = self.phrasegrams
         delimiter = self.delimiter
+
+        if self.recode_to_utf8:
+            s = [utils.any2utf8(w) for w in sentence]
+        else:
+            s = [utils.any2utf8(w) for w in sentence] if self.is_input_bytes else list(sentence)
```
piskvorky: Looks like a bug -- recode is False but recoding happens.

gensim/models/phrases.py
```diff
@@ -157,19 +162,30 @@ def __str__(self):
             self.__class__.__name__, len(self.vocab), self.min_count,
             self.threshold, self.max_vocab_size)
 
-    @staticmethod
-    def learn_vocab(sentences, max_vocab_size, delimiter=b'_', progress_per=10000):
+    def learn_vocab(self, sentences):
         """Collect unigram/bigram counts from the `sentences` iterable."""
+        if not self.recode_to_utf8 and sentences is not None:
+            sentence = list(next(iter(sentences)))
```
jayantj: If I understand correctly, this will raise an exception if sentences is either an empty list or an empty generator.

docs/notebooks/phrases-optimization-benchmark.ipynb
```diff
@@ -0,0 +1,373 @@
...
+     "output_type": "stream",
+     "text": [
+      "1 loop, best of 3: 33.2 s per loop\n"
+     ]
+    }
+   ],
+   "source": [
+    "# currently on develop --- original code\n",
+    "from gensim.models import Phrases\n",
+    "bigram = Phrases(Text8Corpus(text8_file))\n",
```
piskvorky: Better convert the streamed iterable to an in-memory list (using `list()`), so we don't have to iterate over the file on disk every time.

This will make the benchmark conclusions stronger (less noise and delays from other, unrelated parts of code).

gensim/models/phrases.py
jayantj: In that case, ideally a simple test should be catching this too.

gensim/models/phrases.py
prakhar2b: @piskvorky is it true that only `unicode string` is accepted as input token in getitem or export_phrases as [here in phrases.py](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/phrases.py#L257)

gensim/models/phrases.py
```diff
@@ -165,7 +165,10 @@ def __str__(self):
     def learn_vocab(self, sentences):
         """Collect unigram/bigram counts from the `sentences` iterable."""
         if not self.recode_to_utf8 and sentences is not None:
-            sentence = list(next(iter(sentences)))
+            try:
+                sentence = list(next(iter(sentences)))
+            except:
+                raise ValueError("Input can not be empty list or generator.")
```
piskvorky: -1: why should empty input suddenly become a special case?

prakhar2b: @piskvorky  this concern was raised by @jayantj in [review comment above](https://github.com/RaRe-Technologies/gensim/pull/1413#discussion_r125001735)

prakhar2b: @piskvorky this concern was raised here in the [review comment above](https://github.com/RaRe-Technologies/gensim/pull/1413#discussion_r125001735)

piskvorky: Yes, but that's a reason to fix the bug, not change the API.

prakhar2b: oh, I should have just logged a warning `Empty sentences provided as input` in except block and no need to raise the error.

piskvorky: Unless it's really a special case (and I don't think it is), there's no need to treat it in a special way.

No exception, no warning -- there's nothing special to an empty corpus with regard to Phrases, except your new check for its first element.

gensim/models/phrases.py
prakhar2b: @piskvorky this conversion was apparently meant to handle the mismatch between bytestrings training input and unicode token input . Is this an unprecedented use case, should we just raise a warning (like TypeError for differently encoded training and token input)

piskvorky: Sorry, I don't understand. If the user said "I don't want recoding", we shouldn't be recoding.

What is the motivation for this?

prakhar2b: yes, we are not doing `recoding` while training Phrases.

For example,  for `recode_to_utf8=False` and for `unicode input sentences (for training)`, we have `unicode` words in vocab. Now if we provide `bytestrings` tokens in getitem or export_phrases, this mismatch will be a problem [here in phrases.py](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/phrases.py#L289). (or vice-versa)

For, no recoding at all, I think user will have to use same encoded corpus for both training, and phrases retrieval (getitem / export_phrases)

piskvorky: Ah, you're talking about what to do if the user sends a different type of data to `learn_vocab` vs `__getitem__`?

That's an interesting question -- should we try some implicit conversions? Or prefer to fail explicitly?

I think I'd prefer to fail explicitly, because the combination of "user asked not to recode during training" and "we implicitly recode to utf8 anyway during inference" could lead to subtle errors. For example, what happens if the user sent latin2 bytestrings to train, but we silently recode to utf8 during inference?

prakhar2b: `Silent recode to utf8 during inference` is carried out for token input and not for `training input`. The words in `vocab` will remain intact, we only do the conversion for incoming token to match the encoding in vocab (in this PR).

piskvorky: Yes, I see what you mean.

Let's see what the devs think, but it's a subtle point, I can see arguments for both ways (convert, fail).

Other than that this PR looks good, nice job @prakhar2b 👍

prakhar2b: @jayantj @gojomo @menshikh-iv ref - [above comment](https://github.com/RaRe-Technologies/gensim/pull/1413#discussion_r125164269)

cc @piskvorky

piskvorky:
@prakhar2b can you fix the two issues I pointed out last week? Let's get this PR over with finally.

gensim/models/phrases.py
```diff
@@ -157,19 +162,34 @@ def __str__(self):
...
+    def learn_vocab(self, sentences):
         """Collect unigram/bigram counts from the `sentences` iterable."""
+        if not self.recode_to_utf8 and sentences is not None:
+            try:
+                sentence = list(next(iter(sentences)))
+                if not isinstance(sentence[0], bytes):
+                    self.delimiter = utils.to_unicode(self.delimiter)
+                    self.is_input_bytes = False
+                sentences = it.chain([sentence], sentences)
+            except:
```
jayantj: Using catch-all except blocks is generally a bad idea, since you could end up catching unexpected exceptions. So catch only the specific exception expected here (`StopIteration`?)

gensim/models/phrases.py
```diff
@@ -157,19 +162,34 @@ def __str__(self):
...
+                #  No need to raise any exception or log any warning, as it's not a special case.
```
jayantj: Debug message here would serve well too.

gensim/models/phrases.py
jayantj: I think we shouldn't be recoding implicitly in `__getitem__`. The only gain is in the case when `learn_vocab` receives bytestrings, `recode_to_utf8` is `False` and `__getitem__` receives unicode. I don't think that justifies the dangerous subtle errors that it could cause.

Also, if we're going to be implicitly handling different input formats for `learn_vocab` vs `__getitem__`, we're also going to have to take care of the delimiter [here](https://github.com/RaRe-Technologies/gensim/pull/1413/files#diff-cb673cf37d57cdbaaa92b62f019a380dR263)

prakhar2b: I think delimiter issue is sorted in learn_vocab [here](https://github.com/prakhar2b/gensim/blob/c0d17c40ad635d29ce06baa3626266ddd9df33a5/gensim/models/phrases.py#L171). 

Just for information, what could be those dangerous errors ? The idea is simply to do the conversion for incoming token in `__getitem__` to match the encoding in vocab.

jayantj: As @piskvorky mentioned, suppose we had `recode_to_utf8=False`, and learn vocab was called with latin2 bytestrings, and latin2 bytestrings are sent to `__getitem__`. We'd silently recode the latin2 bytestrings to utf8 in `__getitem__` and lookup would fail, even though it should succeed.

prakhar2b: ok, yes I understand. I think `fail explicitly` should be the correct choice then (for mismatch in training input and infer input for recode_to_utf8=False). What kind of error should we raise then for this mismatch ?

prakhar2b:
I have a question regarding behaviour of phrases with  `bytestrings` infer input.
Suppose we have - 

> b_sent = [b'survey', b'user', b'computer', b'system', b'response', b'time']
> sent = [u'survey', u'user', u'computer', u'system', u'response', u'time']

If we do `bigram[sent]`, we get expected `['survey', 'user', 'computer_system', 'response', 'time']`.
However, `bigram[b_sent]` returns `<gensim.interfaces.TransformedCorpus object at 0x7f73a7d466a0>`.

I added this test (for bytestrings infer input) for `recode_to_utf8=False` but the behaviour is same for `recode_to_utf8=True` as well. I find it difficult to understand this behaviour. Therefore, before making any further changes, it would be better to know the POV of person who implemented it, if it is an intended behaviour or a bug ?

cc @gojomo @piskvorky @jayantj

piskvorky:
Not sure how it comes about, but looks like a bug to me @prakhar2b .

menshikh-iv:
What's status @prakhar2b?

piskvorky:
@menshikh-iv should I reopen my #1454 , since this was closed? I'm confused by the issue flow here.

menshikh-iv:
@piskvorky I think no, now Filip works with [#1446](https://github.com/RaRe-Technologies/gensim/pull/1446) as I understand

piskvorky:
@menshikh-iv #1446 is completely orthogonal to this. You can convert to utf8 with/without a memory-bounded counter.

PR #1416, opened 2017-06-14
Word2vec coherence

### LINKED ISSUE #1380: Add a new indirect confirmation measure based on word2vec similarity
#### Description
As introduced in [1], topic coherence can be computed using word2vec similarity of terms. This is roughly compatible with the coherence evaluation framework posited by [2] and currently implemented in gensim. The probability estimation phase will involve training a word2vec model on the given corpus, and the confirmation measure will use the average cosine similarity between context vectors from the word2vec model. [1] did not analyze which segmentation strategy is optimal, but a reasonable guess is to use the same method as "c_v" coherence, since that method also uses context vectors in a semantic space.

The implementation should include the ability to pass in a pre-trained word2vec model or path to the same.

I'm envisioning an API something like the following: 
```
cm = CoherenceModel(corpus=corpus, coherence="c_w2v", embeddings="./path-to-model.bin.gz", embeddings_kwargs=dict(binary=True))
# OR
model = gensim.models.KeyedVectors.load_word2vec_format('./GoogleNews-vectors-negative300.bin.gz', binary=True)
cm = CoherenceModel(corpus=corpus, coherence="c_w2v", embeddings=model)
cm.get_coherence()
```

#### Versions
This should be supported under all versions.

#### References
[1] D. O’Callaghan, D. Greene, J. Carthy, and P. Cunningham, “An analysis of the coherence of descriptors in topic modeling,” Expert Systems with Applications, vol. 42, no. 13, pp. 5645–5657, Aug. 2015.
[2] M. Röder, A. Both, and A. Hinneburg, “Exploring the Space of Topic Coherence Measures,” 2015, pp. 399–408.

macks22:
Resolves #1380.

This PR adds a new coherence measure to `CoherenceModel`, called "c_w2v". This uses word2vec word vectors for computing similarity between terms to calculate coherence. This implementation adds a new accumulator in the `text_analysis` module that either uses pre-trained `KeyedVectors` or trains a `Word2Vec` model on the input corpus to derive `KeyedVectors`. A new keyword argument `keyed_vectors` is added to the `CoherenceModel` to pass in pre-trained vectors.

Rather than requiring parity between the corpus dictionary and the keyed vectors vocabulary, the coherence calculation ignores terms that are missing from the vocabulary (with a warning). This PR also adds a new `with_std` argument to the `get_coherence_per_topic` method to calculate the standard deviation between the various segments for each topic. This can serve as some indication of topics that have good overall coherence but a few notable outlier terms or subsets.

Finally, I noticed that the `CoherenceModel`'s method of getting topics from models was implemented in a switch-style manner and refactored it to use polymorphism. Specifically, I introduced a new `get_topics` method to all topic models that returns the topic-term distributions (except for LSI, which just returns the weights, since its not a real distribution).


gensim/models/wrappers/ldavowpalwabbit.py
```diff
@@ -235,6 +235,14 @@ def log_perplexity(self, chunk):
                  corpus_words)
         return bound
 
+    def get_topics(self):
```
menshikh-iv: The common get_topics method for all TM looks nice, can you add this method to base TM class?

gensim/models/lsimodel.py
```diff
@@ -470,6 +468,24 @@ def __getitem__(self, bow, scaled=False, chunksize=512):
             result = matutils.Dense2Corpus(topic_dist)
         return result
 
+    def get_topics(self):
+        """
```
menshikh-iv: Please use google-style docstring format

gensim/test/test_coherencemodel.py
```diff
@@ -138,6 +142,12 @@ def testCvMalletModel(self):
             return
         CoherenceModel(model=self.malletmodel, texts=self.texts, coherence='c_v')
 
+    def testCw2vMalletModel(self):
+        """Perform sanity check to see if c_w2v coherence works with LDA Mallet gensim wrapper"""
+        if not self.mallet_path:
```
menshikh-iv: It's not correct checking, because if you run test & have no `mallet_path`, the status of this test is `ok`, but should be `skipped`

menshikh-iv:
Nice work @macks22, sorry for late answer.
Please fix my review comments, resolve merge problem & update coherence notebook with new examples.

menshikh-iv:
ping @macks22, what's a status of this PR?

macks22:
@menshikh-iv -- I royally botched the rebase/merge for conflict resolution here, so I'll be reopening this in a new PR.

PR #1420, opened 2017-06-16
 index and index.d data cleanup in AnnoyIndexer (test_similarities.py)

prakhar2b:
cc @menshikh-iv


menshikh-iv:
Thank you @prakhar2b

PR #1422, opened 2017-06-16
Create local random generator for sample_text & add lenght

vlejd:
Changes regarding @gojomo comments at  #308 .  Add option to use local random and option to tell the function how long is the corpus (without running __len__  and going through the whoe corpus).

### DIFF
--- a/gensim/corpora/textcorpus.py
+++ b/gensim/corpora/textcorpus.py
@@ -100,24 +100,25 @@ class TextCorpus(interfaces.CorpusABC):
 
     def sample_texts(self, n, seed=None, length=None):
         """
-        Yields n random documents from the corpus without replacement.
+        Yield n random documents from the corpus without replacement.
 
-        Given the number of remaining documents in corpus, we need to choose n elements.
-        The probability for current element to be chosen is n/remaining.
+        Given the number of remaining documents in a corpus, we need to choose n elements.
+        The probability for the current element to be chosen is n/remaining.
         If we choose it, we just decreese the n and move to the next element.
-        Computing corpus length may be a costly operation so you can use optional paramter
-        length instead.
+        Computing the corpus length may be a costly operation so you can use the optional
+        parameter `length` instead.
 
         Args:
             n (int): number of documents we want to sample.
             seed (int|None): if specified, use it as a seed for local random generator.
-            length (int|None): if specified, use it as guess of corpus length.
+            length (int|None): if specified, use it as a guess of corpus length.
+                It must be positive and not greater than actual corpus length.
 
         Yeilds:
-            list[str]: document represented as list of tokens. See get_texts method.
+            list[str]: document represented as a list of tokens. See get_texts method.
 
         Raises:
-            ValueError: then n is invalid or length was set incorrectly.
+            ValueError: when n is invalid or length was set incorrectly.
         """
         random_generator = None
         if seed is None:
--- a/gensim/corpora/textcorpus.py
+++ b/gensim/corpora/textcorpus.py
@@ -144,8 +144,9 @@ class TextCorpus(interfaces.CorpusABC):
                 yield sample
 
         if n != 0:
-            # This means that length was set to be smaller than number of items in stream.
-            raise ValueError("length smaller than number of documents in stream")
+            # This means that length was set to be greater than number of items in corpus
+            # and we were not able to sample enough documents before the stream ended.
+            raise ValueError("length greater than number of documents in corpus")
 
     def __len__(self):
         if not hasattr(self, 'length'):
--- a/gensim/corpora/textcorpus.py
+++ b/gensim/corpora/textcorpus.py
@@ -98,27 +98,53 @@ class TextCorpus(interfaces.CorpusABC):
                 else:
                     yield utils.tokenize(line, lowercase=True)
 
-    def sample_texts(self, n):
+    def sample_texts(self, n, seed=None, length=None):
         """
-        Yield n random texts from the corpus without replacement.
+        Yields n random documents from the corpus without replacement.
 
-        Given the the number of remaingin elements in stream is remaining and we need
-        to choose n elements, the probability for current element to be chosen is n/remaining.
+        Given the number of remaining documents in corpus, we need to choose n elements.
+        The probability for current element to be chosen is n/remaining.
         If we choose it, we just decreese the n and move to the next element.
-        """
-        length = len(self)
-        if not n <= length:
-            raise ValueError("sample larger than population")
+        Computing corpus length may be a costly operation so you can use optional paramter
+        length instead.
 
+        Args:
+            n (int): number of documents we want to sample.
+            seed (int|None): if specified, use it as a seed for local random generator.
+            length (int|None): if specified, use it as guess of corpus length.
+
+        Yeilds:
+            list[str]: document represented as list of tokens. See get_texts method.
+
+        Raises:
+            ValueError: then n is invalid or length was set incorrectly.
+        """
+        random_generator = None
+        if seed is None:
+            random_generator = random
+        else:
+            random_generator = random.Random(seed)
+
+        if length is None:
+            length = len(self)
+
+        if not n <= length:
+            raise ValueError("n is larger than length of corpus.")
         if not 0 <= n:
-            raise ValueError("negative sample size")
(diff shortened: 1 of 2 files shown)


gensim/test/test_textcorpus.py
```diff
@@ -21,34 +21,48 @@
 class TestTextCorpus(unittest.TestCase):
     # TODO add tests for other methods
 
-    def test_sample_text(self):
-        class TestTextCorpus(TextCorpus):
-            def __init__(self):
-                self.data = [["document1"], ["document2"]]
+    class DumyTextCorpus(TextCorpus):
```
gojomo: spelling: 'Dummy'

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,54 @@ def get_texts(self):
...
+            remaining_in_corpus = length - i
+            chance = random_generator.randint(1, remaining_in_corpus)
             if chance <= n:
                 n -= 1
                 yield sample
 
+        if n != 0:
+            # This means that length was set to be smaller than nuber of items in stream.
+            raise ValueError("length smaller than number of documents in stream")
+
```
gojomo: Isn't this actually the opposite: that the generator has *more* elements than was declared? 

Even so, I doubt this case should be a thrown-exception that must be handled. The user got their `n` docs, within the `length` declared – which might be exactly what they wanted/needed. So perhaps just a logged warning – "documents in excess of declared 'length' not considered for sampling".

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,54 @@ def get_texts(self):
                 else:
                     yield utils.tokenize(line, lowercase=True)
 
-    def sample_texts(self, n):
+    def sample_texts(self, n, seed=None, length=None):
         """
-        Yield n random texts from the corpus without replacement.
+        Yields n random documents from the corpus without replacement.
```
piskvorky: "Yields" => "Yield" (imperative mode).

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,54 @@ def get_texts(self):
...
 
-        Given the the number of remaingin elements in stream is remaining and we need
-        to choose n elements, the probability for current element to be chosen is n/remaining.
+        Given the number of remaining documents in corpus, we need to choose n elements.
```
piskvorky: "a corpus".

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,54 @@ def get_texts(self):
...
+        The probability for current element to be chosen is n/remaining.
```
piskvorky: "the current element"

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,54 @@ def get_texts(self):
...
         If we choose it, we just decreese the n and move to the next element.
+        Computing corpus length may be a costly operation so you can use optional paramter
```
piskvorky: "the corpus length"
"the optional parameter \`length\` instead"

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,54 @@ def get_texts(self):
...
+        length instead.
+
+        Args:
+            n (int): number of documents we want to sample.
+            seed (int|None): if specified, use it as a seed for local random generator.
+            length (int|None): if specified, use it as guess of corpus length.
```
piskvorky: "a guess"

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,54 @@ def get_texts(self):
...
+
+        Yeilds:
+            list[str]: document represented as list of tokens. See get_texts method.
```
piskvorky: "a list"

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,54 @@ def get_texts(self):
...
+
+        Raises:
+            ValueError: then n is invalid or length was set incorrectly.
```
piskvorky: "then" => "when"?

Also, what does "incorrectly" mean? Above we say `length` is "a guess", so how can it be incorrect?

gensim/corpora/textcorpus.py
vlejd: ok

gensim/corpora/textcorpus.py
vlejd: ok

gensim/corpora/textcorpus.py
vlejd: ok

gensim/corpora/textcorpus.py
vlejd: ok

gensim/corpora/textcorpus.py
vlejd: ok

gensim/corpora/textcorpus.py
vlejd: I add some clarification to length parameter description.

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,55 @@ def get_texts(self):
...
+        Computing the corpus length may be a costly operation so you can use the optional
+        parameter `length` instead.
+
+        Args:
+            n (int): number of documents we want to sample.
+            seed (int|None): if specified, use it as a seed for local random generator.
+            length (int|None): if specified, use it as a guess of corpus length.
+                It must be positive and not greater than actual corpus length.
+
+        Yeilds:
```
rasto2211: *Yields. Spellchecker can be useful.

vlejd: Ok.
I was wandering that I should finally install it (F6 in sublime).

piskvorky: "Wondering". Definitely :)

gensim/corpora/textcorpus.py
vlejd: ok

gensim/test/test_textcorpus.py
vlejd: ok

gensim/corpora/textcorpus.py
vlejd: It should be _length was set to be **greater** than number of items in stream_  indeed.  

It throws an exception because user received less than `n` docs. Note that if `length` is greater than actual length of the corpus, it will not trigger the exception on its own. The exception triggers when we still need to sample some documents, but the stream run out.

The case when `length` was smaller than the actual corpus length does not throw anything right now (the problem was only in the documentation). Printing a warning is awkward, because we would have to actually go through all of the `length` elements which is unnecessary (we may sample enough elements earlier).

gensim/corpora/textcorpus.py
vlejd: I have no words.

menshikh-iv:
Thank you @vlejd

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,56 @@ def get_texts(self):
...
+            remaining_in_corpus = length - i
+            chance = random_generator.randint(1, remaining_in_corpus)
             if chance <= n:
                 n -= 1
                 yield sample
 
+        if n != 0:
+            # This means that length was set to be greater than number of items in corpus
+            # and we were not able to sample enough documents before the stream ended.
+            raise ValueError("length greater than number of documents in corpus")
```
piskvorky: Can you make the message more concrete? Include the actual lengths.

gensim/corpora/textcorpus.py
```diff
@@ -98,28 +98,56 @@ def get_texts(self):
...
+        if seed is None:
+            random_generator = random
+        else:
+            random_generator = random.Random(seed)
+
+        if length is None:
+            length = len(self)
 
+        if not n <= length:
+            raise ValueError("n is larger than length of corpus.")
```
piskvorky: Can you make the message more specific? Include the actual lengths.

vlejd: ok

gensim/corpora/textcorpus.py
vlejd: ok

Issue #1364, opened 2017-05-24
Better support for training word2vec models on a corpus consisting of multiple files


michaelwsherman:
models.word2vec.LineSentence offers a simple generator that takes either a file object or a path to a single object and reads each line as a sentence where the tokens are separated by whitespaces. This assumes that the corpus to be trained on is all in a single file, which may not be the case. And adding additional files to a word2vec model after its been trained is very complex, requiring multiple calls and manual management of learning rates.

Rather than force the user to write their own generator, to concatenate their files, or to manually manage learning rates across multiple files, it is desirable if word2vec could take a directory reference or a list of files directly, through LineSentence or a similar generator.

I have code for a generator that takes all files in a directory and makes a generator suitable for training a word2vec model (or any other model compatible with LineSentence). If desired, I can create a pull request with this generator. I can also (attempt to) change LineSentence to detect if it has been passed a file-like object, a file path, or a directory path (and also possibly a list of files) and then generate lists of tokens appropriately, and then submit the pull request.

Please let me know if this is desired. Thank you.

menshikh-iv:
Hi @michaelwsherman, I think this will be useful. Please, submit your PR

michaelwsherman:
Ok, give me a week or two, I've never submitted anything to open source before and I want to make sure I don't create extra work for others by messing something up. But it will come.

Right now its a separate method that just assumes all files in a directory are text files. Should I add some improvements (gzip support? merging it into the LineSentence method and adding input detection?) or just start with what I have?

menshikh-iv:
It's better to start with what you already have, we discuss about another features after your PR

gojomo:
When the work is in a single git branch, you can create the PR, get feedback, and continue to update the branch/PR with improvements – so always OK to start with something rough. (You won't mess anything up unless/until it's reviewed and accepted, and you can mark it as 'in progress'/'for review' to indicate it's not yet ready.)

You should use the `smart_open` package for file-opening - it will automatically detect gzip name-extensions and do the right thing. 

But it's ok to start with whatever you've got, and see what suggestions for extension/refactoring/etc come up once that's concretely viewable via github!

michaelwsherman:
Pull request at https://github.com/RaRe-Technologies/gensim/pull/1423 . Would love to hear any feedback--first time OSS contributor.

menshikh-iv:
Resolved in #1423

PR #1423, opened 2017-06-16
Add word2vec.PathLineSentences for reading a directory as a corpus (#1364)

michaelwsherman:
word2vec.LineSentencePath(path) will read all the files in a directory in the same fashion as word2vec.LineSentence reads a file. This provides an easy way to use a corpus of multiple files when training a word2vec model (or any model compatible with word2vec.LineSentence). 

Minimal exeception handling, but some logging.


gojomo:
Thanks for the contrib!

The failing automatic check looks like some style issues - you can click the 'Details' link or red 'X' to view the failure logs for more hints. 

More substantive comments inline.

gensim/models/word2vec.py
```diff
@@ -1521,6 +1521,54 @@ def __iter__(self):
                         i += self.max_sentence_length
 
 
+class LineSentencePath(object):
```
gojomo: Personally I'd consider the name `PathLineSentences` more descriptive, but other may have an even better name.

gensim/models/word2vec.py
```diff
@@ -1521,6 +1521,54 @@ def __iter__(self):
...
+
+        The files in the directory should be either text files, .bz2 files, or .gz files.
+
+        """
+        self.source = source
+        self.max_sentence_length = max_sentence_length
+        self.limit = limit
+
+        try:
+            self.source = os.path.join(source, '') # ensures os-specific slash is at end of path
```
gojomo: Thoughts on coverage of all related needs:

* perhaps this should accept a path to a single file, too, and still work in that case?
* by deferring the actual resolution of initialization parameters to the beginning of `__iter__()`, the object might be more robust for cases where files are arriving in the target directory between instantiation & 1st iteration. OTOH, that would also mean repeated iterations – as in the common Word2Vec/Doc2Vec multi-pass training, could find different files each time. No strong opinion yet on which approach is better – just pointing out the choice.

gensim/test/test_word2vec.py
```diff
@@ -763,6 +763,15 @@ def testLineSentenceWorksWithNormalFile(self):
                 sentences = word2vec.LineSentence(fin)
                 for words in sentences:
                     self.assertEqual(words, utils.to_unicode(orig.readline()).split())
+
+    def testLineSentencePath(self):
+        """Does LineSentencePath work with a path argument?"""
+        logging.debug(word2vec)
+        with utils.smart_open(datapath('lee_background.cor')) as orig:
+            sentences = word2vec.LineSentencePath(datapath('LineSentencePath'))
```
gojomo: Perhaps rather than creating a custom new split (and duplication) of the `lee_background.cor`, a new subdir of test-data `lee` could be added, with just the `lee.cor` and `lee_background.cor` files. The test would make sure this new class on the directory yields the same set of docs as the *two* other files read individually. And eventually, the duplication of these files in two places could be eliminated by making all the other tests/demos/tutorials just grab the lee* files from this new canonical place. (That is: aim for less duplication in the long run by re-using the same files everywhere.)

gensim/models/word2vec.py
michaelwsherman: Thank you for taking the time to comment. I'll get to these next week.

michaelwsherman: Change made, will be reflected in next pull request

gensim/test/test_word2vec.py
michaelwsherman: I thought about this, but I think making a lee directory may be worse than having a special PathLineSentences directory.

It seems there are quite a few lee* files in test_data already. I could move those two files you mentioned, but then we'd have a bunch of lee files still in the main test_data directory. Those files could be moved as well, except some of them aren't compatible with PathLineSentences (.bin, .vec) which meant that they couldn't be moved without breaking the test. Then we would have lee* files in two places, which could lead to confusion later on. There is also the issue that something that makes sense to add to this lee directory in the future could break the test.

Given that the PathLinesSentences class operates on a full directory (which makes it somewhat unique) I think it makes more sense to keep it's test dependent on a directory that isn't overloaded for use with other tests.

What I will do is use smaller files that aren't duplicated.

michaelwsherman: Change made, put in a new tiny test corpus, and changed the test to read the files and combine them manually and compare to PathLineSentences, rather than having a single file duplicating the contents of the PathLineSentences folder in test_data

gensim/models/word2vec.py
michaelwsherman: Change made to accept a single file and still work, including an additional test case.

I think it is better to resolve the initialization parameters in the `__init__()`. While there could be some use in not requiring the files to all be present when the object is initialized, I think that possibly changing the files processed every time a new iteration starts is likely to cause confusion. It seems more natural that the default behavior would be to get a list of files and not change them as long as the object is used. This would match the behavior of LineSentence--if you change the contents of the file between iterations, you'll get different results, but you can't change the reference to the file after the object has been created.

I would personally be caught off guard if the files changed between iterations. While this could be useful in some cases, I think it is a risky default behavior. Adding some capabilities to do this, however, may make sense. But I'd rather not do that unless a compelling use case is presented.

What I've done instead is log the list of files read when the object is created at the info level, so there's some sort of explicit record available of what the object is reading.

michaelwsherman:
I think I've addressed all the comments. Please re-review.

I'm having some problems with figuring out what travis-ci is angry about. I did find an inline comment missing a whitespace and fixed that. But I think the other comments are due to extra spaces in [i : i +...] line and that's elsewhere in the files. I also see errors with the new corpus file--not sure how to tell travis-ci to not treat those files as code. I'll review after the current style check is done and see if I can get to the bottom of things.

michaelwsherman:
Nevermind, looks like travis just checks my changes -- not code already in the repo. But I did fix another style issue elsewhere in the code. 

I think the style check should pass now.

michaelwsherman:
Cleaning up my noob git branching fail, no more dangling branches not actually getting used.

michaelwsherman:
@gojomo or other maintainers--Any other changes? Any questions? Please let me know. Thank you.

menshikh-iv:
Looks good for me, thank you @michaelwsherman, congratz with your first PR:1st_place_medal:

gensim/models/word2vec.py
```diff
@@ -1517,7 +1517,57 @@ def __iter__(self):
...
+
+        The files in the directory should be either text files, .bz2 files, or .gz files.
+
+        """
+        self.source = source
+        self.max_sentence_length = max_sentence_length
+        self.limit = limit
+
+        if os.path.isfile(self.source):
+            logging.warning('single file read, better to use models.word2vec.LineSentence')
```
piskvorky: Should be `logger` (each module has its own logger in gensim).

Applies here and elsewhere in this PR.

gensim/models/word2vec.py
```diff
@@ -1517,7 +1517,57 @@ def __iter__(self):
...
+            self.input_files = [self.source]  # force code compatibility with list of files
+        elif os.path.isdir(self.source):
+            self.source = os.path.join(self.source, '')  # ensures os-specific slash at end of path
+            logging.debug('reading directory ' + self.source)
+            self.input_files = os.listdir(self.source)
+            self.input_files = [self.source + file for file in self.input_files]  # make full paths
```
piskvorky: `file` is a reserved keyword in Python. Better use `filename` or something like that.

gensim/models/word2vec.py
```diff
@@ -1517,7 +1517,57 @@ def __iter__(self):
...
+            self.input_files.sort()  # makes sure it happens in filename order
+        else:  # not a file or a directory, then we can't do anything with it
+            raise ValueError('input is neither a file nor a path')
+
+        logging.info('files read into PathLineSentences:' + '\n'.join(self.input_files))
```
piskvorky: Better to pass the formatting arguments as arguments: `logger.info("%s", y)` instead of `logger.info("%s" % y)` .

Here and elsewhere.

gensim/models/word2vec.py
```diff
@@ -1517,7 +1517,57 @@ def __iter__(self):
...
+    """
+    Simple format: one sentence = one line; words already preprocessed and separated by whitespace.
+    Like LineSentence, but will process all files in a directory in alphabetical order by filename
+    """
+
+    def __init__(self, source, max_sentence_length=MAX_WORDS_IN_BATCH, limit=None):
+        """
+        `source` should be a path to a directory (as a string) where all files can be opened by the
+        LineSentence class. Each file will be read up to
+        `limit` lines (or no clipped if limit is None, the default).
```
piskvorky: `no` => `not`.

michaelwsherman:
Fixes from @piskvorky in PR #1573 .

PR #1424, opened 2017-06-16
Change docs build command

rasto2211:
Add gensim source code path to `${PYTHONPATH}` bash variable before running `sphinx-build`.
This allows Sphinx to use `.py` files from `../..` instead of the `.py` files from the virtual environment or
system-wide installed version of gensim.

Previously, you needed to run:
`python3 setup.py install`
whenever you wanted to build the docs from the current version of docstrings.
Now, you don't need to do it. Sphinx automatically extracts docstrings from the
gensim source code in `../..`. This makes debugging of docstrings formating in
sphinx documentation a lot easier. Just run `make html`. SymPy is also using
this trick with changing the `${PYTHONPATH}`.

### DIFF
--- a/docs/src/Makefile
+++ b/docs/src/Makefile
@@ -6,7 +6,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
+SPHINXBUILD   = PYTHONPATH=../..:$(PYTHONPATH) sphinx-build
 PAPER         =
 BUILDDIR      = _build
 


gojomo:
A relative-path in PYTHONPATH seems somewhat risky in terms of predictable-behavior, unless it's absolutely certain it will only take effect from a certain invocation working-directory. Shouldn't attempts to `make html` only happen when in a venv with the gensim-src-dir-of-interest already on the venv's paths?

rasto2211:
Well, the point is that running `python3 setup.py install` in gensim directory every time I want to build the documentation gets annoying. With this change I just need to `cd docs/src; make html`. I don't need to install gensim every time I want to build the documentation. 

I don't think that using a relative path is risky in this case. You are only going to run `make html` when you are in `gensim/docs/src` and the source code is in `gensim/gensim`. Sphinx-build looks for directory named `gensim`.

Btw. travis checks are failing because travis runs flake8 also on Makefiles which is weird.

gojomo:
Is it because `python3 setup.py install` from the gensim directory takes longer to complete than `cd docs/src; make html` from the gensim directory? 

I do now realize the PYTHONPATH change is low risk since that Makefile will only be active when already in that working-directory.

(Tangential thought: I'm no setup.py expert, but maybe an alternate target besides `install` can be added that just does the right minimal doc-build – even obviating the `cd docs/src`?)

Looks like the failing test is due to spurious Python-style-checking on the non-Python `Makefile`.

rasto2211:
Yes, gensim installation takes quite long and is not really necessary to build the documentation with this fix.

When you have only `sphinx-build` in the Makefile then it extracts docstrings from the installed version of gensim. When you have `PYTHONPATH=../..:$(PYTHONPATH) sphinx-build` then it extracts docstrings from the `.py` files in the `../..`.

I cannot come up with a scenario when you would want to build the documentation using python files in the installed version of gensim. When you are in the git folder of gensim you can switch to any commit in git and then run `cd docs/src; make html`.

As I said SymPy is also doing it this way and it seems like nobody is complaining about it.

Besides the advantage which I mentioned above, it also allows you to build the docs from git without creating a virtual environment.

rasto2211:
Python checks should not be ran on non-python files. Should I create a github issue on this problem?

rasto2211:
@menshikh-iv What do you think of this change?

menshikh-iv:
> Tangential thought: I'm no setup.py expert, but maybe an alternate target besides install can be added that just does the right minimal doc-build – even obviating the cd docs/src?

@gojomo As I know, It's possible, I like this idea:+1: 
@rasto2211 `make` works only in dir with `Makefile` (i.e. I must to `cd docs/src` before make), I'm correct? 
If we will add `.../...` to `PYTHONPATH`, install gensim to the environment and run Sphinx, where Spinx will be collect docs (from repo folder OR from installed gensim version)?

menshikh-iv:
ping @rasto2211

rasto2211:
I'm sorry Ivan but I'm quite busy these days. Maybe, I'll find some time next week to look at the `setup.py` script and find out if we can add `doc-build` flag.

If you run `make html` in the command line then it automatically takes `Makefile` from the current directory (`.`). If you want to run `make html` using Makefile in a different directory then you have to use a flag `-f` and run it this way: `make -f [filename] html`. Does this answer your question?

menshikh-iv:
@rasto2211 yes, but what happens with a relative path for this case (with your current edition)?

I think I should close this PR. I hope when you have a time, you will add a command to `setup.py` for doc-building.

PR #1432, opened 2017-06-19
Added word ngram parameter to fasttext

fsonntag:
Simply add the word ngram parameter to fasttext


menshikh-iv:
Thank you @fsonntag, I congratulate you on the first PR :1st_place_medal:

PR #1434, opened 2017-06-20
[MRG] Implement 'Translation Matrix' 

robotcator:
This PR adds translation matrix feature to gensim model.


gojomo:
Very excited to see this! Note that it may be useful for more than just sets of word-vectors – for example, if you were to Doc2Vec-train a batch of new documents, but include a bunch (100s? 1000s? TBD) of reference documents from a prior larger run, you could presumably translate the new doc-vecs back into the older/larger space using those reference docs to learn the back-mapping. 

The comments & documentation need a lot of review/improvement for clarity/completeness. (Variable and method names might benefit from some careful consideration, too.) When it's thought to be ready for merging – passing tests and feature complete – I'll be happy to give it a close top-to-bottom read.

robotcator:
@gojomo Thanks, happy to have you help. You mean the translation matrix not only restrict to translate the word, but can apply to the document translation? Am I right?

gojomo:
Not document 'translation' as in from one natural-language to another, but from alternate Doc2Vec runs on different sets of (same-language) documents - but featuring enough overlapping texts to learn the mapping.

robotcator:
You mean when I were to train on a batch of new documents, and get the new doc vector. And There are also many reference documents, we can use those to learn to back-mapping. But I don't quite understand the reference documents( in same language), can you explain in more detail? Thanks.

gojomo:
Let's say you train doc-vecs for 10 million documents. Then you get a batch of 500,000 more. 

If you train the old model on just the new 500K, the model is improving for those – but drifting away from modelling the older docs. (If some docs/doc-types are not re-presented, in interleaved fashion, the model will lose its 'tuning' for them.)

Alternatively, if you train a new model on just the new 500K, the doc-vectors from this new model will have no essential relationship with the older model. (There's no absolute vector positioning from any training session - the exact same document can wind up in very different places based on the inherent randomness of initialization and training. It's only the *relative* positions of doc-vectors from the same combined training that is meaningful.)

So, I'm proposing it could be effective to train the new 500K, with some small set of "reference point" documents from the older set mixed-in. (Let's say for now: 10,000 docs.) When the training of the all-new model is done, those 10K ref docs will have points in the "new" space that render them comparable to the 500K new docs. These points will quite probably be very *unlike* their points in the older space. But, using this `TranslationMatrix`, those 10K points can be used to learn a mapping back to the older (10M doc) space, and then transform all 500K new docs into the original, older space for comparison to any of the now 10.5M documents.

robotcator:
Thank for you explaining. I understand what you said now. The "reference documents" from the older set will be mixed into the new training set, those can be seen as pivot to learn the "TranslationMatrix". It is worth trying.

robotcator:
Hi, @gojomo recently, I have doing the experiment about produce new document vectors via the translation matrix which learned by the back-mapping method. If I train on new 500K document , with 10k of "reference point" documents taken from the 10 million docs. Then, by using TranslationMatrix, we can transform all 500K new docs into the original, older space. So we got 500k more document vector in the original space.  But Is there any need measure/evaluate the new docs' vector in older space (perform better)? Or we just use this method to produce more doc vectors rather than train the entire corpus(10.5 million document )again?

piskvorky:
Hi guys, what's the context here? Can you please add some description of the purpose of this PR? 

What are its use-cases, from gensim users' POV?

robotcator:
@piskvorky This PR is to implement the [translation matrix](https://github.com/RaRe-Technologies/gensim/wiki/GSOC-2017-project-ideas). If needed, I will add more description in this PR.

gojomo:
@robotcator - Yes, an experiment to evaluate the usefulness of such reference-projected doc-vectors would be good – especially in comparison to inferring-doc-vectors. (Maybe inference works better. Or maybe if there are a lot of new words in the incremental docs, the back-projection works better... it'd certainly be able to bootstrap word-vectors, in the original space, that inference alone would not.) But, I don't have a proposed experiment design ready.

robotcator:
Hi, I use the  IMDB sentiment dataset to train two models, one with 15k documents and the other with 25k documents, all other parameters are the same. The code is below(some are from the notebook)
```
import gensim
from gensim.models.doc2vec import TaggedDocument
from gensim.models import Doc2Vec
from collections import namedtuple
from gensim import utils

def read_sentimentDocs():
    SentimentDocument = namedtuple('SentimentDocument', 'words tags split sentiment')

    alldocs = []  # will hold all docs in original order
    with utils.smart_open('alldata-id.txt', encoding='utf-8') as alldata:
        for line_no, line in enumerate(alldata):
            tokens = gensim.utils.to_unicode(line).split()
            words = tokens[1:]
            tags = [line_no] # `tags = [tokens[0]]` would also work at extra memory cost
            split = ['train','test','extra','extra'][line_no//25000]  # 25k train, 25k test, 25k extra
            sentiment = [1.0, 0.0, 1.0, 0.0, None, None, None, None][line_no//12500] # [12.5K pos, 12.5K neg]*2 then unknown
            alldocs.append(SentimentDocument(words, tags, split, sentiment))

    train_docs = [doc for doc in alldocs if doc.split == 'train']
    test_docs = [doc for doc in alldocs if doc.split == 'test']
    doc_list = alldocs[:]  # for reshuffling per pass

    print('%d docs: %d train-sentiment, %d test-sentiment' % (len(doc_list), len(train_docs), len(test_docs)))

    return train_docs, test_docs, doc_list

train_docs, test_docs, doc_list = read_sentimentDocs()

import multiprocessing
from random import shuffle

cores = multiprocessing.cpu_count()
# PV-DM w/concatenation - window=5 (both sides) approximates paper's 10-word total window size
model = Doc2Vec(dm=1, dm_concat=1, size=100, window=5, negative=5, hs=0, min_count=2, workers=cores)

small_train_docs = train_docs[:15000]

# train for small corpus
model.build_vocab(small_train_docs)
for epoch in xrange(50):
    shuffle(small_train_docs)
    model.train(small_train_docs, total_examples=len(small_train_docs), epochs=1)
# save the model
model.save("small_doc_15000_iter50.bin")


# train for large corpus
model.build_vocab(train_docs)
for epoch in xrange(50):
    shuffle(train_docs)
    model.train(train_docs, total_examples=len(train_docs), epochs=1)
# save the model
model.save("large_doc_25000_iter50.bin")
```
I use those 15k documents as reference document to mix in the new 15k documents, so I got the model2 with 25k documents. Then using those 15k documents and their corresponding doc vectors in model1 and model2, we can train a translation matrix. The code is below:
```
# coding=utf-8
from gensim.models import Doc2Vec
from sklearn.linear_model import LogisticRegression
from collections import namedtuple

import numpy as np

model1 = Doc2Vec.load("small_doc_15000_iter50.bin")
model2 = Doc2Vec.load("large_doc_25000_iter50.bin")

l = model1.docvecs.count
l2 = model2.docvecs.count
m1 = np.array([model1.docvecs[i] for i in range(l)])
m2 = np.array([model2.docvecs[i] for i in range(l)])

# train the translation matrix
tm = np.linalg.lstsq(m2, m1, -1)[0]
# back-mapping for the 15k documents
model1_15k_docvec = []
for i in range(l, l2):
    x = model2.docvecs[i]
    y = np.dot(x, tm)
    model1_15k_docvec.append(y)
```
The problem is how to evaluate the 15k document vector learned by the method, I think we can train a classifier to see whether those vector works. But we need a benchmark for comparison. The code is rough but did I explain the problem clear?@menshikh-iv

gensim/models/translation_matrix.py
```diff
@@ -208,8 +208,8 @@ def translate(self, source_words=None, topn=5, additional=None, source_lang_vec=
         if source_words is None:
             raise RuntimeError("The words to be translated must be provided.")
 
-        if not isinstance(source_words, [string_types, list]):
-            raise ValueError("The word to be translated must be string type or a list of string.")
+        if not isinstance(source_words, (string_types, list, tuple)):
```
piskvorky: This looks rather unpythonic; why is such test needed?

Python uses duck-typing.

robotcator: I see, will remove this type checking.

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+            raise ValueError("Each training data item must contain two different language words.")
+
+        self.source_word, self.target_word = zip(*word_pair)
+        if source_lang_vec is None or target_lang_vec is None:
+            raise RuntimeError("you must provide the source language vectors and target language vectors")
+
+        self.source_lang_vec = source_lang_vec
+        self.target_lang_vec = target_lang_vec
+
+        self.translation_matrix = None
```
menshikh-iv: useless line

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+
+        self.source_space = self.build_space(self.source_lang_vec, set(self.source_word))
```
menshikh-iv: Please call `Space.build(lang_vec, words)` directly

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+        self.target_space = self.build_space(self.target_lang_vec, set(self.target_word))
+
+        self.translation_matrix = self.train(self.source_space, self.target_space)
+
+    def build_space(self, lang_vec, words=None):
```
menshikh-iv: useless method (according to below comment)

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+            target_lang_vec: you can specify the target language vector for retrieving the most similar word,
+            the default is to use the model's target language vector.
+        Returns:
+            A OrderedDict object, each item is (word : topn translated words)
+
+        [1] Dinu, Georgiana, Angeliki Lazaridou, and Marco Baroni. "Improving zero-shot learning by mitigating the
+        hubness problem." arXiv preprint arXiv:1412.6568 (2014).
+        """
+
+        if source_words is None:
```
menshikh-iv: If this argument is mandatory, please don't use `kwargs`

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,86 @@
...
+        with utils.smart_open(train_file, "r") as f:
+            word_pair = [tuple(utils.to_unicode(line).strip().split()) for line in f]
+
+        source_word_vec_file = datapath("EN.1-10.cbow1_wind5_hs0_neg10_size300_smpl1e-05.txt")
+        source_word_vec = KeyedVectors.load_word2vec_format(source_word_vec_file, binary=False)
+
+        target_word_vec_file = datapath("IT.1-10.cbow1_wind5_hs0_neg10_size300_smpl1e-05.txt")
+        target_word_vec = KeyedVectors.load_word2vec_format(target_word_vec_file, binary=False)
+
+        translation_matrix.TranslationMatrix(word_pair, source_word_vec, target_word_vec)
```
menshikh-iv: You create a translation matrix but don't use it, please add the translation as a test. Also add basic check (for translation matrix size for example)

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,86 @@
...
+        source_word_vec = KeyedVectors.load_word2vec_format(source_word_vec_file, binary=False)
+
+        target_word_vec_file = datapath("IT.1-10.cbow1_wind5_hs0_neg10_size300_smpl1e-05.txt")
+        target_word_vec = KeyedVectors.load_word2vec_format(target_word_vec_file, binary=False)
+
+        transmat = translation_matrix.TranslationMatrix(word_pair, source_word_vec, target_word_vec)
+
+        test_word_pair = [("one", "uno"), ("two", "due")]
+        test_source_word, test_target_word = zip(*test_word_pair)
+        transmat.translate(test_source_word, topn=3)
```
menshikh-iv: Add an assert (check that your translation is correct)

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def test_translate_GC(self):
```
menshikh-iv: What's a main difference between NN and GC?

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,86 @@
...
+        source_word_vec = KeyedVectors.load_word2vec_format(source_word_vec_file, binary=False)
+
+        target_word_vec_file = datapath("IT.1-10.cbow1_wind5_hs0_neg10_size300_smpl1e-05.txt")
+        target_word_vec = KeyedVectors.load_word2vec_format(target_word_vec_file, binary=False)
+
+        transmat = translation_matrix.TranslationMatrix(word_pair, source_word_vec, target_word_vec)
+
+        test_word_pair = [("one", "uno"), ("two", "due")]
+        test_source_word, test_target_word = zip(*test_word_pair)
+        transmat.translate(test_source_word, topn=3, additional=10)
```
menshikh-iv: Add an assert (check that your translation is correct)

docs/notebooks/translation_matrix.ipynb
```diff
@@ -0,0 +1,528 @@
...
...hxBeCiHU\nhhBKWmh3SghhWQhhRQjhsiRrzEa+/Z7kWz2QfzVZT8vyrZ5sZDsCdjkwP8Z4OPAEMLVxgxBCX+A8\n4OgY4yCgCDgry/3mvfb0S+Kx5J/2chx54EXgDOBPzTUIIXQCZgFjgSOAiSGEfsmUl518+z3Jt3og\n/2qynpblWz3ZyDaAnQ7MqVueA3y1iTbrgS1AjxBCEbAn8EaW+5WkrMUYl8cYVwKhhWbDgJUxxtUx\nxq3AnWT6Pklqs2wD2H4xxrcAYoxvAvs1bhBjXAvMAF4D/g68F2Ocn+V+JSkpBwFrGtx+vW6dJLVZ\niDG23CCEx4A+DVcBEbgCuCXG+OkGbd+JMe7b6PGfAx4EjgfWAfcAd8cYb29mfy0XJKndiTG2NAKV\nlRb6sB/GGB+oa7MA+F6MsaKJx38NGBtjPL/u9j8Dw2KMFzfR1v5L6oDa0ocVtWKjJzV3XwjhrRBC\nnxjjWyGE/YH/aaLZMcAzMcZ36x7ze+BYoMkAtjs7YkkdT0t9WCv9HTikwe2D69Y1tS/7L0mtku0p\nyHnAOXXLk4A/NNFmOTAihLBHCCEAJwJVWe5XknKtufC0GPh8CKFv3Se4zyLT90lSm2UbwH4OnBRC\nWE4mWF0FEEI4IITwIECMcSlwK/A8sJRMJ/erLPcrSVkLIXw1hLAGGAE8GEJ4uG59wz6sFpgCPAr8\nFbgzxugfkZKysss5YJIkScqtVK+E354ugvgJLkpbHUJYGkL4SwjhuaTrbElrnucQwnUhhJUhhBdC\nCIOTrrE1dnUcIYTRIYT3QggVdT9XpFFna4QQyurmWla20KYQXpMWj6OQXpOG8q0Py5d+KN/6knzr\nE/LtfZ1v788QwsEhhCdC5gLuL4YQPvaBl7p2iTxHramnTc9RjDG1H+Bw4AtkLuJa0kybTsArQF+g\nC/AC0C/Nupup8+fAD+qWLwOuaqbdKqBX2vW25XkGxgF/rFseDvw57brbeByjgXlp19rK4zkeGAxU\nNnN/3r8mrTyOgnlNGtWdV31YPvRD+daX5GOfkG/v63x7fwL7A4Prlj9FZi55mr9DrannEz9HqY6A\nxfZ1EcTWXJQWMseaj9/B2Zrn+XQy8/mIMS4C9g4h9CG/tPb3pSA+rRZjfBpY20KTQnhNWnMcUCCv\nSUN52IflQz+Ub31J3vUJ
```
menshikh-iv: Please add non-trivial examples (fruits, animals, random words, etc). Also please add your experiments in the distinct notebook.

gensim/test/test_translation_matrix.py
robotcator: Given a source element x, the NN method to retrieve a target space label y is to returning the most similar word (according to the cosine similarity in this). The GC method is to return the target space label y which has x  ranked highest. The implementation of GC can computed as :  
GC(x, T) = arg min (Rank_{y, P} （x） - cos(x, y) ), where x is the source element to be translated, T the target retrieval space, P is set of elements sampled from the source space. Sorry for it didn't support Latex format. You can refer this paper in formulation 5. [website](https://arxiv.org/pdf/1412.6568.pdf)

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+    """
+    def __init__(self, matrix, index2word):
+        """
+        matrix: N * length_of_word_vec, which store the word's vector
+        index2word: a list of words int the Space
+        word2index: a dict which for word indexing
+        """
+        self.mat = matrix
+        self.index2word = index2word
+        self.build_word2index()
```
robotcator: Yes, I found that the Exception check is redundant and will remove it.

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+        construct a space class for the lexicon, if it's provided.
+        Args:
+            lang_vec: word2vec model that extract word vector for lexicon
+            lexicon: the default is None, if it is not provided, the lexicon
+                    is all the lang_vec's word, i.e. lang_vec.vocab.keys()
+        Returns:
+            space object for the lexicon
+        """
+        if lang_vec is None:
+            raise RuntimeError("the word vector must be provided!")
```
robotcator: I see, I will fix this here and below.

docs/notebooks/Translation Matrix Revisit.ipynb
```diff
@@ -0,0 +1,341 @@
...
+    "def read_sentimentDocs():\n",
+    "    SentimentDocument = namedtuple('SentimentDocument', 'words tags split sentiment')\n",
+    "\n",
+    "    alldocs = []  # will hold all docs in original order\n",
+    "    with utils.smart_open('aclImdb/alldata-id.txt', encoding='utf-8') as alldata:\n",
+    "        for line_no, line in enumerate(alldata):\n",
+    "            tokens = gensim.utils.to_unicode(line).split()\n",
+    "            words = tokens[1:]\n",
+    "            tags = [line_no] # `tags = [tokens[0]]` would also work at extra memory cost\n",
+    "            split = ['train','test','extra','extra'][line_no//25000]  # 25k train, 25k test, 25k extra\n",
```
menshikh-iv: Missing spaces

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,93 @@
...
+
+        source_word_vec_file = datapath("EN.1-10.cbow1_wind5_hs0_neg10_size300_smpl1e-05.txt")
+        source_word_vec = KeyedVectors.load_word2vec_format(source_word_vec_file, binary=False)
+
+        target_word_vec_file = datapath("IT.1-10.cbow1_wind5_hs0_neg10_size300_smpl1e-05.txt")
+        target_word_vec = KeyedVectors.load_word2vec_format(target_word_vec_file, binary=False)
+
+        transmat = translation_matrix.TranslationMatrix(word_pair, source_word_vec, target_word_vec)
+
+        test_word_pair = [("one", "uno"), ("two", "due")]
```
menshikh-iv: Add more word pairs (and maybe not only numbers)

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,93 @@
...
+        target_word_vec = KeyedVectors.load_word2vec_format(target_word_vec_file, binary=False)
+
+        transmat = translation_matrix.TranslationMatrix(word_pair, source_word_vec, target_word_vec)
+        transmat.save(temp_save_file())
+
+        loaded_transmat = translation_matrix.TranslationMatrix.load(temp_save_file())
+
+        self.assertTrue(np.allclose(transmat.translation_matrix, loaded_transmat.translation_matrix))
+
+    def test_translate_NN(self):
```
menshikh-iv: lower case for method names (here and below)

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def temp_save_file():
+    # temporary data will be stored to this file
+    return os.path.join(tempfile.gettempdir(), 'transmat-en-it.pkl')
+
+
+class TestTranslationMatrix(unittest.TestCase):
+    def test_translation_matrix(self):
+        train_file = datapath("OPUS_en_it_europarl_train_one2ten.txt")
```
menshikh-iv: Please load all needed models/files only once in `setUp` method (because you re-use this several times)

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+# encoding: utf-8
+import warnings
+import numpy as np
+import random
+
+from collections import OrderedDict
+from gensim import utils
+from six import string_types
+
+random.seed(2333)
```
menshikh-iv: Please don't hardcode global random seed. You should create personal `Random` object and use it

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+
+        self.source_word, self.target_word = zip(*word_pair)
+        self.source_lang_vec = source_lang_vec
+        self.target_lang_vec = target_lang_vec
+
+        self.translation_matrix = None
+        self.source_space = Space.build(self.source_lang_vec, set(self.source_word))
+        self.target_space = Space.build(self.target_lang_vec, set(self.target_word))
+
+        self.translation_matrix = self.train(self.source_space, self.target_space)
```
menshikh-iv: Don't run `train` in `__init__` method.

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+        """
+
+        if isinstance(source_words, string_types):
+            # pass only one word to translate
+            source_words = [source_words]
+
+        # if the language word vector not provided by user, use the model's
+        # language word vector as default
+        if source_lang_vec is None:
+            warnings.warn("the parameter source_lang_vec didn't specified,"
```
menshikh-iv: We ignore line size limitation, please write all comment in one line (here and below). Also, use full stops and capitals where your start your sentence.

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+        mapping the source word vector to the target word vector using translation matrix
+        Args:
+            words_space: the Space object that constructed for those words to be translate
+
+        Returns:
+            A Space object that constructed for those mapped words
+        """
+        return Space(np.dot(words_space.mat, self.translation_matrix), words_space.index2word)
+
+    def translate(self, source_words, topn=5, additional=None, source_lang_vec=None, target_lang_vec=None):
```
menshikh-iv: please add argument for algorithm (`additional=None` looks strange)

gensim/models/translation_matrix.py
```diff
@@ -181,16 +182,17 @@ def apply_transmat(self, words_space):
...
          Args:
             source_words(str/list): single word or a list of words to be translated
             topn: return the top N similar words. By default (`topn=5`)
-            additional: defines the training algorithm. By default (`additional=None`), use standard NN retrieval.
-            Otherwise use corrected retrieval(as described in[1]), additional is an int that specify the number of
-            word to sample from the source lexicon.
+            gc: defines the training algorithm. By default (`gc=0`), use standard NN retrieval.
+            Otherwise use globally corrected neighbour retrieval method(as described in[1]).
+            additional: additional is an int that specify the number of word to sample from the source lexicon.
+            if gc=1, then additional must be provided.
```
piskvorky: Additional what?

gensim/models/translation_matrix.py
```diff
@@ -209,20 +211,20 @@ def translate(self, source_words, topn=5, additional=None, source_lang_vec=None,
...
-            warnings.warn("the parameter target_lang_vec isn't specified,"
-                          " use the model's target language word vector as default")
+            warnings.warn("The parameter target_lang_vec isn't specified, use the model's target language word vector as default")
             target_lang_vec = self.target_lang_vec
 
         # if additional is provided, bootstrapping vocabulary from the source language word vector model.
-        if additional is not None:
+        if gc:
+            if additional is None:
+                raise RuntimeError("When using the globally corrected neighbour retrieval method, the additional parameter which the number of words sampled from source space must be provided.")
```
piskvorky: There seems to be a word missing after `which` -- the sentence doesn't make sense now.

gensim/models/translation_matrix.py
robotcator: Sorry for missing understanding and I will use another word. The parameter 'additional' is the number of words which sampled from the source language(usually the vocabulary in the source language's word2vec object) for the globally corrected neighbor retrieval method.

gensim/models/translation_matrix.py
robotcator: Thanks, I will revise it. And if there is any other inappropriate expression, please point out.

gensim/models/translation_matrix.py
piskvorky: Ah, I see. You meant literal `additional`.

It's best to format code as literals (using ``` backticks), to avoid misunderstandings like this.

docs/notebooks/Translation Matrix Revisit.ipynb
```diff
@@ -0,0 +1,344 @@
+{
```
menshikh-iv: Remove spaces in the filename `Translation_Matrix_Revisit.ipynb`

docs/notebooks/Translation Matrix Revisit.ipynb
```diff
@@ -0,0 +1,344 @@
...
+    "train_array = np.zeros((25000, 100))\n",
+    "train_label = np.zeros((25000, 1))\n",
+    "test_array = np.zeros((25000, 100))\n",
+    "test_label = np.zeros((25000, 1))\n",
+    "\n",
+    "for i in range(12500):\n",
+    "    train_array[i] = m2[i]\n",
+    "    train_label[i] = 1\n",
+    "\n",
+    "    train_array[i+12500] = m2[i+12500]\n",
```
menshikh-iv: Spaces before/after `+` (here and anywhere)

docs/notebooks/Translation Matrix Revisit.ipynb
```diff
@@ -0,0 +1,344 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "100000 docs: 25000 train-sentiment, 25000 test-sentiment\n",
+      "25000 25000 100000 15000 50000\n"
+     ]
+    }
+   ],
+   "source": [
```
menshikh-iv: Need more explanation between cells (what happens)

docs/notebooks/translation_matrix.ipynb
```diff
@@ -0,0 +1,1352 @@
...
+       "<script>requirejs.config({paths: { 'plotly': ['https://cdn.plot.ly/plotly-latest.min']},});if(!window.Plotly) {{require(['plotly'],function(plotly) {window.Plotly=plotly;});}}</script>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from sklearn.decomposition import PCA\n",
+    "import matplotlib.pyplot as plt\n",
```
menshikh-iv: Please use plotly for all viz

docs/notebooks/translation_matrix.ipynb
```diff
@@ -0,0 +1,1352 @@
+{
```
menshikh-iv: Also please add a description before each image (what I see now).

docs/notebooks/translation_matrix.ipynb
```diff
@@ -0,0 +1,1352 @@
...
...Ne4Xrn3uHO\ncOfcOd9PMmF+vHfmed/zzjw8c86cdy3QSa/iNSLGAr+fmYsBMvMjYE+DrytJktRy3d3dPPnkk1x5\n5ZUcc8wx/bY9/PDDARgxYgQfffRRn232L1j33/P6wQcfHNDm6aefZtSoUY2ELknDTqO/eZ2QmTsB\nMvMNYEIfbU4B3oyIf46IzRFxb0SMafB1JUmSWmrZsr9k6tTfZcmSf6SjYx2rV3+Pt956C6icwKke\n55xzDh0dHQBs3bqVrq6unscmTpzItm3b+Pjjj1m/fn3P/eeff/4BhyRv2bKlGd2RpMIbsHiNiEci\nomu/y/PVfy/qo3lfx7eMBGYC38vMmcB79H1osSRJ0rDQ3d3N6tX3Ar8AtgEb2b17L3PmzGHGjBnc\ncMMNB+w5rXX24aVLl7J3715OP/10li9fzqxZs3oeu+WWW7jgggs4++yzmTRpUs/9d955J8888wxn\nnnkm06ZN45577jlEvZSkYmloqZyI6AbmZebOiJgIPJ6ZU3q1ORb4z8z8YvX22cCNmflHNZ4zb775\n5p7b8+bNY968eYOOUZJUHp2dnXR2dvbcXrFihUvlNIFL5XzW2rVrWbz4H6gUrp84le9//yYWLVo0\n6OedP38+t99+OzNnzmw4RkkqokaWymn0N68bgMXASmAR8GDvBtXC9pWIODUztwNfBfo9Jd7y5csb\nDEuSVEa9v/BcsWJF64JRW5s9ezbwCtAFnFH999Xq/YPn+rCSVFuje16/APwIOBF4mcpSObsj4jjg\nvsy8sNruTCpL5YwCXgK+0deSOtW2frsrSWqKRr7d1afMzX1btux6Vq++DzgBeJVrr72aVavuHOjP\nJKnUGsnNDRWvh4IJUpLULBavzWFurq27u5uNGzcye/ZspkyZMvAfSFLJWbxKktQHi9fmMDdLkpql\nkdzc6FI5kiRJkiQdchavkiRJkqTCs3iVJEmSJBWexaskSZIkqfAsXiVJkiRJhWfxKkmSJEkqPItX\nSZIkSVLhWbxKkiRJkgrP4lWSJEmSVHgWr5IkSZKkwrN4lSRJkiQVnsWrJEmSJKnwLF4lSZIkSYVn\n8SpJkiRJKjyLV0mSJElS4Vm8SpIkSZIKz+JVkiRJklR4Fq9qqVWrVjF16lSOPvpobr311laHI0mS\nJKmgIjNbHcMBIiKLFp
```
menshikh-iv: E251 unexpected spaces around keyword / parameter equals (here and anywhere)

menshikh-iv: E226 missing whitespace around arithmetic operator (here and anywhere)

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,118 @@
...
+        self.source_doc_vec_file = datapath("small_tag_doc_5_iter50")
+        self.target_doc_vec_file = datapath("large_tag_doc_10_iter50")
+
+        self.source_doc_vec = Doc2Vec.load(self.source_doc_vec_file)
+        self.target_doc_vec = Doc2Vec.load(self.target_doc_vec_file)
+
+    def test_translation_matrix(self):
+        model = translation_matrix.BackMappingTranslationMatrix(self.train_docs[:5], self.source_doc_vec, self.target_doc_vec)
+        transmat = model.train(self.train_docs[:5])
+        self.assertEqual(transmat.shape, (100, 100))
```
menshikh-iv: Why only "sanity check"? Please add another tests (similar with translation tests from previous class)?

docs/notebooks/translation_matrix.ipynb
robotcator: done

docs/notebooks/translation_matrix.ipynb
robotcator: Done

gensim/test/test_translation_matrix.py
robotcator: This is a little different with the translation of word. We build the mapping `F` between the target model and original model(old). Given a document vector `x` from the target model,  we infer the vector by multiple `F` with `x` to get a new vector for the original vector. By now, we didn't have a good evaluation metric to assess the vector. So I only give the "sanity check" now.

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+            translated_target_word = []
+            # search the most `topn` similar words
+            for j in range(topn):
+                map_space_id = sim_matrix_idx[j, source_space.word2index[word]]
+                translated_target_word.append(target_space.index2word[map_space_id])
+            translated_word[word] = translated_target_word
+        return translated_word
+
+
+class BackMappingTranslationMatrix(utils.SaveLoad):
```
menshikh-iv: Please add example with viz for it (to translation matrix notebook)

gensim/test/test_translation_matrix.py
menshikh-iv: You can calculate cosine distance between vector from model and vector from translation

docs/notebooks/Translation_Matrix_Revisit.ipynb
```diff
@@ -0,0 +1,366 @@
+{
```
menshikh-iv: Please add this test to `translation_matrix.ipynb` (as another section) and remove  `Translation_Matrix_Revisit.ipynb`

menshikh-iv:
In general, LGTM for me, have you any suggestions @gojomo @piskvorky?

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+import numpy as np
+
+from collections import OrderedDict
+from gensim import utils
+from six import string_types
+
+
+"""
+Produce translation matrix to translate the word from one language to another language, using either
+standard nearest neighbour method or globally corrected neighbour retrieval method [1].
```
gojomo: Could note that this is potentially more useful than just word-to-word translation; it'll work for any two sets of named-vectors where there are some paired-guideposts to learn the transformation.

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+This method can be used to augment the existing phrase tables with more candidate translations, or
+filter out errors from the translation tables and known dictionaries [2].
+
+Initialize a model with e.g.::
+
+    >>> transmat = TranslationMatrix(word_pair, source_word_vec, target_word_vec)
+
+Train a model with e.g.::
+
+    >>> transmat.train(word_pair)
```
gojomo: This API that supplies `word_pair` twice - both at construction & a later necessary `train()` step – seems odd to me, and in contrast to other gensim practice (where training can be combined with initialization, if training data is supplied.)

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+"""
+
+
+class Space(object):
+    """
+    An auxiliary class for storing the the words space
+
+    Attributes:
+        `mat` (ndarray): each row is the word vector of the lexicon
+        `index2word` (list): a list of lexicon
```
gojomo: Terms like 'lexicon' unclear: is `index2word` a list of strings?

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+
+        else:
+            for item in lang_vec.vocab.keys():
+                words.append(item)
+                mat.append(lang_vec.syn0[lang_vec.vocab[item].index])
+
+        return Space(mat, words)
+
+    def normalize(self):
+        """ normalized the word vector's matrix """
```
gojomo: Capitalization, verb-tense: 'Normalize...'

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+    """
+    def __init__(self, word_pair, source_lang_vec, target_lang_vec, random_state=None):
+        """
+        Initialize the model from a list pair of `word_pair`. Each word_pair is tupe
+         with source language word and target language word.
+
+        Examples: [("one", "uno"), ("two", "due")]
+
+        Args:
+            `word_pair` (list): a list pair of words
```
gojomo: As a list of many word-pairs, a more natural name for this parameter would be plural  `word_pairs`.

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+        self.source_word, self.target_word = zip(*word_pair)
+        self.source_lang_vec = source_lang_vec
+        self.target_lang_vec = target_lang_vec
+
+        self.random_state = utils.get_random_state(random_state)
+        self.translation_matrix = None
+        self.source_space = None
+        self.target_space = None
+
+    def train(self, word_pair):
```
gojomo: This `word_pair` argument, which I noted seemed odd above, doesn't seem to be consulted at all in the following code. (Only the word-pairs supplied at initialization have any effect.)

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+
+        m1 = self.source_space.mat[[self.source_space.word2index[item] for item in self.source_word], :]
+        m2 = self.target_space.mat[[self.target_space.word2index[item] for item in self.target_word], :]
+
+        self.translation_matrix = np.linalg.lstsq(m1, m2, -1)[0]
+        return self.translation_matrix
+
+    def save(self, *args, **kwargs):
+        """
+        Save the model to file but ignoring the souce_space and target_space
```
gojomo: `source_space`

gensim/models/translation_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+
+    @classmethod
+    def load(cls, *args, **kwargs):
+        """ load the pre-trained translation matrix model"""
+        model = super(TranslationMatrix, cls).load(*args, **kwargs)
+        return model
+
+    def apply_transmat(self, words_space):
+        """
+        mapping the source word vector to the target word vector using translation matrix
```
gojomo: Capitalization, tense: "Map…"

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,125 @@
...
+
+        loaded_model = translation_matrix.TranslationMatrix.load(temp_save_file())
+        self.assertTrue(np.allclose(model.translation_matrix, loaded_model.translation_matrix))
+
+    def test_translate_nn(self):
+        # test the nearest neighbor retrieval method
+        model = translation_matrix.TranslationMatrix(self.word_pair, self.source_word_vec, self.target_word_vec)
+        model.train(self.word_pair)
+
+        test_word_pair = [("one", "uno"), ("two", "due"), ("apple", "mela"), ("orange", "aranicione"), ("dog", "cane"), ("pig", "maiale"), ("cat", "gatto")]
```
gojomo: Every one of these word-pairs is in the training mappings. A 'truer' test of 'translation' functionality would verify that using these known signpost-equivalences creates a mapping that's useful for getting *other* word-to-word translations approximately right.

gensim/test/test_translation_matrix.py
```diff
@@ -0,0 +1,125 @@
...
+        test_source_word, test_target_word = zip(*test_word_pair)
+        translated_words = model.translate(test_source_word, topn=3)
+
+        self.assertTrue("uno" in translated_words["one"])
```
gojomo: This could be a loop driven by `test_word_pair`.

gojomo:
The constant stream of tiny tweak change-notifications, without any deeper review of the docs/naming issues I'd mentioned way-back, made me think this work was still far from review-readiness.  

I just now see it was it marked for `[MRG]` and merged in the last 10 hours. 

I think the API, naming, comments, and testing are all still a bit off. I've added some comments, but all of the code-doc-comments should get additional style/clarity proofreading.

robotcator:
@gojomo Thank you for reviewing, there still need more improvements. And I will fix the problem according to your comments.

gensim/models/translation_matrix.py
robotcator: Yes, index2word is a list of words(strings).

gensim/models/translation_matrix.py
robotcator: Done.

gensim/models/translation_matrix.py
robotcator: fixed

gensim/models/translation_matrix.py
robotcator: As you said in the second comment, if training was combined with initialization, then this parameter would be used in `train` method.

gensim/models/translation_matrix.py
robotcator: Thanks, I will fix all the grammar errors and typos here and below.

gensim/models/translation_matrix.py
gojomo: Looking at the code body of the `train()` method, I see no references to this `word_pair` parameter. Hence, the parameter is unused and (unless the code changes) unnecessary here. If the intent was that this be an alternate way to supply the list-of-word-pairs, the code would have to change to reflect that intent.

robotcator: Okay, I change the code where the `word_pairs` is optional  parameter in the initialization, and if provided, then `train` method will run followed.

robotcator:
@gojomo Hi, I open a new [PR](https://github.com/RaRe-Technologies/gensim/pull/1594) to fix the comment according to you. I change the `train` method to make it compulsory and in the `__init__` method to optional, but If the `word_pairs` is provided, the `train` method will also run. And I will make change in the notebook later.

PR #1437, opened 2017-06-21
[WIP] Added sklearn wrapper for w2v model

chinmayapancholi13:
This PR adds scikit-learn wrapper for Gensim's Word2Vec model.


gensim/sklearn_integration/sklearn_wrapper_gensim_w2vmodel.py
```diff
@@ -0,0 +1,116 @@
...
+import numpy as np
+import six
+from sklearn.base import TransformerMixin, BaseEstimator
+from sklearn.exceptions import NotFittedError
+
+from gensim import models
+from gensim.sklearn_integration import BaseSklearnWrapper
+
+
+class SklW2VModel(BaseSklearnWrapper, TransformerMixin, BaseEstimator):
```
gojomo: As this is already inside the `sklearn_integration` package, and will usually serve (in sklearn pipelines) as a `Transformer`, I'd suggest the name `W2VTransformer` as a more simple, direct, and non-abbreviation-reliant class name.

gensim/sklearn_integration/sklearn_wrapper_gensim_w2vmodel.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    Base Word2Vec module
+    """
+
+    def __init__(self, size=100, alpha=0.025, window=5, min_count=5,
+            max_vocab_size=None, sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
+            sg=0, hs=0, negative=5, cbow_mean=1, hashfxn=hash, iter=5, null_word=0,
+            trim_rule=None, sorted_vocab=1, batch_words=10000):
```
gojomo: To ensure these values are always kept in sync with the underlying `Word2Vec` class, it might be appropriate to use `inspect.getargspec()` on the `Word2Vec.__init__()` method. Perhaps this method could then just take `**kwargs`, with the defaults dict from `getargspec()` merged together. (Keeping these as a dict might further make many of the explicit attributes unnecessary, and the implementation of `get_params()` trivial.)

gensim/sklearn_integration/sklearn_wrapper_gensim_w2vmodel.py
```diff
@@ -0,0 +1,116 @@
...
+            X[k] = word_vec
+
+        return np.reshape(np.array(X), (len(words), self.size))
+
+    def partial_fit(self, X):
+        """
+        Update model using newly added sentences.
+        """
+        if self.gensim_model is None:
+            self.gensim_model = models.Word2Vec(size=self.size, alpha=self.alpha,
```
gojomo: Since `train()` (below) requires additional model vocab initialization (as by `build_vocab()`), I can't see this path, where the model doesn't already exist, possibly working. More generally, I'm not sure a `partial_fit()` method makes sense for Word2Vec - incremental training is at best experimental, and not at all the usual mode of operation.

gensim/sklearn_integration/sklearn_wrapper_gensim_w2vmodel.py
chinmayapancholi13: I agree surely that `W2VTransformer` would be a simpler and non-abbreviation-reliant class name however `SklW2VModel` is consistent with the naming of the scikit-learn wrappers added for other Gensim models like `SklLdaModel`, `SklLsiModel`, `SklRpModel`. `SklLdaSeqModel` etc.

gensim/sklearn_integration/sklearn_wrapper_gensim_w2vmodel.py
chinmayapancholi13: Yes, this did come up during some earlier discussions about the implementation as well. :)
I have now modified the `partial_fit` function to raise a `NotImplementedError` on being invoked.

gensim/sklearn_integration/sklearn_wrapper_gensim_w2vmodel.py
gojomo: Hmm, are those classes old & already-widely relied upon, or also new/recent? It may make sense to fix all their names to avoid an idiosyncratic abbreviation (`Skl`) and match sklearn's role-terminolog! cc @piskvorky

chinmayapancholi13: All these classes are relatively recent and I don't think they are currently widely relied upon in the codebase. So yes it might be a good idea to change the names for all the existing wrapper classes.

piskvorky: I'm not aware of the history of those classes but agree with @gojomo we don't need the (sub)package repeated in class names.

PR #1444, opened 2017-06-22
Random sample on stream

vlejd:
Related to #308.

Fix comments in  #1422.


gensim/corpora/textcorpus.py
```diff
@@ -130,11 +130,14 @@ def sample_texts(self, n, seed=None, length=None):
...
-            raise ValueError("n is larger than length of corpus.")
+
+            raise ValueError("n {0:d} is larger than length of corpus {1:d}.".format(n, length))
         if not 0 <= n:
-            raise ValueError("Negative sample size.")
+            raise ValueError("Negative sample size n {0:d}.".format(n))
 
+        seen_documents = 0
         for i, sample in enumerate(self.get_texts()):
+            seen_documents = i + 1
```
piskvorky: What is the purpose of introducing this new variable?

piskvorky: Oh I think I see -- you can directly re-use the counter `i` from enumerate (just give it a better name).

vlejd: It may be only my personal opinion, but I do not like this type of "variable leaking" out of the for loop scope.

piskvorky: OK, sure. Then I suggest `seen_documents += 1` at least.

menshikh-iv:
Ping @vlejd, when you finish your PR?

menshikh-iv:
Thank you @vlejd

gensim/corpora/textcorpus.py
```diff
@@ -305,7 +302,7 @@ def sample_texts(self, n, seed=None, length=None):
             # This means that length was set to be greater than number of items in corpus
             # and we were not able to sample enough documents before the stream ended.
             raise ValueError(
-                "length {0:d} greater than number of documents in corpus {1:d}".format(length, seen_documents))
+                "length {0:d} greater than number of documents in corpus {1:d}".format(length, i + 1))
```
piskvorky: This will fail with an exception if there are no texts (`i` will be uninitialized).

menshikh-iv: `i` undefined if only `self.getstream()` is empty

PR #1446, opened 2017-06-23
Fast object counting + Phrases

piskvorky:
Create a new class for fast, memory-efficient object counting (bounded RAM footprint, even for super large corpora).

To be used in Phrases, Dictionary, Word2vec etc. Basically everywhere where a vocabulary larger than RAM can be expected (long tail, Zipf's law).

Perhaps we can release this as a separate, stand-alone Python library (widely useful even outside of gensim)?

* [x] design APIs
* [ ] choose efficient algorithms and data structures
* [ ] implement reference implementation in raw Python (slow but correct)
* [ ] optimize hotspots in C/Cython
* [ ] parallelize, if necessary
* [ ] implement Dictionary, Phrases, Word2vec, Doc2vec using this new counter

### DIFF
--- a/gensim/models/fast_counter.py
+++ b/gensim/models/fast_counter.py
@@ -15,7 +15,7 @@ Such counts are useful in various other modules, such as Dictionary, TfIdf, Phra
 
 import sys
 import os
-from collections import Counter
+from collections import defaultdict
 import logging
 
 from gensim import utils
@@ -51,7 +51,7 @@ class FastCounter(object):
         self.doc2items = doc2items
         self.max_size = max_size
         self.min_reduce = 0
-        self.hash2cnt = Counter()  # TODO replace by some GIL-free low-level struct
+        self.hash2cnt = defaultdict(int)
 
     def hash(self, item):
         return hash(item)
@@ -63,10 +63,8 @@ class FastCounter(object):
         If the memory structures get too large, clip them (then the internal counts may be only approximate).
         """
         for document in documents:
-            # TODO: release GIL, so we can run update() in parallel threads.
-            # Or maybe not needed, if we create multiple FastCounters from multiple input streams using
-            # multiprocessing, and only .merge() them at the end.
-            self.hash2cnt.update(self.hash(ngram) for ngram in self.doc2items(document))
+            for item in self.doc2items(document):
+                self.hash2cnt[self.hash(item)] += 1
             self.prune_items()
 
         return self  # for easier chaining
@@ -103,8 +101,8 @@ class Phrases(object):
         self.min_count = min_count
         self.max_vocab_size = max_vocab_size
         # self.counter = FastCounter(iter_gram12, max_size=max_vocab_size)
-        # self.counter = FastCounterCython()
-        self.counter = FastCounterPreshed()
+        self.counter = FastCounterCython()
+        # self.counter = FastCounterPreshed()
 
     def add_documents(self, documents):
         self.counter.update(documents)
--- a/gensim/models/fast_counter.py
+++ b/gensim/models/fast_counter.py
@@ -7,7 +7,9 @@
 """
 Fast & memory efficient counting of things (and n-grams of things).
 
-This module is designed to count *item* and *document* frequencies over large, streamed corpora (lazy iteration).
+This module is designed to count item frequencies over large, streamed corpora (lazy iteration).
+
+Such counts are useful in various other modules, such as Dictionary, TfIdf, Phrases etc.
 
 """
 
@@ -42,12 +44,9 @@ class FastCounter(object):
     Fast counting of item frequency and document frequency across large, streamed iterables.
     """
 
-    def __init__(self, doc2items=iter_gram1, collect_df=False):
+    def __init__(self, doc2items=iter_gram1):
         self.doc2items = doc2items
-        self.collect_df = collect_df
-
-        self.item_counts = Counter()  # TODO replace by some GIL-free low-level struct
-        self.doc_counts = Counter()  # TODO replace by some GIL-free low-level struct
+        self.hash2cnt = Counter()  # TODO replace by some GIL-free low-level struct
 
     def hash(self, key):
         return hash(key)
@@ -62,11 +61,7 @@ class FastCounter(object):
             # TODO: release GIL, so we can run update() in parallel threads.
             # Or maybe not needed, if we create multiple FastCounters from multiple input streams using
             # multiprocessing, and only .merge() them at the end.
-            item_cnts = Counter(self.hash(ngram) for ngram in self.doc2items(document))
-            self.item_counts.update(item_cnts)
-            if self.collect_df:
-                # increment by 1 per unique key ("document frequency")
-                self.doc_counts.update(iterkeys(item_cnts))
+            self.hash2cnt.update(self.hash(ngram) for ngram in self.doc2items(document))
 
             # self.prune_vocab()
 
@@ -79,7 +74,7 @@ class FastCounter(object):
 
     def get(self, key, default=None):
         """Return the item frequency of `key` (or `default` if key not present)."""
-        return self.item_counts.get(self.hash(key), default)
+        return self.hash2cnt.get(self.hash(key), default)
 
     def merge(self, other):
         """
@@ -89,7 +84,7 @@ class FastCounter(object):
         raise NotImplementedError
 
     def __len__(self):
-        return len(self.item_counts)
+        return len(self.hash2cnt)
 
     def __str__(self):
         return "%s<%i items>" % (self.__class__.__name__, len(self))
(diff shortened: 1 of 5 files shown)


piskvorky:
@gojomo @menshikh-iv @jayantj WDYT?

This could be a starting point for the new API, to be optimized and then used in Phrases and Dictionary.

piskvorky:
Using [preshed](https://github.com/explosion/preshed) seems to bring another ~2x speedup over plain `Counter` (58s for FastCounter vs 18s for cythonized FastCounter vs 8s for cythonized Preshed, on the [text8 corpus](http://mattmahoney.net/dc/text8.zip)).

gensim/models/fast_counter.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def update(self, documents):
+        """
+        Update the relevant ngram counters from the iterable `documents`.
+
+        If the memory structures get too large, clip them (then the internal counts may be only approximate).
+        """
+        for document in documents:
+            for item in self.doc2items(document):
+                self.hash2cnt[self.hash(item)] += 1
```
gojomo: That this implementation ignores hash-collisions is important detail to document.

gensim/models/fast_counter_cython.pyx
```diff
@@ -0,0 +1,137 @@
...
+        hash2cnt = self.hash2cnt
+        for document in documents:
+            l = len(document)
+            if l:
+                h1 = chash(document[0])
+                hash2cnt[h1] += 1
+                for idx in range(1, l):
+                    h2 = chash(document[idx])
+                    hash2cnt[h2] += 1
+                    hash2cnt[h1 + h2] += 1
```
gojomo: Simple addition to determine hash of bigram gives (A, B) and (B, A) same hash value – problematic for Phrase-promotion.

gojomo:
Worth testing against `collections.Counter`, too, for reference.

Certainly worth trying as drop-in for Phrases - and If/when cythonized code or preshed-ized code gives identical output on representative inputs, in less time or memory, replace existing steps.

gensim/models/fast_counter_cython.pyx
piskvorky: Yeah, I had XOR here before, but then (A, A) == (B, B), which is also not good. This requires more thought, too hackish.

piskvorky:
More interested in feedback for the API now, not performance.

Is the interface complete, anything else we should support? Do you envision any issues when plugging this into Dictionary? Or in to other use-cases that need counting?

Btw tried Counter (see commit history), it's quite a bit slower than defaultdict (more Python layers).

piskvorky:
On second non-4-am thoughts, how about simply mimicking the API of stdlib `Counter`?

Just slap one more ctr parameter (`max_size`, either #items or #RAM bytes), and one more method (`update_from_documents`) and keep the rest standard?

jayantj:
About the API, I'd planned on the same, using the same API as `Counter` is the most intuitive, I think. (though I'm not sure what stdlib `Counter` means, I can't find anything about it either - is it the `Counter` from `collections`?)
Simply updating counts from an iterable of hashables is the most common use case IMO, the `update_from_documents` is more specific to Gensim/Phrases. It might even be a good idea to keep `update_from_documents` out of `FastCounter` completely, and add a simple wrapper/subclass over `FastCounter` for Gensim with the method. The notion of "documents" is again specific to the Gensim use case, I think.

Also not sure about `BoundedCounter` since `max_size` should only be an optional param. The current `FastCounter` sounds good to me.

gensim/models/fast_counter.py
```diff
@@ -0,0 +1,150 @@
...
+            self.prune_items()
+
+        return self  # for easier chaining
+
+    def prune_items(self):
+        """Trim data structures to fit in memory, if too large."""
+        # XXX: Or use a fixed-size data structure to start with (hyperloglog?)
+        while self.max_size and len(self) > self.max_size:
+            self.min_reduce += 1
+            utils.prune_vocab(self.hash2cnt, self.min_reduce)
```
jayantj: This can potentially be very slow, as very `utils.prune_vocab` call is going to iterate over the entire vocabulary. Could be worthwhile to store a `min_count` property. Depending on the data structures used, this might not be trivial though.
Anyway, the discussion right now is API/design and not optimization, I suppose.

gensim/models/fast_counter.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def get(self, item, default=None):
```
jayantj: `__getitem__` would be good too.

piskvorky:
Without `max_size`, what would be the difference to `Counter`? Why not just use `defaultdict`?

I think the point of this class is that it can process very large iterables (corpora).

gensim/models/fast_counter.py
piskvorky: What would this `min_count` property do?

gojomo: I'd want to see profiling that indicates this survey is a major time-cost before spending much time optimizing. If it needed to be optimized, it might be possible to retain a `min_reduce`/`min_count` threshold and, during tallying, track when a key's tally exceeds this count, perhaps removing it from a 'most-eligible-for-pruning' list. 

With this existing algorithm, where each prune, everything under the count reached last time is automatically discarded, this hypothetical "most eligible" list could be discarded 1st – without even looking at the existing counts again. Maybe that'd be enough... though if keeping this 'escalating min_reduce' you'd still need to iterate over all the remaining keys. (It seems a lot like GC, with a 'tenured' set that might be able to skip some collections.)

But as I've mentioned in previous discussions around Phrases: I suspect this 'escalating floor prune' is a somehat undesirable algorithm, injecting more imprecision/bias in final counts than necessary, and (maybe) losing a lot of keys that would survive a truly unbounded counting method. (TLDR: because of typical freq-distributions plus the always-escalating floow, every prune may wind up eliminating 90%+ of all existing words, starving some words of the chance to ever survive a prune, or in some cases inefficiently leaving the final dict at a small fraction of the desired `max_size`.) Maybe it's necessary to prevent too-frequent prunes, but I'd prefer an option to be more precise via less-frequent prunes – especially if I'd hypothetically minimized the full-time-cost of Phrases-analysis via other process improvements or effective parallelism.

piskvorky: It looks like we're on to optimizations and choosing algorithms, so let me tick off the "API design" checkbox. Conclusion: mimic Counter/defaultdict API, except add a parameter that allows restricting its size.

gensim/models/fast_counter.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+
+    def __init__(self, doc2items=iter_gram1, max_size=None):
+        self.doc2items = doc2items
+        self.max_size = max_size
+        self.min_reduce = 0
+        self.hash2cnt = defaultdict(int)
+
+    def hash(self, item):
+        return hash(item)
```
menshikh-iv: Maybe use mmh3 for this purposes?

gensim/models/fast_counter.py
menshikh-iv: @piskvorky I agree with you
Also, maybe need to add a static method for a size calculation (for example, if we store X tokens what's megabytes we needed for this in RAM). It's can be useful if we choose a `max_size` parameter.

gensim/models/fast_counter_cython.pyx
menshikh-iv: We can use `h1 ** h2`, but this will lead to overflow

piskvorky: That's not a native operation (instruction), probably not a good idea.

Google shows me this for `boost::hash_combine`
```C
size_t hash_combine( size_t lhs, size_t rhs ) {
  lhs^= rhs + 0x9e3779b9 + (lhs << 6) + (lhs >> 2);
  return lhs;
}
```

But probably anything that breaks the symmetry would be enough, like `7*lhs + rhs`.

piskvorky: Here's a repo with a bunch of useful hashing algorithms & their comparison: https://github.com/rurban/smhasher

gensim/models/fast_counter.py
piskvorky: Yes, though the best place for that is probably the docstring. Something like `1GB gives you room for approximately 100,000,000 item, and the number scales linearly for more/less RAM` (or whatever the case may be).

piskvorky:
Additional experiments and timings from @jayantj , building on top of this PR: https://github.com/jayantj/gensim/blob/fast_counter_modifications/docs/notebooks/counter_benchmarks.ipynb

gensim/models/fast_counter_cython.pyx
piskvorky: One more interesting resource on hashing (in the context of Bloom filters / min-sketch counting in #508 ): https://www.eecs.harvard.edu/~michaelm/postscripts/rsa2008.pdf
CC @isamaru

menshikh-iv:
Continued in a distinct repo.

PR #1448, opened 2017-06-25
Add flag parameters for topic diff

parulsethi:
Add flag parameters to optionally calculate diff matrix, diff diagonal and annotations.


menshikh-iv:
The main idea of this PR - for diagonal case calculates only diagonal values (with/without annotation for diagonal), it's needed for free speed-up this function for concrete case (for example - visdom integration). Please change logic (first, check for diagonal case) and refactor code (remove duplication)

gensim/models/ldamodel.py
```diff
@@ -1020,28 +1020,38 @@ def diff(self, other, distance="kullback_leibler", num_words=100, n_ann_terms=10
...
-                z[topic1][topic2] = distance_func(d1[topic1], d2[topic2])
-
-        if normed:
-            if np.abs(np.max(z)) > 1e-8:
-                z /= np.max(z)
-
-        annotation = [[None] * t1_size for _ in range(t2_size)]
+        if diagonal:
+            t_size = min(t1_size, t2_size)
+            z = np.zeros(t_size)
```
parulsethi: If the no. of topics is different between two models (diff will not be a square matrix), then what should be returned in the `diagonal` case? Either raise an error for num_topics to be same or just return the diff of identical topic no.s till the smaller value of num_topic of both model (latter one implemented right now)?

menshikh-iv: I think you should raise error

parulsethi: Replaced to raise the error instead

gensim/models/ldamodel.py
```diff
@@ -1020,28 +1022,41 @@ def diff(self, other, distance="kullback_leibler", num_words=100, n_ann_terms=10
...
                 neg_tokens = fst_topics[topic1].symmetric_difference(snd_topics[topic2])
 
                 pos_tokens = sample(pos_tokens, min(len(pos_tokens), n_ann_terms))
                 neg_tokens = sample(neg_tokens, min(len(neg_tokens), n_ann_terms))
 
-                annotation[topic1][topic2] = [pos_tokens, neg_tokens]
+                diff_terms[topic] = [pos_tokens, neg_tokens]
+
+        if normed:
+                if np.abs(np.max(z)) > 1e-8:
```
menshikh-iv: excess indent

gensim/models/ldamodel.py
```diff
@@ -1020,28 +1022,41 @@ def diff(self, other, distance="kullback_leibler", num_words=100, n_ann_terms=10
...
+            assert t1_size == t2_size, "Both input models should have same no. of topics, as the diagonal will only be valid in a square matrix"
+            # initialize z and annotation array
+            z = np.zeros(t1_size)
+            if annotation:
+                diff_terms = np.zeros(t1_size, dtype=list)
+        else:
+            # initialize z and annotation matrix
+            z = np.zeros((t1_size, t2_size))
+            if annotation:
+                diff_terms = np.zeros((t1_size, t2_size), dtype=list)
```
menshikh-iv: This terms for `diff` and `intersection` (not only `diff_terms`), please rename this variable to old variant (`annotation`)

gensim/models/ldamodel.py
```diff
@@ -1020,28 +1022,41 @@ def diff(self, other, distance="kullback_leibler", num_words=100, n_ann_terms=10
...
+                topic2 = topic[1]
 
-        for topic1 in range(t1_size):
-            for topic2 in range(t2_size):
+            z[topic] = distance_func(d1[topic1], d2[topic2])
+            if annotation:
                 pos_tokens = fst_topics[topic1] & snd_topics[topic2]
                 neg_tokens = fst_topics[topic1].symmetric_difference(snd_topics[topic2])
 
                 pos_tokens = sample(pos_tokens, min(len(pos_tokens), n_ann_terms))
```
menshikh-iv: As I remember, you want to remove `sample` for another PR, you already do it in different PR?

parulsethi: Yep, it's in https://github.com/RaRe-Technologies/gensim/pull/1484

gensim/models/ldamodel.py
parulsethi: `annotation` is already being used as boolean parameter to indicate if annotations should be calculated or not. So I've renamed it to `annotation_terms`.

Issue #1441, opened 2017-06-22
Windows 2.2.0 issues


menshikh-iv:
We have several errors in tests if I use Windows

2 errors, connected with multithreading in windows (???) from [#1349](https://github.com/RaRe-Technologies/gensim/pull/1349), python2 and python3
```
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python35-x64\lib\site-packages\gensim\test\test_text_analysis.py", line 57, in test_occurrence_counting
    self.assertEqual(3, accumulator.get_occurrences("this"))
AssertionError: 3 != 0
-------------------- >> begin captured logging << --------------------
gensim.topic_coherence.text_analysis: INFO: 1 batches submitted to accumulate stats from 64 documents (3 virtual)
gensim.topic_coherence.text_analysis: INFO: 2 accumulators retrieved from output queue
gensim.topic_coherence.text_analysis: INFO: accumulated word occurrence stats for 4 virtual documents
--------------------- >> end captured logging << ---------------------
======================================================================
FAIL: test_occurrence_counting2 (gensim.test.test_text_analysis.TestParallelWordOccurrenceAccumulator)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python35-x64\lib\site-packages\gensim\test\test_text_analysis.py", line 67, in test_occurrence_counting2
    self.assertEqual(2, accumulator.get_occurrences("human"))
AssertionError: 2 != 0
-------------------- >> begin captured logging << --------------------
gensim.topic_coherence.text_analysis: INFO: 2 accumulators retrieved from output queue
gensim.topic_coherence.text_analysis: INFO: accumulated word occurrence stats for 10 virtual documents
--------------------- >> end captured logging << ---------------------
```
3 errors connected with encoding from [#1402](https://github.com/RaRe-Technologies/gensim/pull/1402), 
python3 only

```
======================================================================
ERROR: `Dictionary` can be loaded from textfile.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python35-x64\lib\site-packages\gensim\test\test_corpora_dictionary.py", line 234, in test_loadFromText
    d = Dictionary.load_from_text(tmpf)
  File "C:\Python35-x64\lib\site-packages\gensim\corpora\dictionary.py", line 358, in load_from_text
    line = utils.to_unicode(line)
  File "C:\Python35-x64\lib\site-packages\gensim\utils.py", line 235, in any2unicode
[...]

PR #1449, opened 2017-06-25
#1441: Fix issues with `WordOccurenceAccumulator` on Windows.

macks22:
Should resolve the first part of #1441.

For some reason, the "view" of the numpy array `_uniq_words` was not being updated on Windows, so I just rebuilt it each time it needed to be used instead of caching it once on init. If anyone can suggest an alternative approach that avoids this, I think that might be a more optimal solution. However, I doubt this added overhead is significant.


menshikh-iv:
Great @macks22, thank you for PR :+1:

PR #1456, opened 2017-06-30
Checkout to docker PR commit of gensim in dockerfile

parulsethi:


tmylk:
please add docker hub instructions to readme

menshikh-iv:
Maybe we should create RaRe account in dockerhub and load docker image, what do you think @piskvorky?

piskvorky:
I think so, yes. Don't we already have an account there @0ri0nX?

PR #1463, opened 2017-07-05
Fix for `TextDirectoryCorpus` test; windows path compatibility.

macks22:
@menshikh-iv -- short PR with fix for windows test failure, as requested. I have not tested it on Windows yet, but I believe this will resolve the issue. Thanks for the SO link!

### DIFF
--- a/gensim/test/test_corpora.py
+++ b/gensim/test/test_corpora.py
@@ -525,6 +525,7 @@ class TestTextDirectoryCorpus(unittest.TestCase):
             'b_folder/3.txt',
             'b_folder/c_folder/4.txt'
         ]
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Yes, that enough for fix issue, thank @macks22 for fast reaction :1st_place_medal:

gensim/test/test_corpora.py
```diff
@@ -525,6 +525,7 @@ def test_non_trivial_structure(self):
             'b_folder/3.txt',
             'b_folder/c_folder/4.txt'
         ]
+        expected = [os.path.normpath(path) for path in expected]
```
piskvorky: Add a comment for why this line is here please.

Issue #1363, opened 2017-05-24
models.Phrases only supports word2vec paper's method of scoring potential n-grams


michaelwsherman:
Based on [this mailing list discussion](https://groups.google.com/forum/#!topic/gensim/6ftKTlIGwJo), it seems desirable if models.Phrases supported alternative methods of scoring potential n-grams instead of forcing the use of the metric in [the original word2vec paper](https://papers.nips.cc/paper/5021-distributed-representations-of-words-and-phrases-and-their-compositionality.pdf) (Section 4).

I have some changes I've made to models.Phrases that support an optional scoring function as an instance variable, which is then used in place of the default scoring. This is a relatively easy change to make, since it just replaces the score calculation in models.Phrases.export_phrases with a call to the scoring function.

I've also implemented [normalized pointwise mutual information](https://papers.nips.cc/paper/5021-distributed-representations-of-words-and-phrases-and-their-compositionality.pdf) as an alternative scoring method in models.Phrases.export_phrases. It does, however, require an additional instance variable to track the corpus size, which is incremented as part of add_vocab.

If these changes are desirable, let me know and I'll submit a pull request. Otherwise, thanks for reading and close away :).

gojomo:
Pluggable scoring sounds interesting and useful. The docs and refactoring of Phrases/etc would have to be handled carefully to avoid user confusion or a drag on performance for people using the classic scoring. 

I believe there may be a Google Summer-of-Code project to make improvements to gensim's Phrases support – which may be focused on optimizations but might be able to consider other features/refactoring. 

A well-designed, conscientious refactoring might also help resolve the open question on #1263/#1258 – whether that functionality should be parameterized in the main class, or in a clearly distinct class. (Something like: swappable creators of candidate-phrases, and swappable scorers.)

@tmylk should comment, but I'd think a PR could be useful even if it might take a while to integrate/reconcile into the main package. (It could be a useful model to others even before any full merge.)

piskvorky:
One thing to look out for: objects must be serializable, so the "injected function as a parameter" must be named (pickle cannot handle lambdas).

menshikh-iv:
Maybe create PMI & NPMI as methods in Phrases and pass a string with a name to constructor (and match it to the method in `__init__`). This is less flexible, but avoid serialization problems

michaelwsherman:
[...]

PR #1464, opened 2017-07-05
models.Phrases multiple scoring methods (#1363)

michaelwsherman:
First attempt at alternative scoring methods, with tests, based on issue #1363 

Currently uses a string to specify the scoring. That can be changed, but I'd like some feedback first. 

Different scoring methods require different inputs, for example the default scoring requires the min_count setting and the length of the vocabulary, and npmi requires a count of all the words in the corpus (which is now counted in learn_vocab, and presumably slows learn_vocab down a tiny bit).

My main argument against pluggable scoring functions is that since different scoring methods require different corpus-based or settings-based constants, all pluggable functions would have to take all these parameters even if they weren't used. Further, if you wanted a scoring function that used a different corpus-based constant, you'd have to implement the determination/counting of that constant in somewhere like learn_vocab. New corpus based constants could also cause backwards compatibility issues if the become part of the standard scoring function call (although there might be a clever way around this by requiring all scoring functions to inherit from a parent function, which could then be updated to fill in missing parameters in existing scoring functions as new ones are created). See the remarks in the code for more about how pluggable functions might work.

My second argument against pluggable functions is that the current implementation as a string setting is more straightforward, and that it might be overkill to create cool pluggable scoring functionality that may never/rarely get used. And even if there is demand for more scoring options later, after a few other ad-hoc scoring functions are implemented the best design of pluggable scoring functions will be more obvious. Backwards compatibility could easily be kept by creating a new scoring_function input to phrases that overrides scorer. 

It's very possible I'm overthinking this and pluggable scoring isn't so complicated. I'm happy to implement something if the feeling is there's a clear best way to do it.


gensim/models/phrases.py
```diff
@@ -226,31 +252,47 @@ def export_phrases(self, sentences, out_delimiter=b' ', as_tuples=False):
...
-                    if bigram_word in vocab and not last_bigram:
-                        pa = float(vocab[word_a])
-                        pb = float(vocab[word_b])
-                        pab = float(vocab[bigram_word])
-                        score = (pab - min_count) / pa / pb * len(vocab)
+                    if bigram_word in vocab:
+                        count_a = float(vocab[word_a])
+                        count_b = float(vocab[word_b])
+                        count_ab = float(vocab[bigram_word])
+                        score = scoring_function(count_a, count_b, count_ab)
```
michaelwsherman: A pluggable scoring function would have to be called with all corpus constants and Phrases settings used in any scoring function. Right now that would look like:
`score = scoring_function(count_a, count_b, count_ab, min_count, len_vocab, corpus_word_count)`.
And the call would grow as the universe of variables considered by all scoring functions grows.

menshikh-iv:
Looks good, thank you @michaelwsherman :+1:

gensim/models/phrases.py
```diff
@@ -120,10 +123,9 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
         `min_count` ignore all words and bigrams with total collected count lower
         than this.
 
-        `threshold` represents a threshold for forming the phrases (higher means
-        fewer phrases). A phrase of words `a` and `b` is accepted if
-        `(cnt(a, b) - min_count) * N / (cnt(a) * cnt(b)) > threshold`, where `N` is the
-        total vocabulary size.
+        `threshold` represents a score threshold for forming the phrases (higher means
+        fewer phrases). A phrase of words `a` followed by `b` is accepted if the score of the
+        phrase is greater than threshold. see the `scoring' setting
```
piskvorky: Capitalize first word in sentence, end in full stop.

gensim/models/phrases.py
```diff
@@ -197,8 +221,10 @@ def add_vocab(self, sentences):
         # directly, but gives the new sentences a fighting chance to collect
         # sufficient counts, before being pruned out by the (large) accummulated
         # counts collected in previous learn_vocab runs.
-        min_reduce, vocab = self.learn_vocab(sentences, self.max_vocab_size, self.delimiter, self.progress_per)
+        min_reduce, vocab, total_words = \
+        self.learn_vocab(sentences, self.max_vocab_size, self.delimiter, self.progress_per)
```
piskvorky: Code style: bad indentation (unneeded line break).

gensim/models/phrases.py
```diff
@@ -226,31 +252,47 @@ def export_phrases(self, sentences, out_delimiter=b' ', as_tuples=False):
...
+        vocab = self.vocab
+        threshold = self.threshold
+        delimiter = self.delimiter  # delimiter used for lookup
+        min_count = self.min_count
+        scoring = self.scoring
+        corpus_word_count = self.corpus_word_count
+
+        if scoring == 'default':
+            scoring_function = \
+            partial(self.original_scorer, len_vocab=float(len(vocab)), min_count=float(min_count))
```
piskvorky: Indentation (unneeded line break).

gensim/models/phrases.py
```diff
@@ -226,31 +252,47 @@ def export_phrases(self, sentences, out_delimiter=b' ', as_tuples=False):
...
+        elif scoring == 'npmi':
+            scoring_function = \
+            partial(self.npmi_scorer, corpus_word_count=corpus_word_count)
```
piskvorky: Indentation (unneeded line break).

gensim/models/phrases.py
piskvorky: I think that's still preferable. This string-passing is inflexible.

We could support some common use-cases by passing a string, but the code underneath should simply translate that string into a `scoring_function` and work with that underneath. Custom `scoring_function`s should be supported IMO.

gensim/models/phrases.py
```diff
@@ -226,31 +252,47 @@ def export_phrases(self, sentences, out_delimiter=b' ', as_tuples=False):
...
+                        # it would not be enforced otherwise
+                        if score > threshold and count_ab >= min_count:
                             if as_tuples:
                                 yield ((word_a, word_b), score)
                             else:
                                 yield (out_delimiter.join((word_a, word_b)), score)
                             last_bigram = True
                             continue
-                    last_bigram = False
+                last_bigram = False
```
piskvorky: Is this on purpose? What is this change about?

gensim/models/phrases.py
```diff
@@ -311,6 +353,20 @@ def __getitem__(self, sentence):
 
         return [utils.to_unicode(w) for w in new_s]
 
+    # calculation of score based on original mikolov word2vec paper
+    # len_vocab and min_count set so functools.partial works
+    @staticmethod
+    def original_scorer(worda_count, wordb_count, bigram_count, len_vocab=0.0, min_count=0.0):
+        return (bigram_count - min_count) / worda_count / wordb_count * len_vocab
```
piskvorky: Beware of integer divisions - this code is brittle.

gensim/models/phrases.py
```diff
@@ -311,6 +353,20 @@ def __getitem__(self, sentence):
...
+
+    # normalized PMI, requires corpus size
+    @staticmethod
+    def npmi_scorer(worda_count, wordb_count, bigram_count, corpus_word_count=0.0):
+        pa = worda_count / corpus_word_count
```
piskvorky: Is this meant to be an integer division?

michaelwsherman:
@piskvorky Thank you for your comments. I'll go though these and respond with some specifics sometime next week, am on vacation now.

piskvorky:
Sounds good. Enjoy your vacation :)

gensim/models/phrases.py
michaelwsherman: What's the number of columns we cap at? I thought it was 100, which I believe this exceeded.

gensim/models/phrases.py
michaelwsherman: I will make this change, hopefully before the end of the week, and make it part of a PR.

gensim/models/phrases.py
michaelwsherman: Yes, this is on purpose. Matches up to line 277. If that test fails we have to set `last_bigram` to false. This positioning sets it to false always--the only time it gets set to true is in line 293 when a passing bigram is found.

gensim/models/phrases.py
piskvorky: There's no hard limit; if the line becomes hard to read, we break it.

If the break would be even harder to read than the original, we don't break it.

Line continuations are indented at one extra level (4 spaces to the right).

gensim/models/phrases.py
piskvorky: Aha, so this is a bug fix at the same time. Thanks! CC @menshikh-iv

michaelwsherman:
Question @piskvorky -- what's the best way to make these changes (and to the other PR)? Submit another PR? Or Is there a way to update this PR even though it has already been merged?

menshikh-iv:
@michaelwsherman Please submit an another PR.

michaelwsherman:
I haven't forgotten about this, just really swamped right now at work. Do you have an expected date of next release? I'll try to do my best to get these fixes (and the fixes from #1423 ) in a new PR before that.

Sorry.

menshikh-iv:
No problem @michaelwsherman, I think next release will be in the first/second week of September.

michaelwsherman:
Fixes from @piskvorky in PR #1573 .

gensim/models/phrases.py
michaelwsherman: I didn't fix this in PR #1573 . Rather, I just cast everything before calling the scoring method in Phrases and Phraser. I think that's the better place to do the casting since then it fixes the problem for all custom scorers as well. 

Of course, I can do the casting in the scoring methods as well. Let me know if you still think I need it here and in `npmi_scorer` and I'll update PR #1573. It's extra steps, but I'd assume the performance hit is infinitesimal.

PR #1467, opened 2017-07-05
update score_sentence_cbow

jmhessel:
I don't think the learning rate parameter belongs in the argument list (unless I'm mistaken). Also, the cython version of this function doesn't have this argument.


menshikh-iv:
You are right, thank you @jmhessel, congratulations on your first PR:1st_place_medal:

PR #1468, opened 2017-07-05
score_cbow_pair unused argument

jmhessel:
I don't think the word indices in the cbow scoring function are actually used (unless I'm mistaken)


menshikh-iv:
Thank you @jmhessel

PR #1470, opened 2017-07-06
Downgrade scipy/numpy requirements. Fix #1450

### LINKED ISSUE #1450: Minimum version for numpy/scipy
In the last release, I had to "bump" versions for numpy/scipy. This may be a problem for *ops people (need to upgrade old versions). 

For this reason, we should downgrade numpy/scipy version and don't break anything.

menshikh-iv:
We should to minimize numpy/scipy version
For test/build we need numpy/scipy wheels, 
Minimal wheels version for linux from PyPI

| *   | numpy  | scipy  |
|-----|--------|--------|
| 2.7 | 1.6.0  | 0.9.0  |
| 3.5 | 1.9.0  | 0.16.0 |
| 3.6 | 1.11.3 | 0.18.1 |

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,8 @@ before_install:
   - export PATH=/home/travis/miniconda2/bin:$PATH
   - conda update --yes conda
 install:
-  - conda create --yes -n gensim-test python=$TRAVIS_PYTHON_VERSION pip atlas numpy scipy
+  - conda create --yes -n gensim-test python=$TRAVIS_PYTHON_VERSION pip atlas numpy==1.11.3 scipy==0.18.1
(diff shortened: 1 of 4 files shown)


menshikh-iv:
We can use different versions of the dependencies depending on the python version, but It will bring pain and suffering.

As more people move to 3.6 from 2.7 + we want to build gensim wheels for all main python versions (2.7, 3.5 and 3.6), for this reasons I'll use versions that anywhere available (i.e. from 3.6).

You agree with me @piskvorky?

piskvorky:
Up to you @menshikh-iv , I have no preference either way.

PR #1473, opened 2017-07-06
Update sklearn API for Gensim models

chinmayapancholi13:
This PR does the following things: 
- Renames scikit-learn wrapper classes for Gensim models as suggested [here](https://github.com/RaRe-Technologies/gensim/pull/1437/#discussion_r123593796).
- Updates existing API by removing the base sklearn-API class and `set_params()` & `get_params()` functions.
- Updates `transform` implementation by making use of `sparse2full` function.
- Adds `testConsistencyWithGensimModel` test for LdaModel.
- Updates ipynb corresponding to these changes.


gensim/test/test_sklearn_integration.py
```diff
@@ -15,12 +15,12 @@
...
-from gensim.sklearn_integration.sklearn_wrapper_gensim_lsimodel import SklLsiModel
-from gensim.sklearn_integration.sklearn_wrapper_gensim_ldaseqmodel import SklLdaSeqModel
-from gensim.sklearn_integration.sklearn_wrapper_gensim_w2vmodel import SklW2VModel
-from gensim.sklearn_integration.sklearn_wrapper_gensim_atmodel import SklATModel
+from gensim.sklearn_integration.sklearn_wrapper_gensim_rpmodel import RpTransformer
+from gensim.sklearn_integration.sklearn_wrapper_gensim_ldamodel import LdaTransformer
+from gensim.sklearn_integration.sklearn_wrapper_gensim_lsimodel import LsiTransformer
+from gensim.sklearn_integration.sklearn_wrapper_gensim_ldaseqmodel import LdaSeqTransformer
+from gensim.sklearn_integration.sklearn_wrapper_gensim_w2vmodel import W2VTransformer
+from gensim.sklearn_integration.sklearn_wrapper_gensim_atmodel import AuthorTopicTransformer
```
gojomo: As long as we're making breaking renames/refactors for clarity/concision, these package-names have a lot of redundancy, too. `gensim.sklearn_integration.sklearn_wrapper_gensim_atmodel` repeats both `sklearn` and `gensim`, and `integration` and `wrapper` have overlapping meanings. 

So perhaps just: `gensim.sklearn_wrappers.atmodel`? And that's if `at` is already easily recognized from use elsewhere as abbreviation; otherwise `gensim.sklearn_wrappers.authortopicmodel` or even `gensim.sklearn_wrappers.authortopic`?

Or maybe even:, given that these wrappers are each pretty slight-in-size, they could either go: (1) in the respective associated model file – that is W2VTransformer goes into `gensim.models.word2vec` (and probably loses its abbreviation, to match the non-abbreviation used in that file's classes); or (2) into a single `sklearn_wrappers` file, alongside the `BaseSklearnWrapper`?

(And regarding `BaseSklearnWrapper` - not sure it adds much beyond alternative of using skearn's own BaseEstimator and TransformerMixin. It forces more rigor in overriding the necessary abstract-methods, which sklearn itself never does, but has less functionality in `set_params()` and other aspects of introspectability.)

gojomo: cc @piskvorky for overall library naming/organization priorities

tmylk: I vote for `from gensim.models.atmodel import AuthorTopicTransformer`, but would agree with  `from gensim.sklearn_api import AuthorTopicTransformer` too. It is neither a wrapper nor an integration, but just an API.

piskvorky: If we were to share the same module for gensim/sklearn classes, we'd have to tiptoe around the sklearn imports, because sklearn is not a dependency of gensim.

An extra subpackage sounds cleaner to me: one subpackage for sklearn / keras / spark / whatever API. Not imported automatically from gensim `__init__`, so that users need to import it explicitly (after installing sklearn/keras/spark/whatever), with `import gensim.sklearn_api`.

And if the subpackage becomes too unwieldy / complex (not the case with sklearn now), a separate library would make sense to me too.

menshikh-iv: I vote for `from gensim.sklearn_api import AuthorTopicTransformer` with subpackage `sklearn_api`.
We will not have "sklearn as dependency", but have "short and uniq import path" for sklearn wrappers.

gensim/sklearn_api/basemodel.py
```diff
@@ -11,7 +11,7 @@
 from abc import ABCMeta, abstractmethod
 
 
-class BaseSklearnWrapper(object):
+class BaseTransformer(object):
```
gojomo: With this refactoring... I'm again wondering, is this class even helpful? Its `set_params()` does less than the implementation that would be inherited from sklearn's own `BaseEstimator`. The other methods just serve to enforce that any subclass override certain methods – a rigor that is possibly useful, but that sklearn does not, itself, impose within its own superclass-hierarchy.

chinmayapancholi13: @gojomo scikit-learn indeed does not impose such a hierarchy among its classes. However, as you stated above, in addition to not having to implement the `set_params()` every time, we know which functions are to be implemented for each model. Also, if in the future, we decide to refine/improve the `set_params()` function and make it more sophisticated (e.g. as in sklearn's `BaseEstimator`), we would only have to do it once in the base class rather than having to make the same change in the sklearn-api class of each Gensim model. Hence, having an abstract base class and deriving classes for Gensim models from it would help us in writing new sklearn-api classes in the future as well.

gojomo: But, the existing `set_params()` looks worse than just doing nothing (and inheriting from `BaseEstimator` as all current subclasses of this class already do) – unless I'm missing something. So there's no need to plan for a future where "we decide to refine/improve" it, and do such improvements in the base class. We can make that decision now, and delete 8 lines of code from the base class, and all the subclasses automatically get the better & more standard (for sklearn) implementation of `set_params()`. 

And then, the only benefit of this shared superclass is forcing an error as a reminder of what to implement. But if sklearn doesn't do that itself, is it that important to do so? It's also over-enforcing – it is not a strict requirement that all transformers support `partial_fit()` - not every algorithm will necessarily support that.

tmylk: In my head I expect all sklearn models to have `partial_fit`, `fit`, `transform` and be used in a pipeline. We constantly get questions on the mailing list about `how to update model X` for almost every sklearn algo. Raising a `NotImplemented ` with an explanation for word2vec and doc2vec `use gensim api update function but it's not recommended as it hasn't been researched.`  And throw an explicitly that RandomProj can't do it because of algo limitations.

gojomo: OK, providing a place to include a error/warning is a potential reason to implement `partial_fit()` even where the underlying algorithm can't (or shouldn't casually) be used that way. BUT, simply not implementing a method is also a pretty good way of signalling that something isn't implemented – and indeed that seems to be what scikit-learn itself does. For that reason, it's even possible other code will use introspection about whether an sklearn pipeline object implements `partial_fit()` as a test of support, in which case having an implemented-method that just errors could be suboptimal. 

But with `set_params()`, again, why have any lesser-featured implementation here, when every subclass (so far) inherits from scikit-learn's own `BaseEstimator`, which provides a more fully-featured implementation? Why wouldn't we want fewer limitations (and match the source library's behavior) for zero lines of code, rather than more limitations with more code?

oxymor0n: hi,

jumping in the conversation a little late here, but I do agree that the old `BaseSklearnWrapper` does very little since

 - the default `get_params` and `set_params` from sklearn's `BaseEstimator` are very reasonable and should only be overriden when necessary; if all the concrete wrapper classes inherit from `BaseEstimator` (as they do now) then there's no need to force them to re-implement these methods
 - @tmylk `partial_fit` is absolutely not a requirement for a class to be compatible with sklearn `Pipeline` since lots of algorithm are not online in nature. If there is a particular algorithm that most people are confused about, then the warning should be raised for that class only.

tmylk: I would struggle to find a gensim algorithm for which I haven't seen a "is it possible to update the model post-training?" question on the mailing list...

gojomo: Right, it's always asked, but not always supported. The scikit-learn convention for not supporting incremental training is to simply not implement `partial_fit()`. Why not adopt the same convention for scikit-learn compatible classes here?

gensim/sklearn_api/w2vmodel.py
```diff
@@ -83,4 +83,5 @@ def transform(self, words):
         return np.reshape(np.array(X), (len(words), self.size))
 
     def partial_fit(self, X):
-        raise NotImplementedError("'partial_fit' has not been implemented for W2VTransformer")
+        raise NotImplementedError("'partial_fit' has not been implemented for W2VTransformer since 'update()' function for Word2Vec class is experimental. "
```
piskvorky: Is this correct? Why would the `update()` method of word2vec be experimental?

menshikh-iv: Yes, it's correct.

piskvorky: I don't even see any `update()` function. Can you elaborate?

chinmayapancholi13: @piskvorky The original error message was not correct (since there is no `update()` function in `Word2Vec` class as you correctly pointed out) and I have updated the message now. My apologies for the confusion.
This change was made in reference to [this comment](https://github.com/RaRe-Technologies/gensim/pull/1437#discussion_r123597260) by @gojomo in one of the older PRs.

piskvorky: @chinmayapancholi13 thanks for the clarification.

So the concern is about initializing the vocabulary. That's a separate concern to incremental training, which is fully supported and not experimental (in fact, it's the same code as non-incremental training).

I don't think `partial_fit` is a high priority, but at the same time, users ask for it all the time, especially in combination with incremental vocab updates (which are indeed experimental). 

So I see two directions (both non-critical, nice-to-have) here:

1. support incremental training with a fixed vocabulary
  * implies finding a natural way to initialize vocabulary before hand
  * implies finding a natural way to control the learning rate (alpha) during the incremental calls.
2. support fully online training, incl. updating the vocabulary incrementally
  * implies changing the w2v algo(s) to support this (hashing trick with fixed hash space?)
  * still implies finding a way to control alpha

chinmayapancholi13: Ah, I see. So the issue is actually "updating the vocab" (and not "training") being experimental. I have updated the error message in `partial_fit` for `W2VTransformer` for this accordingly.
Thanks for your explanation and suggestions about resolving this concern. I guess it would be nice to create an issue for this. :)

gensim/sklearn_api/basemodel.py
piskvorky: I agree with @gojomo 's point of view. The point of these adapters is to assume the "philosophy" of another package, whatever structure or idiosyncrasies that may have. The least amount of surprises and novelty for users of that package.

It is not the place of the adapters to introduce their own philosophy on class structures or missing methods or whatever.

gensim/sklearn_api/w2vmodel.py
```diff
@@ -101,4 +83,4 @@ def transform(self, words):
         return np.reshape(np.array(X), (len(words), self.size))
 
     def partial_fit(self, X):
-        raise NotImplementedError("'partial_fit' has not been implemented for SklW2VModel")
+        raise NotImplementedError("'partial_fit' has not been implemented for W2VTransformer since updating vocabulary incrementally for Word2Vec model is experimental.")
```
tmylk: "However the model can be updated with a fix vocab using Gensim API call"

gensim/test/test_sklearn_api.py
```diff
@@ -295,19 +292,6 @@ def testTransform(self):
         self.assertEqual(transformed_vecs.shape[0], 1)
         self.assertEqual(transformed_vecs.shape[1], self.model.num_topics)
 
-    def testSetGetParams(self):
```
tmylk: Please keep these tests to check that this still works

gensim/sklearn_api/w2vmodel.py
chinmayapancholi13: Done

gensim/sklearn_api/basemodel.py
chinmayapancholi13: I have removed the `BaseTransformer` class now. Thanks for the useful feedback and suggestions. :)

gensim/test/test_sklearn_api.py
chinmayapancholi13: Done.

tmylk:
@menshikh-iv LGTM please

gensim/test/test_sklearn_api.py
tmylk: Looks good. Thanks

gensim/sklearn_api/__init__.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2011 Radim Rehurek <radimrehurek@seznam.cz>
```
piskvorky: @chinmayapancholi13 please add yourself as the author. This is your baby :)

gensim/sklearn_api/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+"""Scikit learn wrapper for gensim.
+Contains various gensim based implementations which match with scikit-learn standards.
+See [1] for complete set of conventions.
+[1] http://scikit-learn.org/stable/developers/
+"""
+
+
+from .ldamodel import LdaTransformer  # noqa: F401
```
piskvorky: What is all that `noqa: F401` for? Is it really necessary?

docs/notebooks/sklearn_api.ipynb
```diff
@@ -0,0 +1,562 @@
...
+   "source": [
+    "This tutorial is about using gensim models as a part of your scikit learn workflow with the help of wrappers found at ```gensim.sklearn_integration```"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The wrappers available (as of now) are :\n",
+    "* LdaModel (```gensim.sklearn_api.ldaModel.SklLdaModel```),which implements gensim's ```LDA Model``` in a scikit-learn interface\n",
```
piskvorky: Haven't these been updated?

gensim/sklearn_api/__init__.py
chinmayapancholi13: A line having `noqa` doesn't issue flake8 errors. Writing `noqa: F401` saves us from flake8 errors saying that we have imports (e.g. `LdaTransformer`) which we haven't used anywhere in the file.

gensim/sklearn_api/__init__.py
chinmayapancholi13: Hehe. Ok! :smile:

gensim/sklearn_api/__init__.py
piskvorky: It feels wrong to litter the gensim code with such constructs -- the gensim code is correct, this is essentially working around some idiosyncrasy (bug?) of an unrelated library.

By the way, how come we don't these errors from all the other `__init__` imports in gensim? Or do we? CC @menshikh-iv

menshikh-iv: @piskvorky because flake8 analyze the only diff every time, if you change same lines in any other `__init__` file you will get same error.

docs/notebooks/sklearn_api.ipynb
chinmayapancholi13: Yes. I had changed the transformer names in the code in the ipynb but missed updating it here. Thanks a lot for pointing this out. :)

gensim/sklearn_api/__init__.py
piskvorky: OK, so the warnings will go away immediately? Why did we even add these comments then.

tmylk:
@menshikh-iv May I ask for a date of a new release with this new API?

menshikh-iv:
@tmylk ~ first half of September

Issue #1477, opened 2017-07-09
Restructure TextCorpus code to share multiprocessing and preprocessing logic.


macks22:
#### Description
This is a follow-up from a conversation that came up around the addition of the `TextDirectoryCorpus` (see #1387). As part of discussion around that ticket, @piskvorky mentioned that there is not a clear plan for the `textcorpus` code. It was originally meant to just be example code for others to work from. However, it seems many people (and gensim tutorial writers) have need of common text processing corpora.  I did some analysis (discussed in #1387) and found that there are several text corpora classes scattered in various places throughout the code. These include the `TextCorpus`, `TextDirectoryCorpus`, `BrownCorpus`, `WikiCorpus`, `LineSentence`, and `Text8Corpus`. 

I propose combining the shared logic of these various corpora into the `textcorpus` module. The `WikiCorpus` can continue to live in the `wikicorpus` module, since its preprocessing is so specific to the wiki markup. However, the others should be moved to the `textcorpus` module. Once #1453 is completed, the `BrownCorpus` would be a good candidate to move to some sort of `datasets` subpackage. But at least for now, `textcorpus` is a better home for it than the `word2vec` module where it currently resides.

piskvorky:
Makes sense, thanks!

To me, it's more about the sharing logic and making things modular and easier to discover, rather than where the implementation lives (module). We should import all the classes from `corpora.__init__` anyway (so `from gensim.corpora import WikiCorpus, LineSentence, BrownCorpus...` works). The particular module name with the actual implementation is not that important, except to point to as a blueprint example for people's own custom extensions.

PR #1478, opened 2017-07-09
Restructure TextCorpus code to share multiprocessing and preprocessing logic.

macks22:
Implements #1477.


macks22:
@piskvorky @menshikh-iv I believe the build failures on this only have to do with importing in the main guard in the `test_word2vec` module. This appears to be for a good reason based on the comment there. Is there anything else that should be changed for this PR? Thanks!

macks22:
I tested out the WikiCorpus before and after on the full wikipedia corpus. I did this by building the Dictionary. My goal was to ensure the speed is comparable now to what was implemented before.

* Before: built Dictionary(2012943 unique tokens: [u'tripolitan', u'ftdna', u'fi\u0250', u'soestdijk', u'farmobil']...) from 4265002 documents (total 2338950602 corpus positions)
    * 206m26.452s
* After: built Dictionary(2014666 unique tokens: [u'tripolitan', u'ftdna', u'fi\u0250', u'soestdijk', u'farmobil']...) from 3765193 documents (total 1379989194 corpus positions)
    * 211m35.043s

The speed is comparable. This implementation also performs deaccenting and stopword removal, which I suspect is why it takes a few minutes longer. The difference in number of documents comes from the removal of stopwords, which results in many more empty documents which are pruned. I think the difference in number of terms is due to the Dictionary pruning encountering different documents.

gensim/corpora/text_processing_pool.py
```diff
@@ -0,0 +1,160 @@
...
+    """Generic process whose target function processes batches of texts."""
+
+    def __init__(self, group=None, target=None, name=None, args=(), kwargs={},
+                 state_kwargs=None):
+        super(TextProcessor, self).__init__(group, target, name, args, kwargs)
+        state_kwargs = {} if state_kwargs is None else dict(state_kwargs)
+        self.init_state(state_kwargs)
+
+    def init_state(self, state_kwargs):
+        for name, value in state_kwargs.items():
```
piskvorky: Why not simply `self.__dict__.update`?

gensim/corpora/text_processing_pool.py
```diff
@@ -0,0 +1,160 @@
+import multiprocessing as mp
```
piskvorky: Missing file header.

gensim/corpora/textcorpus.py
```diff
@@ -409,66 +516,130 @@ def getstream(self):
...
+        kwargs['tokenizer'] = kwargs.get('tokenizer', unicode_and_tokenize)
+        kwargs['character_filters'] = kwargs.get('character_filters', [])
+        TextCorpus.__init__(self, source, **kwargs)
+
+    def _sentence_token_stream(self):
+        # Entire corpus is one gigantic line -- there are no sentence marks at all.
+        # So just split the token sequence arbitrarily into sentences of length
+        # `max_sentence_length`.
+        sentence, rest = [], b''
+        with utils.smart_open(self.source) as fin:
```
piskvorky: Best to open files in binary mode (`rb`), and convert to text explicitly where needed.

gensim/corpora/textcorpus.py
```diff
@@ -409,66 +516,130 @@ def getstream(self):
...
+                text = rest + fin.read(self.chunksize)  # avoid loading the entire file (=1 line) into RAM
+                if text == rest:  # EOF
+                    words = text.split()
+                    sentence.extend(words)  # return the last chunk of words, too (may be shorter/longer)
+                    if sentence:
+                        yield sentence
+                    break
+
+                last_token = text.rfind(b' ')  # last token may have been split in two... keep for next iteration
+                words, rest = (text[:last_token].split(),
```
piskvorky: No vertical indent -- please use hanging indent.

gensim/corpora/wikicorpus.py
```diff
@@ -278,21 +257,38 @@ def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), diction
...
-        if dictionary is None:
-            self.dictionary = Dictionary(self.get_texts())
-        else:
-            self.dictionary = dictionary
+        tokenizer = utils.lemmatize if lemmatize else tokenize
+        if character_filters is None:
+            # no need to lowercase and unicode, because the tokenizer already does that.
+            character_filters = [textcorpus.deaccent, textcorpus.strip_multiple_whitespaces]
+        super(WikiCorpus, self).__init__(source, dictionary, metadata, character_filters, tokenizer,
+                                         token_filters, processes)
```
piskvorky: No vertical indent.

gensim/test/test_corpora.py
```diff
@@ -20,7 +19,7 @@
 import numpy as np
 
 from gensim.corpora import (bleicorpus, mmcorpus, lowcorpus, svmlightcorpus,
-                            ucicorpus, malletcorpus, textcorpus, indexedcorpus)
+                            ucicorpus, malletcorpus, indexedcorpus)
```
piskvorky: No vertical indent please.

gensim/test/test_textcorpus.py
```diff
@@ -0,0 +1,303 @@
...
+    ['trees'],
+    ['graph', 'trees'],
+    ['graph', 'minors', 'trees'],
+    ['graph', 'minors', 'survey']
+]
+
+
+def test_texts_file():
+    fpath = os.path.join(tempfile.gettempdir(), 'gensim_corpus.tst')
+    with open(fpath, 'w') as f:
```
piskvorky: `smart_open` + binary mode please.

gensim/test/test_textcorpus.py
```diff
@@ -0,0 +1,303 @@
...
+            ['tooth', 'spaces']
+        ]
+
+        corpus = self.corpus_from_lines(lines)
+        texts = list(corpus.get_texts())
+        self.assertEqual(expected, texts)
+
+    def corpus_from_lines(self, lines):
+        fpath = tempfile.mktemp()
+        with codecs.open(fpath, 'w', encoding='utf8') as f:
```
piskvorky: Dtto.

gensim/utils.py
```diff
@@ -1263,3 +1269,45 @@ def _iter_windows(document, window_size, copy=False, ignore_below_size=True):
     else:
         for doc_window in doc_windows:
             yield doc_window.copy() if copy else doc_window
+
+
+def walk_with_depth(top, topdown=True, onerror=None, followlinks=False, depth=0):
```
piskvorky: Is this really needed? The depth can be deduced easily from normal `walk()`, by comparing the root directories.

piskvorky:
This looks like a massive PR; are the changes 100% backward compatible?

If not, what is the upgrade plan for users = how do they modify their existing code so it continues to work?

gensim/corpora/text_processing_pool.py
```diff
@@ -0,0 +1,160 @@
...
+                util.debug("Possible encoding error while sending result: %s" % (
+                    wrapped))
+                put((job, i, (False, wrapped)))
+
+            task = job = result = func = args = kwds = None
+            completed += 1
+        util.debug('worker exiting after %d tasks' % completed)
+
+
+class _PatchedPool(mp.pool.Pool):
```
piskvorky: What is this for, what is being patched (and why)?

Needs a clear comment.

gensim/corpora/text_processing_pool.py
```diff
@@ -0,0 +1,160 @@
...
+        self._processor_class = processor_class
+        super(_PatchedPool, self).__init__(processes, initializer, initargs, maxtasksperchild)
+
+    def _repopulate_pool(self):
+        """Bring the number of pool processes up to the specified number,
+        for use after reaping workers which have exited.
+        """
+        for i in range(self._processes - len(self._pool)):
+            w = self.Process(args=(self._inqueue, self._outqueue,
+                                   self._initializer,
```
piskvorky: No vertical indent in gensim.

gensim/corpora/text_processing_pool.py
```diff
@@ -0,0 +1,160 @@
...
+                                   self._initargs, self._maxtasksperchild))
+            self._pool.append(w)
+            w.name = w.name.replace('Process', 'PoolWorker')
+            w.daemon = True
+            w.start()
+            util.debug('added worker')
+
+
+class TextProcessingPool(object):
```
piskvorky: What is this wrapper for? Why not use the default `Pool`?

I'd prefer to stick to built-ins, unless absolutely necessary. And if absolutely necessary, will need a better documentation describing the rationale.

gensim/corpora/text_processing_pool.py
macks22: added file header

gensim/corpora/text_processing_pool.py
macks22: good suggestion; done

gensim/corpora/textcorpus.py
macks22: The default is 'rb', but I updated to set mode explicitly to future-proof.

gensim/corpora/textcorpus.py
macks22: done

gensim/corpora/wikicorpus.py
macks22: done

gensim/test/test_corpora.py
macks22: done

gensim/test/test_textcorpus.py
macks22: this function was actually not being used, so I just removed it

gensim/test/test_textcorpus.py
macks22: changed mode to 'wb'

gensim/utils.py
macks22: This is a very good point; I replaced this with a wrapper on `os.walk` that just deduces the depth in the manner you suggested.

gensim/corpora/text_processing_pool.py
macks22: I've added documentation throughout this module to clarify. Some added context: when I initially implemented this refactor, I was serializing the `token_filters`, `tokenizer`, and `character_filters` used in `TextCorpus` for text preprocessing. This pickling overhead was causing a significant slowdown. So I wanted to include them in each worker process at startup to speed it up. Doing so ruled out the use of the builtin `multiprocessing.Pool` class. 

Rather than write a complicated custom pool, I decided that reuse via patching of the existing pool would be more robust and probably useful elsewhere in the code (for instance, in the `text_analysis` module used by the `probability_estimation` module). That is why this module came about.

gensim/corpora/text_processing_pool.py
macks22: done

gensim/corpora/text_processing_pool.py
macks22: Added documentation with rationale. See comment on `_PatchedPool` above for additional context.

macks22:
@piskvorky I've addressed your review comments; thank you for the quick feedback! If I can add anything else to make your review of the logic easier or otherwise clarify things, I will gladly do so.

gensim/corpora/text_processing_pool.py
piskvorky: Aha, thanks, that's interesting. @gojomo @menshikh-iv can you please review this extended multiprocessing logic?

I'm curious whether others do it this way too, since this seems a very common use-case.

gensim/utils.py
```diff
@@ -1271,43 +1271,16 @@ def _iter_windows(document, window_size, copy=False, ignore_below_size=True):
...
-
-            # Generator so besides the recursive `walk()` call, no additional cost here.
-            for x in walk_with_depth(new_path, topdown, onerror, followlinks, depth + 1):
-                yield x
-    if not topdown:
-        yield depth, top, dirs, nondirs
+    path = os.path.abspath(top)
+    for dirpath, dirnames, filenames in os.walk(path, topdown, onerror, followlinks):
+        sub_path = dirpath.replace(path, '')
+        depth = sub_path.count(os.sep)
```
piskvorky: Is this safe even with unnormalized paths (`/tmp` vs `/tmp/` etc)? How does `walk` handle symlinks?

`os.path.relpath`/`commonprefix` may be safer, I'm not sure.

macks22: docs say this on symlinks:
> By default, os.walk does not follow symbolic links to subdirectories on
> systems that support them.  In order to get this functionality, set the
> optional argument 'followlinks' to true.

I did look at both of the functions you referenced (which I was not familiar with), but I believe the current code handles unnormalized paths correctly. I've added tests to verify this.

macks22:
@piskvorky I believe this is fully backwards-compatible in terms of interfaces. The only thing I expect will be different is the default preprocessing used for the `WikiCorpus`. In particular, it is now removing stopwords and deaccenting non-ascii text. It is also removing tokens shorter than 3 characters, instead of just those shorter than 2. I expect this will be a happy change for most users, but it is also possible to achieve the old behavior by initializing with `WikiCorpus(fname, character_filters=[], token_filters=[])`.

Also, I have updated the PR to address your most recent comments; thank you for your review. I believe you'd asked for thoughts from @gojomo and @menshikh-iv regarding the modified multiprocessing pool; I'm also curious to know if the approach I took here has been used elsewhere and if any alternative approaches might be more suitable for this problem. Thanks!

gensim/utils.py
```diff
@@ -1263,3 +1269,18 @@ def _iter_windows(document, window_size, copy=False, ignore_below_size=True):
...
+    determine the depth relative to `top`. Yields in the `top` directory are at
+    depth 0.
+
+    Returns:
+        generator of tuples of (depth, dirpath, dirnames, filenames).
+    """
+    path = os.path.abspath(top)
+    for dirpath, dirnames, filenames in os.walk(path, topdown, onerror, followlinks):
+        sub_path = dirpath.replace(path, '')
+        depth = sub_path.count(os.sep)
```
piskvorky: This construct still makes me a little uneasy. Can we at least `os.path.normpath`, to get rid of any double/trailing/leading slashes? Or does `os.walk` normalize the `dirpath` somehow? Although in that case, we'd have to normalize `path` and `dirpath` in exactly the same way, so that the `.replace()` above works.

piskvorky:
@macks22 thanks!

`In particular, it is now removing stopwords and deaccenting non-ascii text. It is also removing tokens shorter than 3 characters`. Why this change?

gensim/utils.py
macks22: Ah, I see your concern. I think `os.path.abspath` handles the situation you're worried about:
```
In [4]: os.path.abspath('/test/path/')
Out[4]: '/test/path'

In [5]: os.path.abspath('/test/path')
Out[5]: '/test/path'

In [6]: os.path.abspath('/test/path//')
Out[6]: '/test/path'
```

macks22:
@piskvorky I had mainly made those changes so the preprocessing defaults would be as close to the default for the `TextCorpus` as possible. I've added a commit changing the defaults back to the way they were before. And yes, the new code does support custom tokenization, text normalization, and any other preprocessing desired by the user. The preprocessing pipeline is the same as that used for `TextCorpus`, which consists of 0+ `character_filters`, 1 `tokenizer`, and 0+ `token_filters`. Here is the relevant excerpt from the `TextCorpus.__init__` docstring:
```
character_filters (iterable of callable): each will be applied to the text of each
    document in order, and should return a single string with the modified text.
    For Python 2, the original text will not be unicode, so it may be useful to
    convert to unicode as the first character filter. The default character filters
    lowercase, convert to unicode (strict utf8), perform ASCII-folding, then collapse
    multiple whitespaces.
tokenizer (callable): takes as input the document text, preprocessed by all filters
    in `character_filters`; should return an iterable of tokens (strings).
token_filters (iterable of callable): each will be applied to the iterable of tokens
    in order, and should return another iterable of tokens. These filters can add,
    remove, or replace tokens, or do nothing at all. The default token filters
    remove tokens less than 3 characters long and remove stopwords using the list
    in `gensim.parsing.preprocessing.STOPWORDS`.
```

gensim/utils.py
piskvorky: What if `walk` hits a symlink -- does it return `dirpath` as a canonical path, or is `path` still its prefix?

piskvorky:
Nice! Should `tokenizer` return byte strings, or unicode strings?

piskvorky:
@macks22 is there a way to reach you privately (email)? Please ping me at radim@rare-technologies.com.

gensim/utils.py
macks22: `tree tmp`
```
tmp
├── subdir
│   └── test
└── symlink -> subdir

2 directories, 1 file
```
```
In [56]: list(os.walk('tmp'))
Out[56]: [('tmp', ['subdir', 'symlink'], []), ('tmp/subdir', [], ['test'])]
```
```
In [58]: list(os.walk('tmp', followlinks=True))
Out[58]:
[('tmp', ['subdir', 'symlink'], []),
 ('tmp/subdir', [], ['test']),
 ('tmp/symlink', [], ['test'])]
```

macks22:
@piskvorky Updated to improve the docstrings around the `tokenizer`. I also included the `PathLineSentences` corpus to `textcorpus` and made it inherit from `TextDirectoryCorpus` so it can share the same preprocessing/multiprocessing and filename filtering functionality.

gensim/corpora/textcorpus.py
```diff
@@ -575,62 +610,89 @@ def getstream(self):
...
+        """Yield documents from the underlying plain text collection (of one or more files).
+        Each item yielded from this method will be considered a document by subsequent
+        preprocessing methods.
+        """
+        paths = self.iter_filepaths()
+        if self.sort_filenames:
+            logger.debug("sorting filepaths")
+            paths = list(paths)
+            paths.sort(key=lambda path: os.path.basename(path))
+            logger.debug("found {} files: {}".format(len(paths), paths))
```
piskvorky: The rest of the code uses C-style formatting -- best keep it consistent.

gensim/corpora/textcorpus.py
```diff
@@ -575,62 +610,89 @@ def getstream(self):
...
+        num_files = 0
+        for path in paths:
+            logger.debug("reading file: %s", path)
+            num_files += 1
+            with utils.smart_open(path) as fin:
+                for line in itertools.islice(fin, self.limit):
+                    yield line.strip()
 
-        logging.info('files read into PathLineSentences:' + '\n'.join(self.input_files))
+        logger.debug("finished reading %d files", num_files)
```
piskvorky: Why not `info`?

gensim/corpora/textcorpus.py
```diff
@@ -191,19 +203,34 @@ def __init__(self, source=None, dictionary=None, metadata=False, character_filte
...
-                in `character_filters`; should return an iterable of tokens (strings).
+                in `character_filters`; should return an iterable of tokens. The tokens should
+                be in a format that can be parsed by the first callable in `token_filters`.
             token_filters (iterable of callable): each will be applied to the iterable of tokens
                 in order, and should return another iterable of tokens. These filters can add,
                 remove, or replace tokens, or do nothing at all. The default token filters
                 remove tokens less than 3 characters long and remove stopwords using the list
                 in `gensim.parsing.preprocessing.STOPWORDS`.
+            processes (int): number of processes to use for text preprocessing. The default is
+                -1, which will use (number of virtual CPUs - 1) worker processes, in addition
```
piskvorky: What happens if `number of virtual CPUs == 1`?

gensim/corpora/textcorpus.py
```diff
@@ -191,19 +203,34 @@ def __init__(self, source=None, dictionary=None, metadata=False, character_filte
...
             metadata (bool): True to yield metadata with each document, else False (default).
+                By default, metadata consists only of the line number for each document. This
+                is yielded from `get_texts` as a tuple of (tokens, (line_num,)) and from __iter__
+                as a tuple of (bag_of_words, (line_num,)).
             character_filters (iterable of callable): each will be applied to the text of each
                 document in order, and should return a single string with the modified text.
-                For Python 2, the original text will not be unicode, so it may be useful to
-                convert to unicode as the first character filter. The default character filters
-                lowercase, convert to unicode (strict utf8), perform ASCII-folding, then collapse
+                For Python 2, the original text will not be unicode (unless you modify your
```
piskvorky: I propose dropping the `(unless ...` bracket. This is already very complicated as it is.

Also, why single out Python 2? Is the behaviour different between Python 2 vs Python 3?
If so, I'd consider that a bug.

Let's keep the API as simple as possible: `getstream` returns unicode (no matter the Python version); all filters expect unicode.

gensim/corpora/textcorpus.py
```diff
@@ -191,19 +203,34 @@ def __init__(self, source=None, dictionary=None, metadata=False, character_filte
...
+                `getstream` method to convert it to unicode), so it may be useful to convert to 
+                unicode as the first character filter. The default character filters lowercase, 
+                convert to unicode (strict utf8), perform ASCII-folding, then collapse
```
piskvorky: Why lowercase before converting to unicode? Could lead to bugs for non-ASCII capitals.

gensim/corpora/textcorpus.py
```diff
@@ -191,19 +203,34 @@ def __init__(self, source=None, dictionary=None, metadata=False, character_filte
...
+                be in a format that can be parsed by the first callable in `token_filters`.
             token_filters (iterable of callable): each will be applied to the iterable of tokens
                 in order, and should return another iterable of tokens. These filters can add,
                 remove, or replace tokens, or do nothing at all. The default token filters
                 remove tokens less than 3 characters long and remove stopwords using the list
                 in `gensim.parsing.preprocessing.STOPWORDS`.
+            processes (int): number of processes to use for text preprocessing. The default is
+                -1, which will use (number of virtual CPUs - 1) worker processes, in addition
+                to the master process. If set to a number greater than the number of virtual
+                CPUs available, the value will be reduced to (number of virtual CPUs - 1).
```
piskvorky: -1 on this: why override a user's explicit request?

gensim/corpora/textcorpus.py
macks22: The intention here was to get the auto-formatting for the list of paths, as opposed to having to do my own '[' + ', '.join(paths) + ']', which seemed much messier. Should I still change it do this instead?

gensim/corpora/textcorpus.py
```diff
@@ -552,7 +554,7 @@ def getstream(self):
         """
         for path in self.iter_filepaths():
             logging.debug("reading file: %s", path)
-            with utils.smart_open(path) as f:
+            with utils.smart_open(path, 'rt') as f:
```
piskvorky: This looks fragile. Best to always open files in binary mode `rb`, and convert to text (unicode) explicitly, with an explicit encoding, where needed.

gensim/corpora/textcorpus.py
piskvorky: All these techniques should work identically (call str/repr on their arguments), so I'm not sure what you mean. Are you seeing a difference?

One advantage of the C-style formatter is that the types are immediately apparent to the reader (`%d` and `%s` or `%r` in this case).

gensim/corpora/textcorpus.py
macks22: Done; I've put the unicode conversion in the master process. For the sake of speed, it may make sense to have `getstream` return bytes in all versions, move the encoding parameters to the workers, and have them do the unicode conversion. Based on the ongoing Phrases refactor, that seems to be more of a bottleneck than I would've expected. Despite these considerations, I think it is sensible to do it in the master for the sake of simplicity for now.

gensim/corpora/textcorpus.py
macks22: No worker pool is used; all preprocessing occurs in the master process. I've updated the docstring to inform on this.

gensim/corpora/textcorpus.py
macks22: you're right; this mistrust of users is not suitable for a Python code base! Modified to remove upper bounding

gensim/corpora/textcorpus.py
macks22: Ah, I see; I simply wasn't aware of '%r' as an option to get the `repr`. Updated to use the C-style formatting.

gensim/corpora/textcorpus.py
macks22: changed to 'rb' followed by explicit unicode conversion

macks22:
@piskvorky thank you for your many reviews! I believe I have addressed all your comments and requests for changes. From what I can tell, the build check failures are only due to the imports in the `__main__` block of the `word2vec` module, which seem to be necessary. Is there anything else you'd like me to change for this PR? Thanks!

piskvorky:
Thanks for all the fixes and good work :)

I'll defer to @menshikh-iv for a final thorough review and decision (and fixing the unrelated build errors).

gensim/corpora/stateful_pool.py
```diff
@@ -0,0 +1,190 @@
...
+        self._pool.join()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._pool.terminate()
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
macks22: moved to test class

gensim/corpora/textcorpus.py
```diff
@@ -401,74 +561,223 @@ def iter_filepaths(self):
...
+            logging.debug("reading file: %s", path)
+            with utils.smart_open(path, 'rb') as f:
                 if self.lines_are_documents:
                     for line in f:
                         yield line.strip()
-                        num_texts += 1
                 else:
                     yield f.read().strip()
-                    num_texts += 1
+# endclass TextDirectoryCorpus
```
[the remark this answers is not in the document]
macks22: done

menshikh-iv:
@macks22  please pay attention to Appveyour problems, a lot of tests breaks, but all of this looks like one problem.

menshikh-iv:
Ping @macks22, what's a status here?

menshikh-iv:
Ping @macks22

macks22:
@menshikh-iv I'm hoping to update this in the coming weeks. Having trouble finding time to put towards it on the weekends. I'm thinking to refactor it according to some discussion I had with @michaelwsherman in regards to #1506. He had proposed a decomposition of responsibilities into something like a `TextCorpusLoader` and a `TextPreprocessor`. I think splitting the preprocessing logic out into its own class instead of dynamically generating classes and transplanting methods (as I'm doing now) will resolve the errors on Windows.

menshikh-iv:
@macks22 thanks for clarification, good luck :)

macks22:
@menshikh-iv hope all is well; I'm still working to find time to update this to fix the tests on Windows in the manner I described above. Hopefully next weekend.

menshikh-iv:
ping @macks22, have you time to finish this now?

menshikh-iv:
Ping @macks22, we are waiting you :)

macks22:
@menshikh-iv sorry for the latency in reply. I haven't had sufficient time to finish this. It's still on my Todo list,  but TBH, I may not have time again until end of December holidays.

menshikh-iv:
Ping @macks22, December has come, I remind you of us :)

menshikh-iv:
ping @macks22, I remind you about PR :)

menshikh-iv:
I'm sorry, but I'm closing this PR.
@macks22 feel free to re-open when you'll have time to finish this.

macks22:
Sorry for the delay in responding; I have been busier than expected. I will try to re-open and finish when I can.

PR #1462, opened 2017-07-05
[WIP] sklearn API for Gensim models

chinmayapancholi13:
This PR is for creating scikit-learn API for the following Gensim models : 
- [x] Doc2Vec Model
- [x] Text to Bag-of-words Model
- [x] Term Frequency-Inverse Document Frequency Model
- [x] Hierarchical Dirichlet Process Model
- [x] Phrases
- [ ] Normalization Model
- [ ] LogEntropy Model
- [ ] Dynamic Topic Model
- [ ] Topic Coherence Model

PR #1481, opened 2017-07-12
Add an sklearn wrapper for the Doc2Vec model

oxymor0n:
* TaggedDocumentTransformer transform tokenized documents to TaggedDocuments
 * SklD2VModel transforms TaggedDocuments to doc2vec vectors


oxymor0n:
oh oops sorry wasn't aware of #1462

oxymor0n:
In any case, this is a homebrewed wrapper that we used at @pagerinc and thought you guys were missing. It's not very different than your version in #1462, but I think it still has some value since the code is more concise and the tests are (somewhat) different

menshikh-iv:
@oxymor0n thank you for your work, we looked carefully at your code and changed several things in #1462. 
Now Doc2Vec wrapper merged from #1462, for this reason I close this PR.

Issue #1471, opened 2017-07-06
native fastText (unsupervised) in gensim


prakhar2b:
Currently, gensim has a wrapper for fastText. As discussed [here](https://github.com/RaRe-Technologies/gensim/pull/1153#issuecomment-312628437), we need to implement training code (`subword n-grams`, `hashing trick`)  for unsupervised fastText in gensim in python. As fastText is only a slight modification to word2vec, we will need to refactor the word2vec code to properly reuse the overlapping codes.

However, fastText outputs two files `.vec` and `.bin` which is C-standard. Should the python implementation in gensim provide `pkl` format output ?

This thread is intended to discuss and streamline all the requirements and deliverables regarding native fastText in gensim.

jayantj:
Hi @prakhar2b Let's simply use the pickle-style format from `utils.SaveLoad` we already use for word2vec models to persist the fastText models to disk.
Writing out the models in `.bin` format is a useful step, but it comes later.

Also, as discussed, please look at the word2vec code in detail, figure out what is needed for fastText, formulate a clear plan of action and post it here. It should contain details about -

1. Class structure
   - Do we subclass `Word2Vec`/create a common base class for Word2Vec and FastText/use composition?
2. Refactoring/code reuse
    - Analyze if it makes sense to reuse code from the word2vec training methods (I think it would, there should be a lot of overlap)
    - Analyze what methods can be reused, and how (or whether) they would have to be refactored to make them re-usable
3. Integrating with existing FastText wrapper
    - For actual use of the model (word similarity etc), we should be able to reuse a lot of code from the FastText wrapper
4. API of the new class (IMO, it's going to be pretty much the same as the FastText wrapper class)

piskvorky:
Awesome feature! Let me add that this will open up other unsupervised possibilities, such as  #1376.

gojomo:
Even though the gensim mission is more 'unsupervised', the addition of known-labels in the FastText-for-classification mode is such a small delta I would suggest it be in-scope. It's really just adding in another kind of known-data, during training, as possible 'target' outputs of the internal NN, that may make the resulting vectors better. (Potentially, even if *not* then using the resulting word-vecs for the exact same classification problem, the inclusion of these extra targets during training may have made the word-vecs better for other tasks.)

Also, it will otherwise be a constant exception-to-be-mentioned, in docs/support: "Yes gensim implements FastText except not FastText mode X".

dsouzadaniel:
[...]

PR #1482, opened 2017-07-13
[WIP] unsupervised fasttext 

prakhar2b:
#1471 
Unsupervised version of [facebook's fastText](https://github.com/facebookresearch/fastText)


prakhar2b:
@piskvorky In word2vec skip-gram model, the target is context word which we predict based on input word (exact opposite to cbow model), but in gensim word2vec code, I see that just like cbow model, for skip-gram model too we have taken context as input only. [here in word2vec.py](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec.py#L265) . 

It doesn't affect the performance, but could you please clarify a little for my own understanding.

gojomo:
Skip-gram & CBOW not 'exact opposite' - one uses single words to predict target, other uses averages-of-multiple-words.

There has occasionally been confusion because in there's two ways to loop over the text & window in skip-gram: you can loop over each main-word, then each window-word within that word's window, and make the predictions either (main-word -> window-word), or vice-versa (window-word -> main-word). The original word2vec paper described it one way, but the 1st-released word2vec.c code did it the alternate way, and there was a comment from one of authors that the way the code does it is slightly more cache-efficient. In both cases, the exact same (word->word) pairs are eventually trained, just in a slightly different iteration order. 

The naming `train_cbow_pair()` is somewhat unfortunate, as it isn't really a neat 'pair' but 'window-to-word'. And it'd certainly be better for variables like `word2` to have more functionally-descriptive names, like 'target_word' or 'context_word`.

gensim/models/fasttext.py
```diff
@@ -0,0 +1,374 @@
...
+			word2_indices = [word2.index for pos2, word2 in window_pos if (word2 is not None and pos2 != pos)]
+			# l1 = np_sum(model.wv.syn0[word2_indices], axis=0)  # 1 x vector_size
+			# if word2_indices and model.cbow_mean:
+			#	l1 /= len(word2_indices)
+
+			# for loop for ngrams for words in this window here
+			word2_subwords_indices = []
+
+			for indices in word2_indices:
+				word2_subwords_indices += get_subwords(model.wv.syn0[indices])  # subwords for each word in window except target word
```
jayantj: How is this intended to work? You're passing vector weights to `get_subwords`?

gensim/models/fasttext.py
```diff
@@ -0,0 +1,374 @@
...
+
+				# probably, we need to put subwords in model.wv.syn0 too
+
+			l1 = np_sum(model.wv.syn0[word2_subwords_indices], axis=0)  # 1 x vector_size
```
jayantj: I don't think storing ngram weights in `syn0` is a good idea, because it makes it much harder to use methods from `KeyedVectors`. We should have a separate matrix for ngram weights (maybe `syn0_all`, like in the `FastTextKeyedVectors` from the wrapper, to be able to reuse that class for methods like `most_similar` etc).
Once training is complete, `syn0` should be populated with computed in-vocab-word vectors, to not have to recompute word vectors for in-vocab words on every lookup.

gensim/models/fasttext.py
```diff
@@ -0,0 +1,374 @@
...
+        for pos, word in enumerate(word_vocabs):
+            reduced_window = model.random.randint(model.window)  # `b` in the original word2vec code
+
+            # subword_indices = []
+            # subword_indices += get_subwords(word)
+
+            # now go over all words from the (reduced) window, predicting each one in turn
+            start = max(0, pos - model.window + reduced_window)
+            for pos2, word2 in enumerate(word_vocabs[start:(pos + model.window + 1 - reduced_window)], start):
+                # don't train on the `word` itself
```
prakhar2b: I didn't change the comment here. I think it's correct as `don't train on word itself` means word is the target and rest words in windows are context (taken as input in word2vec code)

jayantj: I meant, it is present before a different line than it talks about.

gensim/models/fasttext.py
```diff
@@ -0,0 +1,394 @@
...
+        self.num_ngram_vectors = len(all_ngrams)
+        logger.info("Total number of ngrams in the vocab is %d", self.num_ngram_vectors)
+
+        ngram_indices = []
+        for i, ngram in enumerate(all_ngrams):
+            ngram_hash = Ft_Wrapper.ft_hash(ngram)
+            ngram_indices.append(len(self.wv.vocab) + ngram_hash % self.bucket)
+            self.wv.ngrams[ngram] = i
+
+        self.wv.syn0_all = self.wv.syn0_all.take(ngram_indices, axis=0)
```
jayantj: This is not going to work if we want to resume training. Extraneous vectors should be discarded only when we are sure no training is going to take place further (probably analogous to `init_sims(replace=True)` in `Word2Vec`.

gensim/test/test_fasttext.py
```diff
@@ -0,0 +1,35 @@
...
+                yield utils.simple_preprocess(line)
+
+
+class TestNativeFastText(unittest.TestCase):
+
+    def testTraining(self):
+        """ Test training with subwords for both skipgram and cbow model"""
+        sentences = LeeCorpus()
+        self.assertTrue(FastText(sentences, min_count=100, size=100, workers=3))
+        self.assertTrue(FastText(sentences, sg=1, min_count=100, size=100, workers=3))
```
jayantj: Model sanity tests similar to word2vec are required here too to ensure the "correct" models are learnt, as well as a comparison to the original FastText vectors.

gensim/models/fasttext.py
prakhar2b: yes, I got it. Thanks

gensim/test/test_fasttext.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # build vocab and train in one step; must be the same as above
+        model2 = FastText(sentences, size=2, min_count=1, hs=1, negative=0)
+        self.models_equal(model, model2)
+
+    def model_sanity(self, model, train=True):
+        """Even tiny models trained on LeeCorpus should pass these sanity checks"""
+        # run extra before/after training tests if train=True
+        """
+        if train:
```
tmylk: is there a reason that there is no test here?

prakhar2b:
For context - model trained using this code on Lee Corpus

> model = FastText(sentences, min_count=1, size=50)

`Total number of ngrams in the vocab is 87485`
`training model with 3 workers on 10781 vocabulary and 50 features, using sg=0 hs=0 sample=0.001 negative=5 window=5`

model.similarity("woman", "woman") >> 1.0
model.similarity("night", "nights") >> 0.99999806698759131
model.doesnt_match("breakfast cereal dinner lunch".split()) >> cereal (as expected)

> model.most_similar("nights")

(result looks somewhat good with syntactic pov

[('night', 0.999998152256012),
 ('lights', 0.9999980926513672),
 ('night.', 0.9999977350234985),
 ("night's", 0.999997615814209),
 ('fighting,"', 0.9999971985816956),
 ('sanctioned', 0.9999971985816956),
 ('night,', 0.9999971389770508),
 ('flights', 0.9999971389770508),
 ('airliner', 0.9999971389770508),
 ('Heights', 0.999997079372406)]

piskvorky:
@prakhar2b we're going for identical with the original, for correctness, to start with (not "somewhat good").

prakhar2b:
@piskvorky yes, what I meant was this PR was very close to  getting the identical result (perhaps a little debugging), the difference IMO is mainly because of the way we have initialized n-gram weights.

gensim/test/test_fasttext.py
prakhar2b: yes, the training code in the test is commented out as it took very long time on travis for pure python code. I'd look to reduce the number of iterations, or maybe a smaller datasets.

prakhar2b:
@jayantj could you look into it and review what needs to be done to get identical results. I think much effort has gone into this so far, and there is no point abandoning this PR.

cc @piskvorky

jayantj:
I agree we don't want to abandon this PR, there has been a lot of effort into it.
As I said before, I think the next step is comparing the `most_similar` results for the FastText models trained using the C code vs Python.

The fact that the results look like this - 
```python
model.most_similar("nights")
-> [('night', 0.999998152256012),
('lights', 0.9999980926513672),
('night.', 0.9999977350234985),
("night's", 0.999997615814209),
('fighting,"', 0.9999971985816956),
('sanctioned', 0.9999971985816956),
('night,', 0.9999971389770508),
('flights', 0.9999971389770508),
('airliner', 0.9999971389770508),
('Heights', 0.999997079372406)]
```
- is not very conclusive. The way that word vectors are defined in FastText (sum of all ngram vectors) means that even without any training (randomly initialized ngram vectors), the word vector for `nights` is likely to be very close to words with similar ngrams (`night`, `lights` etc.)
So it is hard to draw any conclusions from the results so far. A more detailed analysis is necessary.

In case they do differ significantly, the next steps in my opinion would be to check -
1. How the initialization differs in C FastText vs Gensim FastText
2. Whether the gradients are computed in the same way or not
3. If there are any subtle differences caused by the other hyperparameters

gojomo:
Also keep in mind most modes of word2vec/fasttext use internal randomness. Even with deterministic seeding, if multiple worker threads are in play, the training texts will be consumed in slightly different orders, and thus also be paired with slightly different draws in the PRNG stream. So the standard for comparison against a reference implementation will usually be, "very close in observable quality" rather than "identical numerical results".

piskvorky:
Never mind multiple workers for now -- let's aim for identical results in the simplest possible mode to start with, single-threaded.

menshikh-iv:
Continued in #1525

PR #1490, opened 2017-07-19
Remove unittest2 imports

souravsingh:


souravsingh:
This is related to issue #3225 regarding running tests in conda-forge

gensim/test/test_ldamallet_wrapper.py
```diff
@@ -10,7 +10,13 @@
 
 
 import logging
-import unittest2 as unittest
+import sys
+
+if sys.version_info < (2, 7):
```
menshikh-iv: @souravsingh now we don't support Python version that older than 2.7, you no need this `if`

souravsingh:
@menshikh-iv Can you take a look?

menshikh-iv:
@souravsingh yes, we should to replace `unittest2` to `unittest` and remove it from env.

menshikh-iv:
Thanks @souravsingh :+1:

PR #1500, opened 2017-07-23
Fix links & spaces in Quick start guide

iamsanten:
Removed sign of space from "split()" because space is default. There is no sign of space in the text below, so I decided to cut it.


menshikh-iv:
Good job @iamsanten, congratulations with your first PR:1st_place_medal:

PR #1508, opened 2017-07-26
Changing six's viewitems() to iteritems()

HodorTheCoder:
Per issue 1495, using viewitems() from the six package breaks the six package version requirement of >= 1.5 when iteritems() has the same functionality for this use.  

Changing back to iteritems() fixes the requirement properly, in the file gensim/topic_coherence/text_analysis.py


menshikh-iv:
Nice, thank you @HodorTheCoder, congratulations with your first PR :+1:

PR #1510, opened 2017-07-28
Add NLPub to the list of adopters

dustalov:
We actively use Gensim for developing distributional semantic models and their applications to lexical semantics.

* EACL 2017 Short Paper : <http://aclweb.org/anthology/E/E17/E17-2087.pdf>
* ACL 2017 Full Paper: <https://arxiv.org/abs/1704.07157>

It would be great for us to be added to the list of adopters.


menshikh-iv:
@dustalov :+1:

PR #1530, opened 2017-08-13
Word2vec coherence

### LINKED ISSUE #1380: Add a new indirect confirmation measure based on word2vec similarity

macks22:
Resolves #1380.

This PR adds a new coherence measure to CoherenceModel, called "c_w2v". This uses word2vec word vectors for computing similarity between terms to calculate coherence. This implementation adds a new accumulator in the text_analysis module that either uses pre-trained KeyedVectors or trains a Word2Vec model on the input corpus to derive KeyedVectors. A new keyword argument keyed_vectors is added to the CoherenceModel to pass in pre-trained vectors.

Rather than requiring parity between the corpus dictionary and the keyed vectors vocabulary, the coherence calculation ignores terms that are missing from the vocabulary (with a warning). This PR also adds a new with_std argument to the get_coherence_per_topic method to calculate the standard deviation between the various segments for each topic. This can serve as some indication of topics that have good overall coherence but a few notable outlier terms or subsets.

Finally, I noticed that the CoherenceModel's method of getting topics from models was implemented in a switch-style manner and refactored it to use polymorphism. Specifically, I introduced a new get_topics method to all topic models that returns the topic-term distributions (except for LSI, which just returns the weights, since its not a real distribution).


macks22:
@menshikh-iv I believe this PR is now ready to go. I responded to the change requests in the original PR: I've added the `get_topics` method to the `BaseTopicModel` and I've updated the notebook to demonstrate how to use the new coherence measure. I believe the existing CI check failures are all due to the imports in the `corpora.__init__`, which are being viewed as "unused imports" despite their intentionality.

Please let me know if there are any additional changes needed. Thanks!

gensim/models/coherencemodel.py
```diff
@@ -261,6 +260,10 @@ def for_topics(cls, topics_as_topn_terms, **kwargs):
             for topic in topic_list:
                 topn = max(topn, len(topic))
 
+        if 'topn' in kwargs:
+            topn = min(kwargs.get('topn'), topn)
+            del kwargs['topn']
```
piskvorky: Have a look at `dict.pop()`.

macks22:
@piskvorky @menshikh-iv all merge conflicts have been resolved and all requested changes made. Please let me know if there is anything else to be done here. Thanks!

docs/notebooks/topic_coherence_model_selection.ipynb
```diff
@@ -213,24 +154,25 @@
...
+      "Dictionary(24593 unique tokens: [u'woods', u'hanging', u'woody', u'localized', u'gaa']...)\n",
+      "CPU times: user 20 s, sys: 797 ms, total: 20.8 s\n",
+      "Wall time: 21.1 s\n"
      ]
     }
    ],
    "source": [
     "%%time\n",
     "\n",
     "corpus = NewsgroupCorpus(data_path)\n",
```
menshikh-iv: What's a reason to use `NewsgroupCorpus` instead of [sklearn data interface](http://scikit-learn.org/stable/datasets/index.html#the-20-newsgroups-text-dataset)? For comfort?

docs/notebooks/topic_coherence_model_selection.ipynb
```diff
@@ -394,109 +520,3275 @@
...
+   "execution_count": 33,
    "metadata": {
     "collapsed": false
    },
    "outputs": [
     {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "WARNING:gensim.topic_coherence.indirect_confirmation_measure:at least one topic word not in word2vec model\n",
```
menshikh-iv: Please suppress this warning here (and below)

gensim/topic_coherence/indirect_confirmation_measure.py
```diff
@@ -35,30 +37,75 @@
...
+        if with_std:
+            topic_coherences.append((np.mean(segment_similarities), np.std(segment_similarities)))
+        else:
+            topic_coherences.append(np.mean(segment_similarities))
+
+    return topic_coherences
+
+
+def cosine_similarity(segmented_topics, accumulator, topics, measure='nlr', gamma=1,
+                      with_std=False):
```
menshikh-iv: Don't use vertical indent (here you have not very long line) please make definition in one line

docs/notebooks/topic_coherence_model_selection.ipynb
macks22: This was a comfort thing -- I already had code for this. I've updated to use the `sklearn.datasets.fetch_20newsgroups` function for retrieving the data.

docs/notebooks/topic_coherence_model_selection.ipynb
macks22: done

gensim/topic_coherence/indirect_confirmation_measure.py
macks22: I added a new argument here, which made it longer, so I used the hanging indent instead.

macks22:
@menshikh-iv thank you for the review; I've addressed your changes. In updating the notebook, I also noticed a few other minor things to improve the robustness of this code, so I made some other minor changes. Please let me know if there is anything else you'd like modified. Thanks!

menshikh-iv:
Thanks for your work @macks22, you are TOP contributor :fire:

ydennisy:
@macks22 this functionality is not documented - would be great to add the docs 👍

mpenkov:
@ydennisy May I interest you in making a PR? ;)

PR #1538, opened 2017-08-17
Pin sklearn version

chinmayapancholi13:
This PR pins scikit-learn version to 0.18.2 to pass unit-tests which use sklearn's `Pipeline`.


menshikh-iv:
For clarification - it's temporary pinning because we need more time to investigate a problem with new sklearn version (0.19.0)

menshikh-iv:
Fixed in #1462

PR #1544, opened 2017-08-21
Fix doctag unicode problem. Fix 1543

### LINKED ISSUE #1543: Unicode encode problem when saving doc2vec models in plain word2vec format.
#### Description
Exporting doc2vec embeddings in plain word2vec format fails when document title contains unicode chars that cannot be ASCII encoded. This only fails in Python 2.7 because Python 3 defaults to unicode strings. The problem is the call `str(..)` in [doc2vec.py](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/doc2vec.py#L850)

#### Code to Reproduce
Example:
```
from gensim.models import doc2vec

# Load doc2vec model with a document title not ASCII encodable
m = Doc2Vec.load('model')
m.save_word2vec_format('model.out')
```

#### Expected Results
Export works

#### Actual Results
```
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-3-d12938849014> in <module>()
----> 1 m.save_word2vec_format('model.out')

..../.virtualenvs/gensim27_2.3/local/lib/python2.7/site-packages/gensim-2.3.0-py2.7-linux-x86_64.egg/gensim/models/doc2vec.pyc in save_word2vec_format(self, fname, doctag_vec, word_vec, prefix, fvocab, binary)
    848                 # store as in input order
    849                 for i in range(len(self.docvecs)):
--> 850                     doctag = prefix + str(self.docvecs.index_to_doctag(i))
    851                     row = self.docvecs.doctag_syn0[i]
    852                     if binary:

UnicodeEncodeError: 'ascii' codec can't encode character u'\xf3' in position 8: ordinal not in range(128)
```

#### Versions
Linux-4.4.0-59-generic-x86_64-with-Ubuntu-16.04-xenial
('Python', '2.7.12 (default, Nov 19 2016, 06:48:10) \n[GCC 5.4.0 20160609]')
('NumPy', '1.13.1')
('SciPy', '0.19.1')
('gensim', '2.3.0')
('FAST_VERSION', 1)

englhardt:
Fixes #1543

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -847,7 +847,7 @@ class Doc2Vec(Word2Vec):
                     fout.write(utils.to_utf8("%s %s\n" % (total_vec, self.vector_size)))
                 # store as in input order
                 for i in range(len(self.docvecs)):
-                    doctag = "%s%s" % (prefix, self.docvecs.index_to_doctag(i))
+                    doctag = u"%s%s" % (prefix, self.docvecs.index_to_doctag(i))
                     row = self.docvecs.doctag_syn0[i]
                     if binary:
                         fout.write(utils.to_utf8(doctag) + b" " + row.tostring())
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -847,7 +847,7 @@ class Doc2Vec(Word2Vec):
                     fout.write(utils.to_utf8("%s %s\n" % (total_vec, self.vector_size)))
                 # store as in input order
                 for i in range(len(self.docvecs)):
-                    doctag = prefix + str(self.docvecs.index_to_doctag(i))
+                    doctag = "%s%s" % (prefix, self.docvecs.index_to_doctag(i))
                     row = self.docvecs.doctag_syn0[i]
                     if binary:
                         fout.write(utils.to_utf8(doctag) + b" " + row.tostring())
--- a/gensim/test/test_doc2vec.py
+++ b/gensim/test/test_doc2vec.py
@@ -30,11 +30,16 @@ datapath = lambda fname: os.path.join(module_path, 'test_data', fname)
 
 
 class DocsLeeCorpus(object):
-    def __init__(self, string_tags=False):
+    def __init__(self, string_tags=False, unicode_tags=False):
         self.string_tags = string_tags
+        self.unicode_tags = unicode_tags
 
     def _tag(self, i):
-        return i if not self.string_tags else '_*%d' % i
+        if self.unicode_tags:
+            return u'_\xa1_%d' % i
+        elif self.string_tags:
+            return '_*%d' % i
+        return i
 
     def __iter__(self):
         with open(datapath('lee_background.cor')) as f:
@@ -94,6 +99,13 @@ class TestDoc2VecModel(unittest.TestCase):
         model.save_word2vec_format(test_word, doctag_vec=False, word_vec=True, binary=True)
         binary_model_dv = keyedvectors.KeyedVectors.load_word2vec_format(test_word, binary=True)
         self.assertEqual(len(model.wv.vocab), len(binary_model_dv.vocab))
+
+    def test_unicode_in_doctag(self):
+        """Test storing document vectors of a model with unicode titles."""
+        model = doc2vec.Doc2Vec(DocsLeeCorpus(unicode_tags=True), min_count=1)
+        model.save_word2vec_format(testfile(), doctag_vec=True, word_vec=True, binary=True)
+        binary_model_dv = keyedvectors.KeyedVectors.load_word2vec_format(testfile(), binary=True)
+        self.assertEqual(len(model.wv.vocab) + len(model.docvecs), len(binary_model_dv.vocab))
 
     def test_load_mmap(self):
         """Test storing/loading the entire model."""
--- a/gensim/test/test_doc2vec.py
+++ b/gensim/test/test_doc2vec.py
@@ -103,9 +103,10 @@ class TestDoc2VecModel(unittest.TestCase):
     def test_unicode_in_doctag(self):
         """Test storing document vectors of a model with unicode titles."""
         model = doc2vec.Doc2Vec(DocsLeeCorpus(unicode_tags=True), min_count=1)
-        model.save_word2vec_format(testfile(), doctag_vec=True, word_vec=True, binary=True)
-        binary_model_dv = keyedvectors.KeyedVectors.load_word2vec_format(testfile(), binary=True)
-        self.assertEqual(len(model.wv.vocab) + len(model.docvecs), len(binary_model_dv.vocab))
+        try:
+            model.save_word2vec_format(testfile(), doctag_vec=True, word_vec=True, binary=True)
+        except UnicodeEncodeError:
+            self.fail('Failed storing unicode title.')
 
     def test_load_mmap(self):
         """Test storing/loading the entire model."""


gojomo:
Can you add a unit test in `test_doc2vec.py` that fails before the fix, and works thereafter?

englhardt:
I added a test.

On **develop** branch:
```
Launching unittests with arguments python -m unittest test_doc2vec.TestDoc2VecModel.test_unicode_in_doctag in .../gensim/gensim/test

Using TensorFlow backend.

Error
  File "/usr/lib64/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File ".../gensim/gensim/test/test_doc2vec.py", line 106, in test_unicode_in_doctag
    model.save_word2vec_format(testfile(), doctag_vec=True, word_vec=True, binary=True)
  File ".../gensim/gensim/models/doc2vec.py", line 850, in save_word2vec_format
    doctag = prefix + str(self.docvecs.index_to_doctag(i))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa1' in position 1: ordinal not in range(128)
```

gensim/test/test_doc2vec.py
```diff
@@ -30,11 +30,16 @@
...
 class DocsLeeCorpus(object):
-    def __init__(self, string_tags=False):
+    def __init__(self, string_tags=False, unicode_tags=False):
         self.string_tags = string_tags
+        self.unicode_tags = unicode_tags
 
     def _tag(self, i):
-        return i if not self.string_tags else '_*%d' % i
+        if self.unicode_tags:
+            return u'_\xa1_%d' % i
```
menshikh-iv: Looks strange, why is it so different with line 41, why you use a `\xa1_'?

gensim/test/test_doc2vec.py
```diff
@@ -95,6 +100,13 @@ def testPersistenceWord2VecFormat(self):
         binary_model_dv = keyedvectors.KeyedVectors.load_word2vec_format(test_word, binary=True)
         self.assertEqual(len(model.wv.vocab), len(binary_model_dv.vocab))
 
+    def test_unicode_in_doctag(self):
```
menshikh-iv: Please add a test with an exception (and check it with assertraise).

gensim/test/test_doc2vec.py
```diff
@@ -95,6 +100,13 @@ def testPersistenceWord2VecFormat(self):
...
+        """Test storing document vectors of a model with unicode titles."""
+        model = doc2vec.Doc2Vec(DocsLeeCorpus(unicode_tags=True), min_count=1)
+        model.save_word2vec_format(testfile(), doctag_vec=True, word_vec=True, binary=True)
+        binary_model_dv = keyedvectors.KeyedVectors.load_word2vec_format(testfile(), binary=True)
+        self.assertEqual(len(model.wv.vocab) + len(model.docvecs), len(binary_model_dv.vocab))
```
menshikh-iv: How is this assert related with your PR / with reason for test?

gensim/test/test_doc2vec.py
englhardt: I enforce a title that is not ASCII encodable, i.e. a unicode character '¡'.

gensim/test/test_doc2vec.py
englhardt: It is true that lines 107/108 are not related to the PR. I can remove them.

gensim/test/test_doc2vec.py
englhardt: The test should rather be the inverse of assertraise. The exception should not be thrown with this PR. I will add a inverse test.

englhardt:
I have updated the test. Please have a look.

gensim/test/test_doc2vec.py
menshikh-iv: @gojomo wdyt?

gensim/test/test_doc2vec.py
menshikh-iv: It's a good idea to add 2 tests: first with `unicode_tags=False` and assertRaise and second is a current test.

englhardt: I think you are mixing up what really happens, let me explain:
In gensim 2.1 a modified `save_word2vec_format` for Doc2Vec was added in #1256. The storing handled to export any utf8 symbols [doc2vec.py#L853](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/doc2vec.py#L853). Unfortunately, the `str(..)` call in [doc2vec.py#L850](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/doc2vec.py#L850) fails in Python 2.7 when the doctags contain chars not ASCII encodable.
Currently the develop branch is broken and the added test will fail, i.e. `UnicodeEncodeError` is thrown for Python 2.7.
The test now checks whether it can successfully store unicode doctags and fails otherwise. I have to generate them (with the code above) since the default string doctags are ASCII encodable (-> addition of the '¡' character).
So, what should the second test actually be then?

gensim/models/doc2vec.py
```diff
@@ -847,7 +847,7 @@ def save_word2vec_format(self, fname, doctag_vec=False, word_vec=True, prefix='*
                     fout.write(utils.to_utf8("%s %s\n" % (total_vec, self.vector_size)))
                 # store as in input order
                 for i in range(len(self.docvecs)):
-                    doctag = prefix + str(self.docvecs.index_to_doctag(i))
+                    doctag = "%s%s" % (prefix, self.docvecs.index_to_doctag(i))
```
piskvorky: We should pick one type (byte string or unicode) and stick with it.

Supporting both types at the same time like this (`doctag` will be silently upcast to unicode in this solution, if the argument is unicode) looks very brittle.

I'm not as familiar with this codebase as @gojomo is, but converting to one fixed type (bytestring or unicode) during input processing is preferable. At this point, during model save, the type ought to be fixed.

englhardt: I agree that the _hidden_ cast to unicode is not optimal. The problem is that the method [`index_to_doctag`](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/doc2vec.py#L332) can return a string as well as an int. The fix is similar to another place in the class [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/doc2vec.py#L395).
I am open for other suggestions though.

piskvorky: Aha, so the expected types are either unicode or int. A `u"%s" % index_to_doctag` should work either way, right? No change needed (beside dropping that `str`). IIRC that int was an optimization hack (uses less memory).

My suggestion is to keep the tags explicitly as "unicode or int", or "bytestring or int". And then during string formatting, rely on the fact it's unicode-or-int (or byte-or-int), possibly even with an assert.

@gojomo What are the invariants here, how to do this cleanly? That unicode-or-str-or-int is really confusing, and apparently not encapsulated well (fixes in one place leave bugs in other places).

gensim/test/test_doc2vec.py
gojomo: Testing with a non-ASCII character seems sensible/necessary to me!

gensim/models/doc2vec.py
gojomo: IMO the type of `tags` should be as permissive as possible – conceivably even anything that works as a dict key would work everywhere in `Doc2Vec` *except* in this new bit of convenience for shoehorning doc-vecs into the (limited, somewhat-'legacy') word2vec.c format. For people who want to use this format, they should be sure to use tags that can write as a nice string-token – I'd think it'd be OK here if this code is tolerant of any reasonable such choice. (They may also need to make sure, themselves, that their tags don't include internal spaces if using the 'text' word2vec.c format.)

englhardt: @piskvorky I changed it to `u"%s%s" % ..` so that the cast is less hidden.
@gojomo Your last point is especially important, yeah.

menshikh-iv:
@gojomo should we merge this now?

gojomo:
I haven't independently exercised the bug or confirmed the fix, but as all old tests and the well-focused new test pass, and the changes are tiny/well-focused, it looks OK to me!

menshikh-iv:
Thanks @englhardt, congratz with your first PR :+1: 
Please follow an instruction from #1587 and receive gifts!

Issue #1521, opened 2017-08-04
Refactor current gensim code by PEP8


menshikh-iv:
We have a lot of old code which doesn't satisfy PEP8, it's time to fix it!

What we want:
- [x] PEP8 (without a hard 80-char constraint for long lines)
- [x] No vertical indents (only hanging)
- [x] map/filter -> list/set/dict comprehension
- [x] logging message arguments as method arguments (not direct formatting)

zsef123:
Hi, I'm trying to refactor by PEP8

E731 ( do not assign a lambda expression, use a def these error ) is often seen during refactoring

ex) `blas = lambda name, ndarray: scipy.linalg.get_blas_funcs((name,), (ndarray,))[0]`

It looks like code writer prefer lambda expresion than def

Do you think you should change lambda expression to def in case of E731 error?

menshikh-iv:
Hi @zsef123, we ignore E731 in gensim, look at [this line](https://github.com/RaRe-Technologies/gensim/blob/develop/continuous_integration/travis/flake8_diff.sh#L137).

zsef123:
I send PR https://github.com/RaRe-Technologies/gensim/pull/1550
but I got some problems. 

`return buffer[start_index : start_index + nbytes].view(dtype).reshape(shape, order=order)`

In case E203 error occurs in pep8 library( https://pypi.python.org/pypi/pep8 ), but it's fine in documentation( https://www.python.org/dev/peps/pep-0008/#other-recommendations )

I think E203 error corresponds to scope delimeter( like if True: ) 

So First, no change
`return buffer[start_index : start_index + nbytes].view(dtype).reshape(shape, order=order)`
these line still occurs error in pep8 library but it's fine in docs

Second, line changed to
`return buffer[start_index:start_index + nbytes].view(dtype).reshape(shape, order=order)`
these line not occurs error in pep8 library and fine too

Which of the two is better?

piskvorky:
The second option does not conform to PEP8 (see the "No:" section in https://www.python.org/dev/peps/pep-0008/#pet-peeves).

zsef123:
What about ignore F405 and F403 error

> F405 : name may be undefined, or defined from star imports: module
> F403 : ‘from module import *’ used; unable to detect undefined names

In [test_parsing.py](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_parsing.py)

`from gensim.parsing.preprocessing import *`
error occurs F403
[...]

PR #1550, opened 2017-08-27
Refactor by pep8

zsef123:
Fixes https://github.com/RaRe-Technologies/gensim/issues/1521

First checkbox 
> PEP8 (without a hard 80-char constraint for long lines)

Most changes are whitespace adjustment and import order changes.

Then unused var and header are removed except in try catch, named _ var and side effects call. 
Excepted case add comments 'noqa'

And test_atmodel.py, test_ldamallet_wrapper.py, test_ldamodel codes don't use fname like under code
```
fname = testfile()
model = self.model
model.save(testfile(), sep_limit=0)
```
save or load argument changed from testfile() to fname


gensim/corpora/__init__.py
```diff
@@ -3,15 +3,15 @@
...
-from .bleicorpus import BleiCorpus
-from .svmlightcorpus import SvmLightCorpus
-from .lowcorpus import LowCorpus
-from .dictionary import Dictionary
-from .hashdictionary import HashDictionary
-from .wikicorpus import WikiCorpus
-from .textcorpus import TextCorpus
-from .ucicorpus import UciCorpus
-from .malletcorpus import MalletCorpus
+from .mmcorpus import MmCorpus  # noqa:F401
```
piskvorky: -1: we don't want to litter the code with some tool-specific constructs.

Instead, open an issue with the tool that requires this, it's a bug on their side.

If you find such comments in other parts of gensim, please remove them.

gensim/examples/dmlcz/sources.py
```diff
@@ -190,13 +187,13 @@ class DmlCzSource(DmlSource):
 
     See the ArticleSource class for general info on sources.
     """
+
     def idFromDir(self, path):
         assert len(path) > len(self.baseDir)
         dmlczId = open(os.path.join(path, 'dspace_id')).read().strip()
-        pathId = path[len(self.baseDir) + 1 : ]
+        pathId = path[len(self.baseDir) + 1:]
```
piskvorky: This is against PEP8 (see https://www.python.org/dev/peps/pep-0008/#other-recommendations ).

gensim/interfaces.py
```diff
@@ -267,7 +263,7 @@ def __iter__(self):
                 # (unlike numpy). so, clip the end of the chunk explicitly to make
                 # scipy.sparse happy
                 chunk_end = min(self.index.shape[0], chunk_start + self.chunksize)
-                chunk = self.index[chunk_start : chunk_end]
+                chunk = self.index[chunk_start: chunk_end]
```
piskvorky: PEP8: https://www.python.org/dev/peps/pep-0008/#other-recommendations (here and elsewhere).

gensim/corpora/__init__.py
menshikh-iv: @piskvorky it's not a tool bug, it's really "unused import" by PEP

piskvorky: But it is used -- it makes the classes available directly from the package. Without it, you won't be able to `from corpora import MmCorpus`.

Or how can this be done in other way, that doesn't trigger F401?

menshikh-iv: > it makes the classes available directly from the package 
You are right, but now in the code, it's unused import.

I know another way - use this imports (like your example) in gensim internally, but I don't think that this is a good idea. For my opinion, `noqa` suppress is OK.

zsef123:
> E722 | do not use bare except, specify exception instead

For example

```
try:
    "do something"
except:
    pass
```
E722 error occurs except line

In gensim code, some line add Exception class but the others not specify exception

So if not used exception class in exception scope, I just add Exception like this code

```
try:
    "do something"
except Exception:
    pass
```

zsef123:
I miss check in corpora dir files.

I'm not touch about noqa, line slice whitespace yet.

and naming like 'l' changed to suitable name

gensim/matutils.py
```diff
@@ -150,7 +150,7 @@ def zeros_aligned(shape, dtype, order='C', align=128):
     nbytes = np.prod(shape, dtype=np.int64) * np.dtype(dtype).itemsize
     buffer = np.zeros(nbytes + align, dtype=np.uint8)  # problematic on win64 ("maximum allowed dimension exceeded")
     start_index = -buffer.ctypes.data % align
-    return buffer[start_index: start_index + nbytes].view(dtype).reshape(shape, order=order)
+    return buffer[start_index:start_index + nbytes].view(dtype).reshape(shape, order=order)
```
piskvorky: Does not conform to PEP8 (need spaces around `:` here, see https://github.com/RaRe-Technologies/gensim/issues/1521#issuecomment-325270209).

zsef123: Sorry, Line slice white space was not my intention. I missed these line in this commit.

Anyway, I agree to input space around `:`.

but if input space like `return buffer[start_index : start_index + nbytes]`, E203 error occurs. [(pycodestyle issue)](https://github.com/PyCQA/pycodestyle/issues/373)

Is it okay upper line?

piskvorky: I don't think there's any error in the code, the version with spaces is correct. It appears to be a bug in the style-checking software.

zsef123:
Issue https://github.com/RaRe-Technologies/gensim/issues/1551

> By adding # flake8: noqa to the file  [(docs:ignoring-entire-file)](http://flake8.pycqa.org/en/latest/user/violations.html#ignoring-entire-file)

I have adding above comment ignore file

It may be dangerous but I think `__init__.py` files short and simple enough so I can safely ignore flake8 errors 

Excepts [models/__init_.py](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/__init__.py), this file contains short class.
Still, I think it's okay to ignore the error because the file is still short, and support by PEP8.

gensim/corpora/__init__.py
```diff
@@ -1,17 +1,17 @@
 """
 This package contains implementations of various streaming corpus I/O format.
 """
-
+# flake8: noqa
```
piskvorky: This seems to combine the worst of both worlds. Still extra comments, AND no style checking.

I think I prefer even the `noqa:F401` version.

zsef123:
I canceled commit about ignore file.

> F403 : `from module import *` used; unable to detect undefined names

And just `from module import *` disassembled to each functions

zsef123:
Sorry, I clicked delete accidently.

And I think I've finished applying PEP8.

1. white space adjustment
2. delete unused var( except naming `_` ), unused import modules
3. Add specific exceptions( `except Exception:` )
4. `noqa` comments about flake8 F401 error ( Issue https://github.com/RaRe-Technologies/gensim/issues/1551 )
5. disassembled `import *` about flake8 F403 error

menshikh-iv:
Hi @zsef123, please remove file gensim/scripts/make_wiki_lemma.py from your PR ASAP (because you broke current symlink -> we can't build package)

menshikh-iv:
@zsef123 I means to remove this file from your PR (not from the repo), you need to checkout this file to `develop` state.

zsef123:
@menshikh-iv Sorry I check again.
But PEP8 E203 error occurs in travis
```
gensim/examples/dmlcz/sources.py:125:41: E203 whitespace before ':'
        intId = path[1 + path.rfind('#') :]

gensim/models/word2vec.py:1552:37: E203 whitespace before ':'
                        yield line[i : i + self.max_sentence_length]
```
Issue https://github.com/RaRe-Technologies/gensim/issues/1521
>Second, line changed to
return buffer[start_index:start_index + nbytes].view(dtype).reshape(shape, order=order)
these line not occurs error in pep8 library and fine too

And

>piskvorky commented 8 days ago • edited
The second option does not conform to PEP8 (see the "No:" section in https://www.python.org/dev/peps/pep-0008/#pet-peeves).Basically, the : operator has a lower priority than e.g. +, and so shouldn't have less whitespace because that implies more priority in Python, by convention.

So I input space around `:`. Is it problem?

menshikh-iv:
@zsef123 space **before** `:` is problem here.

zsef123:
In matutils.py, I already remove space before `:` ( in first commits )
And I requested like under  code
```
-    return buffer[start_index: start_index + nbytes].view(dtype).reshape(shape, order=order)
+    return buffer[start_index:start_index + nbytes].view(dtype).reshape(shape, order=order)
```

>zsef123 7 days ago • edited Contributor
Sorry, Line slice white space was not my intention. I missed these line in this commit.
Anyway, I agree to input space around :.
but if input space like return buffer[start_index : start_index + nbytes], E203 error occurs. (pycodestyle issue)
Is the above line okay if an error occurs?

>piskvorky 7 days ago • edited Owner
I don't think there's any error in the code, the version with spaces is correct. The version without spaces is incorrect. It appears to be a bug in the style-checking software. CC @menshikh-iv

And Issue https://github.com/RaRe-Technologies/gensim/issues/1521

>The second option does not conform to PEP8 (see the "No:" section in https://www.python.org/dev/peps/pep-0008/#pet-peeves).Basically, the : operator has a lower priority than e.g. +, and so shouldn't have less whitespace because that implies more priority in Python, by convention.

So I changed to input space around `:`

zsef123:
In matutils.py, I already remove space before `:` ( in first commits )
And I requested like under  code
```
-    return buffer[start_index: start_index + nbytes].view(dtype).reshape(shape, order=order)
+    return buffer[start_index:start_index + nbytes].view(dtype).reshape(shape, order=order)
```

>zsef123 7 days ago • edited Contributor
Sorry, Line slice white space was not my intention. I missed these line in this commit.
Anyway, I agree to input space around :.
but if input space like return buffer[start_index : start_index + nbytes], E203 error occurs. (pycodestyle issue)
Is the above line okay if an error occurs?

>piskvorky 7 days ago • edited Owner
I don't think there's any error in the code, the version with spaces is correct. The version without spaces is incorrect. It appears to be a bug in the style-checking software. CC @menshikh-iv

So I changed to input space around `:`

zsef123:
@menshikh-iv Should I remove space before `:`? like `line[i: i + self.max_sentence_length]` 
This line passed Travis

menshikh-iv:
@zsef123 yep

menshikh-iv:
Very large and useful work, thank you very much @zsef123 :fire:

PR #1555, opened 2017-08-31
Fix Mallet wrapper & test for HDPTransform

menshikh-iv:


gensim/models/wrappers/ldamallet.py
```diff
@@ -183,7 +183,7 @@ def __getitem__(self, bow, iterations=100):
 
     def load_word_topics(self):
         logger.info("loading assigned topics from %s", self.fstate())
-        word_topics = numpy.zeros((self.num_topics, self.num_terms), dtype=numpy.float32)
+        word_topics = numpy.zeros((self.num_topics, self.num_terms), dtype=numpy.float64)
```
piskvorky: Is this really necessary? Double precision is rarely useful.

menshikh-iv: Double precision used for all TMs now.

piskvorky: Question is "why"?

menshikh-iv: I don't know, I don't implement TMs in gensim.

piskvorky: That doesn't sound right. I'm pretty sure many models used to use single precision (RP and LSI, for sure).

Can you check when the change to double happened, and why?

PR #1567, opened 2017-09-05
adding common terms to phrases model

alexgarel:
This is an implementation for #1258 (and a replacement to aborted PR #1263).

I've tested different implementation. This one in my view is effective and helps code maintenability.


alexgarel:
Sorry I skrew :-/ making it again !

PR #1570, opened 2017-09-05
Add verification when summarize_corpus returns null. Fix #1531.

### LINKED ISSUE #1531: 'NoneType' object is not iterable when using summarize
When using `gensim.summarization.summarize`, `TypeError: 'NoneType' object is not iterable` is occasionally thrown when summarizing random complaints in [this dataset of consumer complaints](https://catalog.data.gov/dataset/consumer-complaint-database). I'm not sure what it is about the strings that causes this error to be thrown but I feel the error should not be thrown regardless. Here is a self-contained example:

```
from gensim.summarization import summarize
text = "To whom it may concern, Only after weeks of using my new Wal-Mart master " \
    "card was I given the option to opt out of personal information sharing including :" \
    " Social Security #, income, account balance, payment history, credit history and credit" \
    " scores. According to the supervising operator at the wal-mart credit card call center my" \
    " personal information had already been shared. Furthermore, the letter informing me of my " \
    " rights was apart from my billing statement. Seemingly hidden between some flyers in the same" \
    " envelope. I almost threw it in the recycle bin without seeing it. If this is not illegal it " \
    "certainly appears to be an attempt to deceive me and disseminate my private information without " \
    "my permission. Thank you for your time and effort. Sincerely, XXXX XXXX"
summary = summarize(text)
```

I pasted the stack trace below but I'm sure you'll see it when running the above example.
```
<ipython-input-1-15020c19bb47> in <module>()
      1 from gensim.summarization import summarize
      2 text = "To whom it may concern, Only after weeks of using my new Wal-Mart master "     "card was I given the option to opt out of personal information sharing including :"     " Social Security #, income, account balance, payment history, credit history and credit"     " scores. According to the supervising operator at the wal-mart credit card call center my"     " personal information had already been shared. Furthermore, the letter informing me of my "     " rights was apart from my billing statement. Seemingly hidden between some flyers in the same"     " envelope. I almost threw it in the recycle bin without seeing it. If this is not illegal it "     "certainly appears to be an attempt to deceive me and disseminate my private information without "    "my permission. Thank you for your time and effort. Sincerely, XXXX XXXX"
----> 3 summary = summarize(text)

/usr/local/lib/python3.6/site-packages/gensim/summarization/summarizer.py in summarize(text, ratio, word_count, split)
    213 
    214     # Extracts the most important sentences with the selected criterion.
--> 215     extracted_sentences = _extract_important_sentences(sentences, corpus, most_important_docs, word_count)
    216 
    217     # Sorts the extracted sentences by apparition order in the original text.

/usr/local/lib/python3.6/site-packages/gensim/summarization/summarizer.py in _extract_important_sentences(sentences, corpus, important_docs, word_count)
    112 
    113 def _extract_important_sentences(sentences, corpus, important_docs, word_count):
--> 114     important_sentences = _get_important_sentences(sentences, corpus, important_docs)
    115 
    116     # If no "word_count" option is provided, the number of sentences is

/usr/local/lib/python3.6/site-packages/gensim/summarization/summarizer.py in _get_important_sentences(sentences, corpus, important_docs)
     87     hashable_corpus = _build_hasheable_corpus(corpus)
     88     sentences_by_corpus = dict(zip(hashable_corpus, sentences))
---> 89     return [sentences_by_corpus[tuple(important_doc)] for important_doc in important_docs]
     90 
     91 

TypeError: 'NoneType' object is not iterable
```

I'm assuming this is a bug but there's no example in the gensim ~~documentation~~ api page so I'm not entirely sure if I'm using the function correctly.

fbarrios:


gensim/summarization/summarizer.py
```diff
@@ -211,6 +212,11 @@ def summarize(text, ratio=0.2, word_count=None, split=False):
 
     most_important_docs = summarize_corpus(corpus, ratio=ratio if word_count is None else 1)
 
+    # If couldn't get important docs, the algorithm ends.
+    if not most_important_docs:
+        logger.warning("Couldn't get relevant sentences.")
+        return
```
piskvorky: Raising an exception better? I'm not sure whether this is an error state, or just a warning.

Many people don't have logging enabled, and the docstring implies the result of this function is a string (not `None`).

fbarrios: I don't think this is an error. Perhaps it makes even more sense to actually return the entire text as the summary wasn't possible, but this will break compatibility with the old behavior.

Regarding the docstring, the method returns a string or a list if the `split` parameter was set to true, so perhaps the best thing to do is:

```
if not most_important_docs:
    logger.warning("Couldn't get relevant sentences.")
    return [] if split else ""
```

piskvorky: That looks like a good solution 👍 
CC @menshikh-iv

menshikh-iv:
Thank you @fbarrios

Issue #1533, opened 2017-08-15
Phrases __getitem__() method does not respect chosen scoring function


SergeyDidenko:
#### Description
Phrases `__getitem__()` method does not respect chosen scoring function, it has 'default' scoring builtin. So when Phrases object is constructed with `scoring='npmi'` it returns wrong results.

See Phrases `export_phrases()` for correct implementation, i.e. `score = scoring_function(count_a, count_b, count_ab)` instead of `score = (pab - min_count) / pa / pb * len(vocab)`

For a while it's better to construct and use Phraser object, i.e. 
`phraser = Phraser[phrases]; res = phraser[sentences]` instead of  `res = phrases[sentences]`

piskvorky:
You're right, looks like a bug, thanks for reporting. Since you already have the solution, can you open a PR with the fix?

CC @michaelwsherman

SergeyDidenko:
Sorry, I don't have the fix yet. I just use Phraser.

michaelwsherman:
Good find--i totally forgot about getitem when I implemented this. I can get a fix together, should be fairly straightforward. Will be a few weeks though, and I assume this isn't a showstopper since Phraser still works.

@sergeididenko -- are you using npmi? That makes me happy.

SergeyDidenko:
@michaelwsherman Yes, I prefer MPI scores over alternatives. Can't say yet if it's better for my current project though.

michaelwsherman:
Fix in PR #1573 .

PR #1573, opened 2017-09-06
1533 fix and 1464 1423 comments

michaelwsherman:
Fix for #1533 , Phrases.__getitem__ now support custom scoring. Pluggable scoring via a function parameter is Phrases is now supported.

Made fixes in comments in PRs #1464 and #1423, except did not explicitly cast floats in the pre-defined scoring methods. Rather, float casting is now done before calling the scoring method.


michaelwsherman:
One unfortunate omission here is a lack of a test of pickling a Phrases and a Phraser object, to make sure pluggable scoring works properly after unpickling. I don't know enough about pickling to know what the risks here are, or if it is even worth testing this.

I'm worried that if a Phraser (or Phrases) object from an older version of gensim is loaded, it won't work now since older Phraser objects won't have an assigned scoring function and `utils.SaveLoad` is just pickling. I think the backwards compatibility could easily be handled (just a matter of checking for the presence of the scoring function and loading the right one if it isn't present), but I'm not sure of the best place to handle it from an architecture standpoint.

michaelwsherman:
I'm fairly certain that the switch the pluggable scoring will break backwards compatibility (although the switch to string-specified scoring probably did as well). Since we support saving and loading of Phrases and Phraser objects, do we need to implement something that "updates" older versions of Phraser and Phrases classes when they are loaded? If so, what's the best place to add this?

piskvorky:
@michaelwsherman thanks, much appreciated!

The best place to handle backward compatibility is inside `load`. That is, after loading an object, check whether the required attributes exist. If not, this implies an older version, so set the defaults that correspond to the "old" behaviour (=backward compatibility).

You can see that approach in action in word2vec: https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec.py#L1408

michaelwsherman:
I'm failing a test in `test_sklearn_api.py` related to pickling and unpickling (see the end of this comment). It's not surprising that pickling doesn't work, since scorer is now a function and functions don't pickle with default pickling.

The fix for this likely involves some custom pickling. My (brief) research has led me to the `__getstate__()` and `__setstate__()` methods ([as described in the python docs](https://docs.python.org/3/library/pickle.html#pickling-and-unpickling-normal-class-instances)). It looks like this needs to be done to support pluggable scoring through `utils.SaveLoad` as well. This work is in addition to the compatibility checks that need to be handled inside of load.

I can maybe embark on these fixes, but I'd like to know I'm on the right track before I go down this rabbit hole. It's also likely going to be a bit (a few weeks) for this fix, as I'll probably need a couple of days to figure out how this is done (mainly for the pickling part).

> ======================================================================
> ERROR: testPersistence (__main__.TestPhrasesTransformer)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "test_sklearn_api.py", line 946, in testPersistence
>     model_dump = pickle.dumps(self.model)
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 1380, in dumps
>     Pickler(file, protocol).dump(obj)
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 224, in dump
>     self.save(obj)
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 331, in save
>     self.save_reduce(obj=obj, *rv)
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 425, in save_reduce
>     save(state)
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 286, in save
>     f(self, obj) # Call unbound method with explicit self
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 655, in save_dict
>     self._batch_setitems(obj.iteritems())
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 669, in _batch_setitems
>     save(v)
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 331, in save
>     self.save_reduce(obj=obj, *rv)
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 425, in save_reduce
>     save(state)
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 286, in save
>     f(self, obj) # Call unbound method with explicit self
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 655, in save_dict
>     self._batch_setitems(obj.iteritems())
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 669, in _batch_setitems
>     save(v)
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 286, in save
>     f(self, obj) # Call unbound method with explicit self
>   File "C:\Users\msherman49\AppData\Local\Continuum\Anaconda3\envs\ml\lib\pickle
> .py", line 754, in save_global
>     (obj, module, name))
> PicklingError: Can't pickle <function original_scorer at 0x0000000006C71208>: it
> 's not found as gensim.models.phrases.original_scorer
> 
> ----------------------------------------------------------------------
>

piskvorky:
@michaelwsherman that doesn't sound right. Function are picklable no problem, they just need to be named functions (e.g. `gensim.models.phrases.original_scorer` is fine; `lambda x: x + 2` is not fine).

Since you're using named functions, I suspect the problem lies elsewhere.

michaelwsherman:
@piskvorky Sorry, you're right, and thank you for your quick response. This is a good learning experience for me.

I think it is because the [methods are static methods](https://stackoverflow.com/questions/1914261/pickling-a-staticmethod-in-python) in the `models.phrases.Phrases` class. I don't think there is any reason they have to be static methods in the class, so I'm making them regular methods in the base of `models.Phrases`. This may be even better, since any passed scoring method would not be in the Phrases class. All the tests pass if I do this.

I'm going to make this change, add support for the custom scoring to the scikit interface, and add tests for using a custom scorer (including persistence) to the scikit interface, add the save/load support for backwards compatibility, and throw together some tests for that as well. These changes may take a little bit.

I'm not going to add support for backwards compatibility through the scikit interface, as I expect that persistence via pickle across versions is not supported. Tell me if this is incorrect, and I'll start looking into the possibility of it.

gensim/models/phrases.py
```diff
@@ -177,9 +177,9 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
         # to still run the check of scoring function parameters in the next code block
         if type(scoring) is str:
```
piskvorky: `if isinstance(scoring, basestring)` more standard and future-proof.

gensim/models/phrases.py
```diff
@@ -64,8 +64,8 @@
 import warnings
 from collections import defaultdict
 import itertools as it
-from functools import partial
 from math import log
+from inspect import getargspec
```
piskvorky: Import not used?

gensim/models/phrases.py
```diff
@@ -354,15 +386,17 @@ def __getitem__(self, sentence):
...
+    # these two built-in scoring methods don't cast everything to float because the casting is done in the call
+    # to the scoring method in __getitem__ and export_phrases.
+
+
     # calculation of score based on original mikolov word2vec paper
-    # len_vocab and min_count set so functools.partial works
-    @staticmethod
-    def original_scorer(worda_count, wordb_count, bigram_count, len_vocab=0.0, min_count=0.0):
+def original_scorer(worda_count, wordb_count, bigram_count, len_vocab, min_count, corpus_word_count):
         return (bigram_count - min_count) / worda_count / wordb_count * len_vocab
```
piskvorky: Bad indent.

gensim/models/phrases.py
```diff
@@ -354,15 +386,17 @@ def __getitem__(self, sentence):
...
 
-    # normalized PMI, requires corpus size
-    @staticmethod
-    def npmi_scorer(worda_count, wordb_count, bigram_count, corpus_word_count=0.0):
+
+# normalized PMI, requires corpus size
+def npmi_scorer(worda_count, wordb_count, bigram_count, len_vocab, min_count, corpus_word_count):
         pa = worda_count / corpus_word_count
```
piskvorky: Bad indent.

gensim/models/word2vec.py
```diff
@@ -1577,23 +1577,23 @@ def __init__(self, source, max_sentence_length=MAX_WORDS_IN_BATCH, limit=None):
         self.limit = limit
 
         if os.path.isfile(self.source):
-            logging.warning('single file read, better to use models.word2vec.LineSentence')
+            logger.warning('single file read, better to use models.word2vec.LineSentence')
             self.input_files = [self.source]  # force code compatibility with list of files
         elif os.path.isdir(self.source):
             self.source = os.path.join(self.source, '')  # ensures os-specific slash at end of path
-            logging.debug('reading directory ' + self.source)
+            logger.warning('reading directory %s', self.source)
```
piskvorky: Why is this a warning?

gensim/models/word2vec.py
```diff
@@ -1563,7 +1563,7 @@ def __init__(self, source, max_sentence_length=MAX_WORDS_IN_BATCH, limit=None):
         """
         `source` should be a path to a directory (as a string) where all files can be opened by the
         LineSentence class. Each file will be read up to
-        `limit` lines (or no clipped if limit is None, the default).
+        `limit` lines (or not clipped if limit is None, the default).
```
piskvorky: The docs are not clear -- does the "will process all files in a directory" work recursively?

gensim/models/word2vec.py
```diff
@@ -1577,23 +1577,23 @@ def __init__(self, source, max_sentence_length=MAX_WORDS_IN_BATCH, limit=None):
         self.limit = limit
 
         if os.path.isfile(self.source):
-            logging.warning('single file read, better to use models.word2vec.LineSentence')
+            logger.warning('single file read, better to use models.word2vec.LineSentence')
```
piskvorky: If the class API contract supports it, this is no warning (maybe debug, at most).

If it's outside the API contract, this is an error and we should raise an exception, not log a warning.

gensim/models/phrases.py
michaelwsherman: Used in line 188 (in the commit your comments are on) to check for the proper parameters in the pluggable scoring function.

gensim/models/phrases.py
michaelwsherman: Sorry about these, very sloppy on my part.

gensim/models/word2vec.py
michaelwsherman: I made a mistake, changing to info.

gensim/models/word2vec.py
michaelwsherman: It does not. Maybe wishlist? I've clarified the docs.

gensim/models/word2vec.py
michaelwsherman: Clarified this message a bit, made it debug.

gensim/models/phrases.py
```diff
@@ -177,6 +176,13 @@ def __init__(self, sentences=None, min_count=5, threshold=10.0,
         # set scoring based on string
         # intentially override the value of the scoring parameter rather than set self.scoring here,
         # to still run the check of scoring function parameters in the next code block
+
+        # for python 2 and 3 compatibility. basestring is used to check if scoring is a string
```
piskvorky: Almost there :) We use `six` in gensim for py2/py3 compatibility, so `isinstance` on `six.string_types` is probably what we want.

gensim/models/phrases.py
piskvorky: Thanks, I see it now. What is that check for though? Python is duck-typed by convention, so "type checks" are best postponed until truly needed (something breaks).

michaelwsherman: Mostly to save the stress that would result from improperly specifying a scoring function when initializing the phrases object. I know Python will do the type checking when the scoring function is called, but that won't happen until export_phrases or __getitem__ is called. The "normal" workflow for the Phrases object is to just specify sentences on load, or to use add_vocab. Only after that does the scoring function get called.

I could easily see a user specifying a bad scoring method and then making the vocab dictionary from their large corpus. Only after significant time extracting vocab from a corpus do they then discover that something is wrong with how they specified scoring. At this point you could manually specify a correct scoring function, but that requires you to set it directly. Users also wouldn't have an easy bailout in the form of use one of the scorer string settings, since those are only checked when the Phrases object is created--the user would have to figure out how to specify those built in scorers which would mean opening up the code. This seems a bit user unfriendly, I feel it is friendlier to just do the type checking on initialization even if it is less Pythonic.

This could be fixed with a set_scorer method that takes the string or function input, but that seems a bit more awkward than just doing this type check.

There's also an issue with wanting to raise an informative exception when the scoring function is called in __getitem__ or export_phrases and the types don't match, but that means adding a try/except into the main scoring loop and that seems awkward as well. I think its better to just do that try/except once when the object is initialized.

But I defer to your judgement on this--what do you think is best?

piskvorky: Thanks, I see your argument (that checking early a little more convenient).

I'm not sure if it's worth it, but don't care much either way. I'll defer to @menshikh-iv :)

michaelwsherman:
This should be ready to go now, the requested changes have been made (or discussed) and everything is up to date with develop.

menshikh-iv:
Nice work @michaelwsherman :fire:

PR #1574, opened 2017-09-06
fixing input

tebandesade:
I was getting a TypeError: can't concat bytes to str 
because of this line self.wv.syn0[i] = self.seeded_vector(self.wv.index2word[i] + str(self.seed))
and the input was b"''"

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1163,7 +1163,8 @@ class Word2Vec(utils.SaveLoad):
         # randomize weights vector by vector, rather than materializing a huge random matrix in RAM at once
         for i in xrange(len(self.wv.vocab)):
             # construct deterministic seed from word AND seed argument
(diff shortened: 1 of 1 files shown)


gojomo:
The test failure is in exactly the line changed... so unless there's a case the the test should be changed, this (seemingly harmless) fix is breaking previously-working behavior.

menshikh-iv:
@gojomo @tebandesade what's status here? Should I close this PR?

gojomo:
I don't know the bug this is supposed to fix, *and* the code breaks an existing test. So without more details/fixes it should be set aside.

menshikh-iv:
Thanks @gojomo 
@tebandesade can you describe more concretely what's a motivation of this fix (describe concrete bug)?

menshikh-iv:
I close this pull request because it looks abandoned

PR #1579, opened 2017-09-08
Summarize raises a ValueError when appropriate

diegospd:
The `summarize` function used to just fail when the text had less than 10 sentences. Not it warns about the problem via logging, and it also raises an appropriate exception


gensim/summarization/summarizer.py
```diff
@@ -197,12 +197,11 @@ def summarize(text, ratio=0.2, word_count=None, split=False):
 
     # If no sentence could be identified, the function ends.
     if len(sentences) == 0:
-        logger.warning("Input text is empty.")
-        return [] if split else u""
+        raise ValueError("No sentences could be identified, maybe input text is empty")
```
menshikh-iv: Add test for this case & fix current test (same for all `ValueError`)

menshikh-iv:
Ping @diegospd, how is going?

menshikh-iv:
I close this PR because it's abandoned.

PR #1580, opened 2017-09-08
Temporary workaround for bug #1531.

diegospd:
When the bug is encountered a warning will be logged and the original text will be returned as if it had been succesfully summarized.


menshikh-iv:
Very strange "workaround", I think this isn't needed here.

PR #1582, opened 2017-09-11
Add "most_similar_to_given" method for KeyedVectors

TheMathMajor:
Added a function to find the most similar word in a given list to a given word.


.spyproject/codestyle.ini
```diff
@@ -0,0 +1,6 @@
+[codestyle]
```
menshikh-iv: What is it? Please remove all non-relevant files (all from `.spyproject` folder).

gensim/models/keyedvectors.py
```diff
@@ -617,6 +618,22 @@ def similarity(self, w1, w2):
 
         """
         return dot(matutils.unitvec(self[w1]), matutils.unitvec(self[w2]))
+    
+    def most_similar_to_given(self, w1, word_list):
```
menshikh-iv: Please reformat your docstring according to [google-style](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

gensim/models/keyedvectors.py
```diff
@@ -617,6 +618,22 @@ def similarity(self, w1, w2):
...
+        """
+        Return the word from word_list most similar to w1.
+        
+        Example::
+            
+          >>> trained_model.most_similar_to_given('music', ['water', 'sound', 'backpack', 'mouse'])
```
menshikh-iv: What do you this @gojomo, it's a useful feature?

menshikh-iv:
@TheMathMajor  Also, please fix PEP8 issues (look at [travis log](https://travis-ci.org/RaRe-Technologies/gensim/jobs/274089780#L1021))

gojomo:
There's another nearly-complete implementation of similar functionality by @shubhvachher in closed-as-idle PR #1229.

gensim/models/keyedvectors.py
gojomo: If done efficiently it makes sense. If the common case is there's a stable subset against which some projects are doing all of their similarity-searches, the need might be best met by a new method for creating a subset KeyedVectors, with just the words-of-interest. Overlaps with closed-as-idle PR #1229.

menshikh-iv:
Ping @TheMathMajor, what's a status of this PR?

TheMathMajor:
Hi, thanks for the feedback, I have made the committed the changes requested.

menshikh-iv:
Thanks @TheMathMajor LGTM
@gojomo should I merge this change (or you have any suggestions)?

gojomo:
There's no need for "deprecated" forwarding-method in Word2Vec if this is a brand-new feature on `KeyedVectors` that no one has ever learned to call on `Word2Vec`. 

Perhaps the method should have a test, but as a simple 1-liner composed of other well-tested methods, maybe not. 

But that highlights another difference with the earlier #1229 – while that PR had a lot of code-duplication, it did try to do the similarity calculations with array math, and thus might be noticeably faster with long word-lists. If main goal is performance, that approach may have been better; if goal is simply providing a convenience/clarity/example-method, this idiomatic 1-liner is better.

menshikh-iv:
> There's no need for "deprecated" forwarding-method in Word2Vec if this is a brand-new feature on KeyedVectors that no one has ever learned to call on Word2Vec.

Agree, I'll remove it from Word2Vec

For this method, I think clean one-liner is better (IMO we no need performance here)

gojomo:
I think the people who brought this up in #1229 and #481 *were* concerned about performance in their projects, which is why they pursued that path. But seems OK to add the clear 1-liner to start, and see who uses-and-needs-better, thus perhaps contributing a faster alternative.

menshikh-iv:
Thanks @TheMathMajor, congratz with the first contribution:1st_place_medal:

TheMathMajor:
Thanks a lot for the suggestions and guiding me through my first contribution!

PR #1577, opened 2017-09-08
Add preprocessing step to remove \n from the text before summarizing. Fix 1575.

diegospd:
As reported in issue #1575 `_clean_text_by_sentences` cannot properly handle `\n` in the middle of sentences. This PR adds a preprocessing step to remove them before doing anything else.

Fixes issue #1575. 

No tests were written

PR #1590, opened 2017-09-17
Newlines are now ignored while splitting by sentences in summarizing. Fix #1575.

### LINKED ISSUE #1575: Add preprocessing in summarize function to remove newlines in the middle of sentences
#### Description
The `summarize` function depends on the text having at least 10 sentences as measured by `clean_text_by_sentences`.  If the text is shorter than that then the summarization fails in an undocumented way. Moreover  `clean_text_by_sentences` cannot handle properly a text with new lines
at the middle of a sentence. I suggest a preprocessing step to purge those.

I'm currently using this to workaround this bug
```
import re
text = re.sub(r'\n|\r|\t', ' ', text)
text = re.sub(r'\s+', ' ', text)
```

#### Versions
- Python 3.6.2 (default, Jul 17 2017, 16:44:45)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)]
- NumPy 1.13.0
- SciPy 0.19.1
- gensim 2.3.0
- FAST_VERSION 0

fbarrios:
This is an improvement over #1577 for fixing #1575, as this doesn't need to process all text at once for removing newlines.

### DIFF
--- a/gensim/test/test_data/mihalcea_tarau.txt
+++ b/gensim/test/test_data/mihalcea_tarau.txt
@@ -1,11 +1,11 @@
-AP880911-0016
-AP-NR-09-11-88 0423EDT r i
-BC-HurricaneGilbert 09-11 0339
-BC-Hurricane Gilbert,0348
-Hurricane Gilbert Heads Toward Dominican Coast
-By RUDDY GONZALEZ
-Associated Press Writer
-SANTO DOMINGO, Dominican Republic (AP)
+AP880911-0016.
+AP-NR-09-11-88 0423EDT r i.
+BC-HurricaneGilbert 09-11 0339-
+BC-Hurricane Gilbert,0348.
+Hurricane Gilbert Heads Toward Dominican Coast.
+By RUDDY GONZALEZ.
+Associated Press Writer.
+SANTO DOMINGO, Dominican Republic (AP).
 Hurricane Gilbert swept toward the Dominican Republic Sunday, and the Civil Defense alerted its heavily populated south coast to prepare for high winds, heavy rains and high seas.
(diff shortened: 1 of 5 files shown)


gensim/test/test_data/mihalcea_tarau.txt
```diff
@@ -1,11 +1,11 @@
-AP880911-0016
-AP-NR-09-11-88 0423EDT r i
-BC-HurricaneGilbert 09-11 0339
-BC-Hurricane Gilbert,0348
-Hurricane Gilbert Heads Toward Dominican Coast
-By RUDDY GONZALEZ
-Associated Press Writer
-SANTO DOMINGO, Dominican Republic (AP)
+AP880911-0016.
+AP-NR-09-11-88 0423EDT r i.
```
menshikh-iv: Without dots, this doesn't work? You think that this is a good solution?

fbarrios: Not really, but this was reported in #1575 and I created this PR as an improvement in performance over #1577.

I do believe that we should at least write about this behavior in the docstring.

menshikh-iv:
Ping @fbarrios, how is going?

menshikh-iv:
ping @fbarrios, what's status here?

fbarrios:
Sorry @menshikh-iv, I've been busy this week. I'll look into this in the next few days.

menshikh-iv:
Ping @fbarrios

fbarrios:
@menshikh-iv I think the correct behaviour is the current one, so this PR should be closed, so please do if you agree.

I created #1702 for updating the docstring with the current behaviour.

PR #1599, opened 2017-09-25
Fix scan vocab speed issue, build vocab from provided word frequencies

jodevak:
This request has two parts:
1-There was a noticeable speed issue with scan_vocab, and it turned out to be "sum(itervalues(vocab))" because this will iterate through the whole vocab on each completed "progress_per" iterations, which has a high speed cost when dealing with a big vocab. It took only 45 mins to iterate and build the whole vocab on 57 Gigabyte production ready words co-occurrences (window=1) with my modification, whereas it took 270 mins using old implementation.

2- Since build vocab is a single threaded operation, it would be very helpful to have a function that builds a word vocabulary from pre-given word frequencies (build_vocab_from_freq function), for example one could use Spark to count the words in a distributed way and then pipe the word frequencies to gensim word2vec.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -114,16 +114,18 @@ try:
 except ImportError:
     from Queue import Queue, Empty
 
-from numpy import exp, log, dot, zeros, outer, random, dtype, float32 as REAL,\
-    uint32, seterr, array, uint8, vstack, fromstring, sqrt,\
-    empty, sum as np_sum, ones, logaddexp
+from numpy import exp, log, dot, zeros, outer, random, dtype, float32 as REAL, \
+    double, uint32, seterr, array, uint8, vstack, fromstring, sqrt, newaxis, \
+    ndarray, empty, sum as np_sum, prod, ones, ascontiguousarray, vstack, logaddexp
 
 from scipy.special import expit
 
 from gensim import utils, matutils  # utility fnc for pickling, common scipy operations etc
+from gensim.corpora.dictionary import Dictionary
 from six import iteritems, itervalues, string_types
 from six.moves import xrange
 from types import GeneratorType
+from scipy import stats
 
 logger = logging.getLogger(__name__)
 
@@ -135,6 +137,7 @@ except ImportError:
     # failed... fall back to plain numpy (20-80x slower training than the above)
     FAST_VERSION = -1
     MAX_WORDS_IN_BATCH = 10000
+
 
     def train_batch_sg(model, sentences, alpha, work=None, compute_loss=False):
         """
@@ -150,7 +153,7 @@ except ImportError:
         result = 0
         for sentence in sentences:
             word_vocabs = [model.wv.vocab[w] for w in sentence if w in model.wv.vocab and
-                           model.wv.vocab[w].sample_int > model.random.rand() * 2**32]
+                           model.wv.vocab[w].sample_int > model.random.rand() * 2 ** 32]
             for pos, word in enumerate(word_vocabs):
                 reduced_window = model.random.randint(model.window)  # `b` in the original word2vec code
 
@@ -159,12 +162,12 @@ except ImportError:
                 for pos2, word2 in enumerate(word_vocabs[start:(pos + model.window + 1 - reduced_window)], start):
                     # don't train on the `word` itself
                     if pos2 != pos:
-                        train_sg_pair(
-                            model, model.wv.index2word[word.index], word2.index, alpha, compute_loss=compute_loss
-                        )
+                        train_sg_pair(model, model.wv.index2word[word.index], word2.index, alpha,
+                                      compute_loss=compute_loss)
 
             result += len(word_vocabs)
         return result
+
 
     def train_batch_cbow(model, sentences, alpha, work=None, neu1=None, compute_loss=False):
         """
@@ -180,7 +183,7 @@ except ImportError:
         result = 0
         for sentence in sentences:
             word_vocabs = [model.wv.vocab[w] for w in sentence if w in model.wv.vocab and
-                           model.wv.vocab[w].sample_int > model.random.rand() * 2**32]
+                           model.wv.vocab[w].sample_int > model.random.rand() * 2 ** 32]
             for pos, word in enumerate(word_vocabs):
                 reduced_window = model.random.randint(model.window)  # `b` in the original word2vec code
                 start = max(0, pos - model.window + reduced_window)
@@ -192,6 +195,7 @@ except ImportError:
                 train_cbow_pair(model, word, word2_indices, l1, alpha, compute_loss=compute_loss)
             result += len(word_vocabs)
         return result
+
 
     def score_sentence_sg(model, sentence, work=None):
         """
@@ -221,6 +225,7 @@ except ImportError:
                     log_prob_sentence += score_sg_pair(model, word, word2)
 
         return log_prob_sentence
+
 
     def score_sentence_cbow(model, sentence, work=None, neu1=None):
         """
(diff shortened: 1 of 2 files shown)


gensim/models/word2vec.py
```diff
@@ -615,12 +615,26 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
         """
         Build vocabulary from a sequence of sentences (can be a once-only generator stream).
         Each sentence must be a list of unicode strings.
-
         """
         self.scan_vocab(sentences, progress_per=progress_per, trim_rule=trim_rule)  # initial survey
         self.scale_vocab(keep_raw_vocab=keep_raw_vocab, trim_rule=trim_rule, update=update)  # trim by min_count & precalculate downsampling
         self.finalize_vocab(update=update)  # build tables & arrays
 
+
```
menshikh-iv: Too many blank lines

gensim/models/word2vec.py
```diff
@@ -615,12 +615,26 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
...
+    def build_vocab_from_freq(self, word_freq, keep_raw_vocab=False, corpus_count=None, trim_rule=None, update=False):
+        """
```
menshikh-iv: Please add documentation in [numpy-style](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)

gensim/models/word2vec.py
```diff
@@ -615,12 +615,26 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
...
         Build vocabulary from a sequence of sentences (can be a once-only generator stream).
         Each sentence must be a list of unicode strings.
-
         """
         self.scan_vocab(sentences, progress_per=progress_per, trim_rule=trim_rule)  # initial survey
         self.scale_vocab(keep_raw_vocab=keep_raw_vocab, trim_rule=trim_rule, update=update)  # trim by min_count & precalculate downsampling
         self.finalize_vocab(update=update)  # build tables & arrays
 
+
+    def build_vocab_from_freq(self, word_freq, keep_raw_vocab=False, corpus_count=None, trim_rule=None, update=False):
```
menshikh-iv: Add test for this method

gensim/models/word2vec.py
```diff
@@ -629,32 +643,28 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
...
+                        "Each 'sentences' item should be a list of words (usually unicode strings)."
+                        "First item here is instead plain %s.", type(sentence)
                     )
                 checked_string_types += 1
             if sentence_no % progress_per == 0:
-                logger.info(
-                    "PROGRESS: at sentence #%i, processed %i words, keeping %i word types",
-                    sentence_no, sum(itervalues(vocab)) + total_words, len(vocab)
-                )
+                logger.info("PROGRESS: at sentence #%i, processed %i words, keeping %i word types",
```
menshikh-iv: Please use hanging indents

piskvorky:
@jodevak I find 1) weird. Summing a bunch of values should be very fast, no matter the dictionary size.

What was your `progress_per`, how often did this operation take place (in seconds)?

jodevak:
@piskvorky
1- My progress per was 1000, the vocab size was almost 2700000, and the total word co-occurrences were two and a half billion. I haven't measured the time for this operation alone, i just compared the total run time using both implementation.

2- Summing the values will require iterating over the whole dictionary values,  in other words it means iterating over the whole stored word counts which will be definitely slower than incrementing a single counter.

@menshikh-iv 
1- In progress :+1:

piskvorky:
Seems the `progress_per` is too low, that's not its intended use-case. What is the reason for this?

Btw we'll be replacing all the counting stuff by Bounter, so this will be moot.

Only needs some code style fixes (vertical indent) otherwise LGTM 👍

jodevak:
@piskvorky i edited the common, origress per is 10000 which is the default value.

piskvorky:
A sum of 2,700,000 dict values shouldn't take more than a few dozen milliseconds (and it's only once every few seconds). Weird... but a timing is a timing!

In any case, Bounter keeps a `.total` tally for free, so this will be irrelevant.

jodevak:
@piskvorky Yes, using bounter would be more elegant. And in order to understand the speed problem, consider this code.
`from time import time
from collections import defaultdict

counter=0

g=defaultdict(int,dict(enumerate(range(2700000))))

stime=time()

sum(itervalues(g))

ftime=time()

h=((ftime-stime)/60)*(2500000000/10000) 

print h



stime=time()

counter+=1

ftime=time()

h=((ftime-stime)/60)*(2500000000/10000) 

print h
`

gensim/models/word2vec.py
```diff
@@ -329,7 +333,8 @@ def train_sg_pair(model, word, context_index, alpha, learn_vectors=True, learn_h
     return neu1e
 
 
-def train_cbow_pair(model, word, input_word_indices, l1, alpha, learn_vectors=True, learn_hidden=True, compute_loss=False,
+def train_cbow_pair(model, word, input_word_indices, l1, alpha, learn_vectors=True, learn_hidden=True,
```
menshikh-iv: Please use hanging indents (instead of vertical), here and anywhere

jodevak: "When using a hanging indent the following should be considered; there should be no arguments on the first line and further indentation should be used to clearly distinguish itself as a continuation line."

Is this what you need ? if yes, do you recommend any tool other than autopep8 that would auto format the file ?

menshikh-iv: Sorry, incorrect line
vertical indents OK for function/method definition, that's all. For other situations, in gensim, we used hanging indent.

Unfortunately, I can't recommend any tool for it (because we have no tool for check this condition yet), only manually

jodevak:
@menshikh-iv , Is indentation accepted now ? if yes, i think only a test for build_vocab_from_freq is left to include. Many thanks.

menshikh-iv:
@jodevak yes, please make and that's all :+1:

menshikh-iv:
Congratz with first contribution @jodevak :1st_place_medal:

jodevak:
@menshikh-iv Thanks :D

gensim/models/word2vec.py
```diff
@@ -641,16 +678,16 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
             if sentence_no % progress_per == 0:
                 logger.info(
                     "PROGRESS: at sentence #%i, processed %i words, keeping %i word types",
-                    sentence_no, sum(itervalues(vocab)) + total_words, len(vocab)
+                    sentence_no, total_words, len(vocab)
                 )
             for word in sentence:
                 vocab[word] += 1
+                total_words += 1
```
piskvorky: This is not a good idea, will be slow. Add the entire `len(sentence)` at once.

gensim/models/word2vec.py
```diff
@@ -641,16 +678,16 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
...
 
             if self.max_vocab_size and len(vocab) > self.max_vocab_size:
-                total_words += utils.prune_vocab(vocab, min_reduce, trim_rule=trim_rule)
+                utils.prune_vocab(vocab, min_reduce, trim_rule=trim_rule)
```
piskvorky: I don't see any tests for this change during pruning, seems risky. Does it really work?

gensim/models/word2vec.py
```diff
@@ -615,12 +615,49 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
...
+        `update`: bool
+            If true, the new provided words in `word_freq` dict will be added to model's vocab.
+
+        Returns
+        --------
+        None
+
+        Examples
+        --------
+        >>> build_vocab_from_freq({"Word1":15,"Word2":20}, update=True)
```
piskvorky: Code style: PEP8. Also, this is an instance method (cannot be called without an object).

gensim/models/word2vec.py
```diff
@@ -615,12 +615,49 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
...
+        """
+        logger.info("Processing provided word frequencies")
```
piskvorky: Be more concrete in the log: what was provided? (how many entries, total frequencies?)

gensim/models/word2vec.py
```diff
@@ -615,12 +615,49 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
...
+        vocab = defaultdict(int, word_freq)
```
piskvorky: Won't this duplicate (double) the entire dictionary? Is it backward compatible in the sense that this refactoring won't consume much more memory?

gensim/models/word2vec.py
```diff
@@ -615,12 +615,49 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
...
+
+        self.corpus_count = corpus_count if corpus_count else 0
+        self.raw_vocab = vocab
+
+        self.scale_vocab(keep_raw_vocab=keep_raw_vocab, trim_rule=trim_rule, update=update)  # trim by min_count & precalculate downsampling
```
piskvorky: This function could use some comments and invariants: what's the relationship between `vocab` vs `raw_vocab` vs `word_freq`?

gensim/models/word2vec.py
jodevak: Duplicating the entire vocab ? its just assigning a ready raw vocab (word count) dictionary. Is there a part im not getting ?

gensim/models/word2vec.py
jodevak: word_freq is the same as raw_vocab, Vocab is the same as word freq, so yes i think i should use a different naming.

gensim/models/word2vec.py
jodevak: hmmm although it wont noticeably affect the speed, but yes it should be incrementing at once :+1:

gensim/models/word2vec.py
jodevak: hmmm do you really think it needs a new test ? prunce_vocab has not been touched only the counter

piskvorky: Yes, definitely. You changed the semantics of how the `total_words` works; for example, the return value of `prune_vocab` is ignored.

It may be correct, but is not obvious to me and deserves a test.

gensim/models/word2vec.py
piskvorky: I don't think so. The constructor will copy the entire contents of `word_freq`, which may be memory intensive for large vocabularies.

menshikh-iv:
can you create new PR and fix all comments @jodevak?

jodevak:
@menshikh-iv sure

menshikh-iv:
Continued in #1695.

PR #1605, opened 2017-10-02
[WIP] Refactor API reference gensim.corpora

anotherbugmaster:
Issue #1671


gensim/corpora/bleicorpus.py
```diff
@@ -85,15 +97,21 @@ def line2doc(self, line):
         return doc
 
     @staticmethod
-    def save_corpus(fname, corpus, id2word=None, metadata=False):
+    def _save_corpus(fname, corpus, id2word=None, metadata=False):
```
piskvorky: This is an API breaking change (implies a major version increment); why do we want this?

Did you catch all `save_corpus` instances (I'd expect there to be more calls, in different modules, but I'm not sure).

anotherbugmaster: @piskvorky Hi. The reason is these lines in docstrings:

```
-        This function is automatically called by `BleiCorpus.serialize`; don't
-        call it directly, call `serialize` instead.
```

If we shouldn't call it directly, then why keep it public?

As for the second question, yes, I think that I caught all of them.

piskvorky: Yes, I guess you're right.

I think we should keep the old method as "deprecated" for a while, for a few releases, before removing it. WDYT @menshikh-iv ?

menshikh-iv: @piskvorky "deprecation warning" is good, but how we should detect who called this function? If user call this - we should raise deprecation warning, but if this method called from any internal function - deprecation warning shouldn't be raised.
I know 2 solutions:

Add warning, look into "callstack" and see, who called current function and raise warning by condition (black magic, I don't think that this is a good idea);
Add warning and suppress this warning in each "inner" call (this variant is better, but need to patch all "call places").

piskvorky: No internal function should be calling this. They should all use the "new" (non-deprecated) calling style.

anotherbugmaster:
Switched to another branch (fix_1605)

PR #1606, opened 2017-10-02
Retain the raw vocabulary when you feed several data respectively

shinochin:
When you have several data the sizes of which are so large, it is difficult to feed all the data to the model at the same time.

If you feed several data respectively, it is needed to retain the raw vocabulary to see if the number of occurrences of each infrequent word which is included in data is larger than `min_count`.

Thus, I sent this pull request.

If these commits are added, we can do this operation without discarding infrequent words the sizes of which are larger than `min_count` globally but smaller locally.

```python
model = word2vec.Word2Vec(window=1, min_count=1)

def load(file):
    ~
    hoge
    ~
    return sentences

files=["data0.txt", "data1.txt", "data2.txt", ...]
update=False
for file in files:
    sentences = load(file)
    model.build_vocab(sentences, update=update)
    update=True
for file in files:
    sentences = load(file)
    model.train(sentences, total_examples=len(sentences),
                epochs=model.iter)
```

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -663,7 +663,8 @@ class Word2Vec(utils.SaveLoad):
         
     def reset_raw_vocab(self):
         """
-        After building vocabulary, Delete the raw vocabulary to free up RAM
+        Delete the raw vocabulary to free up RAM after building vocabulary completely
+        if `keep_raw_vocab` is set.
         """
         if hasattr(self, 'raw_vocab'):
             self.raw_vocab = defaultdict(int)
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -627,7 +627,10 @@ class Word2Vec(utils.SaveLoad):
         sentence_no = -1
         total_words = 0
         min_reduce = 1
-        vocab = defaultdict(int)
+        if not hasattr(self, 'raw_vocab'):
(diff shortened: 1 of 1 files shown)


gojomo:
It's usually cleaner and less-error-prone to create a single corpus iterator that streams all the combined data to a single `build_vocab()` call. There's an example of a utility class using this technique to virtually-concatenate multiple file in the `PathLineSentences` class:

https://github.com/RaRe-Technologies/gensim/blob/09fddf5c1215fe94f35f46d16747b4ce6c8b32f0/gensim/models/word2vec.py#L1630

This will generally also be better for a making a single call to `train()` - the loop you've shown, where `train()` is called separately for each file, will complete training on one file (from starting `alpha` to `min_alpha` over many iterations) before then doing many-iterations, with alpha going from its max to min, on the next file. Neither the completion of one file before any other are considered, nor the 'saw-tooth' jumping up-and-down of the effective 'alpha' learning-rate, are proper stochastic-gradient-descent and would be likely to get much worse final results.

Because the combined re-iterable full corpus is preferred, we'd probably not want to adapt `build_vocab()` to work this way.

shinochin:
@gojomo Thanks for your kind reply. I understood the circumstances. I will use `PathLineSentences` from now on. I greatly appreciate your kindness.

PR #1607, opened 2017-10-03
[WIP] Subpackage refactor. Fix 1584

### LINKED ISSUE #1584: Remove/refactor useless subpackages
In gensim, we have many sub-packages, but several of this should be a part of another subpackage, another part is broken/useless and should be removed.

Candidates:
`/examples` - Old broken code + confused users with name, should be removed.
`/parsing` - Not relevant for gensim (all of this already implemented in NLTK/etc.), should be removed (but before, need to check that this isn't used).
`/scripts` - Many scripts for Wikipedia + symlinks + some conversions + w2v_standalone. Need to look into all wiki scripts and understand why this needed, remove all that no needed (w2v_standalone too).
`/summarization` - need to refactor code and create one model, no need distinct subpackage for this.
`/topic_coherence` - same as `summarization`

menshikh-iv:
Fixes #1584.


menshikh-iv:
@macks22 please review my commit about [CoherenceModel refactoring](https://github.com/RaRe-Technologies/gensim/pull/1607/commits/50d7592d8b419356850c75482cb77b64963b26fe), maybe you have any suggestions about it.

gensim/models/coherence_utils.py
```diff
@@ -4,25 +4,539 @@
...
+from gensim.utils import iter_windows, flatten
 from gensim.models.word2vec import Word2Vec
+from scipy import sparse as sps
+from six import string_types, iteritems
 
 logger = logging.getLogger(__name__)
+EPSILON = 1e-12  # Should be small. Value as suggested in paper.
+
+
+"""
```
macks22: The presence of these docstrings throughout the module makes me think the existing module structure was actually a pretty useful way to separate these things out. Have you considered adding a `coherence_utils` package with `from <existing_module> import *` instead of throwing them all in the same module? This seems cleaner to me. I think it makes things easier to find and navigate.

gensim/utils.py
```diff
@@ -1278,3 +1277,456 @@ def lazy_flatten(nested_list):
                 yield sub
         else:
             yield el
+
+
+class PorterStemmer(object):
```
macks22: I think a similar comment applies here to my comment above about the `coherence_utils` module. I think this would be better organized as a package, with subpackages for `preprocessing`, `corpora`, etc. I think the existing `utils` module was already getting a bit bloated. I would hypothesize very long files with a wide variety of things makes it hard to find things for folks who are new to the codebase. I might even advocate a more granular decomposition such as: `utils/stemming`, `utils/lemmatizing`, `utils/corpora`, `utils/helpers`, or something of that sort, with aggregation in the `utils/__init__`.

gensim/utils.py
```diff
@@ -1278,3 +1277,456 @@ def lazy_flatten(nested_list):
...
+    'your', 'yours', 'yourself', 'yourselves'
+})
+
+
+def remove_stopwords(s):
+    s = to_unicode(s)
+    return " ".join(w for w in s.split() if w not in STOPWORDS)
+
+
+RE_PUNCT = re.compile('([%s])+' % re.escape(string.punctuation), re.UNICODE)
```
macks22: What are your thoughts on moving this into some kind of namespace object called `regexes`, or something of that sort, either as a module or a class with these as class attributes?

The dedicated namespace would make it easier to find these, provide a single place for users to look for existing regex patterns, and also make it easier to document them. The downside I see is moving their definitions away from where they are used.

macks22:
@menshikh-iv I left a few comments, all regarding modularity. I really like the work you've been doing to make the code better organized and more consistently styled. I think this is a good direction, and I hope my comments are useful. Cheers!

gensim/models/coherence_utils.py
menshikh-iv: So, maybe simple moving & renaming `coherence_model->gensim.models.coherence_utils` OR `gensim.models.coherence_inner`is better and you are right (because I don't think that this variant is sufficiently consistent), for this reason, I'm asking your for review, thanks :+1:

gensim/utils.py
menshikh-iv: I agree with you, maybe simple storage class will be better for this purposes.

gensim/utils.py
menshikh-iv: In general, I agree (maybe not very "detailed" utils, but splitting `gensim.utils` to `gensim.utils.utils` (old) and `gensim.utils.preprocessing` is better)

menshikh-iv:
Continued in #1618

PR #1618, opened 2017-10-10
Subpackage refactor. Fix 1584

menshikh-iv:
Fix for #1584.


menshikh-iv:
I re-create PR according to discussion from #1607 and #1584, please review again @piskvorky @macks22

piskvorky:
Regarding the various `make_wiki_*.py` duplicates: these were originally symlinks. They became duplicates (regular files) only by error.

Removing them will break some functionality, because the script behaves differently based on what filename it was run as.

These parameters are probably better served as proper CLI options, rather than encoded in the filename. Just saying that simply removing the "duplicates" will prevent users from accessing existing functionality.

menshikh-iv:
@piskvorky but how implemented different behavior if all files the same? I don't understand it. Can you show me an example?

menshikh-iv:
Close (because conflict is really big), I'll re-create it with additions when we plan to create major release

PR #1619, opened 2017-10-10
Add Sent2Vec model. Fix #1376

### LINKED ISSUE #1376: sent2vec in Gensim
The EPFL ML group have come up with a sentence embedding technique, [sent2vec](https://arxiv.org/abs/1703.02507). The repo with the code can be found [here](https://github.com/epfml/sent2vec).

It would be interesting to have this feature in Gensim, either with a wrapper or otherwise.

prerna135:
Rough initial code for sent2vec.


gensim/models/sent2vec.py
```diff
@@ -0,0 +1,132 @@
...
+
+from gensim.models.word2vec import Word2Vec
+from gensim.models.fasttext import FastText
+from gensim.models.wrappers.fasttext import FastTextKeyedVectors
+from gensim.models.wrappers.fasttext import FastText as Ft_Wrapper
+from random import randint
+
+logger = logging.getLogger(__name__)
+
+MAX_WORDS_IN_BATCH = 10000
```
menshikh-iv: You can import this constant instead of explicit definition

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,132 @@
...
+
+class Sent2Vec(FastText):
+    def __init__(
+            self, sentences=None, sg=0, hs=0, size=100, alpha=0.2, window=5, min_count=5,
+            max_vocab_size=None, word_ngrams=2, loss='ns', sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
+            negative=5, cbow_mean=1, hashfxn=hash, iter=5, null_word=0, min_n=3, max_n=6, sorted_vocab=1, bucket=2000000,
+            trim_rule=None, batch_words=MAX_WORDS_IN_BATCH, dropoutK=2):
```
menshikh-iv: All parameters should be in lowercase (`dropoutK`)

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,132 @@
...
+
+from __future__ import division
+import logging
+from collections import defaultdict
+from six import itervalues, string_types
+import numpy as np
+from numpy import dot
+from gensim import utils, matutils
+
+from gensim.models.word2vec import Word2Vec
```
menshikh-iv: Useless import

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,132 @@
...
+
+class Sent2Vec(FastText):
+    def __init__(
+            self, sentences=None, sg=0, hs=0, size=100, alpha=0.2, window=5, min_count=5,
+            max_vocab_size=None, word_ngrams=2, loss='ns', sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
+            negative=5, cbow_mean=1, hashfxn=hash, iter=5, null_word=0, min_n=3, max_n=6, sorted_vocab=1, bucket=2000000,
+            trim_rule=None, batch_words=MAX_WORDS_IN_BATCH, dropoutK=2):
+
+        # sent2vec specific params
+        #dropoutK is the number of ngrams dropped while training a sent2vec model
```
menshikh-iv: misssing space after `#` (here and anywhere)

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,132 @@
...
+        self.dropoutK = dropoutK
+
+        super(Sent2Vec, self).__init__(bucket=bucket, word_ngrams=word_ngrams, min_n=min_n, max_n=max_n,
+            sentences=sentences, size=size, alpha=alpha, window=window, min_count=min_count,
+            max_vocab_size=max_vocab_size, sample=sample, seed=seed, workers=workers, min_alpha=min_alpha,
+            sg=sg, hs=hs, negative=negative, cbow_mean=cbow_mean, hashfxn=hashfxn, iter=iter, null_word=null_word,
+            trim_rule=trim_rule, sorted_vocab=sorted_vocab, batch_words=batch_words)
+
+    def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
+        """Do an initial scan of all words appearing in sentences."""
```
menshikh-iv: Please add more description in docstring, what's a difference between fasttext and sent2vec.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,132 @@
...
+                logger.info(
+                    "PROGRESS: at sentence #%i, processed %i words, keeping %i word types",
+                    sentence_no, sum(itervalues(vocab)) + total_words, len(vocab)
+                )
+            for word in sentence:
+                num_discarded = 0
+                line_size = len(sentence)
+                discard = [False] * line_size
+                while (num_discarded < self.dropoutK and line_size - num_discarded > 2):
+                    token_to_discard = randint(0,line_size-1)
```
menshikh-iv: need space after `,` + spaces around `-`

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,132 @@
...
+                    if discard[token_to_discard] == False:
```
menshikh-iv: `if discard[token_to_discard] == False:` -> `if not discard[token_to_discard]:`

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,132 @@
...
+              word_count=0, queue_factor=2, report_delay=1.0):
+        super(Sent2Vec, self).train(sentences, total_examples=total_examples, epochs=epochs, start_alpha=start_alpha, end_alpha=end_alpha)
+
+    def __getitem__(self, word):
+        return self.word_vec(word)
+
+    def word_vec(self, word, use_norm=False):
+        return FastTextKeyedVectors.word_vec(self.wv, word, use_norm=use_norm)
+
+    def sent_vec(self, sentence):
```
menshikh-iv: This method is bad, don't need to split the raw string to tokens, should pass for example list of tokens.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,132 @@
...
+        self.scale_vocab(keep_raw_vocab=keep_raw_vocab, trim_rule=trim_rule, update=update)
+        self.finalize_vocab(update=update)
+        self.init_ngrams(update=update)
+
+    def train(self, sentences, total_examples=None, total_words=None,
+              epochs=None, start_alpha=None, end_alpha=None,
+              word_count=0, queue_factor=2, report_delay=1.0):
+        super(Sent2Vec, self).train(sentences, total_examples=total_examples, epochs=epochs, start_alpha=start_alpha, end_alpha=end_alpha)
+
+    def __getitem__(self, word):
```
menshikh-iv: what's a difference between `__getitem__` and `word_vec` ?

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,307 @@
...
+            for word in sentence:
+                self.add(word)
+                if self.ntokens % 1000000 == 0:
+                    print "Read " + str(self.ntokens / 1000000) + "M words"
+                if self.size > 0.75 * self.max_vocab_size:
+                    minThreshold += 1
+                    self.threshold(minThreshold)
+
+        self.threshold(min_count)
+        self.initTableDiscard()
```
menshikh-iv: same as for variables (no camel case), here and everywhere.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,307 @@
...
+                continue
+            ntokens += 1
+            words.append(wid)
+            hashes.append(self.hash_(word))
+            if ntokens > self.max_line_size:
+                break
+        return ntokens, hashes, words
+
+
+class Model():
```
menshikh-iv: Are you really need this class (why this isn't a part of `Sent2Vec`?)

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,307 @@
...
+from gensim import matutils
+from random import randint
+import sys
+import random
+from gensim.utils import SaveLoad, tokenize
+import time
+
+logger = logging.getLogger(__name__)
+# TODO: add logger statements instead of print statements
+# TODO: add docstrings and tests
```
menshikh-iv: It's time to make resolve this TODO's, start from `logger` and tests.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,307 @@
...
+                        self.model.update(input_=context, target=words[i], lr=lr)
+                if local_token_count > self.lr_update_rate:
+                    self.token_count += local_token_count
+                    local_token_count = 0
+                if self.token_count >= self.epochs * ntokens:
+                    break
+            print "Progress: ", progress * 100, "% lr: ", lr, " loss: ", self.model.loss / self.model.nexamples
+        print "\n\nTotal training time: %s seconds" % (time.time() - start_time)
+
+    def sentence_vectors(self, sentence_string):
```
menshikh-iv: Useless method (no need to make tokenization in model)

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,307 @@
...
+        sentence = tokenize(sentence_string)
+        ntokens_temp, hashes, words = self.dict.get_line(sentence)
+        sent_vec = np.zeros(self.vector_size)
+        line = self.dict.add_ngrams(context=words, n=self.word_ngrams)
+        for word_vec in line:
+            sent_vec += self.model.wi[word_vec]
+        sent_vec *= (1.0 / len(line))
+        return sent_vec
+
+    def similarity(self, sent1, sent2):
```
menshikh-iv: sent1 and sent2 should be **already** list of tokens (no need to tokenize it).

gensim/models/sent2vec.py
prerna135: Done. Kindly verify in the current commit.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,307 @@
...
+import random
+from gensim.utils import SaveLoad, tokenize
+import time
+
+logger = logging.getLogger(__name__)
+# TODO: add logger statements instead of print statements
+# TODO: add docstrings and tests
+
+
+class Entry():
```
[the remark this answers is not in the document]
prerna135: I can't use tuple because it is immutable.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,307 @@
...
+    def __init__(self, word=None, count=0, subwords=[]):
+        self.word = word
+        self.count = 0
+        self.subwords = subwords
+
+
+class Dictionary():
```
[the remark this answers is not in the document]
prerna135: Done. Kindly verify in the current commit.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,307 @@
...
+        if self.word2int[h] == -1:
+            e = Entry(word=word, count=1)
+            self.words.append(e)
+            self.word2int[h] = self.size
+            self.size += 1
+        else:
+            self.words[self.word2int[h]].count += 1
+
+    def read(self, sentences, min_count):
+        minThreshold = 1
```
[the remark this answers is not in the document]
prerna135: Done. Kindly verify in the current commit.

gensim/models/sent2vec.py
prerna135: Done. Kindly verify in the current commit.

gensim/models/sent2vec.py
prerna135: Done. Merged with sent2vec now.

gensim/models/sent2vec.py
prerna135: Done. Kindly verify in the current commit. Now sentence is passed as a list of unicode strings.

gensim/models/sent2vec.py
prerna135: Done. Kindly verify in the current commit.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,453 @@
...
+        for i in input_:
+            self.wi[i] += self.grad
+
+    def train(self, sentences):
+        """
+        Update the model's neural weights from a sequence of sentences.
+        For Sent2Vec, each sentence must be a list of unicode strings.
+        """
+
+        logger.info("Creating dictionary...")
```
menshikh-iv: Change this method according to w2v (train in __init__ if `sentences` is provided and so on)

gensim/models/sent2vec.py
```diff
@@ -75,6 +75,25 @@ class ModelDictionary():
     """
 
     def __init__(self, t, bucket, minn, maxn, max_vocab_size=30000000, max_line_size=1024):
+        """
+        Initialize a sent2vec dictionary.
```
menshikh-iv: Numpy-style please: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt and http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

prerna135: Sorry, my bad! I used the word2vec code as a reference. I've updated the docstrings. Kindly verify in the latest commit.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,453 @@
...
+        self.subwords = subwords
+
+
+class ModelDictionary():
+    """
+    Class for maintaining Sent2Vec's vocbulary. Provides functionality for storing and training
+    word and character ngrams.
+    """
+
+    def __init__(self, t, bucket, minn, maxn, max_vocab_size=30000000, max_line_size=1024):
```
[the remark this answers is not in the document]
prerna135: Done. Kindly verify in the current commit.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,453 @@
...
+        `bucket` = Number of hash buckets for vocabulary. Default is 2000000.
+
+        `minn` = min length of char ngrams. Default is 3.
+
+        `maxn` = max length of char ngrams. Default is 6.
+
+        `dropoutk` = Number of ngrams dropped when training a sent2vec model. Default is 2.
+        """
+
+        random.seed(seed)
```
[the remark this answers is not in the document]
prerna135: Done. Kindly verify in the current commit.

gensim/models/sent2vec.py
prerna135: Done. Kindly verify in the current commit.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,453 @@
...
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+
+class Entry():
+    """
+    Class for populating Sent2Vec's dictionary.
+    """
+
+    def __init__(self, word=None, count=0, subwords=[]):
+        """
```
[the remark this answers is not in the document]
prerna135: Done. Kindly verify in the current commit.

menshikh-iv:
@prerna135  Oh, very boldly, but it's not worth it c99d18f - you need to optimize partially (hottest and slowest part). It makes no sense to completely rewrite the class (all speedup will be from slowest parts).

Also, I waiting for optimization report (**important**).

@piskvorky can you give advice to us? What gives the best performance gain? Maybe is it time to do a multithreaded version?

piskvorky:
I lack context: what is slow now? How does this implementation compare to other implementations (especially the original), when it comes to performance or quality?

menshikh-iv:
@piskvorky we are interested in performance in "execution time" meaning, quality of implementation is very similar to original implementation as I remember.

@prerna135 please publish profile results + update notebooks with the quality benchmark.

prerna135:
@menshikh-iv 

PFA uncythonized sent2vec code and training script. The output of the python profiler can be found [here](https://drive.google.com/open?id=1Y95Uv3qZZoR-Opj3_BwFrtxYwXoTxPGV)

gensim/models/sent2vec_inner.pyx
```diff
@@ -0,0 +1,117 @@
...
+    cdef REAL_t *wo = <REAL_t *>np.PyArray_DATA(model.wo)
+    cdef int negatives_len = <int>len(model.negatives)
+    cdef int context_len = <int>len(context_)
+
+    #hidden_ = zeros(vector_size, dtype=REAL)
+    #grad_ = zeros(vector_size, dtype=REAL)
+    #cdef float *hidden = <float *>np.PyArray_DATA(hidden_)
+    #cdef float *grad = <float *>np.PyArray_DATA(grad_)
+    cdef REAL_t *hidden, *grad
+    hidden = <REAL_t *>calloc(vector_size, cython.sizeof(REAL_t))
```
piskvorky: Dynamic allocations are very expensive (because of book keeping, but also because they ruin caches). Can't we pre-allocate the structures somehow?
(here and elsewhere)

prerna135:
@menshikh-iv 
Evaluation results after cythonization. Speed up achieved with cython + sigmoid table = 7x

| S.No. | Model                          | Training Time (in seconds) | Pearson/Spearman/MSE | SUBJ | MR   | TREC |
|-------|--------------------------------|----------------------------|----------------------|------|------|------|
| 1.    | Original Sent2Vec (20 threads) | 25.75                      | 0.42/0.43/0.82       | 0.78 | 0.59 | 0.59 |
| 2.    | Gensim Uncythonized Sent2Vec   | 319.09                     | 0.48/0.49/0.78       | 0.76 | 0.58 | 0.63 |
| 3.    | Gensim Cythonized Sent2Vec     | 45.54                      | 0.43/0.42/0.82       | 0.71 | 0.57 | 0.60 |

menshikh-iv:
@prerna135 looks good :+1: what's about latest bugs, are you already fix it?

prerna135:
@menshikh-iv What latest bugs? I fixed whatever bugs I was encountering earlier (NaN values, segmentation faults, overflow etc). Now I'm starting with the implementation of the multithreaded version.

piskvorky:
Just a little note: I see you're using dynamic allocations (`zeros(vector_size, dtype=REAL)`). From what I remember, using pre-allocated arrays resulted in a big speed up (dynamic allocations are costly, plus they thrash CPU caches).

Probably not critical, it seems already fast. Just FYI if you wanted to squeeze more speed :)

prerna135:
@piskvorky I removed the dynamic allocation and the training time reduced from 45.54 s to 39.94 s!

gensim/models/sent2vec.py
```diff
@@ -419,6 +419,8 @@ def __init__(self, sentences=None, vector_size=100, lr=0.2, lr_update_rate=100,
         self.minn = minn
         self.maxn = maxn
         self.dropoutk = dropoutk
+        self.hidden = np.zeros(vector_size, dtype=np.float32)
```
piskvorky: This has to match `REAL_t` (if they ever diverge in the future, you'll get segfaults).

Also be careful around multithreading. Each thread will need its own "scratch space", so they don't overwrite the same array at the same time.

piskvorky:
Oh, only 10%? I thought it would be more. But still worthwhile IMO (the change is simple).

menshikh-iv:
@piskvorky probably, change will be more than 10% on the large corpus (right now, all tests on toy (small) corpus).

gensim/models/sent2vec.py
prerna135: I understand that the threads will need different hidden and grad arrays, but what about updating the weights? Won't all threads have access to the same matrices of neural weights? So it's possible that they may be overwriting the same weights at the same time. Is there a workaround for this?

piskvorky: Great point! You are correct, but that's called a feature :)

It's a technique called HogWild and it's used in algorithms too. Word2vec and doc2vec also overwrite the same parameter matrices.

The idea is that the hardware provides enough atomicity (updating a single float / a SIMD operation / cache line) to make it not crash / go horribly wrong. And then the threads stepping on each other's toes cancel out, it turns out it's not a big deal (and surprisingly, sometimes even helps!).

prerna135:
@piskvorky @menshikh-iv I'm getting some weird results. When I use a single thread the training time is 37.6s, but when I use 3 threads the training time increases to 71.2s. It can't possibly be because of the GIL since I'm using nogil everywhere in the ```sent2vec_inner.pyx``` file. What could be the reason for this?

manneshiva:
Your [`update`](https://github.com/prerna135/gensim/blob/sent2vec/gensim/models/sent2vec_inner.pyx#L96) in `sent2vec_inner.pyx` is a python function and contains a `nogil` portion inside it . Any call to this function can't be completely `nogil` (since it isn't a `nogil` function by itself). You call this function an enormous number of times (once for each word in every sentence) causing each worker's call to `update` to wait for other workers to get into `nogil` portions of their code. This is evident/confirmed from the time profiling results your code -- maximum time taken is by the `update` function, also this time increases with increasing number of workers.

The solution is to follow a design similar to `FastText`/`Word2Vec`. Create a function in `sent2vec_inner.pyx` which accepts batches of `sentences` and process each batch in `nogil`. For better clarity, refer to this [portion](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/fasttext_inner.pyx#L478) in `fasttext_inner.pyx`. This will also need you to change your `update` function to a `nogil` `C` function.

Also, as Ivan mentioned [here](https://github.com/RaRe-Technologies/gensim/pull/1619#pullrequestreview-71758691), please try and follow the `FastText` design so that it will be easier to refactor your code in future (already in pipeline).

prerna135:
@menshikh-iv 

Latest results! :)

| S.No. | Model                          | Training time (in seconds) | Pearson/Spearman/MSE | SUBJ | MR   | TREC |
|-------|--------------------------------|----------------------------|----------------------|------|------|------|
| 1.    | Original Sent2Vec (20 threads) | 25.75                      | 0.42/0.43/0.82       | 0.78 | 0.59 | 0.59 |
| 2.    | Gensim Sent2Vec (20 threads)   | 10.0                       | 0.44/0.44/0.81       | 0.71 | 0.57 | 0.59 |

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,827 @@
...
+        self.hidden *= (1.0 / len(input_))
+        loss = self.negative_sampling(target, lr)
+        self.grad *= (1.0 / len(input_))
+        for i in input_:
+            self.wi[i] += self.grad
+        return loss
+
+    def build_vocab(self, sentences):
+        logger.info("Creating dictionary...")
+        self.dict = ModelDictionary(t=self.t, bucket=self.bucket, maxn=self.maxn, minn=self.minn)
```
manneshiva: `max_vocab_size` never set. Pass `self.max_vocab_size` to `ModelDictionary` here.

prerna135: I'm using default `max_vocab_size` as specified in the `__init__()` function of `ModelDictionary`

manneshiva:
@menshikh-iv @prerna135 
I went through the paper, the original code and the code here. Here are my comments:
1. The training part of your code looks all OK to me. Great job!
2. During Vocabulary building, I see that you are also computing and storing character `ngrams` (`subwords`) which according to my understanding from the paper, are not required to be used -- please discard this extra/unnecessary computation. The code released by the authors also seems to neglect this part -- they have built upon Facebook's `FastText` library that requires these subwords information). 
3. Your code closely resembles (method names and design) the code released by the original authors ([repo link](https://github.com/epfml/sent2vec)) and does not re-use existing Gensim code (for example, the vocabulary building/pruning parts) -- @menshikh-iv  Is this fine? Might become difficult to refactor later and adds to the existing code redundancy.

gensim/models/sent2vec.py
manneshiva: Shouldn't this be a parameter that the user can set (look at `Word2Vec`/`FastText`)? I can see that you have mentioned  this parameter's description in `Sent2Vec.__init__()` docstring but is missing from the method.

menshikh-iv:
Big thanks for review @manneshiva !
about (3), I agree with you, we need to reuse some code here (to simplify future refactoring and avoid duplication in codebase)

gensim/models/sent2vec.py
prerna135: Oh yes. My bad. I've fixed it in latest commit.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,832 @@
...
+    def __init__(self, word=None, count=0, subwords=[]):
+        """
+        Initialize a single dictionary entry.
+        Parameters
+        ----------
+        word : str
+            Actual vocabulary word.
+        count : int
+            Number of times the word occurs in the vocabulary.
+        subwords : list
```
menshikh-iv: list of ... ? (here and everywhere)

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,832 @@
...
+    def __init__(self, t, bucket, minn, maxn, max_vocab_size=30000000, max_line_size=1024):
+        """
+        Initialize a sent2vec dictionary.
+        Parameters
+        ----------
+        t : float
+            Threshold for configuring which higher-frequency words are randomly downsampled;
+            default is 1e-3, useful range is (0, 1e-5).
+        bucket : int
+            Number of hash buckets for vocabulary. Default is 2000000.
```
menshikh-iv: No need to duplicate default value in dostring (here and everywhere)

Sent2Vec Tests and Tutorial.ipynb
```diff
@@ -0,0 +1,3008 @@
+{
+ "cells": [
```
menshikh-iv: Notebook should be in `docs/notebooks` (not root of repo) + please rename it to `sent2vec_tutorial.ipynb`

Sent2Vec Tests and Tutorial.ipynb
```diff
@@ -0,0 +1,3008 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "No handlers could be found for logger \"gensim.models.fasttext\"\n",
```
menshikh-iv: Look suspicious, please check it

Sent2Vec Tests and Tutorial.ipynb
```diff
@@ -0,0 +1,3008 @@
...
+    "from gensim.models.fasttext import FastText as ft\n",
+    "from gensim.models.sent2vec import TorontoCorpus\n",
+    "from gensim.utils import tokenize\n",
+    "import scipy\n",
+    "import re\n",
+    "from numpy import dot\n",
+    "from gensim import matutils\n",
+    "import time\n",
+    "import numpy as np\n",
+    "import tensorflow as tf\n",
```
menshikh-iv: Please cleanup import section (remove useless imports)

Sent2Vec Tests and Tutorial.ipynb
```diff
@@ -0,0 +1,3008 @@
...
+      "/Users/prerna135/anaconda/envs/gensim_env/lib/python2.7/site-packages/sklearn/cross_validation.py:44: DeprecationWarning: This module was deprecated in version 0.18 in favor of the model_selection module into which all the refactored classes and functions are moved. Also note that the interface of the new CV iterators are different from that of this module. This module will be removed in 0.20.\n",
+      "  \"This module will be removed in 0.20.\", DeprecationWarning)\n"
+     ]
+    }
+   ],
+   "source": [
+    "import gensim\n",
+    "import os\n",
+    "from gensim.models.word2vec import LineSentence\n",
+    "from gensim.models.sent2vec import Sent2Vec as s2v\n",
```
menshikh-iv: Please don't rename classes when you import it (use original names, not short-cuts)

Sent2Vec Tests and Tutorial.ipynb
```diff
@@ -0,0 +1,3008 @@
...
+    "from gensim.utils import tokenize\n",
+    "import scipy\n",
+    "import re\n",
+    "from numpy import dot\n",
+    "from gensim import matutils\n",
+    "import time\n",
+    "import numpy as np\n",
+    "import tensorflow as tf\n",
+    "import random\n",
+    "import eval_sick\n",
```
menshikh-iv: Before import this, please make notebook "runnable" (i.e. download needed files through concrete commands in cells).

Sent2Vec Tests and Tutorial.ipynb
```diff
@@ -0,0 +1,3008 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# Prepare training data\n",
```
menshikh-iv: Use markdown cell as headers before code (not `#` on head of cells) for big pieces of code

Sent2Vec Tests and Tutorial.ipynb
menshikh-iv: Also, please re-run it with python3

[WIP] Address Detection Evaluation.ipynb
```diff
@@ -0,0 +1,36 @@
+{
```
Comment (menshikh-iv) [inline [WIP] Address Detection Evaluation.ipynb]: Please remove this file from PR

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,847 @@
...
+        for i in range(len(counts)):
+            z += counts[i] ** 0.5
+        for i in range(len(counts)):
+            c = counts[i] ** 0.5
+            for j in range(int(c * self.negative_table_size / z) + 1):
+                self.negatives.append(i)
+        self.random.shuffle(self.negatives)
+        self.negatives = np.array(self.negatives)
+
+    def get_negative(self, target):
```
menshikh-iv: please hide all non-public methods with `_method_name`

gensim/models/sent2vec.py
```diff
@@ -339,7 +338,7 @@ def get_line(self, sentence):
 class Sent2Vec(SaveLoad):
     """Class for training and using neural networks described in [1]_"""
 
-    def __init__(self, sentences=None, size=100, lr=0.2, lr_update_rate=100, epochs=5, min_count=5, neg=10,
+    def __init__(self, sentences=None, size=100, lr=0.2, lr__update_rate=100, epochs=5, min_count=5, neg=10,
```
menshikh-iv: No needed rename, return old version please

menshikh-iv:
Don't forgot about tests @prerna135

menshikh-iv:
@prerna135 problem exactly with `gensim.test.test_sent2vec.TestSent2VecModel)` (I comment it and add move verbosity)

1. Remove all x32 skip for AppVeyour and reduce the model size (let's look how this will work on windows) 
2. Try to investigate what happens with `test-persistence`

gensim/test/test_sent2vec.py
```diff
@@ -0,0 +1,85 @@
...
+        model = Sent2Vec.load(tmpf)
+        self.assertTrue(model.dict.size, 12)
+        model.build_vocab(new_sentences, update=True)  # update vocab
+        model.train(new_sentences)
+        self.assertEqual(model.dict.size, 14)
+
+    def test_sent2vec_for_document(self):
+        model = Sent2Vec(sentences, size=5, min_count=0, seed=42, neg=5, workers=1)
+        for sentence in test_sentences:
+            sent_vec = model.sentence_vectors(sentence)
```
menshikh-iv: It's better to use `__getitem__` instead of `sentence_vector` (gensim way), i.e.
`model[["hello", "my", "friend"]]`

gensim/test/test_sent2vec.py
```diff
@@ -0,0 +1,85 @@
...
+            self.assertEqual(len(sent_vec), 5)
```
menshikh-iv: You should also check that vector have "sence" (more deep sanity check)

gensim/test/test_sent2vec.py
```diff
@@ -0,0 +1,85 @@
...
+        model = Sent2Vec(sentences, size=5, min_count=0, seed=42, neg=5, workers=1)
+        model.save(tmpf)
+        model = Sent2Vec.load(tmpf)
+        self.assertTrue(model.dict.size, 12)
+        model.build_vocab(new_sentences, update=True)  # update vocab
+        model.train(new_sentences)
+        self.assertEqual(model.dict.size, 14)
+
+    def test_sent2vec_for_document(self):
+        model = Sent2Vec(sentences, size=5, min_count=0, seed=42, neg=5, workers=1)
```
menshikh-iv: Change `workers=2` only in this test (I hope that we have no more problems with Travis)

gensim/test/test_sent2vec.py
prerna135: "Sense" as in I should compute sentence vectors of sentences having similar meanings and then check that the sentence vectors have greater cosine similarity value?

menshikh-iv: @prerna135 yep, and also be sure that if you apply model twice to one document - you received same vectors.

menshikh-iv:
@prerna135 I'm still waiting:
- SEGFAULT fix
- return python code for avoiding the problem with `FAST_VERSION` (that we already discussed earlier)
- Refactoring with new API

Also, don't forget to resolve merge-conflict

menshikh-iv:
Also, don't forget to add your files to https://github.com/RaRe-Technologies/gensim/blob/develop/MANIFEST.in (in same manner as w2v or fasttext), this needed for building `sdist`

setup.py
```diff
@@ -262,6 +262,11 @@ def finalize_options(self):
             include_dirs=[model_dir]),
         Extension('gensim._matutils',
             sources=['./gensim/_matutils.c']),
+        Extension('gensim.models.sent2vec_inner',
+            sources=['./gensim/models/sent2vec_inner.cpp'],
+            include_dirs=[model_dir],
+            language="c++",
+            extra_compile_args=['/std:[c++14|c++latest]', '/EHsc'])
```
menshikh-iv: remember that you need different arguments for different OS version here
- `-std=c++11` for non-windows
- `something-else` for windows

You can make something like

`extra_compile_args=['/std:[c++14|c++latest]' if os.name == "nt" else "-std=c++11"]`

for each argument that is OS-special

prerna135:
@menshikh-iv 
Current status:

| S.No. | Model               | Status                                          |
|-------|---------------------|-------------------------------------------------|
| 1.    | Gensim FastText     | Trained and evaluated.                          |
| 2.    | Gensim Doc2Vec DBOW | To be trained. More memory required.            |
| 3.    | Gensim Doc2Vec DM   | To be trained. More memory required.            |
| 4.    | Original Sent2Vec   | Trained and evaluated.                          |
| 5.    | Gensim Sent2vec     | Trained. Not evaluated. NaN values encountered. |

docs/notebooks/sent2vec_tutorial.ipynb
```diff
@@ -0,0 +1,339 @@
+{
```
menshikh-iv: re-run & commit notebook please (to be 100% sure that this works with current code version).

tox.ini
```diff
@@ -34,7 +34,6 @@ setenv =
     SKIP_NETWORK_TESTS={env:SKIP_NETWORK_TESTS:}
 
 commands =
-    python -c "from gensim.models.word2vec import FAST_VERSION; print(FAST_VERSION)"
```
menshikh-iv: return this line please

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,809 @@
...
+The sentence vectors are stored in a numpy array
+
+>>> vector = model[['computer', 'interface']] # vector of a sentence
+
+You can perform the NLP similarity task with the model
+
+>>> similarity = model.similarity(['graph', 'minors', 'trees'], ['eps', 'user', 'interface', 'system'])
+
+
+References
```
menshikh-iv: Please use `some text <http://...>`_ instead of references with `[number]`.

gensim/models/sent2vec_inner.pyx
```diff
@@ -0,0 +1,245 @@
+#!/usr/bin/env cython
```
menshikh-iv: please add a docstrings here too, instruction: https://github.com/RaRe-Technologies/gensim/pull/1944#issuecomment-372909479

docs/src/models/sent2vec.rst
```diff
@@ -0,0 +1,8 @@
+:mod:`models.fasttext` -- Sent2Vec model
```
menshikh-iv: missing rst for `sent2vec_inner` (+same for `apiref`)

gensim/models/sent2vec_inner.pyx
menshikh-iv: still  missed `# cython: embedsignature=True `  + need to re-compile after

gensim/models/sent2vec_inner.pyx
```diff
@@ -0,0 +1,443 @@
...
+                              REAL_t *hidden, const int vector_size, const int neg, int *negatives,
+                              int *negpos, const int negatives_len)nogil:
+    """
+    Get loss using negative sampling.
+    Pararmeters
+    -----------
+    target : const int
+        Word id of target word.
+    lr : REAL_t
+        current learning rate.
```
menshikh-iv: Shoud start with Capital and end with `.` always (here and everywhere).

gensim/models/sent2vec_inner.pyx
```diff
@@ -0,0 +1,443 @@
...
+    if val >= 1.0:
+        return 0.0
+    else:
+        return LOG_TABLE[<int>(val * EXP_TABLE_SIZE)]
+
+
+cdef REAL_t binary_logistic(const int target, const int label, const REAL_t lr,
+                            const int vector_size, REAL_t *wo, REAL_t *grad,
+                            REAL_t *hidden)nogil:
+
```
menshikh-iv: remove empty string

gensim/models/sent2vec.py
menshikh-iv: still not resolved

gensim/models/sent2vec_inner.pyx
prerna135: What empty string?

gensim/models/sent2vec.py
menshikh-iv:
```
`Unsupervised Learning of Sentence Embeddings using Compositional n-Gram Features <https://arxiv.org/abs/1703.02507>`_
```

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+from random import randint
+from gensim.models.base_any2vec import BaseWordEmbeddingsModel
+from scipy.special import expit
+import os
+
+try:
+    from gensim.models.sent2vec_inner import _do_train_job_fast, FAST_VERSION
+except ImportError:
+    # failed... fall back to plain numpy (20-80x slower training than the above)
+    FAST_VERSION = -1
```
menshikh-iv: move `_do_train_job_slow` here (make it function)

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+
+        """
+        if update:
+            self.negatives = list(self.negatives)
+        z = 0.0
+        for i in range(len(counts)):
+            z += counts[i] ** 0.5
+        for i in range(len(counts)):
+            c = counts[i] ** 0.5
+            for j in range(int(c * self.negative_table_size / z) + 1):
```
menshikh-iv: - replace with `min(3 * len(counts), 10000000)` here
- remove `self.negative_table_size` attribute

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+        alpha : float, optional
+            Initial learning rate.
+        epochs : int, optional
+            Number of iterations (epochs) over the corpus.
+        min_count : int, optional
+            Ignore all words with total frequency lower than this.
+        negative : int, optional
+            Specifies how many "noise words" should be drawn (usually between 5-20).
+        word_ngrams : int, optional
+            Max length of word ngram.
```
menshikh-iv: length in words

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+            Target size (in words) for batches of examples passed to worker threads (and thus cython routines).
+            Larger batches will be passed if individual texts are longer than 10000 words, but the standard cython code
+            truncates to that maximum.
+        workers : int, optional
+            Use this many worker threads to train the model (=faster training with multicore machines).
+        max_vocab_size : int, optional
+            Limit RAM during vocabulary building,
+            if there are more unique words than this, then prune the infrequent ones.
+        compute_loss: bool
+            If True, computes and stores loss value which can be retrieved using `model.get_latest_training_loss()`.
```
menshikh-iv:
```
:meth:`~gensim.models.sent2vec.Sent2Vec.get_latest_training_loss`
```

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+        Returns
+        -------
+        numpy.ndarray
+            Sentence vector for input sentence.
+
+        """
+
+        ntokens_temp, words = self.vocabulary.get_line(sentence)
+        sent_vec = np.zeros(self.vector_size)
+        line = self.vocabulary.add_ngrams(context=words, n=self.word_ngrams)
```
menshikh-iv: `*_ngrams` -> `*word_ngrams` (everywhere)

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+        for word_vec in line:
```
menshikh-iv: use `np.mean` instead

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+            sent_vec += self.wi[word_vec]
+        if len(line) > 0:
+            sent_vec *= (1.0 / len(line))
+        return sent_vec
+
+    @classmethod
+    def load(cls, *args, **kwargs):
```
menshikh-iv: remove this method

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+            logger.info("Creating dictionary...")
+            self.corpus_count = self.vocabulary.read(sentences=sentences, min_count=self.min_count)
+            logger.info("Dictionary created, dictionary size: %i, tokens read: %i",
+                        self.vocabulary.size, self.vocabulary.ntokens)
+            counts = [entry.count for entry in self.vocabulary.words]
+            self.wi = self.random.uniform((-1 / self.vector_size), ((-1 / self.vector_size) + 1),
+                                          (self.vocabulary.size + self.bucket, self.vector_size)
+                                          ).astype(np.float32)
+            self.wo = np.zeros((self.vocabulary.size, self.vector_size), dtype=np.float32)
+            self._init_table_negatives(counts=counts, update=update)
```
menshikh-iv: `return`

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+        sentences : iterable of iterable of str
+            Input sentences.
+        update : boolean
+            Update existing vocabulary using input sentences if True
+        """
+        if not update:
+            logger.info("Creating dictionary...")
+            self.corpus_count = self.vocabulary.read(sentences=sentences, min_count=self.min_count)
+            logger.info("Dictionary created, dictionary size: %i, tokens read: %i",
+                        self.vocabulary.size, self.vocabulary.ntokens)
```
menshikh-iv:
```python

logger.info(
    "Dictionary created, dictionary size: %i, tokens read: %i", 
    self.vocabulary.size, self.vocabulary.ntokens
)
```
here and everywhere

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,793 @@
...
+            h = line[i]
+            for j in range(i + 1, line_size):
+                if j >= i + n or discard[j] is True:
+                    break
+                h = h * 116049371 + line[j]
+                line.append(self.size + (h % self.bucket))
+        return line
+
+    def add_ngrams(self, context, n):
+        """Computing word ngrams for given sentence while inferring sentence vector.
```
menshikh-iv: clarify docstring about n-gram generation

gensim/models/sent2vec.py
prerna135: Can't remove this. I need to call `utils.SaveLoad` load function instead of `BaseWordEmbeddingsModel`.

menshikh-iv:
@prerna135 see my message in https://github.com/RaRe-Technologies/gensim/issues/1376#issuecomment-447249741

prerna135:
@menshikh-iv It looks like the `fasttext`, `word2vec` and `doc2vec` tests are failing due to absence of certain modules like the `word2vec_corpusfile`. Any idea why this happens?

gensim/models/sent2vec_inner.pyx
```diff
@@ -0,0 +1,445 @@
...
+            num_discarded += 1
+
+    for i from 0 <= i < line_size:
+        if discard[i] == 1:
+            continue
+        h = line[i]
+        for j from i + 1 <= j < line_size:
+            if j >= i + n or discard[j] == 1:
+                break
+            h = h * 116049371 + line[j]
```
horpto: it's quite not obvious what a magic number is 116049371

gensim/models/sent2vec_inner.pyx
```diff
@@ -0,0 +1,445 @@
...
+        if wids[i] < 0:
+            continue
+        ntokens += 1
+        words.push_back(wids[i])
+        if ntokens >= max_line_size:
+            break
+    return ntokens
+
+
+cdef void add_ngrams_train(vector[int] &line, int n, int k, int bucket, int size)nogil:
```
horpto: shouldn't empty space stand before nogil ?

gensim/models/sent2vec_inner.pyx
menshikh-iv: this is mimic to FB implementation

gensim/models/sent2vec_inner.pyx
menshikh-iv: Not necessary as I remember

horpto: yeah, unnecessary, but r e a d a b i l i t y

menshikh-iv: I'll fix that myself of course

menshikh-iv:
blocked  by #2313 (should be merged before we can continue with current PR)

gensim/models/sent2vec_inner.pyx
piskvorky: A good idea to include that info in a comment -- so someone doesn't accidentally change the magic numbers in the future.

gensim/models/sent2vec.py
```diff
@@ -0,0 +1,788 @@
...
+        for fname in os.listdir(self.dirname):
+            fname = os.path.join(self.dirname, fname)
+            if not os.path.isfile(fname) or ".txt" not in fname:
+                continue
+            sentence = None
+            for line in utils.smart_open(fname):
+                if line not in ['\n', '\r\n']:
+                    sentence = list(tokenize(line))
+                if sentence:
+                    yield sentence
```
horpto: if line is  a new line chars then previous sentence will be yielded twice.

menshikh-iv: hm, I guess that's a bug

prerna135:
> @prerna135 sent2vec built successfully (some issues with FT and python2, but unrelated)
> 
> 1. Bad performance (you can check it yourself, can be related with correct hash function, but I'm not sure)
> 2. Looks like model doesn't really learn, for example
> 
> ```python
> import logging
> from gensim.models import Sent2Vec
> import gensim.downloader as api
> from gensim.utils import simple_preprocess
> import numpy as np
> from scipy.spatial.distance import cdist
> 
> logging.basicConfig(level=logging.INFO)
> 
> corpus = [simple_preprocess(_["data"]) for _ in api.load("20-newsgroups")]
> model = Sent2Vec(corpus)
> 
> c_vectors = np.array([model[d] for d in corpus])
> fst_vector = c_vectors[0]
> 
> similarities = (1 - cdist(new_vector.reshape((1, new_vector.shape[0])), c_vectors, metric='cosine')).reshape(-1)
> print(similarities, similarities.mean())
> 
> # (array([0.99890759, 0.99885709, 0.99865863, ..., 0.99843101, 0.99866404,
> #       0.99762219]), 0998304)
> ```
> 
> I.e. all vectors from corpus super-near, that's very suspicious IMO, even if I try to `model.similarity(random_words, different_random_words)` - it's also too high, any ideas, what's wrong here?
> 
> 1. Original sent2vec released several models that we can't load (important feature I think, see https://github.com/epfml/sent2vec#downloading-pre-trained-models)
> 
> Unfortunatelly, I can't merge PR in current state :( Not ready for `3.7.0`.
> @prerna135 when you'll have a time to resolve mentioned problems?

Hi @menshikh-iv. The bad performance part is surprising since it outperformed `doc2vec` in all the evaluation tasks as you can see [here](https://rare-technologies.com/sent2vec-an-unsupervised-approach-towards-learning-sentence-embeddings/). I'll try to check could be causing the problem while calculating sentence similarities. Can't promise quick results though, as my semester begins next week.

menshikh-iv:
@prerna135 I guess I know what’s a reason of bad performance (problem in hash function), @mpenkov will fix it soon and I’ll update your PR and ping you, ok ?

menshikh-iv:
Possible problem (hash func): 
https://github.com/RaRe-Technologies/gensim/issues/1261#issuecomment-455487894
Will be fixed in #2340 (block current PR)

menshikh-iv:
@prerna135 I fixed a performance issue (this was really hashing issue) & build itself.
Unfortunatelly, quality of result still low (2 from https://github.com/RaRe-Technologies/gensim/pull/1619#issuecomment-454306757), feel free to investigate (btw, check also my changes that I don't break any myself).

So, TODO for you
- check my changes in your PR, make sure that I don't break any
- investigate & fix, what happens with distance
-  **make sure that you calculate ngrams/hashes in exactly same way** (bytes or string ngram?), this is important one, because fasttext use bytes for it (not strings) -> we hit with incompatibility issue with FB models, see #2059 and #2313 + #2340 
- (optional) Implement `load_sent2vec_format` for support pre-trained from https://github.com/epfml/sent2vec#downloading-pre-trained-models, similar to https://radimrehurek.com/gensim/models/fasttext.html#gensim.models.fasttext.FastText.load_fasttext_format


--

menshikh-iv:
ping @prerna135, any updates?

mpenkov:
@prerna135 are you able to finish this PR?

prerna135:
@menshikh-iv @mpenkov I have been very busy with grad school in the past year. Apologies for the delay. I'll try to look into this over the summer.

piskvorky:
@prerna135 ping. This project has been under way for nearly 2 years.

piskvorky:
@prerna135 what's the status? Summer is nearly over.

@mpenkov unless Prerna finishes the PR, we'll have to kill it + her [blog post](https://rare-technologies.com/sent2vec-an-unsupervised-approach-towards-learning-sentence-embeddings/). It's getting ridiculous.

prerna135:
@menshikh-iv @piskvorky I tried to look into the code over the summer (hash function and distance computation issue). I'd have to go over the entire original c++ code to detect the bug, refractor code according to avoid deprecation warnings, retrain models and run benchmarking experiments again. I'm afraid I won't be able to devote the time required to do this. Apologies for dragging this out. I tried to run the existing code to replicate blog post results, but too many things were breaking for me to figure out the issue quickly.

PR #1620, opened 2017-10-10
Enable float32 for LSI - stochastic SVD

isamaru:

Enables using float32 in LSI calculation by exposing dtype as parameter. Default remains float64 and the behavior shouldn't change at all unless different dtype is used.


menshikh-iv:
Thanks @isamaru, very useful PR :fire:

PR #1621, opened 2017-10-10
Add __getitem__ method to Sparse2Corpus to allow direct queries

isamaru:

Adds `__getitem__` method to Sparse2Corpus to enable direct doc querying.
Was necessary so that the CSC corpus could be used for similarity computation, and looks overall useful.

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -342,6 +342,11 @@ class Sparse2Corpus(object):
     def __len__(self):
         return self.sparse.shape[1]
 
+    def __getitem__(self, item):
+        indprev = self.sparse.indptr[item]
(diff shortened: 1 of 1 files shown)


gensim/matutils.py
```diff
@@ -342,6 +342,11 @@ def __iter__(self):
     def __len__(self):
         return self.sparse.shape[1]
 
+    def __getitem__(self, item):
```
piskvorky: What is `item`? Row index, column index, something else?

Missing docstring (method semantics, args, returns).

menshikh-iv:
Thank you @isamaru, congratz with first PR :+1:

PR #1622, opened 2017-10-10
Remove unnecessary assert blocking direct usage of CSC for LSI

isamaru:

Removes the assert which is blocking usage of CSC directly as corpus for LsiModel with multiple power iterations.
I could not figure out why the assert was there in the first place: it doesn't make sense to me: The code path handles CSC exclusively but I don't see why CSC would require a single pass only. It works just fine with multiple power iterations.

### DIFF
--- a/gensim/models/lsimodel.py
+++ b/gensim/models/lsimodel.py
@@ -407,7 +407,6 @@ class LsiModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
                 self.docs_processed += doc_no
         else:
             assert not self.dispatcher, "must be in serial mode to receive jobs"
-            assert self.onepass, "distributed two-pass algo not supported yet"
             update = Projection(
                 self.num_terms, self.num_topics, corpus.tocsc(), extra_dims=self.extra_samples,
                 power_iters=self.power_iters


piskvorky:
@isamaru IIRC, the code path that handles CSC was there for the distributed version only. The master worker prepares batches for workers as CSC matrices; the workers process the batches and send the results back.

It was never meant to be used in this way, and I wouldn't expect it to actually work (do all the necessary model setup etc). But if it does, that's great.

isamaru:
@radim Thanks for the explanation! Makes much more sense now.

Distributed algorithm with not work on CSC directly as written now: the dispatcher code is only enabled in the top branch (`if not scipy.sparse.issparse(corpus):`), and the remaining assert will fail during dispatcher initialization if you try distributed straight on CSC.
Distributed algorithm on corpus (the original use case) will work just fine, it remains unaffected by this change.
Enabling the distributed mode for CSC would take slightly more work and refactoring of the whole method.

To sum up: serial (non-distributed) algorithm will work directly on CSC. Distributed will not work, failing on the very same assert than before ('must be in serial mode to receive jobs').

I could rewrite that assert message to be more meaningful to the user: "Distributed mode not supported for CSC matrices directly. Use csc2corpus."

piskvorky:
I see, thanks. I think that's fine as-is, no changes needed -- nobody really uses the (ancient) distributed mode, AFAIK. It's one of the candidate features for removal.

menshikh-iv:
@piskvorky ok, I close this PR

piskvorky:
Why? We definitely needs this PR.

menshikh-iv:
@piskvorky  as I understand, ` think that's fine as-is, no changes needed -- nobody really uses the (ancient) distributed mode` means that we no need this changes

piskvorky:
No, it means we don't need to rewrite the assert message.

menshikh-iv:
Thanks @isamaru :1st_place_medal:

PR #1633, opened 2017-10-17
Fix duplication and wrong markup in docs

horpto:
- duplication of the citates from word2vec and doc2vec,
- wrong markup of lists in the scripts,
- some typos.


menshikh-iv:
Thanks @horpto, congratz with first contribution :1st_place_medal:

Issue #10, opened 2011-03-04
comply to PEP8


piskvorky:
Change `camelCase` vars and function names and parameters to `python_style`. This will break backward compatibility, so it best be done all at once (rather than incrementally).

http://groups.google.com/group/gensim/browse_thread/thread/73f4301723cea765

mcobzarenco:
This is not the case anymore, (e.g. pyflake shows 63 warnings for `models/wrappers/fasttext.py`). Happy to PEP8 the whole repo and create a pr if that's cool.

piskvorky:
@mcobzarenco what warnings are these? For example, we care more about readability than a strict 80-char-line limit.

But I also noticed some lapses (vertical indent in word2vec). @menshikh-iv what's the status on code style consistency in gensim?

menshikh-iv:
@piskvorky 
```
(math) ivan@P50:~/release/gensim$ git log | head -n5
commit e92b45d3f83ad670da9cf6ae20ae86a2a1c8558c
Author: jodevak <jodevak@users.noreply.github.com>
Date:   Thu Oct 19 09:37:41 2017 +0300

    Add build_vocab_from_freq to Word2Vec, speedup scan_vocab (#1599)
(math) ivan@P50:~/release/gensim$ find gensim/ -name "*.py" | xargs flake8 --ignore=E501,E731,E12,W503,E402 --show-source
gensim/test/test_lsimodel.py:95:5: E303 too many blank lines (2)
    def testCorpusTransform(self):
    ^
```

mcobzarenco:
Right, fair enough, all of the warnings were of the types you ignore above. Do you think we could we add those ignored rules to `setup.cfg` so they're documented and also ignored by default? I created a PR here #1636

PR #1636, opened 2017-10-19
Add pep8 exceptions for flake8 in setup.cfg

mcobzarenco:
#10 suggests that the code style of `gensim` should ignore flake8's `E501,E731,E12,W503,E402`

This PR adds the rules to `setup.cfg` so they're documented and used by default.

### DIFF
--- a/continuous_integration/travis/flake8_diff.sh
+++ b/continuous_integration/travis/flake8_diff.sh
@@ -20,6 +20,7 @@ set -o pipefail
 
 PROJECT=RaRe-Technologies/gensim
 PROJECT_URL=https://github.com/${PROJECT}.git
+FLAKE_CONFIG_FILE=setup.cfg
 
 # Find the remote with the project name (upstream in most cases)
 REMOTE=$(git remote -v | grep ${PROJECT} | cut -f1 | head -1 || echo '')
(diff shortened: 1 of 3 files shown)


menshikh-iv:
@mcobzarenco hi, please update current flake8 run https://github.com/RaRe-Technologies/gensim/blob/develop/continuous_integration/travis/flake8_diff.sh#L143 and https://github.com/RaRe-Technologies/gensim/blob/develop/continuous_integration/travis/flake8_diff.sh#L153

menshikh-iv:
ping @mcobzarenco

mcobzarenco:
Sorry for the delay -- ok, so:
 - I removed "--ignore ..." from the arguments of `flake8` in `flake8_diff.sh` 
 - I checked that `flake8` uses the rules from `setup.cfg` when running `flake8_diff.sh` locally

mcobzarenco:
One thing to keep in mind with the current way of running `flake8` (only run it on the files that were changed by a PR) is that it breaks down if one tries to change the `flake8` rules and/or `flake8` is updated. 

One fix for the former would be for `flake8_diff.sh` to run flake on all files if it itself has been changed.

menshikh-iv:
@mcobzarenco please add `--config` argument with path explicitly for both calls.

mcobzarenco:
`flake8` seems to look for `setup.cfg` / `tox.ini` / `.flake8` in the root of the repo by default (https://travis-ci.org/RaRe-Technologies/gensim/jobs/292001367 seemed to have worked), but cool I also added an explicit `--config` cmd line argument, (I assume `flake_diff.sh` gets executed from the root of the repo by travis).

menshikh-iv:
Thanks, congratz with first contribution @mcobzarenco :+1:

mcobzarenco:
Thanks for merging!

Issue #1624, opened 2017-10-11
Evaluation of existing Pointcaré embedding implementations


janpom:
Implementations
- https://github.com/TatsuyaShirakawa/poincare-embedding
- https://github.com/nishnik/poincare_embeddings

Data sets
- [WordNet](https://wordnet.princeton.edu/wordnet/download/)
  - The version used for training/evaluation is prepared with this [script](https://github.com/TatsuyaShirakawa/poincare-embedding/blob/master/scripts/create_wordnet_noun_hierarchy.py) from the c++ poincare embedding repository 
- [HyperLex](http://people.ds.cam.ac.uk/iv250/hyperlex.html)
- scientific collaboration social networks
  - [AstroPh](https://snap.stanford.edu/data/ca-AstroPh.html)
  - [CondMat](https://snap.stanford.edu/data/ca-CondMat.html)
  - [GrQc](https://snap.stanford.edu/data/ca-GrQc.html)
  - [HepPh](https://snap.stanford.edu/data/ca-HepPh.html)

Evaluation experiments
- WordNet reconstruction
- WordNet link prediction
- link prediction in social networks
- lexical entailment on HyperLex & WordNet (for training)

Deliverables
- Jupyter notebook
  - evaluation code
  - evaluation results
  - summary

Requirements
- runnable notebook with minimal necessary dependencies
- ideally, as part of the setup
  - datasets are downloaded
  - implementations are downloaded and installed
- notebook easy to read, concise (no long printouts)
- clean code (PEP8, DRY)

janpom:
done in #1643

PR #1643, opened 2017-10-23
Evaluation of existing Poincaré embedding implementations

janpom:
#1624


docs/notebooks/Poincare Evaluation.ipynb
```diff
@@ -0,0 +1,1448 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Set this to the path of the directory that is to contain the poincare-embedding directory\n",
+    "parent_directory = '/home/jayant/projects/'"
```
janpom: Please use path relative to the notebook, so that you don't need to hardcode your home dir path.

docs/notebooks/Poincare Evaluation.ipynb
```diff
@@ -0,0 +1,1448 @@
...
+      "./poincare_burn_in_ 100%[===================>]   7.81K  --.-KB/s    in 0s      \n",
+      "\n",
+      "2017-10-18 20:37:11 (59.5 MB/s) - ‘./poincare_burn_in_eps.patch’ saved [7999/7999]\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Download and apply patch\n",
+    "patch_url = \"https://raw.githubusercontent.com/RaRe-Technologies/gensim/poincare_eval/docs/notebooks/poincare_data/poincare_burn_in_eps.patch\"\n",
```
janpom: The patch is in the repo so you can just reference the local file (`poincare_data/poincare_burn_in_eps.patch`). Besides, the URL will get broken once the code is merged to master.

docs/notebooks/Poincare Evaluation.ipynb
```diff
@@ -0,0 +1,1448 @@
...
+   "execution_count": 96,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Results for WordNet Link Prediction task\n",
+      "+--------------------------------------------+---------------------------------+\n",
+      "|                                            |        Model Dimensions         |\n",
```
janpom: Table header misaligned.

docs/notebooks/Poincare Evaluation.ipynb
```diff
@@ -0,0 +1,1448 @@
...
+    "                item_2_index = embedding_vocab[row[1]].index\n",
+    "                relations[item_1_index].add(item_2_index)\n",
+    "                items.update([item_1_index, item_2_index])\n",
+    "        self.items = items\n",
+    "        self.relations = relations\n",
+    "        self.embedding = embedding\n",
+    "    \n",
+    "    \n",
+    "    @staticmethod\n",
+    "    def get_positive_relation_ranks(distances, positive_relations):\n",
```
janpom: There's a bit too much magic in this method implementation. Are you sure it actually does what it's expected to do? Have you tested it?

docs/notebooks/Poincare Evaluation.ipynb
```diff
@@ -0,0 +1,1448 @@
...
+    "            item_relations = list(self.relations[item])\n",
+    "            item_term = self.embedding.kv.index2word[item]\n",
+    "            item_distances = self.embedding.get_all_distances(item_term)\n",
+    "            positive_relation_ranks = self.get_positive_relation_ranks(item_distances, item_relations)\n",
+    "            ranks += positive_relation_ranks\n",
+    "            if max_n is not None and i > max_n:\n",
+    "                break\n",
+    "        return np.mean(ranks)\n",
+    "    \n",
+    "    def evaluate_map(self, max_n=None):\n",
```
janpom: What's the plan with `evaluate_map`?

docs/notebooks/Poincare Evaluation.ipynb
jayantj: Good point, changed.

docs/notebooks/Poincare Evaluation.ipynb
jayantj: Done.

docs/notebooks/Poincare Evaluation.ipynb
jayantj: Fixed

docs/notebooks/Poincare Evaluation.ipynb
jayantj: Yeah I agree the code isn't straightforward. I changed `distances` to `all_distances` to make it slightly clearer. And yes, I've tested it - 
```python
all_distances = np.array([5, 10, 15, 20, 0, 1, 2, 3])
positive_relations = [4, 6, 1, 7]
assert(ReconstructionEvaluation.get_positive_relation_ranks(all_distances, positive_relations) == [1, 2, 3, 2])
```

docs/notebooks/Poincare Evaluation.ipynb
jayantj: Ah yes, I'd forgotten. Will implement.

docs/notebooks/Poincare Evaluation.ipynb
janpom: My concern wasn't about naming, but rather about the used constructions. Mainly
```python
ranks = (negative_relation_distances < positive_relation_distances[:, np.newaxis]).sum(axis=1) + 1
```
is pretty non-obvious. I have tested it myself, though, and indeed does seem to work, so that's good.

menshikh-iv:
@jayantj please ignore CI now, don't worry about it.

docs/notebooks/Poincare Evaluation.ipynb
```diff
@@ -0,0 +1,1553 @@
...
+     "text": [
+      "82115 nouns\n",
+      "743241 hypernyms\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Prepare the WordNet data\n",
+    "wordnet_file = os.path.join(data_directory, 'wordnet_noun_hypernyms.tsv')\n",
+    "! python {parent_directory}/poincare-embedding/scripts/create_wordnet_noun_hierarchy.py {wordnet_file}"
```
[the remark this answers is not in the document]
janpom: After correcting the path I'm getting:
```python
  File "/data_ssd/poincare/gensim/docs/notebooks/poincare/poincare-cpp-embedding/scripts/create_wordnet_noun_hierarchy.py", line 4, in <module>
    from nltk.corpus import wordnet as wn
ImportError: No module named nltk.corpus
```

We may need to add `nltk` to requirements.

docs/notebooks/Poincare Evaluation.ipynb
```diff
@@ -0,0 +1,1553 @@
...
+    "! cmake ..\n",
+    "! make\n",
+    "% cd {current_directory}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You might need to install an updated version of `cmake` to be able to compile the source code. Please make sure that the binary `poincare-embedding` has been created before proceeding."
```
janpom: Please don't ask the user to do something that's trivial to do automatically. Just check if the file exists and raise an exception if it doesn't. Otherwise, the execution flow continues and fails much later. At that point it's no longer obvious what the root cause is.

docs/notebooks/Poincare Evaluation.ipynb
```diff
@@ -0,0 +1,1553 @@
...
+    "if not os.path.exists(np_repo_name):\n",
+    "    ! git clone https://github.com/nishnik/poincare_embeddings.git {np_repo_name}\n",
+    "\n",
+    "cpp_repo_name = 'poincare-cpp-embedding'\n",
+    "if force_setup and os.path.exists(cpp_repo_name):\n",
+    "    ! rm -rf {cpp_repo_name}\n",
+    "if not os.path.exists(cpp_repo_name):\n",
+    "    ! git clone https://github.com/TatsuyaShirakawa/poincare-embedding.git {cpp_repo_name}\n",
+    "\n",
+    "patches_applied = False"
```
janpom: This looks wrong. If this is re-run with `force_setup == False` the patches will get applied twice.

docs/notebooks/Poincare Evaluation.ipynb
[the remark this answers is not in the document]
janpom: ... and `click` as well.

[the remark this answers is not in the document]
janpom: Thinking more about it, it doesn't seem a great idea to make these gensim requirements just because of the notebook. Let's just do
```
import nltk
import click
```
in the first cell in the notebook so that it fails there if these libs are missing. It should be obvious for the user that they need to install these. (I don't expect non-expert users to ever run the notebook.)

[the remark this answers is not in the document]
jayantj: Yeah, makes sense. I added the imports at the beginning, and a link to a separate `requirements.txt` in `poincare/` for easy installation of missing imports.

docs/notebooks/Poincare Evaluation.ipynb
jayantj: Done

docs/notebooks/Poincare Evaluation.ipynb
jayantj: Fixed

PR #1647, opened 2017-10-24
Fix test_filename_filtering test

nehaljwani:
CI tests fail with:
```
FAIL: test_filename_filtering (gensim.test.test_corpora.TestTextDirectoryCorpus)
  File ".../lib/python3.6/site-packages/gensim/test/test_corpora.py", line 462, in test_filename_filtering
      self.assertEqual(expected, filenames)
      AssertionError: Lists differ: ['/tmp/tmp0j1tou_7/test1.log', '/tmp/tmp0j1tou_7/test2.log'] != ['/tmp/tmp0j1tou_7/test2.log', '/tmp/tmp0j1tou_7/test1.log']
```
It's not a real failure, since the files are correct, only their order of
comparison is not same


menshikh-iv:
Nice catch @nehaljwani, congratz with first PR!

PR #1649, opened 2017-10-24
Fix deprecation warnings for regex string literals. Fix #1646

### LINKED ISSUE #1646: Fix Regex Deprecation Warnings in Python 3
#### Description

When I import gensim in Python 3, I get a lot of deprecation warnings like

```
DeprecationWarning: invalid escape sequence \[
  RE_P2 = re.compile("(\n\[\[[a-z][a-z][\w-]*:[^:\]]+\]\])+$", re.UNICODE)
```

which is a 2 to 3 issue which I believe can be fixed by adding a raw flag before the string.

On Python 3...

```
import gensim
```

Observe the warnings.

#### Expected Results

No deprecation warnings.

#### Actual Results

Deprecation warnings.

#### Versions

Python 3.6.3, gensim 3.0.0

franklsf95:
Fix deprecation warnings for regex string literals. Fix #1646  …
Add raw flag before all Regex strings so Python 3 can stop complaining.

### DIFF
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -41,23 +41,23 @@ TOKEN_MIN_LEN = 2
 TOKEN_MAX_LEN = 15
 
 
-RE_P0 = re.compile('<!--.*?-->', re.DOTALL | re.UNICODE)  # comments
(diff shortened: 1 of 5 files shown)


menshikh-iv:
Thanks @franklsf95, need to check that you don't break anything, I mean that all regex are equal now.

menshikh-iv:
Thanks @franklsf95, congratz with first PR :+1:

Issue #1519, opened 2017-08-04
Fix python warnings (occur during testing)


menshikh-iv:
Need to fix warnings from interpreter (during test run).
It's easy to see all:
`python -Wd -t -3` for python2
`python -Wd -b` for python3

menshikh-iv:
I looked at it and realized that it makes no sense to do it separately. This can be done in the process of refactoring tests (in the future).

poornagurram:
@menshikh-iv I would like to submit a PR for this. I am a beginner contributor. can you elaborate on this?.

menshikh-iv:
@poornagurram of course, I'll review #1655

poornagurram:
@menshikh-iv I didn't get a clear idea about this issue can you elaborate more on this.

menshikh-iv:
@poornagurram of course,

Firstly (for simple investigation) you need to run tests with python3 and enabled warnings, howto:
- Clone gensim repo
- Create clean venv
- `pip install -e .[test]` in gensim folder
- python -Wd -b setup.py tests

you'll see many different warnings (about unclosed files, division, etc), you need to go to concrete tests that generate this warning, fix warnings and submit PR.

menshikh-iv:
Many `ResourseWarning` from `smart_open`, this part fixed in [smart_open/140](https://github.com/RaRe-Technologies/smart_open/pull/140), another yet part fixed in #1660

ParseThis:
Is this still a good beginner contrib issue? Would like to have a look. @menshikh-iv

menshikh-iv:
@ParseThis this is a good issue for start, but need to make many small changes (if you have enough time - you are welcome)

menshikh-iv:
I fixed almost all of it in #2191

PR #1655, opened 2017-10-25
PR related to issue #1269

### LINKED ISSUE #1269: Word2Vec model to dict; Adding to the word2vec to production pipeline
A lot of users use their trained `word2vec` model in production environments to get `most_similar` words to (for example) words in a user's entered query, or words in complete documents, on the fly. In times like these, using the word2vec model becomes very cumbersome, *often taking the most amount of time in the pipeline.* [1]

What I propose is a `model_to_dict` method, to be used right at the end of the word2vec pipeline. It would find and store, in preproduction, all most similar words to all words in the trained vocabulary.

The most similar words can be from a custom user list as in #1229 and we can allow the user to define a custom preprocessing function to pass all most similar words through before storing them. Being a dict, the query time will also be minimal which is great for this purpose! Since our dict just stores words, the size of the dict should be comparable to a multiple of the size of the vocab [2]

At the end of this, we expect a dictionary with keys as word2vec vocab and values as the most_similar words to them. Words in vocab that have empty most_similar words will not be stored in the dict. This will happen a lot if there is a custom results list as well as a pass function for result words applied on top of most similar cutoff.

[1] This is because we always calculate all cos distances from the query word to all words in vocab before returning topn most similar words. I can't think of a better way for that, yet.
[2] albeit a large multiple if users do not provide proper preprocessing pass function or have a small similarity cutoff. Maybe we can give them a warning about the same.

poornagurram:
Added function to give preproduction ready dictionary of most similar words and ability to pass custom most similar words dict

### DIFF
--- a/gensim/test/test_fasttext_wrapper.py
+++ b/gensim/test/test_fasttext_wrapper.py
@@ -158,14 +158,14 @@ class TestFastText(unittest.TestCase):
         ]
         self.assertTrue(numpy.allclose(model["rejection"], expected_vec_oov, atol=1e-4))
 
-        self.assertEquals(model.min_count, 5)
-        self.assertEquals(model.window, 5)
-        self.assertEquals(model.iter, 5)
-        self.assertEquals(model.negative, 5)
-        self.assertEquals(model.sample, 0.0001)
-        self.assertEquals(model.bucket, 1000)
-        self.assertEquals(model.wv.max_n, 6)
-        self.assertEquals(model.wv.min_n, 3)
+        self.assertEqual(model.min_count, 5)
+        self.assertEqual(model.window, 5)
+        self.assertEqual(model.iter, 5)
+        self.assertEqual(model.negative, 5)
+        self.assertEqual(model.sample, 0.0001)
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Please don't edit notebooks manyally (you break current notebook, it does not display correctly).
Run ipython notebook, open in browser, make all needed changes, **re-run** notebook (to make sure that everything works) and after - commit/push.

Also, please add example of usage for this function (because now this don't used anywhere) and short description, why this functionality can be useful.

poornagurram:
@menshikh-iv this PR is a reference to the issue #1269, not #1519.

menshikh-iv:
Yes, my mistake, thanks @poornagurram

menshikh-iv:
@poornagurram please firx ipython notebook here and move commits for #1519 to distinct PR (please don't mix up different fixes for unrelated issues)

menshikh-iv:
I close this PR, because it's easier to make new with needed commits for #1269 instead reverting & fixing here.

Please make new PR with example in notebook (don't forget about the comments above)

poornagurram:
Yeah sure thank you :)

Issue #1576, opened 2017-09-08
Check what's a reason to use double-precision in TMs


menshikh-iv:
Our TMs return vectors with double-precision `float64`, it looks like very suspicious, because `float32` is enough for all. Need to check, what's a reason of this behavior and what's a concrete method.

The first step - look at [this line in the test](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/basetests.py#L51), after it - collect all TMs, that depends on this tests and check, where and why `float64` happened.

Result - detailed description (where and why), and fixing this behavior after discussion (if needed)

gojomo:
FWIW, the Word2Vec/Doc2Vec classes use `float32` in vectors/calculations. (I even did experiments converting models to `float16` as an optimization; unfortunately it seems bulk array calculations still coerce back to/from `float32`, meaning no time savings and space savings only at rest.)

xelez:
Started investigation.

Test classes that use `TestBaseTopicModel`

 * `TestAuthorTopicModel`
 * `TestHdpModel`
 * `TestLdaMallet`
 * `TestLdaModel`
 * `TestLsiModel`


Went on to `LdaModel`.

Basically, numpy uses `float64` by default, and in `LdaModel` is never specified explicitly.

 * `LdaState.__init__()` uses `np.zeros()` for `sstats` (which then used for computing topics)
 * `LdaModel.__init__()` does `self.state.sstats = self.random_state.gamma(100., 1. / 100., (self.num_topics, self.num_terms))`
 * `nd.asarray` is used in many places that deduces type from argument by default
 * ndarrays can be passed in `__init__()`, and they will be used as-is (float64, float32 or whatever)

So, as a result, everything just defaults to `float64` so `float64` is returned in `get_topics()`.

menshikh-iv:
Good investigation, thanks @xelez.
I think the best solution is to add `dtype` argument in the constructor and **explicitly** set `dtype` everywhere when needed (instead of default `float64`). Potentially, we'll have same problems for another method (including for training). If this is true - we can easily reduce memory twice & speedup model :fire:

xelez:
@menshikh-iv Are there any plans on refactoring? If yes, I think this should be done with refactoring.
If no, well.. The only way I can see is to examine every line of code identifying places to apply `dtype`.

Well, I think I can do it for `LdaModel`, and maybe other models later. But if anyone will be working on it too, it could easily become a nightmare to merge.

[...]

PR #1656, opened 2017-10-26
Adding dtype to LDAModel to speed it up

xelez:
Started implementing #1576

Current state:

 * added dtype to LdaModel
 * added asserts about dtype everywhere to be sure that haven't missed any conversion
 * probably need help in handling load/save, maybe link to docs about how it works

And I need to somehow rewrite test asserts like this:
`self.assertTrue(all(model.alpha == np.array([0.3, 0.3])))`

Cause `model.alpha` is now converted to `float32` (or whatever dtype) and `np.array` standard dtype is `float64`. Use `np.allclose`, maybe?

And I'm not sure where to discuss things, here or in the issue.


piskvorky:
Good idea with the asserts!

I don't think save/load need any special handling at all. Save just saves the object, and the load loads it back (using the same types as when the object was saved).

Maybe the only tricky part is how to handle backward compatibility: should loading models saved *before* this change work? I'd say yes. We need to test this explicitly: save an "old" model, then load it using your new code with dtypes and asserts, make sure everything works.

The other compatibility direction (load new model in old code) is not necessary.

xelez:
Do I need to do anything to make code save my new `dtype` field as well?

And yes, there is a compatibility problem as tests shown me. To achieve it, I'll need to set dtype to float64 if it's not present in the saved model. I'll need some time to wrap my head around code in load/save methods.

menshikh-iv:
Nice @xelez, my suggestions about it:

1. Default `dtype` **must be** `np.float32`.
2. As I remember, you no need to make anything else for save your `dtype` fields (all will be handled automatically). But you need to modify `load` for **old** models (without this field).
3. About backward compatibility: 
- add a check, this model old or new (check that `dtype` is defined), if defined, no need to make anything else (this model is new)
- if not - check what's `dtype` was used for matrices and fill-up new instance with this `dtype`. Also, add the warning about it (dtype isn't loaded, we'll use ...)

xelez:
Implemented setting of `dtype` for old models.

Two things left to do:
 * modify tests to work with float32
 * cleanup asserts and TODO's that I've added

xelez:
Looked up failing tests:

I'll also need to modify `AuthorTopicModel` and maybe other classes based on `LdaModel`

xelez:
Fixed tests and quick-fixed AuthorTopicModel.

gensim/models/ldamodel.py
```diff
@@ -354,25 +373,25 @@ def init_dir_prior(self, prior, name):
 
         if isinstance(prior, six.string_types):
             if prior == 'symmetric':
-                logger.info("using symmetric %s at %s", name, 1.0 / prior_shape)
-                init_prior = np.asarray([1.0 / self.num_topics for i in xrange(prior_shape)])
+                logger.info("using symmetric %s at %s", name, 1.0 / prior_shape) #TODO: prior_shape?
```
xelez: I have feeling that it should be

```
"using symmetric %s at %s", name, 1.0 / self.num_topics
```

am I right? @menshikh-iv

gensim/matutils.py
```diff
@@ -600,19 +600,13 @@ def jaccard_distance(set1, set2):
 def dirichlet_expectation(alpha):
     """
     For a vector `theta~Dir(alpha)`, compute `E[log(theta)]`.
-
+    Saves dtype of the argument.
```
piskvorky: What does this comment mean? Looks out of place.

xelez: I wanted to add some note that this function returns np.array with the same dtype as input alpha. Well, probably it's not really needed.

piskvorky: If that's the intent, it's not really apparent from `Saves dtype of the argument`.

gensim/models/ldamodel.py
menshikh-iv: Yes @xelez, you are correct!

xelez:
@piskvorky , @menshikh-iv I think I've finished.

gensim/models/ldamodel.py
```diff
@@ -1231,7 +1230,7 @@ def load(cls, fname, *args, **kwargs):
 
         # the same goes for dtype (except it was added later)
         if not hasattr(result, 'dtype'):
-            result.dtype = np.float64 # float64 was used before as default in numpy
+            result.dtype = np.float64  # float64 was used before as default in numpy
             logging.warning("dtype was not set, so using np.float64")
```
piskvorky: A more concrete message please. When reading this warning, users will be left scratching their heads: set where? Why? What does this mean to me?

How about `"dtype not set in saved %s file %s, assuming np.float64" % (result.__class__.__name__, fname)`? 
And only log at INFO or even DEBUG level, since it's an expected state when loading an old model, nothing out of ordinary.

Question: isn't it better to infer the dtype from the loaded object? Can it ever happen that it's something else, not `np.float64`?

gensim/models/hdpmodel.py
```diff
@@ -538,7 +538,8 @@ def suggested_lda_model(self):
         The num_topics is m_T (default is 150) so as to preserve the matrice shapes when we assign alpha and beta.
         """
         alpha, beta = self.hdp_to_lda()
-        ldam = ldamodel.LdaModel(num_topics=self.m_T, alpha=alpha, id2word=self.id2word, random_state=self.random_state)
+        ldam = ldamodel.LdaModel(num_topics=self.m_T, alpha=alpha, id2word=self.id2word,
+                                 random_state=self.random_state, dtype=np.float64)
```
piskvorky: Code style: no vertical indent.

gensim/models/ldamodel.py
```diff
@@ -87,10 +87,11 @@ class LdaState(utils.SaveLoad):
 
     """
 
-    def __init__(self, eta, shape):
-        self.eta = eta
-        self.sstats = np.zeros(shape)
+    def __init__(self, eta, shape, dtype=np.float32):
+        self.eta = eta.astype(dtype, copy=False)
+        self.sstats = np.zeros(shape, dtype)
```
piskvorky: Using positional arguments can lead to subtle bugs with numpy. Better use explicit names for keyword parameters: `dtype=dtype`.

gensim/models/ldaseqmodel.py
```diff
@@ -244,7 +245,8 @@ def lda_seq_infer(self, corpus, topic_suffstats, gammas, lhoods,
         vocab_len = self.vocab_len
         bound = 0.0
 
-        lda = ldamodel.LdaModel(num_topics=num_topics, alpha=self.alphas, id2word=self.id2word)
+        lda = ldamodel.LdaModel(num_topics=num_topics, alpha=self.alphas, id2word=self.id2word,
+                                dtype=np.float64)
```
piskvorky: Code style: no vertical indent.

gensim/models/ldaseqmodel.py
```diff
@@ -419,7 +421,8 @@ def __getitem__(self, doc):
         """
         Similar to the LdaModel __getitem__ function, it returns topic proportions of a document passed.
         """
-        lda_model = ldamodel.LdaModel(num_topics=self.num_topics, alpha=self.alphas, id2word=self.id2word)
+        lda_model = ldamodel.LdaModel(num_topics=self.num_topics, alpha=self.alphas, id2word=self.id2word,
+                                      dtype=np.float64)
```
piskvorky: Code style: no vertical indent.

gensim/models/ldamodel.py
```diff
@@ -165,6 +166,18 @@ def get_lambda(self):
...
         return dirichlet_expectation(self.get_lambda())
+
+    @classmethod
+    def load(cls, fname, *args, **kwargs):
+        result = super(LdaState, cls).load(fname, *args, **kwargs)
+
+        # Check if `dtype` is set after main pickle load
+        # if not, then it's an old model and we should set it to default `np.float64`
+        if not hasattr(result, 'dtype'):
+            result.dtype = np.float64  # float64 was used before as default in numpy
```
menshikh-iv: Old LDA used float64, really?

gensim/models/ldaseqmodel.py
```diff
@@ -130,7 +130,8 @@ def __init__(self, corpus=None, time_slice=None, id2word=None, alphas=0.01, num_
             if initialize == 'gensim':
                 lda_model = ldamodel.LdaModel(
                     corpus, id2word=self.id2word, num_topics=self.num_topics,
-                    passes=passes, alpha=self.alphas, random_state=random_state
+                    passes=passes, alpha=self.alphas, random_state=random_state,
+                    dtype=np.float64
```
menshikh-iv: Maybe it will be a good idea to change default behaviour (to float32)?
CC @piskvorky @xelez

gensim/test/basetmtests.py
```diff
@@ -48,6 +48,7 @@ def test_get_topics(self):
         vocab_size = len(self.model.id2word)
         for topic in topics:
             self.assertTrue(isinstance(topic, np.ndarray))
-            self.assertEqual(topic.dtype, np.float64)
+            # Note: started moving to np.float32 as default
+            # self.assertEqual(topic.dtype, np.float64)
```
menshikh-iv: need to enable + switch to `float32`

gensim/test/test_matutils.py
```diff
@@ -0,0 +1,33 @@
...
+import logging
+import unittest
+
+import numpy as np
+
+from gensim import matutils
+
+
+class TestMatUtils(unittest.TestCase):
+    def test_dirichlet_expectation_keeps_precision(self):
```
menshikh-iv: I don't think that make new file is a good idea. Please move this tests to LDA tests class.

menshikh-iv: Also, please add tests for the load **old** model with **new code** for all models that you changed.

piskvorky: Test not just "load", but also using the loaded models.

The asserts that we newly sprinkled into the code may trigger errors in various places, if something is wrong.

gensim/models/ldamodel.py
xelez: Fixed message, decided info level suites better.

About inferring. Not clear how to do it. Infer from `LdaState.eta` and `LdaState.sstats`? But then we had test that their sum is `np.float64`, so it's safe to assume that we don't loose precision when setting dtype to `np.float64` and `np.float32` is not enough.

Anyway, let's imagine situation some of `nd.arrays` are somehow of different dtype, like `np.float32` and some are `np.float64`. The right `dtype` is still `np.float64`.

gensim/models/ldamodel.py
xelez: Pretty much everything is using `float64` cause it's default dtype when creating arrays.

gensim/models/ldaseqmodel.py
xelez: Not now, `LdaSeqModel` will require modifications similar to those I made in `LdaModel` to handle dtype properly.

gensim/test/basetmtests.py
xelez: This will break other topic models then

gensim/test/test_matutils.py
xelez: @menshikh-iv That test don't involve LDA at all, so it's wrong place for this test. Haven't found any file involving testing matutils functions so I think a separate file is not that bad.

xelez:
@piskvorky , @menshikh-iv see latest commit for backwards compatibility tests.

By the way, I think it's good idea to remove my asserts before merging. They were used mostly during tests to ensure that I haven't missed any place to add dtype. That way we definitely won't break old code or models.

xelez:
@piskvorky , @menshikh-iv see latest commit for backwards compatibility tests.

By the way, I think it's good idea to remove my asserts before merging. They were used mostly during tests to ensure that I haven't missed any place to add dtype. That way we definitely won't break old code or models.

xelez:
By the way, why .npy files are ignored in .gitignore?

gensim/models/ldamodel.py
```diff
@@ -820,6 +856,7 @@ def show_topics(self, num_topics=10, num_words=10, log=False, formatted=True):
 
             # add a little random jitter, to randomize results around the same alpha
             sort_alpha = self.alpha + 0.0001 * self.random_state.rand(len(self.alpha))
+            # random_state.rand returns float64, but converting back to dtype won't speed up anything
```
menshikh-iv: Maybe `.astype` (for consistency only) ?

gensim/models/ldamodel.py
```diff
@@ -1186,9 +1224,14 @@ def load(cls, fname, *args, **kwargs):
             result.random_state = utils.get_random_state(None)  # using default value `get_random_state(None)`
             logging.warning("random_state not set so using default value")
 
+        # dtype could be absent in old models
+        if not hasattr(result, 'dtype'):
+            result.dtype = np.float64  # float64 was implicitly used before (cause it's default in numpy)
+            logging.info("dtype was not set in saved %s file %s, assuming np.float64", result.__class__.__name__, fname)
```
menshikh-iv: Maybe `warn`?

gensim/matutils.py
```diff
@@ -600,13 +600,12 @@ def jaccard_distance(set1, set2):
...
     """
     For a vector `theta~Dir(alpha)`, compute `E[log(theta)]`.
-
     """
     if len(alpha.shape) == 1:
         result = psi(alpha) - psi(np.sum(alpha))
     else:
         result = psi(alpha) - psi(np.sum(alpha, 1))[:, np.newaxis]
-    return result.astype(alpha.dtype)  # keep the same precision as input
+    return result
```
menshikh-iv: Please return `astype`, because
`np.float32 -> np.float32`
`np.float64 -> np.float64`
but
`np.float16 -> np.float32`

gensim/test/test_ldamodel.py
```diff
@@ -242,7 +243,7 @@ def testGetDocumentTopics(self):
             self.assertTrue(isinstance(topic, list))
             for k, v in topic:
                 self.assertTrue(isinstance(k, int))
-                self.assertTrue(isinstance(v, float))
+                self.assertTrue(np.issubdtype(v, float))
```
menshikh-iv: simple `isinstance` is better here (and everywhere).

gensim/test/test_matutils.py
```diff
@@ -0,0 +1,33 @@
...
+        for dtype in (np.float32, np.float64, np.complex64, np.complex128):
```
menshikh-iv: Add np.float16 and you'll see a problem

gensim/matutils.py
xelez: Oh, my bad, you're right!
Then tests that I added in separate file aren't needed.

gensim/models/ldamodel.py
xelez: Consistency vs one additional array copy. I'm not sure :)

gensim/models/ldamodel.py
xelez: See https://github.com/RaRe-Technologies/gensim/pull/1656#discussion_r148215039 for discussion. Although it's expected state when loading old model, maybe a warning still a good thing.

gensim/test/test_ldamodel.py
xelez: simple `isinstance` fails cause `float32` is not an instance of 'float'

gensim/models/hdpmodel.py
xelez: fixed

gensim/models/ldamodel.py
xelez: fixed

gensim/models/ldaseqmodel.py
xelez: fixed

gensim/models/ldaseqmodel.py
xelez: fixed

menshikh-iv:
Thanks a lot @xelez, nice work :+1:

piskvorky:
Great feature!

@xelez how would you summarize it, for layman people who just want "the gist"? The title says `Adding dtype to LDAModel to speed it up` -- what was the final speed-up?

menshikh-iv:
@piskvorky with LdaMulticore (80k words, 100 topics) ~ 20%

piskvorky:
That's neat :) Let's make sure this information makes it into the release notes / tweets etc.

rmalouf:
There are a couple of places in ldamodel.py where 1e-100 gets added to `phinorm` to avoid division by zero.  That constant will need to be adjusted depending on the precision being used.

piskvorky:
Good point. I'd hope this would be caught by the unit tests though -- @menshikh-iv ?

menshikh-iv:
@piskvorky this isn't catched by unittests, because operation `+ 1e-100`  don't change original dtype.
@rmalouf nice catch! thanks for your remark:+1: 
@xelez can you fix this? (need to scale `1e-100` according to arrays dtype https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ldamodel.py#L476 and https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ldamodel.py#L487 in LDA + same fixes in another TMs).

piskvorky:
The unit tests should catch the operation of "add epsilon" not working.

In other words, if the unit tests pass, what is the problem?

menshikh-iv:
For types `< float64`, it can produce problems with zeros if `phinorm` is zero -> division by zero -> `nan`
```
In [1]: import numpy as np
In [2]: np.float64(1e-100)
Out[2]: 1e-100

In [3]: np.float32(1e-100)
Out[3]: 0.0
```
For me, it looks like medium bug.

piskvorky:
That's not my question. My question is: do unit tests catch it?

If not, is it an issue with the unit tests (=> update unit tests), or with the algorithm (=> update gensim code)?

menshikh-iv:
@piskvorky unittests doesn't catch it.

piskvorky:
Then the unit tests should be improved.

menshikh-iv:
The problem here is not in tests at all, it's generally impossible to catch this bug in this code with unittests. Here, perhaps, we need to change the lda code itself, but I do not think this is a good idea.

piskvorky:
I don't understand -- if there's no way to catch a bug, then there is no bug.

rmalouf:
It's definitely test-for-able in principle: I noticed the bug because I started getting division by zero errors in a processing pipeline that used to work.  I don't know how create a minimal corpus that triggers it, though.

menshikh-iv:
Summarizing: this is a bug, we need to fix it.

PR #1660, opened 2017-10-28
Fix ResourceWarnings in tests

horpto:


gensim/utils.py
```diff
@@ -135,7 +134,6 @@ def __exit__(self, type, value, traceback):
 nocm = NoCM()
 
 
-@contextmanager
 def file_or_filename(input):
```
menshikh-iv: As I see in usage, this function used as contextmanager, why you remove this (`@contextmanager` + `yield`) ?

gensim/matutils.py
```diff
@@ -895,8 +895,11 @@ def docbyoffset(self, offset):
             assert previd <= docid, "matrix columns must come in ascending order"
             if docid != previd:
                 if previd >= 0:
-                    return document
+                    break
```
menshikh-iv: Can you add a test for this?

horpto: Hmm, ok. But I don't understand what exactly should I check in it ?

menshikh-iv: I read full method definition, now it's clear for me, thanks.

gensim/utils.py
menshikh-iv: Oh, it really doesn't work as was expected, nice catch!

horpto: After meged comment:

Actually, it was already brokenl.  Compare with code:
```
from contextlib import contextmanager

@contextmanager
def g():
    print("g input")
    yield
    print("g exit")

@contextmanager
def f():
    print("f input")
    yield g()
    print("f exit")


with f():
    print("with block")
```

This code prints:
```
f input
with block
f exit
```

But expected something like:
f input
g input
with block
g exit
f exit
```

Back to the `file_or_filename`, methods `__enter__` and `__exit__` of `smart_open(input)` or `input` are not called at all. Oops.

Issue #1644, opened 2017-10-23
Fix list of ignored problems for flake8


menshikh-iv:
Now we using `--ignore=E501,E731,E12,W503,E402` parameters for `flake8`, we can be more "strict" and reduce list of "ignored" checks

What this checks means (full list [here](http://pep8.readthedocs.io/en/latest/intro.html#error-codes))

| Code | Message | Action (from us)|
|---------|--------------|---------|
|E501|line too long (>80 characters)| remove (but need track size with adequate limit, like 120)|
|E731|do not assign a lambda expression, use a def| remove |
|W503|line break occurred before a binary operator| - |
|E402|module level import not at top of file | remove |
|E12| Many checks ... | remove (need to think about E121, E123, E126) |

Related #1636

menshikh-iv:
For now, need to think about `E12` group and (more important), make extended-E501.

menshikh-iv:
Last part fixed in #1721

PR #1661, opened 2017-10-29
Add max-line-length param to flake8

paragkulkarni:
Fixes a part of #1644 

This is my first patch for Gensim


menshikh-iv:
Hi @paragkulkarni, you make something strange - why you removed `flake8` call?
In this part (based on PR name), we want to track string limit and raise an error, but with less strict limit (120 instead of 80).

menshikh-iv:
@paragkulkarni are you planning to fix your PR?

Issue #1651, opened 2017-10-25
Mutable vector returned by KeyedVectors.word_vector


menshikh-iv:
#### Description
TODO: change commented example

#### Steps/Code/Corpus to Reproduce

Method `KeyedVectors.word_vector` returned "mutable" vector if we call `model['anywords']` (for `model[['anywords']]` works correctly because [vstack make a copy.](https://github.com/RaRe-Technologies/gensim/blob/269028975e0db48e37e01edfb54e66018db7b61b/gensim/models/keyedvectors.py#L600)

Simple example

```python
from gensim.models import KeyedVectors

model = KeyedVectors.load_word2vec_format('path/to/model')
vector = model['word_from_model']
assert ~np.allclose(vector, np.zeros(vector.shape))  # check that our vector isn't zero
vector *= 0.

assert ~np.allclose(model['word_from_model'], np.zeros(vector.shape))  # failed, because now model['word_from_model'] is zero vector

```

#### Expected Results
assert passed

#### Actual Results
assert failed

#### What needs to fix
Add `arr.setflags(write=False)`  in [word_vec](https://github.com/RaRe-Technologies/gensim/blob/269028975e0db48e37e01edfb54e66018db7b61b/gensim/models/keyedvectors.py#L266)

gojomo:
What's wrong with providing a mutable vector? It's very rare to even try to mutate the result, so adding the overhead of a copy, on every access, seems a bad tradeoff to me, and returning a mutable view on the actual backing array the right default behavior.

menshikh-iv:
@gojomo I was approached by a person from one company, they at work caught this bug. As for me, it's very strange if you change a vector - your `model` changed too.

About performance - yes, it will be slightly slower, but I don't think that it will be greatly affect performance, I benchmarked it

### Preparation
```python
import gensim
import numpy as np
from random import sample

model = gensim.models.KeyedVectors.load_word2vec_format("wiki-news-300d-1M.vec")
rand_words = sample(model.index2word, k=100000)
```

### Benchmark

Without any changes
```python
%%timeit

for w in rand_words:
    model[w]
```
`14.4 ms ± 300 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`

With explicit copy
```python
%%timeit

for w in rand_words:
    np.array(model[w])
```
`21.9 ms ± 335 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)`

[...]

PR #1662, opened 2017-10-29
Fix 1651: make word_vec() return immutable vector

CLearERR:
fix for issue about mutable vector returned by KeyedVectors.word_vector

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -284,6 +284,8 @@ class KeyedVectors(utils.SaveLoad):
 
             result.setflags(write=False)
             return result
+        else:
+            raise KeyError("word '%s' not in vocabulary" % word)
 
     def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=None, indexer=None):
         """
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -282,7 +282,7 @@ class KeyedVectors(utils.SaveLoad):
             else:
                 result = self.syn0[self.vocab[word].index]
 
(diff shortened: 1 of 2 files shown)


gensim/models/keyedvectors.py
```diff
@@ -278,11 +278,12 @@ def word_vec(self, word, use_norm=False):
         """
         if word in self.vocab:
             if use_norm:
-                return self.syn0norm[self.vocab[word].index]
+                result = self.syn0norm[self.vocab[word].index]
             else:
-                return self.syn0[self.vocab[word].index]
-        else:
```
menshikh-iv: Why you remove this branch? We must raise an exception if word not in the dictionary.

gensim/models/keyedvectors.py
```diff
@@ -278,11 +278,12 @@ def word_vec(self, word, use_norm=False):
...
-            raise KeyError("word '%s' not in vocabulary" % word)
+                result = self.syn0[self.vocab[word].index]
+
+            result.setflags(write=False)
```
menshikh-iv: Please add test to check this behaviour (with Exception catching)

gojomo:
Ideally PR names should be descriptive even if issue-numbers (like `#1651`) are ignored/unknown. So I've updated the name to be more descriptive.

piskvorky:
On github, PR titles do not link through to the issues (no link in issue, no easy to way to visit the 
 issue).

Please use native Github formatting, it makes maintenance easier.

menshikh-iv:
Thanks @CLearERR, congratz with the first PR :1st_place_medal:

PR #1679, opened 2017-11-01
* optimize __len__ of SlicedCorpus

horpto:
* small refactoring of SlicedCorpus and IndexedCorpus
* remove unnecessary file


menshikh-iv:
@horpto thanks, you are very active for last time :+1:

piskvorky:
Awesome! These type of fixes are sorely needed. Many thanks @horpto .

horpto:
Thanks, @piskvorky! It's just little fix. The Worse problem is SlicedCorpus.__iter__: `docbyoffset` will reopen file for each doc. ><  I cannot fix it as I don't want to destroy the abstraction of IndexedCorpus & not all Corpuses supports `line2doc` of something like this.

piskvorky:
I think we should just implement a binary format, such as binary Matrix Market (.mm, `MmCorpus`). 

It will be efficient and much faster, for those people who care about speed more than readability. The format is pretty trivial too => easy task. CC @menshikh-iv .

menshikh-iv:
Can you describe your proposal more concretely @piskvorky (I'll create an issue)?

piskvorky:
I mean, store the (sparse) matrix represented by a corpus in a binary format: less space + faster processing.

For example, store each document as `number_of_document_nonzeros (int32)` followed by serialized `(feature_id, feature_weight)` pairs as `(int32, float32)` = 8 bytes per non-zero element. So, total size = `num_docs * 4 + 8 * nnz`.

That's one simple option (doesn't support slicing / indexing though).

More optimized binary schemes are possible, such as encoding the feature ids more cleverly, to save space, or adding indexing capabilities, but I don't think that's strictly necessary. Certainly not as the first step.

PR #1686, opened 2017-11-02
Disable using various workers in tests for FastText

horpto:
Current version of FastText is pure-python,
so CPU-bound processing with multiple workers in it is a bad idea.
Using workers=12 in a some tests increased execution time of tests.

### DIFF
--- a/gensim/test/test_fasttext.py
+++ b/gensim/test/test_fasttext.py
@@ -368,7 +368,7 @@ class TestFastTextModel(unittest.TestCase):
 
         model_gensim = FT_gensim(size=50, sg=0, cbow_mean=1, alpha=0.05, window=5, hs=1, negative=0,
             min_count=5, iter=5, batch_words=1000, word_ngrams=1, sample=1e-3, min_n=3, max_n=6,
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Thanks @horpto, can you compare the time of `test_fasttext` before/after?
CC: @chinmayapancholi13 WDYT about this change?

menshikh-iv:
I run each test 3 times and averaged the results.

| Test | Before (sec) | After (sec) | Speedup (x)|
|------|--------------|-------------|------------|
| test_cbow_hs_against_wrapper | 115.7 | 52.5 | 2.2 |
| test_sg_hs_against_wrapper | 189.6 | 92.1 | 2.0 |
| test_sg_hs_online | 23.5 | 13.1 | 1.7 |
| test_sg_neg_online | 29.9 | 16.8 | 1.8 |
| test_cbow_hs_online | 18.0 | 9.0 | 2.0 |
| test_cbow_neg_online | 20.3 | 11.1| 1.8 |

Nice speedup for slowest tests, good job @horpto!

PR #1693, opened 2017-11-05
fix #1682, custom dtype of keyedvectors

### LINKED ISSUE #1682: `datatype` parameter in `load_word2vec_format` doesn't work as expected
#### Description
Using `datatype=np.float64` in a `KeyedVectors.load_word2vec_call` doesn't work as expected, the loaded floats seem to lose precision. The datatype for `syn0` is still `float64` though, so it seems that they are cast to float32 first while loading, then cast to float64 when creating the array.

Using this file - 
[test.kv.txt](https://github.com/RaRe-Technologies/gensim/files/1434953/test.kv.txt)

```python
import numpy as np

kv = KeyedVectors.load_word2vec_format('test.kv.txt', datatype=np.float64)
print(kv['horse.n.01'][0] == -0.0008546282343595379)
# False
print(kv['horse.n.01'].dtype)
# float64
```
#### Expected Results
```python
print(kv['horse.n.01'][0] == -0.0008546282343595379)
# True
```
#### Actual Results
```python
print(kv['horse.n.01'][0] == -0.0008546282343595379)
# False
```

Looking at the code and making a quick hack [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/keyedvectors.py#L253), changing..
```
word, weights = parts[0], [REAL(x) for x in parts[1:]]
```
to..
```
word, weights = parts[0], [datatype(x) for x in parts[1:]]
```
..leads to the correct result. However, I imagine there are other cases to be covered as well.

horpto:


menshikh-iv:
@horpto also please have a look to tests, this looks like real fail (not false-positive).

menshikh-iv:
ping @horpto, how is going?

menshikh-iv:
ping @horpto

horpto:
close this PR as this feature should be rethinked.

PR #1694, opened 2017-11-05
Fix segment-wiki script

menshikh-iv:
What's done:
- Updated field names (now it's more descriptive, instead of `ts`, `sc`);
- Updated documentation about output format (available with `-h` option);
- Added stdout support (as default option);
- Removed prunning through tokenization (and comments about it).

### DIFF
--- a/gensim/scripts/segment_wiki.py
+++ b/gensim/scripts/segment_wiki.py
@@ -54,7 +54,7 @@ from smart_open import smart_open
 logger = logging.getLogger(__name__)
 
 
-def segment_all_articles(file_path):
+def segment_all_articles(file_path, min_article_character=200):
     """Extract article titles and sections from a MediaWiki bz2 database dump.
 
     Parameters
@@ -63,6 +63,9 @@ def segment_all_articles(file_path):
         Path to MediaWiki dump, typical filename is <LANG>wiki-<YYYYMMDD>-pages-articles.xml.bz2
         or <LANG>wiki-latest-pages-articles.xml.bz2.
 
+    min_article_character : int, optional
+        Minimal number of character for article (except titles and leading gaps).
+
     Yields
     ------
     (str, list of (str, str))
@@ -70,14 +73,14 @@ def segment_all_articles(file_path):
 
     """
     with smart_open(file_path, 'rb') as xml_fileobj:
-        wiki_sections_corpus = _WikiSectionsCorpus(xml_fileobj)
+        wiki_sections_corpus = _WikiSectionsCorpus(xml_fileobj, min_article_character=min_article_character)
         wiki_sections_corpus.metadata = True
         wiki_sections_text = wiki_sections_corpus.get_texts_with_sections()
         for article_title, article_sections in wiki_sections_text:
             yield article_title, article_sections
 
 
-def segment_and_write_all_articles(file_path, output_file):
+def segment_and_write_all_articles(file_path, output_file, min_article_character=200):
     """Write article title and sections to output_file,
     output_file is json-line file with 3 fields::
 
@@ -94,6 +97,9 @@ def segment_and_write_all_articles(file_path, output_file):
     output_file : str
         Path to output file in json-lines format.
 
+    min_article_character : int, optional
+        Minimal number of character for article (except titles and leading gaps).
+
     """
     if output_file is None:
         outfile = sys.stdout
@@ -101,7 +107,7 @@ def segment_and_write_all_articles(file_path, output_file):
         outfile = smart_open(output_file, 'wb')
 
     try:
-        for idx, (article_title, article_sections) in enumerate(segment_all_articles(file_path)):
+        for idx, (article_title, article_sections) in enumerate(segment_all_articles(file_path, min_article_character)):
             output_data = {"title": article_title, "section_titles": [], "section_texts": []}
             for section_heading, section_content in article_sections:
                 output_data["section_titles"].append(section_heading)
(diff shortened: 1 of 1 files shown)


gensim/scripts/segment_wiki.py
```diff
@@ -5,12 +5,14 @@
 # Copyright (C) 2016 RaRe Technologies
 
 """
-Construct a corpus from a Wikipedia (or other MediaWiki-based) database dump and extract sections of pages from it
-and save to json-line format.
+Construct a corpus from a Wikipedia (or other MediaWiki-based) database dump (typical filename
+is <LANG>wiki-<YYYYMMDD>-pages-articles.xml.bz2 or <LANG>wiki-latest-pages-articles.xml.bz2),
```
piskvorky: Add a link to the actual place to download those, because it's not obvious.

For example, the English Wiki dump is here: https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2

gensim/scripts/segment_wiki.py
```diff
@@ -5,12 +5,14 @@
...
+extract titles, section names, section content and save to json-line format,
+that contains 3 fields ::
 
-If you have the `pattern` package installed, this module will use a fancy
-lemmatization to get a lemma of each token (instead of plain alphabetic
-tokenizer). The package is available at https://github.com/clips/pattern .
+    'title' (str) - title of article,
+    'section_titles' (list) - list of titles of sections,
+    'section_texts' (list) - list of content from sections.
```
piskvorky: I'd prefer to include a concrete example, something like this:

---

Process a raw Wikipedia dump (XML.bz2 format, for example https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2 for the English Wikipedia) and extract all articles and their sections as plain text::

    python -m gensim.scripts.segment_wiki -f enwiki-20171001-pages-articles.xml.bz2 -o enwiki-20171001-pages-articles.json.gz

The output format of the parsed Wikipedia is *json-lines* = one article per line, serialized into JSON. Here's an example how to load it in Python::

    # iterate over the file we just created
    for line in smart_open('enwiki-20171001-pages-articles.txt.gz'):
        # decode JSON into a Python object
        article = json.loads(line)

        # each article has a "title", "section_titles" and "section_texts" fields
        print("Article title: %s" % article['title']) 
        for section_title, section_text in zip(article['section_titles'], article['section_texts']):
            print("Section title: %s" % section_title)
            print("Section text: %s" % section_text)

gensim/scripts/segment_wiki.py
```diff
@@ -217,15 +225,8 @@ def get_texts_with_sections(self):
...
         for group in utils.chunkize(page_xmls, chunksize=10 * self.processes, maxsize=1):
             for article_title, sections in pool.imap(segment, group):  # chunksize=10):
-                # article redirects and short stubs are pruned here
-                num_total_tokens = 0
-                for section_title, section_content in sections:
-                    if self.lemmatize:
-                        num_total_tokens += len(utils.lemmatize(section_content))
-                    else:
-                        num_total_tokens += len(tokenize(section_content))
-                if num_total_tokens < ARTICLE_MIN_WORDS or \
```
piskvorky: Including redirects and stubs is a bad idea. That's typically (never?) what people want, out of Wikipedia dumps. 

We want to keep only meaningful articles, such as at least 500 plain text characters (~1 paragraph) or something.

menshikh-iv: I don't think so (about short articles), because we provide parsed wikipedia dump "as-is" and short articles can be useful for users for special cases (and easy to filter later if needed), for this reason, I removed this part.

menshikh-iv: But it's needed to filter trash (like the redirect), I'll add fix for this.

piskvorky: Sounds good 👍

gensim/scripts/segment_wiki.py
```diff
@@ -226,7 +249,9 @@ def get_texts_with_sections(self):
         for group in utils.chunkize(page_xmls, chunksize=10 * self.processes, maxsize=1):
             for article_title, sections in pool.imap(segment, group):  # chunksize=10):
                 # article redirects are pruned here
-                if any(article_title.startswith(ignore + ':') for ignore in IGNORED_NAMESPACES):
+                if any(article_title.startswith(ignore + ':') for ignore in IGNORED_NAMESPACES) \
+                        or len(sections) == 0 \
```
piskvorky: `not sections` more Pythonic

gensim/scripts/segment_wiki.py
```diff
@@ -249,10 +249,13 @@ def get_texts_with_sections(self):
...
             for article_title, sections in pool.imap(segment, group):  # chunksize=10):
                 # article redirects are pruned here
-                if any(article_title.startswith(ignore + ':') for ignore in IGNORED_NAMESPACES) \
-                        or len(sections) == 0 \
-                        or sections[0][1].lstrip().startswith("#REDIRECT"):
+                if any(article_title.startswith(ignore + ':') for ignore in IGNORED_NAMESPACES):  # filter non-articles
                     continue
+                if len(sections) == 0 or sections[0][1].lstrip().lower().startswith("#redirect"):  # filter redirect
+                    continue
+                if sum(len(body.strip()) for (_, body) in sections) < 250:  # filter very short articles (thrash)
```
piskvorky: `thrash` => `trash` ; but it's more stubs than trash.

The constant (250) should be configurable, not hardwired like this.

PR #1696, opened 2017-11-06
[MRG] Poincare Model implementation

jayantj:
Pure Python implementation of the Poincare model from [[1]](https://arxiv.org/pdf/1705.08039.pdf).

TODO -

- [x] Unit tests
- [x] API conformity
- [x] More logging

Follow up PR: #1700 

[[1] Poincaré Embeddings for Learning Hierarchical Representations](https://arxiv.org/pdf/1705.08039.pdf)


gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+        negative : int, optional
+            Number of negative samples to use, defaults to 10.
+        iter : int, optional
+            Number of iterations (epochs) over the corpus, defaults to 50.
+        workers : int, optional
+            Number of threads to use for training the model, defaults to 1.
+        epsilon : float, optional
+            Constant used for clipping embeddings below a norm of one, defaults to 1e-5.
+        burn_in : int, optional
+            Number of epochs to use for burn-in initialization (0 means no burn-in), defaults to 0.
```
janpom: defaults to 10 according to method signature

gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+        self.burn_in = burn_in
+        self.seed = seed
+        self.random = random.Random(seed)
+        self.np_random = np_random.RandomState(seed)
+        self.init_range = (-0.001, 0.001)
+        self.loss_grad = grad(PoincareModel.loss_fn)
+        self.load_relations()
+        self.init_embeddings()
+
+    def load_relations(self):
```
janpom: Having to initialize the relations from a file seems unfortunate. If the user happens to have a list of pairs in memory, the API gets clunky. I think we should be able to initialize from an iterable of pairs as well (or instead; getting the iterable of pairs from a file is easy and can be done outside of this class).

gensim/models/poincare.py
menshikh-iv: let's not write default values at all, because this is useless duplication.

gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+
+        Returns
+        --------
+        numpy array
+            numpy array of shape (`self.negative`,) containing indices of negative nodes.
+        """
+
+        if self.negatives_buffer_index >= len(self.negatives_buffer):
+            # Note: np.random.choice much slower than random.sample for large populations, possible bottleneck
+            self.negatives_buffer = self.np_random.choice(self.indices_array, size=self.negatives_buffer_size, p=self.node_probabilities)
```
janpom: line too long

gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+
+        self.wv.vocab = vocab
+        self.wv.index2word = index2word
+        self.indices_set = set((range(len(index2word))))  # Set of all node indices
+        self.indices_array = np.array(range(len(index2word)))  # Numpy array of all node indices
+        counts = np.array([self.wv.vocab[index2word[i]].count for i in range(len(index2word))])
+        self.node_probabilities = counts / counts.sum()
+        self.all_relations = all_relations
+        self.term_relations = term_relations
+        self.negatives_buffer = []  # Buffer to store negative samples, to reduce calls to sampling method
```
janpom: It would be nice to encapsulate the buffer to a separate class. Would remove some complexity from this class.

gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+            numpy array of shape (self.negative,) containing indices of negative nodes for the given node index.
+        """
+        node_relations = self.term_relations[node_index]
+        positive_fraction = len(node_relations) / len(self.term_relations)
+        if positive_fraction < 0.01:
+            # If number of positive relations is a small fraction of total nodes
+            # re-sample till no positively connected nodes are chosen
+            indices = self.get_candidate_negatives()
+            times_sampled = 1
+            while len(set(indices) & node_relations):
```
janpom: `len` redundant

gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+            contains node indices, computed gradients and loss for the batch.
+        """
+        batch_size = len(relations)
+        all_vectors = []
+        indices_u, indices_v = [], []
+        for relation, negatives in zip(relations, all_negatives):
+            u, v = relation
+            indices_u.append(u)
+            indices_v.append(v)
+            for negative in negatives:
```
janpom: `indices_v.extend(negatives)`

gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+        for epoch in range(1, epochs + 1):
+            indices = list(range(len(self.all_relations)))
+            self.np_random.shuffle(indices)
+            avg_loss = 0.0
+            last_time = time.time()
+            for batch_num, i in enumerate(range(0, len(indices), batch_size), start=1):
+                print_check = not (batch_num % print_every)
+                batch_indices = indices[i:i+batch_size]
+                relations = [self.all_relations[idx] for idx in batch_indices]
+                result = self.train_on_batch(relations, check_gradients=print_check)
```
janpom: I don't like reusing `print_check` for this purpose. I understand that it doesn't really matter that much how often we do it, but it's confusing. I would much rather add `check_gradients_every` parameter with the same default and document as appropritate.

gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+            raise NotImplementedError("Multi-threaded version not implemented yet")
+        if epochs is None:
+            epochs = self.iter
+        for epoch in range(1, epochs + 1):
+            indices = list(range(len(self.all_relations)))
+            self.np_random.shuffle(indices)
+            avg_loss = 0.0
+            last_time = time.time()
+            for batch_num, i in enumerate(range(0, len(indices), batch_size), start=1):
+                print_check = not (batch_num % print_every)
```
janpom: I prefer `should_do_x` over `x_check` for boolean flags of this kind. "if should print" reads more naturally than "if print check".

gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+        else:
+            return np.arccosh(
+                1 + 2 * (
+                    (euclidean_dist ** 2) / ((1 - norm_1 ** 2) * (1 - norm_2 ** 2))
+                )
+            )
+    # TODO: Add other KeyedVector supported methods - most_similar, etc.
+
+
+class PoincareModel(utils.SaveLoad):
```
janpom: Minor thing, but IMO it's better to put the main class first in the file and the helper classes last.

gensim/models/poincare.py
```diff
@@ -0,0 +1,637 @@
...
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+
+
+"""
+Python implementation of Poincare Embeddings, an embedding to capture hierarchical information,
+described in [1]
+
+This module allows training a Poincare Embedding from a training file containing relations from
+a transitive closure.
+
```
janpom: Code example would be helpful.

janpom:
This looks very good overall.

gensim/models/poincare.py
jayantj: Removed

gensim/models/poincare.py
jayantj: Good point, extracted it out to a `PoincareData` class. The API accepts an iterable now.
This is also in line with other Gensim models.

gensim/models/poincare.py
jayantj: Fixed (along with a minor change in logic)

gensim/models/poincare.py
jayantj: Useful, thanks

gensim/models/poincare.py
jayantj: It's essentially just a list/numpy array + an index, but yes, it does make it conceptually simpler, and the code more readable. Done.

jayantj: Just confirming if something like this was what you meant - https://github.com/RaRe-Technologies/gensim/pull/1696/files#diff-d698c7f682aaf0f30c4694aa584a8faaR718

jayantj:
Updates: I've added the unittests to this branch, since it's a small file, and I think it actually makes reviewing easier.
I can create a new PR with them if necessary @janpom

gensim/test/test_poincare.py
```diff
@@ -0,0 +1,97 @@
...
+import unittest
+import os
+
+import numpy as np
+
+from gensim.models.poincare import PoincareData, PoincareModel
+
+
+module_path = os.path.dirname(__file__)  # needed because sample data files are located in the same folder
+datapath = lambda fname: os.path.join(module_path, 'test_data', fname)
```
menshikh-iv: Please merge `develop` to you branch and replace this two lines to `from gensim.test.utils import datapath` (yesterday we refactored tests)

gensim/test/test_poincare.py
```diff
@@ -0,0 +1,97 @@
...
+        self.assertEqual(len(relations), 2)
+        self.assertEqual(relations[0], (u'tímto', u'budeš'))
+
+        utf8_file = datapath('poincare_utf8.tsv')
+        relations = [relation for relation in PoincareData(utf8_file)]
+        self.assertEqual(len(relations), 2)
+        self.assertEqual(relations[0], (u'tímto', u'budeš'))
+
+
+class TestPoincareModel(unittest.TestCase):
```
menshikh-iv: Add tests for save/load.

gensim/models/poincare.py
menshikh-iv: Short description of "what's the main idea, what's killer-feature of this representation" will be helpful too.

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+
+import csv
+import itertools
+import logging
+import os
+import random
+import time
+
+from autograd import numpy as np, grad
```
menshikh-iv: Oh, new dependency for gensim. Why is needed (I mean why we can't differentiate loss analytically)?

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+
+    Note that training cannot be resumed from a model loaded via `load_word2vec_format`, if you wish to train further,
+    use `save()` and `load()` methods instead.
+
+    """
+    def __init__(
+        self, train_data, size=50, alpha=0.1, negative=10, iter=50,
+        workers=1, epsilon=1e-5, burn_in=10, burn_in_alpha=0.01, seed=0):
+        """
+        Initialize and train a Poincare embedding model from an iterable of transitive closure relations.
```
menshikh-iv: `"""Initialize ...`, without new line for first sentences ([example](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)), here and anywhere.

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+        epsilon : float, optional
+            Constant used for clipping embeddings below a norm of one.
+        burn_in : int, optional
+            Number of epochs to use for burn-in initialization (0 means no burn-in).
+        burn_in_alpha : float, optional
+            learning rate for burn-in initialization, ignored if `burn_in` is 0.
+        seed : int, optional
+            seed for random to ensure reproducibility.
+
+        Returns
```
menshikh-iv: No needed make `Return` section in `__init__` method (here and everywhere)

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+        --------
+        PoincareModel instance
+
+        Examples
+        --------
+        Initialize a model from a list::
+
+        >>> model = PoincareModel(list_of_relations)
```
menshikh-iv: It will be better if examples will be executable (i.e. with all needed imports + all variables defined)

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+
+    @staticmethod
+    def has_duplicates(array):
+        """
+        Returns whether or not the input array has any duplicates.
+
+        Parameters
+        ----------
+        array : iterable of hashables
+            input array to checked, should contain hashable items.
```
menshikh-iv: `input` -> `Input`, uppercase for first character (here and anywhere).

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+
+        Returns
+        -------
+        bool
+            Whether the input array contains any duplicates
```
menshikh-iv: Don't forget about `.` at the end of the sentence (here and anywhere).

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+        Return a sample of negatives for the given node.
+
+        Parameters
+        ----------
+        node_index : int
+            index of the positive node for which negative samples are to be returned.
+
+        Returns
+        --------
+        numpy array
```
menshikh-iv: numpy.array

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+            numpy array of shape (self.negative,) containing indices of negative nodes for the given node index.
```
menshikh-iv: `numpy array` -> `Array` (you already mention type), here and everywhere.

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+        """
+        node_relations = self.term_relations[node_index]
+        positive_fraction = len(node_relations) / len(self.term_relations)
+        if positive_fraction < 0.01:
+            # If number of positive relations is a small fraction of total nodes
+            # re-sample till no positively connected nodes are chosen
+            indices = self.get_candidate_negatives()
+            times_sampled = 1
+            while (set(indices) & node_relations) or self.has_duplicates(indices):
```
menshikh-iv: Maybe change the order of conditions (if `indices` with duplicates, this will be faster), or this line doesn't affect performance?

gensim/models/poincare.py
```diff
@@ -0,0 +1,803 @@
...
+        return list(indices)
+
+    @staticmethod
+    def loss_fn(matrix):
+        """
+        Given a numpy array with vectors for u, v and negative samples, computes loss value.
+
+        Parameters
+        ----------
+        matrix : numpy array
```
menshikh-iv: `numpy.array` (here and everywhere)

menshikh-iv:
Code look awesome, nice work @jayantj :fire: !
Almost all of my comments connected with docstrings, sorry for nitpicks, but now we try to refactor all documentation in gensim, for this reason, I asked to you make the similar thing with you docstrings.

gensim/models/poincare.py
```diff
@@ -0,0 +1,827 @@
...
+    def __init__(
+        self, train_data, size=50, alpha=0.1, negative=10, iter=50,
+        workers=1, epsilon=1e-5, burn_in=10, burn_in_alpha=0.01, seed=0):
+        """
+        Initialize and train a Poincare embedding model from an iterable of transitive closure relations.
+
+        Parameters
+        ----------
+        train_file : iterable
+            iterable of hypernym pairs, e.g. a list of tuples, or a PoincareData instance streaming from a file.
```
janpom: I guess "hypernym" is more specific then needed here. Can't the embedding be trained on other types of relations?

gensim/models/poincare.py
```diff
@@ -0,0 +1,827 @@
...
+    """
+    def __init__(
+        self, train_data, size=50, alpha=0.1, negative=10, iter=50,
+        workers=1, epsilon=1e-5, burn_in=10, burn_in_alpha=0.01, seed=0):
+        """
+        Initialize and train a Poincare embedding model from an iterable of transitive closure relations.
+
+        Parameters
+        ----------
+        train_file : iterable
```
janpom: "iterable of (str, str)" would make things much clearer.

Also, train_file -> train_data.

gensim/models/poincare.py
```diff
@@ -0,0 +1,827 @@
...
+from smart_open import smart_open
+
+from gensim import utils
+from gensim.models.keyedvectors import KeyedVectors, Vocab
+from gensim.models.word2vec import Word2Vec
+
+logger = logging.getLogger(__name__)
+
+
+class PoincareModel(utils.SaveLoad):
```
janpom: It's unclear what the public API is in this class. Looks like all the methods are private. It would actually make sense to prefix them with an underscore.

Then again, once the embedding is trained, it's not clear to me from the documentation (or even the code) what can I do with it.

gensim/models/poincare.py
janpom: Yes, that's pretty much what I meant. The name is good. Nice.

gensim/models/poincare.py
janpom: @jayantj Could we make it so that `autograd` is imported on demand only if `check_gradients` is requested? Looking at the code, this will require some refactoring (can't initialize `self.loss_grad` in the `__init__`), but it shouldn't be too hard to do.

It would also be nice to fail with a clear message if the import fails, such as "please `pip install autograd` if you want to check gradients" or something along those lines.

Then I suppose we can get away without adding `autograd` to gensim dependencies.

gensim/models/poincare.py
```diff
@@ -0,0 +1,827 @@
...
+        Yields
+        -------
+        2-tuple (unicode)
+            Hypernym relation from input file.
+        """
+        reader = csv.reader(self.stream_lines(), delimiter=self.delimiter)
+        for row in reader:
+            yield tuple(row)
+
+class NegativesBuffer(object):
```
janpom: PEP8: Two empty lines before a class.

gensim/models/poincare.py
```diff
@@ -0,0 +1,827 @@
...
+class PoincareModel(utils.SaveLoad):
+    """
+    Class for training, using and evaluating Poincare Embeddings described in https://arxiv.org/pdf/1705.08039.pdf
+
+    The model can be stored/loaded via its `save()` and `load()` methods, or stored/loaded in the word2vec
+    format via `wv.save_word2vec_format()` and `KeyedVectors.load_word2vec_format()`.
+
+    Note that training cannot be resumed from a model loaded via `load_word2vec_format`, if you wish to train further,
+    use `save()` and `load()` methods instead.
+
```
janpom: `<nitpick>`
You're inconsistent about empty lines. Sometime you leave the last line in the docstring empty; other times you leave the first line after the docstring empty; other times none of the two. Ordnung muss sein! :) 
`</nitpick>`

gensim/models/poincare.py
```diff
@@ -0,0 +1,827 @@
...
+        else:
+            return np.arccosh(
+                1 + 2 * (
+                    (euclidean_dist ** 2) / ((1 - norm_1 ** 2) * (1 - norm_2 ** 2))
+                )
+            )
+    # TODO: Add other KeyedVector supported methods - most_similar, etc.
+
+
+class PoincareData(object):
```
janpom: `PoincareRelations` would be more appropriate, IMO.

gensim/test/test_poincare.py
```diff
@@ -0,0 +1,173 @@
...
```
janpom: Nice test coverage!

What I'm missing is a test that checks the embedding actually works. Would it be possible to train on some very small data, such as this graph:
```
A
| \
B   C
|\  |\
D E F G
```
and assert, for instance, that `distance(A, B) < distance(A, D)`?

gensim/models/poincare.py
jayantj: Agreed, changed.

gensim/test/test_poincare.py
jayantj: I agree, I plan on adding this, it relies on the `PoincareKeyedVector` operations though, which are in the `poincare_model_keyedvectors` branch.

gensim/models/poincare.py
jayantj: Added, also added working (although trivial) examples to the module docstring.

gensim/models/poincare.py
jayantj: IMO, checking gradients should definitely be enabled by default, a user is unlikely to enable it himself in case we disable it by default, therefore running the risk of not knowing in case something goes wrong with the training. In case we are enabling gradient checking by default, it makes sense to have `autograd` in the dependencies, since having the training fail due to an uninstalled dependency would be rather annoying.

If we are completely sure about the correctness of the model, we don't need the gradient checking in the first place - however, I imagine there are cases we haven't yet discovered where something might go wrong with the training. Would it be okay to add the dependency for now, and maybe once we are confident about correctness, we remove the gradient checking and the dependency?

gensim/models/poincare.py
jayantj: Yeah, it's only the `__init__` and `train` for now. I'd like to prefix the private methods too, just going with the existing gensim convention of not doing so. I don't mind changing if @menshikh-iv is okay with it.

We definitely need a clearer explanation of what can be done with the model post-training, for that, I need to resolve and finish the `KeyedVector` operations.

gensim/models/poincare.py
menshikh-iv: Yes, I agree with you. Variant with conditional import is good enough. Let's use `autograd`, but don't add it as a dependency (only in [test_env](https://github.com/RaRe-Technologies/gensim/blob/develop/setup.py#L228) of course).

gensim/models/poincare.py
menshikh-iv: @jayantj I agree with @janpom, prefixes is better, because I see the sad situation in global: with `existing gensim convention of not doing so` new user have many methods and doesn't know what should be called. Unfortunately, we can't fix this in the current code, because this change has broken all, but we can prevent this in new code.

gensim/models/poincare.py
janpom: @jayantj Good points. The way I think about it is we shouldn't need gradient checking in the final version. Thus, it's OK to have it disabled by default. It's a useful feature for developers though, should there be any enhancements in the future, so I would definitely keep it in the code. The developers will know how to enable it.

That said, I believe what I have initially proposed would be ideal -- have the checking disabled by default; do the autograd import on demand; do not add another gensim dependency.

gensim/models/poincare.py
janpom: OK, let's prefix all the methods other than `train()` with underscore.

We could decide the same about the fields. I think it makes sense to keep most fields public since the user may want to read the values. @jayantj Please use your judgement there. Private should be the default. We don't want to clutter the API. On the other hand, we do want to keep things public where it makes sense.

gensim/models/poincare.py
```diff
@@ -0,0 +1,828 @@
...
+
+        """
+        self.train_data = train_data
+        self.wv = KeyedVectors()
+        self.size = size
+        self.train_alpha = alpha  # Learning rate for training
+        self.burn_in_alpha = burn_in_alpha  # Learning rate for burn-in
+        self.alpha = alpha  # Current learning rate
+        self.negative = negative
+        self.iter = iter
```
janpom: `iter` collides with a built-in function. `iterations` would be better, IMO.

gensim/models/poincare.py
jayantj: Sounds good. Let's do conditional import. Maybe even instead of raising an error, just adding a prominent warning that gradient checking isn't happening, and to enable gradient check, the user should install autograd, would be good enough?

gensim/models/poincare.py
jayantj: Was following word2vec API which uses `iter`. Renamed to `epochs` now, shorter and more specific/clear IMO.

gensim/models/poincare.py
jayantj: Done https://github.com/RaRe-Technologies/gensim/pull/1696/commits/98e078dbc24e5cbadb08ed09077b6be561d48ccf

In case gradient checking is enabled and autograd isn't installed, I'm logging a warning that gradients checking isn't occurring and autograd needs to be installed to do so. Can raise an exception if that is better.

gensim/models/poincare.py
jayantj: Done https://github.com/RaRe-Technologies/gensim/pull/1696/commits/530146dd91a859c5c4ccb75547e46f27c7540e0b and https://github.com/RaRe-Technologies/gensim/pull/1696/commits/be0249a42008dfbc6294fad9c246495e3c56fad6

I've kept most properties public. But I'm not sure what you mean by `private should be the default`, since the language itself doesn't have any notion of public/private (AFAIK)
Do you mean simply that we ourselves should mark any new method/property we add as public only if there's a specific reason to do so, and continue to mark them with an underscore if not?

jayantj:
I've finished all the changes we've discussed. I've also added autograd to the test dependencies in `setup.py`, @menshikh-iv could you please let me know if I need to add it anywhere else?

Lastly, I ended up changing `PoincareModel.wv` to `PoincareModel.kv`, this doesn't maintain consistency with `Word2Vec` etc. but I believe it makes more sense conceptually. Please let me know if you'd like me to name it something else.

Is there anything else required before merging this PR?

gensim/models/poincare.py
menshikh-iv: Looks good for me now, thanks!

@piskvorky wdyt about `_` and https://github.com/RaRe-Technologies/gensim/pull/1696#discussion_r149880020 ?

gensim/models/poincare.py
menshikh-iv: @jayantj I think that exception isn't needed here (because model successfully works without `autograd`), the warning is enough.

menshikh-iv:
@jayantj as you already suggested in the first post - ipython notebook (or this is already done in another branch?)

menshikh-iv:
@jayantj I remembered one imporant things, for rendered doc for site, please add `.rst` to `docs/src/models/` + update `docs/src/apiref.rst`

gensim/models/poincare.py
```diff
@@ -0,0 +1,851 @@
...
+        array : iterable of hashables
+            Input array to checked, should contain hashable items.
+
+        Returns
+        -------
+        bool
+            Whether the input array contains any duplicates.
+
+        """
+        seen = set()
```
janpom: `return len(array) != len(set(array))` simpler. Probably not even worth adding a method.

gensim/models/poincare.py
```diff
@@ -0,0 +1,851 @@
...
+
+    """
+    def __init__(
+        self, train_data, size=50, alpha=0.1, negative=10, workers=1,
+        epsilon=1e-5, burn_in=10, burn_in_alpha=0.01, init_range=(-0.001, 0.001), seed=0):
+        """Initialize and train a Poincare embedding model from an iterable of transitive closure relations.
+
+        Parameters
+        ----------
+        train_data : iterable of (str, str)
```
janpom: `str` is ambiguous for Python 2 vs 3. Better to say `unicode` or `bytes` instead.

gensim/models/poincare.py
```diff
@@ -0,0 +1,851 @@
...
+
+    def _load_relations(self):
+        """Load relations from the train data and build vocab."""
+        vocab = {}
+        index2word = []
+        all_relations = []  # List of all relation pairs
+        node_relations = defaultdict(set)  # Mapping from node index to its related node indices
+
+        logger.info("Loading relations from train data..")
+        for hypernym_pair in self.train_data:
```
janpom: Please rename `hypernym_pair` to something more generic such as `relation`.

gensim/models/poincare.py
```diff
@@ -0,0 +1,851 @@
...
+        """
+        Initialize instance with sets of vectors for which distances are to be computed.
+
+        Parameters
+        ----------
+        vectors_u : numpy.array
+            Vectors of all nodes `u` in the batch.
+            Expected shape (batch_size, dim).
+        vectors_v : numpy.array
+            Vectors of all hypernym nodes `v` and negatively sampled nodes `v'`,
```
janpom: just "nodes"

gensim/models/poincare.py
```diff
@@ -0,0 +1,851 @@
...
+                )
+            )
+    # TODO: Add other KeyedVector supported methods - most_similar, etc.
+
+
+class PoincareRelations(object):
+    """Class to stream hypernym relations for `PoincareModel` from a tsv-like file."""
+
+    def __init__(self, file_path, encoding='utf8', delimiter='\t'):
+        """Initialize instance from file containing one hypernym pair per line.
```
janpom: hypernym pair -> relation (here and elsewhere)

jayantj: Done

gensim/models/poincare.py
```diff
@@ -0,0 +1,851 @@
...
+            return np.arccosh(
+                1 + 2 * (
+                    (euclidean_dist ** 2) / ((1 - norm_1 ** 2) * (1 - norm_2 ** 2))
+                )
+            )
+    # TODO: Add other KeyedVector supported methods - most_similar, etc.
+
+
+class PoincareRelations(object):
+    """Class to stream hypernym relations for `PoincareModel` from a tsv-like file."""
```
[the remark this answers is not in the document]
jayantj: Done

gensim/models/poincare.py
jayantj: Done

gensim/models/poincare.py
jayantj: Done

gensim/models/poincare.py
jayantj: Done

gensim/models/poincare.py
jayantj: Both `unicode` and `bytes` are allowed here. Wherever this is true, I've used `str`, wherever a specific type is required returned, I've used `unicode`/`bytes`. Does that sound okay?

jayantj:
I've added the rst files and made some fixes for python2 bugs. The only failing test is the one that requires autograd now (due to it being missing from test dependencies). With autograd added to test dependencies, the build errors (due to some MKL error, as you mentioned).

menshikh-iv:
@jayantj maybe remove this test (because we can't run it correctly in CI)?

gensim/models/poincare.py
janpom: Yes, perfect. Are you sure it works correctly with bytes, though? I suppose if we train on bytes we'll end up with a bytes based model. I wonder if that's common for other gensim models. Won't that cause unexpected behavior with some `KeyedVectors` calls?

gensim/models/poincare.py
```diff
@@ -0,0 +1,831 @@
...
+    format via `wv.save_word2vec_format()` and `KeyedVectors.load_word2vec_format()`.
+
+    Note that training cannot be resumed from a model loaded via `load_word2vec_format`, if you wish to train further,
+    use `save()` and `load()` methods instead.
+
+    """
+    def __init__(
+        self, train_data, size=50, alpha=0.1, negative=10, workers=1,
+        epsilon=1e-5, burn_in=10, burn_in_alpha=0.01, init_range=(-0.001, 0.001), seed=0):
+        """Initialize and train a Poincare embedding model from an iterable of transitive closure relations.
```
janpom: Is the transitive closure a requirement? If not, let's just say "iterable of relations".

jayantj: Fixed.

jayantj:
@menshikh-iv I've instead added a skiptest in case autograd is not installed, that way we can continue to check if the test runs locally, making development easier. Does that seem okay?

PR #1699, opened 2017-11-07
Reduce distributive size. Fix #1698

### LINKED ISSUE #1698: Reduce size of gensim distrib
For the current moment, gensim distribution is very heavy (>60Mb for `.tar.gz`). Need to significantly reduce this value, for example, remove notebooks from tar.gz and so on.

**How to produce `tar.gz`**
`python setup.py sdist`

**How to control content in `tar.gz`**
Change [MANIFEST.in](https://github.com/RaRe-Technologies/gensim/blob/develop/MANIFEST.in)

menshikh-iv:
What's done:
- removed `docker/`
- removed `continious_integration/`
- removed `docs/`

Result: `63MB -> 14MB` for `tar.gz`

Last large part is distrib is `gensim/test/test_data`, biggest file is:
- `large_tag_doc_10_iter50` - `4.8MB`
- `small_tag_doc_5_iter50` - `2.3MB`
- `head500.noblanks.cor` - `2.3MB`
- `pang_lee_polarity_fasttext.vec` - `1.8MB`
- `enwiki-latest-pages-articles1.xml-p000000010p000030302-shortened.bz2` - `1.7MB`
- `lee_fasttext` - `1.1MB`

But I think that we no needed to replace it & fix tests, wdyt @piskvorky?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,4 @@
 recursive-include gensim/test/test_data *
-recursive-exclude . *.sh
-prune docs/src*
-prune docs/notebooks/datasets
 include README.md
 include CHANGELOG.md
 include COPYING
@@ -13,4 +10,3 @@ include gensim/models/word2vec_inner.pyx
 include gensim/models/word2vec_inner.pxd
 include gensim/models/doc2vec_inner.c
 include gensim/models/doc2vec_inner.pyx
-prune docs/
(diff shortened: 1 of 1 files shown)


piskvorky:
Can't we distributed these large data files as `.gz`? We're using `smart_open` anyway.

menshikh-iv:
I think that most of this are binary and will be compressed poorly @piskvorky

piskvorky:
Yeah, that's true. Scratch that :)

PR #1616, opened 2017-10-08
[WIP] Topic model visualization

parulsethi:
This PR adds the visualization for interactive exploration of doc/topic/word entities in topic models (as discussed).

TODOs:

- [ ] Display hooks - to embed visualization in notebook or display in browser
- [ ] Gensim wrapper - API for direct gensim corpus/dictionary inputs

PR #1700, opened 2017-11-08
[MRG] Poincare model keyedvectors

jayantj:
This branch includes the changes to `KeyedVectors` and `PoincareKeyedVectors` for the `PoincareModel`.

Includes commits from #1696 , to be reviewed and merged after #1696 

TODOs:
- [x] Evaluating Gensim implementation and adding results to notebook
- [x] Adding `KeyedVector` methods like `most_similar` to `PoincareKeyedVectors`
- [x] Refactoring `KeyedVectors` into `KeyedVectorsBase` and `EuclideanKeyedVectors` for cleaner class hierarchy
- [x] Tests for `PoincareKeyedVectors`


jayantj:
The addition of `PoincareModel` raises some questions about `KeyedVectors`. Not all the operations provided by `KeyedVectors` make sense for certain kind of vector-sets that might wish to inherit from `KeyedVectors` (`PoincareVectors`, in this case), however they still share a fair amount of overlap, both conceptually and in code -

1. Every vector is associated with a label/node/word ("key")
2. There is a notion of distance/similarity between two vectors (although the exact calculation of distance may be  different for each type of embedding), and therefore, also between two "keys".

The operations that `KeyedVectors` supports that only make sense for word2vec/doc2vec/fastText vectors (which is what `KeyedVectors` had been used for so far) - 
1. `wmdistance`
2. `accuracy`
3. `most_similar_cosmul`
3. Operations that depend on a linear combination of vectors to obtain a "result" vector (`most_similar_cosmul`, `n_similarity`, `doesnt_match`, `most_similar` with multiple input terms)

I think, ideally, `KeyedVectors` should be limited in scope to only allow common operations (lookup, similarity, distance), however it is tricky in practice, since it'd require breaking the `KeyedVectors` API. I see two options - 

1. Extract out the word2vec specific methods from `KeyedVectors` to a new class (say `WordVectors` for now) and use that class from now on for Word2Vec/FastText/Doc2Vec vectors (breaks `KeyedVectors` API)
2. Let `KeyedVectors` stay as it is, and if any class inherits from it, for the unsupported methods, it raises  an error. (not a clean solution, not good in the long term)

gensim/models/keyedvectors.py
```diff
@@ -263,6 +255,121 @@ def add_word(word, weights):
...
+        raise NotImplementedError
+
+    def word_vec(self, word):
+        """
+        Accept a single word as input.
+        Returns the word's representations in vector space, as a 1D numpy array.
+
+        Example::
+
+          >>> trained_model['office']
```
janpom: Weird example. Should be `word_vec` call instead?

gensim/models/keyedvectors.py
```diff
@@ -873,3 +965,6 @@ def get_keras_embedding(self, train_embeddings=False):
             weights=[weights], trainable=train_embeddings
         )
         return layer
+
+# For backward compatibility
+KeyedVectors = EuclideanKeyedVectors
```
janpom: PEP8: no newline at the EOF

gensim/models/keyedvectors.py
```diff
@@ -103,28 +104,19 @@ def __str__(self):
         return "%s(%s)" % (self.__class__.__name__, ', '.join(vals))
 
 
-class KeyedVectors(utils.SaveLoad):
+
+class KeyedVectorsBase(utils.SaveLoad):
```
janpom: PEP8: too many blank lines

gensim/models/keyedvectors.py
```diff
@@ -263,6 +255,121 @@ def add_word(word, weights):
         logger.info("loaded %s matrix from %s", result.syn0.shape, fname)
         return result
 
+    def similarity(self, word_1, word_2):
```
janpom: The parameters are called `word_1, word_2` here, `w1, w1` in Euclidean an `term_1, term_2` in Poincare. A bit of consistency would be nice. The `distance()` is the same story.

gensim/models/poincare.py
```diff
@@ -731,27 +739,457 @@ def compute_loss(self):
...
+        >>> model.similarity('mammal.n.01', 'carnivore.n.01')
+        0.73
+
+        Notes
+        -----
+        Raises KeyError if either of `term_1` and `term_2` is absent from vocab.
+        Similarity lies between 0 and 1.
+
+        """
+        return 1 / (1 + self.distance(term_1, term_2))
```
janpom: IIRC, we wanted to normalize the distance (`norm_distance = distance / max_distance`) and then return `1 - norm_distance` as similarity. I'm re-thinking it now. One feature would be that the similarity would be 0 for the max distance. I'm not sure if that's desired. Maybe the current is OK.

gensim/models/poincare.py
```diff
@@ -731,27 +739,457 @@ def compute_loss(self):
...
+        norm = np.linalg.norm(vector_1)
+        all_norms = np.linalg.norm(vectors_all, axis=1)
+        return np.arccosh(
+            1 + 2 * (
+                (euclidean_dists ** 2) / ((1 - norm ** 2) * (1 - all_norms ** 2))
             )
-    # TODO: Add other KeyedVector supported methods - most_similar, etc.
+        )
+
+    def nodes_closer_than(self, term_1, term_2):
```
janpom: This looks generic. Would it make sense to move to the parent class?

gensim/models/poincare.py
```diff
@@ -731,27 +739,457 @@ def compute_loss(self):
...
+        ['dog.n.01', 'canine.n.02']
+
+        """
+        all_distances = self.distances(term_1)
+        term_1_index = self.vocab[term_1].index
+        term_2_index = self.vocab[term_2].index
+        closer_node_indices = np.where(all_distances < all_distances[term_2_index])[0]
+        return [self.index2word[index] for index in closer_node_indices if index != term_1_index]
+
+    def rank(self, term_1, term_2):
```
janpom: This looks generic. Would it make sense to move to the parent class?

gensim/models/poincare.py
```diff
@@ -731,27 +739,457 @@ def compute_loss(self):
...
+
+        Notes
+        -----
+        Raises KeyError if either of `term_1` and `term_2` is absent from vocab.
+        Similarity lies between 0 and 1.
+
+        """
+        return 1 / (1 + self.distance(term_1, term_2))
+
+    def most_similar(self, word_or_vector, topn=10, restrict_vocab=None):
```
janpom: Is there a reason the API is different from Euclidean?

gensim/models/poincare.py
```diff
@@ -731,27 +739,457 @@ def compute_loss(self):
...
+            closest_indices = matutils.argsort(all_distances, topn=1 + topn)
+        result = [
+            (self.index2word[index], float(all_distances[index]))
+            for index in closest_indices if (not word_index or index != word_index)  # ignore the input word
+        ]
+        if topn:
+            result = result[:topn]
+        return result
+
+    def distances(self, word_or_vector, words_2=[]):
```
janpom: It's a bad practice to use mutable types as default parameter values. `words_2=()` better.

gensim/models/poincare.py
```diff
@@ -47,10 +47,14 @@
 import numpy as np
 from collections import defaultdict, Counter
 from numpy import random as np_random
+from pygtrie import Trie
```
janpom: Not in requirements. Use local import instead?

gensim/models/poincare.py
```diff
@@ -47,10 +47,14 @@
...
+from scipy.spatial.distance import euclidean
```
janpom: unused import

janpom:
Not a part of these changes, but I have only noticed it now:
```
positive_fraction = len(node_relations) / len(self.kv.vocab)
```

This will be integer division in Python 2. Probably not what you want.

gensim/models/poincare.py
```diff
@@ -731,27 +739,457 @@ def compute_loss(self):
...
+
+        """
+        if isinstance(word_or_vector_1, string_types) and isinstance(word_or_vector_2, string_types):
+            input_vector_1 = self.word_vec(word_or_vector_1)
+            input_vector_2 = self.word_vec(word_or_vector_2)
+        elif isinstance(word_or_vector_1, np.ndarray) and isinstance(word_or_vector_2, np.ndarray):
+            input_vector_1 = word_or_vector_1
+            input_vector_2 = word_or_vector_2
+
+        return np.linalg.norm(input_vector_2) - np.linalg.norm(input_vector_1)
```
janpom: `input_vector_1, input_vector_2` may be unitialized here.

gensim/models/poincare.py
```diff
@@ -846,3 +1285,363 @@ def get_items(self, num_items):
...
+
+    @staticmethod
+    def get_positive_relation_ranks_and_avg_prec(all_distances, positive_relations):
+        """
+        Given a numpy array of all distances from an item and indices of its positive relations,
+        compute ranks and Average Precision of positive relations.
+
+        Parameters
+        ----------
+        distances : numpy.array (float)
```
janpom: doesn't match parameter name

gensim/models/poincare.py
janpom: How about just:
```
return position_in_hierarchy(word_or_vector_1) - position_in_hierarchy(word_or_vector_2)
```

I'm not sure this method is needed at all.

gensim/models/poincare.py
jayantj: Yes, keeping this in the current form, due to the significant overhead of computing max distance.

gensim/models/poincare.py
jayantj: Good idea, done

gensim/models/poincare.py
jayantj: Good idea, done

gensim/models/poincare.py
jayantj: As discussed, keeping it different from `EuclideanKeyedVectors` since `EuclideanKeyedVectors` allows passing in multiple positive and negative arguments, where `PoincareKeyedVectors` don't. Removed method from base class.

gensim/models/poincare.py
jayantj: Fixed

gensim/models/poincare.py
jayantj: Thanks, removed

gensim/models/poincare.py
jayantj: Fixed

jayantj: Refactored the method, IMO it makes sense to keep as a convenience method.

jayantj:
Nice catch for the integer division, fixed.

jayantj:
Short summary and rationale for the changes to `KeyedVectors` in this PR - 

1. Refactoring of the previously existing `KeyedVectors` class into `KeyedVectorsBase` and `EuclideanKeyedVectors`. For backwards compatibility, the `keyedvectors` module contains a reference `KeyedVectors` which points to `EuclideanKeyedVectors`. It may be a good idea to rename `KeyedVectors` elsewhere in the codebase to `EuclideanKeyedVectors`.

2. `KeyedVectorsBase` is simply a collection of vectors and associated labels, supporting the following old methods - 
    - `__getitem__`
    - `__contains__`
    - `word_vec`
    - `similarity`
    - `most_similar_to_given`
    - `load_word2vec_format/save_word2vec_format`
    Along with the following new methods -
    - `distance`
    - `distances`
    - `words_closer_than`
    - `rank`
    Note that the `KeyedVectorsBase` class does not provide definitions for the `distance`, `distances` and `similarity` methods, it is upto the child class to define them as appropriate.

3. `EuclideanKeyedVectors` is derived from `KeyedVectorsBase` and contains functionality that is only relevant/meaningful for vectors in Euclidean space. Both word2vec and fasttext vectors fall into this category.

4. `PoincareKeyedVectors` is a new class to contain vectors in hyperbolic space for the Poincare model, which supports operations specific to the vectors for the Poincare model.

5. The `most_similar` method conceptually makes sense for both `PoincareKeyedVectors` and `EuclideanKeyedVectors`, however as the already existing API for `most_similar` could not be supported in `PoincareKeyedVectors`, `PoincareKeyedVectors` provides an implementation of `most_similar` with a different API. For this reason, `most_similar` is not present in the `KeyedVectorsBase` class.

The PR also adds missing tests for some of the older `KeyedVectors` methods.

gensim/models/poincare_visualization.py
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env python
```
menshikh-iv: I think better write this code in ipython notebook (or maybe move to `docs/notebooks` and import from "under the feet").
`gensim.model` isn't a suitable place for this file.

gensim/models/poincare.py
```diff
@@ -514,6 +516,8 @@ def train(self, epochs, batch_size=10, print_every=1000, check_gradients_every=N
         """
         if self.workers > 1:
             raise NotImplementedError("Multi-threaded version not implemented yet")
+        # Some divide-by-zero results are handled explicitly
+        old_settings = np.seterr(divide='ignore', invalid='ignore')
```
menshikh-iv: Why? You mean that division by zero is expected and you process this situation in code, I'm correct?

gensim/models/poincare.py
```diff
@@ -731,27 +737,389 @@ def compute_loss(self):
...
-    # TODO: Add other KeyedVector supported methods - most_similar, etc.
+        )
+
+    def closest_child(self, node):
+        """
+        Returns the node closest to `node` that is lower in the hierarchy than `node`.
+
+        Parameters
+        ----------
+        node : str
```
menshikh-iv: This always `str` (or int possible too)? This question more global (about all methods) that pass `node` argument?

gensim/models/poincare.py
jayantj: Yeah, it happens in `PoincareBatch`. I'm setting it here to avoid repeated calls to `np.seterr`

gensim/models/poincare.py
jayantj: It could be an `int` too in theory, depending on what the vocab keys are. The most common case is `str` though. How would you prefer to handle this?

gensim/models/poincare_visualization.py
jayantj: I agree `gensim.models` isn't a good place for it. I don't want to have it in an ipython notebook or `docs/notebooks` though since that would mean a user can't import and use it, and I think it is definitely useful for a user. Do you think creating a new package `poincare` in `gensim/` would be a good idea? Other models do this too (e.g. `topic_coherence`)

gensim/models/poincare.py
menshikh-iv: maybe `str` or `int` everywhere?

gensim/models/poincare_visualization.py
menshikh-iv: I don't think that this is a very good idea, after refactoring, `topic_coherence` will be moved/renamed in "deep" of `gensim.modules` (coherence contains only inner/secondary functions, not public API). 

I agree about imports (it's really untrivial, how to import from `docs/notebooks` if you in `/randomfolder`, only manually with `importlib` I think.

We have many viz helpers (produced by @parulsethi on GSoC) + now Parul works on very nice with for topic models. Potentially, we can create the distinct repository (like `gensim-data`) and move all viz helpers, or, as you suggest, create submodule `gensim.viz` and move all viz stuff (not only your Poincare viz).

Hard question, I don't know what's better right now.

WDYT @piskvorky @janpom @parulsethi?

jayantj: A submodule `gensim.viz` sounds good to me, keeping in mind we might have future visualizations too. I don't have a good enough perspective on this though, so whatever you decide is okay with me.

gensim/models/poincare.py
jayantj: Sounds good

jayantj: Done. Also made some other changes to docstrings for more clarity.

jayantj:
Some completely unrelated tests seem to be failing on travis (`test_translation_matrix`, `test_lda_model`). Not sure what that is about.

gensim/models/poincare_visualization.py
jayantj: Hi @menshikh-iv , so is it okay if I create a `gensim.py` package and move the poincare visualization there?

jayantj: @piskvorky @parulsethi

piskvorky:
@menshikh-iv we need this resolved & finished -- can you have a look? Cheers.

gensim/models/poincare_visualization.py
parulsethi: `gensim.viz` submodule would be useful for https://github.com/RaRe-Technologies/gensim/pull/1616 also in future, and the long code blocks of [network graph](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/topic_network.ipynb)/[dendrogram](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/Topic_dendrogram.ipynb) could also be wrapped up in a function under this module so that those visualizations can be produced simply using the imports.

menshikh-iv: @jayantj sounds good.

jayantj: Great! Thanks for the feedback @parulsethi @menshikh-iv 
Pushed changes with a new `gensim.viz` package.

PR #1702, opened 2017-11-08
Reworded and added the newline behaviour in the docstring for the…

fbarrios:
…summarize method (see #1575).


menshikh-iv:
Thanks for the link @fbarrios!

PR #1703, opened 2017-11-09
Fix scipy logsumexp deprecation warning

dreamgonfly:
**Problem**
I noticed that when training LDA model with the latest version of scipy I get a deprecation warning.

```python
>>> ldamodel = LdaModel(corpus, num_topics=4, id2word=dictionary, passes=20, alpha=‘auto’, eval_every=5)
gensim/models/ldamodel.py:776: DeprecationWarning: `logsumexp` is deprecated!
```

**Reason**
I dug into the problem and found that scipy 1.0.0 has been released recently.
https://scipy.github.io/devdocs/release.1.0.0.html

In the release they deprecated a number of functions from `scipy.misc`, and `logsumexp` is one of them. The link below is the corresponding commit.
https://github.com/scipy/scipy/pull/7246

**Solution**
I fixed the warning by making `logsumexp` to be imported from `scipy.special`, as they recommend.
Also I checked that, except in LdaModel, `logsumexp` is not used in any other places in gensim .

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -48,7 +48,7 @@ from gensim.models import basemodel, CoherenceModel
 from gensim.models.callbacks import Callback
 
 # log(sum(exp(x))) that tries to avoid overflow
-from scipy.misc import logsumexp
+from scipy.special import logsumexp
 
 
 logger = logging.getLogger('gensim.models.ldamodel')
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -48,7 +48,10 @@ from gensim.models import basemodel, CoherenceModel
 from gensim.models.callbacks import Callback
 
 # log(sum(exp(x))) that tries to avoid overflow
-from scipy.special import logsumexp
+try:
+    from scipy.special import logsumexp
+except ImportError:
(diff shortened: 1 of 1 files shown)


gensim/models/ldamodel.py
```diff
@@ -48,7 +48,7 @@
 from gensim.models.callbacks import Callback
 
 # log(sum(exp(x))) that tries to avoid overflow
-from scipy.misc import logsumexp
+from scipy.special import logsumexp
```
menshikh-iv: For `1.0.0` it works, but as you see in https://github.com/RaRe-Technologies/gensim/blob/develop/setup.py#L290 our minimal version is `0.18.1`, in this version, this import will be failed. 

Import **must** work for any scipy `>=0.18.1`

dreamgonfly:
I fixed the problem with imports

menshikh-iv:
Thank you @dreamgonfly, congratz with first contribution :1st_place_medal:

uduwage:
This fix probably not available on the latest pip installation? or is it?

menshikh-iv:
@uduwage fix will be available in next release (~first week of December). Alternative way -  you can install `develop` branch from git.

Issue #1676, opened 2017-10-31
Bug in sklearn_api.hdp and in sklearn_api.ldamodel


mmunozm:
#### Description
The new `sklearn_api.hdp` (and also `ldamodel`) modules and/or their combination with `matutils.Sparse2Corpus` yield the wrong results when fitting models from sklearn vectorizers or other sklearn-styled sparse matrices, since they are stored in CSR format. This error might occur in other sklearn_api classes, too.

I believe that either the default value of Sparse2Corpus constructor's `documents_columns` parameter should be changed:

```python
class Sparse2Corpus(object):
    """
    Convert a matrix in scipy.sparse format into a streaming gensim corpus.
    This is the mirror function to `corpus2csc`.
    """

    # Change this to def __init__(self, sparse, documents_columns=False) ?
    def __init__(self, sparse, documents_columns=True):
        if documents_columns:
            self.sparse = sparse.tocsc()
        else:
            self.sparse = sparse.tocsr().T  # make sure shape[1]=number of docs (needed in len())
```

or the following call should include that `documents_columns=False`:

```python
# Same happens in lda model
class HdpTransformer(TransformerMixin, BaseEstimator):
    ...
    def fit(self, X, y=None):
        """
        Fit the model according to the given training data.
        Calls gensim.models.HdpModel
        """
        if sparse.issparse(X):
            corpus = matutils.Sparse2Corpus(X) # Change to matutils.Sparse2Corpus(X, False) ?
        ...
```

#### Steps/Code/Corpus to Reproduce

Example: The number of processed documents corresponds to the number of features.

```python
from sklearn.datasets import fetch_20newsgroups
from sklearn.feature_extraction.text import CountVectorizer
from gensim.sklearn_api.hdp import HdpTransformer

# Small dataset configuration
ngs = fetch_20newsgroups()
samples = ngs.data[:100]

# Simple count vectorization
vectorizer = CountVectorizer()
# x is a sparse matrix
x = vectorizer.fit_transform(samples)
print("%d documents, %d features" % x.shape) # 100 documents, 6547 features

inv_vocab = {v: k for k, v in vectorizer.vocabulary_.items()}

# Train a HDP
hdp_transformer = HdpTransformer(inv_vocab)
hdp_transformer.fit(x)

# Should be 100 but got 6547
print("Processed documents %d" % hdp_transformer.gensim_model.m_num_docs_processed) # 6547
```

#### Expected Results
[...]

PR #1704, opened 2017-11-09
Update 'documents_columns' param passed to 'Sparse2Corpus'

chinmayapancholi13:
Fixes issue https://github.com/RaRe-Technologies/gensim/issues/1676.
cc: @mmunozm @menshikh-iv

### DIFF
--- a/gensim/sklearn_api/ldamodel.py
+++ b/gensim/sklearn_api/ldamodel.py
@@ -57,7 +57,7 @@ class LdaTransformer(TransformerMixin, BaseEstimator):
         Calls gensim.models.LdaModel
         """
         if sparse.issparse(X):
-            corpus = matutils.Sparse2Corpus(X, False)
+            corpus = matutils.Sparse2Corpus(sparse=X, documents_columns=False)
         else:
             corpus = X
 
@@ -101,7 +101,7 @@ class LdaTransformer(TransformerMixin, BaseEstimator):
 
         """
         if sparse.issparse(X):
-            X = matutils.Sparse2Corpus(X)
+            X = matutils.Sparse2Corpus(sparse=X, documents_columns=False)
 
         if self.gensim_model is None:
             self.gensim_model = models.LdaModel(
--- a/gensim/sklearn_api/ldamodel.py
+++ b/gensim/sklearn_api/ldamodel.py
@@ -57,7 +57,7 @@ class LdaTransformer(TransformerMixin, BaseEstimator):
         Calls gensim.models.LdaModel
         """
         if sparse.issparse(X):
-            corpus = matutils.Sparse2Corpus(X)
+            corpus = matutils.Sparse2Corpus(X, False)
         else:
             corpus = X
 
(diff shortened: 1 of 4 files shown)


menshikh-iv:
I'm glad to see you @chinmayapancholi13, have you returned to us :smiley: ?

gensim/sklearn_api/ldamodel.py
```diff
@@ -57,7 +57,7 @@ def fit(self, X, y=None):
         Calls gensim.models.LdaModel
         """
         if sparse.issparse(X):
-            corpus = matutils.Sparse2Corpus(X)
+            corpus = matutils.Sparse2Corpus(X, False)
```
piskvorky: Named parameter better, for clarity. (here and elsewhere)

gensim/test/test_sklearn_api.py
```diff
@@ -177,7 +177,7 @@ def testCSRMatrixConversion(self):
         newmodel.fit(sarr)
         bow = [(0, 1), (1, 2), (2, 0)]
         transformed_vec = newmodel.transform(bow)
-        expected_vec = numpy.array([0.35367903, 0.64632097])
+        expected_vec = numpy.array([0.12843782, 0.87156218])
```
piskvorky: Why this change?

menshikh-iv: I think because now this script interprets the input differently -> different model -> different result, but looks suspicious, I agree.

chinmayapancholi13: @piskvorky As Ivan correctly pointed out, since the input matrix used for training (by `fit` method) is being interpreted differently after making this change, the model changes and thus, the output of `transform` changes as well.

gensim/sklearn_api/ldamodel.py
chinmayapancholi13: Thanks! Done now.

chinmayapancholi13:
@menshikh-iv Right now, among the unit-tests, we have a function `testCSRMatrixConversion()` only for LDA model (and not for HDP and LSI models). So should I add tests with CSR matrices for these models as well?

And should we add separate unit-tests to ensure that input matrix is being inpterpreted in the right manner or just having a function like the current version of `testCSRMatrixConversion()` would be fine? I mean that currently `testCSRMatrixConversion()` function compares the output of `transform()` on the trained model (with a fixed expected output vector) after calling `fit()`. So we are inherently verifying that the input matrix is being interpreted in the right manner.

Also, I'm sorry for the delayed response. I am still a little busy in some university projects so it might take a few more weeks to resume the work here. :smile:

menshikh-iv:
I dont think that additional tests is needed here (all changes exactly similar), LGTM for me, @piskvorky wdyt?

piskvorky:
Up to you. Just checking: no default gensim behaviour was changed here, right? Only the wrapper.

menshikh-iv:
@piskvorky you are correct (only wrapper).
@chinmayapancholi13 thanks for fix :+1:

Issue #1453, opened 2017-06-27
Data/Model storage


menshikh-iv:
We want to store **trained models** and popular **dataset** (in raw/preprocessed format). Also, we want to develop a simple API for accessing this data.

This project makes our users a bit happier.

Plan:
---
1. Check how another frameworks deal with sharing data (sklearn, spacy, nltk, etc). Describe the advantages and disadvantages.
2. Look at Google storage offer for open-source projects (or other offers for an open-source project). We want to host data without payments (because traffic might be substantial).
3. Implement the API for downloading these external, potentially large datasets (based on research from 1.)
4. Create popular models and datasets and upload them to the storage
5. Write a clear, beginner-friendly tutorial (Jupyter notebook) on how to use this functionality.

macks22:
Issues #717 and #746 are closely related to this.

souravsingh:
@menshikh-iv sklearn stores smaller datasets and models in a separate folder and also provides a fetcher for datasets which could be large or require preprocessing. The datasets can be downloading by importing the necessary dataset namespace from ``sklearn.datasets``

NTLK provides a downloader which can be imported to download all the datasets available.

For Storing the datasets, we can store them in the repo if they aren't large, or write a downloader script which can do the job.

menshikh-iv:
@souravsingh thanks for info, let's wait detailed comparison from @chaitaliSaini

chaitaliSaini:
NLTK : It provides downloader with several interfaces(interactie installer and installation via command line) which can be used to download corpora, models, and other data packages that can be used with NLTK.(https://github.com/nltk/nltk/blob/develop/nltk/downloader.py)

 sklearn : sklearn comes with a few small standard datasets that do not require to download any file from some external website. They have store other datasets on mldata.org and sklearn.datasets package is able to directly download data sets from the repository.
 (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/datasets/mldata.py)

 spacy :  It allows models to be downloaded and loaded manually, or using spaCy's download and link commands.
 (https://github.com/explosion/spaCy/blob/master/spacy/cli/download.py)

 For Storage :
 Google Storage : They do not provide free cloud services to open source projects and if we want to avail free cloud services, then they have 2 options. 
 1.we can use google cloud free for one year.
 2.we can always use google cloud for free but with some restriction in number of monthly queries,etc.
 (https://cloud.google.com/free/)

[...]

PR #1705, opened 2017-11-10
[MRG] Data/model storage. Fix 1453

menshikh-iv:
Based on #1453.
What's done:
- Docstrings (fixed mistakes, added examples, etc)
- Removed `--catalogue` (instead of this used `--info`, similar to python API)
- Renaming / style fixes
- Remove unpacking (now we save much HDD space)


gensim/downloader.py
```diff
@@ -0,0 +1,437 @@
...
+    corpora = information['corpora']
+    models = information['models']
+    if name in corpora:
+        return information['corpora'][name]["file_name"]
+    elif name in models:
+        return information['models'][name]["file_name"]
+
+
+def load(name, return_path=False):
+    """Download (if needed) dataset/model and load it to memory (managed by return_path).
```
piskvorky: "managed by return_path" doesn't sound correct. Do you mean "unless return_path is set"?

gensim/downloader.py
```diff
@@ -0,0 +1,437 @@
...
+    Download only example
+    >>> import gensim.downloader as api
+    >>>
+    >>> print(api.load("wiki-en", return_path=True))  # output: /home/user/gensim-data/wiki-en/wiki-en.gz
+
+    """
+    _create_base_dir()
+    file_name = _get_filename(name)
+    if file_name is None:
+        raise Exception(
```
piskvorky: Exception too generic. This looks like a `ValueError`.

gensim/downloader.py
```diff
@@ -0,0 +1,437 @@
...
+            "Incorrect model/corpus name. Choose the model/corpus from the list "
+            "\n {}".format(json.dumps(info(), indent=4))
+        )
+    folder_dir = os.path.join(base_dir, name)
+    path = os.path.join(folder_dir, file_name)
+    if not os.path.exists(folder_dir):
+        _download(name)
```
piskvorky: Any way to "force" a download? (fix previously broken / partial / deleted downloads etc)

gensim/downloader.py
```diff
@@ -0,0 +1,437 @@
...
+
+    if return_path:
+        return path
+    else:
+        sys.path.insert(0, base_dir)
+        module = __import__(name)
+        return module.load_data()
```
piskvorky: Where do I find this `load_data` function? For example for Wikipedia, and for text8?

I got a bit lost in the import magic, and there are no comments.

gensim/downloader.py
menshikh-iv: This situation doesn't happen (because data moved from `/tmp/..` to `~/gensim-data` **after** downloading needed files + checking md5). For this reason, we don't add this key (if something happens - run `load` again.

gensim/downloader.py
menshikh-iv: Yes, that's true.

gensim/downloader.py
menshikh-iv: [text8](https://github.com/RaRe-Technologies/gensim-data/releases/tag/text8) and [wiki-en](https://github.com/RaRe-Technologies/gensim-data/releases/tag/wiki-en)

menshikh-iv:
Need to add the last feature: 
- [ ] possibility to use another `list.json`, it's very needed for me for very accurate adding the new things to `gensim-data` (for now, I must commit to master, it's OK now, but its horribly after release, it can produce "race-condition"). 

The process must look like this: 
- add the model to release page
- create PR with an updated `list.json` (new name, hashes and so on).
- check that all download/load correctly
- merge PR -> new model is available for all users.

PR #1707, opened 2017-11-10
[MRG] Improve segment_wiki script

piskvorky:
This PR cleans up the documentation of the `segment_wiki.py` script, continuing from #1694.

I also added a new `-w` option to the script, letting users control the number of workers.

I also fixed some broken formatting in the [release notes](https://github.com/RaRe-Technologies/gensim/releases/tag/3.1.0) in the section on segment_wiki.

### DIFF
--- a/gensim/scripts/segment_wiki.py
+++ b/gensim/scripts/segment_wiki.py
@@ -19,10 +19,10 @@ Examples
 
   python -m gensim.scripts.segment_wiki -f enwiki-latest-pages-articles.xml.bz2 -o enwiki-latest.json.gz
 
-Processing the entire English Wikipedia dump (13.5 GB, https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2) \
-takes about 2.5 hours (about 3 million articles per hour, on i7-6700HQ, SSD).
+Processing the entire English Wikipedia dump (14 GB, https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2) \
+takes 2 hours (about 2.5 million articles per hour, on 8 core Intel Xeon E3-1275@3.60GHz).
 
-You can then read the created output with:
+You can then read the created output (~6.1 GB gzipped) with:
 
 >>> # iterate over the plain text file we just created
 >>> for line in smart_open('enwiki-latest.json.gz'):
--- a/gensim/scripts/segment_wiki.py
+++ b/gensim/scripts/segment_wiki.py
@@ -5,30 +5,25 @@
 # Copyright (C) 2016 RaRe Technologies
 
 """
-Construct a corpus from a Wikipedia (or other MediaWiki-based) database dump (typical filename
-is <LANG>wiki-<YYYYMMDD>-pages-articles.xml.bz2 or <LANG>wiki-latest-pages-articles.xml.bz2),
-extract titles, section names, section content and save to json-line format,
-that contains 3 fields ::
+CLI script for processing a raw Wikipedia dump (the xml.bz2 format provided by MediaWiki).
 
-    'title' (str) - title of article,
-    'section_titles' (list) - list of titles of sections,
-    'section_texts' (list) - list of content from sections.
-
-English Wikipedia dump available
-`here <https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2>`_. Approximate time
-for processing is 2.5 hours (i7-6700HQ, SSD).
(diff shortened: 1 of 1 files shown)


gensim/scripts/segment_wiki.py
```diff
@@ -54,7 +51,7 @@
 logger = logging.getLogger(__name__)
 
 
-def segment_all_articles(file_path, min_article_character=200):
+def segment_all_articles(file_path, min_article_character=200, workers=None):
```
menshikh-iv: Why `workers=None` by default? If default is `number of cores - 1`, this should be `workers=multiprocessing.cpu_count() - 1` or something equal (same for all `workers=...`)

gensim/scripts/segment_wiki.py
```diff
@@ -260,34 +267,53 @@ def get_texts_with_sections(self):
...
+                    # filter stubs (incomplete, very short articles)
+                    skipped_length += 1
                     continue
 
-                articles += 1
+                total_articles += 1
+                total_sections += len(sections)
                 yield (article_title, sections)
+        logger.info(
+            "finished processing %i articles with %i sections (skipped %i redirects, %i stubs, %i ignored namespaces)",
```
menshikh-iv: Why is 'sections' needed here, what is the good of it?

gensim/scripts/segment_wiki.py
```diff
@@ -260,34 +267,53 @@ def get_texts_with_sections(self):
...
     logging.basicConfig(format='%(asctime)s : %(processName)s : %(levelname)s : %(message)s', level=logging.INFO)
     logger.info("running %s", " ".join(sys.argv))
 
     parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter, description=globals()['__doc__'])
+    default_workers = max(1, multiprocessing.cpu_count() - 1)
     parser.add_argument('-f', '--file', help='Path to MediaWiki database dump', required=True)
     parser.add_argument('-o', '--output', help='Path to output file (stdout if not specified)')
     parser.add_argument(
+        '-w', '--workers',
+        help='Number of parallel workers for multi-core systems (default: %i)' % default_workers,
```
menshikh-iv: `'(default: %i)' % default workers` -> `(default: %(default)s)`

gensim/scripts/segment_wiki.py
piskvorky: Sanity checking.

The rule of thumb for logging is: always display statistics and concrete examples, to make debugging and "sanity eyeballing" easier for humans.

gensim/scripts/segment_wiki.py
piskvorky: What, why? That doesn't sound right, it changes the semantics.

gensim/scripts/segment_wiki.py
piskvorky: It is (see the docs).

gensim/scripts/segment_wiki.py
menshikh-iv: Yes, it's automatic substitution `default` value, [doc](https://docs.python.org/3/library/argparse.html#help)

gensim/scripts/segment_wiki.py
```diff
@@ -5,38 +5,35 @@
 # Copyright (C) 2016 RaRe Technologies
 
 """
-Construct a corpus from a Wikipedia (or other MediaWiki-based) database dump (typical filename
-is <LANG>wiki-<YYYYMMDD>-pages-articles.xml.bz2 or <LANG>wiki-latest-pages-articles.xml.bz2),
```
menshikh-iv: Please return concrete "patterns" for filenames, Wikipedia provides many archives, users very often use the incorrect archive and don't understand "why this doesn't work".

gensim/scripts/segment_wiki.py
menshikh-iv: Why not substitute the real default value (instead of `None), when you have it? Why needed `None` here?

piskvorky: I find it more explicit -- a person viewing the signature would see a number, and perhaps not realize this number is "variable" (will be different on a different system).

In other words, I find it a little "less surprising" when only constants are in default parameters. Anything dynamic/mutable is handled by `None` instead.

gensim/scripts/segment_wiki.py
piskvorky: Cool, I didn't know about that. Changed.

gensim/scripts/segment_wiki.py
piskvorky: Done.

Issue #1634, opened 2017-10-18
TextCorpus doesn't provide a way to convert document text to indices as needed for say DL NLP models


roopalgarg:
#### Description
TextCorpus doesn't provide a way to convert text in a document to indices per the dictionary as needed for say Deep Learning NLP models. TextCorpus uses Dictionary objects `doc2bow` function which is great for most ML models but for DL models where we need sequential indices for text its not usable in most cases.
#### Steps/Code/Corpus to Reproduce

sample.txt
```
hello how are you ?
i am good
```

code:
```
from gensim.corpora.textcorpus import TextCorpus

some_file_name = "sample.txt"
some_dictionary = {
    '<UNK>': 0,
    'how': 1,
    'hello': 2,
    'hi': 3,
    'are': 4,
    'you': 5,
    '?': 6,
    'good': 7
}

gensim_dictionary = Dictionary()
gensim_dictionary.token2id = some_dictionary

txt_corpus = TextCorpus(input=some_file_name, dictionary=gensim_dictionary, token_filters=[])

for text in txt_corpus:
    print list(text)
```

#### Expected Results
Some way to simply convert the corpus to indices per the token2id dict object in Dictionary class, also adding in option to provide an unknown token id which replaces all unknown tokens.
So either adding a `doc2idx()` in `TextCorpus` or integrating that in `Dictionary` class along with `doc2bow()`
[2, 1, 4, 5, 6]
[0, 0, 7]

#### Actual Results
[(1, 1), (2, 1), (4, 1), (5, 1), (6, 1)]
[(7, 1)]

#### Versions

Darwin-16.4.0-x86_64-i386-64bit
('Python', '2.7.12 |Anaconda custom (x86_64)| (default, Jul  2 2016, 17:43:17) \n[GCC 4.2.1 (Based on Apple Inc. build 5658) (LLVM build 2336.11.00)]')
('NumPy', '1.13.1')
('SciPy', '0.19.1')

menshikh-iv:
Hi @roopalgarg, `doc2bow` returns data in BagOfWords format (**unordered**), this works fine for many models from gensim. Also, `doc2bow` returns frequency for each word in the document (second element in the tuple).

Add method `doc2idx` may be a good idea, wdyt @piskvorky @gojomo?

roopalgarg:
[...]

PR #1720, opened 2017-11-16
Add functionality in TextCorpus to convert document text to index vectors

roopalgarg:
TextCorpus doesn't provide a way to convert document text to index vector as needed for say DL NLP models.
Adding a 'doc2idx' to Dictionary object and creating modes in TextCorpus to leverage this functionality.
Referencing issue #1634

### DIFF
--- a/gensim/corpora/dictionary.py
+++ b/gensim/corpora/dictionary.py
@@ -176,8 +176,7 @@ class Dictionary(utils.SaveLoad, Mapping):
     def doc2idx(self, document, unknown_word_index=-1):
         """Convert `document` (a list of words) into a list of indexes = list of `token_id`.
 
-        Each word is assumed to be a **tokenized and normalized** string
-        (either unicode or utf8-encoded).
+        Each word is assumed to be a **tokenized and normalized** string (either unicode or utf8-encoded).
         No further preprocessing is done on the words in `document`; apply tokenization, stemming etc. before calling
         this method.
 
@@ -186,19 +185,26 @@ class Dictionary(utils.SaveLoad, Mapping):
 
         Notes
         -----
-            This function is `const`, aka read-only
+        This function is `const`, aka read-only
 
         Parameters
         ----------
-        document : list
-            List of words tokenized, normalized and preprocessed.
+        document : list of str
+            Tokenized, normalized and preprocessed words
         unknown_word_index : int, optional
             Index to use for words not in the dictionary.
 
         Returns
         -------
-        list
-            List of indexes in the dictionary for words in the `document`
+        list of int
+            Indexes in the dictionary for words in the `document` preserving the order of words
+
+        Examples
+        --------
+        >>> dictionary_obj = Dictionary()
+        >>> dictionary_obj.token2id = {'computer': 0, 'human': 1, 'response': 2, 'survey': 3}
+        >>> dictionary_obj.doc2idx(document=['human', 'computer', 'interface'], unknown_word_index=-1)
+        [1, 0, -1]
 
         """
         if isinstance(document, string_types):
--- a/gensim/corpora/dictionary.py
+++ b/gensim/corpora/dictionary.py
@@ -173,6 +173,36 @@ class Dictionary(utils.SaveLoad, Mapping):
         else:
             return result
 
+    def doc2idx(self, document, unk_wrd_idx=0):
+        """
+        Convert `document` (a list of words) into a list of indexes = list
+        of `token_id`. Each word is assumed to be a **tokenized and normalized** string (either unicode or utf8-encoded)
+        No further preprocessing is done on the words in `document`; apply tokenization, stemming etc. before calling
+        this method.
+
+        Replace all unknown ie, words not in the dictionary presently with the index as set via `unk_wrd_idx`, defaults
+        to 0.
+
+        This function is `const`, aka read-only
+        """
+        if isinstance(document, string_types): 
+            raise TypeError("doc2idx expects an array of unicode tokens on input, not a single string") 
+ 
+        token2id = self.token2id
+
+        list_word_idx = list() 
+        for word in document: 
+            word = word if isinstance(word, unicode) else unicode(word, 'utf-8')
+ 
+            wrd_idx = token2id.get(word, None) 
+ 
+            if wrd_idx is not None: 
+                list_word_idx.append(wrd_idx)
+            else:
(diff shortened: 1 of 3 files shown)


roopalgarg:
@menshikh-iv how does this look?

gensim/corpora/dictionary.py
```diff
@@ -173,6 +173,37 @@ def doc2bow(self, document, allow_update=False, return_missing=False):
         else:
             return result
 
+    def doc2idx(self, document, unk_wrd_idx=0):
```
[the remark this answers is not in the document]
menshikh-iv: Also please rename `unk_wrd_idx` to `unknown_word_index` (here and everywhere)

gensim/corpora/textcorpus.py
```diff
@@ -112,7 +114,10 @@ class TextCorpus(interfaces.CorpusABC):
     6.  remove stopwords; see `gensim.parsing.preprocessing` for the list of stopwords
 
     """
-    def __init__(self, input=None, dictionary=None, metadata=False, character_filters=None, tokenizer=None, token_filters=None):
+    def __init__(
```
menshikh-iv: Please use vertical indent (**only** for method/function definition, in all other cases - hanging indent).

gensim/corpora/dictionary.py
```diff
@@ -173,6 +173,37 @@ def doc2bow(self, document, allow_update=False, return_missing=False):
...
+
+        Replace all unknown words i.e, words not in the dictionary with the index as set via `unk_wrd_idx`, defaults to
+        0.
+
+        This function is `const`, aka read-only
+        """
+        if isinstance(document, string_types):
+            raise TypeError("doc2idx expects an array of unicode tokens on input, not a single string")
+
+        token2id = self.token2id
```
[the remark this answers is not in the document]
roopalgarg: was just following convention from 'doc2bow' but I will make the change as you pointed out.

roopalgarg:
@menshikh-iv the idea was that since we are adding the functionality to the `__iter__` of `TextCorpus`, we can process a corpus to get vocabulary indexes of the corpus. Also since `WikiCorpus` inherits `TextCorpus` and doesnt override the `__iter__` the functionality gets added to `WikiCorpus` as well.
Though I agree that the constructor for `WikiCorpus` doesnt explicitly take parameters for this new functionality, but we can set the parameters after creating an instance of the class by explicitly setting the new parameters.

So couple of things here then:

1. Should `WikiCorpus` have the parameters explicitly as well as part of its constructor definition?
2. Where else do you think this functionality should be added since you were asking as to why only `TextCorpus`?

menshikh-iv:
@roopalgarg current problem is more "global", let me describe:

- We have many different classes for corpuses, we need to add it for all corpuses (but I don't think that this is a good idea, but have same interfaces will be nice).
- How to choose the concrete class for support this Dictionary feature (maybe we no need to add it to any corpuses, only new method for Dictionary?)
- Passing arguments (if we want to add it in many places)

I think corpus classes needs global refactoring (bring everything to the same interfaces and simplify, i.e. a minimum of functionality), @roopalgarg it isn't your problem, sorry, but you reminded me about the old and important problem.

@roopalgarg I'm ready to merge only new method for Dictionary right now.

@piskvorky wdyt? corpuses in common are very chaotic, have any idea how to rework it?

roopalgarg:
@menshikh-iv I see your point. For now just adding `doc2idx` to the `Dictionary` class sounds good to me.
Do I need to revert my changes to the other 2 files (or there is a way you can discard the changes to them)?

gensim/corpora/dictionary.py
```diff
@@ -173,6 +173,40 @@ def doc2bow(self, document, allow_update=False, return_missing=False):
...
+        (either unicode or utf8-encoded).
+        No further preprocessing is done on the words in `document`; apply tokenization, stemming etc. before calling
+        this method.
+
+        Replace all unknown words i.e, words not in the dictionary with the index as set via `unknown_word_index`,
+        defaults to -1.
+
+        Notes
+        -----
+            This function is `const`, aka read-only
```
menshikh-iv: No needed indentation here

gensim/corpora/dictionary.py
```diff
@@ -173,6 +173,40 @@ def doc2bow(self, document, allow_update=False, return_missing=False):
...
+
+        Parameters
+        ----------
+        document : list
```
menshikh-iv: `list of str`

gensim/corpora/dictionary.py
```diff
@@ -173,6 +173,40 @@ def doc2bow(self, document, allow_update=False, return_missing=False):
...
+            List of words tokenized, normalized and preprocessed.
```
menshikh-iv: No need to mention type twice

gensim/corpora/dictionary.py
```diff
@@ -173,6 +173,40 @@ def doc2bow(self, document, allow_update=False, return_missing=False):
...
+        unknown_word_index : int, optional
+            Index to use for words not in the dictionary.
+
+        Returns
+        -------
+        list
```
menshikh-iv: `list of int`

gensim/corpora/dictionary.py
```diff
@@ -173,6 +173,40 @@ def doc2bow(self, document, allow_update=False, return_missing=False):
...
+            List of indexes in the dictionary for words in the `document`
```
menshikh-iv: No need to mention type twice + add `preserves order`.

gensim/corpora/dictionary.py
```diff
@@ -173,6 +173,40 @@ def doc2bow(self, document, allow_update=False, return_missing=False):
...
+
```
menshikh-iv: Please add example section (simple example that works how to apply this method)

menshikh-iv:
@roopalgarg yeah, please revert 2 files, fix docstring and that's all :+1:

roopalgarg:
@menshikh-iv a little new to numpy style docstrings so not fully aware of best practices. learnt something new today :)

roopalgarg:
@menshikh-iv good to go ?

roopalgarg:
@menshikh-iv good to merge ?

menshikh-iv:
@roopalgarg yeah, thanks for your contribution:+1:

roopalgarg:
@menshikh-iv awesome! thanks

Issue #1718, opened 2017-11-15
Bug in bm25 implementation.


se4u:
In line 52 of bm25.py the variable `self.corpus_size` should be replaced by `len(document)`.

https://github.com/RaRe-Technologies/gensim/blob/82c394a9085d583e8a75c2bb32ecd37cf61236f0/gensim/summarization/bm25.py#L52

menshikh-iv:
Nice catch @se4u, thanks for the report!

souravsingh:
@se4u Can you tell us how the bug affects the execution of the summarization module? A small code sample reproducing the bug would be nice.

se4u:
@Sourav : Actually, I used the bm25 module independently of the
summarization module just for IR.

Basically, I wanted to quickly compare bm25 with tf-idf, pylucene's api
seemed to be too cumbersome, and I did not want to reinvent the wheel.
Gensim generally has pretty good coding standards that's why I decided to
use bm25 from gensim instead of the various other BM25 implementations on
github. But after implementation I noticed that the results from bm25 were
worse than standard tf-idf and that's why I looked at the code in more
detail. The data that I am working on comes from [https://www.ldc.upenn.edu
](LDC) so its hard to share it.

PR #1726, opened 2017-11-18
Fix bug in bm25.py

souravsingh:
Related to #1718

### DIFF
--- a/gensim/summarization/bm25.py
+++ b/gensim/summarization/bm25.py
@@ -49,7 +49,7 @@ class BM25(object):
                 continue
             idf = self.idf[word] if self.idf[word] >= 0 else EPSILON * average_idf
             score += (idf * self.f[index][word] * (PARAM_K1 + 1)
-                      / (self.f[index][word] + PARAM_K1 * (1 - PARAM_B + PARAM_B * self.corpus_size / self.avgdl)))
+                      / (self.f[index][word] + PARAM_K1 * (1 - PARAM_B + PARAM_B * len(document) / self.avgdl)))
         return score
 
     def get_scores(self, document, average_idf):


piskvorky:
That looks like a serious / critical bug, thanks!

What functionality affected? How did this bug manifest itself in practice?

souravsingh:
Nothing is mentioned in the issue itself. The person filing the issue just mentions that it is a bug, and didn't say anything about how it affects the functionality.

piskvorky:
Probably best to double check and verify.

souravsingh:
The person who filed the issue mentioned the following- 

> Actually, I used the bm25 module independently of the
summarization module just for IR.

> Basically, I wanted to quickly compare bm25 with tf-idf, pylucene's api
seemed to be too cumbersome, and I did not want to reinvent the wheel.
Gensim generally has pretty good coding standards that's why I decided to
use bm25 from gensim instead of the various other BM25 implementations on
github. But after implementation I noticed that the results from bm25 were
worse than standard tf-idf and that's why I looked at the code in more
detail. The data that I am working on comes from [https://www.ldc.upenn.edu
](LDC) so its hard to share it.

piskvorky:
Yeah, I saw.

It's better link to @se4u's [answer](https://github.com/RaRe-Technologies/gensim/issues/1718#issuecomment-345463764), instead of copy-pasting.

souravsingh:
So we confirm the behavior of the bug by running both tf-idf and bm25 correct? It might be a bit difficult since bm25 doesn't have any docstrings to help in understanding the behavior.

menshikh-iv:
BM25 formula:
image
Current variant (in code) looks correct.

eliotpbrenner:
Can you please double check?  I believe it should be `len(self.corpus[index])`, not `len(document)`
Thanks!

se4u:
@eliotpbrenner yes you are right !

menshikh-iv:
@eliotpbrenner https://github.com/RaRe-Technologies/gensim/commit/24a692b2f1160ff699f65ac26fc840d115d17155#commitcomment-26708993

PR #1727, opened 2017-11-20
Typos

youqad:


menshikh-iv:
@youqad congratz with the first contribution :fire:!

PR #1733, opened 2017-11-21
Fix os.rename from gensim.downloader when src and dst on different partitions

anotherbugmaster:
Tests are working on Arch Linux now. There was a problem with new `downloader.py` module and tmpfs partition.

### DIFF
--- a/gensim/corpora/__init__.py
+++ b/gensim/corpora/__init__.py
@@ -1,5 +1,5 @@
 """
-This package contains implementations of various streaming corpus I/O format.
+This package contains implementations of various streaming corpus I/O formats.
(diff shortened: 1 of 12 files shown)


menshikh-iv:
Nice catch @anotherbugmaster :+1: 
- remove all unrelated changes (and fix your `develop` branch)
- need more complicated logic for multi partition case (shutil.move to needed partitions and after - os.rename to needed dir)

anotherbugmaster:
@menshikh-iv, all done. :)

menshikh-iv:
Great job @anotherbugmaster :+1:

PR #1734, opened 2017-11-22
[MRG] Poincare l2 regularization

jayantj:
Adds l2-regularization for Poincare embeddings, which improves results and results in models much more similar to the models visualized in the [original paper](https://arxiv.org/pdf/1705.08039.pdf). To be merged post #1700 

Without regularization, the model looks like this - 
mammals_200_ep

The problem with this is that most of the nodes are too close to the boundary. In order to prevent this, l2-regularization is added to **only the parent node** in the training hypernymy relations, which results in the following model - 

mammals_350_ep_l2_reg

Also adds cleaner handling of autograd, where the import is not conditional, however if autograd cannot be imported, gradient checking is automatically skipped along with a warning.


gensim/models/poincare.py
```diff
@@ -91,6 +91,8 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
             Number of threads to use for training the model.
         epsilon : float, optional
             Constant used for clipping embeddings below a norm of one.
+        regularization_coeff : float, optional
```
janpom: Can the L2 regularization be disabled? Setting `regularization_coeff = 0.0` should do it, correct? Might be worth mentioning here.

jayantj: That's correct. Good point, adding.

PR #1736, opened 2017-11-22
Add flag for hiding outdated data for gensim.downloader.info

menshikh-iv:
We want to have "immutable" data in [gensim-data](https://github.com/RaRe-Technologies/gensim-data) repository (i.e. `api.load("anyname")` **must** return always exactly same data), but some dataset/models periodically updated and we want to support new versions (for example - wikipedia).

For this reason, I added `show_only_latest` flag to `info` function, this allows showing only latest version of datasets (for example, if we have `wiki-english-20170101` and `wiki-english-20170201` in storage, only `wiki-english-20170201` will be showed). 
Besides, the old version will be showed if `show_only_latest==False` and can be loaded always (by name).

### DIFF
--- a/gensim/downloader.py
+++ b/gensim/downloader.py
@@ -154,14 +154,14 @@ def _calculate_md5_checksum(fname):
     return hash_md5.hexdigest()
 
 
-def info(name=None, actual=True):
+def info(name=None, show_only_latest=True):
     """Provide the information related to model/dataset.
 
     Parameters
     ----------
     name : str, optional
         Name of model/dataset.  If not set - shows all available data.
-    actual : bool, optional
+    show_only_latest : bool, optional
         If storage contains different versions for one data/model, this flag allow to hide outdated versions.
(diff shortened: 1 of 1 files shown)


gensim/downloader.py
```diff
@@ -197,9 +201,15 @@ def info(name=None):
             return information['models'][name]
         else:
             raise ValueError("Incorrect model/corpus name")
-    else:
+
+    if not show_only_latest:
         return information
 
+    return {
+        "corpora": {name: data for (name, data) in information['corpora'].items() if data.get("latest", True)},
```
piskvorky: Should the default be `True`?

menshikh-iv: Yes, It is much easier when the new version is loaded set `latest: False` for the old version, instead of set `latest` flag to all data as the required field.

piskvorky:
LGTM (but fill in the PR description with motivation please).

PR #1739, opened 2017-11-25
Fix PyPI long description rendering

edigaryev:
Currently the rendering of gensim's reStructuredText description on PyPI
is broken[[1]](https://pypi.python.org/pypi/gensim/3.1.0/)[[2]](https://pypi.org/project/gensim/3.1.0/), because some characters aren't escaped properly[[3]](http://docutils.sourceforge.net/docs/user/rst/quickref.html#escaping):

>reStructuredText uses backslashes ("\") to override the special meaning
>given to markup characters and get the literal characters themselves.
>To get a literal backslash, use an escaped backslash ("\\").

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -196,7 +196,7 @@ When `citing gensim in academic papers and theses <https://scholar.google.cz/cit
 
   @inproceedings{rehurek_lrec,
         title = {{Software Framework for Topic Modelling with Large Corpora}},
-        author = {Radim {\v R}eh{\r u}{\v r}ek and Petr Sojka},
+        author = {Radim {\\v R}eh{\\r u}{\\v r}ek and Petr Sojka},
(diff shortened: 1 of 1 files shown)


piskvorky:
Thanks! How can we test this, before releasing?

edigaryev:
>Thanks! How can we test this, before releasing?

https://packaging.python.org/guides/using-testpypi/

If you want to test the legacy PyPI you should probably replace `https://test.pypi.org/` with `https://testpypi.python.org/`.

piskvorky:
This is interesting, thanks. Ping @isamaru @menshikh-iv FYI.

menshikh-iv:
Thanks for information @edigaryev, congratz with first contribution to gensim :fire: :1st_place_medal:

PR #1745, opened 2017-11-29
inline a barebones version of logsumexp for improved performance

arlenk:
logsumexp accounts for 50% of the run time of ldamodel.  Much
of this time is spent by "robustness" checks performed by
scipy's logsumexp (eg, _asarray_validated, checks for NaNs, etc.).

Removing these checks greatly improves the overall performance
of ldamodel.  Eg, run time when fitting a lda model on the
enron dataset (from UCI) decreases from 20-40%.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -52,16 +52,25 @@ logger = logging.getLogger('gensim.models.ldamodel')
 
 
 def logsumexp(x):
+    """Log of sum of exponentials
+
+    Parameters
+    ----------
+    x : array_like
+        Input data
+
+    Returns
+    -------
+    float
+        log of sum of exponentials of elements in `x`
+
+    Notes
+    -----
+        for performance, does not support NaNs or > 1d arrays like 
+        scipy.special.logsumexp()
+
     """
-    barebones log-sum-exp that tries to avoid overflows
 
-    Args:
-        x: 1d ndarray
-
-    Note:
-        does not support NaNs
-
-    """
     x_max = np.max(x)
     x = np.log(np.sum(np.exp(x - x_max)))
     x += x_max
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -66,7 +66,7 @@ def logsumexp(x):
 
     Notes
     -----
-        for performance, does not support NaNs or > 1d arrays like 
+        for performance, does not support NaNs or > 1d arrays like
         scipy.special.logsumexp()
 
     """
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Hi @arlenk, can you attach files from cProfile befor/after change + code that you used for profiling.

gensim/models/ldamodel.py
```diff
@@ -47,16 +47,28 @@
...
-except ImportError:
-    from scipy.misc import logsumexp
-
 
 logger = logging.getLogger('gensim.models.ldamodel')
 
 
+def logsumexp(x):
+    """
+    barebones log-sum-exp that tries to avoid overflows
```
menshikh-iv: Please use [numpy-style docstrings](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)

arlenk:
can't seem to attach a py file, so here's a zip of the profiling code and the cProfile output.  The speed up varies based on the number of passes and other parameters.   On my computer, I see speedups of 10-30% on average.  Curious if that's specific to my configuration or not.

[profiling.zip](https://github.com/RaRe-Technologies/gensim/files/1516475/profiling.zip)

menshikh-iv:
I see only 5% speedup (because this method doesn't consume 50% of time, slowest method is `dirichlet_expectation` around 40% of all time), but that's good optimization too:+1:

piskvorky:
Rewriting the whole LDA inner loop in C (Cython) would be super cool.

It should be also quite easy to do (unless I'm missing something). The NumPy overhead will be massive, will all the checks, dynamic allocations, Python overhead...

arlenk:
> Rewriting the whole LDA inner loop in C (Cython) would be super cool.

I took a stab at moving more of the hotspots into cython in the branch below.  From some profiling, most of the time spent is in the following few spots:

- dirichlet_expectation (specifically the call to psi())
- calculating the mean absolute difference to determine when gamma has converged
- logsumexp (even with the inline version)

I broke these functions out into a cython module.  For now, I just created a fastldamodel.py (a copy of ldamodel.py but with the cython functions above) module to see if there is any interest in pursuing this.  Based on some profiling on my pc (attached), the speed up is about 100% (meaning the new code takes half the time -- though this is dependent on number of topics, number of passes, etc.).

I also noticed that the current ldamodel code uses float32 datatypes, but the user can change this if desired.  If we go down the cython route, we'd have to decide if we want to allow the user to change the datatype as this would require different cython functions for each data type (I believe). Alternatively, we could use the cython code when the datatype is float32 (and perhaps float64) and fall back to the slower versions otherwise.

https://github.com/arlenk/gensim/tree/cython

[profiling_results.zip](https://github.com/RaRe-Technologies/gensim/files/1532983/profiling_results.zip)

piskvorky:
Awesome!

Float32 is the default, that should be our focus (at least to start with). Double precision is not that important.

menshikh-iv:
@arlenk can you create PR with your cython optimization, it will be great!

arlenk:
moved to a separate pull request: 

#1767

PR #1754, opened 2017-12-04
Optimize segment_wiki script

horpto:
In previous scheme the main process and workers worked in parallel but not concurrently. Actually work are executed or worker processes (parsing and search sections) or main process (parsing xml and collecting results) but not at the same time. Oops. I try to fix it via imap and constant strain on the workers.

Actually I've checked on small part of wiki (~400 мб) but processing time decreased from 7 min to 3min 40-50s.

### DIFF
--- a/gensim/scripts/segment_wiki.py
+++ b/gensim/scripts/segment_wiki.py
@@ -46,9 +46,14 @@ import re
 import sys
 from xml.etree import cElementTree
 
+
 from gensim.corpora.wikicorpus import IGNORED_NAMESPACES, WikiCorpus, filter_wiki, get_namespace, utils
 from smart_open import smart_open
 
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
 
 logger = logging.getLogger(__name__)
 
@@ -141,16 +146,18 @@ def extract_page_xmls(f):
         XML strings for page tags.
 
     """
-    elems = (elem for _, elem in cElementTree.iterparse(f, events=("end",)))
+    elems = cElementTree.iterparse(f, events=("end",))
 
-    elem = next(elems)
+    _, elem = next(elems)
     namespace = get_namespace(elem.tag)
     ns_mapping = {"ns": namespace}
     page_tag = "{%(ns)s}page" % ns_mapping
 
-    for elem in elems:
+    for _, elem in elems:
         if elem.tag == page_tag:
-            yield cElementTree.tostring(elem)
+            # Pickle object explicitly to send it to the children
+            # as elem.clear() will be called earlier than implicitly serialization of elem
+            yield pickle.dumps(elem)
             # Prune the element tree, as per
             # http://www.ibm.com/developerworks/xml/library/x-hiperfparse/
             # except that we don't need to prune backlinks from the parent
@@ -175,7 +182,8 @@ def segment(page_xml):
         Structure contains (title, [(section_heading, section_content)]).
 
     """
-    elem = cElementTree.fromstring(page_xml)
+    # Unpickling object should be faster then creation from string (via cElementTree.fromstring(page_xml))
+    elem = pickle.loads(page_xml)
     filter_namespaces = ('0',)
     namespace = get_namespace(elem.tag)
     ns_mapping = {"ns": namespace}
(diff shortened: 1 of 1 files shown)


gensim/scripts/segment_wiki.py
```diff
@@ -141,16 +146,18 @@ def extract_page_xmls(f):
...
     ns_mapping = {"ns": namespace}
     page_tag = "{%(ns)s}page" % ns_mapping
 
-    for elem in elems:
+    for _, elem in elems:
         if elem.tag == page_tag:
-            yield cElementTree.tostring(elem)
+            # Pickle object explicitly to send it to the children
+            # as elem.clear() will be called earlier than implicitly serialization of elem
+            yield pickle.dumps(elem)
```
piskvorky: Does using a higher pickle protocol (not the default ASCII) make sense?

gensim/scripts/segment_wiki.py
```diff
@@ -263,25 +271,22 @@ def get_texts_with_sections(self):
...
-                    continue
-                if sum(len(body.strip()) for (_, body) in sections) < self.min_article_character:
-                    # filter stubs (incomplete, very short articles)
-                    skipped_length += 1
-                    continue
-
-                total_articles += 1
-                total_sections += len(sections)
-                yield (article_title, sections)
+        for article_title, sections in pool.imap_unordered(segment, page_xmls, chunksize=10):
```
piskvorky: This is not a good idea -- if the producer/consumer speed is mismatched, the pool will fill up internal queues and run out of memory (too many tasks in the finished queue / too many tasks in the input queue).

gensim/scripts/segment_wiki.py
horpto: I tried but don't find some meaningful profit.

gensim/scripts/segment_wiki.py
horpto: Actually, the main process is still the busiest as there is a parsing xml and collecting results. Children processes are weakly loaded.

I don't think that article order is really important (but can back to `imap` in other case).

menshikh-iv: @piskvorky as I remember, the order of articles have no sense (no semantic), the only order of sections (in one article) have sense. Let me check.

piskvorky: Yes, that's a side note, I also don't think it's important. The queues filling up without control (unbounded size of input/output queues) is the primary concern here. I don't remember exactly the code exactly, but I have no doubt the chunking was added for a reason.

menshikh-iv: @piskvorky probably `unordered` gives the performance boost, but need to check it.

menshikh-iv:
I can't benchmark it now because I found a problem with py2, please fix this @horpto first

```sh
(wiki) ivan@P50:~/test_wiki$ python -m gensim.scripts.segment_wiki -w 4 -f ruwiki-20171001-pages-articles.xml.bz2 -o test_sasha.jsonl.gz
2017-12-13 20:36:24,291 : MainProcess : INFO : running /home/ivan/release/gensim/gensim/scripts/segment_wiki.py -w 4 -f ruwiki-20171001-pages-articles.xml.bz2 -o test_sasha.jsonl.gz
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/ivan/release/gensim/gensim/scripts/segment_wiki.py", line 324, in <module>
    workers=args.workers
  File "/home/ivan/release/gensim/gensim/scripts/segment_wiki.py", line 123, in segment_and_write_all_articles
    for idx, (article_title, article_sections) in enumerate(article_stream):
  File "/home/ivan/release/gensim/gensim/scripts/segment_wiki.py", line 87, in segment_all_articles
    for article_title, article_sections in wiki_sections_text:
  File "/home/ivan/release/gensim/gensim/scripts/segment_wiki.py", line 274, in get_texts_with_sections
    for article_title, sections in pool.imap_unordered(segment, page_xmls, chunksize=10):
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 289, in <genexpr>
    return (item for chunk in result for item in chunk)
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 673, in next
    raise value
TypeError: expected string or Unicode object, NoneType found
```

menshikh-iv:
Unfortunatelly, this PR doesn't gives tangible speedup, for this reason I close it, sorry @horpto.

piskvorky:
@menshikh-iv how do you know, have you tried? Could you replicate the speedup that @horpto reported above? (7m vs 4m)

menshikh-iv:
@piskvorky we talked with @horpto later and he didn't receive any tangible speedup in the last test.

piskvorky:
Oh :(

Issue #1752, opened 2017-12-03
Incorrect learning of word vectors during online training using `FastText` native implementation


manneshiva:
#### Description
A bug in Fasttext native implementation causes `syn0` to be equal to `syn0_vocab` at the end of training. This causes incorrect learning of vectors during online training.

#### Steps/Code/Corpus to Reproduce

```
from gensim.models.word2vec import LineSentence
from gensim.models.fasttext import FastText as FT_gensim
import os
import gensim

data_dir = '{}'.format(os.sep).join([gensim.__path__[0], 'test', 'test_data']) + os.sep
data_file = '{}lee_background.cor'.format(data_dir)
sentences = LineSentence(data_file)

model = FT_gensim(sg=1, hs=0,window=2, negative=5, iter=1)
model.build_vocab(sentences)
model.train(sentences, total_examples=model.corpus_count, epochs=model.iter)

print (model.wv.syn0 == model.wv.syn0_vocab).all()
```
#### Expected Results
`False`

#### Actual Results
`True`

#### Versions
Linux-4.10.0-40-generic-x86_64-with-Ubuntu-16.04-xenial
('Python', '2.7.12 (default, Nov 19 2016, 06:48:10) \n[GCC 5.4.0 20160609]')
('NumPy', '1.13.3')
('SciPy', '1.0.0')
('gensim', '3.1.0')
('FAST_VERSION', 1)

menshikh-iv:
@manneshiva good description, but if you have information/ideas where is the concrete problem / how to solve it, please add it to your report.

piskvorky:
(off topic: regarding these `'{}'.format` constructions, have a look at `os.path.join`)

PR #1756, opened 2017-12-04
[MRG] Fixes incorrect vectors learned during online training for `FastText` models

manneshiva:
This PR addresses #1752. This bug was caused by incorrect assignment of vectors to `syn0` after training a `FastText` model.

### DIFF
--- a/gensim/test/test_fasttext.py
+++ b/gensim/test/test_fasttext.py
@@ -419,6 +419,8 @@ class TestFastTextModel(unittest.TestCase):
         self.assertTrue(all(['terrorism' not in l for l in others]))
         model.build_vocab(others)
         model.train(others, total_examples=model.corpus_count, epochs=model.iter)
+        # checks that `syn0` is different from `syn0_vocab`
+        self.assertFalse(np.all(np.equal(model.wv.syn0, model.wv.syn0_vocab)))
         self.assertFalse('terrorism' in model.wv.vocab)
         self.assertFalse('orism>' in model.wv.ngrams)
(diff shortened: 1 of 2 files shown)


gensim/test/test_fasttext.py
```diff
@@ -419,6 +419,8 @@ def online_sanity(self, model):
         self.assertTrue(all(['terrorism' not in l for l in others]))
         model.build_vocab(others)
         model.train(others, total_examples=model.corpus_count, epochs=model.iter)
+        # checks that `syn0` is different from `syn0_vocab`
+        self.assertFalse(np.all(np.equal(model.wv.syn0, model.wv.syn0_vocab)))
```
janpom: This is a rather weird check. Nobody really cares if `syn0` or `syn0_vocab` are different. What we do care about is that `syn0_vocab` is incorrect because it gets overwritten by `get_vocab_word_vecs()`. A more meaningful check would be either asserting certain properties of `syn0_vocab` that should hold or checking that `syn0_vocab` is the same before and after calling `get_vocab_word_vecs()`.

piskvorky: What @janpom said + check out `np.allclose` :)

manneshiva: I will add another test case but let us also retain the `np.equal` check for `syn0` and `syn0_vocab`.

menshikh-iv:
Typically, we shouldn't use `np.equal` for any stuff with floating point dtype, but for this case, we should have to **exactly** same matrices, for this reason, I think that `np.equal` is OK.

PR #1758, opened 2017-12-04
Fix phrases load, for backward compatibility

### LINKED ISSUE #1751: Gensim 3.1.0 Phrase loader incompatible with older models due to introduction of common_terms
I have a Phrases model that was computed using Gensim 2.2.0 but because of changes in 3.1.0 (I believe), I cannot load it anymore. This is the error I get:

```
<timed exec> in <module>()

~/Stuff/Sources/anaconda3/envs/nlp/lib/python3.6/site-packages/gensim/models/phrases.py in __init__(self, phrases_model)
    546         self.delimiter = phrases_model.delimiter
    547         self.scoring = phrases_model.scoring
--> 548         self.common_terms = phrases_model.common_terms
    549         corpus = self.pseudocorpus(phrases_model)
    550         self.phrasegrams = {}

AttributeError: 'Phrases' object has no attribute 'common_terms'
```

Cheers!

alexgarel:
This should fix #1751.
Also it gives backward compatible load for Phraser.


alexgarel:
@menshikh-iv it's ready to validate.

gensim/models/phrases.py
```diff
@@ -178,7 +178,61 @@ def analyze_sentence(self, sentence, threshold, common_terms, scorer):
...
+    @classmethod
+    def load(cls, *args, **kwargs):
+        """
+        Load a previously saved Phrases class. Handles backwards compatibility from
+            older Phrases versions which did not support  pluggable scoring functions.
+            Otherwise, relies on utils.load
+        """
+
+        # for python 2 and 3 compatibility. basestring is used to check if model.scoring is a string
+        try:
```
piskvorky: Gensim uses the `six` library to bridge py2/py3.

gensim/models/phrases.py
```diff
@@ -178,7 +178,61 @@ def analyze_sentence(self, sentence, threshold, common_terms, scorer):
...
+                        cls.__name__)
+                    logger.info(
+                        'setting scoring method to npmi_scorer pluggable scoring method ' +
+                        'for compatibility')
+                    model.scoring = npmi_scorer
+                else:
+                    raise ValueError(
+                        'failed to load %s model with unknown scoring setting %s' %
+                        (cls.__name__, model.scoring))
+        # if there is non common_terms attribute, inizialize
```
piskvorky: typo - `initialize`

gensim/models/phrases.py
alexgarel: ok, note that it was from a previous commit…

alexgarel:
> What is the rationale behind introducing the new class (PhrasesTransformation)?
Phrases and Phraser needs the same `load` tweaks to preserve backward compatibility, and I wanted to avoid copy/pasting same code in both class (Phraser backward compatible load was not yet supported, but needs to).

alexgarel:
@piskvorky fixed.

gensim/test/test_phrases.py
```diff
@@ -419,9 +419,78 @@ def testSaveLoadNoScoring(self):
...
-        finally:
-            if os.path.exists("test_phrases_testSaveLoadNoScoring_temp_save.pkl"):
-                os.remove("test_phrases_testSaveLoadNoScoring_temp_save.pkl")
+    def testSaveLoadNoCommonTerms(self):
+        """ Saving and loading a Phrases objects without common_terms
+        This should ensure backwards compatibility with old versions of Phrases"""
+
+        with temporary_file("test.pkl") as fpath:
+            bigram = Phrases(self.sentences, min_count=1, threshold=1)
+            del(bigram.common_terms)
```
menshikh-iv: better solution - checkout to the previous version, tran+save model and try to load old model here (in your current implementation, we possibly skip other bugs, that can be hidden due to current bug)

gensim/models/phrases.py
```diff
@@ -178,7 +178,55 @@ def analyze_sentence(self, sentence, threshold, common_terms, scorer):
...
+        # if no scoring parameter, use default scoring
+        if not hasattr(model, 'scoring'):
+            logger.info('older version of %s loaded without scoring function', cls.__name__)
+            logger.info('setting pluggable scoring method to original_scorer for compatibility')
+            model.scoring = original_scorer
+        # if there is a scoring parameter, and it's a text value, load the proper scoring function
+        if hasattr(model, 'scoring'):
+            if isinstance(model.scoring, six.string_types):
+                if model.scoring == 'default':
+                    logger.info(
```
menshikh-iv: nitpick: we use 140 char limits for code, no need to split this call into several lines

gensim/test/test_phrases.py
```diff
@@ -8,17 +8,29 @@
...
 from gensim import utils
-from gensim.models.phrases import SentenceAnalyzer, Phrases, Phraser, pseudocorpus
+from gensim.models.phrases import SentenceAnalyzer, Phrases, Phraser
+from gensim.models.phrases import pseudocorpus, original_scorer
 from gensim.test.utils import common_texts
 
-if sys.version_info[0] >= 3:
-    unicode = str
+
+@contextlib.contextmanager
```
menshikh-iv: This looks useful, please move it to `gensim.test.utils` https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/utils.py

gensim/models/phrases.py
alexgarel: argh travis failed, it was 120 not 140 @menshikh-iv, you joker ! ;-)

alexgarel:
@menshikh-iv this should be ok.

Your proposal to test load backward compatibility was right, for I extend it to scoring and found there was a bug in it.

menshikh-iv:
Big thanks for fast fix @alexgarel, I checked with 2.0.0, 1.0.1 - works fine!

gensim/models/phrases.py
menshikh-iv: @alexgarel oh, sorry, my mistake (120 instead of 80, not 140)

PR #1761, opened 2017-12-05
Fix dtype of `matutils.unitvec`. Fix #1722

### LINKED ISSUE #1722: matutils.unitvec returns vector with different dtype from input vector
#### Description
```python
from gensim import matutils
import numpy as np

print(input_vector.dtype)
>>> float32

print(unit_vector.dtype)
>>> float64
```

accraze:
`matutils.unitvec` currently returns a unitvector of a different dtype from the input vector if the input dtype isn't `np.float`

we should make the return type consistent with the input type.

fixes #1722

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -420,13 +420,13 @@ def unitvec(vec, norm='l2'):
             return vec
 
     if isinstance(vec, np.ndarray):
-        vec = np.asarray(vec, dtype=float)
+        vec = np.asarray(vec, dtype=vec.dtype)
         if norm == 'l1':
             veclen = np.sum(np.abs(vec))
         if norm == 'l2':
             veclen = blas_nrm2(vec)
         if veclen > 0.0:
-            return blas_scal(1.0 / veclen, vec)
+            return blas_scal(1.0 / veclen, vec).astype(vec.dtype)
         else:
             return vec
 
--- a/gensim/test/test_matutils.py
+++ b/gensim/test/test_matutils.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+
+"""
+Automated tests for checking various matutils functions.
+"""
+
+import logging
+import unittest
+
+import numpy as np
+
+from gensim import matutils
+
+
+class TestMatutils(unittest.TestCase):
+    def test_unitvec(self):
+        input_vector = np.random.uniform(size=(100,)).astype(np.float32)
+        unit_vector = matutils.unitvec(input_vector)
+        self.assertEqual(input_vector.dtype, unit_vector.dtype)
+        
+
+if __name__ == '__main__':
+    logging.root.setLevel(logging.WARNING)
+    unittest.main()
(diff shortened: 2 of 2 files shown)


menshikh-iv:
Hi @accraze, please add tests for this fix (you can use example from #1722 as base for your test), also, check case with sparse too.

gensim/matutils.py
```diff
@@ -426,7 +426,7 @@ def unitvec(vec, norm='l2'):
         if norm == 'l2':
             veclen = blas_nrm2(vec)
         if veclen > 0.0:
-            return blas_scal(1.0 / veclen, vec)
+            return blas_scal(1.0 / veclen, vec).astype(vec.dtype)
```
jayantj: Hi @accraze , thanks for the fix! A better solution here might be to modify the hardcoded dtype in line 423 above, it simplifies the logic, and also ensures that the dtype is consistent for vectors with all zeros too (a rather trivial and probably uncommon case, of course)

menshikh-iv:
Ping @accraze, are you planning to finish this PR?

accraze:
@menshikh-iv yes will get it finished this week

gensim/matutils.py
accraze: Hey @jayantj, I looked into this, however `blas_scal` returns an array of type `float` (see line 398). Not sure if there is a better way to handle this...

menshikh-iv:
@accraze tests failed because next problem in `blas_scal` from https://github.com/RaRe-Technologies/gensim/pull/1761/files#diff-346353d71f16d5fe11b7c3efcfef9b4eR429, this call change a type

menshikh-iv:
ping @accraze

menshikh-iv:
Ping @accraze, are you planning to finish this PR?

gensim/matutils.py
pushpankar: @jayantj What else needs to done? `blas_scal` is not being used anywhere. So, should I define `blas_scal` before line 429 and remove hardcoded float from the definition of `blas_scal`

accraze:
@menshikh-iv I'm still looking for a solution, it seems that the `test_normmodel` tests are failing with my changes. I've tried moving `blas_scal` before line 429 and remove hardcoded float from its the definition, however that still did not work either.

menshikh-iv:
Thanks for description @accraze 
@jayantj have you any ideas about it? Probably cast before return should be the best solution for this case.

gensim/matutils.py
```diff
@@ -420,13 +420,13 @@ def unitvec(vec, norm='l2'):
...
     if isinstance(vec, np.ndarray):
-        vec = np.asarray(vec, dtype=float)
+        vec = np.asarray(vec, dtype=vec.dtype)
         if norm == 'l1':
             veclen = np.sum(np.abs(vec))
         if norm == 'l2':
             veclen = blas_nrm2(vec)
         if veclen > 0.0:
-            return blas_scal(1.0 / veclen, vec)
+            return blas_scal(1.0 / veclen, vec).astype(vec.dtype)
```
jayantj: It looks like this cast at the end `astype(vec.dtype)` is responsible for causing the returned matrix to be an int-matrix with all zeros. It is impossible to maintain consistency in types between the input array and the returned array for an int input array (since a normalized vector can almost never be an int array). A solution could be to check if the input array is int-like, and handle it accordingly?

o-P-o: I have issued a new pull request addressing this.

menshikh-iv:
Hey @accraze, how is going? Are you going to finish PR?

similar PR #1866

menshikh-iv:
Looks abandoned, I close this PR.

PR #1763, opened 2017-12-05
Add forgotten encode

horpto:
Fix for #1750 -- use explicit encoding when writing strings to disk.


menshikh-iv:
Fix problem from #1750

menshikh-iv:
Thanks for fast fix @horpto :+1:

PR #1766, opened 2017-12-06
Add inference for new unseen author for `gensim.models.AuthorTopicModel`

Stamenov:
Add function `get_new_author_topics()` to infer topics distribution of a new unseen author, by passing a `corpus` - list of documents in "bag of words" format.


Stamenov:
Sill some open TODOs:

- how should the rho() function be defined inside the `get_new_author_topics()`
- unit test

gensim/models/atmodel.py
```diff
@@ -882,6 +883,98 @@ def get_document_topics(self, word_id, minimum_probability=None):
 
         raise NotImplementedError('Method "get_document_topics" is not valid for the author-topic model. Use the "get_author_topics" method.')
 
+    def get_new_author_topics(self, corpus, minimum_probability=None):
+        """
```
menshikh-iv: Please use numpy-style docstrings: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html, https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

gensim/models/atmodel.py
```diff
@@ -882,6 +883,98 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+        # should not record the sstats, as we are goint to delete the new author after calculated
+        try:
+            gammat, _ = self.inference(
+                corpus, self.author2doc, self.doc2author, rho(),
+                collect_sstats=False, chunk_doc_idx=corpus_doc_idx
+            )
+        except Exception as e:
+            #something went wrong! Rollback temporary changes in object and log
+            rollback_new_author_chages()
+            logging.error(traceback.format_exc())
```
menshikh-iv: no need to use `traceback` here,  use `logging.exception(e)` instead

gensim/models/atmodel.py
```diff
@@ -882,6 +883,98 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+        gamma_new = self.random_state.gamma(100., 1. / 100., (num_new_authors, self.num_topics))
+        self.state.gamma = np.vstack([self.state.gamma, gamma_new])
+
+        # should not record the sstats, as we are goint to delete the new author after calculated
+        try:
+            gammat, _ = self.inference(
+                corpus, self.author2doc, self.doc2author, rho(),
+                collect_sstats=False, chunk_doc_idx=corpus_doc_idx
+            )
+        except Exception as e:
```
menshikh-iv: Too broad clause, please specify the concrete type(s) of exception.

Stamenov:
Of course I will continue my work here. Thanks for quick review @menshikh-iv . I have already addressed your remarks in my code.

olavurmortensen:
Great work @Stamenov. At a glance, it looks good to me, although do I have some comments.

Let me see if I understand the logic of the code, correct me if I'm wrong. First of all, you are inferring `gamma` for a collection of documents, assuming that they are attributed to a single author. So what you do is:
* Add documents to `self.corpus`.
* Add a single temporary author to author dictionaries.
* Randomly initialize `gamma`, as per usual.
* Run `self.inference` to obtain gammas for the documents.
* Remove authors and documents from model.

This is what you do, right?

I don't think adding the documents to `self.corpus` is necessary, because `self.inference` only uses the documents you pass to it. Calling `self.extend_corpus` is also very slow, unfortunately, so always avoid this if possible.

As a side note, the `chunk_doc_idx` argument to `self.inference` has to do with getting an author list from `self.doc2author[doc_no]`, it doesn't have to do with what documents are accessed.

`rho` is a bit tricky, as you mention. The point of `rho` is to interpolate between a previously computed `gamma` and the new one, as in [line 443 in the code](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/atmodel.py#L443). I'm not quite sure about how to use it here this just yet.

It makes sense to make it in a single pass, since you're not updating the lambdas, and only one author is being updated. So it may be a good idea to make sure the number of iterations ([over each document](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/atmodel.py#L431)) is high enough. Maybe let `self.inference` accept `iterations` as an (optional) argument.

Finally, I must note that it isn't obvious what inference on held-out data should be in the author-topic model. I think that is because observations aren't independent. This method doesn't take into account that many author may contribute to single document, which is really the strength of the AT model. That being said, I think this way is the best way to do it.

Sorry for the wall of text :stuck_out_tongue_closed_eyes: Keep up the good work!

gensim/models/atmodel.py
```diff
@@ -882,6 +882,99 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+            logger.warning("AuthorTopicModel.get_new_author_topics() called with an empty corpus")
+            return
+        if not len_input_corpus < self.chunksize:
+            logger.warning("AuthorTopicModel.get_new_author_topics() called with to many documents. Use update().")
+            return
+
+        new_author_name = "placeholder_name"
+
+        # Add new documents in corpus to self.corpus.
+        self.extend_corpus(corpus)
```
olavurmortensen: Remove this line. self.inference uses the corpus supplied as input, and calling self.extend_corpus is very slow.

menshikh-iv:
ping @Stamenov, what's status here?

Stamenov:
Sorry, christmas activities. Will try to push the recommended changes these days.

Stamenov:
https://travis-ci.org/RaRe-Technologies/gensim/jobs/324067238 
Can someone help me out with this build error? Tox package?
Thanks.

piskvorky:
@Stamenov thanks! @menshikh-iv will help out with the build error once he's back from holiday (next week) :)

menshikh-iv:
@Stamenov this isn't a tox problem (pip can't install tox package because the network disappeared in Travis), I re-run job, let's wait for the result.

gensim/models/atmodel.py
```diff
@@ -882,6 +882,84 @@ def get_document_topics(self, word_id, minimum_probability=None):
 
         raise NotImplementedError('Method "get_document_topics" is not valid for the author-topic model. Use the "get_author_topics" method.')
 
+    def get_new_author_topics(self, corpus, minimum_probability=None):
+        """Infers topics for new author.
```
menshikh-iv: please use numpy-style docstrings: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html and https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

gensim/models/atmodel.py
```diff
@@ -882,6 +882,84 @@ def get_document_topics(self, word_id, minimum_probability=None):
 
         raise NotImplementedError('Method "get_document_topics" is not valid for the author-topic model. Use the "get_author_topics" method.')
 
+    def get_new_author_topics(self, corpus, minimum_probability=None):
```
menshikh-iv: What's a reason to `minimum_probability=None` here (instead of 1e-8)?

gensim/models/atmodel.py
```diff
@@ -882,6 +882,84 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+            self.num_authors -= num_new_authors
+
+        try:
+            len_input_corpus = len(corpus)
+        except TypeError:
+            logger.warning("input corpus stream has no len(); counting documents")
+            len_input_corpus = sum(1 for _ in corpus)
+        if len_input_corpus == 0:
+            logger.warning("AuthorTopicModel.get_new_author_topics() called with an empty corpus")
+            return
```
menshikh-iv: Why "empty return" needed? Maybe better raise exception explicitly (for all empty `return`)

gensim/models/atmodel.py
olavurmortensen: @menshikh-iv It is the same in `get_document_topics` in LDA (https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ldamodel.py#L987), and `get_term_topics` as well.

gensim/models/atmodel.py
```diff
@@ -882,6 +882,84 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+        if not len_input_corpus < self.chunksize:
```
olavurmortensen: What is the reason for this limitation? I don't see it implemented anywhere else.

gensim/models/atmodel.py
```diff
@@ -882,6 +882,84 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+            logger.warning("AuthorTopicModel.get_new_author_topics() called with to many documents. Use update().")
+            return
+
+        new_author_name = "placeholder_name"
+        corpus_doc_idx = list(range(0, len_input_corpus))
```
olavurmortensen: Instead set `corpus_doc_idx` to `list(range(self.total_docs, self.total_docs + len_input_corpus))`.

gensim/models/atmodel.py
```diff
@@ -882,6 +882,84 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+        # Add the new placeholder author to author2id/id2author dictionaries.
+        num_new_authors = 1
+        author_id = 0
+        self.author2id[new_author_name] = author_id + self.num_authors
+        self.id2author[author_id + self.num_authors] = new_author_name
+
+        # Add new author in author2doc and doc into doc2author.
+        self.author2doc[new_author_name] = corpus_doc_idx
+        for new_doc_id in corpus_doc_idx:
+            self.doc2author[new_doc_id] = [new_author_name]
```
olavurmortensen: Because of how you set `corpus_doc_idx`, you're overwriting `self.cod2author[0]` (0 through `len_corpus_idx`) here.

olavurmortensen:
@Stamenov I made some comments in the code. 

I'm afraid I've written the `inference` method in a sort of clumsy way, which makes this more cumbersome than it should be, sorry about that. The code really could use a refactor, but I don't have the time to do that.

gensim/models/atmodel.py
Stamenov: so, I should leave it like this, right?

menshikh-iv: @Stamenov yes, leave as is

menshikh-iv:
ping @Stamenov, when you plan to finish this PR?

Stamenov:
I am doing this parallel to my thesis, so mids February.

menshikh-iv:
@Stamenov what else do you need to complete this PR?

Stamenov:
Hi, I am have done some fixing around the builds.
Also in order to satisfy the Travis CI build I had to change the docstring for get_aurhor_topics(), which was not in the scope of my PR. Is this ok, can you please look into it?
All but one is now green, ci/circleci, and it seems to be ldamodel-related, so I could use some help there.
Furthermore, I need some test cases to implement, so if you have some off the top of your head, I will appreciate them.
I am also working on a accompanying evaluation report (paper) for authorship prediction on a couple of datasets, using this inference function, if this will be of any interest for Gensim.

menshikh-iv:
@Stamenov I convert your docstring to numpy-style (please use it always in gensim) + merge fresh `develop` to your PR, that should fix all problems with CI services.
About test cases for this feature @olavurmortensen can you propose something?
About evaluation report - yes, this will be a nice addition for related ipython notebook!

menshikh-iv:
ping @Stamenov, I fix all needed stuff, can you finish your PR?

Stamenov:
Yes, I am working on a simple test case, similar to current for atmodel. Could please provide me with some instruction on how to run the tests locally?

menshikh-iv:
@Stamenov in your case, you need run only one file (not all tests), for this reason, enough run file with tests directly `python gensim/test/test_atmodel.py` or `python -m unittest -v gensim.test.test_atmodel`

Stamenov:
Ahaa so simple, I was a bit confused because I expected some package like "py.test" or similar to be involved, thanks. 
I will really try to push the test case and complete the PR it until the end of the week. Thank you for your patience and support.

menshikh-iv:
@Stamenov in CI we used `py.test` as the runner (but the tests didn't use any specific `py.test` features), for this reason, you can run it with default `unittest`

menshikh-iv:
@Stamenov glad to help :+1:

Stamenov:
I have commited a test-case and I wonder if it is satisfactory for you?
Currently I am iterating over the rho() function and investigate how it influences my prediction quality.

gensim/models/atmodel.py
```diff
@@ -919,15 +919,91 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+        # Should not record the sstats, as we are goint to delete the new author after calculated.
+        try:
+            gammat, _ = self.inference(
+                corpus, self.author2doc, self.doc2author, rho(),
+                collect_sstats=False, chunk_doc_idx=corpus_doc_idx
+            )
+        except ValueError as e:
+            # Something went wrong! Rollback temporary changes in object and log
+            rollback_new_author_chages()
+            logging.exception(e)
```
menshikh-iv: Maybe better raise it (or at least `warning.warn`)? If user has no configured `logging`, this will be silently skipped. How much is this exception critical?

gensim/test/test_atmodel.py
```diff
@@ -450,6 +450,31 @@ def testTermTopics(self):
             self.assertTrue(isinstance(topic_no, int))
             self.assertTrue(isinstance(probability, float))
 
+    def testNewAuthorTopics(self):
```
menshikh-iv: CC: @olavurmortensen, this test looks like "sanity-check" only, can you propose something more significant?

gensim/models/atmodel.py
```diff
@@ -919,15 +919,91 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+        minimum_probability : float, optional
+            Ignore topics with probability below this value, if None - 1e-8 is used.
+
+        Returns
+        -------
+        list of (int, float)
+            Topic distribution for the given `corpus`.
+
+        """
+        # TODO: how should this function look like for get_new_author_topics?
```
menshikh-iv: What's about this TODO?

gensim/test/test_atmodel.py
Stamenov: Besides the sanity check, I tried another test case, where I would infer the topics for a new author, using the exact same documents from the corpus that an author already is assigned to. We can expect that the topic distributions should be very similar, as it is, but not close enough so that we can use np.allclose(..,..). Example:

In [2]: model.get_new_author_topics(corpus=corpus[1:3])
Out[2]: [(0, 0.914887958236441), (1, 0.08511204176355897)]

In [3]: model.get_author_topics(author_name="sally")
Out[3]: [(0, 0.9290420023752036), (1, 0.07095799762479651)]

gensim/models/atmodel.py
Stamenov: fixed

gensim/models/atmodel.py
menshikh-iv: I understand your point, most critical here - rollback, so, let's move rollback to `final` section (and this will be done anyway).

Why don't you want to raise an exception (with the condition that rollback anyway happens)?

gensim/test/test_atmodel.py
menshikh-iv: Sounds reasonable, only one thing - may be better to calculate [JS-divergense](https://en.wikipedia.org/wiki/Jensen%E2%80%93Shannon_divergence) instead of `allclose` with very hight `atol`?

menshikh-iv: https://github.com/RaRe-Technologies/gensim/blob/a54e3369bf832bc632d20f5e1d401267e6bc4882/gensim/matutils.py#L921 - implementation already available in gensim

Stamenov: How would you propose to set the similarity threshold? Just empirically for this one test case?

gensim/models/atmodel.py
Stamenov: ok, sounds good. So I would just let the exception to be raised:
```

        try:
            gammat, _ = self.inference(
                corpus, self.author2doc, self.doc2author, rho(),
                collect_sstats=False, chunk_doc_idx=corpus_doc_idx
            )
        else:
            new_author_topics = self.get_author_topics(new_author_name, minimum_probability)
        finally:
            rollback_new_author_chages()
        return new_author_topics
```

Stamenov:
```
        try:
            gammat, _ = self.inference(
                corpus, self.author2doc, self.doc2author, rho(),
                collect_sstats=False, chunk_doc_idx=corpus_doc_idx
            )
            new_author_topics = self.get_author_topics(new_author_name, minimum_probability)
        finally:
            rollback_new_author_chages()
        return new_author_topics

```
I would propose this. In this case the exception is raised, but rollback is called.

gensim/test/test_atmodel.py
menshikh-iv: yes, I don't know the better solution here.

Stamenov:
@menshikh-iv I have pushed the test and exception improvement, also the rho() function which works best for me. Hope it is satisfactory.

menshikh-iv:
Great work @Stamenov :+1: LGTM for me (only [fix PEP8 mistakes please](https://travis-ci.org/RaRe-Technologies/gensim/jobs/346534464#L510))

@olavurmortensen ping, PR look ready, are you approve?

olavurmortensen:
@Stamenov @menshikh-iv Sorry for not responding, I was sick for a while. I'll take a look soon.

gensim/models/atmodel.py
```diff
@@ -919,15 +919,85 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+        try:
+            len_input_corpus = len(corpus)
+        except TypeError:
+            logger.warning("input corpus stream has no len(); counting documents")
+            len_input_corpus = sum(1 for _ in corpus)
+        if len_input_corpus == 0:
+            raise ValueError("AuthorTopicModel.get_new_author_topics() called with an empty corpus")
+
+        new_author_name = "placeholder_name"
+        corpus_doc_idx = list(range(self.total_docs, self.total_docs + len_input_corpus))
```
olavurmortensen: Add a comment explaining this line. For example: "indexes representing the documents in the input corpus".

gensim/models/atmodel.py
```diff
@@ -919,15 +919,85 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+
+        # Add the new placeholder author to author2id/id2author dictionaries.
+        num_new_authors = 1
+        author_id = 0
```
olavurmortensen: I think this looks confusing. Just say `author_id = self.num_authors`, or something like that.

gensim/models/atmodel.py
```diff
@@ -919,15 +919,85 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+
+        try:
+            len_input_corpus = len(corpus)
+        except TypeError:
+            logger.warning("input corpus stream has no len(); counting documents")
+            len_input_corpus = sum(1 for _ in corpus)
+        if len_input_corpus == 0:
+            raise ValueError("AuthorTopicModel.get_new_author_topics() called with an empty corpus")
+
+        new_author_name = "placeholder_name"
```
olavurmortensen: Just for completeness, check that this author isn't already in the dictionary.

gensim/models/atmodel.py
```diff
@@ -919,15 +919,85 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+        # Add new author in author2doc and doc into doc2author.
+        self.author2doc[new_author_name] = corpus_doc_idx
+        for new_doc_id in corpus_doc_idx:
+            self.doc2author[new_doc_id] = [new_author_name]
+
+        gamma_new = self.random_state.gamma(100., 1. / 100., (num_new_authors, self.num_topics))
+        self.state.gamma = np.vstack([self.state.gamma, gamma_new])
+
+        # Should not record the sstats, as we are goint to delete the new author after calculated.
+        try:
```
olavurmortensen: Does this try block throw an exception if there is a problem?

olavurmortensen:
I made some comments in the code. I think it would also nice to see some example use-cases on real data, hopefully with good results, to see if this actually works.

Stamenov:
@olavurmortensen I will post a notebook, where I train and predict authors on the Reuters 50x50 dataset soon!

gensim/models/atmodel.py
Stamenov: Yes, it does, for example, when corpus is invalid (list of something else), or if the inference is somehow interrupted, or maybe something else. Must make sure to rollback the temporary changes in any case. If makes sense to show the exception to the user, but leave the model state untouched.

menshikh-iv:
@Stamenov how is going? when you plan to finish PR (looks like this almost done).

Stamenov:
Hi, the PR is ready, as far as code and functionality is concerned. 
The only thing is the playbook, which I have prepared yesterday, but I was getting very strange and  inconsistent results. I have to take my time to debug it properly. 
I can commit a preliminary version of it though.
I would say this PR is closable and I can open up a new one for the playbook?

menshikh-iv:
@Stamenov better to add notebook in current PR (this is hardly-related changes), we'll merge together
new functionality & notebook with an example as one PR.

Also, don't forget to resolve cosmetic issues based on @olavurmortensen comments.

menshikh-iv:
@Stamenov great, code looks good to merge :+1:, the last thing - please add an example to https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/atmodel_tutorial.ipynb

Stamenov:
Thanks. I am working on it. I believe this weel it should be done.

menshikh-iv:
Hello @Stamenov 
1. You upload some html (instead of ipynb), look inside your file
2. Please add a new section to https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/atmodel_tutorial.ipynb (instead of adding new notebook)

Stamenov:
Hi @menshikh-iv ,
 I see your point with adding a new section instead of a new notebook, but I find it would be very difficult in keeping it clean. There would be a lot of double code from the atmodel_tutorial.ipynb  and the notebook itself would become very long. Also I use a different dataset, not the NIPS one.

menshikh-iv:
Thanks for your work @Stamenov :+1: nice feature!

Thanks @olavurmortensen for review :+1:

jonaschn (commented):
I am interested in an appropriate value of `rhot` for visualizing the AT model in  https://github.com/bmabey/pyLDAvis/pull/161

gensim/models/atmodel.py
```diff
@@ -919,15 +919,87 @@ def get_document_topics(self, word_id, minimum_probability=None):
...
+        minimum_probability : float, optional
+            Ignore topics with probability below this value, if None - 1e-8 is used.
+
+        Returns
+        -------
+        list of (int, float)
+            Topic distribution for the given `corpus`.
+
+        """
+        def rho():
```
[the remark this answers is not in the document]
jonaschn: In a different place you explain rho as following:
https://github.com/RaRe-Technologies/gensim/blob/75dce4b50174b5a1afb37b163b903f7e2af903b0/gensim/models/atmodel.py#L812-L816

Issue #1697, opened 2017-11-07
New binary corpus format


menshikh-iv:
Based on https://github.com/RaRe-Technologies/gensim/pull/1679#issuecomment-341686324
New corpus in binary format:
Store each document as `number_of_document_nonzeros` `int32` followed by serialized (feature_id, feature_weight) pairs as (`int32`, `float32`) = 8 bytes per non-zero element. So, total serialized corpus `size = num_docs * 4 + 8 * nnz`.

This is the simple solution (without indexing/slicing support).

piskvorky:
Let me add that a crucial part of this task is measuring and optimizing the read/write performance of this new corpus serialization format. Its main purpose is to be efficient and fast (otherwise we might as well continue using MmCorpus).

piskvorky:
Another idea: can we stream directly from/tp the three CSC arrays? That would be the easiest solution, and most flexible (CSC/CSR are the most standard sparse matrix formats).

menshikh-iv:
@piskvorky are you mean scipy sparse implementation of matrices or maybe "custom" solution?

piskvorky:
Nothing custom, I mean standard CSC (=three dense arrays).

abhijeetchauhan:
If no one else is working on this issue. Please assign it to me.

menshikh-iv:
@abhijeetchauhan if you are interested - feel free to submit PR.

abhijeetchauhan:
I have to code this in cython right?

menshikh-iv:
@abhijeetchauhan Cython/C/C++, pick any.
Potentially, this can be done on pure python, but it's not an easy task to make it fast enough.

arlenk:
@abhijeetchauhan if it helps and you are still interested, I have some code for exploring a few different options here: https://github.com/arlenk/gensim/tree/binary_corpus

I tested out the following:

- the python standard library's "struct" module.  
Pros: no cython code, struct module is supported and well tested
Cons: binary file is not portable (hardware specific endian, packing, etc.).  C
- structs using cython and fread
Pros: faster
Cons: binary file is not portable, manual memory management, only works with local, non-zipped, files.
- structs using cython and fread, but returning the results in an cython iterator class instead of a list of tuples
Pros: same as above
Cons: same as above plus result is an iterator so any code that depends on a list will break (though code should probably work for an iterator anyway)
[...]

PR #1767, opened 2017-12-07
Add cython version for "hot" functions from `gensim.models.LdaModel`

arlenk:
Attempt at moving some of the hotspots for ldamodel into a cython module. From some profiling, most of the time in ldamodel.LdaModel is spent in the following few spots:

dirichlet_expectation (specifically the call to psi())
calculating the mean absolute difference to determine when gamma has converged
logsumexp (even with the inline version)

I broke these functions out into a cython module. For now, I just created a fastldamodel.py (a copy of ldamodel.py but with the cython functions above) module to see if there is any interest in pursuing this. Based on some profiling on my pc (attached), the speed up is about 100% (meaning the new code takes half the time -- though this is dependent on number of topics, number of passes, etc.).

I also noticed that the current ldamodel code uses float32 datatypes, but the user can change this if desired. If we go down the cython route, we'd have to decide if we want to allow the user to change the datatype as this would require different cython functions for each data type (I believe). Alternatively, we could use the cython code when the datatype is float32 (and perhaps float64) and fall back to the slower versions otherwise.


arlenk:
fyi, I reworked this change to mirror the implementation of word2vec_inner and doc2vec_inner.

The cython functions now live in ldamodel_inner.pyx and LdaModel tries to use the fast cython versions if possible, but falls back to the "slower" alternatives if it cant.  I also explicitly added separate cython functions for float32 and float64, so the faster code should work with both of these dtypes now (I don't know how to handle float16s nicely in the cython code so for now that falls back to the slower code as well).  

I used cython's fused datatypes to make the cython code a little nicer, but the code still has a lot of duplication in it.  I'm not familiar with other templating methods in cython, but I'm sure there must be a way to make the functions generic without requiring so much duplication.

The _attach_inner_methods() function in LdaModel tries to pick up the right methods to use based on the availability of ldamodel_inner and the dtype.  There's probably a nicer way of doing this as well.

I haven't checked all the edge cases yet, but my simple tests so far seem to work.

This implementation has the benefit of allowing the multicore and distributed versions to use the faster cython code as well.

I also left in the older "fastldamodel.py" files just to make comparisions between the two implementations easier.

gensim/models/ldamodel.py
```diff
@@ -83,6 +83,87 @@ def logsumexp(x):
...
+    # else fall back to python/numpy
+    FAST_VERSION = -1
+    
+    logsumexp_2d_f32 = logsumexp_2d_f64 = logsumexp
+    dirichlet_expectation_1d_f32 = dirichlet_expectation_1d_f64 = dirichlet_expectation
+    dirichlet_expectation_2d_f32 = dirichlet_expectation_2d_f64 = dirichlet_expectation
+    mean_absolute_difference_f32 = mean_absolute_difference_f64 = mean_absolute_difference
+
+
+def _attach_inner_methods(self, dtype):
```
menshikh-iv: Probably better to define needed functions on module level (not "patch" `LdaModel` at runtime), this allows avoiding a problem with backward compatibility.

gensim/models/_fastldamodel.pyx
```diff
@@ -0,0 +1,256 @@
+from __future__ import division
```
menshikh-iv: This file is needed right now (after last changes where you move all "slowest" function to cython)?

menshikh-iv: Also, `.pyd` file (windows compiled) should be here (only `pyx` and `c`)

setup.py
```diff
@@ -13,7 +13,7 @@
 import os
 import sys
 import warnings
-
+import numpy as np
```
menshikh-iv: That's really wired, please don't use numpy in `setup.py` (by this reason all CI broken)

gensim/models/ldamodel.py
```diff
@@ -83,6 +83,87 @@ def logsumexp(x):
...
+    -------
+    float32
+        mean(abs(a - b))
+    
+    """
+    return np.mean(np.abs(a - b))
+
+try:
+    # try to load fast, cythonized code if possible
+    from gensim.models.ldamodel_inner import dirichlet_expectation_1d_f32, dirichlet_expectation_1d_f64
```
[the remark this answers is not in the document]
piskvorky: Up to you. I'd be mindful of the added complexity -- what sort of performance difference are we talking about?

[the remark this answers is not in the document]
menshikh-iv: @arlenk can you compare these variants?

gensim/models/_fastldamodel.pyx
arlenk: no, not need.  Left it in just for comparison, but removed it now.

gensim/models/ldamodel.py
```diff
@@ -83,6 +83,87 @@ def logsumexp(x):
...
+
+    Returns
+    -------
+    float32
+        mean(abs(a - b))
+    
+    """
+    return np.mean(np.abs(a - b))
+
+try:
```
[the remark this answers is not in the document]
arlenk: not sure how to restructure this yet.  The except clause needs to happen after the "slow" versions of logsumexp and mean_absolute_difference are defined, so that's why the code is where it is right now.  I initially included the definition of the slow versions in the except block as well (mirroring what word2vec does), but that breaks if anyone uses the float16 dtype which requires the slow functions even if the cython versions exist (since the cython versions can only handle float32 and float 64).

might make things easier just to move the "inner" functions (logsumexp, dirichlet_expectation, and mean_abs_diff) into their own python module and have that module import the cython versions if available.  Open to any other suggestions as well.

gensim/models/ldamodel.py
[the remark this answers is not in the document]
arlenk: will take me a few days but I will include some benchmark results.

The typed versions (_f32 and _f64) are not entirely for performance reasons though.  The c code needs to be different for float vs. doubles.  I used cython's fused types to remove some of the duplication, but I don't think there's a way to have python call a single function that will dispatch (statically) to the correct float vs. double cython function.  We could check the dtypes at run time, but that does impose a pretty large penalty given how often these functions are called (eg, dirichlet_expectation is called millions of times even for a small corpus).

I can include a benchmark with the run-time dtype check to see how big the impact is though.

gensim/models/ldamodel.py
arlenk: that was my original intention (and that's what the original "fastldamodel.py" code did), but I am not sure how to make this work nicely with LdaModel since a user could create multiple models with different dtypes.  Eg,

model0 = LdaModel(corpus=corpus, id2word=corpus.id2word, num_topics=10, dtype=np.float16)
model1 = LdaModel(corpus=corpus, id2word=corpus.id2word, num_topics=10, dtype=np.float32)
model2 = LdaModel(corpus=corpus, id2word=corpus.id2word, num_topics=10, dtype=np.float64)

each of those objects would need to use different implementations of the inner functions (unless we dispatched at run-time based on dtypes, but as mentioned before, i think that will have a decent impact on speed).

setup.py
arlenk: ok removed this.  Didn't realize that you could include the numpy header files for the cython code without importing numpy directly.

gensim/models/ldamodel.py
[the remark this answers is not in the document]
menshikh-iv: Ok, let's leave it for now as is

gensim/models/ldamodel.py
menshikh-iv: Agree, you depend on `dtype` of LDA

menshikh-iv: Aha, you are right, `dtype` is a really fresh feature (for LDA), I forgot about it.
Ok, probably that's variant is OK (another variant - add this as method to LDA and call in init)

piskvorky: I believe we should decide what dtype makes the most sense, and stick with it (drop support for all other dtype).

It's not a parameter we can ask users to choose for themselves (especially since we give no guidance, no pros/cons/recommendations).

menshikh-iv: @piskvorky probably `float32` should be the best choice (by speed, memory consumption and quality)

piskvorky: I agree, but would like to see some numeric arguments (benchmarks, model quality & performance evals) to back that up.

arlenk: removing the other dtypes would certainly make this change easier. 

I was running some tests with the different dtypes, and float16 (one of the allowed options in the existing code) seems to over/underflow with some reasonable input data.  For example the following code (I left out the imports) will crash with an overflow error:

np.seterr(all='raise')

text = api.load('text8')
id2word = gensim.corpora.Dictionary(text)
corpus = [ id2word.doc2bow(t) for t in text ]

LdaModel(corpus=corpus, id2word=id2word, num_topics=10, dtype=np.float16)

arlenk:
> Can you show benchmark results please (LdaModel/LdaMulticore with numpy/scipy stuff VS LdaModel/LdaMulticore with cythonized functions)

Here are the results for a couple different settings (only using LdaModel for now, not LdaMulticore):
https://gist.github.com/arlenk/ea745c1d05e323d87df96c0e6b490d05

The speed up is about 75% for float64 dtypes.

Here's the code used for the benchmarks:
https://gist.github.com/arlenk/ec02cce1fe571b4c05ef3853c49289a3

menshikh-iv:
@arlenk your results look great! thanks for benchmarking (but please post it as a table in this PR for history).

 Can you make the same benchmark for the different dtype case?

arlenk:
Sure, here are the profiling results for the different dtypes.  Note I haven't checked the accuracy or memory usage yet, so this is just for comparing speed.  Anecdotally, I noticed that some values underflow for float32 when taking np.exp(gammad), but I don't think this is necessarily a problem.

| dataset | num_docs | num_terms | num_topics | passes | dtype   | code     | elapsed seconds |
|---------|----------|-----------|------------|--------|---------|----------|:---------------:|
| nytimes | 300k     | 100k      | 50         | 1      | float64 | original | 3473            |
|         |          |           |            |        |         | cython   | 1975            |
| enron   | 40k      | 30k       | 20         | 3      | float16 | original | 1068            |
|         |          |           |            |        | float32 | original | 804             |
|         |          |           |            |        |         | cython   | 412             |
|         |          |           |            |        | float64 | original | 774             |
|         |          |           |            |        |         | cython   | 437             |

Interesting that at least on my computer, float16 takes more time than the others.  Seems like np.dot is the culprit for me so this may depend on the linear algebra library installed (I am using conda's distribution which i believe includes the mkl).

I will make sure to include tests cases as well once we figure out if we should allow multiple dtypes or restrict to just one.

arlenk:
Ok, did some further refactoring to simplify the code and added unittests.  

This version now drops all the type specific (f32 vs. f64) code and just falls back on run time type checking and "casting" when necessary.  Eg, the cython code is written to accept double or floats (using a fused type).  If LdaModel is created with float32 or float64, the appropriate cython function is called based on checking the dtype of the input.  Turns out the performance penalty from this is relatively small (5%) which I think is worth the reduced complexity of the code.  If the user creates an LdaModel with dtype=float16, the code will just "cast" (using np.astype) this to a double and call the double function.

Since there are no more type-specific functions, determining if the cython code is available is done at the module level (like in doc2vec and word2vec) so there's no need to "attach" functions to LdaModel objects after they are created.

I moved the python versions of logsumexp and mean_absolute_difference up to the matutils module (where dirichlet_expectation is).  I did this so that a version of these functions would always exist as this made the unittests simpler.  But maybe these functions (and dirichlet_expectation) should reside somewhere else since I believe only LdaModel* uses them.

arlenk:
Also ran some comparisons of models built with different dtypes (just based on the reported perplexity from fitting the models).  This is using the nytimes dataset (300k documents, 100k words).  At least with this dataset, there doesn't seem to be any downside to using float32, so it's probably a nice option to keep if memory is scarce relative to the corpus size.

The second chart is the same as the first, with just the first few passes cropped to improve the scale.

dtype_perplexity
dtype_perplexity_cropped

arlenk:
FYI, I realize the CI builds are failing, but I am having trouble scrolling to the bottom of the log output to see what the failures are (the page just hangs for me).  Is this an issue for anyone else?

menshikh-iv:
@arlenk  great work!

- about the place for  `mean_absolute_difference`, etc: yes, this used only in LDA* but `matutils` is a good place for this functions

- about "I noticed that some values underflow for float32 when taking np.exp(gammad), but I don't think this is necessarily a problem": this probably can be a real problem, have you code/data for reproducing this behavior (because this can "break" LDA)? CC: @piskvorky 

- about `float16`: this type looks "useless", thanks for your benchmark, now we know it!

- about failed tests (needed info for the fix)

```
[00:10:28] FAIL gensim/test/test_ldamodel.py::TestLdaModel::testGetDocumentTopics
[00:10:28] FAIL gensim/test/test_ldamodel.py::TestLdaMulticore::testGetDocumentTopics
```

```
[00:10:28] ================================== FAILURES ===================================
[00:10:28] _____________________ TestLdaModel.testGetDocumentTopics ______________________
[00:10:28] 
[00:10:28] self = <gensim.test.test_ldamodel.TestLdaModel testMethod=testGetDocumentTopics>
[00:10:28] 
[00:10:28]     def testGetDocumentTopics(self):
[00:10:28]     
[00:10:28]         model = self.class_(
[00:10:28]             self.corpus, id2word=dictionary, num_topics=2, passes=100, random_state=np.random.seed(0)
[00:10:28]         )
[00:10:28]     
[00:10:28] >       doc_topics = model.get_document_topics(self.corpus)
[00:10:28] 
[00:10:28] model      = <gensim.models.ldamodel.LdaModel object at 0x074A66B0>
[00:10:28] self       = <gensim.test.test_ldamodel.TestLdaModel testMethod=testGetDocumentTopics>
[00:10:28] 
[00:10:28] gensim\test\test_ldamodel.py:219: 
[00:10:28] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[00:10:28] 
[00:10:28] self = <gensim.models.ldamodel.LdaModel object at 0x074A66B0>
[00:10:28] bow = <gensim.corpora.mmcorpus.MmCorpus object at 0x074A6710>
[00:10:28] minimum_probability = 0.01, minimum_phi_value = 0.01, per_word_topics = False
[00:10:28] 
[00:10:28]     def get_document_topics(self, bow, minimum_probability=None, minimum_phi_value=None,
[00:10:28]                             per_word_topics=False):
[00:10:28]         """
[00:10:28]             Args:
[00:10:28]                 bow (list): Bag-of-words representation of the document to get topics for.
[00:10:28]                 minimum_probability (float): Ignore topics with probability below this value
[00:10:28]                     (None by default). If set to None, a value of 1e-8 is used to prevent 0s.
[00:10:28]                 per_word_topics (bool): If True, also returns a list of topics, sorted in
[00:10:28]                     descending order of most likely topics for that word. It also returns a list
[00:10:28]                     of word_ids and each words corresponding topics' phi_values, multiplied by
[00:10:28]                     feature length (i.e, word count).
[00:10:28]                 minimum_phi_value (float): if `per_word_topics` is True, this represents a lower
[00:10:28]                     bound on the term probabilities that are included (None by default). If set
[00:10:28]                     to None, a value of 1e-8 is used to prevent 0s.
[00:10:28]     
[00:10:28]             Returns:
[00:10:28]                 topic distribution for the given document `bow`, as a list of
[00:10:28]                 `(topic_id, topic_probability)` 2-tuples.
[00:10:28]             """
[00:10:28]         if minimum_probability is None:
[00:10:28]             minimum_probability = self.minimum_probability
[00:10:28]         minimum_probability = max(minimum_probability, 1e-8)  # never allow zero values in sparse output
[00:10:28]     
[00:10:28]         if minimum_phi_value is None:
[00:10:28]             minimum_phi_value = self.minimum_probability
[00:10:28]         minimum_phi_value = max(minimum_phi_value, 1e-8)  # never allow zero values in sparse output
[00:10:28]     
[00:10:28]         # if the input vector is a corpus, return a transformed corpus
[00:10:28]         is_corpus, corpus = utils.is_corpus(bow)
[00:10:28]         if is_corpus:
[00:10:28]             kwargs = dict(
[00:10:28]                 per_word_topics=per_word_topics,
[00:10:28]                 minimum_probability=minimum_probability,
[00:10:28]                 minimum_phi_value=minimum_phi_value
[00:10:28]             )
[00:10:28] >           return apply(corpus, **kwargs)
[00:10:28] E           NameError: name 'apply' is not defined
[00:10:28] 
[00:10:28] bow        = <gensim.corpora.mmcorpus.MmCorpus object at 0x074A6710>
[00:10:28] corpus     = <gensim.corpora.mmcorpus.MmCorpus object at 0x074A6710>
[00:10:28] is_corpus  = True
[00:10:28] kwargs     = {'minimum_phi_value': 0.01, 'minimum_probability': 0.01, 'per_word_topics': False}
[00:10:28] minimum_phi_value = 0.01
[00:10:28] minimum_probability = 0.01
[00:10:28] per_word_topics = False
[00:10:28] self       = <gensim.models.ldamodel.LdaModel object at 0x074A66B0>
[00:10:28] 
[00:10:28] gensim\models\ldamodel.py:1015: NameError
```

As I see from your results, `float32` is the best choice, I'm just concerned about the "underflow" problem, need to investigate it in more details.

arlenk:
> about "I noticed that some values underflow for float32 when taking np.exp(gammad), but I don't think this is necessarily a problem": this probably can be a real problem, have you code/data for reproducing this behavior (because this can "break" LDA)? CC: @piskvorky

This code reproduces the underlows for me (note this underflows on the current development branch as well):

```python
import logging, gensim
import gensim.corpora
import numpy as np
from gensim.models.ldamodel import LdaModel

np.seterr(all='raise')

text = api.load('text8')
id2word = gensim.corpora.Dictionary(text)
corpus = [ id2word.doc2bow(t) for t in text ]

LdaModel(corpus=corpus, id2word=id2word, num_topics=100, dtype=np.float32)
```

arlenk:
> about failed tests (needed info for the fix)
>
> [00:10:28] FAIL gensim/test/test_ldamodel.py::TestLdaModel::testGetDocumentTopics
> [00:10:28] FAIL gensim/test/test_ldamodel.py::TestLdaMulticore::testGetDocumentTopics

looks like I accidentally introduced a typo in an unrelated part of the code... this should be fixed now.

menshikh-iv:
@arlenk big thanks for the concrete example :+1: , I reproduced this problem with `python2` on Linux.
@piskvorky apparently we should use `float64` here (by "underflow" of `float32`)

menshikh-iv:
@arlenk generally, we need to investigate why "underflow" happens, fix it and stay only this data type. Thanks for your great work!

Have you any ideas how to understand, why underflow problem happens? Probably, this is a bad design of the algorithm, but I have no idea how to debug this.

arlenk:
> @arlenk generally, we need to investigate why "underflow" happens, fix it and stay only this data type. Thanks for your great work!
> 
> Have you any ideas how to understand, why underflow problem happens? Probably, this is a bad design of the algorithm, but I have no idea how to debug this.

I don't think it's an algorithm issue, but I will try digging into this some more.  

In this particular case, I don't think it's necessarily an issue.  The underflowed value is later used in a dot product, so the term involving the underflowed value is going to be ~zero in any case.  If all the values in the dot product were underflows, that would be an issue.  But I don't think having a few underflows will break the algorithm.

Having said that, it still may make sense to default to float64 to avoid underflows (I have not run into any yet with float64 dtype).  Float32 does save memory, of course, but that could still be left as an option for anyone who is really running low on memory.

menshikh-iv:
Thank you @arlenk!
@piskvorky tell us where to "dig" further, we have no ideas.

setup.py
```diff
@@ -258,7 +257,10 @@ def finalize_options(self):
             include_dirs=[model_dir]),
         Extension('gensim.models.fasttext_inner',
             sources=['./gensim/models/fasttext_inner.c'],
-            include_dirs=[model_dir])
+            include_dirs=[model_dir]),
+        Extension('gensim.models.ldamodel_inner',
```
menshikh-iv: Don't forget to modify `MANIFEST.ini` in root of repo (add your files)

setup.py
```diff
@@ -258,7 +257,10 @@ def finalize_options(self):
...
+            sources=['./gensim/models/ldamodel_inner.c'],
```
menshikh-iv: Maybe better name for it is `_matutils` (because this isn't model at all, this only concrete math functions)?

menshikh-iv: Also, move this file on level of `matutils.py` ? wdyt?

gensim/models/ldamodel.py
```diff
@@ -56,32 +55,26 @@
     np.float64: 1e-100,
 }
 
+try:
```
menshikh-iv: we can move this to `matutils` too (try/exc section)

menshikh-iv:
@arlenk please resolve merge conflict & make suggested fixes.

I think we'll fix underflow problem not soon, what's prevents us now to merge this change?

arlenk:
> @arlenk please resolve merge conflict & make suggested fixes.
> 
> I think we'll fix underflow problem not soon, what's prevents us now to merge this change?

Ok, I made the suggested changes.  

I also removed the use of dirichlet_expectation_1d and _2d and instead just added a cython version of dirichlet_expectation() the calls the _1d or _2d based on the input shape.  Turns out, after some profiling, the added ndim check add less than a few percent of overhead and having a single dirichlet_expectation function simplifies the code.

I think this also means that hdpmodel and atmodel should benefit from the changes as well (as they call matutils.dirichlet_expectation)

menshikh-iv:
@arlenk UPD, I'll do it myself, don't worry (anyway I need to fix docstrings)

menshikh-iv:
@arlenk beautiful! You made a very dotty and important change, which seriously affects the performance of Lda (and related models, that used LDA as the base model). Both of your PRs (MmReader and this one) will be our main features in next release :boom:

Do not want to continue gensim cythonization? We would be very happy if you have time and wish for it :heart:.

piskvorky:
Awesome job! Thanks so much @arlenk .

arlenk:
> @arlenk beautiful! You made a very dotty and important change, which seriously affects the performance of Lda (and related models, that used LDA as the base model). Both of your PRs (MmReader and this one) will be our main features in next release 💥
> 
> Do not want to continue gensim cythonization? We would be very happy if you have time and wish for it ❤️.

good to hear.  LDA model is what I am most familiar with, but if there are other bottlenecks that have come up on the past, let me know and I am happy to take a look.

menshikh-iv:
@arlenk I have several optimization tasks
- (simple) #1697 new binary corpus. Idea - make one more corpus in simple binary format. This should be fast to read/write (reading speed is more important here) and compact (as the file).
- (hard) [Multi-stream API](https://github.com/RaRe-Technologies/gensim/wiki/GSoC-2018-project-ideas#multiple-stream-api), this is really big, but bottleneck for many models (especially for *2vec), look at related issues (that mentioned in description) for understanding, what's happens here

The last task need a really long time, but first looks simple and does not require so much time

PR #1768, opened 2017-12-07
Fix hashing-trick from FastText.build_vocab. Fix #1765

### LINKED ISSUE #1765: Bucket Argument in fasttext not working as expected ?
Hi, For the fasttext native from gensim:

My understanding is that according to the hashing trick, if bucket is < total # of subwords, there will be collisions and some subwords will be mapped to the same integers.  Am I wrong? 
However, it is not what I see on a toy example: 

```python
import gensim
from gensim.models.fasttext import FastText

sent = [['lol', 'dds', 'sdsf'], ['anticonsti']]
model = FastText(min_count = 1, bucket = 20)
model.build_vocab(sentences=sent)
model.train(sentences = sent, epochs = 1, report_delay = 1.0)

model.wv.ngrams
```
#### Expected Results
Dictionary with ngrams and their mappings to integers between 0 and 19 ( buckets = 20)

#### Actual Results
Dictionary with ngrams and their mappings to integers between 0 and 55 ( number of ngrams is 56 here)

Windows-10-10.0.14393-SP0
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  5 2016, 11:41:13) [MSC v.1900 64 bit (AMD64)]
NumPy 1.13.3
SciPy 1.0.0
gensim 3.1.0
FAST_VERSION 0
>>>

manneshiva:
Fixes and adds a unit test for #1765 .

### DIFF
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -138,7 +138,7 @@ class FastText(Word2Vec):
             ngram_indices = []
             new_hash_count = 0
             for i, ngram in enumerate(all_ngrams):
-                ngram_hash = ft_hash(ngram)
+                ngram_hash = ft_hash(ngram) % self.bucket
                 if ngram_hash in self.wv.hash2index:
                     self.wv.ngrams[ngram] = self.wv.hash2index[ngram_hash]
                 else:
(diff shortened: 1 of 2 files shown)


piskvorky:
Whoa, how did this even work? What was it doing before? Implications for the previous evaluations and comparisons?

Thanks for the fix!

PR #1769, opened 2017-12-07
Add explicit deprecation warnings for all outdated stuff. Fix #1753

### LINKED ISSUE #1753: Add explicit deprecation warnings for outdated stuff
Need to add explicit deprecation warnings for all outdated stuff.
Docstrings isn't good enough for this situation.

What should contain the warning:
- What's an alternative (if exists), i.e. "Call X instead of DEPRECATED_Y"
- When this function will be removed (concrete version, for example, 4.0.0)

related message https://github.com/RaRe-Technologies/gensim/issues/1746#issuecomment-348680651

menshikh-iv:
Resolve issue #1753

### DIFF
--- a/gensim/utils.py
+++ b/gensim/utils.py
@@ -1294,13 +1294,11 @@ def deprecated(reason):
 
             @wraps(func)
             def new_func1(*args, **kwargs):
-                warnings.simplefilter('always', DeprecationWarning)
                 warnings.warn(
                     fmt.format(name=func.__name__, reason=reason),
                     category=DeprecationWarning,
                     stacklevel=2
                 )
-                warnings.simplefilter('default', DeprecationWarning)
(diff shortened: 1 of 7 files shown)


gensim/utils.py
```diff
@@ -1280,3 +1281,46 @@ def lazy_flatten(nested_list):
...
+            return new_func1
+        return decorator
+
+    elif inspect.isclass(reason) or inspect.isfunction(reason):
+        func = reason
+        fmt = "Call to deprecated `{name}`."
+
+        @wraps(func)
+        def new_func2(*args, **kwargs):
+            warnings.simplefilter('always', DeprecationWarning)
```
piskvorky: What is this `simplefilter` stuff for? Some comments would be nice.

PR #1770, opened 2017-12-07
Fix epsilon according to dtype in LdaModel

menshikh-iv:
That's small fix for #1656, big thanks @rmalouf for catch this potential bug https://github.com/RaRe-Technologies/gensim/pull/1656#issuecomment-345473831

What's done:
- limit number of types (only `numpy.float16`, `numpy.float32`, `numpy.float64`)
- add adoptive `eps` depends on `dtype` (instead of hardcoded `1e-100`)

CC: @piskvorky

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -287,7 +287,7 @@ class LdaModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
 
         """
         if dtype not in {np.float16, np.float32, np.float64}:
-            raise ValueError("Incorrect 'dtype', please choice one of numpy.float16, numpy.float32 or numpy.float64")
+            raise ValueError("Incorrect 'dtype', please choose one of numpy.float16, numpy.float32 or numpy.float64")
 
         self.dtype = dtype
 
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -50,6 +50,12 @@ from gensim.models.callbacks import Callback
 
 logger = logging.getLogger('gensim.models.ldamodel')
 
+DTYPE_TO_EPS = {
+    np.float16: 1e-5,
+    np.float32: 1e-35,
+    np.float64: 1e-100,
+}
+
 
 def logsumexp(x):
     """Log of sum of exponentials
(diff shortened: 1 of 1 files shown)


gensim/models/ldamodel.py
```diff
@@ -286,6 +286,9 @@ def __init__(self, corpus=None, num_topics=100, id2word=None,
         >>> lda = LdaModel(corpus, num_topics=50, alpha='auto', eval_every=5)  # train asymmetric alpha from data
 
         """
+        if dtype not in {np.float16, np.float32, np.float64}:
+            raise ValueError("Incorrect 'dtype', please choice one of numpy.float16, numpy.float32 or numpy.float64")
```
piskvorky: `choice` => `choose`

gensim/models/ldamodel.py
```diff
@@ -498,7 +501,8 @@ def inference(self, chunk, collect_sstats=False):
             # The optimal phi_{dwk} is proportional to expElogthetad_k * expElogbetad_w.
             # phinorm is the normalizer.
             # TODO treat zeros explicitly, instead of adding 1e-100?
-            phinorm = np.dot(expElogthetad, expElogbetad) + 1e-100
+            eps = 1e-100 if self.dtype == np.float64 else (1e-35 if self.dtype == np.float32 else 1e-5)
```
piskvorky: Easier to read and maintain as a mapping?

```python
dtype_to_eps = {
    np.float64: 1e-100,
    np.float32: 1e-35,
    np.float16: 1e-5,
}
```

And then also `if dtype not in dtype_to_eps: ...`

gensim/models/ldamodel.py
```diff
@@ -498,7 +501,13 @@ def inference(self, chunk, collect_sstats=False):
             # The optimal phi_{dwk} is proportional to expElogthetad_k * expElogbetad_w.
             # phinorm is the normalizer.
             # TODO treat zeros explicitly, instead of adding 1e-100?
```
piskvorky: Should update comment: `1e-100` => `epsilon`.

gensim/models/ldamodel.py
```diff
@@ -498,7 +501,13 @@ def inference(self, chunk, collect_sstats=False):
...
-            phinorm = np.dot(expElogthetad, expElogbetad) + 1e-100
+            dtype_to_eps = {
```
piskvorky: Better defined at module level, so `__init__` can make use of the allowed keys? Now the same information is in three places, not DRY.

PR #1772, opened 2017-12-08
Fix import in `get_my_ip`. Fix #1771

### LINKED ISSUE #1771: Pyro4.naming exception in `get_my_ip`
On a multi home nic computer, get_my_ip() is failing to use socket connection to nameserver to retrieve the external ip address to use. The code is proceeding to the default case to use hostname and to look up by ip address, which is currently resolving to virtualbox ethernet adapter that is not reachable from the dispatcher.

Specifically, the line 
`ns = Pyro4.naming.locateNS()`
is throwing exception 
`module 'Pyro4' has no attribute 'naming'`

when using Pyro4 version 4.62, thus get_my_ip() is defaulting to use
`result = socket.gethostbyname(socket.gethostname())`

The solution is to add 
`import Pyro4.naming`

darindf:
Fixes #1771

### DIFF
--- a/gensim/utils.py
+++ b/gensim/utils.py
@@ -626,6 +626,7 @@ def get_my_ip():
     import socket
     try:
         import Pyro4
+        import Pyro4.naming
         # we know the nameserver must exist, so use it as our anchor point
         ns = Pyro4.naming.locateNS()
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Thanks @darindf, congratz with the first contribution!

PR #1773, opened 2017-12-09
path error handeled

jet2016:


piskvorky:
@jet2016 can you expand the PR description with some motivation and background for this feature request?

I'm -1 here: Gensim is a library, we don't want any interactive input prompts. If a user wants to create some directory path on their disk, it should be easy enough to do outside of Gensim.

menshikh-iv:
I'm -1 too, no need interactive prompts in the bowels of gensim.

PR #1774, opened 2017-12-10
[MRG] Updates to poincare notebooks

jayantj:
Makes the following updates to the tutorial notebook - 

1. Removes minute technical details and evaluation results
2. Adds links to detailed blog post and other useful links

Evaluation notebook - 
1. Adds results for regularized models
2. HTML tables instead of ascii-tables to easily publish in blogs.

### DIFF
--- a/docs/notebooks/Poincare Evaluation.ipynb
+++ b/docs/notebooks/Poincare Evaluation.ipynb
@@ -799,7 +799,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 145,
+   "execution_count": 176,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


janpom:
This looks good. One minor thing: The "Dimensions" label is off in the results table. It should span all the dimension columns and shouldn't span the "Metric" column.

menshikh-iv:
@jayantj 
- "Note that not all the above libraries are part of the gensim dependencies, so they might need to be installed separately. These requirements are listed in the poincare requirements.txt" - please add cell with `pip install ...` line instead (this file already does not exist)
- This is the second rename `evaluation` -> `tutorial` -> `evaluation`, please explain what is happening

jayantj:
@menshikh-iv do you mean I should hardcode the pip installs in the notebook itself?

menshikh-iv:
@jayantj yes, this is simpler and easy-to-edit if needed.

PR #1775, opened 2017-12-10
Add model to dict method

kakshay21:
Adding to the word2vec to production pipeline #1269


gojomo:
With 11,207 lines added, and 254 removed, it's hard to tell what the meaningful change is here. Could you submit a minimal patch instead, perhaps by not re-running all cells after the change?

kakshay21:
Sure, I understand.
Do you want me to do that in separate PR? @gojomo

PR #1776, opened 2017-12-11
Add `model_to_dict` one-liner to word2vec notebook. Fix #1269

### LINKED ISSUE #1269: Word2Vec model to dict; Adding to the word2vec to production pipeline

kakshay21:
Adding to the word2vec to production pipeline #1269

### DIFF
--- a/docs/notebooks/word2vec.ipynb
+++ b/docs/notebooks/word2vec.ipynb
@@ -1299,6 +1299,34 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "### Adding Word2Vec \"model to dict\" method to production pipeline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def model_to_dict():\n",
+    "    # method to find and store all most similar words to all words in the trained vocabulary.\n",
+    "    most_similars_precalc = {word : model.wv.most_similar(word) for word in model.wv.index2word}\n",
+    "    return most_similars_precalc"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "model_to_dict()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "## Conclusion\n",
     "\n",
     "In this tutorial we learned how to train word2vec models on your custom data and also how to evaluate it. Hope that you too will find this popular tool useful in your Machine Learning tasks!\n",
@@ -1336,7 +1364,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython2",
-   "version": "2.7.13"
+   "version": "2.7.10"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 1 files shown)


kakshay21:
I've created a small patch now @gojomo

docs/notebooks/word2vec.ipynb
```diff
@@ -1295,6 +1295,34 @@
...
+    "### Adding Word2Vec \"model to dict\" method to production pipeline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def model_to_dict():\n",
```
menshikh-iv: You no need this function for oneliner in notebook, 
`most_similars_precalc = {word : model.wv.most_similar(word) for word in model.wv.index2word}` enough here

docs/notebooks/word2vec.ipynb
```diff
@@ -1295,6 +1295,34 @@
     "print(train_times_table)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Adding Word2Vec \"model to dict\" method to production pipeline"
```
menshikh-iv: Please add more motivation (why this can be useful, it will be good if you show concrete use-case, how you'll use `most_similars_precalc` )

kakshay21:
Any changes required in the documentation? @menshikh-iv

menshikh-iv:
@kakshay21 concrete example with comparison with/without caching

docs/notebooks/word2vec.ipynb
```diff
@@ -1295,6 +1295,31 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Adding Word2Vec \"model to dict\" method to production pipeline\n",
+    "Suppose, we still want more performance improvement in production. \n",
+    "One good way is to cache all the similar words in a dictionary.\n",
+    "\n",
```
menshikh-iv: No needed `\n` after each sentence + you should add a link to the concrete issue in the notebook.

menshikh-iv:
Thanks @kakshay21, congratz with the first contribution :1st_place_medal:

Issue #1623, opened 2017-10-11
Potential unification/optimization/simplification/enhancement refactor of *2Vec & related algorithms (FastText, Sent2Vec, FastSent, etc)


gojomo:
Word2Vec, Doc2Vec, FastText, FastSent (#612), Sent2Vec (#1376), 'Doc2VecWithCorruption' (#1159) and others are variants on the same core technique. They should share more code, and perhaps even be implemented as alternate parameter-choices on the same refactored core functions.

A big refactoring (including from-scratch API design) could potentially offer some or all of the following:

1. sharing more code between different modes (SG/CBOW/DBOW/DM/FastText-classification/other), by discovering the ways they're parameterized variants of a shared process

2. making other creative variations possible, even if just experimentally (different kinds of context-windows, dropout strategies, alternate learning-optimizations like AdaGrad/etc, re-weightings of individual examples/vectors, separate input/output vocabularies, 'bloom embeddings', more kinds of 'inference', etc)

3. making it easier to use non-natural-language datasets, perhaps by providing ability to supply examples in an interim (raw-int-indexes) format (other than string tokens), and example transformer/caching classes that turn either texts or other corpuses into the right format

4. eliminating the hard-to-maintain dual-path pure-Python & Cython implementations - perhaps by going to something like Numba-only, or removing the (performance-non-competitive) pure-Python paths whenever Cython code is clean enough

5. avoiding common user errors & sources of confusion - by renaming parameters/methods, updating defaults, separating logically distinct steps into independent code/classes – then providing updated demo notebooks showing the new modes of operation

6. throughput optimizations, including getting away from the 'master single-corpus-reader thread', or using processes rather than threads if that's the only way to avoid GIL contention bottlenecks

7. separating vocabulary-management into explicitly different classes/objects, for more control/customization, perhaps including closer integration with new n-gram (phrasing) options

piskvorky:
Re. 4: I agree. I feel we can drop the pure-Python compatibility in general. Running many of the algorithms (word2vec included) in pure Python is rarely what the users need; it's more often a bug, a problem with their setup.

piskvorky:
Re. 6: threads are fine, but we should extend the single-corpus-reader API to multiple-corpus-readers. There has already been some discussion around this, though the issue link escapes me.

gojomo:
[...]

PR #1777, opened 2017-12-11
Re-design "*2vec" implementations

manneshiva:
This PR aims to better the current "*2vec" implementation design (Doc2Vec, Word2Vec, FastText, Sent2Vec, Poincare), making it more modular and ensuring maximum code re-use.
[Link](https://github.com/manneshiva/gensim/wiki/Any2Vec-Refactoring-Summary) summarizing this PR.


manneshiva:
The first commit contains the initial design that will help facilitate discussion. It outlines present thoughts about the design and is in no way complete.

gojomo:
Simply extracting/clarifying the existing implementation could offer value – but also is less-likely to clear the way for all-new algorithmic variations or optimizations. For example, a very generic "2Vec" model might not even have the idea of a 'vocabulary', or a step within itself that builds-a-vocabulary... as opposed to just taking corpuses or configuration prepped by other objects. 

That is, this sort of incremental re-factoring starting from the current design might not offer the same reuse/efficiency/flexibility benefits for #1623 a from-scratch design could give.)

manneshiva:
@gojomo  Agreed. These first few commits were only meant to facilitate discussions in our meetings. The idea is to sketch the design in a top-down fashion -- first, fix the public APIs that are to be exposed to users and then refactor everything else accordingly. This will involve designing a few classes from scratch, at the same time, will try and make use of existing classes/functions by making them more generic. Currently, the idea is to try and maintain backward compatibility (in terms of end-user APIs) unless some design changes make it absolutely necessary to change them (APIs). The scope of this PR may not include every point in #1623 but is definitely a big step in that direction which will make it easier to incorporate many of the listed points in the future.
Current status: We have finalized on the public interfaces. We also discussed providing/incorporating `callbacks` functionality in a way similar to `keras`. You should see be able to see these commits soon. Your review/comments would be helpful.

gojomo:
I'll keep an eye for reviewable updates. I would suggest starting from a long, weakly-prioritized list of potential new functionality, then cull that to the things that can be usefully merged. (That is, more bottom-up from specifics of novel functions/combinations.) Starting from top-level, end-user visible interfaces, *and* maintaining compatibility with the (somewhat clunky) older interfaces, risks severely-limiting what new breakthrough capabilities can be delivered. (If they were easy to fit into the existing API, they could be added incrementally, rather than via a new design.)

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,454 @@
...
+
+            for callback in self.callbacks:
+                callback.on_batch_end(self)
+
+            progress_queue.put((len(sentences), tally, raw_tally))  # report back progress
+            jobs_processed += 1
+        logger.debug("worker exiting, processed %i jobs", jobs_processed)
+
+    def _job_producer(self, data_iterator, job_queue, cur_epoch=0, total_examples=None, total_words=None):
+            """Fill jobs queue using the input `data_iterator`."""
```
janpom: indented extra level

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,446 @@
...
+            epochs=epochs,
+            total_examples=total_examples,
+            total_words=total_words, **kwargs)
+
+        for callback in self.callbacks:
+            callback.on_train_begin(self)
+
+        trained_word_count = 0
+        raw_word_count = 0
+        start = default_timer() - 0.00001
```
menshikh-iv: What's a magic constant here?

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,446 @@
...
+            trained_word_count += trained_word_count_epoch
+            raw_word_count += raw_word_count_epoch
+
+            for callback in callbacks:
+                callback.on_epoch_end(self)
+
+        # Log overall time
+        total_elapsed = default_timer() - start
+        logger.info(
+            "training on a %i raw words (%i effective words) took %.1fs, %.0f effective words/s",
```
menshikh-iv: This is always "words"? Probably "entities" is better for the general case.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,446 @@
...
+        self.train_count += 1  # number of times train() has been called
+        self._clear_post_train()
+
+        for callback in self.callbacks:
+            callback.on_train_end(self)
+        return trained_word_count, raw_word_count
+
+    @classmethod
+    def load(cls, fname_or_handle, **kwargs):
+        model = super(BaseAny2VecModel, cls).load(fname_or_handle, **kwargs)
```
menshikh-iv: `return super(...)`

janpom:
Some thoughts about the parameter naming. Using a generic term such as `entity` (meaning `word` or `sentence` or `document`) is potentially confusing for the user. Also, I see little value in keeping the API exactly the same for all algorithms. Don't get me wrong. There's certainly value in keeping the API consistent so that once the user learns how to use `Word2Vec`, it will be easy for them to use `Doc2Vec`, but it's not necessary (and probably not even desirable) that the public API is **exactly the same** including parameter names.

For example, I consider the following:

```python
doc2vec = Doc2Vec(data)
print(doc2vec.kv.similarity(document1=mydoc1, document2=mydoc2))

word2vec = Word2Vec(data)
print(word2vec.kv.similarity(word1="foo", word2="bar"))
```

better than:

```python
doc2vec = Doc2Vec(data)
print(doc2vec.kv.similarity(entity1=mydoc1, entity2=mydoc2))

word2vec = Word2Vec(data)
print(word2vec.kv.similarity(entity1="foo", entity2="bar"))
```

Should the different parameter naming be a problem for code reuse, we can always keep the same parameter names in internal methods and define public methods as thin wrappers. For example:

```python
class KeyedVectorsBase(object):
    def _similarity(self, entity1, entity2):
        # calculate and return similarity

class Word2VecKeyedVectors(KeyedVectorsBase):
    def similarity(self, word1, word2):
        return self._similarity(word1, word2)

class Doc2VecKeyedVectors(KeyedVectorsBase):
    def similarity(self, document1, document2):
        return self._similarity(document1, document2)
```

gensim/models/word2vec.py
```diff
@@ -508,130 +233,59 @@ def __init__(self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
...
-        self.batch_words = batch_words
-        self.model_trimmed_post_training = False
+        # self.null_word = null_word
         self.compute_loss = compute_loss
         self.running_training_loss = 0
+
+        self.wv = Word2VecKeyedVectors()
+        self.vocabulary = Word2VecVocab(
+            max_vocab_size=max_vocab_size, min_count=min_count, sample=sample,
+            sorted_vocab=sorted_vocab, null_word=null_word)
```
janpom: `sorted_vocab=bool(sorted_vocab)` better since the param is a boolean in `Word2VecVocab`.

gensim/models/word2vec.py
```diff
@@ -437,7 +162,7 @@ class Word2Vec(utils.SaveLoad):
     def __init__(self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
                  max_vocab_size=None, sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
                  sg=0, hs=0, negative=5, cbow_mean=1, hashfxn=hash, iter=5, null_word=0,
-                 trim_rule=None, sorted_vocab=1, batch_words=MAX_WORDS_IN_BATCH, compute_loss=False):
+                 trim_rule=None, sorted_vocab=1, batch_words=MAX_WORDS_IN_BATCH, compute_loss=False, callbacks=()):
```
janpom: `callbacks` not documented in the docstring

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,446 @@
...
+        self.total_train_time = 0
+        self.batch_words = batch_words
+        self.model_trimmed_post_training = False
+        self.callbacks = callbacks
+
+    def build_vocab(self, data_iterable, update=False, progress_per=10000, **kwargs):
+        """Scan through all the data and create/update vocabulary.
+        Should also initialize/reset/update vectors for new vocab entities.
+        """
+        self.vocabulary.scan_vocab(data_iterable, progress_per=progress_per, **kwargs)
```
janpom: The problem with this code is that `self.vocabulary` and `self.trainables` don't even exist in the base class. I understand that we rely on the subclass to define these, but it's pretty confusing.

I'm still thinking of a better way of doing this.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,446 @@
...
+    def _worker_loop(self, job_queue, progress_queue):
+        """Train the model, lifting lists of data from the job_queue."""
+        thread_private_mem = self._get_thread_working_mem()
+        jobs_processed = 0
+        while True:
+            job = job_queue.get()
+            if job is None:
+                progress_queue.put(None)
+                break  # no more jobs => quit this worker
+            sentences, job_parameters = job
```
janpom: It would be nice to unify variable naming. We have `sentences` here, `data_iterable` or `batch_data` elsewhere.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,446 @@
...
+
+    def _job_producer(self, data_iterator, job_queue, cur_epoch=0, total_examples=None, total_words=None):
+        """Fill jobs queue using the input `data_iterator`."""
+        job_batch, batch_size = [], 0
+        pushed_words, pushed_examples = 0, 0
+        next_job_params = self._get_job_params()
+        job_no = 0
+
+        for data_idx, data in enumerate(data_iterator):
+            sentence_length = self._raw_word_count([data])
```
janpom: Not necessarily `sentence`.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,446 @@
...
+                "be sure to provide a corpus that offers restartable iteration = an iterable)."
+            )
+
+        # give the workers heads up that they can finish -- no more work!
+        for _ in xrange(self.workers):
+            job_queue.put(None)
+        logger.debug("job loop exiting, total %i jobs", job_no)
+
+    def _log_epoch_progress(
+        self, progress_queue, job_queue, cur_epoch=0, total_examples=None, total_words=None, report_delay=1.0):
```
janpom: wrong indentation

gensim/models/base_any2vec.py
manneshiva: Sure, will make this change for all variables.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,446 @@
...
+        if total_words and total_words != raw_word_count:
+            logger.warning(
+                "EPOCH - %i : supplied raw word count (%i) did not equal expected count (%i)", cur_epoch,
+                raw_word_count, total_words
+            )
+        self.total_train_time += elapsed
+        return trained_word_count, raw_word_count
+
+    def _train_epoch(self, data_iterable, cur_epoch=0, total_examples=None,
+              total_words=None, queue_factor=2, report_delay=1.0):
```
janpom: wrong indentation

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,446 @@
...
+        """Train one epoch."""
+        job_queue = Queue(maxsize=queue_factor * self.workers)
+        progress_queue = Queue(maxsize=(queue_factor + 1) * self.workers)
+
+        workers = [
+        threading.Thread(
+            target=self._worker_loop,
+            args=(job_queue, progress_queue,))
+        for _ in xrange(self.workers)
```
janpom: wrong indentation

gensim/models/word2vec.py
```diff
@@ -508,130 +233,59 @@ def __init__(self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
...
+        self.vocabulary = Word2VecVocab(
+            max_vocab_size=max_vocab_size, min_count=min_count, sample=sample,
+            sorted_vocab=sorted_vocab, null_word=null_word)
+        self.trainables = Word2VecTrainables(
+            vector_size=size, seed=seed, alpha=alpha, min_alpha=min_alpha, hs=hs, negative=negative, hashfxn=hashfxn)
+
+        super(Word2Vec, self).__init__(
+            workers=workers, vector_size=size, epochs=iter, callbacks=callbacks, batch_words=batch_words)
+
         if sentences is not None:
```
janpom: Fair amount of duplicated code here (and in FastText).

gensim/models/doc2vec.py
```diff
@@ -630,13 +185,18 @@ def __init__(self, documents=None, dm_mean=None, dm=1, dbow_words=0, dm_concat=0
         Note: The rule, if given, is only used prune vocabulary during build_vocab() and is not stored as part
         of the model.
         """
+        # from IPython.core.debugger import set_trace
```
janpom: some leftovers?

gensim/callbacks.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+
+class Callback(object):
+    """Abstract base class used to build new callbacks."""
+
+    def __init__(self):
+        self.model = None
```
janpom: What's the point of this attribute? We pass `model` to all callback methods.

gensim/models/__init__.py
```diff
@@ -13,7 +13,7 @@
 from .logentropy_model import LogEntropyModel  # noqa:F401
 from .word2vec import Word2Vec  # noqa:F401
 from .doc2vec import Doc2Vec  # noqa:F401
-from .keyedvectors import KeyedVectors  # noqa:F401
+from .word2vec import Word2VecKeyedVectors as KeyedVectors  # noqa:F401
```
menshikh-iv: An old variant is better (like from @jayantj) with `KeyedVectors = SomeNewClass` by documentation reasons (KeyedVectors does not exist as a class -> never be in the documentation).

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,792 @@
...
+
+try:
+    from queue import Queue
+except ImportError:
+    from Queue import Queue
+
+logger = logging.getLogger(__name__)
+
+
+# Public Interfaces
```
menshikh-iv: It's better to add this tip to class docstring (instead of comment in code)

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,792 @@
...
+
+    def __init__(self, workers=3, vector_size=100, epochs=5,
+                 callbacks=(), batch_words=10000):
+        """Initialize model parameters.
+        Subclass should initialize the following attributes:
+        - self.kv (instance of concrete implementation of `BaseKeyedVectors` interface)
+        - self.vocabulary (instance of concrete implementation of `BaseVocabBuilder` abstract class)
+        - self.trainables (instance of concrete implementation of `BaseTrainables` abstract clas)
+        """
+        self.vector_size = int(vector_size)
```
menshikh-iv: Why `int` needed here (same for `workers`)?

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,792 @@
...
+        self.total_train_time = 0
+        self.batch_words = batch_words
+        self.model_trimmed_post_training = False
+        self.callbacks = callbacks
+
+    def build_vocab(self, data_iterable, update=False, progress_per=10000, **kwargs):
+        """Scan through all the data and create/update vocabulary.
+        Should also initialize/reset/update vectors for new vocab entities.
+        """
+        raise NotImplementedError
```
menshikh-iv: Better raise instance, not a class (tip from https://www.python.org/dev/peps/pep-0008/), here and everywhere.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,792 @@
...
+
+                # add the sentence that didn't fit as the first item of a new job
+                job_batch, batch_size = [data], data_length
+        # add the last job too (may be significantly smaller than batch_words)
+        if job_batch:
+            job_no += 1
+            job_queue.put((job_batch, next_job_params))
+
+        if job_no == 0 and self.train_count == 0:
+            logger.warning(
```
menshikh-iv: What's a reason to use `logger.warning` here (instead of an exception, or, at least `warning.warn`)?

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,792 @@
...
+        # give the workers heads up that they can finish -- no more work!
+        for _ in xrange(self.workers):
+            job_queue.put(None)
+        logger.debug("job loop exiting, total %i jobs", job_no)
+
+    def _log_progress(self, job_queue, progress_queue, cur_epoch, example_count, total_examples,
+                      raw_word_count, total_words, trained_word_count, elapsed):
+        raise NotImplementedError
+
+    def _log_epoch_end(self, cur_epoch, example_count, total_examples, raw_word_count, total_words,
```
menshikh-iv: This shouldn't be defined as a method of child of `Callback` class?

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,792 @@
...
+                       trained_word_count, elapsed, job_tally):
+        raise NotImplementedError
+
+    def _log_train_end(self, raw_word_count, trained_word_count, total_elapsed):
+        raise NotImplementedError
+
+    def _log_epoch_progress(self, progress_queue, job_queue, cur_epoch=0, total_examples=None, total_words=None,
+                            report_delay=1.0):
+        example_count, trained_word_count, raw_word_count = 0, 0, 0
+        start, next_report = default_timer() - 0.00001, 1.0
```
menshikh-iv: I forgot, why this `- 0.00001` needed?

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,792 @@
...
+        else:
+            if trim_rule is not None:
+                logger.warning(
+                    "The rule, if given, is only used to prune vocabulary during build_vocab() "
+                    "and is not stored as part of the model. Model initialized without sentences. "
+                    "trim_rule provided, if any, will be ignored.")
+
+    # for backward compatibility (aliases pointing to corresponding variables in trainables, vocabulary)
+    @property
+    def iter(self):
```
menshikh-iv: Probably, add "deprecated" decorator for all of this properties (I think we should remove it later).

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,792 @@
...
+        self.vocabulary.index2word = self.wv.__dict__.get('index2word', [])
+
+    def _set_keyedvectors(self):
+        self.wv.vectors = self.trainables.__dict__.get('vectors', [])
+        self.wv.vectors_norm = self.trainables.__dict__.get('vectors_norm', None)
+        self.wv.vector_size = self.trainables.__dict__.get('vector_size', None)
+        self.wv.vocab = self.vocabulary.__dict__.get('vocab', {})
+        self.wv.index2word = self.vocabulary.__dict__.get('index2word', [])
+
+    def _log_progress(self, job_queue, progress_queue, cur_epoch, example_count, total_examples,
```
menshikh-iv: same comment as about callbacks below (for all of this `_log_..`)

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,792 @@
...
+            "training on a %i raw words (%i effective words) took %.1fs, %.0f effective words/s",
+            raw_word_count, trained_word_count, total_elapsed, trained_word_count / total_elapsed
+        )
+        if job_tally < 10 * self.workers:
+            logger.warning(
+                "under 10 jobs per worker: consider setting a smaller `batch_words' for smoother alpha decay"
+            )
+
+    # for backward compatibility
+    @deprecated("Method will be removed in 4.0.0, use self.wv.most_similar() instead")
```
menshikh-iv: Next version will be `4.0.0`, feel free to remove this stuff.
CC: @piskvorky

gensim/models/doc2vec.py
```diff
@@ -55,249 +55,560 @@
...
+from gensim.models.keyedvectors import WordEmbeddingsKeyedVectors
+from types import GeneratorType
 
 logger = logging.getLogger(__name__)
 
 try:
     from gensim.models.doc2vec_inner import train_document_dbow, train_document_dm, train_document_dm_concat
     from gensim.models.word2vec_inner import FAST_VERSION  # blas-adaptation shared from word2vec
-    logger.debug('Fast version of %s is being used', __name__)
+    logger.info("Using FAST VERSION %s", FAST_VERSION)
```
menshikh-iv: `Using FAST_VERSION=%d`

gensim/models/doc2vec.py
```diff
@@ -55,249 +55,560 @@
...
-                                 learn_words=True, learn_hidden=True, word_vectors=None, word_locks=None,
-                                 doctag_vectors=None, doctag_locks=None):
-        """
-        Update distributed memory model ("PV-DM") by training on a single document, using a
-        concatenation of the context window word vectors (rather than a sum or average).
 
-        Called internally from `Doc2Vec.train()` and `Doc2Vec.infer_vector()`.
+        if 'sentences' in kwargs:
+            raise DeprecationWarning(
+                "Parameter 'sentences' was renamed to 'documents', and will be removed in 4.0.0, "
```
menshikh-iv: Why we pass something through `*kwargs` (not as `argument=default_value`) ?

gensim/models/doc2vec.py
```diff
@@ -55,249 +55,560 @@
...
-        Any of `learn_doctags', `learn_words`, and `learn_hidden` may be set False to
-        prevent learning-updates to those respective model weights, as if using the
-        (partially-)frozen model to infer other compatible vectors.
+        super(Doc2Vec, self).__init__(
+            sg=(1 + dm) % 2,
+            null_word=dm_concat,
+            callbacks=callbacks,
+            **kwargs)
+
+        self.load = call_on_class_only
```
menshikh-iv: Why this needed (and only for word2vec/doc2vec)?

gensim/models/doc2vec.py
```diff
@@ -55,249 +55,560 @@
...
-        in the model's vocab dictionary, and `doctag_indexes`, which provide indexes
-        into the doctag_vectors array.
+class Doc2Vec(BaseWordEmbedddingsModel):
+    """Class for training, using and evaluating neural networks described in http://arxiv.org/pdf/1405.4053v2.pdf"""
 
-        Any of `learn_doctags', `learn_words`, and `learn_hidden` may be set False to
-        prevent learning-updates to those respective model weights, as if using the
-        (partially-)frozen model to infer other compatible vectors.
+    def __init__(self, documents=None, dm_mean=None, dm=1, dbow_words=0, dm_concat=0, dm_tag_count=1,
+                 docvecs=None, docvecs_mapfile=None, comment=None, trim_rule=None, callbacks=(), **kwargs):
```
menshikh-iv: What is `comment` ?

gensim/models/doc2vec.py
```diff
@@ -55,249 +55,560 @@
...
+            if self.dm_concat:
+                segments.append('dm/c')  # ...with concatenative context layer
+            else:
+                if self.cbow_mean:
+                    segments.append('dm/m')
+                else:
+                    segments.append('dm/s')
+        segments.append('d%d' % self.trainables.vector_size)  # dimensions
+        if self.negative:
+            segments.append('n%d' % self.negative)  # negative samples
```
menshikh-iv: Doesn't looks like a good solution (and non-oblivious what means `d`, `n`, etc).

gensim/models/keyedvectors.py
```diff
@@ -1036,33 +835,11 @@ def init_sims(self, replace=False):
...
-                for i in xrange(self.syn0.shape[0]):
-                    self.syn0[i, :] /= sqrt((self.syn0[i, :] ** 2).sum(-1))
-                self.syn0norm = self.syn0
+                for i in xrange(self.vectors.shape[0]):
+                    self.vectors[i, :] /= sqrt((self.vectors[i, :] ** 2).sum(-1))
+                self.vectors_norm = self.vectors
             else:
-                self.syn0norm = (self.syn0 / sqrt((self.syn0 ** 2).sum(-1))[..., newaxis]).astype(REAL)
-
-    def get_keras_embedding(self, train_embeddings=False):
```
menshikh-iv: Where this now?

gensim/similarities/docsim.py
```diff
@@ -629,7 +629,10 @@ def get_similarities(self, query):
         result = []
         for qidx in range(n_queries):
             # Compute similarity for each query.
-            qresult = [self.w2v_model.wmdistance(document, query[qidx]) for document in self.corpus]
+            try:
+                qresult = [self.w2v_model.wmdistance(document, query[qidx]) for document in self.corpus]
```
menshikh-iv: What's a reason to check for both  `` and `.wv` ?

gensim/similarities/index.py
```diff
@@ -86,12 +90,12 @@ def __init__(self, model=None, num_trees=None):
         if model and num_trees:
             if isinstance(self.model, Doc2Vec):
                 self.build_from_doc2vec()
-            elif isinstance(self.model, Word2Vec):
+            elif isinstance(self.model, Word2Vec) or isinstance(self.model, FastText):
```
menshikh-iv: Should be `isinstance(self.model, (Word2Vec, FastText))` (here and everywhere)

tox.ini
```diff
@@ -34,7 +34,6 @@ setenv =
     MALLET_HOME={env:MALLET_HOME:}
 
 commands =
-    python -c "from gensim.models.word2vec import FAST_VERSION; print(FAST_VERSION)"
```
menshikh-iv: Why? This is kind of "sanity check".

gensim/test/test_fasttext.py
```diff
@@ -43,8 +40,6 @@ def __iter__(self):
 class TestFastTextModel(unittest.TestCase):
 
     def setUp(self):
-        ft_home = os.environ.get('FT_HOME', None)
```
menshikh-iv: So, remove this from `tox.ini` too

gensim/test/test_fasttext.py
```diff
@@ -325,65 +325,122 @@ def test_doesnt_match(self):
...
+    def test_cbow_hs_training(self):
 
-        # overlap increases as we increase `iter` value, min overlap set to 2 to avoid unit-tests taking too long
-        # this limit can be increased when using Cython code
-        self.assertGreaterEqual(overlap_count, 2)
+        model_gensim = FT_gensim(size=50, sg=0, cbow_mean=1, alpha=0.05, window=5, hs=1, negative=0,
+            min_count=5, iter=5, batch_words=1000, word_ngrams=1, sample=1e-3, min_n=3, max_n=6,
+            sorted_vocab=1, workers=1, min_alpha=0.0)
 
-    def test_cbow_hs_against_wrapper(self):
```
menshikh-iv: All `*_against_*` tests removed (because wrapper dropped), am I right?

gensim/test/test_fasttext.py
```diff
@@ -325,65 +325,122 @@ def test_doesnt_match(self):
...
-            loss='hs', sample=1e-3, negative=0, iter=5, min_n=3, max_n=6, sorted_vocab=1, threads=12)
+        lee_data = LineSentence(datapath('lee_background.cor'))
+        model_gensim.build_vocab(lee_data)
+        orig0 = np.copy(model_gensim.wv.vectors[0])
+        model_gensim.train(lee_data, total_examples=model_gensim.corpus_count, epochs=model_gensim.epochs)
+        self.assertFalse((orig0 == model_gensim.wv.vectors[0]).all())  # vector should vary after training
 
-        model_gensim = FT_gensim(size=50, sg=1, cbow_mean=1, alpha=0.025, window=5, hs=1, negative=0,
+        sims_gensim = model_gensim.wv.most_similar('night', topn=10)
+        sims_gensim_words = (list(map(lambda x: x[0], sims_gensim)))  # get similar words
```
menshikh-iv: `sims_gensim_words = [word for (word, distance) in sims_gensim]` (here and everywhere)

gensim/models/base_any2vec.py
manneshiva: It is a guard against next sample being identical. Look [here](https://github.com/RaRe-Technologies/gensim/blob/5ea3eb2c37cf57665d55c98168ff5b343c479e12/gensim/models/doc2vec.py#L692).

gensim/__init__.py
```diff
@@ -3,7 +3,7 @@
 similarities within a corpus of documents.
 """
 
-from gensim import parsing, matutils, interfaces, corpora, models, similarities, summarization, utils  # noqa:F401
+from gensim import parsing, matutils, interfaces, corpora, models, similarities, summarization, utils, callbacks  # noqa:F401
```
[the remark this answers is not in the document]
manneshiva: There already exists `gensim.models.callbacks`. I can add my `Callback` base class to this file but will require me to rename the base class as another class with the same name exists -- [link](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/callbacks.py#L234).

gensim/callbacks.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Abstract base class to build callbacks. Callbacks are used to apply custom functions over the model
```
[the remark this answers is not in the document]
manneshiva: It may not make sense to make this inherit `ABC` and use `@abstractmethod` since not all the methods (`on_epoch_end`, `on_train_begin` etc.) need to be re-implemented -- just the one required by the user. Changing 'Abstract base class' to 'Base class` in docstring.

gensim/models/base_any2vec.py
manneshiva: Sure, let me complete the doc string for `base_any2vec.py`.

gensim/callbacks.py
```diff
@@ -0,0 +1,49 @@
...
+>>> epoch_saver = ModelEpochSaver('axax')
+>>> model = word2vec.Word2Vec(sentences, iter=5, size=10, min_count=0, seed=42, callbacks=[epoch_saver])
+
+"""
+
+
+class Callback(object):
+    """Abstract base class used to build new callbacks."""
+
+    def __init__(self):
```
janpom: redundant

gensim/callbacks.py
```diff
@@ -0,0 +1,49 @@
...
+>>>     def on_train_begin(self, model):
+>>>         self.cur_epoch = 0
+>>> epoch_saver = ModelEpochSaver('axax')
+>>> model = word2vec.Word2Vec(sentences, iter=5, size=10, min_count=0, seed=42, callbacks=[epoch_saver])
+
+"""
+
+
+class Callback(object):
+    """Abstract base class used to build new callbacks."""
```
janpom: Pls remove `Abstract` since it's not really abstract.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+logger = logging.getLogger(__name__)
+
+
+# Public Interfaces
+class BaseAny2VecModel(utils.SaveLoad):
+
+    def __init__(self, workers=3, vector_size=100, epochs=5,
+                 callbacks=(), batch_words=10000):
+        """Initialize model parameters.
+        Subclass should initialize the following attributes:
```
janpom: Language: "Subclasses" or "A subclass"

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+
+            for callback in self.callbacks:
+                callback.on_batch_end(self)
+
+            progress_queue.put((len(data_iterable), tally, raw_tally))  # report back progress
+            jobs_processed += 1
+        logger.debug("worker exiting, processed %i jobs", jobs_processed)
+
+    def _job_producer(self, data_iterator, job_queue, cur_epoch=0, total_examples=None, total_words=None):
+        """Fill jobs queue using the input `sentences` iterator."""
```
janpom: sentences => data?

gensim/models/base_any2vec.py
janpom: Is this new? I believe this is the old code.

gensim/models/base_any2vec.py
janpom: Would it be feasible to make this a callback? Maybe `EpochProgressLoggingCallback` and use it as the default?

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+
+        self._set_keyedvectors()
+
+        return trained_word_count, raw_word_count, job_tally
+
+    def train(self, data_iterable, epochs=None, total_examples=None,
+              total_words=None, queue_factor=2, report_delay=1.0, callbacks=(), **kwargs):
+        """Handle multi-worker training."""
+        self._set_train_params(**kwargs)
+        if len(callbacks):
```
janpom: What's with the `len`? I see Jayant using it all the time as well. I'm starting to think it's some Indian cultural habit. :) Why not just `if callbacks:`?

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+        trained_word_count = 0
+        raw_word_count = 0
+        start = default_timer() - 0.00001
+        job_tally = 0
+
+        for cur_epoch in range(self.epochs):
+            for callback in self.callbacks:
+                callback.on_epoch_begin(self)
+
+            trained_word_count_epoch, raw_word_count_epoch, job_tally_epoch = self._train_epoch(data_iterable,
```
janpom: Wrong indentation. `data_iterable` should go to the next line.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+
+class BaseVocabBuilder(utils.SaveLoad):
+    """Class for managing vocabulary of a model. Takes care of building, pruning and updating vocabulary."""
+    def __init__(self):
+        self.vocab = {}
+        self.index2word = []
+
+    def scan_vocab(self, data_iterable, progress_per=10000, **kwargs):
+        """Do an initial scan of all words appearing in data_iterable.
+        Sets num_examples(total examples in data_iterable) and
```
janpom: Please put spaces in front of opening parentheses `(`, so that it doesn't look like function calls.

janpom: Do I understand correctly this is meant to set `self.num_examples` and `self.raw_vocab`? If yes, it would be a good idea to create these attributes in the `__init__`, so that they actually exist in the base class. It's OK to initialize to `None`.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+    @cum_table.deleter
+    def cum_table(self):
+        del self.vocabulary.cum_table
+
+    def __str__(self):
+        return "%s(vocab=%s, size=%s, alpha=%s)" % (
+            self.__class__.__name__, len(self.wv.index2word), self.vector_size, self.alpha
+        )
+
+    def build_vocab(self, sentences, update=False, progress_per=10000, **kwargs):
```
janpom: I don't think `sentences` is the right term here.  `data_iterable` instead for consistency with the `train()` method?

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+        """Build vocabulary from a sequence of sentences (can be a once-only generator stream).
+        Each sentence must be a list of unicode strings.
```
janpom: We say the sentence must be a list here and that it must be an iterable below (`iterable of iterables`). Please fix the inconsistency.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+            or a callable that accepts parameters (word, count, min_count) and returns either
+            :attr:`gensim.utils.RULE_DISCARD`, :attr:`gensim.utils.RULE_KEEP` or :attr:`gensim.utils.RULE_DEFAULT`.
+            Note: The rule, if given, is only used to prune vocabulary during build_vocab() and is not stored as part
+            of the model.
+        `update`: bool
+            If true, the new provided words in `word_freq` dict will be added to model's vocab.
+
+        Returns
+        --------
+        None
```
janpom: Is this required by gensim docstring conventions? If not, I would just remove the `Returns` section. Seems redundant to me.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+    def seeded_vector(self, seed_string):
+        """Create one 'random' vector (but deterministic by seed_string)"""
+        raise NotImplementedError
+
+
+class BaseWordEmbedddingsModel(BaseAny2VecModel):
+    def __init__(self, sentences=None, workers=3, vector_size=100, epochs=5, callbacks=(), batch_words=10000,
+                 trim_rule=None, sg=0, alpha=0.025, window=5, seed=1, hs=0, negative=5, cbow_mean=1,
+                 min_alpha=0.0001, compute_loss=False, **kwargs):
+        self.sg = int(sg)
```
janpom: It would be nice to be consistent about the type conversion. If we convert `sg` to `int`, we should probably convert `hs` and `negative` to `int`, too, and convert `compute_loss` to `bool`. Ditto elsewhere.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+        # Instead of scanning text, this will assign provided word frequencies dictionary(word_freq)
+        # to be directly the raw vocab
+        raw_vocab = word_freq
+        logger.info(
+            "collected %i different raw word, with total frequency of %i",
+            len(raw_vocab), sum(itervalues(raw_vocab))
+        )
+
+        # Since no sentences are provided, this is to control the corpus_count
+        self.corpus_count = corpus_count if corpus_count else 0
```
janpom: `self.corpus_count = corpus_count or 0` also possible and shorter.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+        self.compute_loss = compute_loss
+        self.running_training_loss = 0.0
+        return super(BaseWordEmbedddingsModel, self).train(
+            sentences, total_examples=total_examples, total_words=total_words,
+            epochs=epochs, start_alpha=start_alpha, end_alpha=end_alpha, word_count=word_count,
+            queue_factor=queue_factor, report_delay=report_delay, compute_loss=compute_loss, callbacks=callbacks)
+
+    def _get_job_params(self, cur_epoch):
+        """Return the paramter required for each batch."""
+        alpha_ = self.alpha - ((self.alpha - self.min_alpha) * float(cur_epoch) / self.epochs)
```
janpom: Why the trailing underscore? You could just inline the variable.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+        self.min_alpha = end_alpha or self.min_alpha
+        self.compute_loss = compute_loss
+        self.running_training_loss = 0.0
+        return super(BaseWordEmbedddingsModel, self).train(
+            sentences, total_examples=total_examples, total_words=total_words,
+            epochs=epochs, start_alpha=start_alpha, end_alpha=end_alpha, word_count=word_count,
+            queue_factor=queue_factor, report_delay=report_delay, compute_loss=compute_loss, callbacks=callbacks)
+
+    def _get_job_params(self, cur_epoch):
+        """Return the paramter required for each batch."""
```
janpom: *paramter

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+        work = matutils.zeros_aligned(self.trainables.layer1_size, dtype=REAL)  # per-thread private work memory
+        neu1 = matutils.zeros_aligned(self.trainables.layer1_size, dtype=REAL)
+        return work, neu1
+
+    def _raw_word_count(self, job):
+        """Return the number of words in a given job."""
+        return sum(len(sentence) for sentence in job)
+
+    def _check_training_sanity(self, epochs=None, total_examples=None, total_words=None, **kwargs):
+            if self.alpha > self.min_alpha_yet_reached:
```
janpom: Indented one extra level.

gensim/models/base_any2vec.py
```diff
@@ -0,0 +1,781 @@
...
+        if not hasattr(model, 'train_count'):
+            model.train_count = 0
+            model.total_train_time = 0
+        return model
+
+    def _get_keyedvector_instance(self):
+        raise NotImplementedError
+
+    def _set_params_from_kv(self):
+        self.trainables.vectors = self.wv.__dict__.get('vectors', [])
```
janpom: `getattr(self.wv, 'vectors', [])` more common pattern.

gensim/models/doc2vec.py
janpom: No, this is actually good. We have changed the `sentences` argument to `documents`. It might be that the user has: `Doc2Vec(sentences=...)` in their old code, in which case the `sentences` would fall into `**kwargs` and we would just ignore it. This would result in a mysterious bug for the user. It's good that we raise this exception.

gensim/models/doc2vec.py
```diff
@@ -55,249 +55,578 @@
...
-        concatenation of the context window word vectors (rather than a sum or average).
 
-        Called internally from `Doc2Vec.train()` and `Doc2Vec.infer_vector()`.
+        if 'sentences' in kwargs:
+            raise DeprecationWarning(
+                "Parameter 'sentences' was renamed to 'documents', and will be removed in 4.0.0, "
+                "use 'documents' instead."
+            )
+
+        if 'iter' in kwargs:
```
janpom: Is there a specific reason why we accept `iter` and silently map it to `epochs`, but we do not accept `sentences`? I don't like the inconsistency.

gensim/models/doc2vec.py
```diff
@@ -55,249 +55,578 @@
...
+            mapfile_path=docvecs_mapfile, dm=dm, dm_concat=dm_concat, dm_tag_count=dm_tag_count,
+            vector_size=self.vector_size, **trainables_kwargs)
+
+        self.wv = Word2VecKeyedVectors()
+        self.docvecs = docvecs or Doc2VecKeyedVectors()
+
+        self.comment = comment
+        if documents is not None:
+            if isinstance(documents, GeneratorType):
+                raise TypeError("You can't pass a generator as the sentences argument. Try an iterator.")
```
janpom: sentences -> documents

gensim/models/doc2vec.py
```diff
@@ -55,249 +55,578 @@
...
+
+        """
+        return self.docvecs.init_sims(replace=replace)
+
+    @classmethod
+    def load(cls, *args, **kwargs):
+        try:
+            return super(Doc2Vec, cls).load(*args, **kwargs)
+        except AttributeError:
+            # from IPython.core.debugger import set_trace
```
janpom: Pls remove dead code.

gensim/models/doc2vec.py
```diff
@@ -313,33 +642,88 @@ def note_doctag(self, key, document_no, document_length):
...
+class Doc2VecKeyedVectors(BaseKeyedVectors):
+    def __init__(self):
+        self.doctags = {}  # string -> Doctag (only filled if necessary)
+        self.max_rawint = -1  # highest rawint-indexed doctag
+        self.offset2doctag = []  # int offset-past-(max_rawint+1) -> String (only filled if necessary)
+        self.count = 0
+        self.vectors_docs = []
+        self.mapfile_path = None
+
+    @property
```
janpom: These seem to be for backwards compatibility, right? If yes, best to add `@deprecated` annotations. We decide, in which gensim version the deprecated code will be removed and make that information a part of the deprecation message. @menshikh-iv, thoughts?

gensim/models/fasttext.py
```diff
@@ -522,110 +405,560 @@ def train(self, sentences, total_examples=None, total_words=None,
...
+        self.vectors_vocab_norm = None
+        self.vectors_ngrams = None
+        self.vectors_ngrams_norm = None
+        self.ngrams = {}
+        self.hash2index = {}
+        self.ngrams_word = {}
+        self.min_n = 0
+        self.max_n = 0
+
+    @property
```
janpom: `@deprecated` annotations here as well

gensim/models/keyedvectors.py
```diff
@@ -23,7 +23,7 @@
 The vectors can also be instantiated from an existing file on disk
 in the original Google's word2vec C format as a KeyedVectors instance::
 
-  >>> from gensim.models.keyedvectors import KeyedVectors
+  >>> from gensim.models import KeyedVectors
```
janpom: So, what exactly `KeyedVectors` represents now and why is it moved outside of this module?

gensim/test/test_fasttext.py
```diff
@@ -325,65 +325,122 @@ def test_doesnt_match(self):
...
-        sims_wrapper = model_wrapper.most_similar('night', topn=10)
-        sims_wrapper_words = (list(map(lambda x: x[0], sims_wrapper)))  # get similar words
-
-        overlap_count = len(set(sims_gensim_words).intersection(sims_wrapper_words))
+    def test_cbow_hs_training(self):
 
-        # overlap increases as we increase `iter` value, min overlap set to 2 to avoid unit-tests taking too long
-        # this limit can be increased when using Cython code
-        self.assertGreaterEqual(overlap_count, 2)
+        model_gensim = FT_gensim(size=50, sg=0, cbow_mean=1, alpha=0.05, window=5, hs=1, negative=0,
```
janpom: indentation

gensim/test/test_fasttext.py
```diff
@@ -325,65 +325,122 @@ def test_doesnt_match(self):
...
+            u'kilometres',
+            u'in',
+            u'eight',
+            u'according',
+            u'flights',
+            u'during',
+            u'comes']
+        overlap_count = len(set(sims_gensim_words).intersection(expected_sims_words))
+        self.assertGreaterEqual(overlap_count, 3)
+        # self.assertEqual(sims_gensim_words, expected_sims_words)
```
janpom: no dead code please

gensim/callbacks.py
menshikh-iv: Need to add *args, **kwargs (because probably user wants to store something in this class, for example, for evaluation proposes.

gensim/models/base_any2vec.py
menshikh-iv: @janpom exactly same style in Russian institutes :)

gensim/models/base_any2vec.py
menshikh-iv: No, it has no sense to add "Returns" section if you return nothing @manneshiva .

gensim/models/doc2vec.py
menshikh-iv: Agree with you, I already mention it earlier https://github.com/RaRe-Technologies/gensim/pull/1777#discussion_r160505450

gensim/models/keyedvectors.py
menshikh-iv: @janpom this isn't moved, typically, we import all interesting stuff in `__init__.py` of module -> we can import `from gensim.models import KeyedVectors` instead of `from gensim.models.keyedvectors import KeyedVectors`

gensim/models/doc2vec.py
manneshiva: The original code accepts `iter` in `__init__` while it accepts `epochs` in `train`. I have used `epochs` as the primary attribute of the model and created an alias (`iter`) for `epochs`.  I also didn't change "not supporting" `sentences` from the original code. So basically, this inconsistency is caused due to maintaining backward compatibility in terms of parameters accepted by the `__init__` method.

gensim/models/base_any2vec.py
manneshiva: Is this a bad practice -- renaming method parameters in the subclass for better readability/clarity? (`data_iterable`->`documents` in `Doc2Vec` and `data_iterable`->`sentences` in `Word2Vec`).

gensim/models/base_any2vec.py
manneshiva: Made attribute assignments consistent according to [this](https://github.com/RaRe-Technologies/gensim/pull/1777#discussion_r160936086) comment.

gensim/models/base_any2vec.py
manneshiva: Yes, this is copy-pasted from old code.

gensim/models/base_any2vec.py
manneshiva: Logging at different stages of training cannot be handled using `Callback` instances because `callbacks` only have access to the `model`. They do not have access to other temporary variables used during training -- `cur_epoch`, `job_tally`, `time_elapsed` etc.

gensim/models/base_any2vec.py
manneshiva: We will have to decide whether to remove the `deprecated` methods in this PR. It is not as simple as just removing the `deprecated` methods from here. We will also have to update this changes in tutorials, notebooks and unittest cases -- as they use calls to these deprecated methods (`model.most_similar` is used almost everywhere instead of `model.wv.most_similar`).

menshikh-iv: Aha, OK, let's do it in different PR later.

gensim/models/fasttext.py
```diff
@@ -522,110 +405,550 @@ def train(self, sentences, total_examples=None, total_words=None,
...
+
+    def clear_sims(self):
+        """
+        Removes all L2-normalized vectors for words from the model.
+        You will have to recompute them using init_sims method.
+        """
+
+        self._clear_post_train()
+
+    @deprecated("Method will be removed in 4.0.0, use self.wv.__getitem__() instead")
```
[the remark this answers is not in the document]
manneshiva: Look at [comment](https://github.com/RaRe-Technologies/gensim/pull/1777#discussion_r161157185).

gensim/models/base_any2vec.py
menshikh-iv: But I'm confused yet because you already remove some deprecated stuff here, this is OK?
CC: @janpom

gensim/models/keyedvectors.py
manneshiva: This is in `~gensim.models.word2vec.Word2VecKeyedVectors` now. Should I shift this to `~gensim.models.keyedvectors.WordEmbeddingsKeyedVectors`. This shift would only be required if `FastText` models want to use this method. @menshikh-iv

gensim/models/doc2vec.py
manneshiva: This is again consistent with the old code. Do you want me to change it?

menshikh-iv: @manneshiva yes, but not in this PR I think. Need to collect all "suspicious" stuff in a list, create an issue & fix it after merging current PR.

gensim/models/fasttext.py
```diff
@@ -333,111 +282,43 @@ def build_vocab(self, sentences, keep_raw_vocab=False, trim_rule=None, progress_
...
+        self.trainables.vectors_vocab_norm = self.wv.__dict__.get('vectors_vocab_norm', None)
+        self.trainables.vectors_ngrams_norm = self.wv.__dict__.get('vectors_ngrams_norm', None)
+        self.trainables.ngrams = self.wv.__dict__.get('ngrams', {})
+        self.trainables.hash2index = self.wv.__dict__.get('hash2index', {})
+        self.vocabulary.min_n = self.wv.__dict__.get('min_n', None)
+        self.vocabulary.max_n = self.wv.__dict__.get('max_n', None)
+        self.vocabulary.ngrams_word = self.wv.__dict__.get('ngrams_word', None)
+
+    def _set_train_params(self, **kwargs):
+        pass
```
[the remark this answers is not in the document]
manneshiva: This is intended. The base class (`BaseAny2VecModel`) requires this to be implemented (used by `Word2Vec` and possibly useful for future `any2vec` models).

gensim/similarities/docsim.py
manneshiva: Reverting back this change.

gensim/test/test_fasttext.py
manneshiva: Correct and I have replaced them with similar test cases without the wrapper.

tox.ini
manneshiva: Since we are dropping support for pure-Python implementations, this check fails. It is required that we first run `python setup.py build_ext --inplace` before running this command.

gensim/callbacks.py
manneshiva: We can take out `__init__` and let the user define it.

gensim/models/base_any2vec.py
janpom: IMO, it would be best not to override the method but rather create another method. Maybe name the parent method `_build_vocab` and name the subclass method `build_vocab`. Then it's perfectly OK for the two methods to have different parameter names.

gensim/callbacks.py
janpom: @menshikh-iv Not sure what you mean. If anyone wanted to use `Callback`, they would subclass it and would add their own `__init__` with any parameters they need. I can't see how `*args, **kwargs` in the parent method would help.

PR #1780, opened 2017-12-12
Add Pivot Normalization for `gensim.models.TfidfModel`. Fix #220

markroxor:
Pivoted normalization for tfidf model fix #220 
- [x] Return a BOW (the test will fail until then).
- [x] Write a NB to compare this and old approach.
- [x] Write tests.

PR #1791, opened 2017-12-15
Add smart information retrieval system for `TfidfModel`. Fix #1785

### LINKED ISSUE #1785: Add smart information retrieval system for TFIDF
https://en.wikipedia.org/wiki/SMART_Information_Retrieval_System

The current [TFIDF](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/tfidfmodel.py) model uses natural TF and IDF for computing TFIDF. The idea is to try various transformation like logarithmic, augmented,boolean etc. before computing the vectors.

More about this - http://www.cs.odu.edu/~jbollen/IR04/readings/article1-29-03.pdf and https://nlp.stanford.edu/IR-book/pdf/06vect.pdf

Will send a PR tomorrow.

markroxor:
For more information check issue #1785.

Tests have failed locally because `lambda` functions cannot be serialized by `pickle`. 
Can't monkey patch regular functions either as it is not supported either.

TODO:
- [x] write backward compatibility tests.
- [x] Write docs-strings for each.


gensim/models/tfidfmodel.py
```diff
@@ -49,8 +52,8 @@ class TfidfModel(interfaces.TransformationABC):
     Model persistency is achieved via its load/save methods.
     """
 
-    def __init__(self, corpus=None, id2word=None, dictionary=None,
-                 wlocal=utils.identity, wglobal=df2idf, normalize=True):
+    def __init__(self, corpus=None, id2word=None, dictionary=None, smartirs="ntc",
+                 wlocal=None, wglobal=None, normalize=None):
```
menshikh-iv: Better to support backward compatibility, why you change default values?

Good solution - by default support default behavior and `smartirs=None`, but if user set `smartirs` - use it (and ignore wlocal, wglobal, ...)

gensim/models/tfidfmodel.py
```diff
@@ -11,24 +11,27 @@
...
+    if not isinstance(smartirs, str) or len(smartirs) != 3:
+        raise ValueError('Expected a string of length 3 except got ' + smartirs)
 
-      idf = add + log(totaldocs / doc_freq)
-    """
-    return add + math.log(1.0 * totaldocs / docfreq, log_base)
+    w_tf, w_df, w_n = smartirs
+
+    if w_tf not in 'nlabL':
+        raise ValueError('Expected term frequency weight to be one of nlabL, except got ' + w_tf)
```
menshikh-iv: better use `'nlabL'` instead `nlabL` (readability), same for `got ..`

gensim/models/tfidfmodel.py
```diff
@@ -78,10 +81,41 @@ def __init__(self, corpus=None, id2word=None, dictionary=None,
...
         self.id2word = id2word
-        self.wlocal, self.wglobal = wlocal, wglobal
+        self.wlocal, self.wglobal, self.normalize = wlocal, wglobal, normalize
         self.num_docs, self.num_nnz, self.idfs = None, None, None
+        n_tf, n_df, n_n = smartirs
+        self.smartirs = smartirs
+
+        if self.wlocal is None:
+            if n_tf == "n":
+                self.wlocal = lambda tf, mean=None, _max=None: tf
```
menshikh-iv: better to use simple define (instead of lambda) for avoiding pickle problems (here and everywhere)

gensim/models/tfidfmodel.py
```diff
@@ -127,11 +162,6 @@ def initialize(self, corpus):
 
         # and finally compute the idf weights
         n_features = max(dfs) if dfs else 0
-        logger.info(
```
menshikh-iv: why you remove this?

gensim/test/test_sklearn_api.py
```diff
@@ -498,7 +498,6 @@ def testPersistence(self):
         original_matrix = self.model.transform(original_bow)
         passed = numpy.allclose(loaded_matrix, original_matrix, atol=1e-1)
         self.assertTrue(passed)
-
     def testModelNotFitted(self):
```
menshikh-iv: Need to add more tests (for new functionality)

gensim/models/tfidfmodel.py
markroxor: * It is backward compatible. Please check the smartirs value. :)
* This can be followed but I refrained because I removed `df2idf` function totally because it is redundant. Are you sure I should add it back?

gensim/models/tfidfmodel.py
markroxor: This showed the progress of the `pre - computation` step. Now that the values are no longer being pre-computed I think we need to log this in a different way. How do you want it to go?

gensim/test/test_sklearn_api.py
markroxor: I have that in my checklist but before that I need to pass the already present tests.

gensim/models/tfidfmodel.py
```diff
@@ -78,10 +81,47 @@ def __init__(self, corpus=None, id2word=None, dictionary=None,
...
+                if n_tf == "n":
+                    return tf
+                elif n_tf == "l":
+                    return 1 + math.log(tf)
+                elif n_tf == "a":
+                    return 0.5 + (0.5 * tf / _max)
+                elif n_tf == "b":
+                    return 1 if tf > 0 else 0
+                elif n_tf == "L":
+                    return (1 + math.log(tf)) / (1 + math.log(mean))
```
piskvorky: What happens if the value is none of the enumerated options? Will the error make sense to the user?

gensim/models/tfidfmodel.py
```diff
@@ -144,17 +180,16 @@ def __getitem__(self, bow, eps=1e-12):
 
         # unknown (new) terms will be given zero weight (NOT infinity/huge weight,
         # as strict application of the IDF formula would dictate)
+
         vector = [
-            (termid, self.wlocal(tf) * self.idfs.get(termid))
-            for termid, tf in bow if self.idfs.get(termid, 0.0) != 0.0
+            (termid, self.wlocal(tf, mean=np.mean(np.array(bow), axis=1), _max=np.max(bow, axis=1)) * self.wglobal(self.dfs[termid], self.num_docs))
```
piskvorky: This looks wasteful (creating arrays, only to throw them away). What are the performance implications of these changes? Do you have a benchmark before/after?

gensim/models/tfidfmodel.py
markroxor: [`resolve_weights`](https://github.com/RaRe-Technologies/gensim/pull/1791/files#diff-896942304f9417b1b8842823b19d7431R19) take care of that.

gensim/models/tfidfmodel.py
markroxor: Changed the approach.

gensim/models/tfidfmodel.py
```diff
@@ -77,11 +95,69 @@ def __init__(self, corpus=None, id2word=None, dictionary=None,
         If `dictionary` is specified, it must be a `corpora.Dictionary` object
         and it will be used to directly construct the inverse document frequency
         mapping (then `corpus`, if specified, is ignored).
+
+        `smartirs` or SMART (System for the Mechanical Analysis and Retrieval of Text)
```
markroxor: Taken from wikipedia.

menshikh-iv: You can add concrete link to wiki page

gensim/models/tfidfmodel.py
```diff
@@ -6,22 +6,40 @@
...
 
 
+def resolve_weights(smartirs):
+    if not isinstance(smartirs, str) or len(smartirs) != 3:
+        raise ValueError('Expected a string of length 3 except got ' + smartirs)
+
+    w_tf, w_df, w_n = smartirs
+
+    if w_tf not in 'nlabL':
+        raise ValueError('Expected term frequency weight to be one of \'nlabL\', except got ' + w_tf + '\'')
```
menshikh-iv: nitpick: you should use `raise ValueError("...'nlabL'...")` for avoiding `\`.

gensim/models/tfidfmodel.py
```diff
@@ -6,22 +6,40 @@
...
+
 def df2idf(docfreq, totaldocs, log_base=2.0, add=0.0):
     """
     Compute default inverse-document-frequency for a term with document frequency `doc_freq`::
-
-      idf = add + log(totaldocs / doc_freq)
+    idf = add + log(totaldocs / doc_freq)
     """
-    return add + math.log(1.0 * totaldocs / docfreq, log_base)
+    return add + np.log(float(totaldocs) / docfreq) / np.log(2)
```
menshikh-iv: `log_base` doesn't used here

gensim/models/tfidfmodel.py
```diff
@@ -49,8 +67,8 @@ class TfidfModel(interfaces.TransformationABC):
     Model persistency is achieved via its load/save methods.
     """
 
-    def __init__(self, corpus=None, id2word=None, dictionary=None,
-                 wlocal=utils.identity, wglobal=df2idf, normalize=True):
+    def __init__(self, corpus=None, id2word=None, dictionary=None, wlocal=utils.identity,
+                 wglobal=df2idf, normalize=True, smartirs=None):
         """
         Compute tf-idf by multiplying a local component (term frequency) with a
```
menshikh-iv: Can you convert all docstrings in this file to numpy-style, according to my previous comment https://github.com/RaRe-Technologies/gensim/pull/1780#discussion_r157712393

gensim/models/tfidfmodel.py
```diff
@@ -77,11 +95,69 @@ def __init__(self, corpus=None, id2word=None, dictionary=None,
...
         """
-        self.normalize = normalize
+
         self.id2word = id2word
-        self.wlocal, self.wglobal = wlocal, wglobal
+        self.wlocal, self.wglobal, self.normalize = wlocal, wglobal, normalize
         self.num_docs, self.num_nnz, self.idfs = None, None, None
+        self.smartirs = smartirs
+
+        if self.normalize is True:
```
menshikh-iv: What happen if `self.normalize` isn't bool?

markroxor: In that case user if  `smartirs` is not `None`, `self.normalize` will be defined by the values of `smartirs`. If `smartirs` is `None` and  `self.normalize` is not a bool then `self.normalize` must be a function. Just like `wlocal` and `wglobal`.

gensim/models/tfidfmodel.py
```diff
@@ -6,51 +6,103 @@
...
 
 from gensim import interfaces, matutils, utils
 from six import iteritems
 
+import numpy as np
 
 logger = logging.getLogger(__name__)
 
 
+def resolve_weights(smartirs):
```
menshikh-iv: docstrings needed too (for all stuff here)

gensim/models/tfidfmodel.py
```diff
@@ -6,51 +6,103 @@
...
-    """Precompute the inverse document frequency mapping for all terms."""
+    """
+    Precompute the inverse document frequency mapping for all terms.
+    """
     # not strictly necessary and could be computed on the fly in TfidfModel__getitem__.
     # this method is here just to speed things up a little.
     return {termid: wglobal(df, total_docs) for termid, df in iteritems(dfs)}
 
 
+def wlocal_g(tf, n_tf):  # TODO rename it (to avoid confusion)
```
menshikh-iv: Don't forget about renaming

gensim/models/tfidfmodel.py
```diff
@@ -6,51 +6,103 @@
...
+    elif n_df == "p":
+        return np.log((1.0 * totaldocs - docfreq) / docfreq) / np.log(2)
+
+
+def normalize_g(x, n_n):  # TODO rename it (to avoid confusion)
+    if n_n == "n":
+        return x
+    elif n_n == "c":
+        return matutils.unitvec(x)
+    # TODO write byte-size normalisation
```
menshikh-iv: need to fix it too

gensim/models/tfidfmodel.py
```diff
@@ -65,23 +117,69 @@ def __init__(self, corpus=None, id2word=None, dictionary=None,
...
+                    Corpus is a list of sets where each set has two elements. First being the termid and
+                    second being the term frequency of each term in the document.
+        id2word :   dict
+                    id2word is an optional dictionary that maps the word_id to a token.
+                    In case id2word isn’t specified the mapping id2word[word_id] = str(word_id) will be used.
+        dictionary :corpora.Dictionary
+                    If `dictionary` is specified, it must be a `corpora.Dictionary` object
+                    and it will be used to directly construct the inverse document frequency
+                    mapping (then `corpus`, if specified, is ignored).
+        wlocals :   user specified function
```
menshikh-iv: Instead of
```
        wlocals :   user specified function
                         Default for `wlocal` is identity (other options: math.sqrt, math.log1p, ...)
```

should be
```
        wlocals : function, optional
            description of parameter, with links to different implementations (if function defined in gensim, link should be like :func:`~gensim.some.func`)
```

everywhere

gensim/models/tfidfmodel.py
```diff
@@ -65,23 +117,69 @@ def __init__(self, corpus=None, id2word=None, dictionary=None,
...
-        and default for `wglobal` is `log_2(total_docs / doc_freq)`, giving the
-        formula above.
 
-        `normalize` dictates how the final transformed vectors will be normalized.
-        `normalize=True` means set to unit length (default); `False` means don't
-        normalize. You can also set `normalize` to your own function that accepts
-        and returns a sparse vector.
+        Parameters
+        ----------
+        corpus :    dictionary.doc2bow
```
menshikh-iv: type should be `iterable of iterable of (int, int)`

gensim/models/tfidfmodel.py
```diff
@@ -65,23 +117,69 @@ def __init__(self, corpus=None, id2word=None, dictionary=None,
...
+                    Corpus is a list of sets where each set has two elements. First being the termid and
+                    second being the term frequency of each term in the document.
+        id2word :   dict
+                    id2word is an optional dictionary that maps the word_id to a token.
+                    In case id2word isn’t specified the mapping id2word[word_id] = str(word_id) will be used.
+        dictionary :corpora.Dictionary
```
menshikh-iv: type should be 
```
:class:`~gensim.corpora.dictionary.Dictionary`
```

gensim/models/tfidfmodel.py
```diff
@@ -65,23 +117,69 @@ def __init__(self, corpus=None, id2word=None, dictionary=None,
...
+                    Default for `wlocal` is identity (other options: math.sqrt, math.log1p, ...)
+        wglobal :   user specified function
+                    Default for `wglobal` is `log_2(total_docs / doc_freq)`, giving the
+                    formula above.
+        normalize : user specified function
+                    It dictates how the final transformed vectors will be normalized.
+                    `normalize=True` means set to unit length (default); `False` means don't
+                    normalize. You can also set `normalize` to your own function that accepts
+                    and returns a sparse vector.
+        smartirs : {'None' ,'str'}
```
menshikh-iv: str, optional

gensim/models/tfidfmodel.py
```diff
@@ -65,23 +117,69 @@ def __init__(self, corpus=None, id2word=None, dictionary=None,
...
+                    Term frequency weighing:
+                      natural - `n`, logarithm - `l` , augmented - `a`,  boolean `b`, log average - `L`.
+                    Document frequency weighting:
+                      none - `n`, idf - `t`, prob idf - `p`.
+                    Document normalization:
+                      none - `n`, cosine - `c`, byte size - `b`.
+
+                    for more information visit https://en.wikipedia.org/wiki/SMART_Information_Retrieval_System
+
+        Returns
```
menshikh-iv: `Returns` no needed in `__init__` method

gensim/models/tfidfmodel.py
markroxor: I think that `Checks for validity of smartirs parameter.` is enough. Do you have anything else in mind as well?

menshikh-iv: @markroxor need to add "Parameters" (type, description), "Raises" (type, reason), "Returns" (type, description)

gensim/models/tfidfmodel.py
```diff
@@ -6,51 +6,100 @@
...
+
 def df2idf(docfreq, totaldocs, log_base=2.0, add=0.0):
     """
     Compute default inverse-document-frequency for a term with document frequency `doc_freq`::
-
-      idf = add + log(totaldocs / doc_freq)
+    idf = add + log(totaldocs / doc_freq)
     """
-    return add + math.log(1.0 * totaldocs / docfreq, log_base)
+    return add + np.log(float(totaldocs) / docfreq) / np.log(log_base)
```
menshikh-iv: What's a reason to use `np.log` instead of `math.log`?

gensim/models/tfidfmodel.py
```diff
@@ -6,51 +6,100 @@
...
+        return 1 + np.log(tf) / np.log(2)
+    elif n_tf == "a":
+        return 0.5 + (0.5 * tf / tf.max(axis=0))
+    elif n_tf == "b":
+        return tf.astype('bool').astype('int')
+    elif n_tf == "L":
+        return (1 + np.log(tf) / np.log(2)) / (1 + np.log(tf.mean(axis=0) / np.log(2)))
+
+
+def updated_wglobal(docfreq, totaldocs, n_df):  # TODO rename it (to avoid confusion)
```
menshikh-iv: please remove `TODO` (about renaming)

gensim/test/test_tfidfmodel.py
```diff
@@ -66,6 +89,143 @@ def testPersistenceCompressed(self):
...
+        expected_docs = [model[docs[0]], model[docs[1]]]
+
+        self.assertTrue(np.allclose(transformed_docs[0], expected_docs[0]))
+        self.assertTrue(np.allclose(transformed_docs[1], expected_docs[1]))
+
+        # Testing all the variations of `wlocal`
+        # nnn
+        model = tfidfmodel.TfidfModel(self.corpus, smartirs='nnn')
+        transformed_docs = [model[docs[0]], model[docs[1]]]
+        expected_docs = [[(3, 2), (4, 2), (5, 3), (6, 2), (7, 3), (8, 2)],
```
menshikh-iv: please use hanging indents instead of vertical (here and everywhere)

gensim/test/test_tfidfmodel.py
```diff
@@ -58,6 +74,13 @@ def testPersistence(self):
         tstvec = []
         self.assertTrue(np.allclose(model[tstvec], model2[tstvec]))  # try projecting an empty vector
 
+        # Test persistence between old and new model.
+        model3 = tfidfmodel.TfidfModel(self.corpus, smartirs="ntc")
+        model4 = tfidfmodel.TfidfModel.load(datapath('tfidf_model.tst'))
+        self.assertTrue(model3.idfs == model4.idfs)
+        tstvec = []
+        self.assertTrue(np.allclose(model3[tstvec], model4[tstvec]))  # try projecting an empty vector
```
menshikh-iv: Check non-empty case too

gensim/test/test_tfidfmodel.py
```diff
@@ -66,6 +89,143 @@ def testPersistenceCompressed(self):
         self.assertTrue(model.idfs == model2.idfs)
         tstvec = []
         self.assertTrue(np.allclose(model[tstvec], model2[tstvec]))  # try projecting an empty vector
+
+        # Test persistence between old and new compressed model.
+        model3 = tfidfmodel.TfidfModel(self.corpus, smartirs="ntc")
+        model4 = tfidfmodel.TfidfModel.load(datapath('tfidf_model.tst.bz2'))
+        self.assertTrue(model3.idfs == model4.idfs)
+        tstvec = []
+        self.assertTrue(np.allclose(model3[tstvec], model4[tstvec]))  # try projecting an empty vector
```
menshikh-iv: same as in `testPersistence`

gensim/test/test_tfidfmodel.py
```diff
@@ -66,6 +89,143 @@ def testPersistenceCompressed(self):
         self.assertTrue(model.idfs == model2.idfs)
         tstvec = []
         self.assertTrue(np.allclose(model[tstvec], model2[tstvec]))  # try projecting an empty vector
+
+        # Test persistence between old and new compressed model.
+        model3 = tfidfmodel.TfidfModel(self.corpus, smartirs="ntc")
+        model4 = tfidfmodel.TfidfModel.load(datapath('tfidf_model.tst.bz2'))
```
menshikh-iv: It will be better if you add gensim version in model name ("old" in comment isn't really informative)

gensim/test/test_tfidfmodel.py
```diff
@@ -58,6 +74,13 @@ def testPersistence(self):
         tstvec = []
         self.assertTrue(np.allclose(model[tstvec], model2[tstvec]))  # try projecting an empty vector
 
+        # Test persistence between old and new model.
+        model3 = tfidfmodel.TfidfModel(self.corpus, smartirs="ntc")
```
menshikh-iv: also, need to test than models with `smartirs` save/load correctly

gensim/test/test_tfidfmodel.py
```diff
@@ -66,6 +89,143 @@ def testPersistenceCompressed(self):
...
+        model = tfidfmodel.TfidfModel(self.corpus, wlocal=lambda x: x, wglobal=lambda x, y: x * x, smartirs='nnc')
+
+        transformed_docs = [model[docs[0]], model[docs[1]]]
+
+        model = tfidfmodel.TfidfModel(self.corpus, wlocal=lambda x: x * x, wglobal=lambda x, y: x, smartirs='nnc')
+        expected_docs = [model[docs[0]], model[docs[1]]]
+
+        self.assertTrue(np.allclose(transformed_docs[0], expected_docs[0]))
+        self.assertTrue(np.allclose(transformed_docs[1], expected_docs[1]))
 # endclass TestTfidfModel
```
menshikh-iv: please remove this line `# ...`

gensim/models/tfidfmodel.py
markroxor: Consistency I am using `np` everywhere. Anyways I do'nt see any problem with that. Do you?

markroxor:
> @markroxor need to add "Parameters" (type, description), "Raises" (type, reason), "Returns" (type, description)

I don't think that it is required. As this is part of docstrings and this function is implicitly used and can not be invoked by the user.

Are you sure I should add docstrings?

gensim/models/tfidfmodel.py
menshikh-iv: No problem this is only a question :)

menshikh-iv:
@markroxor I understand what you mean, very desirable add docstrings for all stuff (this isn't only for public-api users, this is for contributors too), this isn't critical as public API, but desirable.

gensim/models/tfidfmodel.py
```diff
@@ -144,17 +262,27 @@ def __getitem__(self, bow, eps=1e-12):
...
+        for termid, tf in bow:
+            termid_array.append(termid)
+            tf_array.append(tf)
+
+        tf_array = self.wlocal(np.array(tf_array))
+
         vector = [
-            (termid, self.wlocal(tf) * self.idfs.get(termid))
-            for termid, tf in bow if self.idfs.get(termid, 0.0) != 0.0
+            (termid, tf * self.idfs.get(termid))
```
markroxor: @menshikh-iv can we have something like `(termid, self.normalize(tf * self.idfs.get(termid)))` instead? So that the user can initialize with something like `normalize=np.log` ? Because right now the self.normalize function should be like `lambda tf, wt : tf, np.log(wt)` .

gensim/models/tfidfmodel.py
```diff
@@ -144,17 +262,27 @@ def __getitem__(self, bow, eps=1e-12):
...
+        elif self.normalize is False:
+            self.normalize = utils.identity
+
         # and finally, normalize the vector either to unit length, or use a
         # user-defined normalization function
-        if self.normalize is True:
-            vector = matutils.unitvec(vector)
-        elif self.normalize:
-            vector = self.normalize(vector)
+        vector = self.normalize(vector)
```
markroxor: Instead of normalizing `vector` which is  set of termid  and tfidf. Should we normalize the array of tfidf instead? Which is 1D and can directly be used with some functions like `sklearn.preprocessing.normalize` ? Since vector comprise of `termid`, while writing custom functions user has to take care of it as well.

gensim/models/tfidfmodel.py
```diff
@@ -272,7 +272,7 @@ def __getitem__(self, bow, eps=1e-12):
 
         vector = [
             (termid, tf * self.idfs.get(termid))
-            for termid, tf in zip(termid_array, tf_array) if self.idfs.get(termid, 0.0) != 0.0
+            for termid, tf in zip(termid_array, tf_array) if self.idfs.get(termid, 0.0) > eps
```
piskvorky: Comparing in absolute value safer (easier to read and comprehend, plus less room for error).

markroxor: I had a discussion with @menshikh-iv stating otherwise. @menshikh-iv I am reverting this.

gensim/models/tfidfmodel.py
```diff
@@ -144,17 +262,27 @@ def __getitem__(self, bow, eps=1e-12):
...
+            termid_array.append(termid)
+            tf_array.append(tf)
+
+        tf_array = self.wlocal(np.array(tf_array))
+
         vector = [
-            (termid, self.wlocal(tf) * self.idfs.get(termid))
-            for termid, tf in bow if self.idfs.get(termid, 0.0) != 0.0
+            (termid, tf * self.idfs.get(termid))
+            for termid, tf in zip(termid_array, tf_array) if self.idfs.get(termid, 0.0) != 0.0
```
piskvorky: This is too brittle; compare floats for equality using `eps` instead: For our purposes, two floats are "equal" if their absolute difference is smaller than `eps`.

gensim/models/tfidfmodel.py
markroxor: @piskvorky I did the same here, but you asked me to revert it.

gensim/models/tfidfmodel.py
menshikh-iv: I'm confused, https://github.com/RaRe-Technologies/gensim/pull/1791#discussion_r160090398 and https://github.com/RaRe-Technologies/gensim/pull/1791#discussion_r158868854 contradict each other, what you mean @piskvorky?

piskvorky: To compare float for equality, use `abs(x - y) < eps`. Not `x == y`, nor `x - y < eps`.

PR #1801, opened 2017-12-19
Description of sg parameter in FastText mixed up

akutuzov:
It was erroneously stated that when sg=1, CBOW is used, otherwise skip-gram is used.
In fact, it is vice versa (quite logically, as sg=SkipGram).
Thus, the description should be fixed.


menshikh-iv:
Nice catch @akutuzov, thanks for your fix :+1:

PR #1804, opened 2017-12-19
Fix docstrings for `gensim.matutils`

Cheukting:
Will refine it according to changes request.


menshikh-iv:
Thank you for pull-request @Cheukting, I'll look into on next week, sorry for waiting.

PR #1805, opened 2017-12-19
Fix docstrings for `gensim.models.normmodel`

AustenLamacraft:
First attempt….


gensim/models/normmodel.py
```diff
@@ -13,34 +13,58 @@
...
+        Norm used to normalize. l1 and l2 norms are supported (l2 is default)
+
+    Methods
+    -------
+    __init__(corpus=None, norm='l2')
+        Normalizes the terms in the given corpus document-wise.
+    calc_norm(corpus):
+        Calculates the norm by calling matutils.unitvec with the norm parameter.
 
+    Examples:
```
anotherbugmaster: Redundant `:`

gensim/models/normmodel.py
```diff
@@ -68,8 +97,29 @@ def calc_norm(self, corpus):
...
     def __getitem__(self, bow):
+        """
+        Calls the self.normalize() method.
+
+        Parameters
+        ----------
+        bow : :class:`~interfaces.CorpusABC` (iterable of documents) or list
+        of (int, int).
+
+        Examples:
```
anotherbugmaster: Redundant `:`

gensim/models/normmodel.py
```diff
@@ -13,34 +13,58 @@
...
+    Model persistency is achieved via its load/save methods.
+
+    Parameters
+    ----------
+    corpus : iterable
+        Iterable of documents.
+    norm : {'l1', 'l2'}
+        Norm used to normalize. l1 and l2 norms are supported (l2 is default)
+
+    Methods
```
anotherbugmaster: No need for `Methods` section, there aren't so many methods here :)

menshikh-iv:
Thanks for PR @AustenLamacraft, I'll look into on next week, sorry for waiting.

gensim/models/normmodel.py
AustenLamacraft: OK

gensim/models/normmodel.py
AustenLamacraft: No examples required?

gensim/models/normmodel.py
```diff
@@ -68,8 +79,26 @@ def calc_norm(self, corpus):
         self.norms = norms
 
     def normalize(self, bow):
+        """
+        Normalizes a simple count representation.
```
piskvorky: PEP 257: docstrings should be commands: "do X", "return Y", instead of "does X", "returns Y".
(here and elsewhere in this PR)

menshikh-iv:
Need to resolve #1807 first (because we have formulas in this PR that should be rendered with latex)

menshikh-iv:
Thank you @AustenLamacraft :+1:  congratz with first contribution :fire:

PR #1811, opened 2017-12-23
Fix datatype of some TM. Partial fix #1576

### LINKED ISSUE #1576: Check what's a reason to use double-precision in TMs

Yuktichopra:


gensim/models/atmodel.py
```diff
@@ -65,11 +65,12 @@ class AuthorTopicState(LdaState):
...
-    def __init__(self, eta, lambda_shape, gamma_shape):
-        self.eta = eta
-        self.sstats = np.zeros(lambda_shape)
-        self.gamma = np.zeros(gamma_shape)
+    def __init__(self, eta, lambda_shape, gamma_shape,dtype=np.float64):
+        self.eta = eta.astype(dtype, copy=False)
+        self.sstats = np.zeros(lambda_shape, dtype)
+        self.gamma = np.zeros(gamma_shape, dtype)
         self.numdocs = 0
+        self.dtype = np.float64
```
menshikh-iv: probably you should use `dtype` that passed as parameter

gensim/models/atmodel.py
```diff
@@ -65,11 +65,12 @@ class AuthorTopicState(LdaState):
 
     """
 
-    def __init__(self, eta, lambda_shape, gamma_shape):
-        self.eta = eta
-        self.sstats = np.zeros(lambda_shape)
-        self.gamma = np.zeros(gamma_shape)
+    def __init__(self, eta, lambda_shape, gamma_shape,dtype=np.float64):
```
menshikh-iv: `,dt` -> `, dt`

gensim/models/hdpmodel.py
```diff
@@ -126,7 +127,7 @@ class HdpModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
     def __init__(self, corpus, id2word, max_chunks=None, max_time=None,
                  chunksize=256, kappa=1.0, tau=64.0, K=15, T=150, alpha=1,
                  gamma=1, eta=0.01, scale=1.0, var_converge=0.0001,
-                 outputdir=None, random_state=None):
+                 outputdir=None, random_state=None, dtype = float32):
```
menshikh-iv: maybe np.float32?

gensim/models/ldaseqmodel.py
```diff
@@ -421,7 +422,7 @@ def __getitem__(self, doc):
         """
         lda_model = ldamodel.LdaModel(num_topics=self.num_topics, alpha=self.alphas, id2word=self.id2word)
         lda_model.topics = np.array(np.split(np.zeros(self.vocab_len * self.num_topics), self.vocab_len))
-        ldapost = LdaPost(num_topics=self.num_topics, max_doc_len=len(doc), lda=lda_model, doc=doc)
+        ldapost = LdaPost(num_topics=self.num_topics, max_doc_len=len(doc), lda=lda_model, doc=doc, dtype = float32)
```
menshikh-iv: why float32?

gensim/models/atmodel.py
```diff
@@ -67,12 +67,16 @@ class AuthorTopicState(LdaState):
...
+        self.eta = eta.astype(dtype, copy=False)
+        self.sstats = np.zeros(lambda_shape, dtype)
+        self.gamma = np.zeros(gamma_shape, dtype)
         self.numdocs = 0
-        self.dtype = np.float64  # To be compatible with LdaState
+
+        self.dtype = np.float64
+
+        self.dtype = np.float32  # To be compatible with LdaState
```
menshikh-iv: What happens here (line 76 and 78)?

gensim/models/atmodel.py
```diff
@@ -207,9 +211,13 @@ def __init__(self, corpus=None, num_topics=100, id2word=None, author2doc=None, d
         ... corpus, num_topics=50, author2doc=author2doc, id2word=id2word, alpha='auto', eval_every=5)
 
         """
+
+        self.dtype = np.float64
+=======
```
menshikh-iv: oh, looks like bad resolving of merge conflict (here and everywhere)

Yuktichopra:
I did not understand this, please elaborate. For this model to be
compatible with ldamodel, and the class that is called in each of these
models, shoudn't I change it to float 64?

gensim/models/atmodel.py
Yuktichopra: My bad

menshikh-iv:
@Yuktichopra please make dtype configurable now (probably later we'll "pin" it), don't pin to `float64` or another type.

Also, check properly than dtype isn't change (tests + probably asserts in code)

Yuktichopra:
Shall I look for backward compatibility?

menshikh-iv:
@Yuktichopra of course!

menshikh-iv:
Ping @Yuktichopra, how is going?

menshikh-iv:
PR looks abandoned, for this reason, I close it.

PR #1819, opened 2017-12-27
Fix datatype parameter for `KeyedVectors.load_word2vec_format`. Fix #1682

### LINKED ISSUE #1682: `datatype` parameter in `load_word2vec_format` doesn't work as expected

pushpankar:


menshikh-iv:
Hi @pushpankar, thanks for your PR, what's about "other cases" from #1682? 
CC @jayantj, can you elaborate on additional cases?

menshikh-iv:
Ping @pushpankar, how is going?

pushpankar:
Sorry. I thought my work is done. Should I write a test script in `gensim/test `and check with different types of inputs like float16, float32, ints, etc?

menshikh-iv:
@pushpankar  for two types already enough, also, use example from original issue

gensim/test/test.kv.txt
```diff
@@ -0,0 +1,3 @@
+2 2
```
menshikh-iv: should be in `gensim/test/test_data`

gensim/test/test_datatype.py
```diff
@@ -0,0 +1,18 @@
...
+import numpy as np
+from gensim.models.keyedvectors import KeyedVectors
+import os
+
+
+class TestDataType(unittest.TestCase):
+    def test_datatype(self):
+        path = os.path.join(os.path.dirname(__file__), 'test.kv.txt')
+        kv = KeyedVectors.load_word2vec_format(path, datatype=np.float64)
+        self.assertEqual(kv['horse.n.01'][0], -0.0008546282343595379)
```
menshikh-iv: 1. You should use [assertAlmostEqual](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertAlmostEqual) instead of `assertEqual` for float comparison
2. You also need to check datatype

gensim/test/test_datatype.py
```diff
@@ -0,0 +1,18 @@
...
+import unittest
+
+import numpy as np
+from gensim.models.keyedvectors import KeyedVectors
+import os
+
+
+class TestDataType(unittest.TestCase):
+    def test_datatype(self):
+        path = os.path.join(os.path.dirname(__file__), 'test.kv.txt')
```
menshikh-iv: When you move `test.kv.txt` construction should look like

```python
from gensim.test.utils import datapath

path = datapath('test.kv.txt')
```

jayantj:
Hi @pushpankar thanks for the PR! Apart from the suggestions @menshikh-iv mentioned, it'd also be good to have a test for word2vec files in binary format.

gensim/test/test_datatype.py
```diff
@@ -0,0 +1,18 @@
+import logging
```
jayantj: Please add a file header, like in the other test files.

pushpankar:
Saving model(loaded with `datatype=np.float64`) in binary and then reloading it is raising `unicode` error. I looking for solution. Everything else is done.

pushpankar:
Saving model also causes loss in precision.

pushpankar:
@menshikh-iv Loading issue is solved for both text and binary format. But saving model in text format still causes precision loss. Should I fix that in this pull request only?

menshikh-iv:
I see many failed tests, @pushpankar please have a look into.

pushpankar:
@menshikh-iv I am working on it. Changing the way binary files are loaded is causing it.

gensim/models/keyedvectors.py
```diff
@@ -220,7 +220,7 @@ def add_word(word, weights):
                 result.index2word.append(word)
 
             if binary:
-                binary_len = dtype(REAL).itemsize * vector_size
+                binary_len = dtype(datatype).itemsize * vector_size
```
pushpankar: I am having problem detecting `binary_len` of vector saved with custom datatype. The only clue is that the next vector starts after a `" "` but before the space comes a string(also converted in python bytes) which can be of any length. @menshikh-iv any suggestion?

pushpankar: I tried changing `\n` at the end of each vector after each vector but that broke many other tests.

menshikh-iv: @pushpankar how it works if `datatype=REAL` here?

pushpankar: Since earlier float were only saved with only 32 bit precision, knowing the size of each vector in binary format was easy. Casting to lower precision is only done after loading vectors.
Please also note that in develop branch too, casting vector to lower precision and saving it in binary and then loading leads to some errors. This is because while loading float32 is being assumed but during saving it was saved with lower precision like float16. I am adding some code to make it more clear.
```
model = KeyedVectors.load_word2vec_format('./test_data/test.kv.txt', datatype=np.float16)
print(model['horse.n.01'][0])
model.save_word2vec_format('./test_data/test.kv.bin', binary=True)
model2 = KeyedVectors.load_word2vec_format('./test_data/test.kv.bin', datatype=np.float32, binary=True)
print(model2['horse.n.01'][0])
```
Gives 
```
  File "convert2binary.py", line 7, in <module>
    print(model2['horse.n.01'][0])
  File "/home/pushpankar/gensim/gensim/models/keyedvectors.py", line 326, in __getitem__
    return self.word_vec(words)
  File "/home/pushpankar/gensim/gensim/models/keyedvectors.py", line 453, in word_vec
    raise KeyError("word '%s' not in vocabulary" % word)
KeyError: "word 'horse.n.01' not in vocabulary"

```
This is because float32 was assumed while reading binary vector but originally it was saved with float16. Thus more than necessary bytes was read for every vector.
Let me know if I am not clear enough.

menshikh-iv: so, probably the easiest solution for this case is read/write with `REAL` type & cast it before the end of "load" process, wdyt @jayantj?

pushpankar:
This fixes:
1. Precision loss issue if model is saved in text format.
2. Crash issue during loading binary model with precision other than float32 but precision loss issue still persists if precision is higher than float32 during both loading and saving.

gensim/test/test_datatype.py
```diff
@@ -0,0 +1,30 @@
...
+
+from gensim.test.utils import datapath
+from gensim.models.keyedvectors import KeyedVectors
+
+
+class TestDataType(unittest.TestCase):
+    def test_text(self):
+        path = datapath('test.kv.txt')
+        kv = KeyedVectors.load_word2vec_format(path, binary=False,
+                                               datatype=np.float64)
```
menshikh-iv: What's about different datatypes?

gensim/models/keyedvectors.py
```diff
@@ -147,9 +147,10 @@ def save_word2vec_format(self, fname, fvocab=None, binary=False, total_vec=None)
             for word, vocab in sorted(iteritems(self.vocab), key=lambda item: -item[1].count):
                 row = self.syn0[vocab.index]
                 if binary:
+                    row = row.astype(REAL)
```
menshikh-iv: why this needed?

pushpankar: Because 
```
model = KeyedVectors.load_word2vec_format('./test_data/test.kv.txt', datatype=np.float16)
print(model['horse.n.01'][0])
model.save_word2vec_format('./test_data/test.kv.bin', binary=True)
model2 = KeyedVectors.load_word2vec_format('./test_data/test.kv.bin', datatype=np.float32, binary=True)
```
this causes crash.
This is another bug that exists in develop branch.
Steps to reproduce error:
1. load model with low precision.
2. save that model in binary
3. then finally try to load the binary model.

gensim/test/test_datatype.py
pushpankar: Will np.float16, np.float32, and np.float64 be enough?

menshikh-iv:
#1777 should be merged first

gensim/test/test_datatype.py
menshikh-iv: Yes

menshikh-iv:
Hello @pushpankar, can you resolve merge-conflict?

pushpankar:
I see that loading and saving word vectors has been added in `utils_any2vec.py `. I am making changes in both the files: `keyedvectors.py` and `utils_any2vec.py`.

menshikh-iv:
Good work @pushpankar, LGTM, wdyt @jayantj?

gensim/test/test_datatype.py
```diff
@@ -0,0 +1,52 @@
...
+
+import numpy as np
+
+from gensim.test.utils import datapath
+from gensim.models.keyedvectors import KeyedVectors
+
+
+class TestDataType(unittest.TestCase):
+    def load_model(self, datatype):
+        path = datapath('test.kv.txt')
```
jayantj: A slightly more descriptive name would be helpful, there's already a lot of test data and it can get easily confusing.

gensim/test/test_datatype.py
```diff
@@ -0,0 +1,52 @@
...
+        self.assertEqual(kv['horse.n.01'][0].dtype, np.float16)
+
+    def test_type_conversion(self):
+        path = datapath('test.kv.txt')
+        binary_path = datapath('test.kv.bin')
+        model1 = KeyedVectors.load_word2vec_format(path, datatype=np.float16)
+        model1.save_word2vec_format(binary_path, binary=True)
+        model2 = KeyedVectors.load_word2vec_format(binary_path, datatype=np.float64, binary=True)
+        self.assertAlmostEqual(model1["horse.n.01"][0], np.float16(model2["horse.n.01"][0]))
+
```
jayantj: Another test to verify that the `type` of a value in the loaded array matches the datatype passed to `load_word2vec_format` would explicitly confirm the new behaviour.

gensim/models/utils_any2vec.py
```diff
@@ -109,9 +109,10 @@ def _save_word2vec_format(fname, vocab, vectors, fvocab=None, binary=False, tota
             for word, vocab_ in sorted(iteritems(vocab), key=lambda item: -item[1].count):
                 row = vectors[vocab_.index]
                 if binary:
+                    row = row.astype(REAL)
```
jayantj: What is the reason for the two changes to the `save_word2vec_format` method?

gensim/models/keyedvectors.py
jayantj: Ah, thanks for investigating that and reporting it. Do you have any ideas why that might be occurring?

jayantj:
LGTM to me too (left some minor comments), thanks for the fix @pushpankar

menshikh-iv:
@pushpankar good work, congratz with the first contribution! :+1:

gensim/test/test_datatype.py
```diff
@@ -0,0 +1,54 @@
...
+import numpy as np
+
+from gensim.test.utils import datapath
+from gensim.models.keyedvectors import KeyedVectors
+
+
+class TestDataType(unittest.TestCase):
+    def load_model(self, datatype):
+        path = datapath('high_precision.kv.txt')
+        kv = KeyedVectors.load_word2vec_format(path, binary=False,
```
piskvorky: Hanging indent please (not vertical).

pushpankar: I guess you are talking about line 22-23. I have merged them.

PR #1821, opened 2017-12-28
Test and refactor WikiCorpus

steremma:
In this PR, I try to improve the test coverage of the corpora package, specifically by adding tests for the WikiCorpus class. This is necessary as I soon plan to submit additional non trivial functionality for the same class and some tests are needed to ensure that nothing is broken. I also performed minor style changes in accordance to PEP8.


gensim/test/test_corpora.py
```diff
@@ -400,6 +402,73 @@ def test_indexing(self):
         pass
 
 
+class TestWikiCorpus(TestTextCorpus):
```
menshikh-iv: One of your tests "stuck" in Travis, please check, what's a problem and fix https://travis-ci.org/RaRe-Technologies/gensim/jobs/322641221#L828

gensim/test/test_corpora.py
```diff
@@ -400,6 +402,73 @@ def test_indexing(self):
...
+
+    def test_load(self):
+        corpus = self.corpus_class(self.fname, article_min_tokens=0)
+
+        docs = list(corpus)
+        # the deerwester corpus always has nine documents
+        self.assertEqual(len(docs), 9)
+
+    def test_empty_input(self):
+        tmpf = get_tmpfile('emptycorpus.xml.bz2')
```
menshikh-iv: Better use `with temporary_file('emptycorpus.xml.bz2'): ...` from `gensim.test.utils` here (because we have auto-cleanup in this function.

gensim/test/test_corpora.py
```diff
@@ -400,6 +402,73 @@ def test_indexing(self):
...
+        content = bz2.compress(b'')  # Explicit string to byte conversion needed in python 3
+        fh = open(tmpf, "wb")
```
menshikh-iv: Instead of `open` / `close` please use `with open(...)`

gensim/test/test_corpora.py
menshikh-iv: Also, question: why you need `TextTextCorpus` here?

gensim/test/test_corpora.py
```diff
@@ -400,6 +402,73 @@ def test_indexing(self):
...
+        self.assertEqual(expected, first_text)
+
+    def test_len(self):
+
+        def test_with_limit(article_min_tokens, expected_articles):
+            corpus = self.corpus_class(self.fname, article_min_tokens=article_min_tokens)
+            all_articles = corpus.get_texts()
+            assert (len(list(all_articles)) == expected_articles)
+
+        test_with_limit(0, 9)
```
menshikh-iv: better to make it in "cycle" (instead of defining test in test)

gensim/test/test_corpora.py
steremma: I am inheriting so that we make sure the `WikiCorpus` class passes not only its own tests but also the one's defined for the `TextCorpus` class (since` WikiCorpus` inherits from `TextCorpus`). Else we can inherit from `CorpusTestCase` or even from `unittest.Testcase`

gensim/test/test_corpora.py
steremma: Ok I will change that

gensim/test/test_corpora.py
steremma: I tried to follow the same style as in the base test classes but you are right, context managers are safer.

menshikh-iv: We have a lot of "dirty" code in the codebase, cleanup slow but go forward.

menshikh-iv:
Ping @steremma, how is going with "stuck" bug? Are you already launched tests with a debugger?

steremma:
Hello Ivan. I’m getting back from vacation tomorrow and I will keep working on it. I will let you know on gitter if the problem persists

gensim/test/test_wikicorpus.py
```diff
@@ -1,135 +0,0 @@
...
-        self.assertTrue(u'collectivization' not in next(wc.get_texts()))
-
-    def test_max_token_len_set(self):
-        """
-        set the parameter token_max_len to 16 and check that 'collectivisation' as a token exists
-        """
-        wc = WikiCorpus(datapath(FILENAME), processes=1, token_max_len=16, lemmatize=False)
-        self.assertTrue(u'collectivization' in next(wc.get_texts()))
-
-    def test_custom_tokenizer(self):
```
menshikh-iv: This test completely missing, please restore.

gensim/test/test_wikicorpus.py
```diff
@@ -1,135 +0,0 @@
...
-    #         return
-    #
-    #     wc = WikiCorpus(datapath(FILENAME))
-    #     l = wc.get_texts()
-    #     self.assertEqual(type(l), types.GeneratorType)
-    #     first = next(l)
-    #     self.assertEqual(type(first), list)
-    #     self.assertTrue(isinstance(first[0], bytes) or isinstance(first[0], str))
-
-    def test_first_element(self):
```
menshikh-iv: This test missing too, please restore

gensim/test/test_corpora.py
```diff
@@ -400,6 +400,137 @@ def test_indexing(self):
...
+        self.assertTrue(u'collectivization' not in next(corpus.get_texts()))
+
+    def test_max_token_len_set(self):
+        """
+        Set the parameter token_max_len to 16 and check that 'collectivisation' as a token exists
+        """
+        corpus = self.corpus_class(self.enwiki, processes=1, token_max_len=16, lemmatize=False)
+        self.assertTrue(u'collectivization' in next(corpus.get_texts()))
+
+    # TODO: sporadic failure to be investigated
```
menshikh-iv: What's a problem here?

gensim/test/test_corpora.py
```diff
@@ -400,6 +400,137 @@ def test_indexing(self):
...
+    # def test_get_texts_returns_generator_of_lists(self):
+    #
+    #     corpus = self.corpus_class(self.fname)
+    #     l = corpus.get_texts()
+    #     self.assertEqual(type(l), types.GeneratorType)
+    #     first = next(l)
+    #     self.assertEqual(type(first), list)
+    #     self.assertTrue(isinstance(first[0], bytes) or isinstance(first[0], str))
+
+    def test_sample_text(self):
```
menshikh-iv: Probably better to skip this test (not silently pass)., what do you think @steremma?

gensim/test/test_corpora.py
steremma: This one exists commented on test_wikicorpus, I don't know why. So I decided to keep it in case we want to look into it in the future.

gensim/test/test_corpora.py
steremma: The problem is that this test overrides the one defined in `TestTextCorpus`. Since if we just skip it the parent definition will be called and it will fail. In that sense passing it silently practically ignores it. The same idea is followed in the tests:

```
    def test_save(self):
        pass

    def test_serialize(self):
        pass

    def test_serialize_compressed(self):
        pass

    def test_indexing(self):
        pass
```

of `TestTextCorpus` for these 4 tests defined in the parent class `CorpusTestCase`. So I think its better to keep it as it is.

gensim/test/test_wikicorpus.py
steremma: You are right adding it now

gensim/test/test_corpora.py
menshikh-iv: At least - leave in place.
Best variant - investigate & fix.

gensim/test/test_corpora.py
menshikh-iv: Best variant - change class hierarchy and interfaces, we should'nt give "useless" methods from parent class (change from `TestWikiCorpus -> TestTextCorpus` to 
`TestTextCorpus -> BaseTestTextCorpus <- TestWikiCorpus`), but right now this isn't really needed, stay current variant with "pass".

menshikh-iv:
@steremma LGTM, tell me, should I merge know, or you'll try to investigate a problem with https://github.com/RaRe-Technologies/gensim/pull/1821#discussion_r161023363 first and merge after the fix?

steremma:
I propose that we merge it now for 2 reasons:

1) I think the faulty test is not really Pythonic since it tries to enforce a type using `isinstance` (at least thats how I interpret "ask forgiveness not permission"). Since the behavior is correct (all tests pass) I don't think we should care about the exact types.

2) I would prefer to work on a new feature for now (interlinks) since the tests are at a good point. If in the future we really want to enforce type checks we can come back to it. For this reason lets keep it commented like you said.

menshikh-iv:
Great start @steremma, congratz with first contribution :1st_place_medal:

steremma:
Thanks a lot, it wouldn't be possible without your guidance. Looking forward to contribute more :)

mebrunet:
@steremma - Thanks for the refactor. However, because of [574134e](https://github.com/RaRe-Technologies/gensim/pull/1821/commits/574134e1a49411375d9a599079231934506cd867) the behavior of passing an empty Dict when constructing a WikiCorpus has changed. Calling:
```
wiki = WikiCorpus(wikidumpfile, dictionary={})
```
used to prevent a full pass over the corpus, returning instantly. Now it triggers `Dictionary(self.get_texts())`. The old behavior was useful when you only wanted to use `get_texts` without building a Dictionary, e.g. for training Word2Vec. There are also many tutorials online that have code snippets that assume the old behavior. 

I think it should be changed back for compatibility.

piskvorky:
Agreed, looks like an unintended change (bug).

steremma:
Thanks for pointing this out. I am currently out of base but I will check
it when I get the time (and hopefully fix)

mebrunet:
I don't have time to clone build and test right now... but it looks right to me. Thanks.

steremma:
Is this functionality common enough to get its own test? I would say it is, in which case it would be useful if you could provide the code that triggered the mistaken call to `get_texts()`. I can then wrap it into a unit test

mebrunet:
Sure. The unit test is easy.

spy on `get_texts`
construct a new WikiCorpus with `WikiCorpus(dumpfile, dictionary={})`
assert spy was not called

You could extend that to make sure the spy is called if `dictionary=None`,
though that may already be tested.


Sent while mobile. Please excuse brevity and typos.

On May 9, 2018 16:32, "Stergiadis Manos" <notifications@github.com> wrote:

Is this functionality common enough to get its own test? I would say it is,
in which case it would be useful if you could provide the code that
triggered the mistaken call to get_texts(). I can then wrap it into a unit
test

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/RaRe-Technologies/gensim/pull/1821#issuecomment-387866424>,
or mute the thread

PR #1825, opened 2018-01-04
Add Cython version of `MmReader`

arlenk:
Very naive version of mmreader in cython.

From some profiling, when simply looping through an uncompressed mm corpus, 60% of the time is spent splitting each line and converting the strings to ints (or floats).  For a compressed corpus, about 45% of the time is spent in splitting and converting to ints/floats.

This version simply performs those two steps in cython using sscanf.  The rest of the code is pretty much unchanged -- though still compiled in cython for simplicity.  This change has a decent impact on performance, especially for uncompressed files.  The table below shows some profiling results for looping through a few different corpuses in mm format (the tfidf corpus is used to make sure the code works with an mm file that has floats as well).

|               | num_docs | num_terms | compressed? | time (current) | time (new) | speed up |
|---------------|----------|-----------|:-----------:|:--------------:|:----------:|:--------:|
| enron         | 40k      | 28k       |      no     |      22.3      |     2.6    |    8.7   |
|               |          |           |     yes     |      37.3      |    14.4    |    2.6   |
| nytimes       | 300k     | 100k      |      no     |      419.3     |    49.2    |    8.5   |
|               |          |           |     yes     |      686.2     |    275.1   |    2.5   |
| text8         | 2k       | 250k      |      no     |      25.4      |     2.5    |   10.1   |
|               |          |           |     yes     |      41.9      |    17.0    |    2.5   |
| text8 (tfidf) | 2k       | 250k      |      no     |      30.0      |     5.5    |    5.4   |
|               |          |           |     yes     |      47.8      |    21.1    |    2.3   |

In theory, the rest of the code could be converted into c code as well, though that would require writing c code to handle all the different file types that the mmreader accepts (eg, compressed files) and from profiling results, that portion of the code only accounts for 10% (uncompressed files) to 30% (compressed files) of the time.

The real world impact of this change is much more modest than the results above (unless for some reason you are just looping through a corpus...), but I have seen time savings of 15% in some real world uses of LDA with this change.  Obviously the impact any user sees will depend on the size of their corpus and the type of model they are using.


piskvorky:
This is great! Speeding up the "basic blocks" (MmCorpus included) is a much needed direction.

@arlenk Are there any cons to this change?

@menshikh-iv shall we keep the old "pure-Python" version, or do we drop the slow versions and keep only the compiled code path? I'm +1 for dropping, now that we offer a good system for binary (pre-compiled) wheels to users.

arlenk:
> @arlenk Are there any cons to this change? What are the disadvantages, trade-offs?

Here are the cons I can think of:
- keeping both the python & cython versions means a fair amount of code duplication.  Anyone fixing bugs or adding features would have to remember to update both files
- keeping only the cython version, on the other hand, makes the code less accessible (in my opinion) to end users.  Not sure how often people will care about looking at (or changing) the internals of mmreader, but keeping this only in cython will make that harder.
- dealing with strings in c/cython always makes me worry about encoding issues.  Since mm files should only have numbers, though, I can't think of a reason why we would ever run into non ascii files, but maybe others can think potential encoding issues.

menshikh-iv:
@piskvorky I'm +1 for dropping too (because we have wheels right now), this has no sense to support 2 versions.

setup.py
```diff
@@ -256,9 +256,12 @@ def finalize_options(self):
         Extension('gensim.models.doc2vec_inner',
             sources=['./gensim/models/doc2vec_inner.c'],
             include_dirs=[model_dir]),
+        Extension('gensim.corpora._mmreader',
+            sources=['./gensim/corpora/_mmreader.c'],
+            include_dirs=[model_dir]),
```
menshikh-iv: Why `model_dir` here? This isn't related to `gensim/models`.

gensim/corpora/_mmreader.pyx
```diff
@@ -0,0 +1,175 @@
...
+    matrix at once (unlike scipy.io.mmread). This allows us to process corpora
+    which are larger than the available RAM.
+    """
+    cdef public input
+    cdef public bint transposed
+    cdef public int num_docs, num_terms, num_nnz
+
+    def __init__(self, input, transposed=True):
+        """
+        Initialize the matrix reader.
```
menshikh-iv: Please use numpy-style docstrings (here and everywhere).

gensim/corpora/_mmreader.pyx
```diff
@@ -0,0 +1,175 @@
...
+import logging
+
+cimport cython
+from libc.stdio cimport sscanf
+
+
+logger = logging.getLogger(__name__)
+
+
+cdef class MmReader(object):
```
menshikh-iv: This is almost same as a pure-python version (`sscanf` different), maybe something else?

arlenk: No, nothing else really.  It's a very trivial change.  Just a change to use sscanf instead of split() and int() and a few type definitions (eg, docid, termid, and previd in __iter__).

No doubt converting the actual file access to c would further improve performance, but that will require dealing with all the different file types that smart_open accepts.  Also, I don't think this is the main hotspot in __iter__.

For example, here's a line profiler output of the current (develop branch) code when reading a raw text file:

<pre>
Line #  % Time  Line Contents
==============================
   907     0.0          with utils.file_or_filename(self.input) as lines:
   908     0.0              self.skip_headers(lines)

   911     8.6              for line in lines:
   912    40.4                  docid, termid, val = utils.to_unicode(line).split()  # needed for python3
   913     7.0                  if not self.transposed:
   914                              termid, docid = docid, termid
   915                          # -1 because matrix market indexes are 1-based => convert to 0-based
   916    22.5                  docid, termid, val = int(docid) - 1, int(termid) - 1, float(val)
   917     6.7                  assert previd <= docid, "matrix columns must come in ascending order"
   918     6.4                  if docid != previd:

   931
   932     8.4                  document.append((termid, val,))  # add another field to the current document
   933
</pre>
The majority of time is spent splitting the line and converting the resulting strings to ints (lines 912 and 916).  sscanf handles both of these lines.  Another 20% or so is taken up by the testing self.transposed and comparing docid and previd (lines 913, 917, and 918).   Statically typing these variables in cython helps quite a bt too.

For a compressed text file, line 911 (for line in lines) takes up 33% of the time, but most of the time is still taken by line.split() and int().

arlenk: Here's the line profile output when reading a compressed file

<pre>
Line #  % Time  Line Contents
==============================
   907     0.0          with utils.file_or_filename(self.input) as lines:
   908     0.0              self.skip_headers(lines)

   911    31.5              for line in lines:
   912    30.6                  docid, termid, val = utils.to_unicode(line).split()  # needed for python3
   913     5.1                  if not self.transposed:
   914                              termid, docid = docid, termid
   915                          # -1 because matrix market indexes are 1-based => convert to 0-based
   916    17.0                  docid, termid, val = int(docid) - 1, int(termid) - 1, float(val)
   917     4.9                  assert previd <= docid, "matrix columns must come in ascending order"
   918     4.6                  if docid != previd:

   932     6.3                  document.append((termid, val,))  # add another field to the current document
</pre>

menshikh-iv:
@arlenk please update `develop` branch in your fork and merge fresh develop here (this needed for correct doc building).

arlenk:
@menshikh-iv, i updated my develop branch

If you want to drop the python only version, let me know, and I can update the PR to only include the cython version (perhaps in a gensim.corpora.mmreader module).

menshikh-iv:
@arlenk great, yes, please remove pure python version, and this LGTM, wdyt @piskvorky?

piskvorky:
What are the compatibility implications of this? Any changes needed in existing user code?

arlenk:
> What are the compatibility implications of this?

I don't think there are any compatibility implications.  It should be a drop in replacement for the current mmreader.

> Any changes needed in existing user code? Does it support the same range of inputs (compressed etc)?

Yes, the actual opening of the file and looping through the rows is still handled by python/smart_open.  The only part that is in c is "splitting" each line and converting the values to ints/floats.

I added a few more unittests for MmCorpus as well, including one with a compressed file.

setup.py
```diff
@@ -256,9 +256,11 @@ def finalize_options(self):
         Extension('gensim.models.doc2vec_inner',
             sources=['./gensim/models/doc2vec_inner.c'],
             include_dirs=[model_dir]),
+        Extension('gensim.corpora.mmreader',
+            sources=['./gensim/corpora/mmreader.c']),
```
menshikh-iv: Probably, "include_dirs" should be `gensim/corpora`

gensim/test/test_data/test_mmcorpus_with_index.mm
```diff
@@ -0,0 +1,30 @@
+%%MatrixMarket matrix coordinate real general
```
piskvorky: Can we also add test files in `mm.gz` and `mm.bz2` formats? Just to make sure stuff continues to work.

setup.py
arlenk: Thanks @menshikh-iv .  It looks like it is failing on the bold line below in the tox config:

<pre>
<b>python -c "from gensim.models.word2vec import FAST_VERSION; print(FAST_VERSION)"</b>
python setup.py build_ext --inplace
python -c "from gensim.models.word2vec import FAST_VERSION; print(FAST_VERSION)"
</pre>

This tries to import word2vec before any of the cython modules are compiled.  word2vec  imports gensim.corpor, which import mmreader.  Since I took out the python version of mmreader, this line is failing as the cython version isn't built yet (this happens on the next line of the tox config).

For now, I removed the bold line to confirm the build works.  If the plan is to only include the c/cython versions of code, I am not sure if the bold line above is still necessary.  If it is still necessary, then I can add a try/except to gensim.corpora.__init__, though that would only be used for this line in the tox file.

menshikh-iv: @arlenk aha, this means that now we can't use `FAST_VERSION` freely, this isn't good. 

Probably, we need to move this constant (because this needed for checking that all compiled successfully), to a zone where we didn't receive an exception if something goes wrong.

arlenk: Ok, I added back the FAST_VERSION check to the tox config, so the config is the same as before.

To make the test run, I had to add a try/except block to gensim.corpora.__init__ as that file tries to import MmCorpus.  Since this check happens before the c-extension is built, MmCorpus fails because there is no pure python MmReader any more.  The except block in the __init__ file just creates a dummy MmCorpus class so the code "works" (ie, the imports don't fail, but obviously the dummy classes will not work if used).

This is not a very elegant solution, but I can't think of a better option if that check in the tox file is necessary.  Maybe there is a different way to check if the compilation step failed?  Also, is the compilation step still necessary with a wheel distribution?

menshikh-iv: Yes, this is not a good solution.
Compilation step is a **mandatory** step for wheels, I agree, but we also have repo & developers who work with it.

As alternative, I propose to replace import in `tox.ini` to something like this `from gensim.models.word2vec_inner import FAST_VERSION`, i.e. direct import, this will be helpful for this case?

arlenk: > As alternative, I propose to replace import in tox.ini to something like this from gensim.models.word2vec_inner import FAST_VERSION, i.e. direct import, this will be helpful for this case?

That's a good idea, but unfortunately that fails too.  Importing anything under gensim.* implicitly runs gensim/__init__.py, which imports gensim.corpora, which imports mmreader (which doesn't exist yet at this point in the tox process).

Do you know what this line in the tox config is meant to test?

<pre>
<b>python -c "from gensim.models.word2vec import FAST_VERSION; print(FAST_VERSION)"</b>
python setup.py build_ext --inplace
python -c "from gensim.models.word2vec import FAST_VERSION; print(FAST_VERSION)"
pytest {posargs:gensim/test}
</pre>

It seems like it tests whether all the "pure python" code installed correctly, but I wonder if there is a different way to test that.  Or is that test necessary since either the following line (the setup.py line) or the pytest line will fail if the install was not good.

menshikh-iv: > Do you know what this line in the tox config is meant to test?

I added it when I write the first version of `tox.ini` for sanity checking only

menshikh-iv: Feel free to remove this (and also remove try/except block in python file)

arlenk: > Do you know what this line in the tox config is meant to test?
> 
> I added it when I write the first version of tox.ini for sanity checking only

ah, ok, that makes sense.  I will remove this line for now and drop the try/except. 

thanks

menshikh-iv:
Hello @arlenk, sorry for waiting, can you make last changes
- Merge fresh changes from `develop`
- Return python version of MmReader (that will be used for backward compatibility until major release).
- Return FAST_VERSION first check too (same reason).

and PR will be ready to merge.

gensim/corpora/ucicorpus.py
```diff
@@ -15,7 +15,7 @@
 from gensim import utils
 from gensim.corpora import Dictionary
 from gensim.corpora import IndexedCorpus
-from gensim.matutils import MmReader
```
menshikh-iv: Please move `MmReader` to `matutils`  (yes, this is wired, but needed for backward compatibility). We'll move it to `corpora` with major release.

setup.py
```diff
@@ -256,9 +256,11 @@ def finalize_options(self):
         Extension('gensim.models.doc2vec_inner',
             sources=['./gensim/models/doc2vec_inner.c'],
             include_dirs=[model_dir]),
+        Extension('gensim.corpora._mmreader',
+            sources=['./gensim/corpora/_mmreader.c']),
```
menshikh-iv: `include_dirs` no needed here?

gensim/corpora/ucicorpus.py
arlenk: ok done.  I kept the cython version as "gensim.corpora._mmreader".  Let me know if you want this renamed to gensim.corpora.mmreader or something else.

setup.py
arlenk: no, I don't think it's needed.  The other pyx/c files include model_dir because they reference each other, but this module doesn't need to reference any of the other cython files.

gensim/corpora/mmcorpus.py
```diff
@@ -19,21 +19,78 @@
 logger = logging.getLogger('gensim.corpora.mmcorpus')
 
 
-class MmCorpus(matutils.MmReader, IndexedCorpus):
+try:
+    # try to load fast, cythonized code if possible
+    from gensim.corpora._mmreader import MmReader
+    FAST_VERSION = 1
+    logger.info('Fast version of MmReader is being used')
```
menshikh-iv: No need to log anything here (and after, because this will be shown for each import of gensim at all (because we import all stuff in `__init__.py`)

gensim/corpora/mmcorpus.py
```diff
@@ -19,21 +19,78 @@
 logger = logging.getLogger('gensim.corpora.mmcorpus')
 
 
-class MmCorpus(matutils.MmReader, IndexedCorpus):
+try:
+    # try to load fast, cythonized code if possible
+    from gensim.corpora._mmreader import MmReader
+    FAST_VERSION = 1
```
menshikh-iv: No need `FAST_VERSION` here at all (this is connected with BLAS stuff, that isn't used here).

gensim/corpora/mmcorpus.py
```diff
@@ -19,21 +19,78 @@
 logger = logging.getLogger('gensim.corpora.mmcorpus')
 
 
-class MmCorpus(matutils.MmReader, IndexedCorpus):
+try:
```
menshikh-iv: Can you move this try/except section to `matutils.py` and restore all imports (this is simplest and clearest way)

gensim/test/test_corpora.py
```diff
@@ -229,9 +229,107 @@ def test_load(self):
...
+        self.corpus_class = mmcorpus.MmCorpus
+        self.corpus = self.corpus_class(datapath('test_mmcorpus_corrupt.mm'))
+        self.file_extension = '.mm'
+
+    def test_serialize_compressed(self):
+        # MmCorpus needs file write with seek => doesn't support compressed output (only input)
+        pass
+
+    def test_load(self):
+        self.assertRaises(ValueError, lambda: [doc for doc in self.corpus])
```
menshikh-iv: I don't catch, what is it?

gensim/test/test_corpora.py
```diff
@@ -215,10 +215,10 @@ def _get_slice(corpus, slice_):
             self.assertRaises(RuntimeError, _get_slice, corpus_, 1.0)
 
 
-class TestMmCorpus(CorpusTestCase):
+class TestMmCorpusWithIndex(CorpusTestCase):
```
menshikh-iv: what's about backward compatibility tests?

gensim/corpora/mmcorpus.py
arlenk: I tried moving the import to matutils, but it causes some issues with circular imports.

For example, if I put the following in matutils.py:

```python
try:
    # try loading cython version of mmreader
    from gensim.corpora._mmreader import MmReader
except ImportError:
    # else fallback to python version
    class MmReader(object):
        pass
```

When the try block tries to import gensim.corpora._mmreader, that executes gensim.corpora.\_\_init\_\_, which tries to import gensim.corpora.mmcorpus.MmCorpus, which unfortunately tries to import matutils.MmReader, which isn't defined yet (as we are still in the try block).

You end up with a exception like:

```python
----> 1 import gensim

~/Projects/gensim/gensim-mmreader/gensim/__init__.py in <module>()
      4 """
      5
----> 6 from gensim import parsing, matutils, interfaces, corpora, models, similarities, summarization, utils  # noqa:F401
      7 import logging
      8

~/Projects/gensim/gensim-mmreader/gensim/matutils.py in <module>()
     33 try:
     34     # try loading cython version of mmreader
---> 35     from gensim.corpora._mmreader import MmReader
   
~/Projects/gensim/gensim-mmreader/gensim/corpora/__init__.py in <module>()
      6 from .indexedcorpus import IndexedCorpus  # noqa:F401 must appear before the other classes
      7
----> 8 from .mmcorpus import MmCorpus  # noqa:F401
      9 from .bleicorpus import BleiCorpus  # noqa:F401
   
~/Projects/gensim/gensim-mmreader/gensim/corpora/mmcorpus.py in <module>()
     21
---> 22 class MmCorpus(matutils.MmReader, IndexedCorpus):
     23     """
   
AttributeError: module 'gensim.matutils' has no attribute 'MmReader'
```

I couldn't think of an easy way around this, short of moving _mmreader out of the corpora package, but I'll keep playing around with it.

gensim/test/test_corpora.py
arlenk: Are you asking specifically about TestMmCorpusCorrupt?  That's a test to confirm the code will raise an exception if the input file is malformed.  The new code checks that sscanf reads in 3 values, so I just test that and raise an exception if fewer values are read.

As for the other tests, I added tests for:
- the two compressed formats (gzip and bz2)
- to make sure you can access docs by index for a corpus with an index
- to check the values read back match what we expect (mostly, I was worried about floats getting truncated to ints)

gensim/test/test_corpora.py
arlenk: I made sure all the tests pass with the existing version of MmReader as well, so there shouldn't be any backward compatibility issues.

gensim/corpora/mmcorpus.py
arlenk: ok, will remove

gensim/corpora/mmcorpus.py
arlenk: sounds good will remove this too

gensim/corpora/mmcorpus.py
menshikh-iv: Oh, that's really sad, I understand what happens :( To avoid this, we need to move `_mmreader` from corpora (but I have no idea what's better place for it).

menshikh-iv:
@CLearERR one more task for you (documenting), add this to "docsim PR", you need to go through files & cleanup docstrings after I merge this PR:
- `gensim/corpora/_mmreader.pyx`
- `gensim/corpora/mmcorpus.py`
- `gensim/matutils.py` (`MmReader` & `MmWriter`)

Be accurate:
- Don't forget to fetch fresh upstream with current changes
- Don't forget to check how all rendered (especially for cython part).

menshikh-iv:
@arlenk great work! I fixed import-error, will merge this after CI if all fine, let's return to LDA cythonization PR.

menshikh-iv:
@arlenk great work, thank you very much :fire:!

piskvorky:
@arlenk how did you create the text8 `.mm` corpus? Can you please share it? (or the code for how you generated it from the raw `text8`)

arlenk:
> @arlenk how did you create the text8 .mm corpus? Can you please share it? (or the code for how you generated it from the raw text8)

Sure, here are the files (I had to gzip the index file as well so that github would allow me to attach it).
[text8.mm.gz](https://github.com/RaRe-Technologies/gensim/files/1773458/text8.mm.gz)
[text8.mm.index.gz](https://github.com/RaRe-Technologies/gensim/files/1773462/text8.mm.index.gz)

Here's the code I used for creating the files from the original text8 corpus
```python
import gensim.corpora as gc

import numpy as np

np.seterr(all='raise')

# load data
_corpus_path = '~/corpora/'

def save_corpus(name='text8'):
    texts = api.load(name)

    print("creating dictionary")
    dictionary = gc.Dictionary([[ word.lower() for word in text ] for text in texts ])
    outfile = os.path.join(_corpus_path, "{}.dict".format(name))
    print("saving dictionary to: {}".format(outfile))
    dictionary.save(outfile)

    print("converting to corpus")
    corpus = [ dictionary.doc2bow([ word.lower() for word in text]) for text in texts ]
    outfile = os.path.join(_corpus_path, "{}.mm".format(name))
    print("saving corpus to {}".format(outfile))
    gc.MmCorpus.serialize(outfile, corpus)
```

arlenk:
and here's the code for the tfidf version.  I created this version just to make sure there were no issues with floating point numbers in the input file.

```python
import gensim.models as gm
import gensim.corpora as gc
corpus = gc.MmCorpus('~/corpora/text8.mm',)
tfidf = gm.TfidfModel(corpus)
corpus_tfidf = [ tfidf[doc] for doc in corpus ]
gc.MmCorpus.serialize('~/corpora/text8.tfidf.mm', corpus_tfidf)
```

piskvorky:
Thanks a lot @arlenk !

PR #1709, opened 2017-11-12
[WIP] Refactor documentation API Reference for gensim.summarization

yurkai:
Fix [#1668](https://github.com/RaRe-Technologies/gensim/issues/1668).

PR #1826, opened 2018-01-04
Documentation for Montemurro and Zanette algorithm

PeteBleackley:
Added the Montemurro and Zanette algorithm to the API docs


PeteBleackley:
Just realised this has already been done.

piskvorky:
Oh, has it? How is that possible?

PeteBleackley:
It seems to have been done in https://github.com/RaRe-Technologies/gensim/pull/1709

piskvorky:
Aha. Is there anything here that is not there?

In other words, could we take the best of both PRs?

PeteBleackley:
I don't think there's anything in this one that isn't in #1709

PR #1827, opened 2018-01-06
Implement Soft Cosine Measure

Witiko:
## Introduction
I implemented the Soft Cosine Measure&nbsp;(SCM)&nbsp;[[wiki][], [1][], [2][]] as a part of research for my thesis&nbsp;[[3][]]. Although the original algorithm&nbsp;[[1][]] has a time complexity that is quadratic in the document length, I implemented a linear-time approximative algorithm that I sketch in [[3][], sec.&nbsp;4.4]. Since Gensim was such an indispensable asset in my work, I thought I would give back and contribute code. The implementation is showcased in [a jupyter notebook][softcossim_notebook] on corpora from the [SemEval 2016][semeval16] and [2017][semeval17] competitions.

 [wiki]: https://en.wikipedia.org/wiki/Cosine_similarity#Soft_cosine_measure (Soft cosine measure)
 [1]: http://www.scielo.org.mx/pdf/cys/v18n3/v18n3a7.pdf (Soft Similarity and Soft Cosine Measure: Similarity of Features in Vector Space Model)
 [2]: http://www.aclweb.org/anthology/S17-2051 (Simbow at semeval-2017 task 3: Soft-cosine semantic similarity between questions for community question answering)
 [3]: https://github.com/witiko-masters-thesis/thesis/blob/master/main.pdf (Vector Space Representations in Information Retrieval)
 [semeval16]: http://alt.qcri.org/semeval2016/ (International Workshop on Semantic Evaluation 2016)
 [semeval17]: http://alt.qcri.org/semeval2017/ (International Workshop on Semantic Evaluation 2017)

soft_cosine_tutorial

## Description
My original implementation closely followed the Gensim implementation of the Word Mover's Distance&nbsp;(WMD), which is split into a [`gensim.models.keyedvectors.EuclideanKeyedVectors.wmdistance`][keyedvectors] method that takes two token lists and computes the WMD for them, and into the [`gensim.similarity.WdmSimilarity`][docsim] class that provides batch similarity queries. However, I was not quite happy with this for the following reasons:

 [keyedvectors]: https://github.com/witiko/gensim/blob/softcossim/gensim/models/keyedvectors.py
 [docsim]: https://github.com/witiko/gensim/blob/softcossim/gensim/similarities/docsim.py
 [matutils]: https://github.com/witiko/gensim/blob/softcossim/gensim/matutils.py

1. Not all useful term similarity matrices are constructed using word embeddings. Therefore, putting the entire logic into [`gensim.models.keyedvectors.EuclideanKeyedVectors`][keyedvectors] immediately seemed like a bad idea that would hinder further extensions.
2. By automatically converting token lists into bag-of-words representation behind the curtain, the user is unable to apply document length normalization methods such as tf-idf.

For the above reasons, I ultimately decided to split the implementation into a function, a method, and a class as follows:

1. The [`gensim.matutils.softcossim`][matutils] function takes two documents in the bag-of-words representation, a sparse term similarity matrix in the scipy CSC format, and computes SCM.
2. The [`gensim.models.keyedvectors.EuclideanKeyedVectors.similarity_matrix`][keyedvectors] method takes a corpus of bag-of-words vectors, a dictionary, and produces a sparse term similarity matrix `Mrel` described by Charlet and Damnati, 2017&nbsp;[[1][]].
3. The [`gensim.similarities.SoftCosineSimilarity`][docsim] class takes a corpus of bag-of-words vectors, a sparse term similarity matrix in the scipy CSC format, and provides batch SCM queries against the corpus.

The above design achieves a much looser coupling between the individual components and eliminates the original concerns. I demonstrate the implementation in [a jupyter notebook][softcossim_notebook] on the corpus of [Yelp reviews][yelp]. The approximative linear-time approximative algorithm for SCM achieves about the same speed as the linear-time approximative algorithm for WMD (see [the corresponding jupyter notebook][wmd_notebook]).

 [softcossim_notebook]: https://github.com/witiko/gensim/blob/softcossim/docs/notebooks/soft_cosine_tutorial.ipynb (Finding similar documents with Word2Vec and Soft Cosine Measure)
 [wmd_notebook]: https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/WMD_tutorial.ipynb (Finding similar documents with Word2Vec and WMD)
 [yelp]: https://www.yelp.com/dataset/download

## Future work
The [`gensim.similarities.SoftCosineSimilarity`][docsim] class goes over the entire corpus and computes the SCM between the query and each document separately by calling [`gensim.matutils.softcossim`][matutils]. If performance is a concern, SCM can be computed in a single step by computing `q^T * S * C`, where `q` is the normalized query vector, `S` is the term similarity matrix, `C` is the normalized term-document matrix of the corpus, and “normalized” in this context stands for a vector `v` being divided by `sqrt(v^T * S * v)`. This is similar to what e.g. the [`gensim.similarity.MatrixSimilarity.get_similarities`][docsim] method does, only with the basic cosine similarity rather than SCM.

 [scipy_parallel]: https://groups.google.com/forum/#!topic/gensim/BVEtU6m-QVI

## References
1. Grigori Sidorov et al. *Soft Similarity and Soft Cosine Measure: Similarity of Features in Vector Space Model*, 2014. ([link to PDF][1])
2. Delphine Charlet and Geraldine Damnati, *SimBow at SemEval-2017 Task 3: Soft-Cosine Semantic Similarity between Questions for Community Question Answering*, 2017. ([link to PDF][2])
3. Vít Novotný, *Vector Space Representations in Information Retrieval* (preprint), 2017. ([link to PDF][3])


Witiko:
I added numpy-style documentation and unit tests. Hopefully, the code should be good to go now.

gensim/matutils.py
```diff
@@ -464,6 +464,82 @@ def cossim(vec1, vec2):
...
+        return scipy.sparse.coo_matrix((data, (row, col)), shape=(similarity_matrix.shape[0], 1),
+                                       dtype=similarity_matrix.dtype)
+
+    def softdot(vec1, vec2):
+        vec1 = vec1.tocsr()
+        vec2 = vec2.tocsc()
+        return (vec1.T).dot(similarity_matrix).dot(vec2)[0, 0]
+
+    if not isinstance(similarity_matrix, scipy.sparse.csc_matrix):
+        if isinstance(similarity_matrix, scipy.sparse.csr_matrix):
```
menshikh-iv: What's about other formats (sparse, but not `csc` or `csr`)?

gensim/matutils.py
```diff
@@ -464,6 +464,82 @@ def cossim(vec1, vec2):
...
+    vec1 = sparse2coo(vec1)
+    vec2 = sparse2coo(vec2)
+    vec1len = softdot(vec1, vec1)
+    vec2len = softdot(vec2, vec2)
+    assert vec1len > 0.0 and vec2len > 0.0, u"sparse documents must not contain any explicit zero" \
+            " entries and the similarity matrix S must satisfy x^T * S * x > 0 for any nonzero" \
+            " bag-of-words vector x."
+    result = softdot(vec1, vec2)
+    result /= math.sqrt(vec1len) * math.sqrt(vec2len)  # rescale by vector lengths
+    if result > 1.0:
```
menshikh-iv: This is for avoiding problems with `float` (can be `1.0000003` or something similar)?

gensim/similarities/docsim.py
```diff
@@ -563,6 +564,107 @@ def __str__(self):
...
+        The Soft Cosine Similarity.
+
+    Examples
+    --------
+    >>> from gensim.models import Word2Vec
+    >>> from gensim.similarities import SoftCosineSimilarity
+    >>> from gensim.utils import simple_preprocess
+    >>> from gensim.corpora import Dictionary
+    >>> # Given a document collection "corpus", train a word2vec model.
+    >>> model = Word2Vec(corpus, workers=3, size=100)
```
menshikh-iv: all variables should be defined (this is really important for examples). You can give small corpus from `gensim.test.utils` or use, for example, this one

```python
corpus = api.load("text8")
```

gensim/similarities/docsim.py
```diff
@@ -563,6 +564,107 @@ def __str__(self):
...
+        self.num_best = num_best
+        self.chunksize = chunksize
+
+        # Normalization of features is undesirable, since soft cosine similarity requires special
+        # normalization using the similarity matrix. Therefore, we would just be normalizing twice,
+        # increasing the numerical error.
+        self.normalize = False
+
+        # index is simply an array from 0 to size of corpus.
+        self.index = numpy.array(range(len(corpus)))
```
menshikh-iv: `numpy.arange(len(corpus))`

menshikh-iv: Also, should corpus **always** have `__len__`? 

For example, if corpus something like this
```python
class MyCorpus(object):
    def __init__(self, fname):
        self.fname = fname
    def __iter__(self):
        with smart_open(self.fname) as infile:
            for line in infile:
                yield line
```

gensim/similarities/docsim.py
```diff
@@ -563,6 +564,107 @@ def __str__(self):
...
+
+        # Remove the columns of the similarity matrix that correspond to terms outside corpus.
+        nonzero_columns = sorted(set((index for document in corpus for index, _ in document)))
+        identity_matrix = scipy.sparse.identity(similarity_matrix.shape[0],
+                                                dtype=similarity_matrix.dtype, format="csr")
+        with warnings.catch_warnings():
```
menshikh-iv: Why this block needed?

gensim/models/keyedvectors.py
```diff
@@ -559,6 +560,90 @@ def similar_by_vector(self, vector, topn=10, restrict_vocab=None):
...
+            if w1 not in self.vocab:
+                continue  # A word from the dictionary not present in the word2vec model.
+            # Traverse upper triangle columns.
+            if len(dictionary) <= nonzero_limit + 1:  # Traverse all columns.
+                columns = ((w2_index, self.similarity(w1, dictionary[w2_index]))
+                           for w2_index in range(w1_index + 1, num_rows)
+                           if w1_index != w2_index and dictionary[w2_index] in self.vocab)
+            else:  # Traverse only columns corresponding to the embeddings closest to w1.
+                num_nonzero = similarity_matrix[w1_index].getnnz() - 1
+                columns = ((dictionary.token2id[w2], similarity)
```
menshikh-iv: Please use hanging indents (instead of vertical).

gensim/models/keyedvectors.py
```diff
@@ -559,6 +560,90 @@ def similar_by_vector(self, vector, topn=10, restrict_vocab=None):
...
+                                  self.most_similar(positive=[w1],
+                                                    topn=nonzero_limit - num_nonzero))
+                           if w2 in dictionary.token2id and w1_index < dictionary.token2id[w2])
+                columns = sorted(columns, key=lambda x: x[0])
+            for w2_index, similarity in columns:
+                assert w1_index < w2_index
+                # Ensure that we don't exceed `nonzero_limit` by mirroring the upper triangle.
+                if similarity > threshold and similarity_matrix[w2_index].getnnz() <= nonzero_limit:
+                    element = similarity**exponent
+                    similarity_matrix[w1_index, w2_index] = element
```
menshikh-iv: `similarity_matrix` is symmetrical, maybe better to store only "half" of this matrix -> reduce memory usage twice?

gensim/models/keyedvectors.py
```diff
@@ -559,6 +560,90 @@ def similar_by_vector(self, vector, topn=10, restrict_vocab=None):
...
+        # Traverse rows.
+        num_rows = len(dictionary)
+        for w1_index in range(num_rows):
+            if w1_index % ceil(num_rows / 10) == 0:
+                logger.info("PROGRESS: at row %i / %i", w1_index + 1, num_rows)
+            w1 = dictionary[w1_index]
+            if w1 not in self.vocab:
+                continue  # A word from the dictionary not present in the word2vec model.
+            # Traverse upper triangle columns.
+            if len(dictionary) <= nonzero_limit + 1:  # Traverse all columns.
```
menshikh-iv: `if num_rows` instead of `len(dictionary)` ?

gensim/models/keyedvectors.py
```diff
@@ -559,6 +560,90 @@ def similar_by_vector(self, vector, topn=10, restrict_vocab=None):
         """
         return self.most_similar(positive=[vector], topn=topn, restrict_vocab=restrict_vocab)
 
+    def similarity_matrix(self, corpus, dictionary, threshold=0.0, exponent=2.0,
```
menshikh-iv: `corpus` isn't used

gensim/test/test_similarities.py
```diff
@@ -155,12 +169,7 @@ def testPersistency(self):
             return
 
         fname = get_tmpfile('gensim_similarities.tst.pkl')
-        if self.cls == similarities.Similarity:
-            index = self.cls(None, corpus, num_features=len(dictionary), shardsize=5)
-        elif self.cls == similarities.WmdSimilarity:
-            index = self.cls(texts, self.w2v_model)
-        else:
-            index = self.cls(corpus, num_features=len(dictionary))
+        index = self.factoryMethod()
```
menshikh-iv: are you sure that this an equivalent?

menshikh-iv:
@Witiko also please merge `develop` from upstream to your branch.

gensim/matutils.py
```diff
@@ -464,6 +464,82 @@ def cossim(vec1, vec2):
...
+    def sparse2coo(vec):
+        col = [0] * len(vec)
+        row, data = zip(*vec)
+        return scipy.sparse.coo_matrix((data, (row, col)), shape=(similarity_matrix.shape[0], 1),
+                                       dtype=similarity_matrix.dtype)
+
+    def softdot(vec1, vec2):
+        vec1 = vec1.tocsr()
+        vec2 = vec2.tocsc()
+        return (vec1.T).dot(similarity_matrix).dot(vec2)[0, 0]
```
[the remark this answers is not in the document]
Witiko: The shape is `(1, 1)`, i.e. a single scalar value packed as a sparse matrix.

gensim/matutils.py
Witiko: These would incur a performance hit, because we first need to convert these to `csc` / `csr`. We can either issue a warning and try to convert anything remotely meaningful to the `csc` / `csr` format, which will take a time linear in the number of nonzero entries in `similarity_matrix` and temporarily hold two copies of the (potentially huge) matrix, or we can just refuse anything that is in the wrong format. In the code, I chose the latter approach.

gensim/models/keyedvectors.py
Witiko: I agree that such a saving would be nice, but there seems to be no support in SciPy for a dot product of a symmetrical matrix that only stores the upper / lower triangle (`(vec1.T).dot(similarity_matrix).dot(vec2)[0, 0]`). Even beyond SciPy, I don't know of a sparse matrix format that would allow this kind of access (i.e. both row-wise and column-wise efficient access).

gensim/similarities/docsim.py
Witiko: This was taken straight from the `WmdSimilarity` class. Note that we also assume that a corpus has `__getitem__` (https://github.com/Witiko/gensim/blob/fc31b659f0d635c827e505b6accb17fd3f8b1925/gensim/similarities/docsim.py#L640).

gensim/test/test_similarities.py
Witiko: Yes, this exact if-then-else statement currently appears at six places in `test_similarities.py` at `gensim/develop`. Rather than adding two new lines for `SoftCosineSimilarity` at each of the five locations, I decided to refactor it into a factory method.

gensim/similarities/docsim.py
Witiko: As stated in the comment, the intent is to „remove the columns of the similarity matrix that correspond to terms outside corpus“. This will create a smaller matrix with zeroed-out columns that never appear in the corpus and that will therefore never be accessed. Due to the deficiencies in the SciPy dot product implementation, the subsequent calls to `gensim.matlibs.softcossim` will also be faster with such a matrix; an optimized dot product code would not access these columns either way.

gensim/models/keyedvectors.py
menshikh-iv: Sad but true, thanks for the clarification.

gensim/matutils.py
Witiko: It is.

docs/notebooks/soft_cosine_tutorial.ipynb
```diff
@@ -0,0 +1,613 @@
...
+    "       'Ffhe2cmRyloz3CCdRGvHtA', # Earl of Sandwich\n",
+    "       'K7lWdNUhCbcnEvI0NhGewg', # Wicked Spoon\n",
+    "       'eoHdUeQDNgQ6WYEnP2aiRw', # Serendipity 3\n",
+    "       'RESDUcs7fIiihp38-d6_6g', # Bacchanal Buffet\n",
+    "       '2weQS-RnoOBhb1KsHKyoSQ'] # The Buffet\n",
+    "\n",
+    "w2v_corpus = []  # Documents to train word2vec on (all 6 restaurants).\n",
+    "scs_corpus = []  # Documents to run queries against (only one restaurant).\n",
+    "documents = []  # scs_corpus, with no pre-processing (so we can see the original documents).\n",
+    "with open('/data/review.json') as data_file:\n",
```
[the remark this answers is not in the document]
Witiko: I don't think we can, the [license](https://s3-media2.fl.yelpcdn.com/assets/srv0/engineering_pages/e926cc12796d/assets/vendor/yelp-dataset-license.pdf) seems to explicitly forbid any distribution of the dataset (see section C). The dataset is also used in the Word Mover's Distance notebook, but from the description, it contained different data back when the notebook was created, so it seems the dataset is continuously changing.

Witiko:
I will look into these, thank you for taking the time to do the review.

gensim/test/test_similarities.py
Witiko: Should I make this a separate pull request?

docs/notebooks/soft_cosine_tutorial.ipynb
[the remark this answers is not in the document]
Witiko: Should I update both the WMD and SCS notebooks to use some open dataset available in `gensim-data`?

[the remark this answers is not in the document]
menshikh-iv: @Witiko for this case, I see two possible solutions:
- Leave as is (because Yelp dataset is really good for this task, but we can't store it in gensim-data), probably add link, when user can download this dataset
- Choice other datasets (that will be really nice for demonstration + possible to add it to gensim-data by license). @Witiko have you good candidates for it?

gensim/test/test_similarities.py
menshikh-iv: No (because this is really small change).

Witiko:
I used SemEval 2015–2017 Task 3 Subtask B question answering datasets in
my thesis. These are also a good fit for the semantic similarity task
and are permissively licensed, so it should be possible to add them to
`gensim-data`.

docs/notebooks/soft_cosine_tutorial.ipynb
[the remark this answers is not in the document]
Witiko: I use the [SemEval 2015–2017 Task 3 Subtask B](http://alt.qcri.org/semeval2017/task3/) question answering datasets in my thesis. These are also a good fit for the semantic similarity task and are permissively licensed, so it should be possible to add them to `gensim-data`.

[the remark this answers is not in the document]
menshikh-iv: Nice, please replace Yelp dataset with SemEval + create an issue in `gensim-data` with needed info (guide: https://github.com/RaRe-Technologies/gensim-data#want-to-add-a-new-corpus-or-model), probably add other "tasks" from semeval will be really good too.

gensim/test/test_similarities.py
Witiko: Ok, I was more concerned with the fact that this pull request may take a little longer to finish up, whereas this single refactoring is independent and could be incorporated fast.

menshikh-iv:
ping @Witiko, how is going?

Witiko:
I am planning to push new commits this weekend.

docs/notebooks/soft_cosine_tutorial.ipynb
```diff
@@ -0,0 +1,613 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from time import time\n",
+    "start_nb = time()\n",
```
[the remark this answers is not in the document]
Witiko: Fixed in 08dea4ebf550153b71d581fa29831674b80a9669.

docs/notebooks/soft_cosine_tutorial.ipynb
```diff
@@ -0,0 +1,613 @@
...
+    "Let's take some sentences to compute the similarity between."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sentence_obama = 'Obama speaks to the media in Illinois'\n",
```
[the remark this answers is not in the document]
Witiko: Fixed in 621ed0d150bb938ab61f1ff72a1c48655eb7d8ee.

docs/notebooks/soft_cosine_tutorial.ipynb
```diff
@@ -0,0 +1,613 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Cell took 107.69 seconds to run.\n"
+     ]
+    }
+   ],
+   "source": [
+    "start = time()\n",
```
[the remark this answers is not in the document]
Witiko: Fixed in 8af5f67583d03deb7cfa088e6d93223e710a05d2.

docs/notebooks/soft_cosine_tutorial.ipynb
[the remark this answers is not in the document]
Witiko: I am running the evaluations. If all looks fine, I will submit a PR to `gensim-data`. Until then, please consider the current state of the Jupyter Notebook temporary.

docs/notebooks/soft_cosine_tutorial.ipynb
```diff
@@ -0,0 +1,613 @@
...
+    "print('Cell took %.2f seconds to run.' %(time() - start))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## References\n",
+    "\n",
+    "1. Grigori Sidorov et al. *Soft Similarity and Soft Cosine Measure: Similarity of Features in Vector Space Model*, 2014.\n",
```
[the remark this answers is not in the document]
Witiko: Added in 50a72741d105635409c37fd44bc40c9f392620cd.

gensim/matutils.py
```diff
@@ -464,6 +464,82 @@ def cossim(vec1, vec2):
...
+
+    References
+    ----------
+    Soft Cosine Similarity was perhaps first defined by [sidorovetal14]_.
+
+    .. [sidorovetal14] Grigori Sidorov et al., "Soft Similarity and Soft Cosine Measure: Similarity
+       of Features in Vector Space Model", 2014.
+    """
+
+    def sparse2coo(vec):
```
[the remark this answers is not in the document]
Witiko: I extracted the `sparse2coo` function to the global scope in 5e3973e9d8f37f9dcfdac962291b0c94021402b6. The existing `corpus2csc` function could be (mis)used for this purpose, but this yielded a significant slowdown i my tests.

gensim/models/keyedvectors.py
Witiko: Fixed in 08dea4ebf550153b71d581fa29831674b80a9669.

gensim/models/keyedvectors.py
Witiko: Fixed in effef71902ff9f0d67b79467c8411bbeae9c9445.

gensim/models/keyedvectors.py
Witiko: Fixed in e1eb7cd12452fa48bf4dfb98ee33d67c4ea47ed0 and c7f0ce176ebda8843b725c673f8c6f950fdc85ce.

gensim/models/keyedvectors.py
Witiko: Note that this knowledge can still be useful when storing and transmitting the similarity matrix.

gensim/similarities/docsim.py
Witiko: Fixed in 3cf154eb8c8cdbb3ff68ce5d67ac6ba014646130.

gensim/similarities/docsim.py
Witiko: The first concern has been fixed in 03a967b580e8c5f45a5cbe12ebabf75064ab0d10.

menshikh-iv:
Need to merge #1777 first (intersection by `KeyedVectors`)

Witiko:
As I mentioned above, I still need to update the corresponding Jupyter Notebook, so this PR is not yet ready to merge either. Do you have any estimate when #1777 might be merged into `develop`?

menshikh-iv:
@Witiko yes, I see. Please, don't forget about comparison (your soft-cosine VS different WMD implementations), I really want to see.

About estimates - tomorrow or the day after.

Witiko:
I have merged #1777 and I am waiting for the SemEval datasets to be merged ([#18](https://github.com/RaRe-Technologies/gensim-data/issues/18)) into gensim-data, so that I can push the final Jupyter notebook. 

Preliminary evaluation results for the Soft Cosine Measure compared against the Gensim implementation of Word Mover's Distance, the Source{d} implementation of Word Mover's Distance ([src-d/wmd-relax](https://github.com/src-d/wmd-relax)), and basic cosine similarity are given below:

```
Dataset		Technique	MAP score	Duration
==============================================================
2016-dev	softcossim	76.57		 1.18 sec
2016-dev	wmd-gensim	72.18		67.75 sec
2016-dev	cossim  	70.72		 3.15 sec
2016-dev	wmd-relax	67.33		 6.77 sec
--------------------------------------------------------------
2016-test	softcossim	77.29		 1.78 sec
2016-test	cossim  	76.45		 4.06 sec
2016-test	wmd-gensim	76.07		86.33 sec
2016-test	wmd-relax	74.41		 9.89 sec
--------------------------------------------------------------
2017-test	softcossim	45.99		 1.24 sec
2017-test	wmd-relax	44.48		12.22 sec
2017-test	cossim  	44.22		 4.39 sec
2017-test	wmd-gensim	44.08		98.29 sec
--------------------------------------------------------------
```

I am at loss as to why the html builds keep on failing:

```
~/gensim/docs/src $ make clean html
…
Warning, treated as error:
/home/circleci/gensim/.tox/docs/lib/python2.7/site-packages/gensim/models/keyedvectors.py:docstring of gensim.models.keyedvectors.Word2VecKeyedVectors.similarity_matrix:44:Duplicate explicit target name: "charletdamnati17".
Makefile:37: recipe for target 'html' failed
```

I managed to place the error between the initial commit of #1777 (31943ae, good) and the final commit of #1777 (7a32739, bad). However, the 118 commits in between are difficult to bisect, especially since I do not know what I am looking for.

Witiko:
I currently do not know why the documentation fails to build:

```
~/gensim/docs/src $ make clean html
…
Warning, treated as error:
/home/circleci/gensim/.tox/docs/lib/python2.7/site-packages/gensim/models/keyedvectors.py:docstring of gensim.models.keyedvectors.Word2VecKeyedVectors.similarity_matrix:44:Duplicate explicit target name: "charletdamnati17".
Makefile:37: recipe for target 'html' failed
```

I managed to place the error between the initial commit of #1777 (31943ae, good) and the final commit of #1777 (7a32739, bad). However, the 118 commits in between are difficult to bisect, especially since I do not know what I am looking for and the documentation is broken in most.

Witiko:
To be more specific, the error message calls the reStructuredText citation `charletdamnati17` – a string that appears only once across the entire codebase – to be a duplicate. Adding the method along with its docstring to 31943ae causes no errors, whereas the same version of Sphinx fails to build the documentation at 7a32739. To me, this signals that some change in the configuration of the HTML generation has been made between 31943ae and 7a32739.

evanmiltenburg:
As an outsider who's interested in similarity metrics, could you also show how this metric compares to the others on [semantic textual similarity](http://alt.qcri.org/semeval2017/task1/)?

Witiko:
@evanmiltenburg That should be interesting; I will see if I can run a comparison by the end of the week.

thomasniebler:
@evanmiltenburg @Witiko I already worked on that :) Long story short: The soft cosine with respective weights can improve semantic word similarity (and I'm sure, also semantic textual similarity).
I learned the respective weights as a posdef matrix using a metric learning approach (seeing as this the soft cosine is actually the same) from evaluation datasets like WordSimilarity353 or MEN. If you are interested in the code, see here: https://thomasniebler.github.io/semantics-metriclearning/
The corresponding paper can be read at https://arxiv.org/abs/1705.07425 and has been published as a poster at the ISWC2017: https://www.thomas-niebler.de/pub/niebler2017rrl.pdf
If you have any questions regarding that, let me know.

evanmiltenburg:
@thomasniebler: Note that MEN has been criticized for not being a good measure of *similarity*. Rather, it measures *relatedness* between terms. See the [SimLex-999](https://www.cl.cam.ac.uk/~fh295/simlex.html) dataset for an alternative.

thomasniebler:
@evanmiltenburg You're correct. Indeed, improving relatedness was my first goal with this measure. However, since we're talking about skewing the vector space, learning on SimLex data should allow us to learn similarity. Maybe not as strict similarity scores as with other metric learning based algorithms (Retrofitting, Counterfitting, Paragram Embeddings), as those do not consider intensity scores, but still. Overall, you have a good point. I will consider this in future work.
Just as a side note: If needed, just try an algorithm from the https://github.com/metric-learn/ repositories. There should be plenty of metric learning algorithms that give you a matrix for the soft cosine that is learned on similarity/dissimilarity constraints. Technically, even the retrofitting/counterfitting/paragram algorithms are nothing different than metric learning algorithms applied to semantic tasks, as they learn a (nonlinear) transformation of the embedding vector space.

Naruto-Sasuke:
I wonder it can be used in patch similarity.

evanmiltenburg:
Well, the issue for me is not how to learn a model that can predict good similarity measures. The issue is that when people try soft cosine for themselves, without any transformations of the vector space, what will the model be good at? It's good to know this in advance, and it's important to be precise in your claims about this. (Don't want to give users the wrong impression and then disappoint them.)

The STS task is a decent intrinsic test to see whether the model can predict whether two sentences are similar. It's a nice addition to the (slightly more extrinsic) Community Question Answering task that the author of this issue already tested Soft Cosine Similarity on.

thomasniebler:
Well, technically, the soft cosine measure *is* a vector transformation, as it makes use of the dot product which can be easily parameterized by a quadratic matrix as e.g. done in the [Mahalanobis distance](https://en.wikipedia.org/wiki/Mahalanobis_distance). By using a parameterized similarity or distance measure, you already transforming the vector space, as metric spaces are defined by their metrics.
So basically, you rewrite the soft cosine as <a, Sb> / (sqrt(<a, Sa>) * sqrt(<b, Sb>)) and thus you transformed the vector space. If you don't want to transform your vector space, well, then you shouldn't use the soft cosine. But I admit that there is a difference in learning S and using a simple heuristic.

menshikh-iv:
@Witiko great work :+1:

Issue #1257, opened 2017-04-03
Monte Carlo Word Mover Distance 


robotcator:
This [repo](https://github.com/jwbaum91/MCWMD) have an implementation of Monte Carlo Word Movers Distance in Python with TensorFlow. Would it be integrated as a new feature in gensim?

gojomo:
What kind of tasks does it work well on? Is there a paper/evaluation highlighting its advantages?

robotcator:
@gojomo No, I just found it on a post and didn't find any paper about it.

tmylk:
It would be a very nice addition if it works faster than current Gensim's WMD

piskvorky:
@menshikh-iv this should be a nice Incubator project (the current WMD is somewhat hard to find and use -- Google doesn't even return Gensim at all!).

PR #1829, opened 2018-01-07
[WIP] Monte Carlo Word Mover Distance

kakshay21:
First prototype
https://github.com/RaRe-Technologies/gensim/issues/1257

### DIFF
--- a/docs/notebooks/WMD_new.ipynb
+++ b/docs/notebooks/WMD_new.ipynb
@@ -1,0 +1,275 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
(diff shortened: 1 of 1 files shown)


menshikh-iv:
@piskvorky author began implementation but in ~same time
- We give SoftCosineDistance (that's already pretty fast, significantly faster than WMD)
- We found that current distance will work fast only for GPU

for this reason, we switch project.

piskvorky:
OK, thanks.

Please keep the PRs and descriptions clean—this is where people come when they Google. Your internal discussions and decisions are invisible.

PR #1839, opened 2018-01-13
Add article interlinks to the output of `gensim.scripts.segment_wiki`. Fix #1712

### LINKED ISSUE #1712: Add interlinks to segment_wiki
### Idea
[Users ask about this feature](https://twitter.com/TheShubhanshu/status/929493213689335808), this is really useful to have [interlinks](https://en.wikipedia.org/wiki/Help:Link) in the dump to construct the graph of articles or use relation between articles in any way.

### What's need to implement
Add field `"section_interlinks" (list of str)` that contains a list of article titles referenced by this section.

steremma:
# Interlinks

This PR adds the feature requested in #1712 
The `segment_wiki` script now includes the interlinks (links to other articles in the dump) found in each article it processes. 

## Implementation

The interlink markup parsing takes place in the filtered text rather than the raw one in order to avoid code and logic duplication (similar markup would need to be removed again which is already done in `filter_wiki`). In order to achieve this the function `filter_wiki` must not promote the uncaught markup to plain text. The usefulness of this operation was already doubted (see comment in old version) so I removed the promotion altogether:

```
# the following is needed to make the tokenizer see '[[socialist]]s' as a single word 'socialists'
# TODO is this really desirable?
text = text.replace('[', '').replace(']', '')  
```

Removing this line was important because:

1. Phrases like "Computer Science" which has a meaning as a phrase will lose its meaning if its promoted to plain text (2 tokens)

2. the `[[` and `]]` are the markup used for interlinks. So if we promote them then we lose this information

Since the output of the script and its helper functions changed, I modified the documentation (source comments) accordingly.


gensim/corpora/wikicorpus.py
```diff
@@ -110,9 +110,6 @@ def remove_markup(text):
         if old == text or iters > 2:
             break
 
-    # the following is needed to make the tokenizer see '[[socialist]]s' as a single word 'socialists'
-    # TODO is this really desirable?
```
steremma: Indeed I don't think its desirable since it will split interlinks which should be represented by one token. I.e 
"Computer Science" fits the context better than "computer" and "science"

gensim/corpora/wikicorpus.py
```diff
@@ -69,7 +69,19 @@
 ]
 
 
-def filter_wiki(raw):
+def find_interlinks(raw):
+    """
+    Find all interlinks to other articles in the dump. `raw` is either unicode
+    or utf-8 encoded string.
+    """
+    interlink_regex_capture = r"\[{1,2}(.*?)\]{1,2}"
```
napsternxg: I think this regex can be compiled outside the function.

menshikh-iv:
@steremma please don't forget to
- resolve the merge conflict
- run `segment_wiki` on full wiki dump (with python2 and python3), this needs some time

menshikh-iv:
Looks great, thank you @steremma!

steremma:
👍

PR #1842, opened 2018-01-16
returning non-zero array with np.int input in matutils.unitvec 

o-P-o:
The zeros matrix is coming from the fact that, by nature, a unit vector is going to be less than one, so using vec.dtypes causes use of numpy.int, which rounds all values in the unit vector down to zero. It seems that the only way to stop this from happening (and at the same time maintaining the original bug fix for numpy.float dtypes) is casting before return in order to handle numpy.int types as and when is appropriate, which I have done here. What do you think? (This is my first time using git hub, so please excuse me if I've done something I'm not supposed to do.)

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -705,15 +705,8 @@ def unitvec(vec, norm='l2'):
         if norm == 'l2':
             veclen = blas_nrm2(vec)
         if veclen > 0.0:
-            '''EDIT: included .astype(vec.dtype)'''
-            if vec.dtype == 'int16':
-                vec = vec.astype(np.float16)
-                return blas_scal(1.0 / veclen, vec).astype(vec.dtype)
-            elif vec.dtype == 'int32':
-                vec = vec.astype(np.float32)
-                return blas_scal(1.0 / veclen, vec).astype(vec.dtype)
-            elif vec.dtype == 'int64':
-                vec = vec.astype(np.float64)
+            if np.issubdtype(vec.dtype, np.int) == True:
+                vec = vec.astype(np.float)
                 return blas_scal(1.0 / veclen, vec).astype(vec.dtype)
             else:
                 return blas_scal(1.0 / veclen, vec).astype(vec.dtype)
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -699,13 +699,24 @@ def unitvec(vec, norm='l2'):
             return vec
 
     if isinstance(vec, np.ndarray):
-        vec = np.asarray(vec, dtype=float)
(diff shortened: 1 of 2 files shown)


o-P-o:
travis-ci unsuccessful on Python 3.6

gensim/matutils.py
```diff
@@ -699,13 +699,24 @@ def unitvec(vec, norm='l2'):
             return vec
 
     if isinstance(vec, np.ndarray):
```
menshikh-iv: Please add tests for this function (checking that we normalized correctly & handle original data-type)

gensim/matutils.py
```diff
@@ -699,13 +699,24 @@ def unitvec(vec, norm='l2'):
...
-        vec = np.asarray(vec, dtype=float)
+        vec = np.asarray(vec, dtype=vec.dtype)
         if norm == 'l1':
             veclen = np.sum(np.abs(vec))
         if norm == 'l2':
             veclen = blas_nrm2(vec)
         if veclen > 0.0:
-            return blas_scal(1.0 / veclen, vec)
+            '''EDIT: included .astype(vec.dtype)'''
+            if vec.dtype == 'int16':
```
jayantj: For int-like inputs, IMO we don't necessarily have to match the dtype of the returned array to the corresponding float type (e.g. `int16` -> `float16`). We could simply return a `float` array when the input array is int-like.
It'd also be good to have a more robust way of checking if the input dtype is int-like instead of a hardcoded list of types.

o-P-o: Thanks, I'll work on that final point.

o-P-o: Hi, I edited the branch that I used for this to make changed to unitvec and present tests, but how do I get it to show up?

menshikh-iv: @o-P-o you should to push the code to `patch-1` branch in your fork.

menshikh-iv:
Ping @o-P-o, how is going?

o-P-o:
I've produced a solution and tests as well. I'm trying to get to grips with
using Git at the moment because I'm struggling to understand how to post
this (only recently started using it). If you want it asap I can PM or
email you my codes?

menshikh-iv:
@o-P-o no, we can wait.

You need to do several things
1. Checkout to current branch `git checkout patch-1`
2. Make needed changes and commit it (`git add gensim/matutils.py` and  `git commit -m "my commit message"`)
3. Push result to GitHub `git push origin patch-1`

Probably, you forgot about last step :)

menshikh-iv:
Continued in #1866

PR #1787, opened 2017-12-15
Bug Fix 1779

saroufimc1:
- Not sure about the header. 
- Keeping the whole embedding matrix for subwords, and only creating a dictionary for the subwords using the hashing function. In the previous version, the index was incorrectly assigned in the subword dictionary.

Issue #1779, opened 2017-12-12
Bucket bug for FastText.load_fasttext_format 


saroufimc1:
Hi,
I tried loading a pretrained model from Facebook's fasttext into gemsim using FastText.load_fasttext_format, and it looks like there is a bug to be fixed for bucket here as well. 

I noticed that for bucket = 2,000,000 we had model.wv.syn0_ngrams.shape[0] = 7,221,731 instead of 2,000,000. 

Also, in https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/wrappers/fasttext.py 

I don't see why we do have to compute the n-grams from the word vocabulary again. Aren't these already imported from the fastText .bin file? 

#### Steps/Code/Corpus to Reproduce

from gensim.models.wrappers import FastText 

## Then unzip and put the file 'wiki.en.bin' in your working directory.
model = FastText.load_fasttext_format('wiki.en')

print(model.wv.syn0_ngrams.shape)[0]

#### Expected Results
Expected value of 2,000,000 which is the default value of bucket. 

#### Actual Results
7,221,731 which is here equal to len(model.wv.ngrams).
In other words, it looks like there were no collusions although we had more n-grams than buckets.

Also, please note that it took 10 minutes to load the fasttext model. I wonder if some parts of the code (especially in load_vectors) are really needed.  

Thanks,
Carl

menshikh-iv:
Thanks for report @saroufimc1, we'll try to reproduce and look into with @manneshiva later (file is downloaded for a very long time)

jayantj:
Hi @saroufimc1 thanks for the detailed report and ideas. 
It is possible that `syn0_ngrams` should contain rows greater than the number of buckets. The reason for this is that the [C++ FastText model stores](https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc#L627) `num_words_in_vocab + bucket` vectors in its `input_` matrix, which corresponds to our `syn0_ngrams` matrix.

Is your vocab size 5,221,731? In that case, this is most likely not a bug.

Re: computing ngrams, no the FastText .bin file doesn't include the ngrams themselves. But you are right, it might be possible to get away without computing the ngrams from the vocab ourselves. Will post about it in more detail in #1261

jayantj:
Edit: I'm mistaken, this indeed is a bug.

manneshiva:
[...]

PR #1843, opened 2018-01-17
Fix 1779

saroufimc1:
Addressing issue #1779. Correct the assignment of model.wv.syn0_ngrams after trimming of unused ngrams (model.wv.syn0_ngrams.shape[0] <= self.bucket). 

It is 2x slower on pre-trained Wikipedia models from Facebook's fastText than if we do not trim unused ngrams. Therefore, I recommend dealing with #1261. 

For the test case, we need an actual small Facebook FastText model saved in bin format and make sure that we have model.wv.syn0_ngrams.shape[0] <= self.bucket.

Related to case #1787.


gensim/models/wrappers/fasttext.py
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Author: Jayant Jain <jayantjain1992@gmail.com>
```
menshikh-iv: Why you remove this?

menshikh-iv:
@saroufimc1 #1777 merged, please resolve merge conflict first

menshikh-iv:
Ping @saroufimc1, when you plan to finish this PR?

menshikh-iv:
Unfortunately, you merged conflict incorrectly (look at current state of `gensim/models/wrappers/fasttext.py` from `develop`), your code edition should be in https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/deprecated/fasttext_wrapper.py

Are you planning to finish PR?

saroufimc1:
@menshikh-iv  Sorry about that, I am not familiar with all the changes of the source code that happened in the meantime. Is it possible that you take my code and do the merging yourself since you are more familiar with the changes? It would be much faster I guess.

gensim/models/wrappers/fasttext.py
saroufimc1: Wasn't sure about the header. Feel free to take the code and do whatever you want with it.

menshikh-iv:
@saroufimc1 no problem, I close this with "almost complete" label, we will return to this later and fix it, thanks!

CC: @manneshiva

PR #1860, opened 2018-01-26
Fix symmetrical case for hellinger distance. Fix #1854

### LINKED ISSUE #1854: gensim.matutils.hellinger d(x,y) != d(y,x) if len(x) = len(y)
#### Description

Compute the distance between 2 distribution with gensim.matutils.hellinger. The d(x,y) do not equal to d(y,x) if len(x) = len(y)

from gensim.matutils import hellinger
vec_1 = [(2, 0.1), (3, 0.4), (4, 0.1), (5, 0.1), (1, 0.1), (7, 0.2)]
vec_2 = [(1, 0.1), (3, 0.8), (4, 0.1), (8, 0.1), (10, 0.8), (9, 0.1)]
hellinger(vec_1,vec_2) == hellinger(vec_2,vec_1)

#### Expected Results

True
#### Actual Results

False
#### Versions

Linux-4.13.0-26-generic-x86_64-with-debian-stretch-sid
('Python', '2.7.11 |Anaconda custom (64-bit)| (default, Dec  6 2015, 18:08:32) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]')
('NumPy', '1.13.3')
('SciPy', '1.0.0')
('gensim', '3.1.0')
('FAST_VERSION', 1)

caiyulun:
This changes fix the bugs that return different distance when we call hellinger(x, y) and hellinger(y,x) ( #1854)
The cause of this bug is that we compute the distance based on one distribution's index previously, but we should iterate all the index appears in two probability distributions.

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -897,9 +897,9 @@ def hellinger(vec1, vec2):
     if isbow(vec1) and isbow(vec2):
         # if it is a BoW format, instead of converting to dense we use dictionaries to calculate appropriate distance
         vec1, vec2 = dict(vec1), dict(vec2)
-        indexs = set(vec1.keys() + vec2.keys())
+        indices = set(list(vec1.keys()) + list(vec2.keys()))
         sim = np.sqrt(
-            0.5 * sum((np.sqrt(vec1.get(index, 0.0)) - np.sqrt(vec2.get(index, 0.0)))**2 for index in indexs)
+            0.5 * sum((np.sqrt(vec1.get(index, 0.0)) - np.sqrt(vec2.get(index, 0.0)))**2 for index in indices)
         )
         return sim
     else:
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -897,10 +897,9 @@ def hellinger(vec1, vec2):
     if isbow(vec1) and isbow(vec2):
         # if it is a BoW format, instead of converting to dense we use dictionaries to calculate appropriate distance
         vec1, vec2 = dict(vec1), dict(vec2)
-        if len(vec2) < len(vec1):
-            vec1, vec2 = vec2, vec1  # swap references so that we iterate over the shorter vector
+        indexs = set(vec1.keys() + vec2.keys())
         sim = np.sqrt(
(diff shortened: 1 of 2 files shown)


gensim/matutils.py
```diff
@@ -897,10 +897,9 @@ def hellinger(vec1, vec2):
     if isbow(vec1) and isbow(vec2):
         # if it is a BoW format, instead of converting to dense we use dictionaries to calculate appropriate distance
         vec1, vec2 = dict(vec1), dict(vec2)
-        if len(vec2) < len(vec1):
-            vec1, vec2 = vec2, vec1  # swap references so that we iterate over the shorter vector
+        indexs = set(vec1.keys() + vec2.keys())
```
menshikh-iv: `indexs` -> `indices`

gensim/matutils.py
```diff
@@ -897,10 +897,9 @@ def hellinger(vec1, vec2):
...
         sim = np.sqrt(
-            0.5 * sum((np.sqrt(value) - np.sqrt(vec2.get(index, 0.0)))**2 for index, value in iteritems(vec1))
+            0.5 * sum((np.sqrt(vec1.get(index, 0.0)) - np.sqrt(vec2.get(index, 0.0)))**2 for index in indexs)
```
menshikh-iv: Also please add a test for it (you can use an example from original issue).

caiyulun: Should I add a test in this function [https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_similarity_metrics.py#L106](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_similarity_metrics.py#L106) or anywhere else ？ And it seems that the current two different length BOW inputs test in [Line109](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_similarity_metrics.py#L109) has a wrong expected value, I'm tring to correct it.

menshikh-iv: Yes, please fix current test & add needed tests to `TestHellinger` class

menshikh-iv:
Good work @caiyulun, congratz with first contribution :+1:

PR #1863, opened 2018-01-30
Fix `score_function` from `LexicalEntailmentEvaluation`. Fix #1858

### LINKED ISSUE #1858: Wrong distance in Lexical entailment experiment for Poincare embeddings
The following line https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/poincare.py#L1522
should be:
```
return -1 * (1 + self.alpha * (norm_2 - norm_1)) * min_distance
```
otherwise `distance` is the last encountered distance, not the smallest one.

hachibaka:
Use min_distance in poincare score function.


menshikh-iv:
Congratz with first contribution @hachibaka  :1st_place_medal:

PR #1864, opened 2018-01-30
Update documentation for gensim.similarities.docsim

souravsingh:
Initial start at fixing the docstrings of docsim.py

### DIFF
--- a/gensim/similarities/docsim.py
+++ b/gensim/similarities/docsim.py
@@ -74,8 +74,7 @@ except ImportError:
 
 
 class Shard(utils.SaveLoad):
-    """
-    A proxy class that represents a single shard instance within a Similarity
+    """A proxy class that represents a single shard instance within a Similarity
     index.
 
     Basically just wraps (Sparse)MatrixSimilarity so that it mmaps from disk on
@@ -84,6 +83,14 @@ class Shard(utils.SaveLoad):
     """
 
     def __init__(self, fname, index):
+        """
+        Parameters
+        ----------
+        fname : str
+            Path to corpus.
+        index : int
+            Index of sytactic unit in corpus.
+        """
         self.dirname, self.fname = os.path.split(fname)
         self.length = len(index)
         self.cls = index.__class__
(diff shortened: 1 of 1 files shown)


gensim/similarities/docsim.py
```diff
@@ -74,8 +74,7 @@
 
 
 class Shard(utils.SaveLoad):
-    """
-    A proxy class that represents a single shard instance within a Similarity
+    """A proxy class that represents a single shard instance within a Similarity
```
menshikh-iv: 120 character limit (not 80)
Also, for definitions, first line of docstring **must** end with `.`

gensim/similarities/docsim.py
```diff
@@ -84,6 +83,14 @@ class Shard(utils.SaveLoad):
...
 
     def __init__(self, fname, index):
+        """
+        Parameters
+        ----------
+        fname : str
+            Path to corpus.
+        index : int
+            Index of sytactic unit in corpus.
+        """
```
menshikh-iv: missed `\n` (empty line for the end of each docstring).

gensim/similarities/docsim.py
```diff
@@ -92,22 +99,32 @@ def __init__(self, fname, index):
         self.index = self.get_index()
 
     def fullname(self):
+        """Return fullname of the corpus.
+        """
```
menshikh-iv: if docstring < 120 characters, please write it in one line: `"""Return fullname of the corpus."""`

menshikh-iv: also, try to avoid 'return' word, use 'get' for example (here and everywhere).

menshikh-iv: besides, missing `Returns` section (here and everywhere).

souravsingh: Okay, will do the fixes

menshikh-iv:
Ping @souravsingh, how is going with PR?

menshikh-iv:
Because nothing happens here, this task delegated to @CLearERR, I close this PR.

souravsingh:
@menshikh-iv I was busy with college work, so the whole thing went off my head. I will pick some other file to update if it is alright with you. 

Please let me know which file needs a docstrig update.

PR #1866, opened 2018-01-31
Fix type issue in `gensim.matutils.unitvec`. Fix #1722

### LINKED ISSUE #1722: matutils.unitvec returns vector with different dtype from input vector

o-P-o:
As requested, I have edited the fix to ignore dtype size. I use np.issubtype to check input type and handle appropriately before return to ensure non-integer output.

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -699,13 +699,17 @@ def unitvec(vec, norm='l2'):
             return vec
 
     if isinstance(vec, np.ndarray):
-        vec = np.asarray(vec, dtype=float)
+        vec = np.asarray(vec, dtype=vec.dtype)
         if norm == 'l1':
             veclen = np.sum(np.abs(vec))
         if norm == 'l2':
             veclen = blas_nrm2(vec)
         if veclen > 0.0:
-            return blas_scal(1.0 / veclen, vec)
+            if np.issubdtype(vec.dtype, np.int) == True:
+                vec = vec.astype(np.float)
+                return blas_scal(1.0 / veclen, vec).astype(vec.dtype)
+            else:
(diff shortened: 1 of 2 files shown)


gensim/test_unitvec
```diff
@@ -0,0 +1,33 @@
+import logging
```
menshikh-iv: 1. Should be `.py` file
2. Should be `gensim/test/test_matutils.py`

gensim/matutils.py
```diff
@@ -699,13 +699,17 @@ def unitvec(vec, norm='l2'):
             return vec
 
     if isinstance(vec, np.ndarray):
```
menshikh-iv: What's about other cases (when vectors is BoW or sparse)?

gensim/matutils.py
```diff
@@ -699,13 +699,17 @@ def unitvec(vec, norm='l2'):
...
-        vec = np.asarray(vec, dtype=float)
+        vec = np.asarray(vec, dtype=vec.dtype)
         if norm == 'l1':
             veclen = np.sum(np.abs(vec))
         if norm == 'l2':
             veclen = blas_nrm2(vec)
         if veclen > 0.0:
-            return blas_scal(1.0 / veclen, vec)
+            if np.issubdtype(vec.dtype, np.int) == True:
```
menshikh-iv: doesn't looks like a good idea

menshikh-iv:
ping @o-P-o, how is going? When you plan to finish this PR?

menshikh-iv:
ping @o-P-o

menshikh-iv:
Looks abandoned, I close this PR.

PR #1867, opened 2018-02-01
Remove outdated bz2 examples from tutorials

menshikh-iv:
MmReader support only `filename` as input (not `file-like object`), but in the old documentation (`wiki.rst`/`dist_lsi.rst`) we used `file-like object` too as input.

Current PR remove this outdated usage from examples.

Based on [mailing list post](https://groups.google.com/forum/#!topic/gensim/420pRBwexKQ)

### DIFF
--- a/docs/src/dist_lsi.rst
+++ b/docs/src/dist_lsi.rst
@@ -120,14 +120,13 @@ Distributed LSA on Wikipedia
 First, download and prepare the Wikipedia corpus as per :doc:`wiki`, then load
 the corpus iterator with::
 
-    >>> import logging, gensim, bz2
+    >>> import logging, gensim
     >>> logging.basicConfig(format='%(asctime)s : %(levelname)s : %(message)s', level=logging.INFO)
 
     >>> # load id->word mapping (the dictionary)
     >>> id2word = gensim.corpora.Dictionary.load_from_text('wiki_en_wordids.txt')
     >>> # load corpus iterator
     >>> mm = gensim.corpora.MmCorpus('wiki_en_tfidf.mm')
-    >>> # mm = gensim.corpora.MmCorpus(bz2.BZ2File('wiki_en_tfidf.mm.bz2')) # use this if you compressed the TFIDF output
 
     >>> print(mm)
     MmCorpus(3199665 documents, 100000 features, 495547400 non-zero entries)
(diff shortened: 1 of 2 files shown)


docs/src/dist_lsi.rst
```diff
@@ -120,14 +120,13 @@ Distributed LSA on Wikipedia
...
 
-    >>> import logging, gensim, bz2
+    >>> import logging, gensim
     >>> logging.basicConfig(format='%(asctime)s : %(levelname)s : %(message)s', level=logging.INFO)
 
     >>> # load id->word mapping (the dictionary)
     >>> id2word = gensim.corpora.Dictionary.load_from_text('wiki_en_wordids.txt')
     >>> # load corpus iterator
     >>> mm = gensim.corpora.MmCorpus('wiki_en_tfidf.mm')
-    >>> # mm = gensim.corpora.MmCorpus(bz2.BZ2File('wiki_en_tfidf.mm.bz2')) # use this if you compressed the TFIDF output
```
piskvorky: Removing this line doesn't sound right -- we still support `bz2`!

Just remove the (superfluous) `bz2.BZFile` wrapper. Dtto below.

piskvorky:
Why don't we support file-like objects in `MmCorpus`? That sounds like a bug, I see no reason.

docs/src/dist_lsi.rst
menshikh-iv: @piskvorky fixed in #1868

menshikh-iv:
@piskvorky I see code for support `file-like`, but also see some buggy behavior, look at example
```python
from gensim.corpora import MmCorpus
import bz2

f = bz2.BZ2File("testcorpus.mm.bz2")
print(f.closed)  # 0
corpus = MmCorpus(f)
print(f.closed)  # 1 ???
```
for this reason, if we try to read from this, we'll receive an exception suggested in mailing list.

menshikh-iv:
@piskvorky UPD, I found what's a reason for this behavior:
https://github.com/RaRe-Technologies/gensim/blob/5342153eb4f4b02bb45bfa3951eef8250ac9f6b6/gensim/matutils.py#L1274

in this line, we using `with` here, that automatically call `.close` when we exit from `with` scope, for this reason, after `Mmreader.__init__` we closed passed file-descriptor.

This is a bug anyway (because internally we use `file_or_filename`, instead of simple `smart_open` -> It was assumed that everything should work for `file-like`)

PR #1841, opened 2018-01-15
Fix several coding errors in the doc2vec-lee examples

PeterHamilton:
This change fixes a couple different coding errors found in the code snippets included in the doc2vec-lee tutorial, including array index out-of-bounds selection and passing test document words to vector inferencing.

PR #1870, opened 2018-02-01
Fix `doc2vec-lee` notebook

numericlee:
Remove .words from lines 613 and 617


numericlee:
Without this change, code gives an error "list has no attribute .words"

menshikh-iv:
Nice catch @numericlee! I'll merge it after release.

menshikh-iv:
@numericlee congratz with first PR :+1:

piskvorky:
I don't get it -- wasn't `.words` specifically added in #1841?

This PR effectively reverse that one. They can't both be right.

menshikh-iv:
@piskvorky this fix for the mistake from #1841, `train_corpus` is `list of TaggedDocument` (in this case `.words` needed), but `test_corpus` is `list of list of str`, and here, we have no attribute `.words`.

PR #1875, opened 2018-02-02
Fix python 3 compatibility for `gensim.corpora.UciCorpus.save_corpus`

darindf:
Modified code to not throw an error with python 3


menshikh-iv:
Thank you @darindf

piskvorky:
Glad to see people are actually using these (somewhat arcane) corpus formats :D

PR #1876, opened 2018-02-03
Fix example block for `gensim.models.Word2Vec`

nzw0301:
Current docstring in `models.word2vec.py` seems to be a little bit strange for me, because the code block is nested and includes no python code.
So I remove `::` to avoid being nested code block, then I fix indents.

Before

After


menshikh-iv:
@nzw0301 thanks, congratz with the first contribution :1st_place_medal:

nzw0301:
@menshikh-iv  Thank you! I'm happy to contribute to gensim repo!

PR #1880, opened 2018-02-07
Add windows venv activate command to `CONTRIBUTING.md`

aneesh-joshi:
The current Contribution guide doesn't have a command for windows activation of a virtual env.

I encountered this problem and had to spend some time searching for it.
Here is a correction.

reference:  https://www.codementor.io/tips/2872397814/activating-a-virtualenv-on-windows


menshikh-iv:
Thank you @aneesh-joshi

PR #1892, opened 2018-02-09
LSI documentation

steremma:
In this PR I attempt to fix the source internal documentation for the LSI model. More specifically:

* Numpy docstrings for functions methods and classes

* Including type annotation for arguments and returned objects


gensim/models/lsi_dispatcher.py
```diff
@@ -5,12 +5,18 @@
 # Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
 
 """
-USAGE: %(program)s SIZE_OF_JOBS_QUEUE
+Dispatcher process which orchestrates distributed LSI computations. Run this
```
menshikh-iv: This is executable script, some "special case", please look into `__doc__` and `.. program-output` from https://github.com/RaRe-Technologies/gensim/commit/117d447087b2e9a36fda1279d837e045fefd1558 (same for lsi_worker)

menshikh-iv:
Goond job @steremma :fire:

PR #1895, opened 2018-02-10
Fix docstrings for `gensim.sklearn_api`. Fix #1667

### LINKED ISSUE #1667: Refactor API reference gensim.sklearn_api
**Check this PR for the soon-to-be-deprecated modules:** #1618 

Documented submodules:

- [ ] `__init__.py`
- [x] `atmodel.py`
- [x] `d2vmodel.py`
- [x] `hdp.py`
- [x] `ldamodel.py`
- [x] `ldaseqmodel.py`
- [x] `lsimodel.py`
- [x] `phrases.py`
- [x] `rpmodel.py`
- [x] `text2bow.py`
- [x] `tfidf.py`
- [x] `w2vmodel.py`

steremma:
In this PR I am working on the source documentation for the sklearn API.
More wrappers will be submitted in subsequent pushes to the same PR until they are all complete.


gensim/sklearn_api/lsimodel.py
```diff
@@ -5,11 +5,6 @@
 # Copyright (C) 2017 Radim Rehurek <radimrehurek@seznam.cz>
 # Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
 
-"""
-Scikit learn interface for gensim for easy use of gensim with scikit-learn
```
menshikh-iv: Some `__doc__` definitely needed

gensim/sklearn_api/lsimodel.py
```diff
@@ -20,14 +15,36 @@
 
 
 class LsiTransformer(TransformerMixin, BaseEstimator):
-    """
-    Base LSI module
+    """Base LSI module.
+
+    Scikit learn interface for `gensim.models.lsimodel` for easy use of gensim with scikit-learn.
```
menshikh-iv: please use links 
```
:class:`~gensim.model.lsimodel.LsiModel`
```

here and everywhere

menshikh-iv: Also, explicit mention "if you want to read more about it, please look into original class :class:`...`"

gensim/sklearn_api/lsimodel.py
```diff
@@ -20,14 +15,36 @@
...
 
     def __init__(self, num_topics=200, id2word=None, chunksize=20000,
                  decay=1.0, onepass=True, power_iters=2, extra_samples=100):
-        """
-        Sklearn wrapper for LSI model. See gensim.model.LsiModel for parameter details.
+        """Sklearn wrapper for LSI model.
+
+        Parameters
+        ----------
+        num_topics : int, optional
```
menshikh-iv: Wdyt about the link to original method only (for avoiding duplication)?

steremma: I also thought about that and I am not sure what is better. On one hand now we have duplication but on the other hand its easier for the developer and user to see the documentation in one tab. Because not all parameters are propagated to the inner model, some of the parameters will be visible in the wrapper and some in the original model (you would need 2 tabs open). I am a bit in favor of duplicating but not 100% sure so if you prefer I will remove duplication.

menshikh-iv: So, maybe combine both approaches: mentioned parameter & type here, but for description - sent the user to the parameter from original class?

steremma: Ok that sounds reasonable, I will apply asap

menshikh-iv: @steremma we discuss this questions again and this isn't good idea, because it's OK if user look into documentation online (and have a link), but if user use `python`/`jupyter`, he will call something like `help(model)` or `model?` and for this case, links don't work :( (and this is the main problem). For this reason - can you return descriptions for parameters, copy-paste is the lesser evil than docstring, that exists, but useless, if you can't read it in your interpreter.

Also, the link to original class must be in any case too.

gensim/sklearn_api/lsimodel.py
```diff
@@ -5,9 +5,19 @@
...
-Follows scikit-learn API conventions
+"""Scikit learn interface for `gensim.models.lsimodel`.
+
+Follows scikit-learn API conventions to facilitate using gensim along with scikit-learn.
+
+Examples
+--------
+Integrate with sklearn Pipelines:
+
+    >>> model = LsiTransformer(num_topics=15, id2word=id2word)
```
menshikh-iv: Great (examples is the really nice idea), but please be sure, that example executable (i.e. you write all needed imports & define data).

You can check that all works easily with `python -m doctest path/to/file/with/examples.py`

menshikh-iv:
@steremma please be accurate with PEP8 checks, look to https://travis-ci.org/RaRe-Technologies/gensim/jobs/343473932#L509 and resolve it.
You can easily check it locally before push, like `tox -e flake8`

steremma:
@menshikh-iv this PR is a work in progress I only push to sync between different workstations (I opened it so that you can easily track the status). Sometimes I have to leave one workstation quickly so I just push to continue from another one. In the end the commits will be squashed so I hope it won't be a problem. We can close the PR for now and re-open when its ready for review and merge.

gensim/sklearn_api/atmodel.py
```diff
@@ -18,17 +37,79 @@
...
 
 class AuthorTopicTransformer(TransformerMixin, BaseEstimator):
-    """
-    Base AuthorTopic module
-    """
+    """Base Author Topic module, wraps :class:`~gensim.models.atmodel.AuthorTopicModel`.
 
+    For more information on the inner workings please take a look at the original class. The model's internal workings
+    are heavily based on `"The Author-Topic Model for Authors and Documents", Osen-Zvi et. al 2004
+    <https://mimno.infosci.cornell.edu/info6150/readings/398.pdf>`_.
```
menshikh-iv: Maybe mention paper only in original class (not here), wdyt?

gensim/sklearn_api/hdp.py
```diff
@@ -4,9 +4,21 @@
...
-Follows scikit-learn API conventions
+"""Scikit learn interface for :class:`~gensim.models.hdpmodel.HdpModel`.
+
+Follows scikit-learn API conventions to facilitate using gensim along with scikit-learn.
+
+
+Examples
+--------
+
+    >>> from gensim.test.utils import common_dictionary, common_corpus
```
menshikh-iv: unindent example please (here and everywhere)

gensim/sklearn_api/atmodel.py
steremma: I will replace references with `some text <http://...>`_

menshikh-iv:
Current PR will fix #1895

Issue #557, opened 2015-12-06
Get rid of scipy


piskvorky:
Scipy in gensim is only needed for its `scipy.sparse` toolbox for handling sparse matrices, which is a huge overkill. We'd like to finally get rid of scipy as a dependency, keeping only numpy.

Now that we have gensim wheels for Windows (@tmylk), even a compiled gensim module to replace `scipy.sparse` might no longer be a show stopper.

Or maybe get the scipy people to split out `scipy.sparse` into its own, independent package... that would be the best and most flexible (and easiest for us).

@honnibal How do you deal with sparse matrices in spaCy?

gojomo:
Scipy is also used to get the BLAS operations in the word2vec/doc2vec cython code. (Maybe there's a way to use the BLAS directly, but right now it's via Scipy.)

honnibal:
spaCy hasn't needed complicated vector maths, because I've just been using Averaged Perceptron. The disadvantage of this is that it's been harder to say, try Adagrad instead, or plug in the group lasso, etc. But on the other hand I've been able to write code I can read easily, and performance is pretty good.

Here's my sparse array:
https://github.com/honnibal/thinc/blob/master/thinc/sparse.pyx
https://github.com/honnibal/thinc/blob/master/thinc/sparse.pxd

The idea is that the `cdef class` wrapper is stateless, and everything is contained in the `SparseArrayC` struct. There are then static methods to do all manipulations. These could be isolated functions, but it gives me better feels to collect them on a class. They're inlined due to a limitation in Cython --- it only understands the staticmethod decorator on inlined methods. I've been meaning to report this bug, but haven't gotten around to it. These arrays are stored in a hash table, to make up the sparse matrix.

The truth is I've always found the traditional sparse matrix descriptions confusing, and the `scipy.sparse` docs quite opaque. So I've never really understood the established data structures for this, and I figure my problem has fairly extreme values. So it made sense to me to just write some code.

piskvorky:
Great, thanks! Nothing beats the flexibility and control of rolling your own solution :) I like your approach.

To be honest, I've had my own ideas around sparse matrices as well. There are some code paths in gensim where having to know the final shape complicates things. I believe the concept of shape is often irrelevant, even misleading. Because we don't need the full array API, with all its flexibility and indexing -- just treat the shape as infinite (it's all zeros!) and support those matrix operations we really need, in the specific online algorithms.

[...]

PR #1896, opened 2018-02-11
[DNM] Optimize sparse * random dense matrix multiply in LsiModel

brianmingus:
This is an interim pull request so we can view and discuss changes.


menshikh-iv:
Related to https://github.com/RaRe-Technologies/gensim/issues/557 and https://github.com/RaRe-Technologies/gensim/wiki/GSoC-2018-project-ideas#sparsetools-package.

In perspective, better to create `csparse` as distinct package with python/cython wrappers. 

CC: @piskvorky

brianmingus:
Creating a new package is out of scope for me. For the GSoC project I think your path is to look into what's already been parallelized with OpenMP in GraphBlas in SuiteSparse and wrap that, and to then fill in the gaps by writing parallel operations with CSparse.

menshikh-iv:
@brianmingus thanks for the information, current PR will be nice "base" for creating the distinct package.

About sparse stuff in general: we want to drop scipy completely (we use an only small subset of scipy, for this reason, this isn't very hard).

Can you also add links to original "sourse" of `csparse` code + lisence and describe your work here (how you use `csparse`, what you add and why)?

brianmingus:
@menshikh-iv The reason for using SuiteSparse is that Professor Tim Davis is still developing it and it meets licensing and quality requirements.  He has also given us the go-ahead to use it, confirmed he thinks its the right choice for the Python community, and has ensured us it will continue to be developed.

The GraphBLAS implementation in SuiteSparse is based on some abstract algebra, and not all of it has been parallelized and uses OpenMP yet. I haven't really audited it at all, TBH, but that's what I gleaned from Davis' messages. For that that reason we are starting with CSparse. I feel that improving contributing directly to GraphBLAS will add unnecessary overhead as you'd likely have to learn some crazy math, whereas we can use CSparse + Cython to find embarassingly parallel cheat codes that we can implement while research into GraphBLAS continues.

SuiteSparse is BSD 3-clause, the code that I am building off of for pmultiply is MIT.

I recommend your GSoC student audit GraphBLAS, write code that automatically generates a wrapper for GraphBLAS, but do not contribute directly to GraphBLAS. Instead, for any important operations that are not yet parallelized and don't yet use OpenMP in GraphBLAS, instead implement them using CSparse + Cython akin to the example I'm commiting here.

piskvorky:
I'm +1 on a tighter integration, including the code directly where possible. Depending on external libraries (especially complex ones) is a headache from the maintenance POV. We don't want replace  one SciPy with another :)

menshikh-iv:
@piskvorky I prefer tighter integration too (if we should add 1, 2, 10 files, this is OK), but +300 files it's definitely too much (and in this case, one more dependency looks better).

About replace - this partially contradicts with #557.

menshikh-iv:
@brianmingus thanks for the advice, sounds good (but I'm not sure that we'll found the student with needed C level).

menshikh-iv:
@brianmingus can you please remove Matlab/Examples/non-relevant files from PR, right now it's impossible to review.

brianmingus:
@menshikh-iv No problem - I removed the cruft from CSParse.

gensim/test/test_csparse.py
```diff
@@ -0,0 +1,52 @@
...
+from os import cpu_count
+from timeit import Timer
+import unittest
+import logging
+import scipy
+
+class TestCSparse(unittest.TestCase):
+    def testPmultiply(self):
+
+        def matmul(X, Y): return X @ Y
```
menshikh-iv: This operator comes from `3.5`, but we support `2.7` too, please don't use it

brianmingus:
Benchmarks:

```
----------------------------------------------
cpus: 128
X.shape[0]: 1
X.shape[1]: 100000
X density: 1%
X non-zeros: 1000.0
Y.shape[0]: 100000
Y.shape[1]: 1
matmul: 0.0002566 seconds
pmatmul: 0.01109 seconds
speedup: 0.02314 times faster
asserted: X @ Y == pmultiply(X, Y)
----------------------------------------------
cpus: 128
X.shape[0]: 1
X.shape[1]: 100000
X density: 1%
X non-zeros: 1000.0
Y.shape[0]: 100000
Y.shape[1]: 10
matmul: 0.001683 seconds
pmatmul: 0.01274 seconds
speedup: 0.1321 times faster
asserted: X @ Y == pmultiply(X, Y)
----------------------------------------------
cpus: 128
X.shape[0]: 1
X.shape[1]: 100000
X density: 1%
X non-zeros: 1000.0
Y.shape[0]: 100000
Y.shape[1]: 100
matmul: 0.0439 seconds
pmatmul: 0.01794 seconds
speedup: 2.448 times faster
asserted: X @ Y == pmultiply(X, Y)
----------------------------------------------
cpus: 128
X.shape[0]: 1
X.shape[1]: 100000
X density: 1%
X non-zeros: 1000.0
Y.shape[0]: 100000
Y.shape[1]: 1000
matmul: 0.4982 seconds
pmatmul: 0.03356 seconds
speedup: 14.84 times faster
asserted: X @ Y == pmultiply(X, Y)
----------------------------------------------
cpus: 128
X.shape[0]: 1
X.shape[1]: 100000
X density: 1%
X non-zeros: 1000.0
Y.shape[0]: 100000
Y.shape[1]: 10000
matmul: 11.32 seconds
pmatmul: 0.135 seconds
speedup: 83.82 times faster
asserted: X @ Y == pmultiply(X, Y)
```

gensim/test/test_csparse.py
```diff
@@ -0,0 +1,52 @@
...
+        
+        if cpus > 1:
+            t_matmul = Timer(partial(matmul, X, Y)).timeit(1)
+            t_pmatmul = Timer(partial(pmatmul, X, Y)).timeit(1)
+
+            assert(t_pmatmul < t_matmul)
+
+        assert_array_equal(matmul(X, Y), pmatmul(X, Y))
+
+        print(f'''cpus: {cpus}
```
menshikh-iv: f-strings from 3.6, please use `format` instead.

setup.py
```diff
@@ -258,7 +259,12 @@ def finalize_options(self):
             include_dirs=[model_dir]),
         Extension('gensim.models.fasttext_inner',
             sources=['./gensim/models/fasttext_inner.c'],
-            include_dirs=[model_dir])
+            include_dirs=[model_dir]),
+        Extension('gensim.csparse.psparse',
+            sources=['gensim/csparse/psparse.pyx', 'gensim/csparse/CSParse/Source/cs_gaxpy.c'],
```
menshikh-iv: 1. What's about other files (~50 `.c` files in PR)?
2. Maybe we should remove all unused files (or if we shouldn't do this, why)?

gensim/test/test_csparse.py
```diff
@@ -0,0 +1,52 @@
...
+        X = rand(X_rows, X_cols, format="csc")
+        Y = asfortranarray(normal(0.0, 1.0, (X.shape[1], factors)))
+
+        cpus = cpu_count()
+        
+        if cpus > 1:
+            t_matmul = Timer(partial(matmul, X, Y)).timeit(1)
+            t_pmatmul = Timer(partial(pmatmul, X, Y)).timeit(1)
+
+            assert(t_pmatmul < t_matmul)
```
menshikh-iv: `assert` is operator (not a functions), no need `()` here

gensim/test/test_csparse.py
menshikh-iv: also, this is good as the benchmark (allows to control performance), but printing in tests isn't a good idea, please remove this when all other stuff will be ready to merge.

setup.py
```diff
@@ -258,7 +259,12 @@ def finalize_options(self):
...
+            extra_compile_args=['-fopenmp', '-O3', '-ffast-math'],
+            include_dirs = [csparse_dir, csparse_dir + '/CSParse/Include/'],
```
piskvorky: `os.path.join` better? (for operating systems with other file path delimiters)

piskvorky: @menshikh-iv what happens when the compilation fails (e.g. because of `openmp` issues)?
Are these extensions optional, or will this be a "hard-fail"?

brianmingus: An autoconf-style check or try / except compilation with fallback shouldn't prevent getting rid of the existing sparse library. It's not much faster than using * for the multiplication.

menshikh-iv: @piskvorky I'm not sure, but probably this will be "hard-fail", need to test it

gensim/csparse/psparse.pyx
```diff
@@ -0,0 +1,94 @@
+#-----------------------------------------------------------------------------
```
menshikh-iv: Better to remove `README.*` files, but mention links to original site/repo/any useful information here.

setup.py
```diff
@@ -103,6 +103,7 @@ def finalize_options(self):
 
 
 model_dir = os.path.join(os.path.dirname(__file__), 'gensim', 'models')
+csparse_dir = os.path.join(os.path.dirname(__file__), 'gensim', 'csparse')
```
menshikh-iv: Can you move this deeply to gensim, for example `gensim.models._csparse` instead of `gensim.csparse` (because this is really low-level internal stuff for special cases).

brianmingus:
This is getting close - main issues are 1) in simplifying import structure by enabling

```python
from gensim.csparse import gaxpy
```
This is now broken:
```python
from gensim.csparse.psparse import pmultiply
```

Additionally, I haven't actually tested that lsi is returning the same results.

Also of note, I removed both invocations of sparsetools.csc_matvecs, including in the else clause. We should do some kind of integration testing.

gensim/test/test_csparse.py
```diff
@@ -0,0 +1,40 @@
...
+# Copyright (C) 2010 Radim Rehurek <radimrehurek@seznam.cz>
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+from gensim.csparse import gaxpy
+
+from numpy.testing import assert_array_equal
+from numpy import asfortranarray
+from numpy.random import normal
+from functools import partial
+from scipy.sparse import rand
+from os import cpu_count
```
menshikh-iv: this doesn't works on `python2`, better use https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count

gensim/models/lsimodel.py
```diff
@@ -678,16 +677,15 @@ def stochastic_svd(corpus, rank, num_terms, chunksize=20000, extra_dims=None,
     # first phase: construct the orthonormal action matrix Q = orth(Y) = orth((A * A.T)^q * A * O)
     # build Y in blocks of `chunksize` documents (much faster than going one-by-one
     # and more memory friendly than processing all documents at once)
-    y = np.zeros(dtype=dtype, shape=(num_terms, samples))
+
```
menshikh-iv: You removed `y` from here, but in next line `y` referenced.

menshikh-iv:
@brianmingus about "integration testing" - we can simply add the new test (as the comparison with deprecated `sparsetools`), this is not perfect, but good enough solution (at least for current time), because `sparsetools` available right now and we have `scipy` as core dependency of project.


About simplifying structure - I don't think that this is really needed here because this is internal functionality (not public).

setup.py
```diff
@@ -23,6 +27,30 @@
...
 ez_setup.use_setuptools()
 
+# 3-clause BSD: https://github.com/yt-project/yt/blob/master/COPYING.txt
+def test_openmp():
+  
+    tmpdir = tempfile.mkdtemp()
+    curdir = os.getcwd()
+    os.chdir(tmpdir)
+
+    compiler = os.environ.get('CC', distutils.sysconfig.get_config_var('CC')).split()[0]
```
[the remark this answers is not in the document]
menshikh-iv: Maybe we no need "fallback" here (with `openmpi`), because
- We provide wheels for users (and we can use `openmpi` when we build wheels)
- If the user wants to compile itself, he should have **compiler** + **python headers (dev package)** -> what's a problem to install one more package (**openmpi**)?
- This looks like not really useful complication (I don't see the case when this is really needed)

CC: @piskvorky thoughts?

menshikh-iv:
Hello @brianmingus, how is going? Are you need some help with PR?

brianmingus:
I've been in Mexico for a week. Returning today.

menshikh-iv:
Ping @brianmingus

brianmingus:
Will deliver this weekend. What was the verdict on Windows support?

menshikh-iv:
@brianmingus several tips:
- we definitely should support windows
- I'm not sure that we need any "fallback" for windows, I described my opinion in https://github.com/RaRe-Technologies/gensim/pull/1896#discussion_r170450923 CC: @piskvorky 
- I can try to build wheels (with openmpi on windows) when this PR will be almost done, this is best test for this case I think.

piskvorky:
I can't say I care about Windows too much. For me it's "nice to have", but wouldn't delay this (amazing!) feature because of it.

brianmingus:
I am hoping to wrap this up but haven't managed to get to it. I noticed you
are doing fairly regular point releases - is the next one scheduled?

menshikh-iv:
@brianmingus yes, next release - first week of April

dselivanov:
Hi here. I've noticed this PR and discussion from twitter. Recently I've improved same CSR * dense matrix multiplies for my [reco](https://github.com/dselivanov/reco) R package (in fact it is also mainly heavily used for truncated SVD calculation).  I used Eigen library for that - it does CSR*dense multiplies with OpenMP. Not sure whether it is faster or slower (I think it should be at least on par with current PR). But it is battle-tested and widely used in industry.

brianmingus:
I dont recall stumbling upon or evaluating Eigen. It looks widely used,
although I note that ie Tensorflow doesn't actually have a parallelized
sparse*dense multiply. IIRC the whole thing is one OP.

GraphBLAS has strong surface validity. It went through a standards
committee and has an implementation in C that is increasingly using OpenMP.

This PR does not use GraphBLAS because its implementation of this function
does not use OpenMP.  Tim Davis says that is in progress.

So the proposal is basically GraphBLAS vs Eigen for the long term, and I
think GraphBLAS is more in the spirit of BLAS which we know and love (but
hate to install).

brianmingus:
I did a library search (incomplete :) before I started. Criticism welcome :)

brianmingus:
I have permission to work on this for a half day. Will do today or tomorrow
and wrap up.

I need to:

- Get csparse import working
- Make sure LsiModel is working
- Get tests passing

piskvorky:
Thanks @brianmingus -- a wrap-up summary, detailing any outstanding TODOs and issues so others can pick this up, would be much appreciated!

piskvorky:
I just saw @dselivanov 's PR (seems related): https://github.com/conradsnicta/armadillo-code/pull/66

brianmingus:
Just took a look at that. A very nice library, although it would
effectively add the dependency of Anaconda due to its reliance on multiple
different BLAS libraries.

dselivanov:
Reason for my contribution was that my pkg depends on both Armadillo and Eigen which is too much.  Also code which  depends on Eigen takes quite a lot of RAM to compile - I was unable to compile pkg on machine with 1gb RAM. Then I've rewritten CSR*dense products with arma. And it appeared to be 30%-50% faster compared to Eigen for my test cases (quite big matrices). After that I  proposed similar implementation for dense*CSC matrices to arma - the only sparse matrix format in arma is CSC.

@brianmingus this is not totally true. arma depends on LAPACK (if you call routines like `solve`, etc.). For BLAS it can use different backends, but also it has its own implementation which is slower than OpenBLAS but still very decent - see related [configuration](http://arma.sourceforge.net/docs.html):

> **ARMA_USE_BLAS** - Enable use of BLAS, or a high-speed replacement for BLAS (eg. OpenBLAS, Intel MKL, AMD ACML or the Accelerate framework). BLAS is used for matrix multiplication. Without BLAS, Armadillo will use a built-in matrix multiplication routine, which might be slower for large matrices.
**ARMA_DONT_USE_BLAS** - Disable use of BLAS. Overrides ARMA_USE_BLAS

@piskvorky Overall it will be amazing if Python/gensim community will be able to use arma - this will foster collaboration and allow to work on the shared C++ backend for common models/algorithms. Let's take it to another thread if there is interest.

brianmingus:
@dselivanov What do you think about my benchmarks above? Also, how does arma compare to the generalized plan to use GraphBLAS moving forward?

menshikh-iv:
Sorry for annoying @brianmingus but what's about current PR?

dselivanov:
@brianmingus sorry, I forgot to answer anything. Theoretically GraphBLAS look very interesting, but I've read only  basic descriptions. On other side I used arma for a couple of projects and I feel it is really easy to use it - templated C++ code feels quite natural and high-level. I feel like I will be far less productive with C code... Also arma is not only about sparse matrices  - it is more general scientific library (I believe suitesparse will be more comprehensive).

So my very "god feeling" conclusion - suitesparse/GraphBLAS can be more comprehensive for sparse matrices, arma - more general and much easy/natural to use. **I could be totally wrong - conclusion above is based on very limited experience**.

menshikh-iv:
Unfortunately, Brian has no time now, I hope that anybody with strong engineering skill can help us later.
@brianmingus big thanks for your work, I really appreciate it :+1:, feel free to re-open if you'll have a time.

brianmingus:
Dask is another compatible path to consider. I've been using it for several weeks. http://dask.pydata.org/en/latest/array-sparse.html

See the Requirements section for what we'd need to implement for this PR. Then we'd need to also add support for passing gensim Dask arrays. 

This may be the path to utilizing 1PB of memory.

Dask can also replace Pyro. 

@piskvorky thoughts?

piskvorky:
Yeah, we considered Dask. There is another gensim PR for using Dask already.

@brianmingus what has been your experience with Dask so far? I heard it's still a little rough for real practical use.

brianmingus:
Anaconda Inc have created a nice stack - Dask + xarray + HoloViews + GeoViews + datashader etc. Probably for their own client work, so it is somewhat idiosyncratic. A mix of functional (esp laziness), syntactic sugar, etc so you have to hardcore RTFM cc @jbednar your thoughts on Dask's suitability for gensim?

jbednar:
Dask and Numba are both very suitable for real work nowadays, and at this point in my life I can't imagine trying to run a project like gensim without them.  Specifically, it makes my heart sink to even imagine trying to support any of these  features of gensim without using Dask:

- All algorithms are memory-independent w.r.t. the corpus size (can process input larger than RAM, streamed, out-of-core),
- Efficient multicore implementations of popular algorithms
- Distributed computing

Plus Numba is amazing and now quite stable for optimizing the per-core performance, making it the perfect counterpart to Dask.  I can't say whether it's worth the cost of *switching* gensim to Dask now that you apparently already have a codebase supporting these features, but I can confidently predict that if you wanted to do that work, you'd be able to delete vast quantities of code and have the resulting system be far more capable.

PR #1903, opened 2018-02-14
Add method that show base installation info about gensim & related packages. Fix #1902

### LINKED ISSUE #1902: Add method for showing base info about gensim
Intro
----
When users report something, we need to know more information (not only gensim version, versions of related packages, OS and so on), and user need to copy-paste&run something like this

````
Please run the following snippet and paste the output below.
````

We can easily add it as method to gensim and use as `python -m gensim.scripts.package_info --info`

Todo
-------
Implement this method, information, that should be showed
- gensim version
- python version
- numpy version
- scipy version
- os version
- smart_open version
- `FAST_VERSION` variable
- path (where gensim installed)
- ??? (maybe something else)

### PR DESCRIPTION
**Pull request by sharanry**
Fixes #1902

### CODE REVIEW

File: gensim/__main__.py [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+import gensim
+import argparse
+
+if __name__ == "__main__":
+
+        parser = argparse.ArgumentParser()
+        parser.add_argument("--info", action="store_true", help="Display version of Gensim and its dpendencies")
+        opt = parser.parse_args()
+        if opt.info:
+                print(platform.platform())
```
menshikh-iv: Please move it do distinct function (that returns `dict` with information)
```python

from gesnim import package_info

info = package_info()
```

and print it here in nice manner

gensim/__main__.py
```diff
@@ -0,0 +1,19 @@
+import platform
+import sys
+import numpy
+import scipy
+import gensim
+import argparse
+
+if __name__ == "__main__":
+
+        parser = argparse.ArgumentParser()
```
menshikh-iv: Where is the description? Should use `__doc__` variable like here https://github.com/RaRe-Technologies/gensim/commit/117d447087b2e9a36fda1279d837e045fefd1558

gensim/__main__.py
```diff
@@ -0,0 +1,19 @@
+import platform
+import sys
+import numpy
+import scipy
+import gensim
+import argparse
+
```
menshikh-iv: Missing docstring for the module, should looks like https://github.com/RaRe-Technologies/gensim/commit/117d447087b2e9a36fda1279d837e045fefd1558#diff-cbaf05cc913e951e3205d150e33edf2eR9

menshikh-iv: In addition, don't forget about `.. program-output:: ` section

gensim/__main__.py
```diff
@@ -0,0 +1,19 @@
...
+if __name__ == "__main__":
+
```
menshikh-iv: remove this line

gensim/__main__.py
```diff
@@ -0,0 +1,19 @@
...
+        parser = argparse.ArgumentParser()
+        parser.add_argument("--info", action="store_true", help="Display version of Gensim and its dpendencies")
```
menshikh-iv: indentation should be 4 spaces (not 8)

gensim/__main__.py
```diff
@@ -0,0 +1,19 @@
...
+        opt = parser.parse_args()
+        if opt.info:
+                print(platform.platform())
+                print("Python", sys.version)
+                print("NumPy", numpy.__version__)
+                print("SciPy", scipy.__version__)
+                print("gensim", gensim.__version__)
```
menshikh-iv: missing `smart_open`, `FAST_VERSION` and path where gensim installed

gensim/__main__.py
menshikh-iv: Also, need to add `.rst` to `gensim/docs/src`

sharanry: On it

gensim/__main__.py
sharanry: for smart_open `smart_open.__version__` doesn't work as the variable does not exist.

menshikh-iv: Try to found other way

gensim/__main__.py
sharanry: > Also, need to add .rst to gensim/docs/src

How do I do this?

menshikh-iv: Look into `gensim/docs/src/scripts` for example, you'll see several `*.rst` files, you need to add something similar to needed place (same as in package hierarchy).

menshikh-iv:
Thanks @sharanry :+1:

PR #1909, opened 2018-02-15
setup.py was missing the license field

nils-werner:
While it has no effect on the actual license or the usability of this library, it hinders the use of tools like `pip-licenses`.


menshikh-iv:
thanks @nils-werner, congratz with first contribution :1st_place_medal:

PR #1910, opened 2018-02-15
Refactor documentation for `gensim.similarities.docsim`.

CLearERR:
@menshikh-iv


gensim/corpora/textcorpus.py
```diff
@@ -216,13 +216,31 @@ def __init__(self, input=None, dictionary=None, metadata=False, character_filter
...
+        >>>
+        >>>     def __len__(self):
+        >>>         if 'length' not in self.__dict__:
+        >>>             logger.info("caching corpus size (calculating number of documents)")
+        >>>         self.length = sum(1 for _ in self.get_texts())
+        >>>         return self.length
+        >>>
+        >>> corpus = CorpusMiislita(datapath('head500.noblanks.cor.bz2'))
+        >>> corpus.get_texts()
+        <generator object get_texts at 0x7fa932f397d0>
```
menshikh-iv: bad output, can you show the concrete line of the dataset `next(iter(corpus.get_texts()))` ?

gensim/corpora/textcorpus.py
```diff
@@ -216,13 +216,31 @@ def __init__(self, input=None, dictionary=None, metadata=False, character_filter
...
+        >>> class CorpusMiislita(corpora.TextCorpus):
+        >>>     stoplist = set('for a of the and to in on'.split())
+        >>>
+        >>>     def get_texts(self):
+        >>>         for doc in self.getstream():
+        >>>             yield [word for word in utils.to_unicode(doc).lower().split()
+        >>>                 if word not in CorpusMiislita.stoplist]
+        >>>
+        >>>     def __len__(self):
+        >>>         if 'length' not in self.__dict__:
```
menshikh-iv: no need to write something with logger, this should be simple & small example

gensim/corpora/textcorpus.py
```diff
@@ -216,13 +216,31 @@ def __init__(self, input=None, dictionary=None, metadata=False, character_filter
...
         >>>
-        >>> corpus = TextCorpus(datapath('head500.noblanks.cor.bz2'))
-        >>> for bow in corpus:
-        ...     pass
+        >>> class CorpusMiislita(corpora.TextCorpus):
+        >>>     stoplist = set('for a of the and to in on'.split())
+        >>>
+        >>>     def get_texts(self):
+        >>>         for doc in self.getstream():
+        >>>             yield [word for word in utils.to_unicode(doc).lower().split()
```
menshikh-iv: some issues with formatting

gensim/corpora/textcorpus.py
```diff
@@ -216,13 +216,31 @@ def __init__(self, input=None, dictionary=None, metadata=False, character_filter
...
+        >>>     def __len__(self):
+        >>>         if 'length' not in self.__dict__:
+        >>>             logger.info("caching corpus size (calculating number of documents)")
+        >>>         self.length = sum(1 for _ in self.get_texts())
+        >>>         return self.length
+        >>>
+        >>> corpus = CorpusMiislita(datapath('head500.noblanks.cor.bz2'))
+        >>> corpus.get_texts()
+        <generator object get_texts at 0x7fa932f397d0>
+        >>> corpus.__len__()
```
menshikh-iv: please `len(dorpus)` instead of this one, call "magic" directly is bad pattern (and is justified only for specific cases)

gensim/similarities/docsim.py
```diff
@@ -116,8 +131,20 @@ def get_index(self):
...
         """Return index vector at position `pos`.
 
+        Parameters
+        ----------
+        pos : int
+            Vector position.
+
+        Return
+        ------
+        {:class: `~scipy.sparse.csr_matrix`, :class: `~numpy.array`}
```
menshikh-iv: `numpy.array` -> `numpy.ndarray` here and everywhere. Also, in this case, link shouldn't be rendered -> don't use `~` for numpy/scipy

gensim/similarities/docsim.py
```diff
@@ -154,39 +179,60 @@ class Similarity(interfaces.SimilarityABC):
...
+        num_best : int, optional
+            If set, return only the `num_best` most similar documents, always leaving out documents with similarity = 0.
+            Otherwise, return a full vector with one float for every document in the index.
+        chunksize : int, optional
+            Size of block.
+        shardsize : int, optional
+            Size of shards should be chosen so that a `shardsize x chunksize` matrix of floats fits comfortably into
+            main memory.
+        norm : str, optional
+            Normalization to use. Accepted values: {l1, l2}.
```
menshikh-iv: In this case, instead of 
```
norm : str, optional
    Normalization to use. Accepted values: {l1, l2}.
```

should be
```
norm : {'l1', 'l2'}, optional
    Normalization to use.
```

menshikh-iv: This is better notation when we have several string pre-defined values.

Issue #1901, opened 2018-02-14
Add anaconda-cloud badge & setup build for MacOSX


menshikh-iv:
Into
-----
We build gensim for [conda-forge/conda-cloud](https://github.com/conda-forge/gensim-feedstock), but nobody knows about it.

Todo
-----
- [ ] (easy) Add conda-cloud badge to gensim `README.md`
- [ ] (medium) Setup MacOSX build in gensim-feedstock, related issue https://github.com/conda-forge/gensim-feedstock/issues/2

menshikh-iv:
Another yet try to fix this problem: https://github.com/conda-forge/gensim-feedstock/pull/12

PR #1911, opened 2018-02-16
Fix file-like closing bug from `gensim.corpora.MmCorpus`. Fix #1869

### LINKED ISSUE #1869: MmCorpus file-like object support bug
**Into**
We have some "weird" behavior if a user passes a `file-like` object to `MmCorpus`, based on [this mailing list thread](https://groups.google.com/forum/#!topic/gensim/420pRBwexKQ)

**Demonstration**

```python
from gensim.corpora import MmCorpus
import bz2

f = bz2.BZ2File("testcorpus.mm.bz2")
print(f.closed)  # 0
corpus = MmCorpus(f)
print(f.closed)  # 1 ???
```

**What happens**
File-like object was closed when we call `MmReader`, problem located here

https://github.com/RaRe-Technologies/gensim/blob/5342153eb4f4b02bb45bfa3951eef8250ac9f6b6/gensim/matutils.py#L1274

`with` automatically close `file-like` when we out of scope, **this is OK if we open this file**, but we **shouldn't close file-like passed from user**.

Related PR #1867 

**UPD:** another problem here - call `IndexCopus.__init__`, that didn't support `file-like` object at all.

sj29-innovate:


gensim/matutils.py
```diff
@@ -1343,25 +1343,40 @@ def __init__(self, input, transposed=True):
         """
         logger.info("initializing corpus reader from %s", input)
         self.input, self.transposed = input, transposed
-        with utils.file_or_filename(self.input) as lines:
+
+        # 'with' statement behaviour without closing the file object
+        mgr = (utils.file_or_filename(self.input))
```
menshikh-iv: useless `()`

gensim/matutils.py
```diff
@@ -1343,25 +1343,40 @@ def __init__(self, input, transposed=True):
...
+        exit = type(mgr).__exit__
```
menshikh-iv: `exit` is built-in function, don't re-define it

gensim/matutils.py
```diff
@@ -1343,25 +1343,40 @@ def __init__(self, input, transposed=True):
...
+
+                self.num_docs = self.num_terms = self.num_nnz = 0
+                for lineno, line in enumerate(lines):
+                    line = utils.to_unicode(line)
+                    if not line.startswith('%'):
+                        self.num_docs, self.num_terms, self.num_nnz = (int(x) for x in line.split())
+                        if not self.transposed:
+                            self.num_docs, self.num_terms = self.num_terms, self.num_docs
+                        break
+            except RuntimeError:
```
menshikh-iv: why `RuntimeError` ?

gensim/matutils.py
```diff
@@ -1343,25 +1343,40 @@ def __init__(self, input, transposed=True):
         """
         logger.info("initializing corpus reader from %s", input)
         self.input, self.transposed = input, transposed
-        with utils.file_or_filename(self.input) as lines:
+
+        # 'with' statement behaviour without closing the file object
+        mgr = (utils.file_or_filename(self.input))
+        exit = type(mgr).__exit__
+        value = type(mgr).__enter__(mgr)
+        exc = True
```
menshikh-iv: Shouldn't be inverted (i.e. false by default and True if exception)?

gensim/matutils.py
```diff
@@ -1328,10 +1328,23 @@ class MmReader(object):
     This allows us to process corpora which are larger than the available RAM.
 
     """
+    @contextmanager
```
menshikh-iv: Please move this function to `utils.py`  (this isn't related with MmReader, this can be useful in other cases)

gensim/matutils.py
```diff
@@ -1328,10 +1328,23 @@ class MmReader(object):
...
+    def open_file(self, input):
```
menshikh-iv: docstring in numpy-style please

gensim/matutils.py
```diff
@@ -1328,10 +1328,23 @@ class MmReader(object):
...
+        # Generates 'with' like behaviour excepting closing the file object
+        mgr = utils.file_or_filename(self.input)
+        exc = False
+        try:
+            yield mgr
+        except StandardError:
```
menshikh-iv: please add the comment, what's a reason for exception here

gensim/matutils.py
```diff
@@ -1328,10 +1328,23 @@ class MmReader(object):
...
+            exc = True
+            if not exit(mgr, *sys.exc_info()):
```
menshikh-iv: I'm slightly worried about `exit`, how it works?

gensim/test/test_corpora.py
```diff
@@ -225,6 +225,14 @@ def test_serialize_compressed(self):
         # MmCorpus needs file write with seek => doesn't support compressed output (only input)
         pass
 
+    def test_closed_file_object(self):
```
menshikh-iv: need to add test for your context manager too (with file path (existent, non-existent), file-like object)

gensim/matutils.py
sj29-innovate: In case of the exception , 'with' statement has the ability to silence errors. Here trying to close the file with the exception information if possible , otherwise raise  the exception.

gensim/matutils.py
menshikh-iv: if handle any - why not `Execption` instead of `StandartError` ?

gensim/matutils.py
menshikh-iv: this is call `__exit__` ?

gensim/matutils.py
sj29-innovate: Hmm , ya my bad . I will use except Exception.

gensim/matutils.py
sj29-innovate: Ya , it is __exit__() , I will correct it as in the finally block. Thanks.

gensim/test/test_corpora.py
sj29-innovate: Is it fine if I test whether the context manager throws exception or not for various cases ?

menshikh-iv: I think yes

gensim/test/test_utils.py
```diff
@@ -84,6 +85,48 @@ def test_decode_entities(self):
         expected = u'It\x92s the Year of the Horse. YES VIN DIESEL \U0001f64c \U0001f4af'
         self.assertEqual(utils.decode_htmlentities(body), expected)
 
+    def test_open_file_existent_file(self):
```
menshikh-iv: No need to check that this doesn't raise an exception if something goes wrong - unittest fail the current test and will show needed information (here and everywhere). Sorry, apparently I misunderstood you in the previous question

gensim/test/test_utils.py
```diff
@@ -84,6 +85,48 @@ def test_decode_entities(self):
...
+        exc = False
+        try:
+            with utils.open_file(datapath('testcorpus.mm')):
+                pass
+        except Exception:
+            exc = True
+
+        self.assertFalse(exc)
+
+    def test_open_file_non_existent_file(self):
```
menshikh-iv: Please use something like Please use something like https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises (instead of manual wrapping to try/exc )

gensim/utils.py
```diff
@@ -139,6 +140,36 @@ def file_or_filename(input):
...
+
+    """
+    mgr = file_or_filename(input)
+    exc = False
+    try:
+        yield mgr
+    except Exception:
+        # Handling any unhandled exceptions from the code nested in 'with' statement.
+        exc = True
+        if not type(mgr).__exit__(mgr, *sys.exc_info()):
```
menshikh-iv: why `type` is needed here?

gensim/test/test_utils.py
```diff
@@ -84,6 +85,48 @@ def test_decode_entities(self):
...
+        exc = False
+        try:
+            with utils.open_file(datapath('non_existent_file.txt')):
+                pass
```
menshikh-iv: Please iterate through this file too (not enough only open it), here and everywhere.
Simplest way looks like like
```python
with utils.open_file("...") as infile:
    self.assertEqual(sum(1 for _ in infile), numer_of_lines_in_file)
```

gensim/utils.py
sj29-innovate: As per the so thread , " exit = type(mgr).__exit__ is used because CPython has a history of caching special methods and this just makes it consistent ".

gensim/utils.py
```diff
@@ -139,6 +140,36 @@ def file_or_filename(input):
...
+    try:
+        yield mgr
+    except Exception:
+        # Handling any unhandled exceptions from the code nested in 'with' statement.
+        exc = True
+        if not mgr.__exit__(*sys.exc_info()):
+            raise
+        # Try to introspect and silence errors.
+    finally:
+        if not exc and isinstance(input, string_types):
```
menshikh-iv: if `exc=True` and input is `file-obj`, what happens?

gensim/utils.py
```diff
@@ -139,6 +140,36 @@ def file_or_filename(input):
...
+
+    """
+    mgr = file_or_filename(input)
+    exc = False
+    try:
+        yield mgr
+    except Exception:
+        # Handling any unhandled exceptions from the code nested in 'with' statement.
+        exc = True
+        if not mgr.__exit__(*sys.exc_info()):
```
menshikh-iv: what return `__exit__` ?

gensim/utils.py
sj29-innovate: Hmm. I think I missed checking for string type there. Will do that.

gensim/utils.py
sj29-innovate: Sorry , I didn't understand your query.

menshikh-iv: `X = mgr.__exit__(*sys.exc_info())` , what is `X` in this case?

sj29-innovate: x is True if the exception was handled gracefully , else False

menshikh-iv:
Ping @sj29-innovate, can you resolve merge-conflict please

menshikh-iv:
@sj29-innovate you broken something in your code, tests didn't passed correctly, please have a look to Travis logs

sj29-innovate:
@menshikh-iv  Any idea why this might be happening , that particular test is working fine locally.

sj29-innovate:
@menshikh-iv The test is running fine locally , can it be the cython support for in _mmreader that caused the test to break ?

menshikh-iv:
@sj29-innovate what you did in the last commit? Re-build `_mmcorpus.c` with Cython?

sj29-innovate:
@menshikh-iv  Yes i rebuild it using cython

gensim/corpora/_mmreader.pyx
```diff
@@ -65,7 +65,7 @@ cdef class MmReader(object):
         """
         logger.info("initializing cython corpus reader from %s", input)
         self.input, self.transposed = input, transposed
-        with utils.file_or_filename(self.input) as lines:
+        with utils.open_file(self.input) as lines:
```
sj29-innovate: @menshikh-iv Here is the change

menshikh-iv: @sj29-innovate yes, I see, thanks

menshikh-iv:
Need to merge #1910 first, sorry for waiting @sj29-innovate

sj29-innovate:
@menshikh-iv Can you please suggest some other task ? . Thanks.

menshikh-iv:
@sj29-innovate help guys with #1901 (you can try to setup MacOSX build in distinct PR yourself)

menshikh-iv:
Thank you @sj29-innovate :+1:

PR #1913, opened 2018-02-17
Fix docstrings for `gensim. interfaces`

yurkai:
Hi,
docstrings for core interfaces were added.


menshikh-iv:
Good work @yurkai :+1:

PR #1916, opened 2018-02-19
Fix method `estimate_memory` from `gensim.models.FastText` & huge performance improvement. Fix #1824

### LINKED ISSUE #1824: FastText memory usage greatly exceeds value returned by `estimate_memory`
#### Description
When using `gensim.models.fasttext.FastText`, the actual memory usage is much higher (>2x) than predicted by `FastText.estimate_memory`.
My usage scenario is to generate 300-dimensional word embeddings using SkipGram training with window size 8. My corpus has ~55,000,000 documents with ~4,144,457 word types across ~20,000,000,000 tokens. The machine has 16GB of available memory, 15GB of which are available for the Gensim process, as well as 16GB of Swap space.

The estimated memory usage is ~11.2GB (see below), which is identical to the size estimated for the Word2Vec model with the same parameters. Training with Word2Vec works flawlessly and uses almost exactly as much memory as estimated.

It seems that `FastText` does not implement its own `estimate_memory` method, but inherits it from the `Word2Vec` class, yielding unreliable values as can be seen below.  The critical section where the most memory is used seems to be this part in `FastText.init_ngrams`:

```python
all_ngrams = []
for w, v in self.wv.vocab.items():
    self.wv.ngrams_word[w] = compute_ngrams(w, self.min_n, self.max_n)
    all_ngrams += self.wv.ngrams_word[w]
```

```python
from gensim.models import fasttext

model = fasttext.FastText(size=300, sg=1, window=8, min_count=50, workers=8, iter=5)

# Word frequencies loaded from a finite state transducer on disk, i.e. no memory usage
freqs = load_frequencies()
vocab_size = sum(1 for typ, cnt in freqs.items() if cnt >= 50)
model.estimate_memory(vocab_size=vocab_size, report=True)
# { 'syn0': 4973348400,
#   'syn1neg': 4973348400,
#   'vocab': 2072228500
#   'total': 12018925300 }
# I.e. ~11.2GB, well within the available memory

model.build_vocab_from_freq(freqs, corpus_count=54878750)
# Memory usage is at ~7GB now, identical to Word2Vec

model.init_ngrams()
# ... Killed by OOM killer after swap space has run out
```

#### Expected Results
Should finish training without running out of memory.

#### Actual Results
Runs out of memory.

#### Versions
```
Linux-4.10.0-28-generic-x86_64-with-Ubuntu-16.04-xenial
Python 3.5.2 (default, Nov 23 2017, 16:37:01)
[GCC 5.4.0 20160609]
NumPy 1.13.3
SciPy 1.0.0
gensim 3.2.0
FAST_VERSION (fasttext) 1
FAST_VERSION (word2vec) 1
```

jbaiter:
This PR is an attempt to optimize the memory usage of the FastText model and to provide a more accurate `FastText.estimate_memory` method.

Specifically, it implements the following improvements:

- [x] Cythonize the `ft_hash` function
- [x] Cythonize the `compute_ngrams` function
- [x] Do not pre-compute and store the ngrams with the model, calculate them on the fly when needed
- [x] Do not store ngrams with the model, completely rely on the ngram hashes

In its current state this PR does not merge and does not pass the test suite, due to these issues:

- [x] ~~The improvements were done before the refactoring of the word embedding models, likely some code will have to be moved around~~
- [X] ~~There are some Python2/3 issues with the cythonized `compute_ngrams` function~~
- [ ] The tests currently check for OOV ngrams. In the old code, this relied on the `ngrams` attribute of the model, but in the optimized model I use the ngram hash. Since these hashes are bucketed, an OOV is most often no longer possible (i.e. if all buckets are occupied), i.e. these tests would have to be removed. Is this okay?

I think I can do 1) and 2) on my own when I find the time, but for 3) I'd need some help, since I'm not that familiar with the intentions behind the old code.


menshikh-iv:
@jbaiter Wow! Please resolve merge-conflicts (this is critical right now).  Probably, create new branch (based on fresh `develop`) and apply your changes will be simpler for you than resolve conflicts here.

When you resolve conflicts - please ping me for review / any help.

jbaiter:
@menshikh-iv So I managed to rebase my changes on the latest develop branch.
The Python 2 issue with `_compute_ngrams` has also been fixed.
The test suit now also passes all of the tests on Python 3.6.
On Python 2.7 and 3.5 however, the OOV vector in `testPersistenceForOldVersions` does not match the fixture. I'm currently at a loss at what could cause this, it's working fine in Python 3.6.

gensim/test/test_fasttext_wrapper.py
```diff
@@ -363,8 +355,8 @@ def testPersistenceForOldVersions(self):
             1.19031894, 2.01627707, 1.98942184, -1.39095843, -0.65036952])
         self.assertTrue(numpy.allclose(loaded_model["the"], in_expected_vec, atol=1e-4))
         # out-of-vocab word
-        out_expected_vec = numpy.array([-1.34948218, -0.8686831, -1.51483142, -1.0164026, 0.56272298,
-            0.66228276, 1.06477463, 1.1355902, -0.80972326, -0.39845538])
+        out_expected_vec = numpy.array([-0.33959097, -0.21121596, -0.37212455, -0.25057459, 0.11222091,
```
jbaiter: This is probably not correct, the vector seems to differ between different Python versions (the above was with 3.6)

gensim/test/test_fasttext_wrapper.py
```diff
@@ -363,8 +355,8 @@ def testPersistenceForOldVersions(self):
...
+            0.17517674, 0.26949012, 0.29352987, -0.1930912, -0.09438948])
```
jbaiter: This is probably not correct, the vector seems to differ between different Python versions (the above was with 3.6)

jbaiter:
I now reverted all changes to the deprecated `fasttext_wrapper` and the test suite is now passing in all Linux environments. I don't think it's too bad if it doesn't get the optimizations for now, given that it's deprecated.

On Windows there seems to be a memory-related issues, the memory allocation for the ngram vectors in `test_estimate_memory` causes a `MemoryError`.

menshikh-iv:
@jbaiter thanks, about memory - this is knows issue for appveyour :( I'll look into later, thanks for your patience.

gensim/models/fasttext_inner.pyx
```diff
@@ -317,7 +317,8 @@ def train_batch_sg(model, sentences, alpha, _work, _l1):
                 continue
             indexes[effective_words] = word.index
 
-            subwords = [model.wv.ngrams[subword_i] for subword_i in model.wv.ngrams_word[model.wv.index2word[word.index]]]
+            subwords = [model.wv.hash2index[ft_hash(subword) % model.bucket]
```
menshikh-iv: Please use only hanging indents (no vertical).

gensim/models/utils_any2vec_fast.pyx
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env cython
```
menshikh-iv: This related only with n-grams model, I think it's better to move it to `fasttext_inner.pyx`

gensim/models/utils_any2vec_fast.pyx
```diff
@@ -0,0 +1,21 @@
...
+# cython: boundscheck=False
+# cython: wraparound=False
+# cython: cdivision=True
+# coding: utf-8
+
+def ft_hash(unicode string):
```
menshikh-iv: why not `cdef` with `nogil`?

gensim/models/utils_any2vec_fast.pyx
```diff
@@ -0,0 +1,21 @@
...
+    cdef unsigned int h = 2166136261
+    for c in string:
+        h ^= ord(c)
+        h *= 16777619
+    return h
+
+
+cpdef compute_ngrams(word, unsigned int min_n, unsigned int max_n):
```
menshikh-iv: why not `nogil`, you can fix type for `word`.

gensim/models/utils_any2vec_fast.pyx
```diff
@@ -0,0 +1,21 @@
...
+    cdef unicode extended_word = f'<{word}>'
```
menshikh-iv: f-strings supports only in `3.6` (and we maintain `2.7`, `3.5`, `3.6`), please use simple concatenation (or any alternative) here.

gensim/test/test_fasttext.py
```diff
@@ -448,27 +429,21 @@ def test_sg_neg_training(self):
...
         tmpf = get_tmpfile('gensim_fasttext.tst')
         model_neg = FT_gensim(sentences, size=10, min_count=0, seed=42, hs=0, negative=5)
         model_neg.save(tmpf)
         model_neg = FT_gensim.load(tmpf)
         self.assertTrue(len(model_neg.wv.vocab), 12)
-        self.assertTrue(len(model_neg.wv.ngrams), 202)
         model_neg.build_vocab(new_sentences, update=True)  # update vocab
         model_neg.train(new_sentences, total_examples=model_neg.corpus_count, epochs=model_neg.iter)
         self.assertEqual(len(model_neg.wv.vocab), 14)
-        self.assertTrue(len(model_neg.wv.ngrams), 271)
```
menshikh-iv: Why you removed part of tests (I mean all removed lines in tests)?

gensim/models/keyedvectors.py
```diff
@@ -1535,9 +1535,7 @@ def __init__(self, vector_size, min_n, max_n):
         self.vectors_vocab_norm = None
         self.vectors_ngrams = None
         self.vectors_ngrams_norm = None
-        self.ngrams = {}
```
menshikh-iv: why?

menshikh-iv:
@jbaiter don't forget to resolve merge-conflict too

gensim/models/fasttext_inner.pyx
jbaiter: That is, remove the newline and keep the comprehension on a single line?

gensim/models/utils_any2vec_fast.pyx
jbaiter: I don't think this would work, since both functions are used by both `models.fasttext` and `models.keyedvectors`. Moving the functions into `models.fasttext` would break, since this would cause a circular import.

gensim/models/utils_any2vec_fast.pyx
jbaiter: Both functions need to be called from both Python and Cython, so `cdef` won't work, but `cpdef` should. Will be fixed.

gensim/models/utils_any2vec_fast.pyx
jbaiter: This is in Cython, which since 0.24 automatically generates cross-compatible C-code for it. It works fine under 2.7.

gensim/test/test_fasttext.py
jbaiter: As I mentioned in the PR, one optimization was to remove the storage of ngrams on the model and solely rely on the hashes. This is why any tests that assert the number of ngrams in the model are no longer neccesary.

With the lack of ngrams, the `__contains__` check now also only uses the hashed and bucketed ngrams, which is why a 'real' OOV is a lot rarer (i.e. it will only happen if not all buckets are occupied and none of the ngrams in the token match any occupied bucket).

gensim/models/fasttext_inner.pyx
menshikh-iv: single line (if line <= 120 characters) or something like
```python
subwords = [
    ....
    ....
]
```

gensim/models/utils_any2vec_fast.pyx
menshikh-iv: wow, really? I didn't know about it, thanks for the information!

gensim/models/utils_any2vec_fast.pyx
jbaiter: `nogil` doesn't work, since the function returns a Python object.

gensim/models/utils_any2vec_fast.pyx
jbaiter: Fixing the type for `word` doesn't really work, since the token might be a `str` on Python 2.7. `nogil` won't work either, since the function returns a Python object.

gensim/models/utils_any2vec_fast.pyx
menshikh-iv: Aha, so, in this case, better to name it as `_utils_any2vec.pyx` (similar with `_mmreader.pyx` and `_matutils.pyx`)

gensim/models/utils_any2vec_fast.pyx
jbaiter: You can read about it here: http://cython.readthedocs.io/en/latest/src/tutorial/strings.html#string-literals

jbaiter:
I ran some benchmarks with my optimized version and the current gensim implementation of FastText. 

Initially the performance was about 10x slower, but I implemented an optimization that pre-generates the ngram buckets for each word to avoid calling `compute_ngrams` and `ft_hash` in the training loop. This actually improved performance to 2x that of the original implementation.

I trained on the `text8` corpus with the default settings on an Xeon E5-1620 with 8 cores.
The test script can be found at https://gist.github.com/3d781a311e536b471b24fb4a46c952a4
All measurements were done with GNU time (`/usr/bin/time`).

| Metric                 | unoptimized         | optimized           |
| -----------------------| ------------------: | ------------------: |
| Training time          | 585.5               | **299.3s**          |
| Training words/sec     | 106,804 words/s     | **208,911 words/s** |
| Training peak memory   | 1,409.92 MiB        | **1,181.35 MiB**    |
| Evaluation time        | 60.66s              | **59.25s**          |
| Evaluation peak memory | 980.46 MiB          | **768.42 MiB**      |

As you can see, the goal of reducing memory consumption was achieved and we additionally almost doubled the training speed, while maintaining evaluation speed.

The quality of the vectors seems to suffer a bit, however I think that this might be due to the different random initializations of the two models.


| Benchmark                   | unoptimized            | optimized             |
| --------------------------- | ---------------------: | --------------------: |
| capital-common-countries    | **8.5% (43/506)**      | 7.1% (36/506)         |
| capital-world               | 3.6% (52/1452)         | 3.6% (52/1452)        |
| currency                    | 0.0% (0/268)           | 0.0% (0/268)          |
| city-in-state               | **4.3% (68/1571)**     | 4.1% (64/1571)        |
| family                      | 40.8% (125/306)        | 40.8% (125/306)       |
| gram1-adjective-to-adverb   | **96.0% (726/756)**    | 95.1% (719/756)       |
| gram2-opposite              | **91.8% (281/306)**    | 91.2% (279/306)       |
| gram3-comparative           | 81.4% (1026/1260)      | **82.2% (1036/1260)** |
| gram4-superlative           | **85.2% (431/506)**    | 84.6% (428/506)       |
| gram5-present-participle    | **76.3% (757/992)**    | 75.9% (753/992)       |
| gram6-nationality-adjective | **71.3% (978/1371)**   | 68.6% (940/1371)      |
| gram7-past-tense            | **29.1% (387/1332)**   | 28.4% (378/1332)      |
| gram8-plural                | **80.9% (803/992)**    | 80.5% (799/992)       |
| gram9-plural-verbs          | 82.8% (538/650)        | **82.9% (539/650)**   |
| **total**                   | **50.7% (6215/12268)** | 50.1% (6148/12268)    |

menshikh-iv:
@jbaiter first table looks awesome: almost x2 faster + reduce memory usage, fantastic :fire:!

About second table & random init: can you train/evaluate several times & average results (to exclude random effect) please?

manneshiva:
@jbaiter The speedup looks great! 
Just a few comments:
1. You would also want to include the size of `wv.buckets_ngrams` to `estimate_memory()`  method.
2. The initial setup time before training (calculating hashes and storing ngrams) will definitely be faster now that you have Cythonized `compute_ngrams` and `ft_hash`. But I am not sure what is causing the speedup in training (in terms of words/sec). As far as I remember, the initial code (Gensim `3.3.0`) did not call `compute_ngrams` or `ft_hash` during training (from `fasttext_inner.pyx`), what do you think is the reason for 2x increase in the number of words processed per second (I am assuming you have pasted these numbers (words/sec) from the logs)?  The table values compare your current implementation (`optimized`) with the implementation from Gensim 3.3.0 (`original`), am I right?

menshikh-iv:
@manneshiva @jbaiter maybe try to compare on bigger corpus (something ~1GB, not `text8`, this is more "fair" performance comparison)

jbaiter:
@menshikh-iv 
> About second table & random init: can you train/evaluate several times & average results (to exclude random effect) please?
>  Besides, please make backward-compatibility test (train FastText with old code, and try to load it with current code).
Will do, I'll also do a run each with a fixed seed.

> Also, please have a look at Appveyor, I see MemoryError in test_estimate_memory.
Do you have any idea what could be causing this? `estimate_memory` is called in the training tests as well, but doesn't cause an error there. You mentioned that this tends to happen on AppVeyor?

> @manneshiva @jbaiter maybe try to compare on bigger corpus (something ~1GB, not text8, this is more "fair" performance comparison)
Can you recommend a suitable dataset that works with the tests in `questions-answers.txt`? The corpora I currently have on hand are all historical German and not really suitable for these tests.

@manneshiva 
> You would also want to include the size of wv.buckets_word to estimate_memory() method.
Will do!

> The initial setup time before training (calculating hashes and storing ngrams) should definitely be faster now that you have Cythonized compute_ngrams and ft_hash. But I am not sure what is causing the speedup in training (in terms of words/sec). As far as I know, the initial code (Gensim 3.3.0) did not call compute_ngrams or ft_hash during training (from fasttext_inner.pyx), what do you think is the reason for 2x increase in the number of words processed per second (I am assuming you have pasted these numbers (words/sec) from the logs)? The table values compare your current implementation (optimized) with the implementation from Gensim 3.3.0 (original), am I right?

Yes, the values for `current` are from the latest commit on the `develop` branch, `optimized` is the code of this PR.
I think the reason might be that the new code does not use a list comprehension but instead directly looks up the tuple of ngram buckets:

``` python
subwords = [model.wv.ngrams[subword_i] for subword_i in model.wv.ngrams_word[model.wv.index2word[word.index]]]
```
vs
```python
subwords = model.wv.buckets_word[word.index]
```
The current version does  `num_ngrams + 2` lookups, while the new code always just uses a single lookup. It could be that the previous code was causing a lot of cache misses and the new one doesn't? I could try to look at cache usage patterns with `perf` if I find the time.

menshikh-iv:
> Do you have any idea what could be causing this? estimate_memory is called in the training tests as well but doesn't cause an error there. You mentioned that this tends to happen on AppVeyor with other tests as well?

This happens sometimes with appveyour (by memory limit reasons), so, you can try to use the smaller model in this case.

> Can you recommend a suitable dataset that works with the tests in questions-answers.txt? The corpora I currently have on hand are all historical German and not really suitable for these tests.

Sample from https://github.com/RaRe-Technologies/gensim-data/releases/tag/wiki-english-20171001 should be a good idea (you should pick first 1M articles).

menshikh-iv:
I'll make benchmark myself too (will be updated soon)

**Code**: **TODO, share as gist**
**Input**: 0.5M first articles from wiki (`"title" + "sect_title" + "sect_content" + ...`)
**Preprocessing**: very simple (without stemming)
```python
from gensim.parsing.preprocessing import (
    preprocess_string, strip_punctuation, strip_multiple_whitespaces, remove_stopwords, strip_short
)

prc = partial(
    preprocess_string, 
    filters=[strip_punctuation, strip_multiple_whitespaces, remove_stopwords, strip_short]
)
```
**Model parameters**: almost default `model = FastText(sentences=corpus, iter=1)`

| Metric                 | original         | optimized           |
| -----------------------| ------------------: | ------------------: |
| Training time          |        -      | -         |
| Training words/sec     | -    | - |
| Training peak memory   | 5,173 MB       |  -  |

jbaiter:
So I ran the `text8` texts 100 times with both the current implementation and the optimization, the total score is now at 50.38% (original) and 50.29%, so a pretty small difference.
The averaged tests with the first 1M documents from the wiki corpus are currently running, will be able to report on the outcome tomorrow.

As for the causes of the speedup, it seems that my changes somehow result in a significant increase in parallelism, as can be gathered from these performance counters:

**Perfomance counters for training + evaluation on current**
```
    1404012.973981      task-clock (msec)         #    2.145 CPUs utilized
         1,272,950      context-switches          #    0.907 K/sec
            27,026      cpu-migrations            #    0.019 K/sec
           667,005      page-faults               #    0.475 K/sec
 4,603,142,310,724      cycles                    #    3.279 GHz                      (48.93%)
 5,168,917,813,028      instructions              #    1.12  insn per cycle           (61.42%)
 1,032,270,301,259      branches                  #  735.228 M/sec                    (61.21%)
     8,170,082,299      branch-misses             #    0.79% of all branches          (60.97%)
 1,732,560,787,151      L1-dcache-loads           # 1234.006 M/sec                    (45.44%)
   164,838,925,283      L1-dcache-load-misses     #    9.51% of all L1-dcache hits    (28.92%)
    36,178,103,172      LLC-loads                 #   25.768 M/sec                    (25.83%)
    10,871,658,490      LLC-load-misses           #   30.05% of all LL-cache hits     (36.92%)

     654.542443124 seconds time elapsed
```


**Performance counters for training + evaluation on optimized**
```
    1032653.771199      task-clock (msec)         #    3.002 CPUs utilized
         1,721,311      context-switches          #    0.002 M/sec
             9,352      cpu-migrations            #    0.009 K/sec
           623,004      page-faults               #    0.603 K/sec
 3,411,762,671,386      cycles                    #    3.304 GHz                      (48.48%)
 4,492,762,293,396      instructions              #    1.32  insn per cycle           (60.97%)
   877,973,650,752      branches                  #  850.211 M/sec                    (60.70%)
     6,791,892,588      branch-misses             #    0.77% of all branches          (60.63%)
 1,521,323,964,974      L1-dcache-loads           # 1473.218 M/sec                    (43.68%)
   126,194,704,440      L1-dcache-load-misses     #    8.30% of all L1-dcache hits    (29.18%)
    21,929,347,031      LLC-loads                 #   21.236 M/sec                    (26.23%)
     5,608,723,856      LLC-load-misses           #   25.58% of all LL-cache hits     (36.70%)

     343.979779366 seconds time elapsed
```

menshikh-iv:
I checked current PR with 500,000 articles from wiki, my results https://github.com/RaRe-Technologies/gensim/pull/1916#issuecomment-369153399, this looks exciting :star2:  :fire: impressive work @jbaiter!

I also checked, the model from `3.3.0` loads fine, need to check `3.2.0` only now.

menshikh-iv:
@jbaiter last things that missed here
- [  ] Add test with loading old `FastText` model from `3.2.0`
- [ ] Fix Appveyour memory issue (reduce model size)

menshikh-iv:
This is files from profiler (current relese version + PR version)

**Script:**
```python

data = api.load("text8")
model = FastText(data)
```
[master.txt](https://github.com/RaRe-Technologies/gensim/files/1766834/master.txt)
[optimized.txt](https://github.com/RaRe-Technologies/gensim/files/1766835/optimized.txt)

@manneshiva this is for you

gensim/models/fasttext.py
```diff
@@ -744,54 +766,52 @@ def init_ngrams_weights(self, wv, update=False, vocabulary=None):
...
-            for i, ngram in enumerate(all_ngrams):
-                ngram_hash = _ft_hash(ngram) % self.bucket
-                if ngram_hash in wv.hash2index:
-                    wv.ngrams[ngram] = wv.hash2index[ngram_hash]
-                else:
-                    ngram_indices.append(ngram_hash % self.bucket)
-                    wv.hash2index[ngram_hash] = new_hash_count
-                    wv.ngrams[ngram] = wv.hash2index[ngram_hash]
-                    new_hash_count = new_hash_count + 1
+            wv.num_ngram_vectors = 0
```
jayantj: We could probably reduce some variables here - there seems to be some redundancy, if I understand correctly. `wv.num_ngram_vectors`, `new_hash_count` and `len(ngram_indices)` serve effectively the same purpose.
Maybe we could use `len(ngram_indices)` within the loop and set `wv.num_ngram_vectors` at the end of the loop?

gensim/models/fasttext.py
```diff
@@ -744,54 +766,52 @@ def init_ngrams_weights(self, wv, update=False, vocabulary=None):
...
             self.reset_ngrams_weights(wv)
         else:
-            new_ngrams = []
-            for w, ngrams in iteritems(wv.ngrams_word):
-                new_ngrams += [ng for ng in ngrams if ng not in wv.ngrams]
-
-            new_ngrams = list(set(new_ngrams))
-            wv.num_ngram_vectors += len(new_ngrams)
-            logger.info("Number of new ngrams is %d", len(new_ngrams))
+            if not wv.buckets_word:
```
jayantj: What is the purpose of this?

gensim/models/fasttext.py
```diff
@@ -744,54 +766,52 @@ def init_ngrams_weights(self, wv, update=False, vocabulary=None):
...
             new_hash_count = 0
-            for i, ngram in enumerate(new_ngrams):
-                ngram_hash = _ft_hash(ngram) % self.bucket
-                if ngram_hash not in wv.hash2index:
-                    wv.hash2index[ngram_hash] = new_hash_count + self.old_hash2index_len
-                    wv.ngrams[ngram] = wv.hash2index[ngram_hash]
-                    new_hash_count = new_hash_count + 1
-                else:
-                    wv.ngrams[ngram] = wv.hash2index[ngram_hash]
+            num_new_ngrams = 0
```
jayantj: There seems to be some redundancy again with `new_hash_count`, `num_new_ngrams`.

gensim/models/fasttext.py
```diff
@@ -847,20 +867,21 @@ def init_ngrams_post_load(self, file_name, wv):
...
-            wv.ngrams[ngram] = i
+        wv.num_ngram_vectors = 0
+        for word in wv.vocab.keys():
+            for ngram in _compute_ngrams(word, wv.min_n, wv.max_n):
+                ngram_hash = _ft_hash(ngram) % self.bucket
+                if ngram_hash in wv.hash2index:
+                    continue
+                ngram_indices.append(len(wv.vocab) + ngram_hash)
+                wv.hash2index[ngram_hash] = wv.num_ngram_vectors
+                wv.num_ngram_vectors += 1
```
jayantj: Can be set to `len(ngram_indices)` at the end instead (sorry for nitpicking, but we already have very long code for some of these methods)

gensim/models/keyedvectors.py
```diff
@@ -1603,13 +1602,14 @@ def word_vec(self, word, use_norm=False):
...
                 ngram_weights = self.vectors_ngrams_norm
             else:
                 ngram_weights = self.vectors_ngrams
             for ngram in ngrams:
-                word_vec += ngram_weights[self.ngrams[ngram]]
+                ngram_hash = _ft_hash(ngram) % self.bucket
+                if ngram_hash in self.hash2index:
+                    word_vec += ngram_weights[self.hash2index[ngram_hash]]
             if word_vec.any():
                 return word_vec / len(ngrams)
```
jayantj: This probably needs to be updated to only take into account the ngrams for which hashes were present in `self.hash2index`

gensim/test/test_fasttext.py
```diff
@@ -155,16 +154,16 @@ def test_load_fasttext_format(self):
...
-            -0.58482,
-            -0.22276,
-            -0.41215,
-            0.91015,
-            -1.6786,
-            -0.26724,
-            0.58818,
-            0.57828,
-            0.75801
+            -0.21929,
```
jayantj: What is the reason for this change? Could it because of the `len(ngrams)` issue mentioned in a comment above?

gensim/test/test_fasttext.py
```diff
@@ -206,16 +205,16 @@ def test_load_fasttext_new_format(self):
...
-            -0.19,
-            0.013482,
-            -0.86767,
-            -0.21684,
-            -0.89928,
-            0.45124,
-            0.18025,
-            -0.14128,
-            0.22508
+            -0.49111,
```
jayantj: Ditto:
What is the reason for this change? Could it because of the `len(ngrams)` issue mentioned in a comment above?

gensim/test/test_fasttext.py
jbaiter: That's really a crucial question, but it's not because of the `len(ngrams)` issue.
The change was honestly simply made because the numbers were pretty similar and I thought that the vectors changed a bit since the new code is far more lenient in assigning a vector to unknown ngrams (i.e. once all buckets are occupied, any ngram will result in a vector, even if it was not in the original corpus).

But it looks like there might a bug in the old code: There are **a lot** more ngram vectors in the loaded model (17004) than there are in the model on disk (2762). This is probably because `wv.vectors_ngram = wv.vectors_ngrams.take(ngram_indices, axis=0)` in `init_ngrams_post_load` will result in a `(num_ngrams_total, ngram_vec_len)` matrix. Shouldn't `vectors_ngram` have a shape of `(num_buckets, ngram_vec_len)`?

menshikh-iv:
Some missed stuff

- Add `.c` and `.pyx` files to https://github.com/RaRe-Technologies/gensim/blob/develop/MANIFEST.in
- Add `.rst` for `_utils_any2vec.py` & `_utils_any2vec.pyx` to `gensim/docs/src/models`
- Update `apiref.rst` according to previous changes

gensim/test/test_fasttext.py
jayantj:
```
...since the new code is far more lenient in assigning a vector to unknown ngrams (i.e. once all buckets are occupied, any ngram will result in a vector, even if it was not in the original corpus).
```
Ahh right, that makes sense, thanks for explaining.

Re: the number of ngram vectors being greater than `num_buckets` (or the number of vectors on disk) - I see why that might have been happening. With a ngram vocab larger than the number of buckets, a lot of ngrams will be mapped to the same indices. And when `.take` is passed a list that contains multiple occurrences of the same index, the vector at that index is "taken" multiple times.
For example - 
```python
all_vectors = np.array([[0.1, 0.3], [0.3, 0.1]])
taken_vectors = all_vectors.take([0, 1, 0], axis=0)
taken_vectors.shape
>>> (3, 2)
```
So it wouldn't result in incorrect results, but yeah, it'd result in unexpectedly high memory usage (and kind of blowing out the whole idea of keeping memory usage constant even with increasing ngram vocabs, out of the water). Thanks for investigating this and explaining it!

@menshikh-iv we're good to merge from my side

PR #1919, opened 2018-02-20
Correct docstring re: initialization parameter `sg`

mdcclv:
Very minor toggling of a boolean in documentation


menshikh-iv:
Thanks @mdcclv, congratz with first contribution :+1:

PR #1926, opened 2018-02-22
Add token frequencies to the corpora.Dictionary

LvanWissen:
The Dictionary object now only stores document frequency information (i.e. in how many documents a token occurs) for tokens. This commit adds a token frequency dictionary `tfs` to the Dictionary object that contains a `tokenid: frequency` mapping. 

This allows for easy corpus inspection and this might be helpful for implementing another way of filtering tokens (not implemented).

### DIFF
--- a/gensim/corpora/dictionary.py
+++ b/gensim/corpora/dictionary.py
@@ -38,6 +38,8 @@ class Dictionary(utils.SaveLoad, Mapping):
         Reverse mapping for token2id, initialized in lazy manner to save memory.
     dfs : dict of (int, int)
         Document frequencies: token_id -> in how many documents contain this token.
+    tfs : dict of (int, int)
+        Token frequencies: token_id -> token frequency from all documents.
     num_docs : int
         Number of documents processed.
     num_pos : int
@@ -70,6 +72,7 @@ class Dictionary(utils.SaveLoad, Mapping):
         self.token2id = {}
         self.id2token = {}
         self.dfs = {}
+        self.tfs = {}
 
         self.num_docs = 0
         self.num_pos = 0
@@ -257,6 +260,11 @@ class Dictionary(utils.SaveLoad, Mapping):
             for tokenid in iterkeys(result):
                 dfs[tokenid] = dfs.get(tokenid, 0) + 1
 
+            # increase token frequency for each token that appeared in the document
+            tfs = self.tfs
+            for tokenid, count in iteritems(result):
+                tfs[tokenid] = tfs.get(tokenid, 0) + count
+
         # return tokenids, in ascending id order
         result = sorted(iteritems(result))
         if return_missing:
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Hello @LvanWissen,

Questions:
- Why did this need?  What's a motivation?
- This attribute never used, why we should add it to class? User can calculate it distinctly with really simple code
  ```python
  tfs = dict()
  for doc in documents:
      for word in doc:
          tfs[word] = tfs.get(word, 0) + 1
   ```
- How this affect to performance at all (need benchmark with small/big corpus by speed & memory consumption) ? This is one of "core" classes, for this reason, adding any additional calculations should be well-grounded

menshikh-iv:
CC: @piskvorky

LvanWissen:
Hi @menshikh-iv, 
Thank you for your comment. To further explain my PR:

- If the token frequencies are calculated separately one needs to iterate over the entire corpus again and this can be quite time consuming. Since the Dictionary class already updates itself by iterating over the corpus, but ignores the result of the doc2bow function (which outputs token-frequency information), incorporating a token frequency dict in the Dictionary class would make sense. This also means that the same tokenids can be used.
- I do realise that the Dictionary class is a core component of Gensim and that any changes to it should be well-supported. I do not know how this affects performance, but I am sure my code can be optimized if the term frequency dict is seen as a useful addition to the class!

menshikh-iv:
@LvanWissen Thanks for explanation
- I'm sure that this change slowdown code & increase memory consumption (and this **will be noticeable, not 1-2 %**). 
- `"This also means that the same tokenids can be used"` - used by than? This no needed for any gensim model (and this class in the base for using with models that pass documents as iterable of BoW), remember that in the first place this is "technical object" with the concrete target.  
- `"If the token frequencies are calculated separately one needs to iterate over the entire corpus again and this can be quite time-consuming"` - that's True, I understand, but If you want to calculate token-freq and fill `Dictionary` in one pass, you can use something like this:
  ```python

  corpus = [["a", "a", "c"], ["c", "b", "c"]]
  dct = Dictionary()


  tf = dict()
  for doc in corpus:
      for (token_id, freq) in dct.doc2bow(doc, allow_update=True):
          tf[token_id] = tf.get(token_id, 0) + freq
   ```

In general, it seems to me that this is not a very good addition because this didn't make possible something useful (I showed an alternative way in the code below), but degrades performance, I'm -1 for this change.

If you disagree with me - please argue, I'm always open to discussion.
If you agree - please close PR.

piskvorky:
@LvanWissen do you need the term frequencies to be exact?

Having them is a good idea, but the memory footprint will be massive.

We're considering replacing all these counters in Gensim by [Bounter](https://github.com/RaRe-Technologies/bounter), which uses approximate counting. Would that fit your bill?

LvanWissen:
Hi @piskvorky, I will keep an eye on Bounter, it looks promising! For now I will settle with an option outside of the class as suggested by @menshikh-iv. Many thanks for your comments!

PR #1928, opened 2018-02-24
Fix `test_similarities.py`

menshikh-iv:
What's done

- [x] Relax precision of similarity (fix wrong "exact" float comparison)
- [x] Replace FastTextWrapper to our naitive implementation

Related PR - https://github.com/MacPython/gensim-wheels/pull/9

### DIFF
--- a/gensim/test/test_similarities.py
+++ b/gensim/test/test_similarities.py
@@ -16,6 +16,7 @@ import os
 import numpy
 import scipy
 
+from smart_open import smart_open
 from gensim.corpora import Dictionary
 from gensim.models import word2vec
 from gensim.models import doc2vec
@@ -542,7 +543,7 @@ class TestWord2VecAnnoyIndexer(unittest.TestCase):
                 self.fn = fn
 
             def __iter__(self):
-                with open(self.fn) as infile:
+                with smart_open(self.fn, 'r', encoding="cp1251") as infile:
                     for line in infile:
                         yield line.lower().strip().split()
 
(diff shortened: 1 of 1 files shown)


gensim/test/test_similarities.py
```diff
@@ -538,12 +538,16 @@ def testWord2Vec(self):
...
-        if not ft_path:
-            return
-        corpus_file = datapath('lee.cor')
-        model = fasttext.FastText.train(ft_path, corpus_file)
+        class LeeReader(object):
+            def __init__(self, fn):
+                self.fn = fn
+
+            def __iter__(self):
+                with smart_open(self.fn, 'r', encoding="cp1251") as infile:
```
piskvorky: This doesn't look correct. Why do you think the encoding is `cp1251`?

menshikh-iv: Because UTF-8 doesn't work here, I pick `cp1251` randomly and this works fine.

piskvorky: Pick randomly? O_o

No, that encoding doesn't seem correct.

menshikh-iv:
```bash
ivan@P50:~/release/gensim$ file -i gensim/test/test_data/lee.cor 
gensim/test/test_data/lee.cor: text/plain; charset=iso-8859-1
```
[iso-8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1), although `cp1251` may not be correct, but it works quite correctly, surprisingly.

piskvorky: Not so surprising, both are 1 byte encodings. Please fix the encoding before we forget.

menshikh-iv: Done: https://github.com/RaRe-Technologies/gensim/pull/1931

PR #1809, opened 2017-12-20
Add new API-reference page, automated generation of `.rst` files, enable `sphinx-gallery`. Fix #1808

anotherbugmaster:
Fix #1808 

Reformatted API reference, but there's a little bit ugly workaround there: reference (like `[1]_`) doesn't work correctly if there's another reference with the same number in the upper section. I explored sklearn repo, it seems like they're aware of it and trying not to do so or making the same workaround as me.

[Autosummary sphinx module](http://www.sphinx-doc.org/en/1.5.1/ext/autosummary.html)

[Templating](http://www.sphinx-doc.org/en/1.5.1/templating.html#templating)

PR #1933, opened 2018-02-25
Refactor documentation for `gensim.models.coherencemodel`.

CLearERR:
@menshikh-iv


gensim/models/coherencemodel.py
```diff
@@ -84,21 +84,23 @@ class CoherenceModel(interfaces.TransformationABC):
 
     The main methods are:
 
-    1. constructor, which initializes the four stage pipeline by accepting a coherence measure,
-    2. the ``get_coherence()`` method, which returns the topic coherence.
+    1. Constructor, which initializes the four stage pipeline by accepting a coherence measure,
```
menshikh-iv: Useless construction, better to show it as example (not list some methods)

gensim/models/coherencemodel.py
```diff
@@ -215,7 +218,7 @@ def __init__(self, model=None, topics=None, texts=None, corpus=None, dictionary=
         self._topics = None
         self.topics = topics
 
-        self.processes = processes if processes > 1 else max(1, mp.cpu_count() - 1)
+        self.processes = processes if processes >= 1 else max(1, mp.cpu_count() - 1)
```
menshikh-iv: why?

gensim/models/coherencemodel.py
```diff
@@ -4,9 +4,9 @@
 # Copyright (C) 2010 Radim Rehurek <radimrehurek@seznam.cz>
 # Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
 
-"""
-Module for calculating topic coherence in python. This is the implementation of
+"""Module for calculating topic coherence in python. This is the implementation of
 the four stage topic coherence pipeline from the paper [1]_.
+
```
menshikh-iv: please use
```
`some text <http://...>`_
```
instead of `[1]_`, because `[num]` can't be parsed correctly with `autosummary` plugin (this related to #1809), here and everywhere

gensim/models/coherencemodel.py
CLearERR: Looks like misclick, will be fixed in next commit.

PR #1934, opened 2018-02-26
Fix OOV pairs counter in `WordEmbeddingsKeyedVectors.evaluate_word_pairs`

akutuzov:
`Evaluate_word_pairs` function returned incorrect number of out-of-vocabulary pairs when `dummy4unknown` parameter was set to True.


menshikh-iv:
Thanks @akutuzov, good catch!

PR #1935, opened 2018-02-26
Add `evaluate_word_analogies` (will replace `accuracy`) method for `gensim.models.KeyedVectors`

akutuzov:
The `accuracy` function evaluates the performance of word2vec models in analogy task. `restrict_vocab` parameter defines which part of the model vocabulary will be used for evaluation. The previous default was 30 000 top frequent words (analogy questions containing words beyond this threshold are simply skipped). It indeed makes sense to use some kind of limit here, as the evaluation running time depends on the size of the used vocabulary.
However, 30 000 is a very small value, with typical models nowadays featuring hundreds of thousands or even millions of words in their vocabularies. This leads to unrealistic evaluation scores, calculated only on small parts of a test set and a model. 
Therefore, I suggest increasing the default value of `restrict_vocab` 10-fold, up to 300 000. This will be more in line with the typical vocabulary size of contemporary word embedding models, and also will be consistent with the default `restrict_vocab` value for the `evaluate_word_pairs` function.

Note that although the original C word2vec does mention  30 000 as a good threshold value for analogies evaluation, the default behavior of its `compute-accuracy` executable is still not to use any threshold (=evaluate on the whole vocabulary).

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -50,7 +50,9 @@ Correlation with human opinion on word similarity::
 
 And on analogies::
 
-  >>> word_vectors.accuracy(os.path.join(module_path, 'test_data', 'questions-words.txt'))
+  >>> word_vectors.evaluate_word_analogies(os.path.join(module_path, 'test_data', 'questions-words.txt'))[0]
+  0.58
+  
 
 and so on.
 
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -859,7 +859,7 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
                 section['section'], 100.0 * correct / (correct + incorrect), correct, correct + incorrect
             )
 
-    def accuracy(self, questions, restrict_vocab=30000, most_similar=most_similar, case_insensitive=True):
+    def accuracy(self, questions, restrict_vocab=300000, most_similar=most_similar, case_insensitive=True):
         """
         Compute accuracy of the model. `questions` is a filename where lines are
         4-tuples of words, split into sections by ": SECTION NAME" lines.
@@ -898,17 +898,17 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
                 section = {'section': line.lstrip(': ').strip(), 'correct': [], 'incorrect': []}
             else:
                 if not section:
-                    raise ValueError("missing section header before line #%i in %s" % (line_no, questions))
+                    raise ValueError("Missing section header before line #%i in %s" % (line_no, questions))
                 try:
                     if case_insensitive:
                         a, b, c, expected = [word.upper() for word in line.split()]
                     else:
                         a, b, c, expected = [word for word in line.split()]
                 except ValueError:
-                    logger.info("skipping invalid line #%i in %s", line_no, questions)
+                    logger.info("Skipping invalid line #%i in %s", line_no, questions)
                     continue
                 if a not in ok_vocab or b not in ok_vocab or c not in ok_vocab or expected not in ok_vocab:
-                    logger.debug("skipping line #%i with OOV words: %s", line_no, line.strip())
+                    logger.debug("Skipping line #%i with OOV words: %s", line_no, line.strip())
                     continue
 
                 original_vocab = self.vocab
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -860,38 +860,60 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
             return score
 
     def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensitive=True, dummy4unknown=False):
-        """
-        Compute performance of the model on an analogy test set
-        (see https://aclweb.org/aclwiki/Analogy_(State_of_the_art)).
-        `analogies` is a filename where lines are 4-tuples of words, split into sections by ": SECTION NAME" lines.
-        See questions-words.txt in
-        https://storage.googleapis.com/google-code-archive-source/v2/code.google.com/word2vec/source-archive.zip
-        for an example.
+        """Compute performance of the model on an analogy test set.
 
-        The accuracy is reported (=printed to log and returned as a score) for each section separately,
-        plus there's one aggregate summary at the end.
+        The accuracy is reported (=printed to log and returned as a score)
+        for each section separately, plus there's one aggregate summary
+        at the end.
+        This method corresponds to the `compute-accuracy` script of the
+        original C word2vec.
 
-        Use `restrict_vocab` to ignore all 4-tuples containing a word not in the first `restrict_vocab`
-        words (default 300,000). This may be meaningful if you've sorted the model vocabulary by descending frequency
-        (which is standard in modern word embedding models).
+        Parameters
+        ----------
+        `analogies` is a filename where lines are 4-tuples of words,
+        split into sections by ": SECTION NAME" lines.
+        See questions-words.txt for an example:
+        `from gensim.test.utils import datapath
+        datapath("questions-words.txt")`
 
-        If `case_insensitive` is True, the first `restrict_vocab` words are taken first, and then
-        case normalization is performed.
-        Use `case_insensitive` to convert all words in 4-tuples and vocabulary to their uppercase form before
-        evaluating the performance (default True). Useful to handle case-mismatch between training tokens
-        and words in the test set. In case of multiple case variants of a single word, the vector for the first
-        occurrence (also the most frequent if vocabulary is sorted) is taken.
+        Use `restrict_vocab` to ignore all 4-tuples containing a word
+        not in the first `restrict_vocab` words (default 300,000). This
+        may be meaningful if you've sorted the model vocabulary by
+        descending frequency (which is standard in modern word embedding
+        models).
 
-        Use `dummy4unknown=True` to produce zero accuracies for 4-tuples with out-of-vocabulary words.
-        Otherwise (default False), these tuples are skipped entirely and not used in the evaluation.
+        Use `case_insensitive` to convert all words in 4-tuples and
+        vocabulary to their uppercase form before evaluating
+        the performance (default True). Useful to handle case-mismatch
+        between training tokens and words in the test set.
+        In case of multiple case variants of a single word, the vector
+        for the first occurrence (also the most frequent if vocabulary
+        is sorted) is taken. If `case_insensitive` is True, the first
+        `restrict_vocab` words are taken first, and then case normalization
+        is performed.
 
-        This method corresponds to the `compute-accuracy` script of the original C word2vec.
+        Use `dummy4unknown=True` to produce zero accuracies for 4-tuples
+        with out-of-vocabulary words. Otherwise (default False), these
+        tuples are skipped entirely and not used in the evaluation.
+
+        References
+        -------
+        See <https://aclweb.org/aclwiki/Analogy_(State_of_the_art)>.
+
+        Returns
+        -------
+        float
+            Overall evaluation score
+        list
+            Full lists of correct and incorrect predictions
+
         """
         ok_vocab = [(w, self.vocab[w]) for w in self.index2word[:restrict_vocab]]
         ok_vocab = {w.upper(): v for w, v in reversed(ok_vocab)} if case_insensitive else dict(ok_vocab)
         oov = 0
         logger.info("Evaluating word analogies for top %i most frequent words on %s", restrict_vocab, analogies)
         sections, section = [], None
+        line_no = 1
         for line_no, line in enumerate(utils.smart_open(analogies)):
             line = utils.to_unicode(line)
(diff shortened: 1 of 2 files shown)


menshikh-iv:
Hello @akutuzov,
I understand your point, sounds reasonable to me but looks like this change breaks backward compatibility (in the case when the user didn't specify `restrict_vocab` parameter). Why this important (I mean, why we should change the default value, a user can specify `300000` if user needs)?

CC: @piskvorky @gojomo @manneshiva

gojomo:
If making this change, people using the same data & same eval-method will, after an upgrade, get 'worse' scores - which is likely to cause alarm, confusion, and support requests. (It's not quite 'backward compaTIbility' that's being broken, but 'backward compaRAbility'.)

But the case for being more realistic, and especially matching the original `word2vec.c` `compute-accuracy` script behavior is reasonable. 

I'd suggest adding a new, fixed method that's more directly analogous to the `compare-accuracy` of `word2vec.c` – and using the 'right' default there, and suggesting it for future use. The older `accuracy()` could be marked deprecated. (The existing `accuracy()`/`log_accuracy()` split, and strange deep-nested-dictionary-of-sections reporting of `accuracy()`, might also be cleaned up at the same time to be more useful/descriptive for the common cases of just reporting on overall correct, or just syntactic/semantic correct.)

akutuzov:
May be adding a new correct method and marking `accuracy` as deprecated is indeed  a good idea. `accuracy` is a bad name for the analogy task evaluation anyway :-)

In fact, this silent skipping of OOV questions is wrong even if the threshold is very permissive (or even if there is no threshold at all). The current implementation (and the original word2vec implementation as well) allows to get high scores on Google Analogies even if your model's vocabulary contains only 10 words, for example. If these 10 words cover at least 1 question from the test set and produce the correct answer, the method will output the score of 100%. I think that the fair way to evaluate is to punish the models for lacking words from the test set. This is what the `dummy4unknown` parameter in the `evaluate_word_pairs` method does: if True, predictions of 0.0 are produced for OOV word pairs, making it impossible to achieve high scores with only a handful of known words. I think the same parameter should be added to the analogies evaluation method.
I even think that in the future this parameter should be set to True by default both for `evaluate_word_pairs` and for analogies evaluation (whatever its name will be), as it really ensures fair comparison of models. But this is a radical change not inherited from the original word2vec, so I would really like to hear what others think.

piskvorky:
That evaluation process was selected *precisely* to match `word2vec.c` behaviour. Did something change?

akutuzov:
As I've said above, the default `compute-accuracy` behavior is to use the threshold=0 (that is, no threshold at all). I quote directly from the comment in its source code:
_threshold is used to reduce vocabulary of the model for fast approximate evaluation (0 = off, otherwise typical value is 30000)_
30000 is only suggested, but by default the evaluation is done on the full model vocabulary. 
Thus, the current Gensim default setting does not mimic the C tool behavior anyway.

piskvorky:
Alright, thanks. That means `compute-accuracy` must have changed in the meanwhile.

If that's the case, mimicking its current behaviour is OK.

menshikh-iv:
@piskvorky @gojomo And what about the total, can I merge this one?

piskvorky:
I had a look and `compute-accuracy` [recommends 30,000](https://github.com/nicholas-leonard/word2vec/blob/master/compute-accuracy.c#L36).

Mikolov's word2vec paper [also used 30,000](https://arxiv.org/pdf/1301.3781.pdf).

If that is the established standard, I'm -1 on deviating at this point.

One other option would be to change the default to match the C tool default ("off"), or create a new, non-conflicting evaluation method/process.

akutuzov:
@piskvorky yes, `compute-accuracy` does recommend 30,000, but uses no threshold by default, as I've already mentioned.
Mikolov's paper evaluates on both 30,0000-limited vocabulary **and** on full vocabulary, see Tables 2 and 4.
It is difficult to tell what is the 'established standard' here.  Here is the quote from highly influential 2014 [paper](http://www.aclweb.org/anthology/P14-1023) by Baroni et al.:

_'because  of  the  way  the  task is framed,  performance also depends on the size of the vocabulary to be searched:  Mikolov et al. (2013a) pick the nearest neighbour among vectors for 1M words, Mikolov et al. (2013c) among 700K words, and we among 300K words.'_

I would say there is unfortunately no established standard here and most users will simply run evaluation script with default parameters.

Overall, I support the suggestion of @gojomo to implement new evaluation method and mark the current `accuracy` method as deprecated. This new method can be called something like `evaluate_word_analogies` to be more consistent with `evaluate_word_pairs`, it will feature a higher  default threshold and the `dummy4uknown` parameter (which probably at some point in the future can be set to True by default). It would also be great for it to have other methods for computing analogies, besides the currently used _3CosAdd_: for example, _3CosMul_ and set-based methods.

If everyone agrees to this plan, I can start implementing this new method.

piskvorky:
Sounds good to me, thanks for investigating @akutuzov .

gojomo:
Despite the word2vec.c precedent, calling this just `accuracy()` has always bugged me, and tends to mislead casual users to think this score is some ‘true north’ of vector quality, the one thing that makes one-set-of-parameters/vectors better than others. And, its name/prominence/ease-of-use forestalls consideration of the limitations of this evaluation, and the tradeoffs in using more/fewer of the most-frequent words.

For example, a plausible non-running-time reason for clipping the analogies evaluation to more-frequent words is that the ‘long tail’ of words includes many that might crowd-out the ‘right’ answer, without being wholly ‘wrong’. They may be near-synonyms of the ‘best’ answer, or just be idiosyncratically-placed because of their few training examples. But they still help on most tasks, even if they hurt analogies! And you might not want to discard them before training - because they still have other value, and perhaps even help improve the ‘fat head’ words. (For example, `model60kvocab.analogies_score(restrict_vocab=30000)` may score better than `model30kvocab.analogies_score()`, and be a fairer comparison, that’s also better correlated with word-vector-quality from the top to the bottom for more other tasks.)

menshikh-iv:
@akutuzov you'll make changes in current PR or in new? If in new - please close current PR.

akutuzov:
@menshikh-iv I think I will work in this PR

akutuzov:
OK, so as discussed before, I implemented a new `evaluate_word_analogies` method. In comparison to the `accuracy` method, it:

1. Defaults to 300,000 top frequent words from the model vocabulary, instead of 30,000.
2. Outputs the percentage of OOV quadruplets, and implements the `dummy4unknown` parameter which makes OOV quadruplets to yield zero accuracy when set to True.
3. Runs about 2 times faster, as it calculates similarities only for top 5 nearest neighbors, not for the whole vocabulary.
4. Has slightly updated docs and logging.

I marked the `accuracy` method as deprecated, but I am not sure I did it the right way, so I ask @menshikh-iv to double-check this.

gensim/models/keyedvectors.py
```diff
@@ -859,6 +967,7 @@ def log_accuracy(section):
                 section['section'], 100.0 * correct / (correct + incorrect), correct, correct + incorrect
             )
 
+    @deprecated("Method will be removed in 4.0.0, use self.evaluate_word_analogies() instead")
```
menshikh-iv: This is the correct way, all fine :+1:

gensim/models/keyedvectors.py
```diff
@@ -850,6 +851,113 @@ def n_similarity(self, ws1, ws2):
         v2 = [self[word] for word in ws2]
         return dot(matutils.unitvec(array(v1).mean(axis=0)), matutils.unitvec(array(v2).mean(axis=0)))
 
+    @staticmethod
+    def log_evaluate_word_analogies(section):
```
menshikh-iv: Maybe better hide this method (with `_`)?

gensim/models/keyedvectors.py
```diff
@@ -850,6 +851,113 @@ def n_similarity(self, ws1, ws2):
...
+        correct, incorrect = len(section['correct']), len(section['incorrect'])
+        if correct + incorrect > 0:
+            score = correct / (correct + incorrect)
+            logger.info("%s: %.1f%% (%i/%i)", section['section'], 100.0 * score, correct, correct + incorrect)
+            return score
+
+    def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensitive=True, dummy4unknown=False):
+        """
+        Compute performance of the model on an analogy test set
```
menshikh-iv: Please use numpy-style docstrings (http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html and https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)

gensim/models/keyedvectors.py
```diff
@@ -850,6 +851,113 @@ def n_similarity(self, ws1, ws2):
...
+        (see https://aclweb.org/aclwiki/Analogy_(State_of_the_art)).
```
menshikh-iv: this should be rendered as a link, should look like
```
 `Analogy (State of the art) <https://aclweb.org/aclwiki/Analogy_(State_of_the_art)>`_
```

gensim/models/keyedvectors.py
```diff
@@ -850,6 +851,113 @@ def n_similarity(self, ws1, ws2):
...
+        `analogies` is a filename where lines are 4-tuples of words, split into sections by ": SECTION NAME" lines.
+        See questions-words.txt in
```
menshikh-iv: this file also provided in the current repo: https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_data/questions-words.txt + this is part of the gensim package, i.e. path on local machine can be retrieved as 
```python
from gensim.test.utils import datapath
datapath("questions-words.txt")
```

No need to download source-code of C version for looking into this file

gensim/models/keyedvectors.py
```diff
@@ -850,6 +851,113 @@ def n_similarity(self, ws1, ws2):
...
+        total = {
+            'section': 'Total accuracy',
+            'correct': sum((s['correct'] for s in sections), []),
+            'incorrect': sum((s['incorrect'] for s in sections), []),
+        }
+
+        oov_ratio = float(oov) / line_no * 100
+        logger.info('Quadruplets with out-of-vocabulary words: %.1f%%', oov_ratio)
+        if not dummy4unknown:
+            logger.info('NB: analogies containing OOV words were skipped from evaluation! '
```
menshikh-iv: nitpick: please use hanging indents (instead of vertical)

gensim/models/keyedvectors.py
akutuzov: What exactly do you mean? Or may be you can point to some example of such hiding in the existing Gensim code?

gensim/models/keyedvectors.py
akutuzov: Done.

gensim/models/keyedvectors.py
akutuzov: Done

gensim/models/keyedvectors.py
akutuzov: Done.

gensim/models/keyedvectors.py
akutuzov: Done.

gensim/models/keyedvectors.py
menshikh-iv: I mean why not `_log_evaluate_word_analogies`? I asking because this method looks like a helper for `evaluate_word_analogies`, not more.

akutuzov: Done.

gensim/models/keyedvectors.py
```diff
@@ -850,6 +851,137 @@ def n_similarity(self, ws1, ws2):
...
+
+        The accuracy is reported (=printed to log and returned as a score)
+        for each section separately, plus there's one aggregate summary
+        at the end.
+        This method corresponds to the `compute-accuracy` script of the
+        original C word2vec.
+
+        Parameters
+        ----------
+        `analogies` is a filename where lines are 4-tuples of words,
```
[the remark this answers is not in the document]
akutuzov: Done.

gensim/models/keyedvectors.py
```diff
@@ -850,6 +851,137 @@ def n_similarity(self, ws1, ws2):
...
+        for the first occurrence (also the most frequent if vocabulary
+        is sorted) is taken. If `case_insensitive` is True, the first
+        `restrict_vocab` words are taken first, and then case normalization
+        is performed.
+
+        Use `dummy4unknown=True` to produce zero accuracies for 4-tuples
+        with out-of-vocabulary words. Otherwise (default False), these
+        tuples are skipped entirely and not used in the evaluation.
+
+        References
```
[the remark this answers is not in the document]
akutuzov: Done.

akutuzov:
@menshikh-iv  Is everything OK now?
(Travis build for Python 3.6 seems to get stuck on test_ldamodel.py, don't know why)

menshikh-iv:
@akutuzov I fixed Travis problem. Looks good to me (I'll slightly cleanup docstring & merge it soon), thanks for your work @akutuzov :+1:

PR #1957, opened 2018-03-06
Add `gensim.models.BaseKeyedVectors.add_entity` method for fill `KeyedVectors` in manual way. Fix #1942

### LINKED ISSUE #1942: Make add_word part of keyedvectors.py interface
Hi, I think it would be very useful to define `add_word` inside the class instead of inside the `load` function in [keyedvectors](https://github.com/RaRe-Technologies/gensim/blob/7127eeef8304852da589465b338e2bba14e881f1/gensim/models/keyedvectors.py).

Thanks ✌🏻

persiyanov:
Pull request related to [this](https://github.com/RaRe-Technologies/gensim/issues/1942) issue.

With these changes, it would be possible to add new word vectors into a `KeyedVectors` object.

@menshikh-iv please, take a look. I will write tests on it after addressing your comments.

Moreover, I have some questions/doubts:

1. **Method name**: is it okay right now or maybe it should be named like `add_entity/add_word`?
2. **Weights parameter**: should we check the type and shape of `weights` and raise an exception in a bad case?
3. **`self.vectors` contiguity**: [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/utils_any2vec.py#L231) `vectors` list is casted to C-contiguous array. Does numpy preserve C-contiguity after operations such as `np.vstack`? If so, I think I should add `numpy.ascontiguousarray` cast in my code.


menshikh-iv:
Hello @persiyanov 
1. I think both (`add_word` & `add_entity`) is OK for this case
2. I don't think so (because you have no predefined length if you'll create the empty class). Anyway, this needs only for some special cases.

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,19 @@ def get_vector(self, entity):
         else:
             raise KeyError("'%s' not in vocabulary" % entity)
 
+    def add(self, entity, weights):
```
menshikh-iv: What's about re-using this function in  (this is duplication right now from https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/utils_any2vec.py#L182)?

menshikh-iv: Also, need to add tests to check this functionality:
- load some kv, add more vectors in a manual way and check that this added fine
- create empty kv, fill it manually and check that all fine

persiyanov: @menshikh-iv 

About reusing this function:

It's a bit difficult because in this function `vstack` is used to append new word vector to `self.vectors`, while `add_word` in utils_any2vec creates `vectors` array at first and then it just inserts vectors into zeros array (https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/utils_any2vec.py#L197).

While it's possible to follow DRY here, the interface of `BaseKeyedVectors.add()` method will be more complicated (or I can change the logic in `utils_any2vec` -- not to create `vectors = np.zeros(...)` but append each word to the array, but it could decrease the performance of `load_word2vec_format` function). 

If some of these two options is okay, I'll implement it.

gensim/test/test_keyedvectors.py
```diff
@@ -153,6 +153,27 @@ def test_wv_property(self):
         """Test that the deprecated `wv` property returns `self`. To be removed in v4.0.0."""
         self.assertTrue(self.vectors is self.vectors.wv)
 
+    def test_add_word(self):
+        """Test that adding word in a manual way works correctly."""
+        from numpy.random import randn
```
menshikh-iv: can you remove import and use `np.random.randn`

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,28 @@ def get_vector(self, entity):
         else:
             raise KeyError("'%s' not in vocabulary" % entity)
 
+    def add_entity(self, entity, weights):
+        """Accept an entity specified by string tag and vector weights as 1D numpy array with shape (`vector_size`,).
```
menshikh-iv: Please use [numpy-style](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) docstrings

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,28 @@ def get_vector(self, entity):
...
+
+        if len(self.vectors) == 0:
+            self.vectors = [weights]
+        else:
+            self.vectors = vstack((self.vectors, weights))
+
+        self.vocab[entity] = Vocab(index=entity_id, count=1)
+        self.index2entity.append(entity)
+
+    def add_word(self, entity, weights):
```
menshikh-iv: Maybe this doesn't need?

gensim/models/keyedvectors.py
menshikh-iv: Aha, thanks for suggestion, let's stay it as is.

menshikh-iv:
CC: @gojomo wdyt?

gojomo:
This sort of functionality is a natural and useful addition; ideally it'd also be joined with other new APIs to assist initial building of a `KeyedVectors` from external data sources (rather than the just the direct property-tampering of current usage). 

There should be a bulk addition option: otherwise doing lots of single-adds in a loop requires many wasteful reallocations of 1-vector-larger-arrays. 

Should perhaps use `__setitem__()` rather than a named method, or at least have that as an idiomatic option. 

Longer-term, a `KeyedVectors` that actually splits its contents into multiple segments of smaller arrays for more efficient add/delete grow/shrinks would also make sense. But, that'd require more hiding of internal implementation details in a way that could break lots of the current direct-property-accesses by other code.  

re: @persiyanov your questions - (1) I'd avoid 'word' in any method names, to stay loyal to intent that this class accepts keys other than just words. (2) if the `vstack()` exception which results from attempting a mismatched shape is already sufficiently descriptive, no need for extra checking here... but if it's unclear/confusing, then a compat-shape-check would make sense. (3) I don't know but that's worth checking.

persiyanov:
@menshikh-iv @gojomo please, take a look.

I've implemented `add_entities` method and its alias `__setitem__`. Also, I've added bool flags which specifies vectors replacement strategy for those entities which are already in vocabulary.

persiyanov:
Ping @gojomo @menshikh-iv

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,75 @@ def get_vector(self, entity):
...
+
+    def add_entities(self, entities, weights, replace=False):
+        """Add entities and theirs vectors in a manual way.
+        If some entity is already in the vocabulary, old vector is keeped unless `replace` flag is True.
+
+        Parameters
+        ----------
+        entities : list of str
+            Entities specified by string tags.
+        weights: list of np.array or np.array
```
menshikh-iv: `{list of numpy.ndarray, numpy.ndarray}`

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,75 @@ def get_vector(self, entity):
         else:
             raise KeyError("'%s' not in vocabulary" % entity)
 
+    def add_entity(self, entity, weights, replace=False):
```
menshikh-iv: maybe remove this method (`add_entities` looks enough, wdyt @gojomo?)

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,75 @@ def get_vector(self, entity):
...
+
+        Parameters
+        ----------
+        entities : list of str
+            Entities specified by string tags.
+        weights: list of np.array or np.array
+            List of 1D np.array vectors or 2D np.array of vectors.
+        replace: bool, optional
+            Boolean flag indicating whether to replace vectors for entities which are already in the vocabulary.
+            Default, False, means that old vectors for those entities are keeped.
```
menshikh-iv: No need to duplicate "default" value for trivial case in docstring, maybe better to write something like

```
Flag indicating whether to replace vectors for entities which are already in the vocabulary, if True - replace vectors, otherwise - keep old vectors.

```

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,75 @@ def get_vector(self, entity):
...
+            self.vectors = weights[~in_vocab_mask]
+        else:
+            self.vectors = vstack((self.vectors, weights[~in_vocab_mask]))
+
+        # change vectors for in_vocab entities if `replace` flag is specified
+        if replace:
+            self.vectors[in_vocab_idxs] = weights[in_vocab_mask]
+
+    def __setitem__(self, entities, weights):
+        """Idiomatic way to call `add_entities` with `replace=True`.
```
menshikh-iv: better to write full docstring

menshikh-iv:
@gojomo please have a look (for me LGTM, except `add_entity` method, because I don't see any reason for this alias).

persiyanov:
ping @gojomo

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,84 @@ def get_vector(self, entity):
...
+            if True - replace vectors, otherwise - keep old vectors.
+        """
+        if isinstance(weights, list):
+            weights = np.array(weights)
+
+        in_vocab_mask = np.zeros(len(entities), dtype=np.bool)
+        in_vocab_idxs = []
+        out_vocab_entities = []
+
+        for idx, entity in zip(range(len(entities)), entities):
```
gojomo: `enumerate()` more idiomatic.

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,84 @@ def get_vector(self, entity):
...
+        for entity in out_vocab_entities:
+            entity_id = len(self.vocab)
+            self.vocab[entity] = Vocab(index=entity_id, count=1)
+            self.index2entity.append(entity)
+
+        # add vectors for new entities
+        if len(self.vectors) == 0:
+            self.vectors = weights[~in_vocab_mask]
+        else:
+            self.vectors = vstack((self.vectors, weights[~in_vocab_mask]))
```
gojomo: Might this line work even in the case where `len(self.vectors)==0`, making the check/branch unnecessary?

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,84 @@ def get_vector(self, entity):
...
+        replace: bool, optional
+            Flag indicating whether to replace vectors for entities which are already in the vocabulary,
+            if True - replace vectors, otherwise - keep old vectors.
+        """
+        if isinstance(weights, list):
+            weights = np.array(weights)
+
+        in_vocab_mask = np.zeros(len(entities), dtype=np.bool)
+        in_vocab_idxs = []
+        out_vocab_entities = []
```
gojomo: This method might be simpler without separate `in_vocab_idxs` and `out_vocab_entities` – just driving those ops from the mask, using options like `where()` or `nonzero()`.

gojomo:
Functionality seems good. 

I think class has a preexisting terminology issue with the use of word 'entity' where often 'key' would be more-consistent/more-specific. Also, as the 'items' inside this are definitionally 'vectors', generally `vector`/`vectors` better terms than `weights` . While I'm not sure this PR can/should fix all of that, I'd prefer `add_entity()` & `add_entities()` be replaced with a single `add(keys, vectors, replace=True)` (that could also for convenience tolerate a single key/vector).

gensim/models/keyedvectors.py
persiyanov: I think it's not obvious how to do that, because when empty `KeyedVectors` object is created, `self.vectors = []` is true. In that case, we can't use `vstack(([], weights[~in_vocab_mask]))` and `ValueError: all the input array dimensions except for the concatenation axis must match exactly` is raised.

persiyanov:
@gojomo @menshikh-iv please, take a look.

I've removed `add_entities & add_entity` but kept `add` method which can be used for all cases. Also, I got rid of `out_vocab_entities` variable, doing several `np.nonzero(...)` calls.

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,66 @@ def get_vector(self, entity):
...
+        Parameters
+        ----------
+        entities : list of str
+            Entities specified by string tags.
+        weights: {list of numpy.ndarray, numpy.ndarray}
+            List of 1D np.array vectors or 2D np.array of vectors.
+        replace: bool, optional
+            Flag indicating whether to replace vectors for entities which are already in the vocabulary,
+            if True - replace vectors, otherwise - keep old vectors.
+        """
```
menshikh-iv: nitpick: multiline docstring should ends with empty line, i.e.
```
"""
...
last text

"""
```

gensim/models/keyedvectors.py
```diff
@@ -154,6 +154,66 @@ def get_vector(self, entity):
...
+        if isinstance(entities, string_types):
+            entities = [entities]
+            weights = weights.reshape(1, -1)
```
menshikh-iv: probably, should be `weights = np.array(weights).reshape(1, -1)` for case if weights, for example, list of floats

persiyanov:
@menshikh-iv Fixed

gensim/models/keyedvectors.py
gojomo: Is it possible for an empty `KeyedVectors` to have a `self.vectors` that is already a proper-dimensioned `(0, vector_size)` empty ndarray? (Not sure myself, but would simplify things in later places like this.)

gensim/models/keyedvectors.py
```diff
@@ -73,7 +73,7 @@
     PYEMD_EXT = False
 
 from numpy import dot, zeros, float32 as REAL, empty, memmap as np_memmap, \
-    double, array, vstack, sqrt, newaxis, integer, \
+    double, array, zeros, vstack, sqrt, newaxis, integer, \
```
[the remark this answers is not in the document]
persiyanov: @gojomo yeah, i've fixed it

menshikh-iv:
Congratz with first time contribution @persiyanov :+1: :1st_place_medal:

gojomo:
@persiyanov Yes, and thanks for your patience through all the subtle refinements!

piskvorky:
Is the word `entity` still in? I'm strongly -1 on that, that should not be merged.

"Entity" has an established meaning in NLP. Introducing a new concept into gensim like this muddles the waters considerably. Especially since the concept is not really introduced at all in the PR AFAICS—what is an "entity" here? The docstrings that I can see only mention the word, never explain it.

gojomo:
The widespread use of 'entity' was introduced in #1777; this PR is just following the code's existing practice. The code desperately needs a refactoring for consistency/clarity, even moreso since the #1777 attempt!

menshikh-iv:
@piskvorky this modification for `BaseKeyedVectors` class (that doesn't say anything about words, only abstract "entities" that can be "words" in the subclasses).

Check out other code from this class, for example 

https://github.com/RaRe-Technologies/gensim/blob/2e08f4d3b218c9675d4f842f724af40a4f4ec1ee/gensim/models/keyedvectors.py#L124-L144

`BaseKeyedVectors` used as base class for `PoincareKeyedVectors` (where we have no "words")
https://github.com/RaRe-Technologies/gensim/blob/2e08f4d3b218c9675d4f842f724af40a4f4ec1ee/gensim/models/poincare.py#L772

piskvorky:
I agree with @gojomo , the code desperately needs a proper refactoring. The current post-#1777  situation is not tenable.

@menshikh-iv your snippet doesn't explain `entity` in any way, so not sure how that's helpful. I'm strongly -1 on introducing new fundamental concepts in this way.

menshikh-iv:
@piskvorky I do not explain it, I just show that the current code mimics the already existing and doesn't introduce any new concepts.

piskvorky:
Understood. It's an issue with #1777 , not this PR.

Issue #1654, opened 2017-10-25
Use Bounter for approx frequency counting


piskvorky:
Multiple models in gensim do a full corpus scan as their first step, to get the vocabulary frequencies / counts: word2vec, doc2vec, tfidf, phrases, make_wiki...

This step can require a lot of memory and be slow, because it's typically not parallelized.

Replace all such scanning by [Bounter](https://github.com/RaRe-Technologies/bounter). Let users specify how much memory they want to dedicate in an optional parameter, with some sane default like "1 GB" or "available 0.25 * total RAM" or something.

This is also a good place to revisit which algorithms need only the `counts['abc']` functionality, versus full `keys`/`items` iteration. The current implementations of counting in gensim are probably unnecessary demanding (use iteration), because there's no difference for dict or Counter as they support both operations.

But there is a significant difference for Bounter (counts-only is more efficient), and unless the algorithm in gensim *really needs* the keys, we should rewrite it using only `bounter(need_iteration=False)`.

aneesh-joshi:
Hey @menshikh-iv @piskvorky 
I would like to take this up.
(I hope it doesn't require any GPU support, etc.)

Could you direct me to a good point to start? 
I am familiar with `word2vec.py`

menshikh-iv:
Hello @aneesh-joshi,

1. This doesn't require GPU support, don't worry
2. Start from more simple variant - `gensim.corpora.Dictionary`, this looks like simplest variant.

aneesh-joshi:
hey @menshikh-iv 

I am trying to apply bounter to [gensim.corpora.dictionary](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/corpora/dictionary.py) and had some doubts

https://github.com/RaRe-Technologies/gensim/blob/f9669bb8a0b5b4b45fa8ff58d951a11d3178116d/gensim/corpora/dictionary.py#L203

implies that the it will return index, frequency

But the example shows 
https://github.com/RaRe-Technologies/gensim/blob/f9669bb8a0b5b4b45fa8ff58d951a11d3178116d/gensim/corpora/dictionary.py#L227

frequency as 1 although mama is there twice

Is this a documentation error or am I missing something?


Also, I cannot understand where I should implement bounter. Could you give me a specific idea?

Thanks!

menshikh-iv:
@aneesh-joshi this is frequency in "current" document, i.e.

```python
[...]

PR #1962, opened 2018-03-07
Adresses #1654 Use Bounter for approx frequency counting

aneesh-joshi:
Here is a basic implementation of bounter on `word2vec.py`
It handles the `scan_vocab` function for frequency counts

I had to specify a new `utils` function as it wasn't possible to use the current function with bounter

Here is the code I used to evaluate:

```
from nltk.corpus import brown
import logging

sentences = brown.sents() + brown.sents()
# sentences = sentences[:10]


model =  word2vec.Word2Vec(size=10, sentences=sentences)
```


gensim/models/word2vec.py
```diff
@@ -105,6 +105,7 @@
 import sys
 import os
 import heapq
+from bounter import bounter
```
menshikh-iv: you need to install it first (modify `setup.py` file)

gensim/models/word2vec.py
```diff
@@ -1146,7 +1147,9 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
         sentence_no = -1
         total_words = 0
         min_reduce = 1
+        use_bounter = True
```
menshikh-iv: why? this should be parametrized (same question about parameter lower)

gensim/utils.py
```diff
@@ -1708,6 +1708,36 @@ def prune_vocab(vocab, min_reduce, trim_rule=None):
     )
     return result
 
+def prune_bounter_vocab(counts, min_reduce, trim_rule=None):
```
menshikh-iv: what's a difference between `prune_vocab` and current implementation?

gensim/models/word2vec.py
```diff
@@ -1160,15 +1163,37 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
...
                 )
-            for word in sentence:
-                vocab[word] += 1
+
             total_words += len(sentence)
 
-            if self.max_vocab_size and len(vocab) > self.max_vocab_size:
-                utils.prune_vocab(vocab, min_reduce, trim_rule=trim_rule)
-                min_reduce += 1
+            if use_bounter:
```
menshikh-iv: bounter should be "drop-in" replacement (i.e. `if/else` looks suspicious here)

aneesh-joshi: does that mean the user should not be given a choice about whether bounter should be used.
(Especially considering bounter gives an approximate count)

menshikh-iv: No, the user must choose what's he needed explicitly.

aneesh-joshi:
@menshikh-iv 
Made changes such that it can be called with the model like:
`model =  word2vec.Word2Vec(size=10, sentences=sentences, bounter_size=1024)`

Will handle backward compatibility once changes are approved.

menshikh-iv:
@aneesh-joshi please make several things
- merge fresh develop to your branch
- make benchmark (train model on the mid-size corpus (maybe a 500,000 articles sample from Wikipedia) and compare the speed of training, memory consumption & evaluation results).

menshikh-iv:
CC: @gojomo wdyt about this PR?

gojomo:
A bunch of `if bounter_size is not None:` branches is a messy way to enable/disable this. I suspect a proper refactoring would separate out the exact-count (traditional) and approximate-bounded-memory-count (bounter) processes into separate functional classes, with compatible interfaces, allowing Word2Vec/etc to be indifferent as to which implementation is in use. (The prior introduction of a `Word2VecVocab`-specific class may have made the right abstraction/decoupling harder to achieve rather than easier.)

menshikh-iv:
@gojomo I agree, this is a good idea to move distinctly counters that will be available for all *2vec models (I'm not sure about distinct class, but definitely need to move in this direction).

About performance: @aneesh-joshi can you benchmark it please (what's a speedup we receive from this change) on medium/big corpuses?

aneesh-joshi:
hey @menshikh-iv 
working on the benchmarking. Trying to find a good corpus. Wikipedia will take too long for preprocessing.
Might use twitter or shakespeare from nltk.

menshikh-iv:
Wikipedia is good enoug I think (with tokenizing and lowercasing, that’s enough), you can use https://github.com/RaRe-Technologies/gensim-data/releases/tag/wiki-english-20171001 as source (and slightly preprocess it)

piskvorky:
Decoupling the counting behind interfaces was the original idea, yes.

@gojomo how did `Word2VecVocab` make the decoupling harder? I haven't followed the recent refactorings closely, but that's the opposite of what we wanted to achieve there. CC @manneshiva @janpom 

Performance (time) is not a terribly interesting metric, at least not as long as Gensim uses only a single stream as input. These Bounter changes should help primarily with memory, so we can get rid of the current thresholding/merging complexity.

gojomo:
I listed some doubts about the refactoring that added lines/classes/files/indirection without any significant new functionality (or demos of things that were now clearly easier) in comments starting at <https://github.com/RaRe-Technologies/gensim/issues/1623#issuecomment-363172999>.  `Word2VecVocab` sticks out to me as a very-specialized (rather than generically-reusable), very tightly-coupled, and somewhat confusingly-named class. (There's at least distinct-but-related things, in close proximity, going by the same abbreviation 'vocab'.)

gensim/models/word2vec.py
```diff
@@ -1158,18 +1171,37 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
...
 
+            if bounter_size is not None:
+                word_counts.update(sentence)
+            else:
+                for word in sentence:
+                    vocab[word] += 1
+
             if self.max_vocab_size and len(vocab) > self.max_vocab_size:
+                if bounter_size is not None:
+                    vocab = dict(word_counts)
```
gojomo: People are unlikely to use both `max_vocab_size` & a bounded-counter... but *if* they did, this branch could be hit many times during a scan, forcing multiple conversions-back-to-a-normal-dict. And after being hit once, what does it mean that both 'word_counts' and 'vocab' exist, with different sets of words? And then a few lines later, any pruned `vocab` is clobbered with another `word_counts`-conversion. So this branch is currently nonsensical.

menshikh-iv:
Hi @aneesh-joshi, how is going? 

@gojomo @piskvorky what we'll plan to do with this PR?

piskvorky:
The integration needs to be a little smoother, as outlined in the comments above. A series of `is None` switches are too messy.

Using a dict-like interface (whether dict or bounter) is the way to go here.

In the evaluation above, I'm missing the baseline of "dict"—something to compare the bounter numbers against (RAM, CPU time, wallclock time).

aneesh-joshi:
@menshikh-iv 
Hi
I will complete this PR. But i am currently travelling which is making it hard.
I will continue the work in 4 days.

aneesh-joshi:
@piskvorky @menshikh-iv @gojomo 

So, if I understand correctly,
>Refactoring the vocabulary counting/filtering to stay behind a clean dict-like interface (whether implemented by dict or bounter or something else) is the way to go here.

some sort of new wrapper object needs to be created to prevent the if branches


As to 
>In the evaluation above, I'm missing the baseline of "dict = the current implementation"—something to compare the bounter numbers against (RAM, CPU time, wallclock time).

Are there any performance testing tools you have in mind?
It get's a little difficult for doing these tests on an 8GB RAM, considering the corpus can be huge.

aneesh-joshi:
There seems to be some problem with my `git`ting  which has lead to the gitignore being added.
Can't figure out why.

aneesh-joshi:
ping @menshikh-iv

gensim/utils.py
```diff
@@ -381,6 +383,35 @@ def call_on_class_only(*args, **kwargs):
     raise AttributeError('This method should be called on a class object.')
 
 
+class DictWrapper():
```
menshikh-iv: missing docstrings

gojomo: 'Wrapper' is not very descriptive of its role/functionality. Seems it should be 'Counter' – and maybe even just using the Python built-in `Counter`?

gensim/utils.py
```diff
@@ -381,6 +383,35 @@ def call_on_class_only(*args, **kwargs):
...
+    def __init__(self, bounter_size):
+        self.word_counts = bounter(size_mb=bounter_size)
+
+    def __len__(self):
+        return self.word_counts.cardinality()
+
+    def update(self, sentence):
+        self.word_counts.update(sentence)
+
+    def get(self):
```
gojomo: Something like `as_dict()` would be more descriptive/accurate here.

gensim/models/word2vec.py
```diff
@@ -1174,18 +1185,30 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
...
                 )
-            for word in sentence:
-                vocab[word] += 1
+
             total_words += len(sentence)
+            wrapped_dict.update(sentence)
 
-            if self.max_vocab_size and len(vocab) > self.max_vocab_size:
+            if self.max_vocab_size and len(wrapped_dict) > self.max_vocab_size:
+                vocab = wrapped_dict.get()
```
gojomo: This looks very, very wrong. It's performing the exact same crude, slow premature pruning-to-save-memory when using a Bounter as when using a dictionary. And that pruning requires converting the whole Bounter to a memory-consuming regular dict each prune. And then the results of each costly prune are ignored, because nothing's done with the 'vocab' local. So this seems to break `max_vocab_size` for the old case, while retaining it as the source of a time- and space-consuming no-op in the bounter case. To truly test/evaluate a change here will require some testing (ad hoc or unit) of cases where `max_vocab_size` is operative, including timing & peak memory effects of the change vs original implementation.

gensim/utils.py
aneesh-joshi: @gojomo 
I'm sorry, I don't understand what you mean by
> Seems it should be 'Counter' – and maybe even just using the Python built-in Counter

especially about the "Python built-in Counter"
Please could you clarify that?

Would a name like `DictCounter` and `BounterCounter` be more suitable?

gensim/models/word2vec.py
aneesh-joshi: Ok
I will think about this and work on it.

gensim/utils.py
gojomo: https://docs.python.org/2/library/collections.html#collections.Counter

menshikh-iv:
Unfortunately, current PR is abandoned, for this reason, I close it.

piskvorky:
What was missing here?
@aneesh-joshi @menshikh-iv @gojomo can you comment with a short summary, so people who come to this PR in the future understand what happened and how to pick it up? Thanks.

menshikh-iv:
So, what I see from my PoV

1. (not really related with current PR, but **mandatory** condition) Build wheels for `bounter` (otherwise, gensim wheel distribution have no sense). Need to create `MacPython/Bounter` and `conda-forge/bounter-feedstock` & build wheels/packages for same versions as gensim support. Also, need to add Appveyor directly to `bounter` repo
2. Testing
3. Integration with other *2vec models
4. Integration with `gensim.corpus.Dictionary`

Just an idea: maybe we should split *2vec models in same way as topic models (i.e. Dictionary + Model itself), now we already have something similar internally, but we have no way to construct *2vec model using some predifined vocab.

aneesh-joshi:
@menshikh-iv 
Sorry for the delay. I would like to not abandon the PR. Just take some time with it. But I would welcome others working on it.

I made a wheel for it [here](https://drive.google.com/drive/folders/1-tMkwcOQFGDNXr2oUyASVcJq_K9NZHsZ?usp=sharing)

I made it using command: `python3 setup.py sdist bdist_wheel`

Is this sufficient?

gensim/models/word2vec.py
```diff
@@ -1146,12 +1150,13 @@ def __iter__(self):
...
 class Word2VecVocab(utils.SaveLoad):
     def __init__(self, max_vocab_size=None, min_count=5, sample=1e-3, sorted_vocab=True, null_word=0,
-        max_final_vocab=None):
+                 bounter_size=None, max_final_vocab=None):
         self.max_vocab_size = max_vocab_size
         self.min_count = min_count
         self.sample = sample
         self.sorted_vocab = sorted_vocab
         self.null_word = null_word
+        self.bounter_size = bounter_size
```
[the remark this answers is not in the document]
aneesh-joshi: >  (in the same way as for gensim in https://github.com/MacPython/gensim-wheels) 
I have made a wheel with Ubuntu 18.
For making MacPython wheel I build on a Mac and push to MacPython repo?

aneesh-joshi:
Current status of PR:
There is one bug relating to vocab pruning:

> This looks very, very wrong. It's performing the exact same crude, slow premature pruning-to-save-memory when using a Bounter as when using a dictionary. And that pruning requires converting the whole Bounter to a memory-consuming regular dict each prune. And then the results of each costly prune are ignored, because nothing's done with the 'vocab' local. So this seems to break max_vocab_size for the old case, while retaining it as the source of a time- and space-consuming no-op in the bounter case. To truly test/evaluate a change here will require some testing (ad hoc or unit) of cases where max_vocab_size is operative, including timing & peak memory effects of the change vs original implementation.

The bug was an oversight on my part. 

From the time I wrote the PR, there might have been changes in word2vec.py which will need to be checked. I will need some time to check this.

I suppose if a proof of concept can be made for word2vec, we can extend it to *2vec.

menshikh-iv:
@aneesh-joshi 

> I would like to not abandon the PR

Feel free to re-open and continue when you'll have a time :+1: 

> I made a wheel for it here

That's cool, but not really what we needed (we support 3 OS & 4 Python version and need to build wheel for each of it automatically). For bounter, we need to create similar repo as for gensim - https://github.com/MacPython/gensim-wheels and setup all CI testing. I'll do it myself when we needed (i.e. bounter included in gensim as deps)

PR #1966, opened 2018-03-08
Fix test for `gensim.models.KeyedVectors.similarity_matrix` method. Fix #1961

### LINKED ISSUE #1961: similarity_matrix: unit tests not run
#### Description
The unit tests for the gensim.models.keyedvectors similarity_matrix cannot currently be found and are not being run.

https://github.com/RaRe-Technologies/gensim/blob/f9669bb8a0b5b4b45fa8ff58d951a11d3178116d/gensim/test/test_keyedvectors.py#L30

Fix is underway: 
https://github.com/nataliedurgin/gensim/commit/5bb6aea3a92b8e7ff7cecfca054a6057bb9b8fa8

nataliedurgin:
Resolves #1961.

Changes allow the similarity_matrix unit tests to be found and run.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -511,6 +511,7 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
             if w1 not in self.vocab:
                 num_skipped += 1
                 continue  # A word from the dictionary is not present in the word2vec model.
+
             # Traverse upper triangle columns.
             if matrix_order <= nonzero_limit + 1:  # Traverse all columns.
                 columns = (
--- a/gensim/test/test_keyedvectors.py
+++ b/gensim/test/test_keyedvectors.py
@@ -27,19 +27,19 @@ class TestEuclideanKeyedVectors(unittest.TestCase):
         self.vectors = EuclideanKeyedVectors.load_word2vec_format(
             datapath('euclidean_vectors.bin'), binary=True, datatype=np.float64)
 
-    # TODO: These tests are not currently being run
     def test_similarity_matrix(self):
         """Test similarity_matrix returns expected results."""
 
-        corpus = [["government", "denied", "holiday"], ["holiday", "slowing", "hollingworth"]]
-        dictionary = Dictionary(corpus)
-        corpus = [dictionary.doc2bow(document) for document in corpus]
+        documents = [["government", "denied", "holiday"],
+                  ["holiday", "slowing", "hollingworth"]]
+        dictionary = Dictionary(documents)
 
         # checking symmetry and the existence of ones on the diagonal
         similarity_matrix = self.vectors.similarity_matrix(dictionary).todense()
         self.assertTrue((similarity_matrix.T == similarity_matrix).all())
-        # TODO: Nonsense. How is this testing for ones on the diagonal?
-        # self.assertTrue((np.diag(similarity_matrix) == similarity_matrix).all())
+        self.assertTrue(
+            (np.diag(similarity_matrix) ==
+             np.ones(similarity_matrix.shape[0])).all())
 
         # checking that thresholding works as expected
         similarity_matrix = self.vectors.similarity_matrix(dictionary, threshold=0.45).todense()
@@ -55,6 +55,8 @@ class TestEuclideanKeyedVectors(unittest.TestCase):
 
         similarity_matrix = self.vectors.similarity_matrix(dictionary, nonzero_limit=3).todense()
         self.assertEquals(20, np.sum(similarity_matrix == 0))
+
+        # TODO: Add unit test to check that supplied tfidf has desired effect
 
     def test_most_similar(self):
         """Test most_similar returns expected results."""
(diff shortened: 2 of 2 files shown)


menshikh-iv:
CC: @Witiko

gensim/test/test_keyedvectors.py
```diff
@@ -27,33 +27,37 @@ def setUp(self):
...
 
         # checking that thresholding works as expected
-        similarity_matrix = self.similarity_matrix(corpus, dictionary, threshold=0.45).todense()
+        similarity_matrix = self.vectors.similarity_matrix(dictionary, threshold=0.45).todense()
         self.assertEquals(18, np.sum(similarity_matrix == 0))
 
         # checking that exponent works as expected
-        similarity_matrix = self.similarity_matrix(corpus, dictionary, exponent=1.0).todense()
+        similarity_matrix = self.vectors.similarity_matrix(dictionary, exponent=1.0).todense()
         self.assertAlmostEqual(9.5788956, np.sum(similarity_matrix))
```
menshikh-iv: Better to add `places=5` to relax threshold (to avoiding FP failures like this one https://ci.appveyor.com/project/piskvorky/gensim-431bq/build/1.0.1883/job/03r7weolpj81olp7)

gensim/test/test_keyedvectors.py
```diff
@@ -27,33 +27,37 @@ def setUp(self):
...
         # checking that nonzero_limit works as expected
-        similarity_matrix = self.similarity_matrix(corpus, dictionary, nonzero_limit=4).todense()
+        similarity_matrix = self.vectors.similarity_matrix(dictionary, nonzero_limit=4).todense()
         self.assertEquals(4, np.sum(similarity_matrix == 0))
 
-        similarity_matrix = self.similarity_matrix(corpus, dictionary, nonzero_limit=3).todense()
+        similarity_matrix = self.vectors.similarity_matrix(dictionary, nonzero_limit=3).todense()
         self.assertEquals(20, np.sum(similarity_matrix == 0))
 
+        # TODO: Add unit test to check that supplied tfidf has desired effect
```
menshikh-iv: we are waiting for :)

gensim/test/test_keyedvectors.py
```diff
@@ -27,33 +27,37 @@ def setUp(self):
...
             datapath('euclidean_vectors.bin'), binary=True, datatype=np.float64)
 
-    def similarity_matrix(self):
+    def test_similarity_matrix(self):
         """Test similarity_matrix returns expected results."""
 
-        corpus = [["government", "denied", "holiday"], ["holiday", "slowing", "hollingworth"]]
-        dictionary = Dictionary(corpus)
-        corpus = [dictionary.doc2bow(document) for document in corpus]
+        documents = [["government", "denied", "holiday"],
```
menshikh-iv: nitpick: we using 120-characters limit, no need to move part of `documents` to new line

gensim/models/keyedvectors.py
```diff
@@ -511,6 +511,7 @@ def similarity_matrix(self, dictionary, tfidf=None, threshold=0.0, exponent=2.0,
             if w1 not in self.vocab:
                 num_skipped += 1
                 continue  # A word from the dictionary is not present in the word2vec model.
+
```
menshikh-iv: unrelated change

Witiko:
Looks ok to me except for the exponent test, which requires smaller precision (`places=…`) in order to pass.

nataliedurgin:
Regarding the tfidf test: It seems like the test should highlight that fewer operations are being performed when the tfidf model is supplied. Does python have a good way to measure such a thing (it seems that things like timeit are not exactly what we want here)? If python does, what would be the smallest example for which it would make a difference? Does the benefit kick in for the common_corpus in any measurable way? Are there examples of similar testing scenarios in the repository that I could look at?

Witiko:
> A low-hanging test would be to check that the output is consistent whether or not the tfidf is supplied.

I suggest that the test should check that the output is *not* consistent. Different processing order of the lines will yield a different similarity matrix:

tfidf_test

nataliedurgin:
Cool. I beg your patience for a few more questions so I can make sure I understand the big picture. If we are doing reordering of any kind (whether rows or columns) is it correct to say that before we go to compute softcossim here,

https://github.com/RaRe-Technologies/gensim/blob/2ee7a2c3b376d62020992f4b31bdaa79cf9de8db/gensim/matutils.py#L825

both the documents (vec1, vec2) and the similarity matrix (both rows and columns) need to have a consistent notion of term/feature order? If so, is that part of what this block of code is meant to accomplish through word_indices?   

https://github.com/RaRe-Technologies/gensim/blob/2ee7a2c3b376d62020992f4b31bdaa79cf9de8db/gensim/matutils.py#L810

Witiko:
It is assumed that the term ids are consistent between the vectors and the similarity matrix, i.e. that they were both produced using the same dictionary. With `vec1 = dict(vec1)`, I am just extracting the set `word_indices` of all term ids in the word vectors. Later in the `softcossim` function, I take a small slice of the term similarity matrix that corresponds to just the terms in the two vectors.

This is for performance reasons, because scipy's `x^T * S * y` is suboptimal and computes `x^T * S` for all columns even though most corresponding elements in `y` are zero.

Witiko:
Do you want to finish the PR, @nataliedurgin, or should I add the finishing touches?

nataliedurgin:
I think I will have some time today/tonight, but if I don't have it by tomorrow morning, feel free to finish it off.

Witiko:
Ok, I will finish the missing test during the weekend.

menshikh-iv:
@nataliedurgin continued by @Witiko in #1984.

PR #1968, opened 2018-03-09
Fix inheritance chain for `load_word2vec_format` (return correct class in case when you create an child class based on kv)

DennisChen0307:
If I create a child class from Word2VecKeyedVectors/KeyedVector this method will always return a instance of its parent class, which is not correct


menshikh-iv:
Nice catch, thank you @DennisChen0307, congratz with the first contribution :1st_place_medal:

Issue #1159, opened 2017-02-22
Efficient vector representation for documents through corruption


Threynaud:
I came across [this paper ](https://openreview.net/pdf?id=B1Igu2ogg) and I think it is a very promising alternative to Doc2Vec which doesn't scale at all with the size of the corpus !

The model architecture seems very similar to Word2Vec and it would be a nice addition to Gensim.
If you think it is a good idea, I would be glad to code a PoC for it!

tmylk:
Thanks for the link. There should at least be an integration with gensim of this code. Hoping the code will appear soon in this github repo:

https://github.com/mchen24/iclr2017/issues/1

It is indeed very interesting way to expand on the paragaraph vectors and tf-idf-averaged word2vec. Would be good to compare it to tf-idf -averaged word2vec(not just idf) when the code becomes available.

Threynaud:
I sent her an email about this, I'll keep you in touch! If she doesn't plan to make it public in a near future I'll try to implement it!

ChijunZhang:
Author made code public 6 hours ago

Threynaud:
Looks like it is directly integrated in Mikolov's code but I don't feel like I can make the change in Gensim by myself, I am too bad with Cython..

gojomo:
From a quick glance at the paper, the idea of composing a document-vector from a (random subset) of whole-doc words made sense to me and would be pretty easy to add as an option in gensim. However, I found the paper's description of the drop-out "randomly overwrites each dimension of the original document x with probability q" process a bit unclear – does it only happen to the composed doc-vec? Or the combined doc-vec + neighboring-words-context? (The paper also mentions "each word in a document is corrupted independently of others", which as far as I can tell doesn't match the earlier descriptions of leaving-out-words or zeroing-out-dimensions-of-the-docvec.) 

But, doing a diff of the https://github.com/mchen24/iclr2017 code & Mikolov's "--sentence-vectors" word2vec.c patch should precisely clarify. 

The paper's results on IMDB sentiment-prediction are only very-slightly better than plain Paragraph Vectors, though on classification and semantic-relatedness, there seem to be bigger improvements. However, the exact metaparameters used aren't specified, so it's unclear if perhaps the new method, highly parameter optimized, is being compared against just one set of PV parameters. Knowing for sure the method often/usually helps should be a prerequisite before adding complexity to the existing PV-Doc2Vec code.

Threynaud:
[...]

PR #1985, opened 2018-03-19
[WIP] Doc2VecC ("Doc2Vec with Corruption")

Saftophobia:
This is a Work-in-Progress on #1159 

> Doc2vecC (Corruption) represents each document as a simple average of word embeddings. A corruption model is included, which introduces a data-dependent regularization that favors informative or rare words while forcing the embeddings of common and non-discriminative ones to be close to zero. - Minmin Chen, 2017 

While training a word2vec model, a percentage of the document (noted by the variable `doc2vecc`) is randomly sampled. The average of these word vectors is used as a sentence vector along with the center word to predict the context (in case of Skip-gram model). Since word vectors incorporate some global information while training, averaging word vectors within a document yields better document embeddings (as shown in the [paper](https://openreview.net/pdf?id=B1Igu2ogg)) than Doc2Vec on IMDB dataset. 

- I outlined the changes in the C code in this [gist](https://gist.github.com/Saftophobia/7c440dad6a055883c00e72ba415efc8e).


menshikh-iv:
Good start, thanks @Saftophobia, @gojomo let's discuss integration, I think that's better to add it as +1 mode for Doc2Vec (instead of new class) wdyt?

gojomo:
To the extent new variants can be just another (flag/parameter/pluggable-policy) mode of a shared meta-algorithm is ideal – so long as it doesn't add too much complexity to the code or overhead for people not using the new variants. But the core Doc2Vec/Word2Vec stuff is getting cruftier & harder-to-extend barring some major refactoring. (A new mode like 'Doc2VecC' could be one of the test cases for whether a new arrangement is making new functionality easier.)

gensim/models/word2vec.py
```diff
@@ -144,7 +144,7 @@
     FAST_VERSION = -1
     MAX_WORDS_IN_BATCH = 10000
 
-    def train_batch_sg(model, sentences, alpha, work=None, compute_loss=False):
+    def train_batch_sg(model, sentences, alpha, work=None, compute_loss=False, doc2vecC=None):
```
menshikh-iv: please use lowercase for parameter name

gensim/models/word2vec.py
```diff
@@ -164,14 +168,25 @@ def train_batch_sg(model, sentences, alpha, work=None, compute_loss=False):
                 for pos2, word2 in enumerate(word_vocabs[start:(pos + model.window + 1 - reduced_window)], start):
                     # don't train on the `word` itself
                     if pos2 != pos:
-                        train_sg_pair(
-                            model, model.wv.index2word[word.index], word2.index, alpha, compute_loss=compute_loss
-                        )
+
+                        # handle the sentence using Doc2VecC
+                        if doc2vecC:
+                            doc2vecc_indices = [w.index for w in random.choice(word_vocabs,
```
menshikh-iv: nitpick: use hanging indents (as from line 177) instead of vertical

menshikh-iv:
Ping @Saftophobia, do you plan to finish PR?

menshikh-iv:
Looks abandoned, for this reason, I close this PR.
@Saftophobia feel free to re-open when you'll have time to continue.

Saftophobia:
Just saw https://github.com/RaRe-Technologies/gensim/issues/1159#issuecomment-432083694, I apologise for abandoning this, got distracted with publishing/traveling/work/etc. Will continue working on this in (~2nd week) November. I already updated the branch with the latest `develop` changes and tended to the comments.

I didn't get https://github.com/RaRe-Technologies/gensim/pull/1985#issuecomment-374321316 very well. Should I go with the +1 mode for doc2vecC in the `word2vec.py` class? or create a separate class?

menshikh-iv:
@Saftophobia great :+1:

> I didn't get #1985 (comment) very well. Should I go with the +1 mode for doc2vecC in the word2vec.py class? or create a separate class?

If this possible to add like a "mode" of current d2v - no need to create distinct class for it (i.e. continue what you do before).


Just re-open PR (not create new one) when you are will be ready for review :+1:

Saftophobia:
Hi @menshikh-iv, could you please re-open the PR? I made sure all tests are working. 

Doc2vecC trains word vectors with the corruption trick using the word2vec architecture. Users will have to average the word vectors within a document to construct the document vector. 

Would be great if you have a closer look at the changes, if you can propose a way to test doc2vecC I would be very grateful. (Would it make sense to compare it with the C output?).

Saftophobia:
@menshikh-iv @gojomo I believe I'm done. I'll be working on a new notebook, tend to your PR comments and compare its performance against `Doc2Vec` and `Word2Vec+AVG`.

docs/notebooks/doc2vecc_newsgroup20.ipynb
```diff
@@ -0,0 +1,372 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 141,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "def infer_vectors(model, documents):\n",
```
gojomo: I'd avoid calling this `infer_vector()`, since it's just using a simple average, rather than an iterative 'inference' process (like the same-named `Doc2Vec` method). Also: an 'average_vector(str[])` utility method probably belongs on `Word2Vec` or one of its parent interfaces, since it's such a common need. It might already exist or be factorable-out-of one of the existing methods that already averages-several-words-together. It could potentially include optional parameters to limit the considered words to those above a certain frequency or in the top-N of all words, or toggle on or off vector norming before or after averaging.

docs/notebooks/doc2vecc_newsgroup20.ipynb
```diff
@@ -0,0 +1,372 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "print(\"Doc2vecC with 0.2 corruption classified:\\t %i correctly,\\t %f%%\" % (doc2vecc_correctly_classified, doc2vecc_correctly_classified/len(df_positive_negative) * 100))\n",
+    "print(\"Word2Vec classified:\\t\\t\\t %i correctly,\\t %f%%\" % (word2vec_correctly_classified, word2vec_correctly_classified/len(df_positive_negative) * 100))\n",
+    "print('-'*75)\n",
+    "print(\"Doc2vecC with 0.2 corruption average distance: \\t %f\" % doc2vecc_avg_distance)\n",
+    "print(\"Word2Vec average distance:\\t\\t\\t %f\" % word2vec_avg_distance)\n"
+   ]
```
gojomo: Results so far seem to suggest this mode hurts classification accuracy. Do you suspect some other refinement or alternative measure will show an advantage of this technique?

GeneralSemantics: Was 0.2 corruption the best-performing of several values tried, or just the first one tested so far?  I think the author said the corruption rate should scale with document length, and be closer to 1 for longer documents:

https://github.com/RaRe-Technologies/gensim/issues/1159#issuecomment-297901638

Saftophobia:
### Test summary

- This test retrieves [20newsgroup](http://qwone.com/~jason/20Newsgroups/) dataset, creates a smaller dataset of triplets in this format:
     - Document A 
     - Document B sampled from the same category as A's
     - Document C randomly sampled from the dataset with category different than A's. 

- What I'm trying to compare against:
    - Percentage of samples where Document B is similar (closer) to Document A in the vector space than Document C. 
        - `similarity(A, B) > similarity(B, C)` or ` dist(A, B) < dist (A, C)`

    - I assumed that clusters (categories) should be distant apart in the vector space. i.e `distance(A,B)` should be minimal while `distance(A, C)` should be maximum. 
        - One metric could be `1.0 - sigmoid( dist(A, B) - dist(A, C))` (1.0 is best).

- I compared 5 Models
    - Word2Vec 
    - 2x Doc2vecC with 0.2 corruption value both via original C code and Gensim's fast version.
    - 2x Doc2vecC with 0.6 corruption value both via original C code and Gensim's fast version.

#### Results:
| Model                                  | # Correct classifications | % correct classifications | Average Distance |
|----------------------------------------|---------------------------|---------------------------|------------------|
| Word2Vec Model                         | 52,490                    | 70.439%                   | 0.527471         |
| Doc2vecC gensim - 0.2 corruption value | 48,222                    | 64.712%                   | 0.522506         |
| Doc2vecC C      - 0.2 corruption value | 49,147                    | 65.953%                   | **0.627540**         |
| Doc2vecC gensim - 0.6 corruption value | **57,402**                    | **77.031%**                   | 0.555006         |
| Doc2vecC C      - 0.6 corruption value | 43,088                    | 57.822%                   | 0.561585         |

#### Remarks:
- Gensim's fast version has better classification accuracy (by far) than the C code with the same parameters with 0.6 corruption value. 
- Doc2vecC code with 0.2 corruption value has +20% increase in average distance, which means clusters are further distant apart in the vector space.

#### Take Home:
- The results doesn't make sense for me, the difference between the C code and the one I wrote is not consistent. I may have done a mistake here or there, but I believe the notebook is reproducible if someone wants to go over it, extend it, please feel free.

- The dataset and the experiment set up may not be the most ideal, if someone wants to contribute with a better notebook, please do so.

- Regarding https://github.com/RaRe-Technologies/gensim/issues/1159#issuecomment-442253228 I totally agree that doc2vecc will add a maintenance overhead in the future and maybe not worth the trouble of adding it to the `Word2Vec` classes. I'm not emotionally attached to the PR (yet) if the decision is to not merge it 😄

docs/notebooks/doc2vecc_newsgroup20.ipynb
Saftophobia: I reran the experiment and updated the ipynb. I summarized it here https://github.com/RaRe-Technologies/gensim/pull/1985#issuecomment-442482214 for visibility.

docs/notebooks/doc2vecc_newsgroup20.ipynb
gojomo: The KeyedVectors functions `most_similar()` and `doesnt_match()` both internally compute a mean-of-multiple-vectors, in slightly different ways. (In `most_similar()`, some of the `negative` contributors to the mean are negated before mixed-in. In both, numpy functions offer compactness & efficiency that might be better (due to increased use of native parallel array operations) than an explicit python accumulate-loop-then-divide.

docs/notebooks/doc2vecc_newsgroup20.ipynb
```diff
@@ -164,47 +349,6 @@
     "    print(\"%s classified:\\t\\t %i correctly,\\t %f%%\" % (k,true_positives, true_positives/len(df_positive_negative) * 100))\n",
     "    print(\"%s average distance: \\t %f\" % (k, avg_distance))\n"
    ]
```
GeneralSemantics: Thanks very much for your hard work on this.

Just to make sure I'm following, here you're comparing the results from Chen's C implementation of Doc2VecC with your Gensim implementation, the best results for which were observed with corruption rate = 0.2.  And "classified correctly" means articles from the same news category were closer to one another than to an article from a random different category.  Have I got this right?

If so, is 77.012265% better or worse than the word2vec averaging method you were using as a baseline?

Also, while this method of comparison seems reasonable, it might be worth trying the method Chen uses in the paper to evaluate Doc2VecC for document classification -- training a linear SVM on the Doc2VecC representations of these news articles and then comparing its performance to one trained on the averaged word2vec vectors (or whatever baseline you prefer)

gojomo: It look like this latest notebook *doesn't* compare with Chen's C-patch results – which would still be useful as a sanity check, because i remember seeing in a prior verision of the notebook a comparison where there was a large discrepancy between vectors that should have performed similarly. 

It would be useful for comparisons to also include timings – as I believe the summed-docvec may add a noticeable overhead – and to compare against plain Word2Vec with a larger window (as the effect of using "20% of words from anywhere in doc" instead of just "small context window" is a little bit like "larger context window").

Saftophobia: It's a bit weird but `Chen's C-patch results` are the in last commit 7 hours ago. I'll list down the timings in a sec.

Saftophobia:
### Results:
- I didn't record the time for the Word2Vec model, it was somewhere around 5m, I can rerun it tomorrow.

|                       | % Correctly Classified | Avg Distance         | Time       |
|-----------------------|------------------------|----------------------|------------|
| Gensim - Word2Vec     | 70.450093%             | 0.527342             | -          |
| Gensim - Doc2vecC 0.2 | **77.012265%**         | 0.553244             | ~34m       |
| Gensim - Doc2vecC 0.4 | 77.005556%             | 0.554434             | ~62m       |
| Gensim - Doc2vecC 0.6 | 76.966639%             | 0.554916             | ~87m       |
| Gensim - Doc2vecC 0.8 | 76.680802%             | 0.554535             | ~110m      |
| Gensim - Doc2vecC 1.0 | 76.439250%             | 0.554650             | ~132m      |
| --------------------  | --------------------   | -------------------- |            |
| C - Doc2vecC 0.2      | 66.630881%             | **0.632914**         | 8m24.315s  |
| C - Doc2vecC 0.4      | 59.557422%             | 0.572672             | 9m58.016s  |
| C - Doc2vecC 0.6      | 57.796774%             | 0.561308             | 10m31.056s |
| C - Doc2vecC 0.8      | 57.175448%             | 0.557564             | 10m30.842s |
| C - Doc2vecC 1.0      | 56.182399%             | 0.557967             | 10m15.160s |

### Remarks:
- Both Gensim and Chen's C code decays when doc2vecC value increases. 
- However, Chen's clusters are more distant in the vector space, and decays unlike gensim's (I don't know why).
- I assume the +16% accuracy boost is due to non dirty writes, but at the expense of the time taken. 

The notebook is not close to ready for documentation purposes, but it's reproducable (I added the params I used for Chen's C models). I'm open for suggestions (experiment setup, refactoring etc).

docs/notebooks/doc2vecc_newsgroup20.ipynb
Saftophobia: @GeneralSemantics Sorry for late. Yes, that's what correcly classified means. I wanted a quick experiment with a clean dataset that is easy to import. 

SVM on top could be something to try out, I'll reread the paper next week and see if it's manageable, thank you.

gojomo:
There's a bit of a mystery in these results. I believe the Chen & gensim implementation should have similar performance on similar inputs/metaparameters. (In particular, I believe both do clobbering-writes from competing threads.) That vanilla gensim Word2Vec does better than any of Chen's doc2vecc numbers is also odd. 

Perhaps since Chen's code is based on the original word2vec.c, also trial word-vectors made by plain word2vec, to see if there's a discrepancy there, too? 

Paralleling the evaluation done in the paper would be ideal. Adding comparisons against word2vec-with-larger-windows, or other metaparameter optimizations, would better test if the time-cost here is the only way to get the benefit. 

My main concern is still that mentioned in my [issue comment](https://github.com/RaRe-Technologies/gensim/issues/1159#issuecomment-442253228) – extra complexity (and maybe a performance impact), to the main word2vec code paths, for a narrow/murky benefit. (I'd have other comments on smaller issues of naming/code-style if this bigger concern were addressed.)

Saftophobia:
@gojomo @GeneralSemantics Thanks for the feedback, I'll address the comparisons before the code refactoring.

gojomo:
@Saftophobia Thanks - & it's preferable to put any new results in a new comment... retaining the history of incremental progress & generating notifications for new material are both good things!

Saftophobia:
I used the author's [evaluation script](https://github.com/mchen24/iclr2017/blob/master/go.sh) to train gensim's on the same clean text with `cbow` and `sg`. ([gist](https://gist.github.com/Saftophobia/c42f896e0d4e95e32d00e21d5fec1342))

|   | gensim | C code|
|-----| -------|-------|
| Doc2vecC CBOW| 86.732%| 89.344%|
|Word2vec CBOW| 86.012% | - |
| Doc2VecC SG | 88.548% | 84.3% |

Went line by line in both implementations. I'm out of ideas why my implementation performs much worse in CBOW and better in sg. 

I'll close the PR till I find a reasonable justification. Thank you for your patience.

PR #1987, opened 2018-03-20
Fix `_is_single` from `Phrases` for case when corpus is numpy array of tokens

rmalouf:
Transforming a corpus using `Phraser[]` fails with:
````
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
````
if the corpus is a numpy array, a pandas Series, or anything else that defaults to element-wise comparison.  

The problem is this check here:
````
if temp_iter == obj:
````
A simple fix is replacing `==` with `is`

### CODE REVIEW

File: gensim/models/phrases.py [hunk 1]
```diff
@@ -118,7 +118,7 @@ def _is_single(obj):
     if isinstance(peek, string_types):
         # It's a document, return the iterator
         return True, obj_iter
-    if temp_iter == obj:
+    if temp_iter is obj:
```
menshikh-iv: Need to add a clear test for the case when this doesn't work correctly without this fix (and now it works).
Also, return mode of file please (755 -> 644).

rmalouf: Fixed.

menshikh-iv:
Good catch @rmalouf, congratz with the first contribution :1st_place_medal:

PR #1992, opened 2018-03-21
Fix return dtype for `matutils.unitvec` according to input dtype. Fix #1722

### LINKED ISSUE #1722: matutils.unitvec returns vector with different dtype from input vector

o-P-o:
This returns the float32 dtypes for float32 inputs for both normal and sparse arrays. The output unit vectors are also the same as those outputted from doing manual unit vector calculations.

How does this look? Any further suggestions for improvement or why this isn't acceptable?


gensim/matutils.py
```diff
@@ -667,45 +667,51 @@ def ret_log_normalize_vec(vec, axis=1):
...
 def unitvec(vec, norm='l2'):
     """Scale a vector to unit length.
-
     Parameters
     ----------
     vec : {numpy.ndarray, scipy.sparse, list of (int, float)}
         Input vector in any format
     norm : {'l1', 'l2'}, optional
         Normalization that will be used.
-
```
menshikh-iv: Please return empty lines

gensim/matutils.py
```diff
@@ -667,45 +667,51 @@ def ret_log_normalize_vec(vec, axis=1):
...
+    
     if scipy.sparse.issparse(vec):
         vec = vec.tocsr()
         if norm == 'l1':
             veclen = np.sum(np.abs(vec.data))
         if norm == 'l2':
             veclen = np.sqrt(np.sum(vec.data ** 2))
         if veclen > 0.0:
-            return vec / veclen
+            if np.issubdtype(vec.dtype, np.int) == True:
```
menshikh-iv: No need to `== True`

gensim/matutils.py
```diff
@@ -667,45 +667,51 @@ def ret_log_normalize_vec(vec, axis=1):
...
+                return vec / veclen
+            else:
+                vec /= veclen
+                return vec.astype(vec.dtype)
         else:
             return vec
 
     if isinstance(vec, np.ndarray):
-        vec = np.asarray(vec, dtype=float)
+        vec = np.asarray(vec, dtype=vec.dtype)
```
menshikh-iv: maybe this have no sense (because later you'll cast it again)

gensim/test_unitvec.py
```diff
@@ -0,0 +1,27 @@
+import numpy as np
+from scipy import sparse
+import unittest
+import matutils
+
+class UnitvecTestCase(unittest.TestCase):
+
+	def manual_unitvec(self, vec):
+		self.vec = vec
```
menshikh-iv: please use 4 spaces for indentation

gensim/test_unitvec.py
```diff
@@ -0,0 +1,27 @@
+import numpy as np
```
menshikh-iv: https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_matutils.py is more suitable place for this test

gensim/test_unitvec.py
```diff
@@ -0,0 +1,27 @@
...
+		if sparse.issparse(self.vec):
+			vec_sum_of_squares = self.vec.multiply(self.vec)
+			unit = 1. / np.sqrt(vec_sum_of_squares.sum())
+			return self.vec.multiply(unit)
+		elif not sparse.issparse(self.vec):
+			sum_vec_squared = np.sum(self.vec ** 2)
+			self.vec /= np.sqrt(sum_vec_squared)
+			return self.vec
+
+	def test_unitvec(self):
```
menshikh-iv: what's about different vectors (sparse) + different types (more floats + int too)?

menshikh-iv:
@o-P-o also, please resolve merge-conflict

menshikh-iv:
CC: @jayantj please have a look

gensim/matutils.py
o-P-o: Fixed

gensim/test_unitvec.py
o-P-o: Created pull request there

gensim/matutils.py
```diff
@@ -667,45 +667,54 @@ def ret_log_normalize_vec(vec, axis=1):
 
 def unitvec(vec, norm='l2'):
     """Scale a vector to unit length.
-
+
```
menshikh-iv: please don't make unrelated changes, this empty line(s) is correct by docstring convention.

gensim/matutils.py
```diff
@@ -667,45 +667,54 @@ def ret_log_normalize_vec(vec, axis=1):
...
+                vec = vec.astype(np.float)
+                return vec / veclen
+            else:
+                vec /= veclen
+                return vec.astype(vec.dtype)
         else:
             return vec
 
     if isinstance(vec, np.ndarray):
-        vec = np.asarray(vec, dtype=float)
```
menshikh-iv: I think that this line is needed especially for `blas_*` calls, please return it

gensim/test/test_matutils.py
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env python
```
menshikh-iv: looks strange, can you merge fresh `develop` to current branch please?

gensim/test/test_matutils.py
```diff
@@ -0,0 +1,187 @@
...
+
+                # 2 dimensional case
+                input_2d = rs.uniform(.01, 10000, size=(1, self.num_topics,))
+                known_good = dirichlet_expectation(input_2d)
+                test_values = matutils.dirichlet_expectation(input_2d)
+
+                msg = "dirichlet_expectation_2d failed for dtype={}".format(dtype)
+                self.assertTrue(np.allclose(known_good, test_values), msg)
+		
+class UnitvecTestCase(unittest.TestCase):
```
menshikh-iv: what's a name?

gensim/test/test_matutils.py
```diff
@@ -0,0 +1,187 @@
...
+	    vec_sum_of_squares = self.vec.multiply(self.vec)
+            unit = 1. / np.sqrt(vec_sum_of_squares.sum())
+	    return self.vec.multiply(unit)
+        elif not sparse.issparse(self.vec):
+	    sum_vec_squared = np.sum(self.vec ** 2)
+            self.vec /= np.sqrt(sum_vec_squared)
+	    return self.vec
+
+    def test_inputs(self):
+	input_dtypes = [np.float32, np.float64, np.int32, np.int64]
```
menshikh-iv: please add `float` and `int` (not numpy, python default)

gensim/test/test_matutils.py
```diff
@@ -0,0 +1,187 @@
...
+	input_arrtypes = ['sparse', 'normal']
```
menshikh-iv: `dense` better name (instead of `normal`)

gensim/test/test_matutils.py
```diff
@@ -0,0 +1,187 @@
...
+			input_vector = np.random.uniform(size=(5,)).astype(dtype_)
+			unit_vector = unitvec_with_bug.unitvec(input_vector)
+			man_unit_vector = self.manual_unitvec(input_vector)
+			self.assertEqual(input_vector.dtype, unit_vector.dtype)
+			self.assertTrue(np.allclose(unit_vector, man_unit_vector))
+    		    else:
+		        input_vector = np.random.randint(10, size=5).astype(dtype_)
+			unit_vector = unitvec_with_bug.unitvec(input_vector)
+			man_unit_vector = self.manual_unitvec(input_vector)
+			self.assertTrue(np.allclose(unit_vector, man_unit_vector))
```
menshikh-iv: you should check the dtype for this case too (you know that this will be default)

menshikh-iv:
@o-P-o please resolve merge-conflict first before you start to change a code according to my review.

gensim/test/test_matutils.py
```diff
@@ -141,6 +142,43 @@ def testDirichletExpectation(self):
...
+        if sparse.issparse(self.vec):
+            vec_sum_of_squares = self.vec.multiply(self.vec)
+            unit = 1. / np.sqrt(vec_sum_of_squares.sum())
+            return self.vec.multiply(unit)
+        elif not sparse.issparse(self.vec):
+            sum_vec_squared = np.sum(self.vec ** 2)
+            self.vec /= np.sqrt(sum_vec_squared)
+            return self.vec
+
+    def test_inputs(self):
```
jayantj: IMO we should split this test into multiple tests (one per combination of `(arr_type, dtype)` maybe) so the logic is simpler and so that it's possible to look at a failing test log and see exactly what kind of input caused the test to fail.

gensim/test/test_matutils.py
```diff
@@ -141,6 +142,43 @@ def testDirichletExpectation(self):
                 self.assertTrue(np.allclose(known_good, test_values), msg)
 
 
+class UnitvecTestCase(unittest.TestCase):
+    # test unitvec
+    def manual_unitvec(self, vec):
```
jayantj: Definitely should be simplified - why use `self` at all?
IMO should just be - 
```python
vec = vec.astype(np.float)
if sparse.issparse(vec):
    vec_sum_of_squares = vec.multiply(vec)
    unit = 1. / np.sqrt(vec_sum_of_squares.sum())
    return vec.multiply(unit)
else:
    sum_vec_squared = np.sum(vec ** 2)
    vec /= np.sqrt(sum_vec_squared)
    return vec
```

gensim/matutils.py
```diff
@@ -692,34 +692,49 @@ def unitvec(vec, norm='l2', return_norm=False):
...
                 return vec
 
     if isinstance(vec, np.ndarray):
-        vec = np.asarray(vec, dtype=float)
         if norm == 'l1':
             veclen = np.sum(np.abs(vec))
         if norm == 'l2':
             veclen = blas_nrm2(vec)
         if veclen > 0.0:
-            if return_norm:
```
jayantj: This entire construct (and a similar construct above) seems to have some unnecessary redundant code. We could simplify to something like - 

```python
if veclen > 0.0:
    if np.issubdtype(vec.dtype, np.int):
        vec = vec.astype(np.float)
    if return_norm:
        return blas_scal(1.0 / veclen, vec).astype(vec.dtype), veclen
    else:
        return blas_scal(1.0 / veclen, vec).astype(vec.dtype)
```

gensim/matutils.py
jayantj: I agree - this pretty much seems like a no-op, effectively. Does `vec = np.asarray(vec, dtype=vec.dtype)` have any effect on `vec` at all?

gensim/matutils.py
```diff
@@ -692,34 +692,42 @@ def unitvec(vec, norm='l2', return_norm=False):
...
         if norm == 'l1':
             veclen = np.sum(np.abs(vec.data))
         if norm == 'l2':
             veclen = np.sqrt(np.sum(vec.data ** 2))
         if veclen > 0.0:
+            if np.issubdtype(vec.dtype, np.int):
+                vec = vec.astype(np.float) / veclen
+            else:
+                vec /= veclen
+                vec = vec.astype(vec.dtype)
```
jayantj: Slightly confusing. For consistency, I'd prefer -
```python
if np.issubdtype(vec.dtype, np.int):
    vec = vec.astype(np.float) / veclen
else:
    vec = vec.astype(vec.dtype) / veclen
```
Does that make sense?
In fact, if dividing by `veclen` cannot change the dtype (is this the case?), even something like -
```python
if np.issubdtype(vec.dtype, np.int):
    vec = vec.astype(np.float)
vec /= veclen
```

jayantj: Also, sorry I'm really nitpicking here :)
But small things like this cause a slow decline in overall code quality over time.

gensim/test/test_matutils.py
```diff
@@ -141,6 +142,106 @@ def testDirichletExpectation(self):
...
+        vec_sum_of_squares = vec.multiply(vec)
+        unit = 1. / np.sqrt(vec_sum_of_squares.sum())
+        return vec.multiply(unit)
+    elif not sparse.issparse(vec):
+        sum_vec_squared = np.sum(vec ** 2)
+        vec /= np.sqrt(sum_vec_squared)
+        return vec
+
+
+class UnitvecTestCase(unittest.TestCase):
```
jayantj: Thanks for reorganizing the tests! Looks much better now IMO

gensim/matutils.py
o-P-o: Unfortunately dividing by `veclen` actually changes the dtype, causing the original problem to manifest itself (i.e. float32 inputs outputting float64). Even trying something like this:

```
if np.issubdtype(vec.dtype, np.int):
     vec = vec.astype(np.float) / veclen
else:
     vec = vec.astype(vec.dtype) / veclen.astype(vec.dtype)
```

causes the same problem. This is why I divide by `veclen` and then cast the dtype. Do you have any suggestions to get around this?

o-P-o: Actually, I really like your second suggestion, and it passes all tests:

```
if np.issubdtype(vec.dtype, np.int):
    vec = vec.astype(np.float) 
vec /= veclen
```

That's very neat indeed, I'll commit it.

o-P-o: Also, don't worry about the nitpicking! When I am experienced enough to pick up on such things I'm sure I will be nitpicking like this haha.

jayantj: Thanks for the fix and the explanation! Looks good.

menshikh-iv:
Congratulation @o-P-o with first contribution :1st_place_medal:

o-P-o:
Thanks for assisting my first ever merge guys!

Issue #1991, opened 2018-03-21
Word2Bits benchmark


menshikh-iv:
#### Description
Pretty interesting paper [Word2Bits - Quantized Word Vectors by Maximilian Lam](https://arxiv.org/abs/1803.05651), looks like it possible to apply "quantization" to the current w2v algorithm and receive a memory-compact representation without sacrificing quality.

#### ToDo
1. Make needed changes in current w2v code (according to this article), only for testing
2. Compare this approach by embedding quality (+memory consumption) with current w2v implementation (reproduce evaluation from paper)
   - **Train corpus:** English wikipedia
   - **Benchmark:**
      - `accuracy` method (classical approach)
      - SQuAD task (more detailed described in the paper)

If benchmark shows good-enough results, this will be a part of Gensim.

menshikh-iv:
Looks like very good task for you @persiyanov :)

menshikh-iv:
Fixed by #2011 (benchmark)

PR #2011, opened 2018-03-31
[WIP #1991]: Word2Bits benchmarks

persiyanov:
Pull request with intermediate results on [this issue](https://github.com/RaRe-Technologies/gensim/issues/1991).

I've implemented quantization techniques as stated in [the paper](https://arxiv.org/pdf/1803.05651.pdf).

I've just benchmarked it on `text8` corpus with 10 iterations over the dataset.

Other params (almost as in paper except embedding size):
* size = 128
* 10 iterations
* Window = 10
* Negative = 12
* Min count = 5
* Subsampling = 1e-4
* Learning rate = 0.05 (linearly decayed to 0.0001)

Based on my initial benchmark results are not promising:

| Training method                  | Quantization after training | Semantic accuracy | Syntactic accuracy |
|----------------------------------|-----------------------------|-------------------|--------------------|
| Original w2v (num_bits=0)        | No quantization             | 49.96%            | 46.99%             |
|                                  | 1 bit                       | 13.06%            | 15.16%             |
|                                  | 2 bits                      | 13.31%            | 15.04%             |
| 1 bit quantization (num_bits=1)  | No quantization             | 29.51%            | 29.85%             |
|                                  | 1 bit                       | 19.18%            | 13.28%             |
| 2 bits quantization (num_bits=2) | No quantization             | 44.38%            | 36.46%             |
|                                  | 2 bits                      | 18.79%            | 21.52%             |

* Quantized vectors work better if training phrase was done with quantization techniques.
* But there is a large accuracy drop when training is done with quantization. I suppose that embedding size (128) is too small to encode all information when quantization is applied.

So, currently I'm training the same models but for **25 iterations** and with **size=400**. I'll post the results asap.

After that, I will train the same setup on EN wikipedia (as was done in the paper).

@menshikh-iv please, take a look at my code. Maybe there are some bugs and this leads to low accuracy.


measure_quantized_w2v.py
```diff
@@ -0,0 +1,88 @@
+from __future__ import unicode_literals
```
menshikh-iv: notebook is enough here

gensim/models/base_any2vec.py
```diff
@@ -298,7 +298,7 @@ def _set_train_params(self, **kwargs):
         raise NotImplementedError()
 
     def __init__(self, sentences=None, workers=3, vector_size=100, epochs=5, callbacks=(), batch_words=10000,
-                 trim_rule=None, sg=0, alpha=0.025, window=5, seed=1, hs=0, negative=5, cbow_mean=1,
+                 trim_rule=None, sg=0, alpha=0.025, window=5, seed=1, hs=0, negative=5, num_bits=0, cbow_mean=1,
```
menshikh-iv: this parameter only for w2v, isn't?  Maybe better to add it directly to w2v?

gensim/models/base_any2vec.py
```diff
@@ -309,6 +309,7 @@ def __init__(self, sentences=None, workers=3, vector_size=100, epochs=5, callbac
         self.min_alpha = float(min_alpha)
         self.hs = int(hs)
         self.negative = int(negative)
+        self.num_bits = int(num_bits)
```
menshikh-iv: need to be sure that old model will be loaded successfully (need to modify `load` function) + backward compatibility test (this modification needed only for w2v class I think).

gensim/models/keyedvectors.py
```diff
@@ -121,6 +121,18 @@ def save(self, fname_or_handle, **kwargs):
     def load(cls, fname_or_handle, **kwargs):
         return super(BaseKeyedVectors, cls).load(fname_or_handle, **kwargs)
 
+    def quantize_vectors(self, num_bits=0):
```
menshikh-iv: missing docstring

gensim/models/word2vec.py
```diff
@@ -424,7 +424,7 @@ class Word2Vec(BaseWordEmbeddingsModel):
 
     def __init__(self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
                  max_vocab_size=None, sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
-                 sg=0, hs=0, negative=5, cbow_mean=1, hashfxn=hash, iter=5, null_word=0,
+                 sg=0, hs=0, negative=5, num_bits=0, cbow_mean=1, hashfxn=hash, iter=5, null_word=0,
```
menshikh-iv: this is a good idea to have `assert` here (because this works only with CBOW and negative sampling) + num_bits accept 1 and 2 (not more)

gensim/models/keyedvectors.py
menshikh-iv: Also, how you reduce memory here (`self.vectors` is float32 anyway)?

gensim/models/word2vec.py
```diff
@@ -529,20 +529,20 @@ def __init__(self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
...
+            hs=hs, negative=negative, num_bits=num_bits, cbow_mean=cbow_mean, min_alpha=min_alpha, compute_loss=compute_loss,
             fast_version=FAST_VERSION)
 
     def _do_train_job(self, sentences, alpha, inits):
         """
         Train a single batch of sentences. Return 2-tuple `(effective word count after
         ignoring unknown words and sentence length trimming, total word count)`.
         """
-        work, neu1 = inits
+        work1, work2, neu1 = inits
```
menshikh-iv: is it possible without `work2` here?

gensim/models/word2vec.py
```diff
@@ -529,20 +529,20 @@ def __init__(self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
...
         tally = 0
         if self.sg:
-            tally += train_batch_sg(self, sentences, alpha, work, self.compute_loss)
+            tally += train_batch_sg(self, sentences, alpha, work1, self.compute_loss)
         else:
-            tally += train_batch_cbow(self, sentences, alpha, work, neu1, self.compute_loss)
+            tally += train_batch_cbow(self, sentences, alpha, work1, work2, neu1, self.compute_loss, self.num_bits)
```
menshikh-iv: I think it's better to extract `num_bits` directly in cython from `self` (for save more consistent interfaces), like `cdef int negative = model.negative`

persiyanov:
Guys, don't lose me, I'm here. Models are training on English Wikipedia right now, it takes quite a long time. I suppose that results will be ready in a day or few.

piskvorky:
@persiyanov is it that much slower than "normal" word2vec? How many words per second are you seeing (on what hardware?

persiyanov:
@piskvorky For "normal" word2vec it was around 140-150k words per sec.

Now the model with 1bit quantization is training, its speed is 85k words per sec.

piskvorky:
OK, so half the speed. That's not so bad.

persiyanov:
Here are the results on **english wikipedia**. Trained for **5 epochs, embedding size 400**:

| Training method                  | Quantization after training | Semantic accuracy | Syntactic accuracy |
|----------------------------------|-----------------------------|-------------------|--------------------|
| Original w2v (num_bits=0)        | No quantization             | 68.79%            | 38.97%             |
|                                  | 1 bit                       | 44.48%            | 25.22%             |
|                                  | 2 bits                      | 44.48%            | 25.22%             |
| 1 bit quantization (num_bits=1)  | No quantization             | 77.22%            | 61.36%             |
|                                  | 1 bit                       | 67.16%            | 51.97%             |
| 2 bits quantization (num_bits=2) | No quantization             | 65.12%            | 47.09%             |
|                                  | 2 bits                      | **72.54%**            | **54.99%**             |

menshikh-iv:
Great, awesome results @persiyanov :fire: :fire: :fire:, definitely nice addition to Gensim!

gojomo:
Some thoughts: 

* `text8` is so small for word-vector purposes I'd be reluctant to make any strong conclusions based on comparisons using it
* parameters `window=10, negative=12, alpha=0.05` are sufficiently different from more-common defaults that I wonder what motivated their choice in the Word2Bits paper. It'd be interesting to compare results with more-default parameters, or with (best non-quantized params) vs (best quantized params)
* it'd be useful for tables of results to include runtimes – and for different options that each improve results, at a cost in runtime, to be compared. (For example: if quantization doubles runtime but improves accuracy, do other ways of spending twice-as-much time – especially the naive approach of twice-as-many training epochs – do just as well in the same amount of time?)

piskvorky:
@persiyanov what's the status here? This is a pretty cool feature. CC @menshikh-iv

persiyanov:
Attaching email discussion with author of Word2Bits paper. He said that quantization during training is unnecessary, the same effect could be achieved via l2 regularization (it's much faster).


------------------------
Hi Dmitry and Menshikh,

This is Max, author of the Word2Bits paper. I've been following the Word2Bits benchmark pull request (https://github.com/RaRe-Technologies/gensim/pull/2011). 

Firstly, thanks so much for benchmarking all this -- really appreciate it!

I have some updates to Word2Bits. The main finding is that you can train with l2 regularization and then just threshold the resulting word vector values (using the same quantization function as before) and still get pretty similar results as training with quantization and then thresholding. So the main finding is that training with quantization is equivalent to regularized training, and regularized training really helps improve word vector quality.

Thought this piece of information might be helpful, as training with l2 regularization is much faster than training with the quantization function! Hope this helps Gensim's word2vec!

Thanks!
Max

-------------------

Greetings, Max! 

This is really great, because my current implementation runs with 80/50k words per second for 1 and 2 bits quantization respectively. To compare, classic word2vec runs with 150k words per second.

I'll definitely try this out. Can you share the value of l2 coefficient from your experiments?


Thank you for reaching out, looking forward for your answer

Dmitry 


--------------------

Max Lam <agnusmaximus@gmail.com> | Tue, Apr 17, 2018 at 3:10 AM
To: Dmitry Persiyanov <dmitry.persiyanov@gmail.com>Cc: menshikh.iv@gmail.com
Awesome! I used an l2 coefficient of .001 for my experiments (which seemed to work pretty well for full wikipedia and for text8). 


--------------------

ivan <menshikh.iv@gmail.com> | Tue, Apr 17, 2018 at 8:13 AM


To: Max Lam <agnusmaximus@gmail.com>, dmitry.persiyanov@gmail.com
Hi Max,thanks for the advice, this will be pretty useful for us! Have you GitHub account? |

---------------------


Hi Ivan,

Yeah I have a github -- my id is agnusmaximus.

Thanks!
Max

piskvorky:
Thanks @persiyanov . I guess that means we drop this PR, and focus on adding regularization instead? CC @menshikh-iv @gojomo

persiyanov:
No need to drop this PR. As I remember, I already added l2 regularization in this PR, so we can move within it. Remember, that only training should be done with l2 regularization, but after that we should apply quantization, as I understood.

Currently, I almost don't have any free time to work on it (moved to a new job + my university classes started), maybe will have in a month or two.

menshikh-iv:
Unfortunately, @persiyanov still have no time :worried: for this reason, I close this PR :(
@persiyanov feel free to re-open when you'll have an time (I really hope that this will happen).

Issue #1955, opened 2018-03-05
SoftCosineSimilarity: Unexpected Behavior for Query Input


nataliedurgin:
#### Description
Expected SoftCosineSimilarity self[query] syntax to accept any iterable of list of (int, number) as indicated by: https://github.com/RaRe-Technologies/gensim/blob/f9669bb8a0b5b4b45fa8ff58d951a11d3178116d/gensim/similarities/docsim.py#L925

#### Steps/Code/Corpus to Reproduce
```
import scipy as sp
import gensim

# import the common_corpus as a TextCorpus object
# We assume common_corpus is a text file containing the lines:
#
# Human machine interface for lab abc computer applications
# A survey of user opinion of computer system response time
# The EPS user interface management system
# System and human system engineering testing of EPS
# Relation of user perceived response time to error measurement
# The generation of random binary unordered trees
# The intersection graph of paths in trees
# Graph minors IV Widths of trees and well quasi ordering
# Graph minors A survey

corpus = gensim.corpora.TextCorpus(input="path/common_corpus")
tfidf = gensim.models.TfidfModel(
    corpus, id2word=corpus.dictionary, normalize=True)

# Behavior of MatrixSimilarity
cos_index = gensim.similarities.MatrixSimilarity(tfidf[corpus])
cos_index[tfidf[corpus]]  # :)

# Behavior of SoftCosineSimilarity
dim = len(corpus.dictionary)
M = sp.sparse.identity(dim).tocsc() # generate the identity matrix
softcos_index = gensim.similarities.SoftCosineSimilarity(tfidf[corpus], M)
softcos_index[tfidf[corpus]] # :(
softcos_index[tfidf[[bow for bow in corpus]]] # :|
softcos_index[[vec for vec in tfidf[corpus]]] # :|
```

#### Expected Results
I believe that the TransformedCorpus tfidf[corpus] is an iterable of list of (int, number):
```
In [248]: next(iter(tfidf[corpus]))
Out[248]:
[(0, 0.4500498962785324),
 (1, 0.4500498962785324),
 (2, 0.3080749612015952),
 (3, 0.3080749612015952),
 (4, 0.4500498962785324),
 (5, 0.4500498962785324)]
```
I would expect that to be taken as query input without incident. It seems that SoftCosineSimilarity was adapted from a copy of MatrixSimilarity and was meant to function in a similar spirit, however, they do not accept the same types of input. 

#### Actual Results
```
[...]

PR #2016, opened 2018-04-04
Implement Levenshtein term similarity matrix and fast SCM between corpora

Witiko:
This is a follow-up of #1827 (Implement Soft Cosine Measure). This PR is a work in progress and not yet ready to merge (although the tests should pass). Its intended purpose is to discuss and improve the current implementation.

## Introduction
The original implementation included only a single term similarity matrix based on word embeddings. The new commits add a Levenshtein term similarity matrix. To prevent code duplication and to reduce complexity, I have separated the matrix building algorithm from the code that retrieves most similar terms. In reaction to #1955, the Soft Cosine Measure (SCM) can now be computed not only between a pair of vectors, but also between a corpus, and a vector and between a pair of corpora. This last point is also the future work suggested in #1827. Issues to discuss are with regards to the placement of new code, deprecation of old code, and speeding up the Levenshtein distance implementation.

## TODO list
- ~~The `gensim.models.term_similarity` module~~
  - ~~Decide on the final location of the `gensim.models.term_similarity` module~~
    - ~~@menshikh-iv suggests moving to `gensim.similarity.term_similarity`, or the less redundant `gensim.similarity.terms`, or `gensim.similarity.termsim`.~~
  - ~~Profile the `SparseTermSimilarityMatrix` constructor.~~
  - ~~Cherry-pick non-contiguous dictionary support from #2047.~~
- The `gensim.models.levenshtein` module
  - ~~Decide on the final location of the `gensim.models.levenshtein` module~~
    - ~~@menshikh-iv suggests moving to `gensim.similarity.levenshtein`.~~
  - ~~Decide whether to deprecate or to keep and maintain the `similarity_matrix` [facade][facade].~~
  - ~~Profile the speed of the `LevenshteinSimilarityIndex.most_similar` method.~~
  - Improve the speed of the `LevenshteinSimilarityIndex.most_similar` method.
- ~~The `gensim.models.keyedvectors` module~~
  - ~~Decide whether to deprecate or to keep and maintain the `WordEmbeddingsKeyedVectors.similarity_matrix` [facade][].~~
  - ~~Profile the `WordEmbeddingSimilarityIndex.most_similar` method.~~
- ~~The `SparseTermSimilarityMatrix.inner_product` method~~
  - ~~Decide whether to keep or deprecate the `gensim.matutils.softcossim` function.~~
  - ~~Update the implementation of `gensim.similarities.docsim.SoftCosineSimilarity` using the `inner_product` method.~~
  - ~~Profile the `SparseTermSimilarityMatrix.inner_product` method.~~
      - ~~When computing a document × corpus inner product, use the document × document inner product dense matrix trick. Then, profile the `inner_product` method again.~~
- ~~Miscellaneous~~
  - ~~Update the [`soft_cosine_benchmark.ipynb`][soft-cosine-benchmark] for `inner_product` between between a document and a corpus, and for the non-parallelized Levenshtein similarity.~~
  - ~~Update the [`soft_cosine_tutorial.ipynb`][soft-cosine-tutorial] Jupyter notebook.~~
  - ~~Add examples to the documentation of `SparseTermSimilarityMatrix`.~~

 [soft-cosine-tutorial]: https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/soft_cosine_tutorial.ipynb (Finding similar documents with Word2Vec and Soft Cosine Measure)
 [soft-cosine-benchmark]: https://github.com/Witiko/gensim/blob/levenshtein-softcossim/docs/notebooks/soft_cosine_benchmark.ipynb (Benchmark: Implement Levenshtein term similarity matrix and fast SCM between corpora)
 [facade]: https://en.wikipedia.org/wiki/Facade_pattern (Facade pattern)

## The `gensim.similarities.termsim` module
A major structural change is the addition of the `gensim.similarities.termsim` module. Before this addition, there existed the `WordEmbeddingsKeyedVectors.similarity_matrix` method that contained both the matrix building algorithm and the algorithm for retriving most similar terms for a given term. Now the matrix building algorithm has been moved into a separate `SparseTermSimilarityMatrix` [director][builder] class and the algorithm for retrieving most similar terms has been moved into a separate `WordEmbeddingSimilarityIndex` [builder][] class that implements the `TermSimilarityIndex` interface. This change follows the [single responsibility principle][srp].

 [builder]: https://en.wikipedia.org/wiki/Builder_pattern (Builder pattern)
 [srp]: https://en.wikipedia.org/wiki/Single_responsibility_principle (Single responsibility principle)

Two new classes implementing the `TermSimilarityIndex` interface have also been added. The `LevenshteinSimilarityIndex` retrieves the most similar terms according to the “Levenshtein similarity” described by Charlet and Damnati, 2017 [[1][], sec.&nbsp;2.2]. The `UniformTermSimilarityIndex` assumes all distinct terms are equally similar and its main use is in testing `SparseTermSimilarityMatrix`.

The following UML class diagram captures the new structure:

soft-cosine-uml

The `WordEmbeddingsKeyedVectors.similarity_matrix` method and the `similarity_matrix` function in the `gensim.similarities.levenshtein` module currently serve as [facades][facade] that construct a `SparseTermSimilarityMatrix` using the appropriate `TermSimilarityIndex` behind the scenes. This keeps the code backwards-compatible. I marked the functions for deprecation in 4.0.0, but if the Gensim policies allow, we can get rid of them sooner. Besides backwards compatibility, the facades are also convenient for the users, but there is an associated maintenance cost if we decided to keep them.

## The `gensim.similarities.levenshtein` module
The `gensim.similarities.levenshtein` module contains code for computing the “Levenshtein similarity” described by Charlet and Damnati, 2017 [[1][]]. See [the benchmark][benchmark] for a detailed performance analysis.

 [benchmark]: https://github.com/Witiko/gensim/blob/levenshtein-softcossim/docs/notebooks/soft_cosine_benchmark.ipynb (Benchmark: Implement Levenshtein term similarity matrix and fast SCM between corpora)

## The `SparseTermSimilarityMatrix.inner_product` method
The `SparseTermSimilarityMatrix.inner_product` method contains code for computing the inner product between two vectors or corpora expressed in a non-orthogonal basis.

There is a bit of “smart” linear algebra involved in computing the inner product between two L2-normalized m×n corpus matrices **X** and **Y**, which I will briefly describe here. We need to normalize each column document vector **x** in **X** by sqrt(**x**ᵀ ⋅ **S** ⋅ **x**), which is equivalent to the [entrywise (Hadamard) division][hadamard] of each row in **X** by the diagonal of sqrt(**X**ᵀ ⋅ **S** ⋅ **X**), where **S** is the m×m term similarity matrix. However, sqrt(**X**ᵀ ⋅ **S** ⋅ **X**) is an O(mn² ≈ m⁵) operation. We can instead directly compute the column vector of the diagonal as sqrt(**X**ᵀ ⋅ **S** \* **X**ᵀ) summed along the row axis, where \* is the [entrywise product][hadamard], which is an O(m²n ≈ m⁴) operation.

 [hadamard]: https://en.wikipedia.org/wiki/Hadamard_product_(matrices) "Hadamard product (matrices)"

The `SparseTermSimilarityMatrix.inner_product` method method is a more general variant of the `gensim.matutils.softcossim` function, which only computes the inner product between two normalized vectors, not general vectors or corpora. I marked the `matutils.softcossim` function for deprecation in 4.0.0, but if the Gensim policies allow, we can get rid of it sooner.

Since the `gensim.similarities.termsim` model imports the `corpus2csc` function from the `matutils` module, importing `SparseTermSimilarityMatrix.inner_product` from `gensim.matutils` would result in a cross-dependency. Therefore, the `softcossim` function carries all the old code instead of calling `inner_product`.

~~If we prefered not to deprecate the `softcossim` function, an alternative solution would be to move the `SparseTermSimilarityMatrix.inner_product` method to a `gensim.matutils.softdot` function, and make `softcossim` a [facade][] for `softdot`.~~

## Future work
The `SparseTermSimilarityMatrix` class constructor uses the dictionary of keys (DOK) sparse matrix format to incrementally build a sparse matrix. This is convenient, but, [as explained by Maciej][kula15], space-inefficient. I observed a 100-fold increase in RAM usage compared to using three dynamic arrays (two for indices, and one for the data) with the shortest possible unsigned integer data types. Similar technique will be implemented to the `SparseTermSimilarityMatrix` class constructor.

 [kula15]: http://maciejkula.github.io/2015/02/22/incremental-sparse-matrices/

The `levdist` function admits a `max_distance` parameter that allows us to terminate the computation of the Levenshtein distance early. This optimization will be introduced to the [python-Levenshtein][] module from Antti Haapala (see also the [discussion below][max_distance discussion]).

The `LevenshteinSimilarityIndex.most_similar` method currently uses pointwise Levenshtein distance to retrieve the most similar terms to a given term in average time O(mn²), where m is the number of terms in a dictionary and n is the average word length. When used to compute a term similarity matrix, this results in average time O(mmn² = m²n²). A more time-efficient procedure for computing the Levenshtein distance between all terms in a dictionary will be implemented.

 [max_distance discussion]: https://github.com/RaRe-Technologies/gensim/pull/2016#discussion_r202326763
 [python-Levenshtein]: https://pypi.org/project/python-Levenshtein/

## References
1. Delphine Charlet and Geraldine Damnati, *SimBow at SemEval-2017 Task 3: Soft-Cosine Semantic Similarity between Questions for Community Question Answering*, 2017. ([link to PDF][1])
2. Vít Novotný. *Implementation Notes for the Soft Cosine Measure*, 2018. ([link to PDF][2])

 [1]: http://www.aclweb.org/anthology/S17-2051 (Simbow at semeval-2017 task 3: Soft-cosine semantic similarity between questions for community question answering)
 [2]: https://arxiv.org/pdf/1808.09407 (Implementation Notes for the Soft Cosine Measure)


gensim/test/test_similarities.py
```diff
@@ -371,6 +372,7 @@ def testIter(self):
 
 
 class TestSoftCosineSimilarity(unittest.TestCase, _TestSimilarityABC):
+    @deprecated("Method will be removed in 4.0.0")
```
menshikh-iv: I don't think that this is a good idea to deprecate something in tests. Anyway, if we'll remove old code - tests will be broken and we'll see it clearly.

gensim/test/test_levenshtein.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python
```
menshikh-iv: why not place it in `test_similarities` (same for `test_term_similarity`)?

gensim/models/term_similarity.py
```diff
@@ -0,0 +1,271 @@
...
+import numpy as np
+from scipy import sparse
+
+from gensim.matutils import corpus2csc
+from gensim.utils import SaveLoad, is_corpus
+
+logger = logging.getLogger(__name__)
+
+
+class TermSimilarityIndex(SaveLoad):
```
menshikh-iv: I don't think that `gensim.models` is the right place for it (same for `levenshtein`), we have `gensim.similarities` for this kind of stuff (exception only for `KeyedVectors`)

CC: @piskvorky

gensim/models/term_similarity.py
```diff
@@ -0,0 +1,271 @@
...
+    -----
+    This class is mainly intended for testing SparseTermSimilarityMatrix and other classes that
+    depend on the TermSimilarityIndex.
+
+    """
+    def __init__(self, dictionary, term_similarity=0.5):
+        self.dictionary = dictionary
+        self.term_similarity = term_similarity
+
+    def most_similar(self, t1, topn=10):
```
menshikh-iv: Looks like this isn't ready, am I right?

Witiko: This is quite ready, just not very useful. `UniformTermSimilarityIndex` assigns a constant similarity to any pair of distinct words; the main use is for testing `SparseTermSimilarityMatrix`.

setup.py
```diff
@@ -309,6 +309,7 @@ def finalize_options(self):
         'scipy >= 0.18.1',
         'six >= 1.5.0',
         'smart_open >= 1.2.1',
+        'python-Levenshtein >= 0.10.2'
```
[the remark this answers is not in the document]
Witiko: This dependency might be temporary, since the current Levenshtein distance implementation is inefficient. We discussed more efficient bulk Levenshtein distance implementation in #1955.

gensim/models/term_similarity.py
Witiko: The module placement is one of the things that I hoped we could discuss, since there is little documentation about where things should go. Both modules implemented term similarity models; if you think they would feel more at home in `gensim.similarities`, then to `gensim.similarities` they will go.

gensim/matutils.py
```diff
@@ -775,6 +776,9 @@ def cossim(vec1, vec2):
     return result
 
 
+@deprecated(
+    "Function will be removed in 4.0.0, use " +
```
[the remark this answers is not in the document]
Witiko: I will fix this once we figure out what to actually deprecate.

Witiko:
## Adding the `gensim.models.term_similarity` module
A major structural change is the addition of the `gensim.models.term_similarity` module, which @menshikh-iv suggests moving to `gensim.similarity.term_similarity`, or the less redundant `gensim.similarity.terms`.

Before this addition of the module, there existed the `WordEmbeddingsKeyedVectors.similarity_matrix` method that contained both the matrix building algorithm and the algorithm for retriving most similar terms for a given term. Now the matrix building algorithm has been moved into a separate `SparseTermSimilarityMatrix` director class and the algorithm for retrieving most similar terms has been moved into a separate `WordEmbeddingSimilarityIndex` builder class that implements the `TermSimilarityIndex` interface. This change follows the [single responsibility principle][s].

 [s]: https://en.wikipedia.org/wiki/Single_responsibility_principle (Single responsibility principle)

Two new classes implementing the `TermSimilarityIndex` have also been added. The `LevenshteinSimilarityIndex` retrieves the most similar terms according to the levenshtein similarity described by Charlet and Damnati, 2017 [[1][]], and the `UniformTermSimilarityIndex` assumes all distinct terms are equally similar and its main use is in testing `SparseTermSimilarityMatrix`.

The following UML class diagram captures the new structure:

soft-cosine-uml

The `WordEmbeddingsKeyedVectors.similarity_matrix` method and the `similarity_matrix` function in the `gensim.models.levenshtein` module currently serve as facades that construct a `SparseTermSimilarityMatrix` using the appropriate `TermSimilarityIndex` behind the scenes. This keeps the code backwards-compatible. I marked the functions for deprecation in 4.0.0, but if the Gensim policies allow, we can get rid of them sooner.

Beside backwards compatibility, the facades are also convenient for the users, but there is a maintenance cost associated with them if we decided to keep them instead of deprecating them.

## References
1. Delphine Charlet and Geraldine Damnati, *SimBow at SemEval-2017 Task 3: Soft-Cosine Semantic Similarity between Questions for Community Question Answering*, 2017. ([link to PDF][1])

 [1]: http://www.aclweb.org/anthology/S17-2051 (Simbow at semeval-2017 task 3: Soft-cosine semantic similarity between questions for community question answering)

gensim/models/term_similarity.py
menshikh-iv: Yes, I think better to place it in `gensim.similarities`, because
- This is purely about the similarity (and we already have submodule for it)
- This doesn't "quack" like a standard gensim model

menshikh-iv:
@Witiko please write TO-DO list for this PR (what else need to change) for simpler navigation / better understanding

Witiko:
I will later today.

gensim/test/test_similarities.py
Witiko: After we have decided what to actually deprecate, I will remove the annotations from tests. They are currently there just to make it clear what parts of the code are proposed for deprecation.

Witiko:
I have finished the description and the TODO list, and I inserted both into the first post for better legibility.

Witiko:
I have finished the description and the TODO list as requested, and I inserted both into the first post for improved legibility. Any comments, especially on the deprecation TODO list items, are welcome.

gensim/models/term_similarity.py
Witiko: It is also quite useful for benchmarking the maximum throughput of the `SparseTermSimilarityMatrix` constructor.

Witiko:
I assume at several places that `range(len(dictionary))` is equivalent to `sorted(dictionary.keys())`. Is this a valid assumption, or can we have “sparse” dictionaries, where not every id between 0 and `max(dictionary.keys())` is mapped to a term?

Witiko:
I added [a benchmark for the director and builder classes][benchmark]. With the default parameters and a dictionary of 2.01M terms:

- the [`SparseTermSimilarityMatrix` constructor][SparseTermSimilarityMatrix.__init__] consumes term similarities at **195.91 Kword pairs&nbsp;/&nbsp;s**,
- the [`WordEmbeddingSimilarityIndex.most_similar`][WordEmbeddingSimilarityIndex.most_similar] method produces term similarities at:
    - **368.69 Kword pairs&nbsp;/&nbsp;s** with [ANNOY][], and
    - **0.76 Kword pairs&nbsp;/&nbsp;s** without ANNOY, and
- the naïve implementation of the [`LevenshteinSimilarityIndex.most_similar`][LevenshteinSimilarityIndex.most_similar] method produces term similarities at **8.51 word pairs&nbsp;/&nbsp;s**.

 [SparseTermSimilarityMatrix.__init__]: https://github.com/Witiko/gensim/blob/levenshtein-softcossim/gensim/similarities/termsim.py#L156
 [WordEmbeddingSimilarityIndex.most_similar]: https://github.com/Witiko/gensim/blob/levenshtein-softcossim/gensim/models/keyedvectors.py#L1094
 [LevenshteinSimilarityIndex.most_similar]: https://github.com/Witiko/gensim/blob/levenshtein-softcossim/gensim/similarities/levenshtein.py#L112

The `SparseTermSimilarityMatrix` constructor currently uses the [dictionary-of-keys (DOK) format][DOK matrix] to incrementally build the term similarity matrix. DOK is [known for to be space-inefficient][incremental-sparse-matrices] (pointers taking up up to 10 times the size of the stored data). Using the [coordinate (COO) format][COO matrix] should improve both space- and time-efficiency, so that the `WordEmbeddingSimilarityIndex.most_similar` with ANNOY is not bottlenecked.

The current speed of `LevenshteinSimilarityIndex.most_similar` is usable for small corpora (<50K&nbsp;terms), where the construction of a term similarity matrix will take a week. However, real-world corpora such as the Ennglish Wikipedia would take years. I will appreciate help with implementing efficient retrieval of top N most similar terms for all terms in a dictionary as discussed in #1955, @piskvorky. Ping @nataliedurgin in case this is still relevant for you.

The above results are just a rough overview, please see [the benchmark][benchmark] for details. A benchmark of the fast SCM between corpora is currently underway, I expect it to finish later today.

[ANNOY]: https://github.com/spotify/annoy (Approximate Nearest Neighbors in C++/Python optimized for memory usage and loading/saving to disk)
[benchmark]: https://github.com/Witiko/gensim/blob/ee33db8862f431f0e8113cbfa7256880c8aca0e9/docs/notebooks/soft_cosine_benchmark.ipynb (Benchmark: Implement Levenshtein term similarity matrix and fast SCM between corpora)
[DOK matrix]: https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.sparse.dok_matrix.html (scipy.sparse.dok_matrix)
[COO matrix]: https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_matrix.html (scipy.sparse.coo_matrix)
[incremental-sparse-matrices]: http://maciejkula.github.io/2015/02/22/incremental-sparse-matrices/ (Incremental construction of sparse matrices)

Witiko:
I added [a benchmark][benchmark] for the [`SparseTermSimilarityMatrix.inner_product`][inner_product] method. These are the main results with the default parameters, a dictionary of 100K terms, and a corpus of 830 documents:

- `inner_product(vector1, vector2, normalized=True)` produces document similarities at **1.36 Kdoc&nbsp;pairs&nbsp;/&nbsp;s**,
- `inner_product(vector, corpus, normalized=True)` produces document similarities at **0.64 Kdoc&nbsp;pairs&nbsp;/&nbsp;s**, and
- `inner_product(corpus1, corpus2, normalized=True)` produces document similarities at **604.96 Kdoc&nbsp;pairs&nbsp;/&nbsp;s**.

Computing iterated inner product between two vectors is faster than computing the inner product between a vector and a corpus matrix because in the vector case, [we construct a smaller dense term similarity matrix that only contains rows and columns corresponding to the union of nonzero element vector indices][dense_trick]. No such trick is used in the inner product between a vector, so we pay for the overhead associated with sparse matrices and the speed is proportional to matrix sparsity.

The above results are just a quick overview, please see [the benchmark][benchmark] for details.

[benchmark]: https://github.com/Witiko/gensim/blob/ee33db8862f431f0e8113cbfa7256880c8aca0e9/docs/notebooks/soft_cosine_benchmark.ipynb (Benchmark: Implement Levenshtein term similarity matrix and fast SCM between corpora)
[inner_product]: https://github.com/Witiko/gensim/blob/levenshtein-softcossim/gensim/similarities/termsim.py#L220
[dense_trick]: https://github.com/Witiko/gensim/blob/levenshtein-softcossim/gensim/similarities/termsim.py#L262

Witiko:
I added [a benchmark][benchmark] for the [`SparseTermSimilarityMatrix.inner_product`][inner_product] method. These are the main results with the default parameters, a dictionary of 100K terms, and a corpus of 830 documents:

- `inner_product(vector1, vector2, normalized=True)` produces document similarities at **1.36 Kdoc&nbsp;pairs&nbsp;/&nbsp;s**,
- `inner_product(vector, corpus, normalized=True)` produces document similarities at **0.64 Kdoc&nbsp;pairs&nbsp;/&nbsp;s**, and
- `inner_product(corpus1, corpus2, normalized=True)` produces document similarities at **604.96 Kdoc&nbsp;pairs&nbsp;/&nbsp;s**.

Computing iterated inner product between two vectors is faster than computing the inner product between a vector and a corpus matrix because in the vector case, [we construct a smaller dense term similarity matrix that only contains rows and columns corresponding to the union of nonzero element vector indices][dense_trick]. No such trick is used in the inner product between a vector and a corpus matrix, so we pay the overhead associated with sparse matrix dot product and the speed is inversely proportional to matrix density. At small matrix density, the speed exceeds that of the inner product between two vectors.

The above results are just a quick overview, please see [the benchmark][benchmark] for details.

[benchmark]: https://github.com/Witiko/gensim/blob/aa7becd/docs/notebooks/soft_cosine_benchmark.ipynb (Benchmark: Implement Levenshtein term similarity matrix and fast SCM between corpora)
[inner_product]: https://github.com/Witiko/gensim/blob/levenshtein-softcossim/gensim/similarities/termsim.py#L220
[dense_trick]: https://github.com/Witiko/gensim/blob/levenshtein-softcossim/gensim/similarities/termsim.py#L262

Witiko:
The extended benchmarks are nearly finished and, disregarding the efficient Levenshtein distance implementation, the remaining points at the TODO list are minor and can be finished over a weekend. Since the naive Levenshtein distance implementation is fast enough for my research and there does not seem to be interest in doing this proper, I suggest we move forward with this PR without the efficient Levenshtein distance implementation. @piskvorky, would you prefer that I keep the naive implementation in, or that I revert it before the merge?

Witiko:
@menshikh-iv The code should be ready to merge.

gensim/similarities/levenshtein.py
```diff
@@ -0,0 +1,107 @@
...
+    LEVENSHTEIN_EXT = True
+except ImportError:
+    LEVENSHTEIN_EXT = False
+
+from gensim.similarities.termsim import TermSimilarityIndex
+
+logger = logging.getLogger(__name__)
+
+
+def levsim(t1, t2, alpha=1.8, beta=5.0):
```
piskvorky: The method signature (even if not implemented/used at the moment) should contain a "max_distance" threshold. With the semantics that values above this threshold are uninteresting to the upstream application, computation can terminate early.

Such "early out" optimization is super cheap and makes a huge difference in practical applications, in my experience.

Witiko: I added the `max_distance` parameter in commit ac60615. Currently, [the Levenshtein distance library I use](https://pypi.org/project/python-Levenshtein/) does not support early termination, but it should be a straightforward future addition ([`_levenshtein.c`](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c) at lines [2304](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L2304), [2332](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L2332), [2352](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L2352), [2448](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L2448), [2483](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L2483), [2511](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L2511), [2531](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L2531), [3223](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L3223), [3856](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L3856), [4815](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L4815), [4935](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L4935), [5810](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L5810), and [5999](https://github.com/ztane/python-Levenshtein/blob/master/Levenshtein/_levenshtein.c#L5999)).

gensim/similarities/levenshtein.py
```diff
@@ -39,6 +39,9 @@ def levsim(t1, t2, alpha=1.8, beta=5.0):
         The multiplicative factor alpha defined by Charlet and Damnati (2017).
     beta : float
         The exponential factor beta defined by Charlet and Damnati (2017).
+    max_distance : {int,float}
+        The maximum Levenshtein distance between t1, and t2. Larger distances
+        will result in the Levenshtein similarity of zero.
```
piskvorky: Why zero? That sounds like a bad idea -- zero distance means the opposite = "terms identical", not "terms too far apart to care".

A better value is `max_distance` or `inf`.

gensim/similarities/levenshtein.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Parameters
+    ----------
+    dictionary : :class:`~gensim.corpora.dictionary.Dictionary`
+        A dictionary that specifies the considered terms.
+    alpha : float, optional
+        The multiplicative factor alpha defined by Charlet and Damnati (2017).
+    beta : float, optional
+        The exponential factor beta defined by Charlet and Damnati (2017).
+    threshold : float, optional
```
piskvorky: Inconsistent naming with `max_distance` above.

gensim/similarities/levenshtein.py
piskvorky: Ah, I see you're converting distance to <0.0, 1.0> similarity.

In that case 0.0 is OK (same as dist=inf), but the change of units is confusing. Why not have the user input `min_similarity` instead of `max_distance`? If similarity is the conceptual unit we chose (as opposed to "distance"), we should stick to it.

It looks like we can freely convert between similarity and distance (bijection), so this is a more a design/conceptual question than a technical one.

Witiko: In commit 5154569, I ended up splitting the original function into the `levdist` function (with the `max_distance` parameter), which serves as a wrapper for the `distance` function from `python-Levenshtein`, and the `levsim` function (with the `min_similarity` parameter), which computes pointwise Levenshtein similarity in the sense of Damnati et al., 2017.

Instead of returning `float("inf")` when the distance exceeds `max_distance`, the current implementation of `levdist` returns `max(len(t1), len(t2))`, which is the upper bound on the distance between terms `t1`, and `t2`. This is a little more difficult to check for on the side of the application, but has the upside of removing the return type polymorphism (we always get an `int`) and it nicely plugs into the `levsim` function, where `max(len(t1), len(t2))` produces the similarity of zero.

gensim/similarities/levenshtein.py
Witiko: This should be resolved as per the above discussion.

gensim/similarities/levenshtein.py
```diff
@@ -0,0 +1,157 @@
...
+    number between <1.0, inf>, higher is less similar.
+
+    Parameters
+    ----------
+    t1 : {bytes, str, unicode}
+        The first compared term.
+    t2 : {bytes, str, unicode}
+        The second compared term.
+    max_distance : {int, float}
+        The maximum Levenshtein distance between `t1`, and `t2`. Greater
```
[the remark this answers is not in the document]
Witiko: I updated the wording as suggested in 155dc589ce6046e8dd8788cdff82357127693deb.

gensim/similarities/levenshtein.py
Witiko: I added the above notes to the future work section of the original post.

Witiko: Since there does not seem to be any rush to merge this, I started with the [modifications of the Levenshtein library](https://github.com/Witiko/python-Levenshtein/commit/0f9de6848d3bb5c9cde23ee05f8683e5563e7578). The interface, and the basic implementation is in place. Currently, the code only terminates prematurely, when `abs(len(t1), len(t2)) > max_distance`. There are likely other opportunities for early termination.

menshikh-iv:
@Witiko what's status here?

Witiko:
@menshikh-iv Waiting for review. Some of the future work items are being tackled (see [the discussion](https://github.com/RaRe-Technologies/gensim/pull/2016#discussion_r202326763)), but that does not block this pull request.

gensim/similarities/levenshtein.py
Witiko: In the light of this, I suggest we revert / forcibly roll back the last couple of commits (ac60615, 5154569, 729d185, 155dc58, and 7e52ef8) introducing `max_distance`, and `min_similarity`

Witiko:
@menshikh-iv @piskvorky Just a reminder that this PR:
- passes tests on Python 2.7.9, and 3.4.4 (CI services are false positives) and has no conflicts with the `develop` branch,
- improves the design of the SCM by decomposing it into a matrix builder, a number of term similarity indices, and a routine for computing the inner product of corpora / documents,
- improves the speed of the inner product of two corpora by a factor of thousands for large corpora,
- adds preliminary support for the Levenshtein distance,
- adds support for producing positive definite matrices that can later be factorized.

In other words, this PR should be good to go for the upcoming release. What are your thoughts?

menshikh-iv:
@Witiko please resolve an merge conflict

Witiko:
@menshikh-iv All checks should be passing now.

gensim/similarities/levenshtein.py
```diff
@@ -0,0 +1,158 @@
...
+        `max_distance`".
+        Default: always compute distance exactly, no threshold clipping.
+
+    Returns
+    -------
+    int
+        The Levenshtein distance between `t1` and `t2`.
+
+    """
+    if not LEVENSHTEIN_EXT:
```
menshikh-iv: make it simpler, just import here without try/catch (`*_EXT` have sense only for testing)

gensim/similarities/levenshtein.py
```diff
@@ -0,0 +1,158 @@
...
+        raise ImportError("Please install python-Levenshtein Python package to compute the Levenshtein distance.")
+
+    distance = Levenshtein.distance(t1, t2)
+    if distance > max_distance:
```
menshikh-iv: what can be bigger than `float('inf')`?

gensim/similarities/levenshtein.py
```diff
@@ -0,0 +1,158 @@
...
+    -------
+    float
+        The Levenshtein similarity between `t1` and `t2`.
+
+    Notes
+    -----
+    This notion of Levenshtein similarity was first defined in section 2.2 of
+    `Delphine Charlet and Geraldine Damnati, "SimBow at SemEval-2017 Task 3:
+    Soft-Cosine Semantic Similarity between Questions for Community Question
+    Answering", 2017 <http://www.aclweb.org/anthology/S/S17/S17-2051.pdf>`__.
```
menshikh-iv: `__.` -> `_.`

gensim/similarities/levenshtein.py
```diff
@@ -0,0 +1,158 @@
...
+
+    """
+    assert alpha >= 0
+    assert beta >= 0
+
+    max_lengths = max(len(t1), len(t2))
+    if max_lengths == 0.0:
```
menshikh-iv: `== 0`, len is integer-like always

gensim/similarities/levenshtein.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self, dictionary, alpha=1.8, beta=5.0, threshold=0.0):
+        self.dictionary = dictionary
+        self.alpha = alpha
+        self.beta = beta
+        self.threshold = threshold
+        super(LevenshteinSimilarityIndex, self).__init__()
+
+    def most_similar(self, t1, topn=10):
+        for _, (t2, similarity) in zip(range(topn), (
+                (t2, similarity) for similarity, t2 in sorted(
```
menshikh-iv: a bit hard to read, can you split this into several code-lines (for readability proposes)

docs/notebooks/soft_cosine_tutorial.ipynb
```diff
@@ -291,41 +313,1410 @@
...
    "outputs": [
     {
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "2018-02-05 10:52:53,477 : INFO : built Dictionary(462807 unique tokens: ['reclarify', 'depeneded', 'autralia', 'cloudnight', 'openmoko']...) from 2274338 documents (total 40096354 corpus positions)\n",
-      "2018-02-05 10:56:50,633 : INFO : training on a 200481770 raw words (192577574 effective words) took 224.3s, 858402 effective words/s\n",
-      "2018-02-05 11:13:14,895 : INFO : constructed a term similarity matrix with 0.003564 % nonzero elements\n"
+      "2018-09-11 22:05:07,212 : INFO : adding document #0 to Dictionary(0 unique tokens: [])\n",
+      "2018-09-11 22:05:07,485 : INFO : adding document #10000 to Dictionary(20088 unique tokens: ['authours', 'chimney', 'bombard', 'euro', 'lies']...)\n",
```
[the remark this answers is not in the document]
piskvorky: To be clear: remove it from the notebook, not from the code.

gensim/similarities/levenshtein.py
Witiko: This code path will only be taken if the caller provides their own `max_distance` that is less than `float('int')`. See lines 97 and 98.

gensim/similarities/levenshtein.py
Witiko: I extracted two generators from the loop.

Witiko:
We should be good to go.

gensim/similarities/levenshtein.py
```diff
@@ -0,0 +1,153 @@
...
+            (levsim(t1, t2, self.alpha, self.beta, self.threshold), t2)
+            for t2 in self.dictionary.values()
+            if t1 != t2
+        )
+        most_similar = (
+            (t2, similarity)
+            for (similarity, t2) in sorted(similarities, reverse=True)
+            if similarity > 0
+        )
+        for _, (t2, similarity) in zip(range(topn), most_similar):
```
piskvorky: What is the `range` for? It seems to be ignored.

If you want to clip `most_similar` without materializing it as a list (is that some optimization?), simply use `itertools.islice`.

gensim/models/keyedvectors.py
```diff
@@ -1386,6 +1331,49 @@ def init_sims(self, replace=False):
...
+        self.threshold = threshold
+        self.exponent = exponent
+        self.kwargs = kwargs or {}
+        super(WordEmbeddingSimilarityIndex, self).__init__()
+
+    def most_similar(self, t1, topn=10):
+        if t1 not in self.keyedvectors.vocab:
+            logger.debug('an out-of-dictionary term "%s"', t1)
+        else:
+            for _, (t2, similarity) in zip(
```
piskvorky: Please fix this one too (too complicated / hard to read).

gensim/similarities/levenshtein.py
Witiko: Not an optimization, since `levsim` currently dominates the time complexity and similarities are still materialized for sorting. I used `argsort` to make the code more concise as suggested.

gensim/models/keyedvectors.py
Witiko: Fixed in 3f04940

menshikh-iv:
Awesome work @Witiko, I hope your research going well :)

Witiko:
Below, you will find the suggested changelog.

### :star2: New Features

- Added similarity search using the Levenshtein distance in `gensim.similarities.LevenshteinSimilarityIndex`. (@Witiko, #2016)

### :+1: Improvements

- Performance optimizations to `gensim.similarities.SoftCosineSimilarity`. 🚀 (@Witiko, #2016)

    | dictionary size | corpus size | speed         |
    |-----------------|-------------|--------------:|
    | 1000            | 100         | 1.0×          |
    | 1000            | 1000        | **53.4×**     |
    | 1000            | 100000      | **156784.8×** |
    | 100000          | 100         | **3.8×**      |
    | 100000          | 1000        | **405.8×**    |
    | 100000          | 100000      | **66262.0×**  |

    - [Soft Cosine notebook with detailed benchmarks](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/soft_cosine_benchmark.ipynb)

### :warning: Deprecations (will be removed in the next major release)

- `gensim.matutils.softcossim`
- `gensim.models.keyedvectors.WordEmbeddingsKeyedVectors.similarity_matrix`

gensim/models/term_similarity.py
Witiko: Another use would be the construction of an identity term similarity matrix by setting `term_similarity` to zero.

PR #2027, opened 2018-04-11
Add `dtype` argument for `chunkize_serial`, fix `LdaModel`

darindf:
Reduce the size of data being sent to distributed workers, when as_numpy is true, is defaulting to float64

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -50,11 +50,6 @@ from gensim.models.callbacks import Callback
 
 logger = logging.getLogger('gensim.models.ldamodel')
 
-# DD
-USE_DASK = True
-logging.basicConfig(
-    format='%(asctime)s : %(levelname)s : %(message)s', level=logging.INFO)
-
 DTYPE_TO_EPS = {
     np.float16: 1e-5,
     np.float32: 1e-35,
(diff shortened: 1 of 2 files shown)


gensim/utils.py
```diff
@@ -1119,7 +1119,7 @@ def substitute_entity(match):
     return RE_HTML_ENTITY.sub(substitute_entity, text)
 
 
-def chunkize_serial(iterable, chunksize, as_numpy=False):
+def chunkize_serial(iterable, chunksize, as_numpy=False,dtype=np.float32):
```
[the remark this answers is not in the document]
darindf: with the new code, this is the dtype of the numpy arrays
```
wrapped_chunk[0][0].dtype
dtype('float32')
```

with the original code, using numpy dtype default, this is
```
wrapped_chunk[0][0].dtype
dtype('float64')
```

The iterator itself is returning documents that are list of tuples of type (int,float) (tokenId,tokenFreq) from corpora.MmCorpus

gensim/utils.py
```diff
@@ -1148,7 +1148,7 @@ def chunkize_serial(iterable, chunksize, as_numpy=False):
         if as_numpy:
             # convert each document to a 2d numpy array (~6x faster when transmitting
             # chunk data over the wire, in Pyro)
-            wrapped_chunk = [[np.array(doc) for doc in itertools.islice(it, int(chunksize))]]
+            wrapped_chunk = [[np.asarray(doc, dtype=dtype) for doc in itertools.islice(it, int(chunksize))]]
```
piskvorky: Doesn't this change the existing behaviour? I'm -1 on that, we should keep compatibility. Both in terms of the array `copy` and `dtype`.

gensim/models/ldamodel.py
```diff
@@ -698,7 +698,9 @@ def rho():
             dirty = False
 
             reallen = 0
-            for chunk_no, chunk in enumerate(utils.grouper(corpus, chunksize, as_numpy=chunks_as_numpy)):
+            for chunk_no, chunk in enumerate(utils.grouper(
+                                                       corpus, chunksize, as_numpy=chunks_as_numpy,
```
piskvorky: Bad indent. Also, line too long -- best split this long expression into two smaller expressions (first assign `utils.grouper`, then `enumerate`).

gensim/utils.py
darindf: This is being used as part of LdaState within LdaModel. These function/classes are already parameterized with dtype, so this makes this more consistent with that code base.

piskvorky: Sure, being parametrized is fine. It's a good addition.

But changing the existing semantics (potentially not copying arrays, changing float precision) is not fine.

darindf: Term frequency should always be an integer value, as you can't have half a word in a document. Not sure why mmcorpus is returning documents as (int,float) while bow returns (int,int)

In both cases this is creating a numpy array, for use in serializing in a distributed environment, which means a copy of the array is being received on the lda_worker process.

Reading asarray documentation,  `copy=False` is being ignored as there is datatype change being made between the source and destination array.

If it makes it clearer, I can change this to 
```
np.array(doc, dtype=dtype)
```

docs/src/corpora/indexedcorpus.rst
```diff
@@ -8,9 +8,3 @@
     :undoc-members:
     :show-inheritance:
 
-
-.. autoclass:: IndexedCorpus
```
menshikh-iv: Merge current `develop` to your PR please (this part will go away, because this is already merged)

gensim/models/ldamodel.py
menshikh-iv: still not changed

gensim/utils.py
menshikh-iv: @darindf yes, this will be better, I agree

docs/src/corpora/indexedcorpus.rst
darindf: I merged the changes, and commited back, but I can't get this message to go away.

menshikh-iv:
Thanks @darindf, good work :+1:

PR #2030, opened 2018-04-12
Fix format & links for `gensim.similarities.docsim`

CLearERR:
@menshikh-iv


menshikh-iv:
Thank you @CLearERR :+1:

PR #2033, opened 2018-04-19
Remove duplication of class documentation for `IndexedCorpus`

darindf:
Class documentation is repeated twice.


menshikh-iv:
Good catch, thanks @darindf :+1:

PR #2035, opened 2018-04-23
Dask

darindf:
Replace pyro distributed clusters with dask based clusters.

I have 2 implementations of distributing workload using a dask cluster.

The first approach currently has an outstanding bug/issue with dask distributed, in that the non-blocking queue (the queue get times out) will cause the dask scheduler to crash. We're currently trying to get this resolved. The advantage of this approach is that it mirrors the pyro implementation in that the dask workers keep running a lda model, and respond to queue events to either perform work on a cluster or to perform a lda state synchronization.

The second approach, is less efficient in that a lda model is created each and every time a chunk is operated on, but is more dask like, in that dask submit operations are performed, i.e. futures are created and waited until they complete.  Currently the number of dask workers is limited to prevent over-provisioning of a worker. This may be related to mkl threads. This approach has been tested against against 1.3 million documents, and found that the 3 dask workers (1 worker on 3 machines) outperformed a similar 3 pyro configuration. 

If there is interest in this, this can be cleaned up, and reviewed to ensure correctness.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -50,6 +50,11 @@ from gensim.models.callbacks import Callback
 
 logger = logging.getLogger('gensim.models.ldamodel')
 
+# DD
+USE_DASK = True
+logging.basicConfig(
+    format='%(asctime)s : %(levelname)s : %(message)s', level=logging.INFO)
+
 DTYPE_TO_EPS = {
     np.float16: 1e-5,
     np.float32: 1e-35,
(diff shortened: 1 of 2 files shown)


piskvorky:
Awesome! Dask is something I've been meaning to look into. It looks like a great Pythonic fit for Gensim.

menshikh-iv:
Hi @darindf,
this can be a pretty nice addition, be careful with
- Testing (this is probably the hardest part, but definitely needed)
- Tutorial (how to setup & run it from scratch)
- Performance benchmark (time/model quality), compare it with `LdaModel`, `LdaMulticore` and `Pyro4` distributed version

gensim/models/ldamodel.py
```diff
@@ -182,6 +187,366 @@ def load(cls, fname, *args, **kwargs):
         return result
 # endclass LdaState
 
+# DD
+# wrapper to just return lda.state after performing lda
+from distributed import local_client
+
+
+class DaskWorker(object):
```
menshikh-iv: it's better to place classes in distinct place (like with `Pyro4`)

gensim/models/ldamodel.py
```diff
@@ -182,6 +187,366 @@ def load(cls, fname, *args, **kwargs):
         return result
 # endclass LdaState
 
+# DD
+# wrapper to just return lda.state after performing lda
+from distributed import local_client
```
menshikh-iv: don't forget to update `setup.py`

gensim/models/ldamodel.py
```diff
@@ -182,6 +187,366 @@ def load(cls, fname, *args, **kwargs):
...
+class DaskWorker(object):
+    #from enum import Enum
+    #class MsgType(Enum):
+    RESET = 'RESET' 
+    EVAL = 'EVAL'
+    GETSTATE ='GETSTATE'
+    DONE = 'DONE'
+
+    
+    def __init__(self, state,inqueue,outqueue,myid, id2word, num_topics, chunksize, alpha, eta, distributed, random_state=None):
```
menshikh-iv: No default values? Are you sure that user **must** pass all of this?

darindf: The LdaModel defines the default values for these, and the dask worker should use the values that the LdaModel is created. Normally in this case I refactor the config information, then LdaModel would be split into 2, LdaModelConfig configuration, and LdaModelCompute. I see a separate class was created for LdaState, to hold the state of the lda algorithm.

gensim/models/ldamodel.py
darindf: Ok, I'll look what needs to be done there.

gensim/models/ldamodel.py
darindf: I agree. This was done to speed up development, as the code needs to copied to the dask workers file system, as the python file itself isn't pickled and sent.

darindf:
How does one want to distinguish between using pyro vs dask. Should this be defined by refining the distributed flag, now true/false into a enum Pyro, Dask, None? Or should this be configured via a config file?

gensim/models/ldamodel.py
menshikh-iv: `LdaState` used for different propose (this is "inner state", used only for computing).
I understand your point, but in this case, you need to make all parameters optional (or do it in a different way).

menshikh-iv:
modify distributed flag is a good idea I think (at least for the current variant, this can be changed at any moment).

mrocklin:
> The first approach currently has an outstanding bug/issue with dask distributed, in that the non-blocking queue (the queue get times out) will cause the dask scheduler to crash

Yup, thanks for identifying this and bringing it to our attention.  This is now resolved in https://github.com/dask/distributed/pull/1950.  We're releasing later today actually, so it should be easy to progress past this.

That being said using inter-worker coordination primitives are somewhat rare among new users; they tend to be a bit of an advanced technique and often overkill.  Most people using inter-worker coordination primitives like queues and variables are better served by just using client.submit and something like `as_completed`.  

Is there a plain-english or psuedocode explanation of the algorithm that you're trying to parallelize somewhere?  If so I'd be happy to take a look and possibly recommend a dask example that might be more appropriate.

> Testing (this is probably the hardest part, but definitely needed)

For testing I strongly recommend using [dask's testing utilities](http://distributed.readthedocs.io/en/latest/develop.html#writing-tests).  They start up and tear down clusters either in multiple processes or in the same thread (using tornado) cleanly and robustly.  There are examples in that section of the documentation that are hopefully easy to copy-paste.  *If* you are familiar with async programming and tornado I recommend the `gen_cluster` approach.  It will allow you to introspect the state of the cluster at any point.  However if you're *not* familiar with async programming then you probably want to stick with the context manager `with cluster(...)` approach.

menshikh-iv:
@darindf any updates?

@mrocklin thanks for the advice, we'll ping you if any "dask-related" problems happens :)

darindf:
I'm in the middle of testing. The design pattern I used is similar to that from LdaMulticore. I thought I had it working but came across a test case where the word distribution seems odd, in that one word appears significantly in multiple topics, and when I resort back to using LdaModel, this distribution disappears.

menshikh-iv:
@darindf interesting effect, is this happens always?
Can you post some benchmark results of the current version (for understand, how this improve performance)?

menshikh-iv:
ping @darindf, any updates?

darindf:
There's some issue when computing in chunksizes that I'm trying to work out.

PR #2058, opened 2018-05-23
Fix loading fasttext models that don't contain ngrams

leezu:
If the fasttext model was trained without ngrams, the bucket size == 0.
Subsequently the current code would fail with a ZeroDivisionError when trying to
look up vectors in the bucket of size 0.

This PR just wraps the ngram loading part in `if self.bucket != 0:`.

### DIFF
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -866,33 +866,30 @@ class FastTextTrainables(Word2VecTrainables):
         for w, vocab in wv.vocab.items():
             wv.vectors[vocab.index] += np.array(wv.vectors_ngrams[vocab.index])
 
-        ngram_indices = []
-        wv.num_ngram_vectors = 0
-        for word in wv.vocab.keys():
-            for ngram in _compute_ngrams(word, wv.min_n, wv.max_n):
-                ngram_hash = _ft_hash(ngram) % self.bucket
-                if ngram_hash in wv.hash2index:
-                    continue
-                wv.hash2index[ngram_hash] = len(ngram_indices)
-                ngram_indices.append(len(wv.vocab) + ngram_hash)
-        wv.num_ngram_vectors = len(ngram_indices)
-        wv.vectors_ngrams = wv.vectors_ngrams.take(ngram_indices, axis=0)
+        logger.info("loading weights for %s words for fastText model from %s",
+                    len(wv.vocab), file_name)
 
-        ngram_weights = wv.vectors_ngrams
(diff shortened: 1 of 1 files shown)


leezu:
Are there any concerns about this PR?

piskvorky:
Is it common to train Fasttext without ngrams though?

Otherwise the PR looks OK to me. CC @manneshiva .

gensim/models/fasttext.py
```diff
@@ -866,33 +866,30 @@ def init_ngrams_post_load(self, file_name, wv):
...
-                ngram_hash = _ft_hash(ngram) % self.bucket
-                if ngram_hash in wv.hash2index:
-                    continue
-                wv.hash2index[ngram_hash] = len(ngram_indices)
-                ngram_indices.append(len(wv.vocab) + ngram_hash)
-        wv.num_ngram_vectors = len(ngram_indices)
-        wv.vectors_ngrams = wv.vectors_ngrams.take(ngram_indices, axis=0)
-
-        ngram_weights = wv.vectors_ngrams
+        logger.info("loading weights for %s words for fastText model from %s",
```
piskvorky: Hanging indent please (here and below).

gensim/models/fasttext.py
```diff
@@ -866,33 +866,30 @@ def init_ngrams_post_load(self, file_name, wv):
...
+                    ngram_indices.append(len(wv.vocab) + ngram_hash)
+            wv.num_ngram_vectors = len(ngram_indices)
+            wv.vectors_ngrams = wv.vectors_ngrams.take(ngram_indices, axis=0)
 
-        for w, vocab in wv.vocab.items():
-            word_ngrams = _compute_ngrams(w, wv.min_n, wv.max_n)
-            for word_ngram in word_ngrams:
-                vec_idx = wv.hash2index[_ft_hash(word_ngram) % self.bucket]
-                wv.vectors[vocab.index] += np.array(ngram_weights[vec_idx])
+            ngram_weights = wv.vectors_ngrams
```
piskvorky: Is this to save one attribute lookup? For visual or performance reasons?

menshikh-iv:
@leezu can you please add a test where you load ft model without n-grams and use it

menshikh-iv:
Ping @leezu, are you plan to finish PR?

menshikh-iv:
@mpenkov I close current PR (abandoned), but please check this case too in your #2313 (not sure how common is it, but deserve to check and process it correctly, not raise an exception)

PR #2064, opened 2018-05-25
CUDA support via cupy

rmalouf:
Here's a first draft of adding CUDA acceleration to LdaModel using cupy. There's still lots of work to do on it (e.g., persistence doesn't work), but I wanted to post this to get feedback before I get too far into it.  

As a nice side effect, the changes necessary to get this to work also speed things up a bit even without using a GPU.

### DIFF
--- a/gensim/_matutils.c
+++ b/gensim/_matutils.c
@@ -753,7 +753,7 @@ static const char *__pyx_filename;
 
 
 static const char *__pyx_f[] = {
-  "gensim/_matutils.pyx",
+  "_matutils.pyx",
   "__init__.pxd",
   "stringsource",
   "type.pxd",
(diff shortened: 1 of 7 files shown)


menshikh-iv:
Hi @rmalouf, thanks for PR, I have several questions:

**Performance**:
1. Can you make benchmark that compare 
  - without `cupy`
  - with `cupy` & without `GPU`
  - with `cupy` + `GPU`
2. How this will work with `LdaMulticore` (or multicore have no sense in this case?), I'm worried about resourses (overutilization -> slowdown).

**Compatibility**:
1. How this will change packaging process (Add CUDA/cupy dependency are serious step). Have we any problems with building an wheels for `Gensim` in this case?

menshikh-iv:
ping @rmalouf, please answer to https://github.com/RaRe-Technologies/gensim/pull/2064#issuecomment-399311872

rmalouf:
Sorry to vanish on you like that.  In running benchmarks I ran into a race condition somewhere in cupy, I think, but I wasn't able to isolate it before getting pulled off on other projects.  

I can say that I was getting modest but real speedups using a Titan X over multicore CPU processing (in the 2x-10x range, depending on lots of factors). So, not worth buying a GPU just for topic models, but maybe useful to have the option if you've already got a bunch of them anyway.

If you'd like to close out this PR for now, go ahead and I'll open a new one if and when I have solid reproducible performance numbers.

menshikh-iv:
Ok, anyway, thanks for PR @rmalouf, please let me know if you will have some results (create or re-open current PR) or want to discuss something (just ping me on GitHub)

I think I'll have a look to cupy myself too (looks promising, but I'm not sure how much we need it)

PR #2066, opened 2018-05-25
Bug inconsistent shape of matrix returned by scipy2scipy_clipped Fixes #2065

### LINKED ISSUE #2065: scipy2scipy_clipped may return a matrix with a different shape to that of the input matrix
#### Description
The function scipy2scipy_clipped may return a clipped matrix with a different shape if the last dimension item is not among the top similar items of any row of the input matrix. This is particularly possible while chunking in SparseMatrixSimilarity, as the similarity matrix is incomplete and thus we don't get to see the last column of the last row (which, being a similarity matrix, usually contains 1)

Example:
```
from scipy.sparse import random, vstack
from gensim.matutils import scipy2scipy_clipped
from sklearn.metrics.pairwise import cosine_similarity

#Some random sparse matrix
X = random(1000, 2000, density=.2, format="csc")

#Getting its similarity matrix
X_sim = cosine_similarity(X, dense_output=False)

#Splitting it to simulate chunking
X_sim_chunk1 = X[:500, :]
X_sim_chunk2 = X[500:, :]

#Assuring that in the first chunk no row is similar to the last item
X_sim_chunk1[:, -1] = 0

X_clipped1 = scipy2scipy_clipped(X_sim_chunk1, 100)
print(X_clipped1.shape) # (500, 1999)

X_clipped2 = scipy2scipy_clipped(X_sim_chunk2, 100)
print(X_clipped2.shape) # (500, 2000)

#While trying to recreate the matrix, this fails because of dimensions' inconsistency
vstack([X_clipped1, X_clipped2])
# ValueError: incompatible dimensions for axis 1

```

#### Expected Results
```
X_clipped1 = scipy2scipy_clipped(X_sim_chunk1, 100)
print(X_clipped1.shape) # (500, 1000)
```
#### Actual Results
```
X_clipped1 = scipy2scipy_clipped(X_sim_chunk1, 100)
print(X_clipped1.shape) # (500, 999)
```
#### Versions
Linux-4.4.0-116-generic-x86_64-with-debian-stretch-sid
('Python', '2.7.14 |Anaconda, Inc.| (default, Nov  8 2017, 22:44:41) \n[GCC 7.2.0]')
('NumPy', '1.13.3')
('SciPy', '1.0.0')
('gensim', '3.4.0')
('FAST_VERSION', 1)

psorianom:
Fixed inconsistent shape of matrix returned by scipy2scipy_clipped, as described in issue #2065 
Solved by explicitly setting the second dimension of the new matrix to that of the input matrix.

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -310,7 +310,7 @@ def scipy2scipy_clipped(matrix, topn, eps=1e-9):
         # Instantiate and return a sparse csr_matrix which preserves the order of indices/data.
         return scipy.sparse.csr.csr_matrix(
             (matrix_data, matrix_indices, matrix_indptr),
-            shape=(matrix.shape[0], np.max(matrix_indices) + 1)
+            shape=(matrix.shape[0], matrix.shape[1])
         )
 
 


menshikh-iv:
hi @psorianom, can you please add tests for this function (at least a case that you described in issue)?

psorianom:
hey @menshikh-iv , thanks for checking this out.  Sure, I will do it as soon as possible.

piskvorky:
The entire function looks strange. Its documentation doesn't match the code comments:

`Return a scipy.sparse vector/matrix consisting of 'topn' elements of the greatest magnitude (absolute value).`
versus
`# Sort and clip each row vector first.`
versus
`# use np.argpartition/argsort and only form tuples that are actually returned.` (??)

@menshikh-iv do we have good unit tests here? The code looks dodgy.

menshikh-iv:
@piskvorky we have only one test for this function

https://github.com/RaRe-Technologies/gensim/blob/9c6db73919d032ab2f6ea35b3a9043e3b0d2aed5/gensim/test/test_similarities.py#L94-L110

psorianom:
> The entire function looks strange. Its documentation doesn't match the code comments:
> 
> `Return a scipy.sparse vector/matrix consisting of 'topn' elements of the greatest magnitude (absolute value).`
> versus
> `# Sort and clip each row vector first.` (so it's actually `topn` per row?)
> versus
> `# use np.argpartition/argsort and only form tuples that are actually returned.` (??)
> 
Hi all, sorry for the delay.

Regarding the comments, yes, it is per row, so if the input is an array of n rows, the output will have n rows. This differs from `full2sparse` which takes row by row (see  in `SimilarityABC` interface:             `return [matutils.full2sparse_clipped(v, self.num_best) for v in result]`)

All in all, I realize that the bug I created is not actually a bug per se, i.e., the code does what it should do according to its description (return the 'topn' elements of the greatest magnitude) without regards to the original size of the matrix.

In other to keep the same matrix dimension, maybe adding a flag to  `full2sparse_clipped`  and `scipy2scipy_clipped` such as "keep_dimensions" would do the trick ? What do you think ?

> @menshikh-iv do we have good unit tests here? The code looks dodgy.

menshikh-iv:
@psorianom hello, sorry for delay

> In my use case, I do need the original size of the matrix, as I use it later for other calculations. So maybe adding a flag to full2sparse_clipped and scipy2scipy_clipped such as "keep_size" would do the trick ?

maybe, sounds compatible :+1:

piskvorky:
If so, it needs a clear description + motivation. I still find the docstrings confusing, and adding extra parameters won't help.

PR #2070, opened 2018-05-28
Fixed Typo and increased performance in analyze_sentence

JonathanHourany:
Type on  line 147. Sentence is "...happy ending has it won't..." where "has" should be "as"

### DIFF
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -141,11 +141,12 @@ class SentenceAnalyzer(object):
 
         """
         s = [utils.any2utf8(w) for w in sentence]
-        last_uncommon = None
-        in_between = []
         # adding None is a trick that helps getting an automatic happy ending
         # as it won't be a common_word, nor score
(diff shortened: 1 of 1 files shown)


JonathanHourany:
Fixing a minor typo

piskvorky:
Thanks @JonathanHourany . Looking at the code, that whole "trick" is questionable though -- adding two lists together will create a new list, which is a very costly operation in such a tight spot.

So I don't think that's a good idea in the first place, but we wanted to replace that code by an optimized Cython loop anyway…

JonathanHourany:
I agree, seems it'd be better to append `None` instead. If a Cython implementation is far off I could make the change in a new commit if you feel it's worth it.

piskvorky:
Yes please!

piskvorky:
Thanks @JonathanHourany !

JonathanHourany:
Thank you!

PR #2071, opened 2018-05-29
Improve "predict_output_word" to be possible to predict output word from context vectors not only context words.

ukiuki-satoshi:
hi.

i would like to predict output word probability from modified or retrofitted vectors of context words.

but Word2Vec.predict_output_word can predict output word just from original vectors of cotext words in vocab.

i fixed it also possible to recieve vectors not only words as input in this PR.

i'm a github newbie, so if any nit picking exists i will humbly fix it. thank you.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -124,7 +124,7 @@ except ImportError:
 
 from numpy import exp, dot, zeros, random, dtype, float32 as REAL,\
     uint32, seterr, array, uint8, vstack, fromstring, sqrt,\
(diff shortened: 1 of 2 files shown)


menshikh-iv:
Hi @ukiuki-satoshi,

I'm not sure that this is a good idea because in the `develop` you use some fitted model X and vectors FROM X (this is fully consistent), your change allows some mistakes/miss-consistency from user-side, any motivation/usecases (i.e. when this needed and why) for this change?

menshikh-iv:
Closing as abandoned.

PR #2076, opened 2018-06-03
self.metadata field was overwritten to False

tarikaltuncu:
It was overwritten to false and this makes contradiction to documents

### DIFF
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -512,7 +512,7 @@ class WikiCorpus(TextCorpus):
 
     def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), dictionary=None,
                  filter_namespaces=('0',), tokenizer_func=tokenize, article_min_tokens=ARTICLE_MIN_WORDS,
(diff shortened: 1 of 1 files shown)


gensim/corpora/wikicorpus.py
```diff
@@ -512,7 +512,7 @@ class WikiCorpus(TextCorpus):
 
     def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), dictionary=None,
                  filter_namespaces=('0',), tokenizer_func=tokenize, article_min_tokens=ARTICLE_MIN_WORDS,
-                 token_min_len=TOKEN_MIN_LEN, token_max_len=TOKEN_MAX_LEN, lower=True):
+                 token_min_len=TOKEN_MIN_LEN, token_max_len=TOKEN_MAX_LEN, lower=True, metadata=True):
```
piskvorky: This seems to change the existing behaviour from `False` to `True`. Why this change?

menshikh-iv:
hi @tarikaltuncu

> this makes contradiction to documents

what's document?

also, please answer to https://github.com/RaRe-Technologies/gensim/pull/2076#discussion_r204221096 comment

tarikaltuncu:
Hi, on double check either I was confused or the documentation has been fixed. So it seems as it should be now.

PR #2078, opened 2018-06-04
[GSoC 2018] Multistream API for vocabulary building in *2vec

persiyanov:
This PR introduces __multi-stream API__ for building vocabularies in __*2Vec models__. 

# Main idea

Training *2vec models involves two main steps:

1. One pass over your training corpus to build (and filter) the vocabulary
2. Train the model

This pull request optimizes the __first step__ (dictionary building).

The main idea is to parallelize the `scan_vocab` method, by allowing users to pass _several streams of data_ (i.e. several `LineSentence` iterators for different files). Under the hood, the implementation uses `multiprocessing` python module creating `multiprocessing.Pool(processes=min(workers, len(input_streams)))` to __read and process all input streams in parallel__.

# Why use it?

This new feature substantially reduces the vocabulary building time, especially if you have complex data iterators (more CPU-bound, i.e. reading a line from a file, then doing complex preprocessing, removing stop words, lemmatizing, stemming etc).

In our benchmarks below, __multi-stream mode__ produces up to __3x boost__ in the vocabulary building step. This is significant because this step can take hours for larger corpora (e.g. on https://dumps.wikimedia.org/enwiki/20180320/ :)

The more CPU-complex data iterators you use, the more speedup you will get. For example, I've experimented with building vocab for `Word2Vec` model in two setups, **without preprocessing** and **with preprocessing**:

1. **Without preprocessing**. Iterate with `gensim.models.word2vec.LineSentence`, which just reads a line from disk and splits it on whitespace. I ran this benchmark on already preprocessed dataset. Using multi-stream resulted in **2.7x speedup**.

| # workers | total time, secs |
|:---------:|:-:|
|     1     |      249.51     |
|     2     |      159.87     |
|     3     |      126.47     |
|     4     |      109.90     |
|     5     |      96.79      |
|     8     |      88.74      |
|     10    |      93.00      |

2. **With preprocessing**. Iterate over lines and preprocess each one with `gensim.parsing.preprocessing.preprocess_string`. In this case, multistream mode has provided **6.9x speedup**.

| # workers | total time, secs |
|:---------:|:-:|
|     1     |      4398.54     |
|     2     |      2252.73     |
|     3     |      1596.28     |
|     4     |      1157.02     |
|     5     |      926.66      |
|     8     |      704.07      |
|     10    |      638.45      |

# How to use it?

It's really simple to start using the multi-stream mode. All you need is to have multiple document streams (training corpora, e.g. files on disk) you want to train on, and pass them as a list into the new `input_streams` constructor parameter. `input_streams` supersedes the  old `sentences` parameter which is now equivalent to `input_streams = [sentences]`:

### Old single stream approach

1. Build vocabulary and train at the same time

```python
sentences = LineSentence('myfile.txt')
model = Word2Vec(sentences, workers=10, ...other params...)
```

2. Initialize model, build vocab and train as separate steps

```python
sentences = LineSentence('myfile.txt')
model = Word2Vec(workers=10, ...other params...)
model.build_vocab(sentences)
model.train(sentences, ...train params...)
```

### New multi-stream approach

1. Build vocab and train at the same time (parallelized internally)

```python
input_streams = [LineSentence('myfile1.txt'), LineSentence('myfile2.txt'), ...]
model = Word2Vec(input_streams=input_streams, workers=10, ...other params...)
```

2. Initialize model, build vocab in parallel and train as separate steps

```python
input_streams = [LineSentence('myfile1.txt'), LineSentence('myfile2.txt'), ...]
model = Word2Vec(workers=10, ...other params...)
model.build_vocab(input_streams=input_streams)  # Optionally, you can pass `workers` param here which differs from the one passed in constructor.
model.train(input_streams=input_streams, ...train params...)
```

The examples above used `Word2Vec`, but `Doc2Vec` and `FastText` work analogously.


persiyanov:
```
* 2018-06-04 14:11:10,727 : INFO : Model = word2vec	Workers = 1	Vocab time = 220.48 secs
*  2018-06-04 14:13:39,131 : INFO : Model = word2vec	Workers = 2	Vocab time = 147.40 secs
*  2018-06-04 14:15:44,110 : INFO : Model = word2vec	Workers = 3	Vocab time = 123.86 secs
*  2018-06-04 14:17:30,654 : INFO : Model = word2vec	Workers = 4	Vocab time = 105.28 secs
*  2018-06-04 14:19:00,277 : INFO : Model = word2vec	Workers = 5	Vocab time = 88.51 secs
*  2018-06-04 14:20:24,006 : INFO : Model = word2vec	Workers = 8	Vocab time = 82.50 secs
*  2018-06-04 14:21:43,277 : INFO : Model = word2vec	Workers = 10	Vocab time = 77.99 secs

*  2018-06-04 15:09:04,340 : INFO : Model = doc2vec	Workers = 1	Vocab time = 223.81 secs
*  2018-06-04 15:11:39,150 : INFO : Model = doc2vec	Workers = 2	Vocab time = 153.86 secs
*  2018-06-04 15:13:51,637 : INFO : Model = doc2vec	Workers = 3	Vocab time = 131.49 secs
*  2018-06-04 15:15:43,956 : INFO : Model = doc2vec	Workers = 4	Vocab time = 111.28 secs
*  2018-06-04 15:17:23,360 : INFO : Model = doc2vec	Workers = 5	Vocab time = 98.30 secs
*  2018-06-04 15:18:55,534 : INFO : Model = doc2vec	Workers = 8	Vocab time = 91.04 secs
*  2018-06-04 15:20:22,721 : INFO : Model = doc2vec	Workers = 10	Vocab time = 86.03 secs

* 2018-06-04 14:43:46,260 : INFO : Model = fasttext	Workers = 1	Vocab time = 309.86 secs
*  2018-06-04 14:47:52,583 : INFO : Model = fasttext	Workers = 2	Vocab time = 244.43 secs
*  2018-06-04 14:51:32,450 : INFO : Model = fasttext	Workers = 3	Vocab time = 217.69 secs
*  2018-06-04 14:54:55,467 : INFO : Model = fasttext	Workers = 4	Vocab time = 201.10 secs
*  2018-06-04 14:58:06,800 : INFO : Model = fasttext	Workers = 5	Vocab time = 189.07 secs
*  2018-06-04 15:01:08,160 : INFO : Model = fasttext	Workers = 8	Vocab time = 179.37 secs
*  2018-06-04 15:04:05,421 : INFO : Model = fasttext	Workers = 10	Vocab time = 175.31 secs
```

persiyanov:
@gojomo Hi! I'd like to ask for your advice. There is one bug in the code in this PR related to `TaggedLineSentence` and tags produced by it. If a user wants to use multistream + `TaggedLineSentence`, then there will be several documents with the same tags (any two documents on the same line in different files will have the same tags, because of `TaggedLineSentence` logic). Could you suggest how to fix this problem?

gojomo:
Clearly incrementing-serial-numbers-per-line can't be used if you have multiple files that all start with line 0. You could require tags to be specified inside the files – as an extra field/fields each line. If the files are considered in a stable order, and the count of texts (lines) in each can be pre-determined (perhaps with another extra pass over all data), each could be given a unique starting-id. Text tags could also be a deterministic function of the hash of the contained text – for additional per-line overhead, and the extra cost of keeping those (now much longer) tags as string keys.

gensim/models/base_any2vec.py
```diff
@@ -330,12 +334,14 @@ def __init__(self, sentences=None, workers=3, vector_size=100, epochs=5, callbac
                 self.neg_labels[0] = 1.
 
         if sentences is not None:
-            if isinstance(sentences, GeneratorType):
+            if multistream and not isinstance(sentences, (tuple, list)):
+                raise TypeError("If multistream=True, you must pass tuple or list as the sentences argument.")
+            if not multistream and isinstance(sentences, GeneratorType):
```
menshikh-iv: This check (about generators) should be applied for each multistream element too (i.e. to each stream)

gensim/models/base_any2vec.py
```diff
@@ -459,25 +465,36 @@ def __str__(self):
...
             The `sentences` iterable can be simply a list of lists of tokens, but for larger corpora,
             consider an iterable that streams the sentences directly from disk/network.
             See :class:`~gensim.models.word2vec.BrownCorpus`, :class:`~gensim.models.word2vec.Text8Corpus`
             or :class:`~gensim.models.word2vec.LineSentence` in :mod:`~gensim.models.word2vec` module for such examples.
+            If `multistream=True`, `sentences` must be list or tuple of input streams described above.
+        multistream : bool
+            If True, use `sentences` as list of input streams and speed up vocab building by parallelization
+            with `min(len(sentences), self.workers)` processes. This option can lead up to 2.5x reduction
+            in vocabulary building time.
+        workers : int
```
menshikh-iv: maybe better use different naming like `multistream_workers` (for avoiding potential collision by parameter names) or this has no sense @persiyanov?

gensim/models/base_any2vec.py
```diff
@@ -213,6 +214,9 @@ def _train_epoch(self, data_iterable, cur_epoch=0, total_examples=None,
             for _ in xrange(self.workers)
         ]
 
+        # Chain all input streams into one, because multistream training is not supported yet.
+        if multistream:
```
menshikh-iv: Better to raise `warnings.warn` about lack of support (I mean explicitly to the user, not only comment)

gensim/models/base_any2vec.py
```diff
@@ -459,25 +465,36 @@ def __str__(self):
...
+        workers : int
+            Used if `multistream=True`. Determines how many processes to use for vocab building.
         update : bool
             If true, the new words in `sentences` will be added to model's vocab.
         progress_per : int
             Indicates how many words to process before showing/updating the progress.
 
         """
+        if workers is None:
+            workers = self.workers
```
menshikh-iv: `workers = workers or self.workers`, here and everywhere.

gensim/utils.py
```diff
@@ -1709,6 +1709,28 @@ def prune_vocab(vocab, min_reduce, trim_rule=None):
     return result
 
 
+def merge_dicts(dict1, dict2):
+    """Merge `dict1` of (word, freq1) and `dict2` of (word, freq2) into `dict1` of (word, freq1+freq2).
+    Parameters
+    ----------
+    dict1 : dict
```
menshikh-iv: `dict of (str, int)`, here and everywhere.

gensim/models/word2vec.py
```diff
@@ -1203,7 +1280,7 @@ def sort_vocab(self, wv):
             wv.vocab[word].index = i
 
     def prepare_vocab(self, hs, negative, wv, update=False, keep_raw_vocab=False, trim_rule=None,
-                      min_count=None, sample=None, dry_run=False):
+                      min_count=None, sample=None, dry_run=False, **kwargs):
```
menshikh-iv: why `**kwargs` ?

gensim/models/word2vec.py
```diff
@@ -1186,12 +1219,56 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
...
+
+    def _scan_vocab_multistream(self, input_streams, workers, trim_rule):
+        manager = multiprocessing.Manager()
+        progress_queue = manager.Queue()
+
+        logger.info("Scanning vocab in %i processes.", min(workers, len(input_streams)))
+        pool = multiprocessing.Pool(processes=min(workers, len(input_streams)))
+
+        results = [
+            pool.apply_async(_scan_vocab_worker,
```
menshikh-iv: why `apply_async` ?

gensim/models/doc2vec.py
```diff
@@ -789,14 +802,101 @@ def build_vocab_from_freq(self, word_freq, keep_raw_vocab=False, corpus_count=No
...
+            elif isinstance(report, string_types):
+                logger.warning(report)
+            else:
+                num_words, num_documents = report
+                total_words += num_words
+                document_no += num_documents
+
+        corpus_count = document_no + 1
+        results = [res.get() for res in results]  # pairs (vocab, doclen2tags)
+        self.raw_vocab = reduce(utils.merge_dicts, [r[0] for r in results])
```
menshikh-iv: `from functools import reduce`, Guido doesn't like `reduce` function, for this reason `reduce` was hidden in python3.

menshikh-iv:
@persiyanov great work! Please add tests & merge fresh `master` to your branch (just in case).

@gojomo problem mostly defined by user-behavior, i.e. if I (as gensim-user) heard about new multistream feature and want to try it with d2v - I'll use something like  
`[TaggedLineDocument(..), TaggedLineDocument(..)], multistream=True` (that's definitely will not work as we expected in this case).

gensim/models/base_any2vec.py
persiyanov: Hm, I don't think so:

We add new functionality, allowing users to use `multistream` mode. We added this functionality to the gensim core, so we are sure that it works correctly. Why should we __warn__ the user that something is wrong? Furthermore, user __can't do anything__ to get rid of this warning message while continuing using multistream mode.

gensim/models/word2vec.py
persiyanov: What else?

gensim/models/base_any2vec.py
persiyanov: I think there is no point in this. In most cases, by setting `workers` parameter user means the same degree of parallelization for both scan vocab (multiprocessing) and training the model (multithreading), IMO

gensim/models/doc2vec.py
persiyanov: done

persiyanov:
hooray! Tests are passing now. I will update the description of the PR in a proper way (what's the feature, how it helps, how to use it etc.)

gensim/models/base_any2vec.py
menshikh-iv: About **can't do anything** - not true, see https://docs.python.org/2/library/warnings.html#temporarily-suppressing-warnings (that's FYI), but I agree with your point in this case.

gensim/models/word2vec.py
menshikh-iv: still a question

gensim/test/test_word2vec.py
```diff
@@ -480,6 +494,31 @@ def testTraining(self):
...
+        input_streams = [sentences[:len(sentences) // 2], sentences[len(sentences) // 2:]]
+        model = word2vec.Word2Vec(size=2, min_count=1, hs=1, negative=0, workers=1, seed=42)
+        model.build_vocab(input_streams, multistream=True)
+
+        self.assertTrue(model.wv.syn0.shape == (len(model.wv.vocab), 2))
+        self.assertTrue(model.syn1.shape == (len(model.wv.vocab), 2))
+
+        model.train(input_streams, total_examples=model.corpus_count, epochs=model.iter, multistream=True)
+        sims = model.most_similar('graph', topn=10)
+        # self.assertTrue(sims[0][0] == 'trees', sims)  # most similar
```
menshikh-iv: maybe remove commented lines?

gensim/models/word2vec.py
menshikh-iv: apply/map/something else, I don't understand your pick here.

gensim/models/base_any2vec.py
```diff
@@ -17,6 +17,7 @@
 from types import GeneratorType
 from gensim.utils import deprecated
 import warnings
```
menshikh-iv: Don't forget to raise warning for d2v-tags case

gensim/models/word2vec.py
persiyanov: apply/map are blocking.

menshikh-iv:
Nice @persiyanov, LGTM :+1:, @piskvorky please have a look.

gensim/models/word2vec.py
```diff
@@ -1144,9 +1152,43 @@ def __iter__(self):
...
+    sentence_no = -1
+    total_words = 0
+    for sentence_no, sentence in enumerate(stream):
+        if not checked_string_types:
+            if isinstance(sentence, string_types):
+                log_msg = "Each 'sentences' item should be a list of words (usually unicode strings). " \
+                          "First item here is instead plain %s." % type(sentence)
+                progress_queue.put(log_msg)
+
+            checked_string_types += 1
```
jayantj: This seems to be functioning as a boolean in this construct. Any particular reason for defining it as an integer and incrementing?

gensim/models/word2vec.py
```diff
@@ -1186,12 +1226,56 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
...
+        results = [
+            pool.apply_async(_scan_vocab_worker,
+                             (stream, progress_queue, self.max_vocab_size, trim_rule)
+                             ) for stream in input_streams
+        ]
+        pool.close()
+
+        unfinished_tasks = len(results)
+        total_words = 0
+        sentence_no = -1
```
jayantj: The semantics might be clearer here if this was called `total_sentences` and initialized to 0. Unless I'm misunderstanding something

gensim/utils.py
```diff
@@ -1709,6 +1709,28 @@ def prune_vocab(vocab, min_reduce, trim_rule=None):
     return result
 
 
+def merge_dicts(dict1, dict2):
```
jayantj: IMO `merge_dicts` seems too general a name for an operation as specific as this, it's slightly confusing. What do you think about `aggregate_counts` or `merge_counts`?

gensim/models/base_any2vec.py
```diff
@@ -330,12 +334,18 @@ def __init__(self, sentences=None, workers=3, vector_size=100, epochs=5, callbac
                 self.neg_labels[0] = 1.
 
         if sentences is not None:
-            if isinstance(sentences, GeneratorType):
+            if multistream:
+                if not isinstance(sentences, (tuple, list)):
+                    raise TypeError("If multistream=True, you must pass tuple or list as the sentences argument.")
+                if any(isinstance(stream, GeneratorType) for stream in sentences):
+                    raise TypeError("You can't pass a generators as input streams. Try an iterator.")
```
jayantj: Minor language fix: `a generator`

gensim/models/word2vec.py
```diff
@@ -1144,9 +1152,43 @@ def __iter__(self):
...
+                progress_queue.put(log_msg)
+
+            checked_string_types += 1
+
+        for word in sentence:
+            vocab[word] += 1
+
+        if max_vocab_size and len(vocab) > max_vocab_size:
+            utils.prune_vocab(vocab, min_reduce, trim_rule=trim_rule)
+            min_reduce += 1
```
jayantj: If I understand correctly, this step should also be applied to the final merged vocab, in addition to each individual vocab?

gensim/models/word2vec.py
persiyanov: I've just copied this logic from an old version of the code (I don't know why it's not bool, but I could fix it if you insist)

gensim/models/word2vec.py
persiyanov: Agree

gensim/models/base_any2vec.py
persiyanov: No, here I explicitly check that `streams` (elements of `sentences` list) are not generators.

gensim/utils.py
persiyanov: Agree, `merge_counts` sounds good

gensim/models/word2vec.py
persiyanov: `max_vocab_size` by itself doesn't guarantee that after vocabulary building its size will be `<= max_vocab_size`. This value is only used to throw away rare tokens during the building time. 

After "reduce" step (self.raw_vocab = `reduce(utils.merge_counts(....))`) another vocabulary method `prepare_vocab` is called, which prunes vocabulary appropriately.

gensim/models/base_any2vec.py
gojomo: @jayantj means that it should be 'generator' singular rather than 'generators' plural.

gensim/models/word2vec.py
gojomo: The `max_vocab_size` is a crude scan-time cap, but for a while it was the only cap available, and it does have the effect in prior code (since all vocabulary was discovered via the single-threaded process) that the final vocabulary could not possibly be larger than than its value. 

I don't see a way the later `prepare_vocab()` would apply that same cap (unless the user also specifies the newer, more precise `max_final_vocab` parameter). So the concern here is that a previously-reliable invariant – you'd never wind up with a vocabulary larger than `max_vocab_size` – may no longer apply.

persiyanov: Yeah, I meant exactly `max_final_vocab` parameters. But it's true that if `max_final_vocab` isn't passed, the property `len(vocab) <= max_vocab_size` may not hold. 

Is it okay to prune the vocab after merging by sorting and selecting top `max_vocab_size` words by their frequency?

persiyanov: @gojomo Actually, current single stream code can't guarantee `len(vocab) <= max_vocab_size` after vocabulary building. I think there is a bug in the code (https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec.py#L1185)

Say, you have only one sentence `a a a b b b b` and you set `max_vocab_size=1`. After only one iteration in scan_vocab, vocabulary will be the same `{"a": 3, "b": 4}` (pruning is called like this `utils.prune_vocab({"a": 3, "b": 4}, min_reduce=1) => {"a": 3, "b": 4}`).

piskvorky: Yes, it's OK. Don't forget the main point of this parameter is not to blow up the RAM, it's there for a reason (applies even more to multistream).

persiyanov:
@gojomo @jayantj @piskvorky Please, take one more look. I've added final vocabulary pruning

gensim/models/word2vec.py
gojomo: @persiyanov Yes, that looks like a bug, though one that would be vanishingly rare with more typical parameters (& Zipfian frequencies). I happen to really dislike this mechanism, because it's prone to misunderstanding - someone who, for example, would be happy with a 100,000 word vocabulary and sets `max_vocab_size=100000` might wind up with a vocab of far less if one of its draconian, escalating prunes (which often take out 90% of all words) happens just before the end-of-scan. But it'd be wise to either retain its (usual) hard-cap effect (which some may be relying upon), or better document that any true hard-cap should be enforced through the other parameter.

@piskvorky & @persiyanov In the current code `max_vocab_size` is enforcing any RAM-cap at all in the multistream case, is it? I'm not sure it has a straightforward interpretation for multi-stream, either - neither giving each stream its own `max_vocab_size` dict-size cap, nor giving each stream `max_vocab_size / num_streams` as a cap, does something intuitive/sensible.

jayantj:
Looks fine to me

gensim/utils.py
```diff
@@ -1709,6 +1709,29 @@ def prune_vocab(vocab, min_reduce, trim_rule=None):
...
+    topk : int
+        Number of words with highest frequencies to keep.
+    trim_rule : function, optional
+        Function for trimming entities from vocab, default behaviour is `vocab[w] <= min_count`.
+
+    """
+    if topk >= len(vocab):
+        return
+
+    sorted_vocab = sorted(vocab.keys(), key=lambda word: vocab[word], reverse=True)
```
piskvorky: This looks inefficient, creating two extra copy of all keys: one in `keys()` (which should be `iterkeys()`), another in `sorted_vocab`. As I wrote before, the whole point of these vocabulary trimming dances is to save memory, otherwise it loses the point.

gensim/models/word2vec.py
piskvorky: The objective is to limit the use of memory, so "total RAM used" would be the ideal parameter (ala Bounter). Barring that, a global `max_vocab_size` is a reasonable approximation.

gensim/utils.py
persiyanov: How do you suggest I can eliminate the second copy, `sorted_vocab` variable? The first one is easy: `vocab.keys() -> iterkeys(vocab)`, what about the second?

piskvorky: `heapq.nlargest` over a generator of integer counts (not strings) is one option. That still creates a copy, but its size depends only on `topk`, not `len(vocab)`.

gensim/models/word2vec.py
gojomo: @piskvorky But, note that as currently written in this PR, `max_vocab_size` may be a single global parameter, but is only enforced locally in each stream-reader. So actual potential max-number-of-vocab-entries is `num_readers * max_vocab_size` – which is not likely what a user expects, especially if familiar with its previous single-thread interpretation. And, if any single reader hits the cap locally, it will be pruning its local counts only - other even smaller counts, of the exact same token in other threads, will survive – making any final tallies even more arbitrarily incorrect than in the single-thread application.

gojomo:
If it works, great, but I'm a little uncomfortable with the manner-of-activation: the `multistream` boolean toggle, and thus differing if-this-type-then overloading of a single `sentences` parameter, all as new branches inside existing code.

Ultimately, even if not right away, I'd prefer a mechanism where people opting-into an alternative-vocabulary-building-optimization use some other explicit code path, not complicating existing interfaces, then drop a completed vocabulary-object into the Word2Vec/etc model. Then the simple/legacy path stays simple, the new alternative is clearly opted-into, and the point-of-entanglement is limited to a single injection of what the main model needs to proceed – which also allows the main model to be indifferent as to the specifics of how its individual  steps were completed. That could also allow further experimentation with other vocab-building approaches, without adding more switches-and-branches to a single entry point every time. (That there's only a 2.5X-3X speedup in moving from 1 to 10 threads makes me think further improvements may be possible.)

persiyanov:
@gojomo Could you give some examples on what you mean by `opting-into an alternative-vocabulary-building-optimization use some other explicit code path...`? Something like `vocab = gensim.models.word2vec.build_vocab_multistream_mode(input_streams, <vocab params>)` and then `model.vocab = vocab`? Or something different?

Why do you think the interface I proposed is complicated? It's just toggling one flag (`multistream=True`) and start passing _multiple streams_ instead of single stream. This looks very clear and simple, isn't it?

Is it okay to merge this PR in its current state (after addressing @piskvorky comments) and possibly change the interfaces and API in the future (according to new research & results in __multistream training__ direction) if needed?

gojomo:
@persiyanov Yes, your example utility function `build_vocab_multistream_mode()` that returns a usable vocabulary-object is more along the lines of what I'd prefer. 

The switches-and-branches approach works, and if there's just one alt mode, not too complex. But, it's expanding some (already-gigantic) method-parameter lists, and some already-twisty methods - making them harder to read/maintain/expand in the future. 

If just a switch, the docs for this mode will be interleaved with long lists of other options – which may tend to tempt those who don't quite need this, but shortchange those who do want full descriptions. Having the docs on a specialized function or class may be more clear – and better able to discuss subtleties like the potentially different memory requirements of this approach, or its sensitivity to the various parallel streams being of roughly-equal size.

Also, adding new switches for new approaches in the future, which might or might not be compatible with each other, will be harder than having a separate, swappable step that hides its own details and just leaves things in a proper state. For example, there's another effort in #1962 to offer the approximate-counting `Bounter` as a vocab-surveying option... but it *also* adds a toggle-on-parameter and bunch of if-thens to nearby code. As edits to a single class, these would be hard to merge; as truly standalone paths-to-a-working-vocabulary-object, they'd be easily swappable as either (1) a policy/implementation option supplied to the main class; or (2) a separate step whose output is just provided to the mail class when done. (You can see I've raised similar modularity concerns at that PR.)

The preexisting and not-so-flexible tangle of the *2Vec initialization steps long predates your work, so it may be outside your project to fully fix it, but I'm wary of making the tangle any thicker.

gensim/utils.py
persiyanov: @piskvorky Fixed.

gensim/utils.py
```diff
@@ -1712,7 +1712,7 @@ def prune_vocab(vocab, min_reduce, trim_rule=None):
 
 def trim_vocab_by_freq(vocab, topk, trim_rule=None):
     """Retain `topk` most frequent words in `vocab`.
-    If there are more words with the same frequency as `topk`-th one, they will be dropped.
+    If there are more words with the same frequency as `topk`-th one, they will be keeped.
```
piskvorky: `keeped` => `kept`

gensim/models/base_any2vec.py
```diff
@@ -459,25 +469,34 @@ def __str__(self):
...
-    def build_vocab(self, sentences, update=False, progress_per=10000, keep_raw_vocab=False, trim_rule=None, **kwargs):
+    def build_vocab(self, sentences, multistream=False, workers=None, update=False, progress_per=10000,
+                    keep_raw_vocab=False, trim_rule=None, **kwargs):
         """Build vocabulary from a sequence of sentences (can be a once-only generator stream).
         Each sentence is a iterable of iterables (can simply be a list of unicode strings too).
 
         Parameters
         ----------
-        sentences : iterable of iterables
+        sentences : {iterable of iterables, list or tuple of iterable of iterables}
```
piskvorky: I'd prefer not to override the same parameter with a different type (with the type dependant on an external flag).

It's already complicated enough; `iterable of iterables, or a list or tuple of iterable of iterables` is no longer humanly parseable.

My original suggestion was to have it a sequence of iterables *always*. If backward compatibility is hard to achieve (is it?), keep the legacy parameter `sentences` (maybe deprecate it in time) and promote a new parameter `multistream`. If backward compatibility can be achieved transparently, then just keep using `sentences`.

persiyanov: Correct me if I got you wrong, but you propose smth like this:

`Word2Vec(sentences=None, input_streams=None, ...)` (instead of old `Word2Vec(sentences=None, ...)` )

with the following logic:
1. both `sentences` and `input_streams` can't be passed
2. if `sentences` is passed, then do `input_streams = [sentences]` and go to (3)
3. if `input_streams` is passed, then perform multistream mode


Am I correct about your logic?

piskvorky: Yes. I originally thought users would only pass `sentences`, and we decide automatically whether it's a single stream (legacy, promote to `input_stream = [sentences]`) or already multiple streams (new-style). And keep only that single parameter.

But introducing extra `input_streams` parameter may be a cleaner option. More explicit.

persiyanov: Okay, I'll implement the logic in this way, sounds good

persiyanov: @piskvorky Currently, in a majority of methods (almost all except constructor) the `sentences` parameter is required, not optional. If we introduce new `input_streams` param, both of them need to be optional.

This requires us to do some extra checks e.g. `assert sentences is not None or input_streams is not None` in all of the methods. It sounds a bit dirty

What do you think about that?

piskvorky: Do you see any ambiguity/problems with detecting "legacy" (single stream) `sentences` automatically?

persiyanov: • `sentences` is like a `[['first', 'sentence'], ['second', 'sentence']]`
• `input_streams` is like a `[ [ [ 'first', 'sentence1' ], [ 'second', 'sentence1' ] ] , [ ['first', 'sentence2' ], [ 'second', 'sentence2' ] ] ]`

So, to determine the nesting level (whether its like `sentences` or like `input_streams`) it's necessary to obtain at least the first element from this argument (execute `next(...)`, maybe twice), if this is okay then no problems. 

Pseudocode for this logic:

```
def is_multistream(sentences):
  nesting_level = 0
  current_elem = sentences
  while not isinstance(current_elem, string_types):
    current_elem = next(current_elem)
    nesting_level += 1
  if nesting_level == 2:
      return False  # this is single stream like [['first', 'sentence'], ...]
  elif nesting_level == 3:
      return True   # this is multistream like [ [['first', 'sentence'], ...], ...]
  else:
      raise RuntimeError('this is not possible')
```

piskvorky: Well, we'd have to put the "peeked" items back of course. We cannot just discard them :)

We're already doing something similar in `utils.is_corpus()`.

Question is, is it possible to do something similar here as well? What is the risk of unintended consequences, such as accidentally materializing some huge corpus?

Because it is a bit hackish, so the tradeoff (pros and cons) between having an extra explicit parameter vs being simpler for user is not clearcut to me.

persiyanov: > Question is, is it possible to do something like that here as well? What is the risk of unintended consequences, such as accidentally materializing some huge streamed object in RAM?

I don't see any problems here. This object will be "materialized" in the code, anyway, when we will be iterate over the data.

> Because the peeking is a bit hackish, so the tradeoff (pros and cons) between having an extra explicit parameter vs being simpler for user is not clearcut to me.

Agree, maybe two parameters `sentences=None, input_streams=None` + check that one of them is not None, is more appropriate?

piskvorky: Yeah, alright.

Btw how will N streams interact with M workers? Or do we force N==M?

That could make life harder for the most common use-case, where user has only one stream, N=1. Then they couldn't train in parallel at all.

persiyanov: N streams are joined into one https://github.com/RaRe-Technologies/gensim/pull/2078/files#diff-ce0adce9f4ad6c3715b44dced29c2759R219

piskvorky: Making sure I understand:
1. if M==1, we use the existing algorithm global queue and merge all streams into one, forcing N==1
2. if M > 1, we use an algorithm that does no queuing. And the N streams are distributed among M workers how?

persiyanov: In this pull request, there are no training optimizations at all, only during dictionary building stage. We still use the existing "global queue" algorithm for any number of input_streams (merging them into one).

I don't know yet the way we will go, because currently I don't have positive experiments with multistream training.

persiyanov:
@piskvorky pls, take a look. I've addressed your comments

gensim/models/base_any2vec.py
piskvorky: Dictionary building is a part of training. I'm still unclear what happens when `N != M` (or do we force `N==M` always?).

gensim/models/word2vec.py
piskvorky: Should definitely be `max_vocab / workers`, to keep the same semantics (~ same max memory footprint).

gensim/models/base_any2vec.py
persiyanov: No, we don't force anything. I just use `multiprocessing.Pool`, put `len(input_streams)` tasks in it and that's all.

persiyanov: https://github.com/RaRe-Technologies/gensim/pull/2078/files#diff-673fdc31e9aae23291039b143f451b9eR1236 here is the exact logic I use when creating the pool

menshikh-iv: Small clarification: this is **only** for dictionary building (see PR title), training process stay as is, the goal of PR is parallelized vocab building only.

gensim/models/word2vec.py
persiyanov: @piskvorky fixed

gensim/models/base_any2vec.py
piskvorky: OK, so the task distribution is managed by Pool (at least during this training stage), got it.

gensim/models/word2vec.py
```diff
@@ -1144,9 +1150,43 @@ def __iter__(self):
...
+
+        total_words += len(sentence)
+
+    progress_queue.put((total_words, sentence_no + 1))
+    progress_queue.put(None)
+    return vocab
```
piskvorky: This will try to pickle the entire result (huge!) and send it back to master. Can you think of a way to use shared memory instead?

That should improve both memory (in a big way) as well as maybe performance (some extra locking, but less data sending).

gensim/models/word2vec.py
```diff
@@ -1186,12 +1224,60 @@ def scan_vocab(self, sentences, progress_per=10000, trim_rule=None):
...
+                logger.info("scan vocab task finished, processed %i sentences and %i words;"
+                            " awaiting finish of %i more tasks", total_sentences, total_words, unfinished_tasks)
+            elif isinstance(report, string_types):
+                logger.warning(report)
+            else:
+                num_words, num_sentences = report
+                total_words += num_words
+                total_sentences += num_sentences
+
+        self.raw_vocab = reduce(utils.merge_counts, [res.get() for res in results])
```
piskvorky: Does `.get()` create another copy of each (potentially large) dictionary, or only a reference to an already existing object?

persiyanov: Only returns the reference, as far as I know

gensim/models/base_any2vec.py
```diff
@@ -330,7 +330,37 @@ def _log_epoch_progress(self, progress_queue, job_queue, cur_epoch=0, total_exam
 
     def _train_epoch(self, data_iterable=None, data_iterables=None, cur_epoch=0, total_examples=None,
                      total_words=None, queue_factor=2, report_delay=1.0):
-        """Train one epoch."""
+        """Train the model for a single epoch.
```
gojomo: Not a comment about the doc-improvements here, which all look good, but a side observation about this method, which I only noticed during related reviews last week, is that its strategy of re-launching a fresh 'producer' thread and fresh 'worker' threads for each epoch, as I believe was introduced in #1777, likely drives down overall throughput and CPU utilization compared to the prior strategy. The one potential advantage I'd see for adopting such a full-sync teardown&restart between epochs would be allowing the user to specify some callback for mid-training reporting at each epoch's end – but that hasn't yet ben added.

piskvorky: @gojomo why would that drive down throughput & CPU utilization?

gojomo: When some threads have finished an epoch, but others haven't, cores will be idle not because of GIL/etc but because there's no thread even trying to move forward onto the next epoch's data. Plus any overhead of re-launching threads (magnitude unclear). Old strategy launched exactly `workers + 1` threads. This one launches `epochs * (workers + 1)` threads.

piskvorky: If I understand correctly, you're worried that at the end of each epoch, some threads may be idle (while other threads finish their last job) until the next epoch starts.

Isn't that idleness infinitesimal, since any single job takes almost no time at all?

gensim/models/word2vec.py
piskvorky: Sorry, missed this notification :)

My thoughts: you mean processes, not threads, right? Shrinking could be tricker, I agree. There'd have to be a lock around that (workers pause while shrinking under way -- hopefully not that frequent).

Disabling pruning with multistream is also an elegant solution :-)

persiyanov: @piskvorky 

Pausing all workers while doing shrinking -- sounds very slow. Why are we sure that shrinking is performed not that frequent?

Do you want me to implement this kind of logic and benchmark it? Then decide which logic we will merge into develop branch?

piskvorky: I would hope so! It's a costly operation. That's also why we do it with a growing threshold, instead of always clipping against a constant, e.g. `remove if count <= 5`.

An experiment would be worthwhile IMO. Not necessarily a full-fledged merge-able implementation, but some insight into whether we can get away with sharing a single structure among all workers. It would certainly save us lots of memory.

persiyanov: Okay, I'll do that :wink:

gensim/models/base_any2vec.py
gojomo: I'm not sure of the magnitude, only the direction: this means more idle cores, every time an epoch rolls over. Of course only measuring could tell for sure, and the proportionate impact becomes smaller with larger corpuses. 

As it's not yet clear to me the relative interplay of existing GIL/queue sync bottlenecks that have been preventing higher throughput (or else I would have tried more to fix them), adding yet more thread launches/waits/syncs-against-a-not-yet-filled-queue is something I'd have been wary of doing without measurement at the time. Even the coarse once-a-second progress logging tended to show slower throughput at the beginning of training; that slow-start might now be repeated at each epoch - for example, via GIL-contention between the 1st new thread getting a job and the new producer thread, trying to get ahead of the workers again.

gensim/models/word2vec.py
persiyanov: @piskvorky I've tried two approaches with `manager.dict()`.

1. Replace local dicts in each worker with `manager.dict()` and perform `vocab[word] += 1` on it. Also, in this case, I don't return `vocab` from each worker, but use shared dict at the end of scanning. This is the approach we were talking about above. The code is here https://gist.github.com/persiyanov/76560df460e8b80d4c6c6124452518c0

2. The second approach is still using local `dict()`-s in each worker and not returning them from worker's functions, but perform `manager.dict().update(local_vocab)`. The code is here https://gist.github.com/persiyanov/8537a81fe9f579aa5470ff27287607bf


Unfortunately, both approaches run very slowly and definitely have worse performance than the approach in current PR.

• 1st way (with updating shared vocab immediately with each word) runs `> 15 minutes` for all runs (`workers=2,3,4,5,8,10`), so I didn't even wait till finish.
• 2st way has the following timings:

```
2018-06-24 15:17:05,682 : INFO : Model = word2vec	Workers = 2	Vocab time = 291.13 secs
2018-06-24 15:21:21,838 : INFO : Model = word2vec	Workers = 3	Vocab time = 255.81 secs
2018-06-24 15:25:17,943 : INFO : Model = word2vec	Workers = 4	Vocab time = 235.76 secs
2018-06-24 15:29:01,697 : INFO : Model = word2vec	Workers = 5	Vocab time = 223.41 secs
2018-06-24 15:32:36,754 : INFO : Model = word2vec	Workers = 8	Vocab time = 214.70 secs
2018-06-24 15:35:58,231 : INFO : Model = word2vec	Workers = 10	Vocab time = 201.11 secs
```

It's clearly worse than what we have.

gojomo: Alternate idea: discard `mac_vocab_size`-style trimming entirely, figure out a way to do a precise count, by spilling extras to disk temp files. Strawman implementation: wherever the old code would've trimmed, instead dump the counts that would've been discarded to a key-sorted temp file. At the end, merge all the temp files to a final combined sorted count temp file – and while doing that, retain a histogram of word-frequencies. From that histogram, the portion of the structure in heap can calculate exactly how many words must be dropped to exactly match the specified cap, and all surviving words' counts will be precisely accurate. 

I doubt the performance hit would be very much, and it's fairly parallelizable (there's just more temp files to include in the final merge), the final counts are accurate, and there are fewer gotchas that need documenting because in the end it "just works" as one might expect.

Maybe, using a LevelDB-like log-structured-merge DB might achieve a similar benefit at the same or close performance.

persiyanov: I forget to note I had commented `max_vocab_size`-trimming lines of code in both approaches above. So it's very slow even without trimming at all (and then without its synchronization overhead)

piskvorky: Hm, strange. Must be due to the locking, right?

@persiyanov Any way to turn locking off (since we don't care about accurate counts, especially in the high-frequency tokens where collisions are more likely)?

persiyanov: @piskvorky Yes, I bet this happens because of locks.

> Any way to turn locking off?

I can only see dirty ways to do that, e.g. make nogil Cython analogue for `defaultdict(int)`.

piskvorky: I see. Have we exhausted ideas for memory sharing? Anything else we could try easily?

If not, are we finished with parallelized dict-building?

persiyanov: @piskvorky done, see the updated description

piskvorky: I made some minor language edits, looks good :) Thanks!

persiyanov:
@piskvorky Are you OK with this PR? If so, let's merge it. I need this changes for my next pull request related to multistream training.

piskvorky:
Yeah, I think so. I don't see any outstanding unresolved comments/issues. @menshikh-iv ?

Issue #1172, opened 2017-02-27
Monitoring Training loss


already-taken-m17:
Hi,
How can I monitor training loss of the model if I am using Word2Vec model from gensim.models for training ?
Also, how can I see the logs generated once training has started ? Can I redirect these logs to stdout ?

gojomo:
Reporting of training loss isn't yet offered, but it's a wishlist item discussed in other issues like #999.

gensim uses standard Python logging, and logs many details at INFO or DEBUG levels - so you can configure logging in various ways – https://docs.python.org/2/howto/logging.html – to see those messages wherever you'd like.

menshikh-iv:
Resolved in #1201

villmow:
To monitor loss while training the model set `compute_loss=True` and change [the following lines](https://github.com/RaRe-Technologies/gensim/blob/aaab5b6c54d947a93980b8ce34d43b894cfdbb89/gensim/models/word2vec.py#L1093) to 
```python
logger.info(                                                                                                                                                
    "PROGRESS: at %.2f%% examples, %.0f words/s, in_qsize %i, out_qsize %i, current_loss %.3f",                                                             
    100.0 * example_count / total_examples, trained_word_count / elapsed,                                                                                   
    utils.qsize(job_queue), utils.qsize(progress_queue), self.get_latest_training_loss() / example_count                                                    
)
```
@menshikh-iv Maybe this could be automatically done when `compute_loss=True`?

menshikh-iv:
@villmow maybe, can you create PR and make this change?

kirk86:
Can someone explain why the last training loss that I get is really so high no matter if I use `log-loss` or the `negative sampling`?

`loss: 8837240.0`

gojomo:
@kirk86 If I remember correctly, the values reported are a tally of the loss over a full epoch, so its absolute value needn't necessarily become "small looking", just smaller at the end than the beginning. If you're not seeing it smaller at the end, you may be doing something wrong in your code, such as mismanaging the `alpha`/iterations such that it goes negative. Most users should never need to change the defaults for `alpha` and `min_alpha`, do any alpha-decrements themselves, or call `train()` more than once – if you're doing any of these, you may be following broken examples.  Also, published work on adequately-sized datasets tends to use 10-20 epochs only. If you're still having a problem a good place to discuss (with example code) would the project discussion list: https://groups.google.com/forum/#!forum/gensim

Holmeyoung:
[...]

PR #2081, opened 2018-06-04
override _log_progress() in word2vec to track training loss

usul83:
I came across this suggestion from @villmow in this thread https://github.com/RaRe-Technologies/gensim/issues/1172 and thought I'd implement since @menshikh-iv thought it was a good idea and doesn't seem to have been done by anyone else.

First time I've contributed so appologies if not followed correct process etc.

### DIFF
--- a/gensim/models/base_any2vec.py
+++ b/gensim/models/base_any2vec.py
@@ -645,16 +645,16 @@ class BaseWordEmbeddingsModel(BaseAny2VecModel):
         if total_examples:
             # examples-based progress %
             logger.info(
-                "EPOCH %i - PROGRESS: at %.2f%% examples, %.0f words/s, in_qsize %i, out_qsize %i, current_loss %.3f",
+                "EPOCH %i - PROGRESS: at %.2f%% examples, %.0f words/s, in_qsize %i, out_qsize %i",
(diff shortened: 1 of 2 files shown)


menshikh-iv:
Hi @usul83,

as I see, this is definition from https://github.com/RaRe-Technologies/gensim/blob/408a714e836484e914a7e3622b4f0f99a6d27ddd/gensim/models/base_any2vec.py#L1148, why this copy-paste needed (this method is inherited automatically)?

usul83:
It should be a little different from the any2vec version, it adds: self.get_latest_training_loss() / example_count

menshikh-iv:
@usul83 but in this case, `compute_loss` **must be** True, otherwise - this will not work, need to check this case.

usul83:
It won't work, but it will give a 0 rather than raise an error (since self.running_training_loss = 0 at line 556 whether compute_loss true or not). Maybe could add a check and a warning for that situation?

menshikh-iv:
No warning needed, simply check, if `compute_loss` is False - use parent method, otherwise - your.

menshikh-iv:
this PR fully overlapped by #2135, for this reason, I close current.
@usul83 thanks for your work!

PR #1944, opened 2018-03-01
Refactor documentation for `*2Vec` models

steremma:
In the end this PR will fix the documentation for `BaseWordEmbeddingsModel`, `BaseAny2Vec`, `Word2Vec` and possible `Doc2Vec`. At the moment of opening only the first is done, however its useful to keep it online to get feedback as early as possible

PR #2087, opened 2018-06-09
Fixes to word2vec doc

piskvorky:
Fixes to formatting, language, hyperlinks.

Some parts I didn't understand, so I couldn't fix:
* what is the difference between the `iter` and `epoch` parameters?
* what does `update_weights()` do?
* what is `delete_temporary_training_data()` for (compared to KeyedVectors)?
* `callbacks: List of callbacks to be executed at specific stages during training.` -- how is it used, what are "specific stages"?


piskvorky:
I pushed my doc fixes for all *2vec models.

The duplication of documentation I saw during the review is formidable. I don't think the current architecture is maintainable. It is also too opaque and convoluted for anyone to contribute meaningfully (including myself, to be honest).

gensim/models/word2vec.py
```diff
@@ -535,8 +771,21 @@ def __init__(self, sentences=None, size=100, alpha=0.025, window=5, min_count=5,
 
     def _do_train_job(self, sentences, alpha, inits):
         """
-        Train a single batch of sentences. Return 2-tuple `(effective word count after
-        ignoring unknown words and sentence length trimming, total word count)`.
+        Train the model on a single batch of sentences. Return`.
```
menshikh-iv: please remove Return`.

gensim/models/word2vec.py
```diff
@@ -558,11 +807,13 @@ def _set_train_params(self, **kwargs):
     def train(self, sentences, total_examples=None, total_words=None,
               epochs=None, start_alpha=None, end_alpha=None, word_count=0,
               queue_factor=2, report_delay=1.0, compute_loss=False, callbacks=()):
-        """Update the model's neural weights from a sequence of sentences (can be a once-only generator stream).
-        For Word2Vec, each sentence must be a list of unicode strings. (Subclasses may accept other examples.)
+        """
+        Update the model's neural weights from a sequence of sentences. `sentences` may be a generator stream.
```
menshikh-iv: please use `"""Update ...`, no need newline (here and everywhere when you have description of method).

docs/src/apiref.rst
```diff
@@ -45,6 +45,7 @@ Modules:
     models/word2vec
     models/keyedvectors
     models/doc2vec
+    models/doc2vec_inner
```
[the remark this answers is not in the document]
piskvorky: This change in was merged from #1944 . That's actually something I wanted to ask -- do we want to be documenting these `_inner` modules?

gensim/models/word2vec.py
```diff
@@ -5,48 +5,76 @@
...
 
-NOTE: It is impossible to continue training the vectors loaded from the C format because hidden weights,
-vocabulary frequency and the binary tree is missing::
+Gensim can also load word vectors in the "word2vec C format", as this `KeyedVectors` instance::
 
     >>> from gensim.models import KeyedVectors
-    >>> word_vectors = KeyedVectors.load_word2vec_format('/tmp/vectors.txt', binary=False)  # C text format
-    >>> word_vectors = KeyedVectors.load_word2vec_format('/tmp/vectors.bin', binary=True)  # C binary format
+    >>> # load from C text format
+    >>> word_vectors = KeyedVectors.load_word2vec_format('/tmp/vectors.txt', binary=False)
```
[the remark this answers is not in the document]
piskvorky: I'd prefer the examples to be simple and clear. Using some internal Gensim constructs could give the reader the wrong idea.

docs/src/apiref.rst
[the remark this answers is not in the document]
menshikh-iv: Yes, we want (this is useful for contributors)

gensim/models/word2vec.py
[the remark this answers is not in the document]
menshikh-iv: I'm +1 for simplicity, but huge -1 for examples that not working:
1. We plan to run all examples in code using `doctests`, this is one more important check (nobody like an example that doesn't work)
2. As a user, I have no idea, what's "C text" format, I want to have a chance to look into.

piskvorky:
@gojomo can you commit a short explanation of these choices into this PR? 1-2 sentences.

I mean a user-facing noob explanation, "why are these parameters like this and what it means for me", rather than deep internal genesis/notes. If I was confused, others must be too.

docs/src/apiref.rst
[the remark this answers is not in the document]
piskvorky: OK. Can I leave this with @steremma ?

gensim/models/word2vec.py
piskvorky: Is this a part of the NumPy standard?

Because if not, I'd prefer to have the first line indented the same way as all other lines. The inconsistency makes the docs harder to read (and the first line is usually the most important one).

gojomo:
@piskvorky Any explanation of `iter` is obsolete unless the #1777 changes are rolled back. It's just `epochs` everywhere now.

 I'm just intuiting what `update_weights()` might mean; I didn't add it and have always had reservations about the way this feature works. I'm not sure it needs any extra doc-comments, as an internal detail of the `update=True` code path that non-experts shouldn't call directly.  (Existing comment of "Copy all the existing weights, and reset the weights for the newly added vocabulary." should be sufficient.)

For `start_alpha`/`end_alpha`, this should be plenty:

        start_alpha : float, optional
            If supplied, this replaces the starting `alpha` from original initialization, for just this one call to `train()`. Used only if a caller is making multiple calls to `train()`, and thus managing the alpha learning-rate themselves (not recommended).
        end_alpha : float, optional
            If supplied, this replaces the final `min_alpha` from original initialization, for just this one call to `train()`. Used only if a caller is making multiple calls to `train()`, and thus managing the alpha learning-rate themselves (not recommended).

(I'm not currently well set up to push to your PR myself.)

piskvorky:
But it's not "`epochs` everywhere now".  I see both `iter` and `epochs` there, which is the source of my confusion. Both seem to have the same semantics, but a different name.

A comment inside `train()` says "explicit `epochs` argument **MUST** be provided", but actually `epochs` has a default value of `None`, and its description is `epochs : int, optional`.

gojomo:
I suppose I was confused by https://github.com/RaRe-Technologies/gensim/blob/916e423a1654d87646fc522d8862a81c7a7cb4fc/gensim/models/base_any2vec.py#L348 - I really don't know the future of iter vs epochs, so this is a question for @manneshiva.

Yes, `epochs` must be provided to any version of `train()` I last touched, later parameter-checking code would raise an error if it isn't, and it wasn't described as 'optional' at that time. I think the same error-raising check still exists in `_check_training_sanity()`. So 'optional' appears to be incorrect.

piskvorky:
@menshikh-iv can you comment on the `epochs` vs `iter`? Optional, not optional? Keep two different parameter names, use a single name?

gensim/models/keyedvectors.py
```diff
@@ -123,30 +159,49 @@ def load(cls, fname_or_handle, **kwargs):
...
 
     def get_vector(self, entity):
-        """Accept a single entity as input, specified by string tag.
-        Returns the entity's representations in vector space, as a 1D numpy array.
+        """
+        Return the entity's representations in vector space, as a 1D numpy array.
+
+        Parameters
+        ----------
+
```
menshikh-iv: remove newline

gensim/models/keyedvectors.py
```diff
@@ -123,30 +159,49 @@ def load(cls, fname_or_handle, **kwargs):
...
+        entity : str
+            Identifier of the entity to return the vector for.
+
+        Returns
+        -------
+        numpy array
```
menshikh-iv: `numpy.ndarray`

gensim/models/keyedvectors.py
```diff
@@ -196,14 +253,16 @@ def add(self, entities, weights, replace=False):
             self.vectors[in_vocab_idxs] = weights[in_vocab_mask]
 
     def __setitem__(self, entities, weights):
-        """Add entities and theirs vectors in a manual way.
+        """
+        Add entities and theirs vectors in a manual way.
+
         If some entity is already in the vocabulary, old vector is replaced with the new one.
-        This method is alias for `add` with `replace=True`.
+        This method is alias for :meth:`~add()` with `replace=True`.
```
menshikh-iv: I'm not sure that current link works correctly (typically need full path) like `~gensim.models.keyedvectors.BaseKeyedVectors.add`

gensim/models/keyedvectors.py
```diff
@@ -296,9 +355,9 @@ def __contains__(self, word):
         return word in self.vocab
 
     def save(self, *args, **kwargs):
-        """Saves the keyedvectors. This saved model can be loaded again using
-        :func:`~gensim.models.*2vec.*2VecKeyedVectors.load` which supports
-        operations on trained word vectors like `most_similar`.
+        """
+        Save the keyed vectors. This saved model can be later loaded again using
+        `load()`.
```
menshikh-iv: missing link to method

gensim/models/keyedvectors.py
```diff
@@ -339,7 +397,7 @@ def get_vector(self, word):
 
     def words_closer_than(self, w1, w2):
         """
-        Returns all words that are closer to `w1` than `w2` is to `w1`.
+        Return all words that are closer to `w1` than `w2` is to `w1`.
```
menshikh-iv: Maybe better to use `Get` instead of `Return` here and everywhere in method description?

gensim/models/keyedvectors.py
```diff
@@ -1304,8 +1369,8 @@ def load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8',
 
         Returns
         -------
-        :obj: `~gensim.models.word2vec.Wod2Vec`
-            Returns the loaded model as an instance of :class: `~gensim.models.word2vec.Wod2Vec`.
+        gensim.models.word2vec.Word2Vec
```
menshikh-iv:
```
:class:`~gensim.models.word2vec.Word2Vec`
```

menshikh-iv:
@piskvorky `iter` deprecated, `epochs` should be used always. We keep both parameters only by backward-compatibility reason.
About "optional": confusing situation - both optional, but one of it should be provided.

piskvorky:
But I see no `epochs` in Word2Vec constructor, for example. How can `iter` be deprecated? I still don't get it.

gensim/models/keyedvectors.py
```diff
@@ -23,23 +24,29 @@
...
-| concurrency | ✅ | ✅ | Thread-safe, allows concurrent vector queries. |
-+-----------|--------------|------------|------------------------------+
-| shared RAM | ✅ | ✅ | Multiple processes can re-use the same data, keeping only a single copy in RAM using `mmap <https://en.wikipedia.org/wiki/Mmap>`_. |
-+-----------|--------------|------------|------------------------------+
-| fast load | ✅ | ✅| Supports `mmap <https://en.wikipedia.org/wiki/Mmap>`_ to load data from disk instantaneously. |
-+-----------|--------------|------------|------------------------------+
++---------------------------+--------------+------------+-------------------------------------------------------------+
+|        capability         | KeyedVectors | full model |                               note                          |
++---------------------------+--------------+------------+-------------------------------------------------------------+
+| continue training vectors | no           | yes        | You need the full model to train or update vectors.         |
```
piskvorky: Why not ✅ and ❌?

menshikh-iv: @piskvorky because the table breaks

gensim/models/fasttext_inner.pyx
```diff
@@ -2,8 +2,9 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: cdivision=True
+# cython: embedsignature=True
 # coding: utf-8
-
+"""Optimized cython functions for training :class:`~gensim.models.fasttext.FastText` model."""
```
piskvorky: Missing blank line after docstring (here and elsewhere).

gensim/models/doc2vec_inner.pyx
```diff
@@ -2,33 +2,27 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: cdivision=True
+# cython: embedsignature=True
 # coding: utf-8
 #
 # Copyright (C) 2013 Radim Rehurek <me@radimrehurek.com>
 # Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
-
+"""Optimized cython functions for training :class:`~gensim.models.doc2vec.Doc2Vec` model."""
```
piskvorky: Missing blank line after docstring (here and elsewhere).

gensim/models/utils_any2vec.py
```diff
@@ -3,7 +3,6 @@
 #
 # Author: Shiva Manne <s.manne@rare-technologies.com>
 # Copyright (C) 2018 RaRe Technologies s.r.o.
-
```
piskvorky: -1: why this change? There definitely should be a blank line.

gensim/models/doc2vec_inner.pyx
menshikh-iv: No needed, see https://github.com/numpy/numpy/blob/master/numpy/ma/mrecords.py#L10

gensim/models/utils_any2vec.py
menshikh-iv: no, see http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

gensim/models/doc2vec_inner.pyx
piskvorky: Strong -1: goes against standard Python docstring logic (blank line to delimit unrelated blocks). Also, ugly / harder to read for no good reason (or is there a reason?).

menshikh-iv: > goes against standard Python docstring logic

any PEP for it? Also see https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

PR #2089, opened 2018-06-13
[MRG] Wikicorpus custom filters

mattilyra:
I wanted to add a simple method for filtering out wiki articles based on arbitrary rules. The idea is to be able to easily produce datasets like the [WikiText-103](https://einstein.ai/research/the-wikitext-long-term-dependency-language-modeling-dataset).

I am currently working on making a German equivalent of that corpus (call it `DeWiki103` for now), but that requires filtering out articles based on regular expressions applied to the text of each XML element (short of parsing the Wikimedia format).

This PR adds a simple `filter_articles` method to `WikiCorpus` that gets called in `extract_pages`. Returning `None` will filter out that specific article.

Let me know if you think this might be a useful addition.


gensim/corpora/wikicorpus.py
```diff
@@ -380,6 +380,13 @@ def extract_pages(f, filter_namespaces=False):
                 if ns not in filter_namespaces:
                     text = None
 
+            if callable(filter_articles) and text:
```
piskvorky: What if `filter_articles` isn't a callable? If I understand correctly, it should be a callable always, no need to check.

piskvorky:
Definitely useful! And I like the solution (callable injection): simple and flexible 👍

gensim/corpora/wikicorpus.py
```diff
@@ -544,10 +551,15 @@ def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), diction
             Maximal token length.
         lower : bool, optional
              If True - convert all text to lower case.
+        filter_articles: callable, optional
+            If set each XML article element will be passed to this callable before being processed. Only articles
+            where the callable returns an XML element are processed, returning None allows filtering out
+            some articles based on customised rules.
```
piskvorky: Are there any restrictions on the callable? Can it be a lambda function? (troublesome with parallelized multiprocessing)

gensim/corpora/wikicorpus.py
mattilyra: That's just catching the default case where `filter_articles=None` - I didn't see the point of introducing a dummy callable that always returns true. I guess you could make the default `filter_articles=False` but that confuses the type semantics of `filter_articles` is it `bool` or a `callable`

gensim/corpora/wikicorpus.py
mattilyra: True, for the parallel `self.process` to work `filter_articles` needs to be serializable, what's the practice for checking this in `gensim`? Try to `pickle` the callable and raise an informative exception if that fails? Is `pickle` the standard serialiser in `gensim`?

On a related "fun" note, see https://stackoverflow.com/questions/50328386/python-typing-pickle-and-serialisation

piskvorky: I was thinking a simple note in the docs. What does the error look like now, if a user passes something not pickleable? If it looks reasonable, I'd just keep that.

@menshikh-iv what is our policy here in general?

mattilyra: Well there's no specific catch for this so it'll just be python's `PicklingError` - that plus a note in the docs I think should be enough, but I don't know how this is handled elsewhere in `gensim`

piskvorky: I think so too. Someone advanced enough to be setting this parameter better be advanced enough to RTFM :)

gensim/corpora/wikicorpus.py
piskvorky: `is not None` preferred (since that's what we really want to test). But a dummy `filter_articles=gensim.utils.identity` is also OK.

piskvorky:
LGTM for me, thanks Matti! Let's wait for @menshikh-iv 's final verdict & merge.

gensim/corpora/wikicorpus.py
```diff
@@ -544,10 +555,15 @@ def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), diction
             Maximal token length.
         lower : bool, optional
              If True - convert all text to lower case.
+        filter_articles: callable or None, optional
```
menshikh-iv: `callable, optional`

gensim/corpora/wikicorpus.py
```diff
@@ -544,10 +555,15 @@ def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), diction
...
+            If set, each XML article element will be passed to this callable before being processed. Only articles
+            where the callable returns an XML element are processed, returning None allows filtering out
+            some articles based on customised rules.
```
menshikh-iv: Can you create an example of this function and link it in docstring like
```
:func:`gensim.corpora.wikicorpus.filter_example`
```
this can be useful for better understanding, how function works
- input parameters
- return values (probably True/False better that XML element/None)

mattilyra: you mean add an example `filter_example` to the `wikicorpus` module?

mattilyra: got it, very busy at work but will try to get it done this week

menshikh-iv: don't forget to resolve merge conflict too

mattilyra:
@menshikh-iv `ERROR: InvocationError for command '/home/travis/build/RaRe-Technologies/gensim/.tox/py36-linux/bin/pytest gensim/test' (exited with code 1)` ????

gensim/corpora/wikicorpus.py
```diff
@@ -365,6 +426,13 @@ def extract_pages(f, filter_namespaces=False):
                 if ns not in filter_namespaces:
                     text = None
 
+            if filter_articles is not None:
+                if not filter_articles(elem, namespace=namespace, title=title,
+                                       text=text, page_tag=page_tag,
```
piskvorky: We use hanging indent in Gensim (not vertical alignment): https://www.python.org/dev/peps/pep-0008/#id3

Why? Makes code easier to maintain (people forget to re-indent after modifying), and some people (me!) also find it more consistent and easier to read :)

mattilyra:
Have to respect properly enforced coding style 😄 👍 

is there a `gensim` code style guide somewhere, I didn't see anything in the contribution guidelines and the linter does not complain about that.

piskvorky:
Both vertical and hanging indents are OK by Python standards (PEP8), so linters wouldn't complain. Consistent hanging indent is just our project choice.

We keep developer instructions here on Gtihub: https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#code-style

piskvorky:
Awesome job, thanks @mattilyra !

PR #2095, opened 2018-06-19
fix git command in readme

aneesh-joshi:
`git add ...` -> `git add .`


piskvorky:
The "…" was meant as an ellipsis, not a copy&paste command :)

I don't think adding the current directory (`.`) is something we want to prescribe. That could lead to many errors, adding unwanted untracked files etc.

aneesh-joshi:
Haha.
Now I feel extremely silly for suggesting that!
I use [perforce VCS](https://www.perforce.com/) where the typical command for adding is:
`p4 add ...` which made me think the author made a mistake. I didn't realize `...` has a different meaning in python

PR #2096, opened 2018-06-20
Fix documentation for various modules

piskvorky:
Reviewing the remaining high-profile modules (utils, corpora, interfaces).


gensim/corpora/dictionary.py
```diff
@@ -463,7 +458,7 @@ def save_as_text(self, fname, sort_by_word=True):
         fname : str
             Path to output file.
         sort_by_word : bool, optional
-            Sort words in lexicographical order before writing them out?
+            Sort words in lexicographical order before writing them out.
```
piskvorky: -1: bool should be a question, to make it clear what positive answer means (True).

gensim/corpora/wikicorpus.py
```diff
@@ -526,17 +510,15 @@ def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), diction
...
         lemmatize : bool
-            Use lemmatization instead of simple regexp tokenization?
-
-            Defaults to `True` if you have the `pattern package <https://github.com/clips/pattern>`_ installed.
+            Use lemmatization instead of simple regexp tokenization.
+            Defaults to `True` if you have the `pattern <https://github.com/clips/pattern>`_ package installed.
         dictionary : :class:`~gensim.corpora.dictionary.Dictionary`, optional
             Dictionary, if not provided,  this scans the corpus once, to determine its vocabulary
-            (this needs **really long time**).
+            (**IMPORTANT: this needs really long time**).
```
piskvorky: If it's important, it should be in brackets :)

`really long time` => `a really long time`

gensim/corpora/wikicorpus.py
```diff
@@ -546,7 +528,11 @@ def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), diction
         token_max_len : int, optional
             Maximal token length.
         lower : bool, optional
-             Convert all text to lower case?
+             Convert all text to lower case.
```
piskvorky: -1: should be a question (here and elsewhere).

gensim/utils.py
```diff
@@ -1260,12 +1259,12 @@ def chunkize(corpus, chunksize, maxsize=0, as_numpy=False):
...
             If > 0, prepare chunks in a background process, filling a chunk queue of size at most `maxsize`.
         as_numpy : bool, optional
-            Yield chunks as `np.ndarray`s instead of lists?
+            Yield chunks as `np.ndarray` instead of lists?
 
         Yields
         ------
         list OR np.ndarray
-            `chunksize`-ed chunks of elements from `corpus`.
+            "chunksize"-ed chunks of elements from `corpus`.
```
piskvorky: Why this change? `chunksize` is a parameter name (just like ``corpus``, not `"corpus"`).

menshikh-iv: in Sphinx construction
```
`something`other
```
without space between "`" and "o" break render

gensim/corpora/dictionary.py
menshikh-iv: Fixed

gensim/corpora/wikicorpus.py
menshikh-iv: fixed

gensim/corpora/wikicorpus.py
menshikh-iv: fixed

PR #2098, opened 2018-06-20
Fix citation of Mikolov paper

davidchall:
The default scorer used by the phraser is Equation 6 from https://arxiv.org/pdf/1310.4546.pdf

I've updated citations in the gensim documentation to reflect this. I also fixed a typo in the docstring, which led to an incorrect formula.


menshikh-iv:
Thanks for contribution @davidchall, this will be fixed as part of #2096.

PR #2103, opened 2018-06-25
[MRG] Doc2Vec inference, notebook cleanup

gojomo:
Inference improvements: default parameters to analogous model parameters

Notebook updates: test operation, correct errors, provide updated executions

PR #2116, opened 2018-07-03
Fix the enumeration over 'alldata'

soumenganguly:
Without the splitlines(), the enumerate goes over each word rather than each line/sentence. As a result, the split list goes out of range as it encounters values 0,1,2,3, and 4 at which point it throws the list out of bounds error as the number of elements in the list is 4 and the enumerate function gives it values more than 4.


menshikh-iv:
thanks @soumenganguly, but tutorial already fixed in #2103

Issue #1947, opened 2018-03-02
AuthorTopicModel memory issue


menshikh-iv:
### Intro

Recently, I often get negative feedback about ATM.
Еhe main reason is memory issues (too much memory consuming), related mailing list threads (latest):
- https://groups.google.com/forum/#!searchin/gensim/author|sort:date/gensim/gG7aiNI1v-Y/SWPMuP8BAwAJ
- https://groups.google.com/forum/#!topic/gensim/wI3x1fqR5wk

I decided to figure out what was going on.

### Investigation

I run ATM based on data provided by the author of https://groups.google.com/forum/#!searchin/gensim/author|sort:date/gensim/gG7aiNI1v-Y/SWPMuP8BAwAJ (so far I can't publish it right now, I'm waiting for permission from its owner).

Basic stats of data:
- Size of dictionary: 12211
- Size of `author2doc` mapping: 106133
- Size of `author2doc` mapping: 73248

I run it with a debugger and found that **hugest** memory-consuming happens here:
https://github.com/RaRe-Technologies/gensim/blob/f9669bb8a0b5b4b45fa8ff58d951a11d3178116d/gensim/models/atmodel.py#L680-L684

I stop it when process already consume **8GB** of RAM, interesting statistics

| expr | value | comment |
|--------|---------|-------------|
|`len(author2doc.keys())`|106133 | |
|`author2doc.keys().index(_)`|3649| index of current processed element, when process already, i.e. 3649 of 106133 (~3%)  |
|`len(train_corpus_idx) ` | 1119955735 | this is hugest memory consumer (also this isn't only/batch approach, before training, this will be full in RAM |

By simple calculations, it turns out that by the end of the construction of `train_corpus_idx`, the process will consume `~232GB of RAM`, this is definitely unacceptable and doesn't allow to use model even for learning tasks (I'm not even talking about "real" tasks).

@olavurmortensen can you look into this problem, this is **super critical**?

Related PR - #893.

menshikh-iv:
Updates:
1. Unfortunately, Olavur have no time to resolve this issue
2. More information (about ATM) can be founded here: http://www2.imm.dtu.dk/pubdb/views/edoc_download.php/6971/pdf/imm6971.pdf

PR #2122, opened 2018-07-07
Fix memory consumption in AuthorTopicModel

philipphager:
@menshikh-iv 
I submitted this PR to address the memory consumption issue faced when using the `AuthorTopicModel` as also recognized in #1947. We had to fix this issue for a research project and maybe you are interested to use this simple fix in the main project. The concatenation of the entire corpus for all authors and then removing unique values (resulting in all unique values in the corpus), can be reduced to this, and the results did not change.

This fix reduced the memory consumption on our project with ≈400.000 docs from 32GB to 2GB for the entire duration of the training.

### DIFF
--- a/gensim/models/atmodel.py
+++ b/gensim/models/atmodel.py
@@ -767,9 +767,9 @@ class AuthorTopicModel(LdaModel):
 
             # Train on all documents of authors in input_corpus.
             train_corpus_idx = []
-            for _ in author2doc.keys():  # For all authors in input corpus.
-                for doc_ids in self.author2doc.values():  # For all documents in total corpus.
-                    train_corpus_idx.extend(doc_ids)
+
+            for doc_ids in self.author2doc.values():  # For all documents in total corpus.
+                train_corpus_idx.extend(doc_ids)
 
             # Make the list of training documents unique.
             train_corpus_idx = list(set(train_corpus_idx))
--- a/gensim/models/atmodel.py
+++ b/gensim/models/atmodel.py
@@ -767,8 +767,8 @@ class AuthorTopicModel(LdaModel):
 
             # Train on all documents of authors in input_corpus.
             train_corpus_idx = []
-
-            for doc_ids in self.author2doc.values():  # For all documents in total corpus.
+            # Collect all documents of authors.
+            for doc_ids in self.author2doc.values():
                 train_corpus_idx.extend(doc_ids)
 
             # Make the list of training documents unique.
--- a/gensim/models/atmodel.py
+++ b/gensim/models/atmodel.py
@@ -766,13 +766,13 @@ class AuthorTopicModel(LdaModel):
                 self.doc2author[d] = a_list
 
             # Train on all documents of authors in input_corpus.
-            train_corpus_idx = []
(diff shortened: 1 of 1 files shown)


gensim/models/atmodel.py
```diff
@@ -767,9 +767,9 @@ def update(self, corpus=None, author2doc=None, doc2author=None, chunksize=None,
 
             # Train on all documents of authors in input_corpus.
             train_corpus_idx = []
-            for _ in author2doc.keys():  # For all authors in input corpus.
-                for doc_ids in self.author2doc.values():  # For all documents in total corpus.
-                    train_corpus_idx.extend(doc_ids)
+            # Collect all documents of authors.
+            for doc_ids in self.author2doc.values():
+                train_corpus_idx.extend(doc_ids)
```
piskvorky: Why keep `train_corpus_idx` as `list` in the first place? It looks like it's converted to a set right below, so maybe better to make it a `set` from the start?

piskvorky:
Nice! The original code looks strange indeed -- @olavurmortensen any particular reason for that nested quadratic loop?

gensim/models/atmodel.py
philipphager: That is true, indeed. I would propose something like this:
```
# Train on all documents of authors in input_corpus.
train_corpus_idx = set()
# Collect all documents of authors.
for doc_ids in self.author2doc.values():
    train_corpus_idx.update(doc_ids)

train_corpus_idx = list(train_corpus_idx)
```

piskvorky: I'm not familiar with the algo, but for the sake of reproducibility, I'd replace `list(train_corpus_idx)` with `sorted(train_corpus_idx)`. That will remove randomness from the ordering of values, while still producing a list.

philipphager:
I updated the loop to use a set and to sort the resulting list. At least on my current dataset this is also a little bit faster :), and reduces the temporary memory overhead even further.

olavurmortensen:
I can't imagine there was a reason for that nested loop, must have just slipped my mind. I only tested scalability w.r.t. running time, if I'd tested memory consumption as well I should have caught this.

There is some stuff in my thesis about asymptotic complexity of memory consumption ([pdf](http://www2.imm.dtu.dk/pubdb/views/edoc_download.php/6971/pdf/imm6971.pdf), section 2.4.2.6). The algorithm doesn't scale terribly well w.r.t. memory consumption. The empirical results showed that running time scaled as expected, compared to the theoretical scalability, but as I said I didn't test memory consumption.

I'm glad this problem was caught, hopefully it fixes the issues people are having. Thanks @philipphager.

philipphager:
No problem! And also, awesome job done on the implementation of the AuthorTopicModel @olavurmortensen, the API is an absolute pleasure to work with 👍 ! And thank you both for being so responsive on this PR :)!

piskvorky:
Thanks @philipphager . What happens next: we'll wait for @menshikh-iv to come back from holiday, so he can review and merge this 👍

piskvorky:
Actually, let me merge right away. The fix is simple enough that hopefully @menshikh-iv won't be angry :)

Issue #1958, opened 2018-03-07
gensim.scripts.word2vec2tensor TypeError: write() argument must be str, not bytes 


ttpro1995:
**How I make article_body_w2v_300.txt**  

```
import gensim
from gensim.models import Word2Vec
from gensim.models.word2vec import LineSentence

sentences = LineSentence("./data/article_body_corpus.txt")

model = Word2Vec(sentences, size=300, window=5, min_count=5, workers=4)

model.wv.save_word2vec_format("article_body_w2v_300.txt", binary=False)
```

**Command I use to run gensim.scripts.word2vec2tensor ** 
python -m gensim.scripts.word2vec2tensor -i article_body_w2v_300.txt -o meow/
2018-03-07 16:30:29,484 - word2vec2tensor - INFO - running 

Console output
```
word_embedding  python -m gensim.scripts.word2vec2tensor -i article_body_w2v_300.txt -o meow/
2018-03-07 16:30:29,484 - word2vec2tensor - INFO - running /home/cpu11453local/anaconda3/envs/gensim/lib/python3.6/site-packages/gensim/scripts/word2vec2tensor.py -i article_body_w2v_300.txt -o meow/
2018-03-07 16:30:29,484 - utils_any2vec - INFO - loading projection weights from article_body_w2v_300.txt
2018-03-07 16:30:41,992 - utils_any2vec - INFO - loaded (56543, 300) matrix from article_body_w2v_300.txt
Traceback (most recent call last):
  File "/home/cpu11453local/anaconda3/envs/gensim/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/cpu11453local/anaconda3/envs/gensim/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/cpu11453local/anaconda3/envs/gensim/lib/python3.6/site-packages/gensim/scripts/word2vec2tensor.py", line 93, in <module>
    word2vec2tensor(args.input, args.output, args.binary)
  File "/home/cpu11453local/anaconda3/envs/gensim/lib/python3.6/site-packages/gensim/scripts/word2vec2tensor.py", line 73, in word2vec2tensor
    file_metadata.write(gensim.utils.to_utf8(word) + gensim.utils.to_utf8('\n'))
TypeError: write() argument must be str, not bytes

```

menshikh-iv:
[...]

PR #2125, opened 2018-07-09
Dealing with typeError: must be str, not bytes

FMCumhaill:
Dealing with typeError: must be str, not bytes which occurred for me when running this file. https://stackoverflow.com/questions/5512811/builtins-typeerror-must-be-str-not-bytes

This may be a local requirement and therefore not sufficient for a pull request?

### DIFF
--- a/gensim/scripts/word2vec2tensor.py
+++ b/gensim/scripts/word2vec2tensor.py
@@ -67,8 +67,8 @@ def word2vec2tensor(word2vec_model_path, tensor_filename, binary=False):
     outfiletsv = tensor_filename + '_tensor.tsv'
     outfiletsvmeta = tensor_filename + '_metadata.tsv'
 
-    with open(outfiletsv, 'w+') as file_vector:
-        with open(outfiletsvmeta, 'w+') as file_metadata:
+    with open(outfiletsv, 'w') as file_vector:
+        with open(outfiletsvmeta, 'wb') as file_metadata:
             for word in model.index2word:
                 file_metadata.write(gensim.utils.to_utf8(word) + gensim.utils.to_utf8('\n'))
                 vector_row = '\t'.join(str(x) for x in model[word])
--- a/gensim/scripts/word2vec2tensor.py
+++ b/gensim/scripts/word2vec2tensor.py
@@ -44,6 +44,7 @@ import logging
 import argparse
 
 import gensim
+from smart_open import smart_open
 
 logger = logging.getLogger(__name__)
 
@@ -67,8 +68,8 @@ def word2vec2tensor(word2vec_model_path, tensor_filename, binary=False):
     outfiletsv = tensor_filename + '_tensor.tsv'
     outfiletsvmeta = tensor_filename + '_metadata.tsv'
 
-    with open(outfiletsv, 'w') as file_vector:
-        with open(outfiletsvmeta, 'wb') as file_metadata:
+    with smart_open(outfiletsv, 'w') as file_vector:
+        with smart_open(outfiletsvmeta, 'wb') as file_metadata:
             for word in model.index2word:
                 file_metadata.write(gensim.utils.to_utf8(word) + gensim.utils.to_utf8('\n'))
                 vector_row = '\t'.join(str(x) for x in model[word])


piskvorky:
All I/O writing or reading should definitely be happening in binary mode (`rb`, `wb`). With explicit encoding conversions where necessary. This is true for both Python 2 and Python 3.

@loretoparisi was there a particular reason for the `w+`? Can we replace it with `wb`?

One more request: @FMCumhaill can you please replace all calls to `open()` with `smart_open()` (`from smart_open import smart_open`)?

The reason is that in Gensim, we're trying to abstract away from the storage, so people can save files to S3, HDFS, compressed etc, using the [smart_open library](https://github.com/RaRe-Technologies/smart_open).

FMCumhaill:
@piskvorky done as requested! 👍

gensim/scripts/word2vec2tensor.py
```diff
@@ -67,8 +68,8 @@ def word2vec2tensor(word2vec_model_path, tensor_filename, binary=False):
     outfiletsv = tensor_filename + '_tensor.tsv'
     outfiletsvmeta = tensor_filename + '_metadata.tsv'
 
-    with open(outfiletsv, 'w') as file_vector:
-        with open(outfiletsvmeta, 'wb') as file_metadata:
+    with smart_open(outfiletsv, 'w') as file_vector:
```
piskvorky: No `w` -- always use binary I/O (`rb`, `wb`). Encode text to utf8 where necessary.

FMCumhaill: Changing that w to wb gives me the following error: 

  File "word2vec2tensor.py", line 94, in <module>
    word2vec2tensor(args.input, args.output, args.binary)
  File "word2vec2tensor.py", line 76, in word2vec2tensor
    file_vector.write(vector_row + '\n')
TypeError: a bytes-like object is required, not 'str'

piskvorky: Yes -- please encode text to binary (utf8) where necessary.

gensim/scripts/word2vec2tensor.py
```diff
@@ -44,6 +44,7 @@
 import argparse
 
 import gensim
+from smart_open import smart_open
```
menshikh-iv: please add a test for `word2vec2tensor` function (for avoid this issue in future)

gensim/scripts/word2vec2tensor.py
```diff
@@ -67,8 +68,8 @@ def word2vec2tensor(word2vec_model_path, tensor_filename, binary=False):
     outfiletsv = tensor_filename + '_tensor.tsv'
     outfiletsvmeta = tensor_filename + '_metadata.tsv'
 
-    with open(outfiletsv, 'w+') as file_vector:
-        with open(outfiletsvmeta, 'w+') as file_metadata:
+    with smart_open(outfiletsv, 'w') as file_vector:
```
menshikh-iv: must be `wb`

menshikh-iv: also, good idea to make a  code more "flat", like this
```python
with smart_open(...) as file_vector, smart_open(...) as file_metadata:
    ...
```

menshikh-iv:
@FMCumhaill thanks for PR, do you plan to finish it?

gensim/scripts/word2vec2tensor.py
FMCumhaill: @menshikh-iv  sorry this escaped me! ATM I don't have time to finish it sorry. I have no idea how to add a test and don't have time to research it. My apologies!

menshikh-iv: @FMCumhaill That's sad, but anyway, thanks for the answer, good luck!

menshikh-iv:
Close by https://github.com/RaRe-Technologies/gensim/pull/2125#discussion_r206426486 reason

piskvorky:
@menshikh-iv this is a bug, please don't close like this. We need to fix this either way (and the fix looks really trivial).

@loretoparisi was there a particular reason for the `w+`? Can we replace it with `wb`?

menshikh-iv:
@piskvorky we need fix bugs, I agree, but this is PR, not a issue, if contributor don't have time to complete (and isn't close to finish) - I close it without "almost complete" badge.

Also, we already have opened issue for this bug #1958 

About "amost_complete" - I don't agree, because:
- incorrect file modes
- missing tests
- 3 lines diff where correct only one (import), I don't think that we should mark it with badge

piskvorky:
OK. Let me remove the label.

PR #2128, opened 2018-07-12
Make clear prune_at documentation

yxonic:
According to the code, the `prune_at` parameter in `Dictionary.__init__` and `add_documents` is only for reducing memory usage, and has no guarantee on correctness, but the documentation of this parameter was confusing to users.


menshikh-iv:
Good catch, thanks @yxonic :+1: Congratz with your first PR :1st_place_medal:

PR #2132, opened 2018-07-18
Fix 1589

probinso:
Empty numpy arrays default to `dtype=np.float`, rather than infering `dtype=np.integer` or `dtype=np.bool`. This means that empty arrays with out `dtype` casting cannot be used as index arrays.


probinso:
I performed a poor pull request, I am resubmitting.

PR #2133, opened 2018-07-18
Correctly process empty documents in `AuthorTopicModel`

### LINKED ISSUE #1589: Problem using bound function in Author Topic model!!
Hi
I am trying to use author topic model. When I used bound function to evaluate my model I received this error in python:

  File "C:\Users\Sara\Desktop\E-COM\Final Project\Working on data\at_clustering.py", line 161, in <module>
    test_corpus = test_corpus, test_d2a = test_doc2author, test_a2d = test_author2doc, limit = 25)
  File "C:\Users\Sara\Desktop\E-COM\Final Project\Working on data\at_clustering.py", line 70, in evaluate_k
    pr = np.exp2(-model.bound(test_corpus, doc2author=test_d2a, author2doc=test_a2d)/number_of_words)
  File "C:\Python27\lib\site-packages\gensim\models\atmodel.py", line 835, in bound
    phinorm = self.compute_phinorm(ids, authors_d, expElogtheta[authors_d, :], expElogbeta[:, ids])
**IndexError: arrays used as indices must be of integer (or boolean) type**

 I think the problem is **expElogbeta[:, ids]** because this is not acceptable in python!

Please test bound function and fix source code.
Thanks

probinso:
This is a fix #1589 

initialized empty numpy arrays defualt to `dtype=np.float` making them ineligible for use as index arrays (which must be of `dtype=np.integer` or `dtype=np.bool`)

### DIFF
--- a/gensim/test/test_atmodel.py
+++ b/gensim/test/test_atmodel.py
@@ -117,8 +117,7 @@ class TestAuthorTopicModel(unittest.TestCase, basetmtests.TestBaseTopicModel):
         a2d = author2doc.copy()
         a2d['joaquin'] = [len(local_texts) - 1]
 
-        _ = self.class_(corpus, author2doc=a2d, id2word=dictionary, num_topics=2)
-        assert(_)
+        self.class_(corpus, author2doc=a2d, id2word=dictionary, num_topics=2)
 
     def testAuthor2docMissing(self):
         # Check that the results are the same if author2doc is constructed automatically from doc2author.


probinso:
@piskvorky is there anything else I need to do for this pull request?

gensim/models/atmodel.py
```diff
@@ -460,11 +460,12 @@ def inference(self, chunk, author2doc, doc2author, rhot, collect_sstats=False, c
                 # make sure the term IDs are ints, otherwise np will get upset
                 ids = [int(idx) for idx, _ in doc]
             else:
-                ids = [idx for idx, _ in doc]
-            cts = np.array([cnt for _, cnt in doc])
+                ids = [id for id, _ in doc]
+            ids = np.array(ids, dtype=np.integer)
+            cts = np.array([cnt for _, cnt in doc], dtype=np.integer)
```
piskvorky: I'm not familiar with this `np.integer` type. How does it differ from normal `np.int`? What's the difference?

piskvorky:
It looks good, thanks @probinso . Just a little clarification around `np.integer` / `np.int` for my sake please.

probinso:
@piskvorky 

That is a good question. I'll read through the `numpy` code. I used what I expected to be the most general correct type. However I can tell that they are different because `np.int is np.integer == False`.

gensim/models/atmodel.py
menshikh-iv: No difference in our case
```python
import numpy as np

arr1, arr2 = [1, 2, 3], []

assert np.array(arr1, dtype=np.int).dtype == \
       np.array(arr1, dtype=np.integer).dtype == \
       np.array(arr2, dtype=np.int).dtype == \
       np.array(arr2, dtype=np.integer).dtype
```

gensim/models/atmodel.py
```diff
@@ -460,11 +460,12 @@ def inference(self, chunk, author2doc, doc2author, rhot, collect_sstats=False, c
                 # make sure the term IDs are ints, otherwise np will get upset
                 ids = [int(idx) for idx, _ in doc]
             else:
-                ids = [idx for idx, _ in doc]
-            cts = np.array([cnt for _, cnt in doc])
+                ids = [id for id, _ in doc]
```
menshikh-iv: Please revert back `idx` (`id` is built-in function name)

gensim/test/test_atmodel.py
```diff
@@ -110,6 +109,19 @@ def testBasic(self):
         jill_topics = matutils.sparse2full(jill_topics, model.num_topics)
         self.assertTrue(all(jill_topics > 0))
 
+    def testEmptyDocument(self):
+        _local_texts = common_texts + [['only_occurs_once_in_corpus_and_alone_in_doc']]
```
menshikh-iv: why vars starts from underscore? please remove underscores from start

gensim/test/test_atmodel.py
```diff
@@ -110,6 +109,19 @@ def testBasic(self):
...
+        _dictionary = Dictionary(_local_texts)
+        _dictionary.filter_extremes(no_below=2)
+        _corpus = [_dictionary.doc2bow(text) for text in _local_texts]
+        _a2d = author2doc.copy()
+        _a2d['joaquin'] = [len(_local_texts) - 1]
+        try:
```
menshikh-iv: No need try/except section in test, if test raise unexpected exception - this means that test broken

gensim/test/test_atmodel.py
```diff
@@ -110,6 +109,19 @@ def testBasic(self):
...
+            _ = self.class_(_corpus, author2doc=_a2d, id2word=_dictionary, num_topics=2)
+        except IndexError:
+            raise IndexError("error occurs in 1.0.0 release tag")
```
menshikh-iv: ???

gensim/models/atmodel.py
menshikh-iv: all of it "casted" to `int64` on my x64 linux

piskvorky:
I'm still -1 on using `np.integer` -- what is that and why should we use it, instead of the standard `int`?

menshikh-iv:
@piskvorky fixed in #2145

PR #2127, opened 2018-07-11
File-based fast training for Any2Vec models

persiyanov:
_note: all preliminary discussions are here https://github.com/RaRe-Technologies/gensim/pull/2048_

This PR summarizes all my work during GSoC 2018. For more understanding what's going on, follow the links:
* My proposal: https://persiyanov.github.io/jekyll/update/2018/04/24/accepted-to-gsoc-2018.html
* First benchmarks: https://persiyanov.github.io/jekyll/update/2018/05/28/gsoc-first-weeks.html
* Last blog post about the almost final solution: https://persiyanov.github.io/2018/07/06/gsoc-midreport.html
* Links to all benchmarks: https://gist.github.com/persiyanov/84b806233947e0069a243433579b35db
* Previous PR about vocab building : https://github.com/RaRe-Technologies/gensim/pull/2078 (reverted these changes in current PR because of API design issues)
* Previous PR about multistream training (all useful changes in this PR): https://github.com/RaRe-Technologies/gensim/pull/2048

# Summary

In this pull request, new argument `corpus_file` is proposed for `Word2Vec`, `FastText` and `Doc2Vec` models. It is supposed to use `corpus_file` instead of standard `sentences` argument if you have the preprocessed dataset on disk and want to get **significant speedup during model training**.

On our benchmarks, training `Word2Vec` on English Wikipedia dump is **370% faster** with `corpus_file` than training with `sentences` (see the attached jupyter notebook with the code).

Look at this chart for Word2Vec:

# Usage

The usage is really simple. I'll provide examples for `Word2Vec` while the usage for `FastText` and `Doc2Vec` is identical. The `corpus_file` argument is supported for:

### Constructor 

```python3
# Standard way
model = Word2Vec(sentences=my_corpus, <...other arguments...>)

# New way
model = Word2Vec(corpus_file='my_corpus_saved.txt', <...other arguments...>)

# You can save your own corpus using
gensim.utils.save_as_line_sentence(my_corpus, 'my_corpus_saved.txt')

```

### build_vocab

```python3
# Create the model without training
model = Word2Vec(<...other arguments...>)

# Standard way
model.build_vocab(sentences=my_corpus, ...)

# New way
model.build_vocab(corpus_file='my_corpus_saved.txt', ...)
```

### train

```python3
# Create the model without training
model = Word2Vec(<...other arguments...>)

# Build vocab (with `sentences` or `corpus_file` way, choose what you like)
model.build_vocab(corpus_file='my_corpus_saved.txt')

[...]

PR #2135, opened 2018-07-19
Fixing the computation of Word2Vec losses.

alreadytaikeune:
The current computation of word2vec losses has flaws. I address them in this PR.

This PR re-writes the computation of the loss for both CBOW and SG Word2Vec. The loss that is computed and reported is the running average NCE loss within the epoch. This means that for each new epoch, the counters are reset to 0, and the new average is computed. This was not the case before, and the loss was incremented during the whole training (but the total_examples used to average was only incremented during on epoch), which is not very informative, beside being also incorrect in the implementation. Moreover, reporting the average loss on an epoch appeared to me more in the spirit of what was trying to be achieved before.

The computation of the word2vec loss was flawed in many ways:
- race condition on the running_training_loss parameter (updated concurrently in a
  GIL-free portion of the code)
- As mentioned above, there was an incoherence between the accumulation of the loss on the whole run, and the reset of the averaging factor at each epoch.
- Even if the points above were fixed, remains the following: the dividing factor for the average in the case of SG is wrong. The averaging factor in the case of SG should not be the effective words, but the effective samples (a new variable I introduce), because the loss is incremented as many times as there are positive examples that are sampled for an effective word.

Additionnally, I add the logging of the current value of the loss in the progress logger, when compute_loss is set to True, and I add a parameter to the word2vec_standalone script to trigger the reporting of the loss.

As a hint towards the fact that the current implementation is now correct, one can look at the first reported values of the loss, when the word embeddings are still relatively uniformly distributed. In this situation, the expectancy of the NCE loss (for Skip-Gram) should be -(N+1)\log(\sigma(0)). Which is 5.545 for N=7, 14.556 for N=20... which corresponds to the reported loss values in the current implementation.


piskvorky:
Thanks @alreadytaikeune . The training loss is a recent addition and may contain errors. What is your use-case for using it? How did you find about this issue?

What are the performance implications of your changes? Can you post a before/after benchmark?

alreadytaikeune:
Thanks for your reply. I started digging into it because I needed to compare the evolution and values of the loss between two word2vec implementations. Therefore I needed to understand exactly how the loss was computed in gensim which lead me to uncover some issues and proposing these changes. Are the issues I pointed out clear to you? I don't know if I expressed them well enough.

As to the benchmark, I haven't really done any since the changes are very minor and should really not impact performances, but I'll do one if needed. Do you have some standard tools for this purpose that I can use, or do I need to write mine?

Finally, it seems the documentation has a hard time building. From what I understand, it is the lines
```
-loss 
                If present, the loss will be computed and printed during training
```
in the `word2vec_standalone.py` that are the cause. The problem is that there isn't any type after `loss`, because it is only a flag. I am not sure what the sphinx syntax is for this kind of stuff. Any idea?

piskvorky:
Thanks! Can you run the benchmark too? Simply train a model on the [text9 corpus](https://markroxor.github.io/gensim/static/notebooks/Word2Vec_FastText_Comparison.html) with/without this PR, using 4 parallel workers, post the two logs to [gist](https://gist.github.com/) and add the links here in a comment.

There'll be some minor variance in the timings, even with same parameters, that's expected. But we're interested in a sanity check here (<<10% difference).

alreadytaikeune:
@piskvorky 

I have run the tests you said in the following environment
```
FROM python:2
RUN mkdir -p /home/gensim_user/gensim
WORKDIR /home/gensim_user
RUN apt-get update && apt-get install -y wget build-essential 
RUN pip install numpy scipy six smart_open
```

Using the script below:

```
#! /bin/bash

mkdir -p data
cur_dir=$(pwd)

if [ ! -f data/text9 ]; then
  if [ ! -f data/enwik9.zip ]; then
    wget -c http://mattmahoney.net/dc/enwik9.zip -P data
  fi
  if [ ! -f data/enwik9 ]; then
    unzip data/enwik9.zip -d data
  fi
  perl wikifil.pl data/enwik9 > data/text9
fi

GENSIM=/home/gensim_user/gensim
ARGS="-train data/text9 -output /tmp/test -window 5 -negative 5 -threads 4 -min_count 5 -iter 5 -cbow 0"
SCRIPT=/usr/local/lib/python2.7/site-packages/gensim-3.5.0-py2.7-linux-x86_64.egg/gensim/scripts/word2vec_standalone.py
if [ -d $GENSIM ]; then
  rm -r $GENSIM
fi
cd /home/gensim_user/ && git clone https://github.com/RaRe-Technologies/gensim && cd gensim &&\
git checkout 96444a7d0357fc836641ec32c65eb2fbffbee68d && python setup.py install
cd $cur_dir
echo "RUNNING WORD2VEC BASE NO LOSS"
python $SCRIPT $ARGS 2>&1 | tee logs_base_word2vec_no_loss
pip uninstall -y gensim
cd /home/gensim_user/ && git clone https://github.com/alreadytaikeune/gensim && cd gensim && git checkout develop && python setup.py install
cd $cur_dir
echo "RUNNING WORD2VEC NEW NO LOSS"
python $SCRIPT $ARGS 2>&1 |tee logs_new_word2vec_no_loss
echo "RUNNING WORD2VEC NEW+LOSS"
python $SCRIPT $ARGS -loss 2>&1 |tee logs_new_word2vec_loss
```

The results of the run can be found here:
(Base no loss): https://gist.github.com/alreadytaikeune/554f21e95aa73ed414482d07b2e6314b
(New no loss): https://gist.github.com/alreadytaikeune/a184e88c059e038e2023170bb29a1eb7
(New loss): https://gist.github.com/alreadytaikeune/62ac3ca2da0d5d860404f0993acc81ac

I've only run them once, because it takes a bit of time. If we need more solid proof, probably we should settle on a smaller size corpus and average more runs. Anyways, it seems that the cost incurred to the computation of the loss amounts to around 3% of the total runtime, while the computation time is left unaffected (ignoring small variations between runs) as was expected.

I haven't run the benchmark on the base code with the loss computation, because if we agree on the fact that the previous computation is flawed, then I don't really see the relevance. Maybe we should discuss more on this point to double check my reasoning and implementation.

piskvorky:
Thanks a lot @alreadytaikeune ! That sounds good to me.

Let's wait for @menshikh-iv 's final verdict & merge, once he gets back from holiday.

gensim/models/base_any2vec.py
```diff
@@ -124,7 +124,17 @@ def _clear_post_train(self):
         raise NotImplementedError()
 
     def _do_train_job(self, data_iterable, job_parameters, thread_private_mem):
-        """Train a single batch. Return 2-tuple `(effective word count, total word count)`."""
+        """Train a single batch. Return 3-tuple
```
menshikh-iv: Please follow [numpy-style](https://numpydoc.readthedocs.io/en/latest/) for docstrings, more links available [here](https://github.com/RaRe-Technologies/gensim/projects/4#column-1683544)

gensim/models/base_any2vec.py
```diff
@@ -167,12 +177,23 @@ def _worker_loop(self, job_queue, progress_queue):
...
+                # gives coherent outputs with previous implementaitons
+                tally, raw_tally = stats_tuple
+                effective_samples = tally
 
             for callback in self.callbacks:
                 callback.on_batch_end(self)
 
-            progress_queue.put((len(data_iterable), tally, raw_tally))  # report back progress
+            # report back progress
+            progress_queue.put(
```
menshikh-iv: no need to break line (we use 120 char limit for gensim), here and everywhere

gensim/models/base_any2vec.py
```diff
@@ -260,6 +281,7 @@ def _log_train_end(self, raw_word_count, trained_word_count, total_elapsed, job_
 
     def _log_epoch_progress(self, progress_queue, job_queue, cur_epoch=0, total_examples=None, total_words=None,
                             report_delay=1.0):
+
```
menshikh-iv: please revert

gensim/models/word2vec_inner.pyx
```diff
@@ -491,8 +491,8 @@ def train_batch_sg(model, sentences, alpha, _work, compute_loss):
     cdef int negative = model.negative
     cdef int sample = (model.vocabulary.sample != 0)
 
-    cdef int _compute_loss = (1 if compute_loss else 0)
-    cdef REAL_t _running_training_loss = model.running_training_loss
+    cdef int _compute_loss = (1 if compute_loss is True else 0)
```
menshikh-iv: Why? old variant work correctly (same for cbow)

gensim/models/word2vec_inner.pyx
```diff
@@ -720,8 +721,8 @@ def train_batch_cbow(model, sentences, alpha, _work, _neu1, compute_loss):
                 if negative:
                     next_random = fast_sentence_cbow_neg(negative, cum_table, cum_table_len, codelens, neu1, syn0, syn1neg, size, indexes, _alpha, work, i, j, k, cbow_mean, next_random, word_locks, _compute_loss, &_running_training_loss)
 
-    model.running_training_loss = _running_training_loss
-    return effective_words
+    model.running_training_loss += _running_training_loss
+    return effective_words, effective_words
```
menshikh-iv: It is worth writing a comment, why return same value twice

gensim/models/word2vec.py
```diff
@@ -771,12 +773,11 @@ def _do_train_job(self, sentences, alpha, inits):
...
         work, neu1 = inits
-        tally = 0
         if self.sg:
-            tally += train_batch_sg(self, sentences, alpha, work, self.compute_loss)
+            (tally, effective_samples) = train_batch_sg(self, sentences, alpha, work, self.compute_loss)
         else:
-            tally += train_batch_cbow(self, sentences, alpha, work, neu1, self.compute_loss)
-        return tally, self._raw_word_count(sentences)
+            (tally, effective_samples) = train_batch_cbow(self, sentences, alpha, work, neu1, self.compute_loss)
+        return tally, self._raw_word_count(sentences), effective_samples
```
menshikh-iv: Need to update an docstrings everywhere when you change returning type

gensim/models/base_any2vec.py
```diff
@@ -966,6 +989,9 @@ def train(self, sentences=None, input_streams=None, total_examples=None, total_w
             total_words=total_words, epochs=epochs, start_alpha=start_alpha, end_alpha=end_alpha, word_count=word_count,
             queue_factor=queue_factor, report_delay=report_delay, compute_loss=compute_loss, callbacks=callbacks)
 
+    def get_latest_training_loss(self):
+        return 0
```
menshikh-iv: should raise `NotImplementet` (loss feature works only for w2v now, not for d2v/fasttext/etc) or  return `-1` maybe?

gensim/models/base_any2vec.py
```diff
@@ -167,12 +177,23 @@ def _worker_loop(self, job_queue, progress_queue):
...
 
-            tally, raw_tally = self._do_train_job(data_iterable, job_parameters, thread_private_mem)
+            stats_tuple = self._do_train_job(
+                data_iterable, job_parameters, thread_private_mem)
+            if len(stats_tuple) == 3:
+                tally, raw_tally, effective_samples = stats_tuple
+            else:
+                # Model doesn't implement the samples tallying. We assume
+                # that the number of samples is the effective words tally. This
+                # gives coherent outputs with previous implementaitons
```
menshikh-iv: please add TODO here - if/else should be removed when compute_loss implemented for all models

gensim/models/base_any2vec.py
```diff
@@ -1172,24 +1198,46 @@ def _log_progress(self, job_queue, progress_queue, cur_epoch, example_count, tot
...
+        else:
+            loss = self.get_latest_training_loss() / total_samples
         if total_examples:
             # examples-based progress %
-            logger.info(
-                "EPOCH %i - PROGRESS: at %.2f%% examples, %.0f words/s, in_qsize %i, out_qsize %i",
-                cur_epoch + 1, 100.0 * example_count / total_examples, trained_word_count / elapsed,
-                utils.qsize(job_queue), utils.qsize(progress_queue)
-            )
+            if self.compute_loss:
```
menshikh-iv: can you fully refactor this function please (make it clearer & shorter, not `if { if { } else { } } else { if { } else { } } `)?

Hint
- generate pattern (logging template first)
- collect needed parameters to list/tuple
- use `*my_parameters`

gensim/models/base_any2vec.py
alreadytaikeune: I changed it to raise an exception. I think -1 will be confusing since it is not the value that will be displayed. The value that is displayed is the running training loss divided by the number of samples. That is why I chose to return 0 and -1. -1 will look like the loss is decreasing as the number of processed words increases.

gensim/models/word2vec_inner.pyx
alreadytaikeune: I added the reason for that in the new docstring.

alreadytaikeune:
I have made the changes you requested, however I am a bit surprised by the outcomes of the tests. They all run smoothly on my machine and most of them pass in travis, except the py35-win one, for some reason that don't seem related at all with my changes.

gensim/models/base_any2vec.py
menshikh-iv: still here

gensim/models/base_any2vec.py
```diff
@@ -1172,24 +1202,28 @@ def _log_progress(self, job_queue, progress_queue, cur_epoch, example_count, tot
...
         else:
-            # words-based progress %
-            logger.info(
-                "EPOCH %i - PROGRESS: at %.2f%% words, %.0f words/s, in_qsize %i, out_qsize %i",
-                cur_epoch + 1, 100.0 * raw_word_count / total_words, trained_word_count / elapsed,
-                utils.qsize(job_queue), utils.qsize(progress_queue)
-            )
+            div = total_words
+
+        msg = "EPOCH %i - PROGRESS: at %.2f%% examples, %.0f words/s, in_qsize %i, out_qsize %i"
```
menshikh-iv: This can be PROGRESS: at %.2f%% **words** (not only `examples`)

gensim/models/base_any2vec.py
```diff
@@ -1172,24 +1202,28 @@ def _log_progress(self, job_queue, progress_queue, cur_epoch, example_count, tot
...
+        args = (cur_epoch + 1, 100.0 * example_count / div, trained_word_count / elapsed,
+                utils.qsize(job_queue), utils.qsize(progress_queue))
+        if self.compute_loss:
+            if total_samples == 0:
+                loss = -1
+            else:
+                loss = self.get_latest_training_loss() / total_samples
+            msg += ", current_loss %.3f"
```
menshikh-iv: maybe add loss to logging only if `self.compute_loss`?

gensim/models/word2vec.py
```diff
@@ -771,12 +778,11 @@ def _do_train_job(self, sentences, alpha, inits):
 
         """
         work, neu1 = inits
-        tally = 0
         if self.sg:
-            tally += train_batch_sg(self, sentences, alpha, work, self.compute_loss)
+            (tally, effective_samples) = train_batch_sg(self, sentences, alpha, work, self.compute_loss)
```
menshikh-iv: `()` no needed here (and same below)

gensim/models/base_any2vec.py
menshikh-iv: @alreadytaikeune yes, you are right, sorry

menshikh-iv:
@alreadytaikeune thanks, looks good! 

Next steps:
- Wait for merge #2127 (this almost done, I hope it will be merged soon)
- Resolve merge-conflicts 
- Add same calculations for new training mode (from #2127)
- Merge current PR

@alreadytaikeune I'll ping you when we'll be done with #2127.

menshikh-iv:
Hi @alreadytaikeune, we finally merged & release #2127, please do steps mentioned in https://github.com/RaRe-Technologies/gensim/pull/2135#issuecomment-413075975 and I'll merge your PR

alreadytaikeune:
Hi, I've completed the steps you mentioned, but when running the tests it seems to me that one is hanging, in the doc2vec test set. I don't really have a clue why, it wasn't the case before the merge.

menshikh-iv:
@alreadytaikeune I see no commits after my comment & merge conflict in PR, please do needed changes.
About any test hangs - reproduce it here first, please.

alreadytaikeune:
Ah it seems the CI server experiences the same stalling issues as me. I am not sure when I will have time to investigate that though...

menshikh-iv:
@alreadytaikeune CI issues fixed, hang reproduced in Travis, is you have time to fix & finalize?

alreadytaikeune:
I've found the bug, but I messed up a little bit when rebasing. I'm on it. I'll keep you posted.

alreadytaikeune:
Ok @menshikh-iv seems good now. Checks are passing and the loss value seems legit now. I guess some further development will be required on different models so they also can have their loss value logged. But at least there's no API incompatibility.

alreadytaikeune:
@menshikh-iv Hi, what's the status now? Do you have an idea where the conflicts come from and how to solve them. It seems to me that we could just regenerate these files with Cython, but I don't know if it is the principled way.

mpenkov:
@alreadytaikeune Hi, I'll be taking care of this PR from now. Please give me a few days to read through it.

gensim/models/base_any2vec.py
```diff
@@ -130,7 +130,21 @@ def _do_train_epoch(self, corpus_file, thread_id, offset, cython_vocab, thread_p
         raise NotImplementedError()
 
     def _do_train_job(self, data_iterable, job_parameters, thread_private_mem):
-        """Train a single batch. Return 2-tuple `(effective word count, total word count)`."""
+        """Train a single batch.
+        `
```
mpenkov: I think we should document the input parameters as well.

gensim/models/base_any2vec.py
```diff
@@ -208,12 +231,23 @@ def _worker_loop(self, job_queue, progress_queue):
             for callback in self.callbacks:
                 callback.on_batch_begin(self)
 
-            tally, raw_tally = self._do_train_job(data_iterable, job_parameters, thread_private_mem)
+            stats_tuple = self._do_train_job(
+                data_iterable, job_parameters, thread_private_mem)
```
mpenkov: I think there is no need to break the line here. We have a hard limit of 120 characters per line in gensim, and this is well within that limit.

gensim/models/base_any2vec.py
```diff
@@ -1080,6 +1115,10 @@ def train(self, sentences=None, corpus_file=None, total_examples=None, total_wor
             queue_factor=queue_factor, report_delay=report_delay, compute_loss=compute_loss, callbacks=callbacks,
             **kwargs)
 
+    def get_latest_training_loss(self):
+        raise NotImplementedError(
```
mpenkov: No need to line break here, still within 120 chars.

gensim/models/word2vec.py
mpenkov: @alreadytaikeune Still not done, please check

gensim/models/base_any2vec.py
```diff
@@ -1297,20 +1338,31 @@ def _log_progress(self, job_queue, progress_queue, cur_epoch, example_count, tot
         always be equal to -1.
 
         """
+        if self.compute_loss:
```
mpenkov: This feels redundant. Why is it here? How is this different from what is on line 1358?

gensim/models/base_any2vec.py
alreadytaikeune: I agree, though there are no clear constraints on what the inputs actually are. They can actually be pretty much anything depending on how the derived class implements `_get_thread_working_mem` and `_get_job_params`, the duty is to the derived class to figure out what to do with them. I'll figure out something.

gensim/models/base_any2vec.py
alreadytaikeune: Yes absolutely, this should not be here, it is a mishap of my rebasing.

alreadytaikeune:
Hi @mpenkov thanks for your feedback. I've incorporated some changes. Although I am a bit clueless about the reasons for the conflicts here. How should I proceed about solving them? Is there a preferred way?

About testing, I agree that it may be a good idea although it seems quite hard to do. The only thing I've done so far to convince myself that the values I now get for the loss I've explained in my very first message. Maybe we could write a test around the same ideas.

gensim/models/base_any2vec.py
```diff
@@ -130,7 +130,27 @@ def _do_train_epoch(self, corpus_file, thread_id, offset, cython_vocab, thread_p
         raise NotImplementedError()
 
     def _do_train_job(self, data_iterable, job_parameters, thread_private_mem):
-        """Train a single batch. Return 2-tuple `(effective word count, total word count)`."""
+        """Train a single batch.
+
+        Parameters
+        ----------
+        data_iterable : an iterable chunk of data representing a batch
```
mpenkov: Please replace "object" by the appropriate type.

```suggestion
        data_iterable: object
          an iterable chunk of data representing a batch
```

and do the same for the lines below.

gensim/models/word2vec.py
```diff
@@ -809,17 +816,17 @@ def _do_train_job(self, sentences, alpha, inits):
 
         Returns
         -------
-        (int, int)
-             2-tuple (effective word count after ignoring unknown words and sentence length trimming, total word count).
+        (int, int, int)
```
mpenkov: Isn't it simpler to just write:

```
int
    effective word count after ignoring unknown words and sentence length trimming
int
    total word count
int
    total effective samples
```

?

gensim/models/word2vec.py
```diff
@@ -787,13 +794,13 @@ def _do_train_epoch(self, corpus_file, thread_id, offset, cython_vocab, thread_p
...
+            train_fct = train_epoch_sg
         else:
-            examples, tally, raw_tally = train_epoch_cbow(self, corpus_file, offset, cython_vocab, cur_epoch,
-                                                          total_examples, total_words, work, neu1, self.compute_loss)
+            train_fct = train_epoch_cbow
+        examples, tally, raw_tally, effective_samples = train_fct(self, corpus_file, offset, cython_vocab,
+            cur_epoch, total_examples, total_words, work, neu1, self.compute_loss)
 
-        return examples, tally, raw_tally
+        return examples, tally, raw_tally, effective_samples
```
mpenkov: Please document the new return type in the docstring of this method.

mpenkov:
Don't worry about the conflicts. I'll take care of them during the merge: they are in autogenerated code, so it's easy to resolve.

Regarding tests: I think the actual loss output may be difficult to test (and not worth it) because it's being output by logs. What _is_ worth unit testing is the new behavior you added to the training functions (counting and returning the effective number of samples). Can you add tests to cover that new behavior?

For example:

- word2vec.py:train_batch_sg
- word2vec.py:train_batch_cbow
- word2vec.py:score_sentence_sg
- word2vec.py:_do_train_job
- ... and their counterparts in the Cython code.

You use the same testing logic and data for Python/Cython (they should be 100% compatible).

mpenkov:
@alreadytaikeune What is the status of this PR? Are you able to finish it?

tridelt:
@alreadytaikeune What do you think of the current status of this PR?
best regards from austria

alreadytaikeune:
Hello @mpenkov and @tridelt. Sorry for not being more involved in this PR. I've had plenty on my plate in my job, and in my personal life as well, and didn't feel like spending free/family time working on tests. Although I'm not comfortable with leaving this as is either. I'll do my best in the coming days to address your comments. Sorry not handling this earlier. Best.

alreadytaikeune:
Hello, @mpenkov. I found a bit of time to throw myself back into the PR. I thought I would rebase my develop branch on the current one and work on the tests from there, but it was not a good idea.... anyway, I've restored my branch as it was before.

Just one thing, when rebasing it seemed to me that all the non-cython implementation of word2vec was removed. Is it correct? Regarding your last comment about checks, I imagine I should now only care about writing tests for the cython functions.

Sorry about the mess...

gojomo:
@alreadytaikeune Yes,, the non-cython variants of many of the algorithms have been eliminated to reduce duplication/maintenance effort – so work going forward need only concern itself with the cython versions!

alreadytaikeune:
Alright thank you @gojomo . I also see that `word2vec.py` became a script as well as a module, and it does the same thing as `scripts/word2vec_standalone.py`, the file I had originally modified. Which one should I focus on?

gojomo:
AFAIK, `word2vec.py` has always implemented a `main` & been runnable as a script. Unsure of the reason for (partial) overlap with `scripts/word2vec_standalone.py` – & neither version has been updated recently to keep-up with newer `Word2Vec` initialization options (like `ns_exponent` or `corpus_file`). I suspect `gensim` usage is far more via imported code (vs command-line invocation), and new functions only get added to the `main` paths when someone specifically needs them. @piskvorky or @mpenkov would have to comment on whether both, or just one or the other, should be updated going forward.

piskvorky:
@gojomo is right. I'm not aware of any changes to the CLI version of word2vec in Gensim. IIRC, the command line version was created to match the CLI of the original C tool by Google.

I don't remember what `word2vec_standalone.py` was. I tracked that script down to PR #593, which doesn't talk about its motivation. But a standalone script makes more sense to me, and is also cleaner (avoids pickling issues when run as `__main__` etc).

So my preference would be to get rid of `__main__` in `word2vec.py`, and keep `word2vec_standalone.py`. We definitely don't need both.

alreadytaikeune:
Thank you for the clarification. So, going forward, I'll just write some tests for cython only implementations (which is a bit confusing since I'm working on a version with still the two implementations, hence why I wanted to rebase in the first place), and leave to you clarifying between what's to be kept and dropped between `word2vec.py` and `word2vec_standalone.py`.

geopapa11:
I was wondering about the status of this PR. Will it be fixed in gensim 4.0.0? Erroneous loss computation happens only in Word2Vec or in other classes too (i.e., FastText)? Thanks in advance

piskvorky:
Yes, ideally we want to clean this up for 4.0 too. @alreadytaikeune will you able to finish this PR?

Anyone else helping is welcome.

geopapa11:
Thank you very much @piskvorky! Does wrong loss computation affect other classes too (e.g., FastText)?

piskvorky:
I don't remember, but I think the loss tallying was idiosyncratic – different for each class, missing from some models.

geopapa11:
Thanks @piskvorky that's very helpful. On an unrelated topic, do you guys plan (or find use) in supporting different initialization schemes for embeddings? Right now you are only supporting uniform initialization I think between [-0.5*embedding_size, 0.5*embedding_size]. I have created other types of initializations as well ('glorot_uniform', 'lecun_uniform', 'he_uniform', 'glorot_normal', 'lecun_normal', 'he_normal'). If you find it might add something in the tooling, I can give some ideas on a different PR.

piskvorky:
Possibly; @Witiko did some work on weight initialization. That's out of topic for this ticket though.

gojomo:
FYI #2617 is a sort of 'umbrella issue' mentioned/referencing all the tangled loss-issues. Ideally we'd want our measure/reporting to be similar to the output in Facebook's FastText logging. (Right now Gensim's `FastText` hasn't ever tracked loss, and whatever Gensim's `Word2Vec` is doing doesn't report similar tallies to FB FastText-in-plain-Word2Vec mode - so there's some discrepancy to be tracked-down & rationalized.)

Regarding alternate weight initializations, it shouldn't be hard to externally re-initialize (or include an option to not initialize for when a user is planning to do this themselves) between vocabulary-discovery and training. We could add some standard options if there's a strong case for how they improve things. (Do they noticeably hasten convergence?)

geopapa11:
> FYI #2617 is a sort of 'umbrella issue' mentioned/referencing all the tangled loss-issues. Ideally we'd want our measure/reporting to be similar to the output in Facebook's FastText logging. (Right now Gensim's `FastText` hasn't ever tracked loss, and whatever Gensim's `Word2Vec` is doing doesn't report similar tallies to FB FastText-in-plain-Word2Vec mode - so there's some discrepancy to be tracked-down & rationalized.)

Regarding this, how are you updating the embedding weights on every iteration if you don't compute the loss? Do you use some kind of ready formula for gradient that's directly applied to update the embedding weights?

> Regarding alternate weight initializations, it shouldn't be hard to externally re-initialize (or include an option to not initialize for when a user is planning to do this themselves) between vocabulary-discovery and training. We could add some standard options if there's a strong case for how they improve things. (Do they noticeably hasten convergence?)

Not sure if they improve or change convergence in a meaningful way. I just mentioned it because `Embeddings ` as defined in other libraries (e.g., TensorFlow) offer different initialization schemes.
https://www.tensorflow.org/api_docs/python/tf/keras/layers/Embedding (see `embeddings_initializer ` argument)
https://www.tensorflow.org/api_docs/python/tf/keras/initializers

Also, your initialization which is more like `[-0.5*fan_out, 0.5*fan_out]` does not really correspond to one of the more known ones: He, Glorot, Lecun but maybe that's ok! :-)

Witiko:
> Possibly; @Witiko did some work on weight initialization. That's out of topic for this ticket though.

@piskvorky @geopapa11 I evaluated different initizations of the positional model of [Mikolov et al. (2018)](https://arxiv.org/pdf/1712.09405.pdf) and [Grave et al. (2018)](https://arxiv.org/pdf/1802.06893.pdf), since no reference implementation existed and it's unclear how the weights should be initialized, see pages 59 through 62 in [the RASLAN 2020 proceedings](https://nlp.fi.muni.cz/raslan/raslan20.pdf#page=68).

nicocheh:
I was wondering about the status of this PR. Is there any news about this PR? @mpenkov

mpenkov:
Not that I'm aware of. I think we're waiting for the original contributor to finish the PR: https://github.com/RaRe-Technologies/gensim/pull/2135#issuecomment-743767056

PR #2146, opened 2018-08-06
Add multiprocessing support for `BM25`

Shiki-H:
I realized that computing BM25 for large corpus can be quite time consuming, so I added multiprocessing support for the original BM25. I did not open an issue as this is a very quick fix and thought that I would probably make a PR directly. I also added a test to verify the result of using multiprocessing is identical to the original approach. Probably not many people use this function, but hopefully it helps :)

### DIFF
--- a/gensim/summarization/bm25.py
+++ b/gensim/summarization/bm25.py
@@ -38,7 +38,7 @@ import math
 from six import iteritems
 from six.moves import xrange
 from functools import partial
-from multiprocessing import Pool
+from multiprocessing import Pool, cpu_count
 
 
 PARAM_K1 = 1.5
@@ -153,6 +153,7 @@ class BM25(object):
             scores.append(score)
         return scores
 
+
 def _get_scores(bm25, document, average_idf):
     scores = []
     for index in xrange(bm25.corpus_size):
@@ -161,7 +162,18 @@ def _get_scores(bm25, document, average_idf):
     return scores
 
 
-def get_bm25_weights(corpus):
+def _effective_n_jobs(n_jobs):
+    """Determine the number of jobs which are going to run in parallel"""
+    if n_jobs == 0:
+        raise ValueError('n_jobs == 0 in Parallel has no meaning')
+    elif n_jobs is None:
+        return 1
+    elif n_jobs < 0:
+        n_jobs = max(cpu_count() + 1 + n_jobs, 1)
+    return n_jobs
+
+
+def get_bm25_weights(corpus, n_jobs=1):
     """Returns BM25 scores (weights) of documents in corpus.
     Each document has to be weighted with every document in given corpus.
 
@@ -189,13 +201,19 @@ def get_bm25_weights(corpus):
     bm25 = BM25(corpus)
     average_idf = sum(float(val) for val in bm25.idf.values()) / len(bm25.idf)
 
+    if _effective_n_jobs(n_jobs) == 1:
+        weights = [bm25.get_scores(doc, average_idf)]
+        return weights
+
     # weights = []
     # for doc in corpus:
     #     scores = bm25.get_scores(doc, average_idf)
     #     weights.append(scores)
 
     get_score = partial(_get_scores, bm25, average_idf=average_idf)
-    pool = Pool()
+    pool = Pool(n_jobs)
     weights = pool.map(get_score, corpus)
+    pool.close()
+    pool.join()
 
     return weights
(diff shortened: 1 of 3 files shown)


gensim/summarization/bm25.py
```diff
@@ -174,15 +198,21 @@ def get_bm25_weights(corpus):
...
-        weights.append(scores)
+    n_processes = _effective_n_jobs(n_jobs)
+    if n_processes == 1:
+        weights = [bm25.get_scores(doc, average_idf) for doc in corpus]
+        return weights
+
+    get_score = partial(_get_scores, bm25, average_idf=average_idf)
+    pool = Pool(n_processes)
+    weights = pool.map(get_score, corpus)
+    pool.close()
```
menshikh-iv: strange order, you `close` and `join` after, why?

gensim/test/test_BM25.py
```diff
@@ -44,6 +44,15 @@ def test_disjoint_docs_if_weight_zero(self):
         self.assertAlmostEqual(weights[0][1], 0)
         self.assertAlmostEqual(weights[1][0], 0)
 
+    def test_multiprocessing(self):
+        """ Result should be the same using different processes """
+        weights1 = get_bm25_weights(common_texts)
+        weights2 = get_bm25_weights(common_texts, n_jobs=2)
+        weights3 = get_bm25_weights(common_texts, n_jobs=-1)
+        self.assertEqual(weights1, weights2)
```
menshikh-iv: should be `assertAlmostEqual` instead (never compare floating point values using "strict" equal)

gensim/summarization/bm25.py
```diff
@@ -152,14 +154,36 @@ def get_scores(self, document, average_idf):
...
+
+
+def _effective_n_jobs(n_jobs):
+    """Determine the number of jobs which are going to run in parallel"""
+    if n_jobs == 0:
+        raise ValueError('n_jobs == 0 in Parallel has no meaning')
+    elif n_jobs is None:
+        return 1
+    elif n_jobs < 0:
+        n_jobs = max(cpu_count() + 1 + n_jobs, 1)
```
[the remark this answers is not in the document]
Shiki-H: @menshikh-iv I was trying to stick to `sklearn` style here, where `n_jobs=-1` means using all cores. In fact, this way of determining the number of effective jobs was borrowed from `sklearn` which can be found [here](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/externals/joblib/_parallel_backends.py). Please let me know if you think this is ok.

gensim/test/test_BM25.py
Shiki-H: @menshikh-iv That was my bad. Fixed now.

gensim/summarization/bm25.py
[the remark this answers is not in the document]
menshikh-iv: ok, that's fine

gensim/summarization/bm25.py
```diff
@@ -152,14 +154,68 @@ def get_scores(self, document, average_idf):
...
+
+def get_bm25_weights(corpus, n_jobs=1):
     """Returns BM25 scores (weights) of documents in corpus.
     Each document has to be weighted with every document in given corpus.
 
     Parameters
     ----------
     corpus : list of list of str
         Corpus of documents.
+    n_jobs : int
```
menshikh-iv: `int, optional`

gensim/summarization/bm25.py
menshikh-iv: thanks!

gensim/summarization/bm25.py
```diff
@@ -152,14 +154,68 @@ def get_scores(self, document, average_idf):
...
+
+    """
+    scores = []
+    for index in xrange(bm25.corpus_size):
+        score = bm25.get_score(document, index, average_idf)
+        scores.append(score)
+    return scores
+
+
+def _effective_n_jobs(n_jobs):
```
menshikh-iv: Can you move this function to `gensim.utils` and rename it to `effective_n_jobs` (looks useful for later usage).

gensim/summarization/bm25.py
```diff
@@ -152,14 +154,68 @@ def get_scores(self, document, average_idf):
...
+
+    Parameters
+    ----------
+    n_jobs : int
+        Number of workers requested by caller.
+
+    Returns
+    -------
+    int
+        number of effective jobs
```
menshikh-iv: `n` -> `N` + `.` at the end of sentence

gensim/summarization/bm25.py
Shiki-H: @menshikh-iv thanks. I have fixed them.

menshikh-iv:
Thanks @Shiki-H, congratz with first contribution :1st_place_medal:

Shiki-H:
@menshikh-iv Thanks :)

PR #2147, opened 2018-08-06
Make `word2vec2tensor` script compatible with `python3`

### LINKED ISSUE #1958: gensim.scripts.word2vec2tensor TypeError: write() argument must be str, not bytes

vsocrates:
Fixes #1958. The word2vec2tensor script didn't support Python 3 due to unicode encoding. This was fixed and tests were added to ensure that the script functions correctly in the future. The usage of the script on the user side remains the same.


gensim/scripts/word2vec2tensor.py
```diff
@@ -63,16 +67,15 @@ def word2vec2tensor(word2vec_model_path, tensor_filename, binary=False):
         True if input file in binary format.
 
     """
-    model = gensim.models.KeyedVectors.load_word2vec_format(word2vec_model_path, binary=binary)
+    model = gensim.models.KeyedVectors.load_word2vec_format(word2vec_model_path, binary=binary, datatype=np.float64)
```
menshikh-iv: why `np.float64` instead of `np.float32` (that's default parameter)?

gensim/test/test_scripts.py
```diff
@@ -1,108 +1,186 @@
...
+        # Check interlinks
+        self.assertTrue(interlinks['self-governance'] == 'self-governed')
+        self.assertTrue(interlinks['Hierarchy'] == 'hierarchical')
+        self.assertTrue(interlinks['Pierre-Joseph Proudhon'] == 'Proudhon')
+
+
+class TestWord2Vec2Tensor(unittest.TestCase):
+    def setUp(self):
+        self.datapath = datapath('word2vec_pre_kv_c')
+        self.output_folder = get_tmpfile('')
```
menshikh-iv: better to add an explicit name (for example `w2v2t_test`)

gensim/test/test_scripts.py
```diff
@@ -1,108 +1,186 @@
...
+        self.metadata_file = self.output_folder + '_metadata.tsv'
+        self.tensor_file = self.output_folder + '_tensor.tsv'
+        self.vector_file = self.output_folder + '_vector.tsv'
+
+    def testConversion(self):
+        word2vec2tensor(word2vec_model_path=self.datapath, tensor_filename=self.output_folder)
+
+        try:
```
menshikh-iv: No need try/except in current test. if exception raised - test failed, this is expected behavior.

gensim/test/test_scripts.py
```diff
@@ -1,108 +1,186 @@
...
+                    'Tensor file %s creation failed. \
+                    Check the parameters and input file format.' % self.tensor_file
+                    )
+
+        # check if number of words and vector size in tensor file line up with word2vec
+        with smart_open(self.datapath, 'rb') as f:
+            first_line = f.readline().strip()
+
+        number_words, vector_size = map(int, first_line.split(b' '))
+        if not len(metadata) == len(vectors) == number_words:
```
menshikh-iv: `assert <CONDITION>, <MESSAGE>`

gensim/test/test_scripts.py
```diff
@@ -1,108 +1,186 @@
...
+            self.fail(
+                'Metadata file %s and tensor file %s \
```
menshikh-iv: we using 120 character limit, no reasons to make short lines

gensim/test/test_scripts.py
```diff
@@ -1,108 +1,186 @@
...
+                imply different number of rows.' % (self.metadata_file, self.tensor_file)
+                )
+
+        # write word2vec to file
```
menshikh-iv: The strange part of the test, you copy part of the code from script to test this script, why?

gensim/scripts/word2vec2tensor.py
vsocrates: For some reason, when the word2vec model is loaded, if we leave the default, using the test data file `word2vec_pre_kv_c` I run into parsing issues. For instance, for the word `the`, I get `array([-0.56110603, -1.97569799,  1.66395497, -1.23224604,  0.75475103,
        0.98576403,  2.26144099, -0.59829003, -0.47433099, -1.41610503], dtype=float32)` instead of `-0.561106 -1.975698 1.663955 -1.232246 0.754751 0.985764 2.261441 -0.598290 -0.474331 -1.416105`

gensim/test/test_scripts.py
vsocrates: I'm, sorry, could you be a bit more specific? Do you mean the lines from 154-157? If so, the idea was to take the metadata and tensor components that were separated and put the back together in w2v style without any gensim functions and make sure they are the same as what is created by the `word2vec2tensor.py` script. I guess it is basically just tested `to_utf8` though. Is it fine to remove it?

gensim/scripts/word2vec2tensor.py
menshikh-iv: And what's a problem here? I still don't catch, sorry

gensim/test/test_scripts.py
menshikh-iv: Not in the current case, stack trace (if something goes wrong) have enough information in current tests.

gensim/test/test_scripts.py
menshikh-iv: I think it's fine to remove it, yes

gensim/scripts/word2vec2tensor.py
vsocrates: Sorry, should have been more clear. In the first dimension, as an example, the original word2vec model has `-0.561106` but for some reason, without the `np.float64` when it is read in, it displays `-0.56110603`, so the test fails equality. I'm not sure on the reason, though I assume something to do with the `load_word2vec_format` internals.

menshikh-iv: what's assertion you mean? can you link concrete line of code?

Also, you can fix assertion (slightly relax "almost equal" condition), but anyway, you shouldn't change dtype here.

vsocrates:
@menshikh-iv The changes you mentioned have been made. 

With regards to the `float32` issue, I thought it'd be easier if I showed you the issue I was getting. The CI tests above show what happens when I don't include the `float64` data type. The `np.testing` module tests to 7 decimal places, and what is in the original file vs. what is written to the two tensor files differs by less than 7 decimal places.

menshikh-iv:
@vsocrates ok, I see, simply "relax" `decimal` parameter of `assert_almost_equal` function (to 6 or 5), that's not an big deal (and I can merge current PR after CI are passed).

menshikh-iv:
Thank you @vsocrates, congtarz with the first contribution :1st_place_medal: !

PR #2148, opened 2018-08-06
Fix minor semantic issue in docs for phrases

RunHorst:
The impact of the `threshold` param is heavily influenced by the choice of the scoring function - the docs should reflect this (instead of saying that this _hardly_ matters, wich means the opposite).
It's a tiny change, but might prevent some confusion in future readers of the docs


piskvorky:
Can you please fix `socring` => `scoring` in the same sentence, while we're at it?

RunHorst:
@piskvorky fixed in 2d066d5, together with other minor corrections

piskvorky:
Nice, I appreciate that. Many of our open source contributors are not native speakers (myself included), so such corrections and clean-ups always help a lot!

RunHorst:
Glad to help! With that in mind, I'll see if I can find time to do more non-code "fixes"

PR #2151, opened 2018-08-09
Remove duplicate count from `Phraser` log message

robguinness:
I suspect this duplicate count was there for debugging purposes, but perhaps it should be removed now.


menshikh-iv:
Yes, looks like you are right, thanks @robguinness, congratz with the first contribution!

robguinness:
Thanks.

PR #2157, opened 2018-08-19
Fix typo in documentation

KenjiOhtsuka:
Fixed misspelling.

`/dics/src/changes_080.rst` contains the word "mmapped". Is it the misspelling of "mapped"? If so, I'll fix it too.  

> Large matrices (numpy/scipy.sparse, in `LsiModel`, `Similarity` etc.) are now **mmapped** to/from disk when doing `save/load`. The `cPickle` approach used previously was too `buggy <http://groups.google.com/group/gensim/browse_thread/thread/3c4c6c0f76c5938c#>`_ and `slow <http://dieter.plaetinck.be/poor_mans_pickle_implementations_benchmark.html>`_.

### DIFF
--- a/docs/src/changes_080.rst
+++ b/docs/src/changes_080.rst
@@ -69,7 +69,7 @@ Other changes (that you're unlikely to notice unless you look)
 ----------------------------------------------------------------------
 
 * Improved efficiency of ``lsi[corpus]`` transformations (documents are chunked internally for better performance).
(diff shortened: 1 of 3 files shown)


piskvorky:
Thanks for the typo fix.

[mmap](https://en.wikipedia.org/wiki/Mmap) is not a misspelling, but if you see it used somewhere prominent, feel free to hyperlink the term to https://en.wikipedia.org/wiki/Mmap (in case it helps others).

menshikh-iv:
Thanks @KenjiOhtsuka, congratz with first contribution :1st_place_medal:

piskvorky:
Eh, changing the release notes of Gensim 0.8.0 (> 7 years ago) was not really necessary :)

I meant hyperlinking `mmap` in actual documentation -- where it matters, where people can see it.

Issue #2162, opened 2018-08-24
Fix nested inline markup in `translation_matrix.py`


nzw0301:
#### Description

reStructuredText parser cannot deal with nested inline markup according to http://docutils.sourceforge.net/FAQ.html#is-nested-inline-markup-possible .

So, [this line](https://github.com/RaRe-Technologies/gensim/blob/46124f4895b4bec6eee58c23e1686e5c9efb84fd/gensim/models/translation_matrix.py#L50) does not show correctly:

![2018-08-25 1 17 10](https://user-images.githubusercontent.com/7121753/44596212-60acaf00-a806-11e8-8525-936d894c8ed4.png)

There are two ways to deal with this nested inline problem, but they are not recommended by [this site](http://docutils.sourceforge.net/FAQ.html#is-nested-inline-markup-possible).

I guess that the most simple way is to avoid to use the bold style for the link part:
`**How to make translation between two** :class:`~gensim.models.doc2vec.Doc2Vec` **models**`.

![2018-08-25 1 42 56](https://user-images.githubusercontent.com/7121753/44596715-3eb42c00-a808-11e8-8614-5452d3314f7c.png)

What do you think about it?

piskvorky:
In fact, there's no need for that line to be bolded at all.

Can you make that line a heading instead? Or if that doesn't work, simply drop the `**`.

nzw0301:
Thank you for your comment! 
I will do that and open a PR for this issue!

menshikh-iv:
Fixed by #2164

PR #2164, opened 2018-08-25
Use heading instead of bold style in `translation_matrix.py`

nzw0301:
This PR solves #2162 .

- Use heading instead of nested inline code
- Use also heading for the same level bolded sentence

----

# Preview


menshikh-iv:
Thanks for PR @nzw0301 :+1:

Issue #2172, opened 2018-09-03
Allow asymmetrical windows for word2vec.


generall:
I found that [WordRank ](https://radimrehurek.com/gensim/models/wrappers/wordrank.html) model could be trained with `symmetric` parameter. I allows to predict next word in sequence based only on the left ones. I think, this option should be also helpful for other models like Word2Vec and FastText.

gojomo:
This has come up before; is there any evidence that asymmetric windows perform better for certain tasks, and which tasks?

MatthieuMontecot:
Hi, I would like to use asymnetric windows for IP embedding (in order to replicate IP2Vec implementation). This approach implies to specifie pairs of words as I/O of  W2V model (lets say skip-gram). Is there a way to force asymnetric role in pairs of words? (context/prediction)? I saw on blogs that I could modifye myself the librairies but I'm not comfortable with messing with the licence.

gojomo:
There's no built-in support for varying the context windows from the classic 'symmetric around a target word' - so you'd have to modify the code for most such customization. (If such modifications were clean & flexible, with demonstrations of situations where they offer a clear benefit, they could be welcome contributions to the project's main codebase.)

I'm not familiar with the specific 'IP2Vec' approach you've mentioned, but in some case, corpus preprocessing may be able to simulate other policies, especially with regard to skip-gram training. 

For example, the 'sentence' `['a', 'b', 'c', 'd', 'e']` with `window=5` would normally result in a particular set of skip-gram pairs being used for training (which happens to include every 'word' with every other, though with some probabilistic weighting where the nearer-pairs are trained more often). 

If you were, completely outside of `Word2Vec`, split that 'sentence' into your own smaller sentences before passing to `Word2Vec`, you could effectively overweight or remove unwanted pairings. For example, if `['a', 'd']` is known to be significant, despite the 3-distance, while `['a', 'c']` is known to be irrelevant, your expansion of the original 5-token sentences into N 2-token sentences could repeat `['a', 'd']` more than once, and leave out `['a', 'c']` completely, to have more influence over the actual skip-grams used without needing to modify Gensim's `Word2Vec` code at all.

MatthieuMontecot:
Thanks for this reply, what I want in the case you are talking about is that I want the network to be trained with 'a' as input and 'b' as target but NOT the opposite (I don't want it to train it with 'b' as input and 'a' as target which I think it still does if I give a sentence  ['a','b'] in the current implementation.
[...]

PR #2173, opened 2018-09-03
Allow asymmetric window for Word2Vec and FastText

generall:
Solves issue https://github.com/RaRe-Technologies/gensim/issues/2172


generall:
Strange error occurred during last test: https://travis-ci.org/RaRe-Technologies/gensim/jobs/423968507#L969 It didn't appear in any other tests in this PR, didn't appear in other envs and has no relations to my changes. So I assume this is pre-existing randomly appearing bug.

menshikh-iv:
@generall this is a non-deterministic test, Travis job restarted

gensim/models/base_any2vec.py
```diff
@@ -580,6 +580,8 @@ def __init__(self, sentences=None, input_streams=None, workers=3, vector_size=10
             The beginning learning rate. This will linearly reduce with iterations until it reaches `min_alpha`.
         window : int, optional
             The maximum distance between the current and predicted word within a sentence.
+        symmetric : {0, 1}, optional
```
menshikh-iv: Why only symmetric + only left, can we extend this (since we still add a new attribute for it)?

gensim/models/base_any2vec.py
```diff
@@ -612,6 +614,7 @@ def __init__(self, sentences=None, input_streams=None, workers=3, vector_size=10
             logger.warning("consider setting layer size to a multiple of 4 for greater performance")
         self.alpha = float(alpha)
         self.window = int(window)
+        self.symmetric = bool(symmetric)
```
menshikh-iv: need to modify `load` function (for backward compatibility reason) + don't forget about test

gensim/models/fasttext.py
```diff
@@ -81,11 +81,10 @@
 import numpy as np
 from numpy import ones, vstack, empty, float32 as REAL, sum as np_sum
 
-from gensim.models.word2vec import Word2VecVocab, Word2VecTrainables, train_sg_pair, train_cbow_pair
-from gensim.models.keyedvectors import Vocab, FastTextKeyedVectors
 from gensim.models.base_any2vec import BaseWordEmbeddingsModel
+from gensim.models.keyedvectors import Vocab, FastTextKeyedVectors
 from gensim.models.utils_any2vec import _compute_ngrams, _ft_hash
-
+from gensim.models.word2vec import Word2VecVocab, Word2VecTrainables, train_sg_pair, train_cbow_pair
```
menshikh-iv: please revert unrelated changes (like this)

gensim/models/fasttext.py
```diff
@@ -823,8 +826,7 @@ def _load_vectors(self, file_handle):
         num_vectors, dim = self.struct_unpack(file_handle, '@2q')
         # Vectors stored by [Matrix::save](https://github.com/facebookresearch/fastText/blob/master/src/matrix.cc)
         assert self.wv.vector_size == dim, (
-            'mismatch between vector size in model params ({}) and model vectors ({})'
-            .format(self.wv.vector_size, dim)
+            'mismatch between vector size in model params ({}) and model vectors ({})'.format(self.wv.vector_size, dim)
```
menshikh-iv: and this too (all changes that not related to PR)

gensim/models/base_any2vec.py
menshikh-iv: also, if you put it here, don't forget to check how it works for all *2vec models (includes doc2vec)

gensim/models/word2vec.py
```diff
@@ -190,7 +190,11 @@ def train_batch_sg(model, sentences, alpha, work=None, compute_loss=False):
 
                 # now go over all words from the (reduced) window, predicting each one in turn
                 start = max(0, pos - model.window + reduced_window)
-                for pos2, word2 in enumerate(word_vocabs[start:(pos + model.window + 1 - reduced_window)], start):
+                if not hasattr(model, 'symmetric') or model.symmetric:
```
menshikh-iv: part with `hasattr` should be in `load` function and initialize it with 1 as default value

gensim/models/base_any2vec.py
menshikh-iv: CC @gojomo

piskvorky: We had a PR with a more general solution, that allowed you to input arbitrary word pairs (so it's up to you how you construct these "contexts"). The left/right context was just one special case.

I think the PR wasn't merged and you later closed it, and I cannot find it now. This was maybe 3 years ago.

generall:
@menshikh-iv I guess I fixed PR by your notes

gensim/models/base_any2vec.py
menshikh-iv: Um, if it was 3 years ago and I close it  - this PR was abandoned a long time ago (several years), it is already completely irrelevant (but I would like to know what approach was used there), because I prefer more general solution.

gensim/models/word2vec.py
```diff
@@ -190,7 +190,11 @@ def train_batch_sg(model, sentences, alpha, work=None, compute_loss=False):
 
                 # now go over all words from the (reduced) window, predicting each one in turn
                 start = max(0, pos - model.window + reduced_window)
-                for pos2, word2 in enumerate(word_vocabs[start:(pos + model.window + 1 - reduced_window)], start):
+                if model.symmetric:
```
menshikh-iv: You also need to make similar changes for `*_inner.pyx` files, because this doesn't affect to "fast" versions

generall:
> Don't forget about doc2vec (just add it to test)
I have added checks for loading old Doc2Vec models here: https://github.com/RaRe-Technologies/gensim/pull/2173/files#diff-7268265e58ffaa87707f3d358f25ae17
New doc2vec model has no `symmetric` param, it uses default value every time, so I don't think any additional tests are required.

> Can you propose more general approach
I can't come up with any use cases for other window shapes. Right symmetric version could be easily obtained by reverting input.

> How this affect to model (some examples, benchmarks, etc)?
I tried to prepare test with a demonstration of this affect, but it seems that gensim training works stochasticly even with fixed seed. So Travis tests have failed. Can you please help me with this?
You can check my attempts here https://github.com/RaRe-Technologies/gensim/pull/2173/commits/c5af2cfc67603f855330e0883c8824459879f0c5

generall:
> I tried to prepare test with a demonstration of this affect, but it seems that gensim training works stochasticly even with fixed seed. So Travis tests have failed. Can you please help me with this?

Looks like I forgot to regenerate cython code. Should work now.

menshikh-iv:
@generall
- about doc2vec - you can pass it through `kwargs` and this should works. Sanity check: train model (with `symmetric=0`) and infer vector, that's enough for simple test
- about benchmark:
  1. You can do it without travis: write small script, train model and publish results here (if issue only with travis), if you train it on your own - model can be larger (we are interested in results in this case)
  2. If you want to have "determenistic" training, you need to pin seed, set 1 worker and use https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHASHSEED variable (for py3 only).

BTW, I re-run appveyour (non-determenistic test again).

generall:
> You can do it without travis: write small script, train model and publish results here (if issue only with travis), if you train it on your own - model can be larger (we are interested in results in this case)

Probably I will do it some later

menshikh-iv:
@generall ok, we'll wait for benchmark from you (this is really interesting, how "left" window improve results)

schmiflo:
This would be a very useful feature. Can we expect a pull soon?

generall:
@menshikh-iv  Hi, unforgettably I could not demonstrate asymmetric window improvements on text datasets. In my experiments I used transactions data which I can't share. Maybe @schmiflo can explain why this functionality would be useful for him?

schmiflo:
Any chance this will be released soon?

mpenkov:
@generall What is the status of this PR? Are you able to finish it?

generall:
Hi, unfortunately i have no time to work on this PR right now. Maybe some time later

PR #2194, opened 2018-09-22
Add missing content to tarball

menshikh-iv:
The main reason why conda packages not builded - missing files in tarball, because conda use `tar.gz` from PyPI, unlike gensim-wheels, where we use repository itself. Anyway, missing files in tarball is a bug (because user can't build all from sources).

With adding new files, conda build & test package succesfully (see https://github.com/conda-forge/gensim-feedstock/pull/14 for more info)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,14 +4,29 @@ include CHANGELOG.md
 include COPYING
 include COPYING.LESSER
 include ez_setup.py
+
 include gensim/models/voidptr.h
+include gensim/models/fast_line_sentence.h
+
 include gensim/models/word2vec_inner.c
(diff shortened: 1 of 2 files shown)


yurivict:
This patch doesn't apply on top of the release.
You should make another release.

menshikh-iv:
@yurivict I know (fix added after the last release), this will be included in next release.

mino98:
Apologies for commenting in a closed issue, but I was wondering if you're planning a 3.6.0 re-release or a minor subversion (even better?) which includes this patch.

As of today, three months after this PR, the `3.6.0` source tar.gz still cannot be built and we're resorting to a dump of the `develop` branch. This is not ideal, I guess.

Thanks a lot.

PR #2195, opened 2018-09-22
Rename vocab attribute to bigram_counts

### LINKED ISSUE #2189: Phraser requires unnecessary memory
Currently, `Phraser` objects (= the trimmed-down version of the full bigram finder `Phrases`) contains the actual bigrams in an internal attribute called [phrasegrams](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/phrases.py#L775). This is the biggest and most memory-intense part of a `Phraser` object.

`phrasegrams` is a dict of `{tuple of strings => (frequency [int], score [float])}`. But the `int` (the frequency count of that particular bigram) is unused. This means we're constructing that int, plus the wrapping tuple, for no good reason, inflating the necessary RAM. See also [mailing list discussion](https://groups.google.com/forum/#!topic/gensim/vsWXxUzuK6s).

Task:
* Drop the `int` from Phraser values, leaving only the `float`.
* And while at it, rename (deprecate) the `.vocab` attribute of Phrases to something more appropriate, for example `bigram_counts`.

souravsingh:
Fixes #2189

### DIFF
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -772,7 +772,7 @@ class Phraser(SentenceAnalyzer, PhrasesTransformation):
         for bigram, score in phrases_model.export_phrases(corpus, self.delimiter, as_tuples=True):
             if bigram in self.phrasegrams:
                 logger.info('Phraser repeat %s', bigram)
-            self.phrasegrams[bigram] = (phrases_model.vocab[self.delimiter.join(bigram)], score)
+            self.phrasegrams[bigram] = (phrases_model.bigram_count[self.delimiter.join(bigram)], score)
             count += 1
             if not count % 50000:
                 logger.info('Phraser added %i phrasegrams', count)


gensim/models/phrases.py
```diff
@@ -772,7 +772,7 @@ def __init__(self, phrases_model):
         for bigram, score in phrases_model.export_phrases(corpus, self.delimiter, as_tuples=True):
             if bigram in self.phrasegrams:
                 logger.info('Phraser repeat %s', bigram)
-            self.phrasegrams[bigram] = (phrases_model.vocab[self.delimiter.join(bigram)], score)
+            self.phrasegrams[bigram] = (phrases_model.bigram_count[self.delimiter.join(bigram)], score)
```
menshikh-iv: of course this doesn't work: `bigram_count` is not defined

menshikh-iv:
Hi @souravsingh, please
- read issue again (and related mailing list thread), the main goal isn't renaming: main point here is reduce memory usage.
- don't rename anything (at least for now), this is really non-trivial (as you can see, you breaks all code-CI now)
- Try to implement improvement suggested by @piskvorky and don't forget about benchmarks (old vs new) and tests (we must be 100% sure that we don't break compatibility).

menshikh-iv:
@souravsingh is this abandoned? Should I close this?

souravsingh:
@jenishah really wants to work on this. Closing

PR #2196, opened 2018-09-23
Fix non-deterministic test failures

menshikh-iv:
Some of the non-deterministic tests failed frequently, I spend much time to rerun it again and again (especially in release time when we need to build gensim many times), most frequent fails are:
- `TestFastTextModel.test_cbow_hs_online`
- `TestWord2VecModel.test_cbow_hs_fromfile`
- `TestDoc2VecModel.testLoadOldModel`

This fails almost always happens on `python3`.

Here I pin `PYTHONHASHSEED` for testing purposes (will do the same thing for `gensim-wheels` repo too).

CC: @gojomo @piskvorky hope that you have no objection.

See also https://github.com/MacPython/gensim-wheels/pull/12

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -32,6 +32,7 @@ setenv =
     DTM_PATH={env:DTM_PATH:}
     MALLET_HOME={env:MALLET_HOME:}
     SKIP_NETWORK_TESTS={env:SKIP_NETWORK_TESTS:}
+    PYTHONHASHSEED=1
 
 commands =
     python -c "from gensim.models.word2vec import FAST_VERSION; print(FAST_VERSION)"


piskvorky:
* Pros: More reliable builds, faster releases, less confusion for maintainers and contributors investigating unrelated CI failures.
* Cons: moving away from functional tests (does the algo still work acceptably?) to more menial bit-checking (algo may be broken for a different seed).

Definitely worth it IMO (pros >> cons).

gojomo:
I'd have to see the individual failures to be sure, but if it really is a small set of commonly-failing tests, they could probably improved to be 'tighter', rather than for the whole environment to be forced into string-hash-determinism. In particular, that it's two variations of the "cbow_hs" test commonly failing makes me very suspicious of that test's design. 

And for `TestDoc2VecModel.testLoadOldModel`, I'm a bit more baffled - as there's no new training there, normal sources of randomness shouldn't risk changes in substantive results. It's loading all static models! There must be a true problem in the test, I'd think.

PR #2197, opened 2018-09-23
Fix `SvmLightCorpus.serialize` if `labels` instance of numpy.ndarray 

### LINKED ISSUE #2113: Found error in check labels condition while creating corpus using SvmLightCorpus.serialize
#### Description
Found error in check labels condition while creating corpus using SvmLightCorpus.serialize 

- Use twenty_newsgroups dataset from sklearn.datasets
- tokenize data and create corpus using gensim dictionary.doc2bow 
- create tfidf of the corpus 
- Pass corpus_tfidf and labels in:
 gensim.corpora.SvmLightCorpus.serialize('svm_20_news_groups.train', corpus_tfidf, labels=labels)

Suggestion: 
Change line 117 to :
` label = labels[docno] if labels.any() else 0  # target class is 0 by default`
or 
` label = labels[docno] if labels is not None and len(labels)>0 else 0  # target class is 0 by default`

#### Expected Results
Should save a gensim corpus

#### Actual Results
File "svm.py", line 40, in <module>
    gensim.corpora.SvmLightCorpus.serialize('svm_20_news_groups.train', corpus_tfidf, labels=labels)
File "C:\Users\shiwangisingh\AppData\Local\Programs\Python\Python36\lib\site-packages\gensim\corpora\indexedcorpus.py", line 117, in serialize
    offsets = serializer.save_corpus(fname, corpus, id2word, **kwargs)
File "C:\Users\shiwangisingh\AppData\Local\Programs\Python\Python36\lib\site-packages\gensim\corpora\svmlightcorpus.py", line 117, in save_corpus
    label = labels[docno] if labels else 0  # target class is 0 by default
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

aquatiko:
Fix #2113 

Changed numpy.ndarray to list.

### DIFF
--- a/gensim/corpora/svmlightcorpus.py
+++ b/gensim/corpora/svmlightcorpus.py
@@ -99,7 +99,7 @@ class SvmLightCorpus(IndexedCorpus):
             Corpus in BoW format.
         id2word : dict of (str, str), optional
             Mapping id -> word.
-        labels : list, numpy.ndarray or False
+        labels : list, numpy.ndarray, False
             An SVMlight `<target>` class tags or False if not present.
         metadata : bool
             ARGUMENT WILL BE IGNORED.
@@ -112,8 +112,8 @@ class SvmLightCorpus(IndexedCorpus):
         """
         logger.info("converting corpus to SVMlight format: %s", fname)
 
-        if type(labels) == np.ndarray:
-            labels = labels.tolist()    
+        if isinstance(labels, np.ndarray):
+            labels = labels.tolist()
         offsets = []
         with utils.smart_open(fname, 'wb') as fout:
             for docno, doc in enumerate(corpus):
--- a/gensim/corpora/svmlightcorpus.py
+++ b/gensim/corpora/svmlightcorpus.py
@@ -112,9 +112,8 @@ class SvmLightCorpus(IndexedCorpus):
         """
         logger.info("converting corpus to SVMlight format: %s", fname)
 
-        if type(labels)== np.ndarray:
-            labels= labels.tolist()
-            
+        if type(labels) == np.ndarray:
+            labels = labels.tolist()    
         offsets = []
         with utils.smart_open(fname, 'wb') as fout:
             for docno, doc in enumerate(corpus):
--- a/gensim/corpora/svmlightcorpus.py
+++ b/gensim/corpora/svmlightcorpus.py
@@ -99,7 +99,7 @@ class SvmLightCorpus(IndexedCorpus):
             Corpus in BoW format.
         id2word : dict of (str, str), optional
             Mapping id -> word.
-        labels : list, numpy.ndarray, False
+        labels : iterable of ELEMENT_TYPE or False
             An SVMlight `<target>` class tags or False if not present.
         metadata : bool
             ARGUMENT WILL BE IGNORED.
@@ -112,8 +112,9 @@ class SvmLightCorpus(IndexedCorpus):
         """
         logger.info("converting corpus to SVMlight format: %s", fname)
 
-        if isinstance(labels, np.ndarray):
-            labels = labels.tolist()
+        if labels is not False:
+            # Cast any sequence (incl. a numpy array) to a list, to simplify the processing below.
+            labels = list(labels)
         offsets = []
         with utils.smart_open(fname, 'wb') as fout:
             for docno, doc in enumerate(corpus):
--- a/gensim/corpora/svmlightcorpus.py
+++ b/gensim/corpora/svmlightcorpus.py
@@ -14,6 +14,7 @@ import logging
 
 from gensim import utils
 from gensim.corpora import IndexedCorpus
+import numpy as np
 
 
 logger = logging.getLogger(__name__)
@@ -98,7 +99,7 @@ class SvmLightCorpus(IndexedCorpus):
             Corpus in BoW format.
         id2word : dict of (str, str), optional
             Mapping id -> word.
-        labels : list or False
+        labels : list, numpy.ndarray or False
             An SVMlight `<target>` class tags or False if not present.
         metadata : bool
             ARGUMENT WILL BE IGNORED.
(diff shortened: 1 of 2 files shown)


gensim/corpora/svmlightcorpus.py
```diff
@@ -98,7 +99,7 @@ def save_corpus(fname, corpus, id2word=None, labels=False, metadata=False):
             Corpus in BoW format.
         id2word : dict of (str, str), optional
             Mapping id -> word.
-        labels : list or False
+        labels : list, numpy.ndarray or False
```
menshikh-iv: also please fix documentation building (current variant are breaking sphinx, see doc build log https://circleci.com/gh/RaRe-Technologies/gensim/1535?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

gensim/corpora/svmlightcorpus.py
```diff
@@ -112,8 +112,8 @@ def save_corpus(fname, corpus, id2word=None, labels=False, metadata=False):
         """
         logger.info("converting corpus to SVMlight format: %s", fname)
 
-        if type(labels) == np.ndarray:
-            labels = labels.tolist()    
+        if isinstance(labels, np.ndarray):
```
menshikh-iv: Isn't better to just call `list(labels)` without any type cheks? In this case, `labels` can be any iterable (and no need to check types).

gensim/corpora/svmlightcorpus.py
```diff
@@ -111,6 +112,8 @@ def save_corpus(fname, corpus, id2word=None, labels=False, metadata=False):
         """
         logger.info("converting corpus to SVMlight format: %s", fname)
 
+        if isinstance(labels, np.ndarray):
+            labels = labels.tolist()
```
piskvorky: Needs a comment: np arrays are sequences too, so it's not obvious why they should be a special case.

By the way, should they? Can't we check for "emptiness" in a way that works for arbitrary sequence, without singling out numpy arrays?

aquatiko: @piskvorky if we simply do,
```
if  labels:
    labels = list(labels)
```
it would surely work for arbitrary sequences, so should the documentation also be changed to reflect that support?

piskvorky: I don't think the `if` will work -- won't it have the same problem as the current version, if `labels` is a numpy array?

aquatiko: @piskvorky  would this be good enough to use?? seems to working fine 
```
try:
    labels = list(labels)
except:
    pass
```

piskvorky: What the `except` for?

menshikh-iv: @piskvorky `labels` can be `False`/`None` (in case if no labels), maybe for this we really need to use `isinstance`.

piskvorky: Aha, I see. Then the try/catch makes sense (but needs to catch a specific Exception, and of course a comment explaining what's happening).

aquatiko: @piskvorky would this be good for a commit
```
# If arbitrary sequences (e.g numpy array) is given as input to labels, casting
# to list will generate TypeError if labels is False (empty)
try:
    labels = list(labels)
except TypeError:
    pass
```

piskvorky: I don't really understand that nested `if` in the comment. Why should casting a numpy array to list generate a TypeError?

menshikh-iv: @piskvorky because `labels` can be "empty" (see my previous comment https://github.com/RaRe-Technologies/gensim/pull/2197#discussion_r220801693)

See an examples

```python
>>> list(None)
TypeError: 'NoneType' object is not iterable
```

```python
>>> list(False)
TypeError: 'bool' object is not iterable
```

piskvorky: I understand how TypeError is thrown when `labels` is `False`. But @aquatiko 's code comment implies an exception in case `labels` is a numpy array. But maybe it's just badly phrased?

aquatiko: @piskvorky will this comment work?
```
# When labels=False, the operation for casting arbitrary sequences (e.g numpy array)
# to list will cause a TypeError 
```

aquatiko: @piskvorky  This looks better than the try/except method. Do the docstring needs to be changed? As ```labels``` will support arbitrary sequences now and its written-- labels : list or False

piskvorky: I think so, yes. @menshikh-iv do we write `sequence` or `iterable` in such cases?

menshikh-iv: @piskvorky `iterable of TYPE`

gensim/corpora/svmlightcorpus.py
```diff
@@ -14,6 +14,7 @@
 
 from gensim import utils
 from gensim.corpora import IndexedCorpus
+import numpy as np
```
piskvorky: I don't think this is needed.

gensim/corpora/svmlightcorpus.py
```diff
@@ -98,7 +99,7 @@ def save_corpus(fname, corpus, id2word=None, labels=False, metadata=False):
             Corpus in BoW format.
         id2word : dict of (str, str), optional
             Mapping id -> word.
-        labels : list or False
+        labels : iterable of ELEMENT_TYPE or False
```
piskvorky: ELEMENT_TYPE is a placeholder: please replace by the actual type that is expected inside `labels`.

menshikh-iv:
Ping @aquatiko, are you planning to finish PR?

menshikh-iv:
@aquatiko :worried: why you close it?

aquatiko:
@menshikh-iv Sorry about that, while learning the git workflow I accidentally deleted it. If you would allow I would like to submit another PR.

PR #2205, opened 2018-10-02
Update documented return types of evaluation funcs

### LINKED ISSUE #2203: Wrong return types in docstrings for wv.evaluate_word_analogies and wv.evaluate_word_pairs
This is the documented return type for `KeyedVectors.evaluate_word_analogies`:
https://github.com/RaRe-Technologies/gensim/blob/6a4424d3e465dea0afcddc4f4e2410cf0f52dba1/gensim/models/keyedvectors.py#L1070

The dicts in the second element of the return tuple have as values either a string or a list of 4-tuples. For instance:
```
{
    'section': 'family',
    'correct': [('he', 'she', 'dad', 'mom'), ...],
    'incorrect': [('man', 'woman', 'groom', 'bride'), ...]
}
```
So I believe the return value of `evaluate_word_analogies` is actually `Tuple[float, List[Dict[str, Union[str, List[Tuple[str, str, str, str]]]]]` (in PEP-484 style annotation).

Then, in `KeyedVectors.evaluate_word_pairs`, the documented return type is:
https://github.com/RaRe-Technologies/gensim/blob/6a4424d3e465dea0afcddc4f4e2410cf0f52dba1/gensim/models/keyedvectors.py#L1284

Here, the second and third elements of the return tuple are actually pairs of floats, since they are the values returned from `scipy.stats.pearsonr` and `spearmanr`, which return both the correlation and a p-value. I don't know whether the p-value should be discarded or the documented return type be updated to `(float, (float, float), (float, float))`.

Stigjb:
Fixes #2203

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -1067,8 +1067,12 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
 
         Returns
         -------
-        (float, list of dict of (str, (str, str, str))
-            Overall evaluation score and full lists of correct and incorrect predictions divided by sections.
+        score : float
+            The overall evaluation score on the entire evaluation set
+        sections : list of dict of {str : str or tuple of (str, str, str, str)}
+            Results broken down by each section of the evaluation set. Each dict contains the name of the section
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Looks good, thanks @Stigjb, congratz with first contribution :1st_place_medal:

PR #2208, opened 2018-10-03
Reduce `Phraser` memory usage (drop frequencies)

### LINKED ISSUE #2189: Phraser requires unnecessary memory

jenishah:
Fix #2189 , removed the frequency from the `phrasregrams` dict and keeping it as a dict of float, that keeps the scores only.


menshikh-iv:
Hello, thanks for PR @jenishah, can you add test to check than this change are compatible with old phrases? I mean you need to train & save phrases with old gensim code (3.6.0 without your change) and load with your new code and check than this works as expected.

menshikh-iv:
@jenishah you should add it as a test here ;) Just add the old model file to `gensim/test/test_data` + add the test where you load this model and use it (apply to some text)

menshikh-iv:
Ping @jenishah, are you planning to finish PR?

gensim/test/test_phrases.py
```diff
@@ -646,6 +646,19 @@ def testEncoding(self):
         self.assertTrue(isinstance(transformed, six.text_type))
 
 
+class TestPhraserModelCompatibilty(unittest.TestCase):
+
+    def TestCompatibilty(self):
+        with temporary_file("phraser_model_3dot6") as fpath:
```
menshikh-iv: This isn't a temporary file. You should use `datapath` for retrieve path to model.

gensim/test/test_phrases.py
```diff
@@ -646,6 +646,19 @@ def testEncoding(self):
...
+            bigram_loaded = Phraser.load(fpath)
+            test_sentences = [u'trees', u'graph', u'minors']
+            prev_ver = bigram_loaded(test_sentences)
+            phrase_model = Phrases(test_sentences, threshold=1)
+            bigram = Phraser(phrase_model)
+            curr_ver = bigram(test_sentences)
```
menshikh-iv: but `Phraser` isn't callable, `[test_sentences]` should be used here (same for 655). How this pass tests?

gensim/models/phrases.py
```diff
@@ -805,7 +805,7 @@ def __init__(self, phrases_model):
         for bigram, score in phrases_model.export_phrases(corpus, self.delimiter, as_tuples=True):
             if bigram in self.phrasegrams:
                 logger.info('Phraser repeat %s', bigram)
-            self.phrasegrams[bigram] = (phrases_model.vocab[self.delimiter.join(bigram)], score)
+            self.phrasegrams[bigram] = (None, score)
```
piskvorky: -1: this doesn't solve the problem, tuples use a lot of memory (as does the extra None pointer). 80 extra bytes per entry, on my machine.

If you want to keep backward compatibility, check the dict format on `load` and convert from tuple to simple float if necessary.

gensim/test/test_phrases.py
```diff
@@ -14,7 +14,6 @@
 import six
 
 import numpy as np
-
```
piskvorky: Keep the blank please. We separate third party modules from internal modules by a blank line (another visual block).

The blank line between `six` and `numpy` shouldn't be there though :)

gensim/models/phrases.py
```diff
@@ -848,7 +848,10 @@ def score_item(self, worda, wordb, components, scorer):
 
         """
         try:
-            return self.phrasegrams[tuple(components)][1]
+            if list(self.phrasegrams.values())[0].__class__ is tuple:
```
horpto: 1. this check is very tricky to just check that value is instance of tuple. Python has **isinstance** builtin function for that.
2. creation of list of values is not easy operation. It would be better to avoid this.

```suggestion
            score = self.phrasegrams[tuple(components)]
            return score[-1] if isinstance(score, tuple) else score
```

jenishah: But *isinstance* has to be applied to dictionary value, which requires using *list*.

gensim/models/phrases.py
```diff
@@ -848,7 +848,11 @@ def score_item(self, worda, wordb, components, scorer):
 
         """
         try:
-            return self.phrasegrams[tuple(components)][1]
+            score = self.phrasegrams[tuple(components)]
+            if isinstance(score, tuple):
```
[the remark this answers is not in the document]
piskvorky: No, I think that's fine: look at the type when loading the object, and if it's the "old" type (tuple), convert to the new type (floats) right away.

gensim/models/phrases.py
```diff
@@ -208,6 +208,10 @@ def load(cls, *args, **kwargs):
         """
         model = super(PhrasesTransformation, cls).load(*args, **kwargs)
         # update older models
+        # if value in phrasegrams dict is a tuple, load only the scores.
+        if len(model.__dict__['phrasegrams']):
+            if isinstance(list(model.__dict__['phrasegrams'].values())[0], tuple):
+                model.__dict__['phrasegrams'].update((k, v[1]) for k, v in model.__dict__['phrasegrams'].items())
```
piskvorky: This is hard to read. Can you rephrase this with normal dict access syntax, and using better variable names? (not `k` and `v` and `v[1]`).

Also, changing the dict at the same time you're iterating over it sounds tricky. Probably safer to iterate over each value, and if it's a tuple, change it right away (in-place, without a full dict copy).

jenishah: Would there be any case where some values are tuple and some are not?
If not, we can check only one instance and based on that modify the entire dictionary.

piskvorky: I don't think so -- all values should always be in the same format (all old, all new).

My suggestion was more around iteration and per-element assignment, so whether you check each value makes no difference (in fact, per-value checks may look cleaner, because there's no special case of "dictionary empty").

gensim/models/phrases.py
```diff
@@ -210,10 +210,12 @@ def load(cls, *args, **kwargs):
         # update older models
         # if value in phrasegrams dict is a tuple, load only the scores.
         try:
-            if isinstance(list(model.__dict__['phrasegrams'].values())[0], tuple):
-                model.__dict__['phrasegrams'].update((k, v[1]) for k, v in model.__dict__['phrasegrams'].items())
+            for components, scores in model.__dict__['phrasegrams'].items():
```
piskvorky: Why this strange `__dict__` access? Why not use the pattern I showed in my last review?

piskvorky: Does this really work? It mutates the collection it's iterating over, which is usually a bad idea.

As mentioned previously, I'd make a copy of the original keys (not items) and iterate over that, while mutating the original (large) dict.

gensim/models/phrases.py
```diff
@@ -209,12 +209,12 @@ def load(cls, *args, **kwargs):
...
         # if value in phrasegrams dict is a tuple, load only the scores.
-        try:
-            for components, scores in model.__dict__['phrasegrams'].items():
-                if isinstance(scores, tuple):
-                    model.__dict__['phrasegrams'][components] = scores[1]
-        except KeyError:
-            pass
+        if model.phrasegrams:
+            components = model.phrasegrams.keys()
+            for component in components:
```
piskvorky: These two lines are better merged into one (so the temporary variable is released as soon as it's not needed).

gensim/models/phrases.py
```diff
@@ -209,12 +209,12 @@ def load(cls, *args, **kwargs):
...
+                score = model.phrasegrams[component]
+                if isinstance(score, tuple):
+                    model.phrasegrams[component] = score[1]
```
piskvorky: Deserves a comment: what is `score[1]`?

Or even better, unroll the tuple into properly named variables (`x, y, z = score`) and then assign that.

menshikh-iv:
thanks for work @jenishah :muscle:

PR #2212, opened 2018-10-04
use itertools.chain to concatenate lists

Stigjb:
The Python docs discourage `sum` for concatenating lists, and suggests `itertools.chain` instead<sup>[[1]](https://docs.python.org/3.5/library/functions.html#sum)</sup>. I think this makes it clearer that the value is a list and not a number.

Also, the unit test for the `evaluate_word_analogies` function seemed to have been written for the deprecated `accuracy` function, where Word2Vec accuracy and KeyedVector accuracy may have been separate methods (?). It seemed pointless to compute analogies twice on the same model, so I changed the test to instead check important aspects of the return value from `evaluate_word_analogies`.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -149,6 +149,7 @@ and so on.
 from __future__ import division  # py3 "true division"
 
 from collections import deque
+from itertools import chain
 import logging
 
 try:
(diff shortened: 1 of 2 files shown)


menshikh-iv:
Hello @Stigjb, thanks for PR

Can you do the same changes for all `sum(..., [])` please, here is examples that I found (maybe somewhere else?)
```
gensim/models/keyedvectors.py:1147:            'correct': sum((s['correct'] for s in sections), []),
gensim/models/keyedvectors.py:1148:            'incorrect': sum((s['incorrect'] for s in sections), []),
gensim/models/keyedvectors.py:1253:            'correct': sum((s['correct'] for s in sections), []),
gensim/models/keyedvectors.py:1254:            'incorrect': sum((s['incorrect'] for s in sections), []),
gensim/models/wrappers/ldamallet.py:225:                tokens = sum(([self.id2word[tokenid]] * int(cnt) for tokenid, cnt in doc), [])
gensim/models/wrappers/ldamallet.py:227:                tokens = sum(([str(tokenid)] * int(cnt) for tokenid, cnt in doc), [])
gensim/models/deprecated/keyedvectors.py:985:            'correct': sum((s['correct'] for s in sections), []),
gensim/models/deprecated/keyedvectors.py:986:            'incorrect': sum((s['incorrect'] for s in sections), []),
gensim/test/test_corpora_dictionary.py:261:        all_tokens = sum(texts, [])
```

PR #2214, opened 2018-10-05
Cleanup (get rid py26 stuff)

menshikh-iv:
Just small cleanup
- Remove `ez_setup.py` (no more relevant)
- Fix loggers (use `__name__` everywhere)
- Add 3.7 to classifier
- Use built-in `bit_length`


gensim/models/lda_dispatcher.py
```diff
@@ -74,7 +74,7 @@
 from gensim.models.lda_worker import LDA_WORKER_PREFIX
 
 
-logger = logging.getLogger("gensim.models.lda_dispatcher")
+logger = logging.getLogger(__name__)
```
piskvorky: This is not a good idea in CLI scripts (`__name__` will be just `__main__` there).

menshikh-iv: fixed: ee87342

PR #2215, opened 2018-10-05
[WIP] Fix `FastText`

### LINKED ISSUE #2059: fasttext ft_hash and unicode handling
Fasttext uses the hashing trick to map ngrams to a an index in [0, N]. Gensim supports loading models trained with original fasttext implementation from facebook research. It is therefore important that both gensims and the original implementation use the same hash function to make sure that ngrams are associated with the correct vectors.

The [original implementation](https://github.com/facebookresearch/fastText/blob/53dd4c5/src/dictionary.cc#L143) is in C++ and considers ngrams as `std::basic_string<char>`, ie. a sequence of bytes:

```
uint32_t Dictionary::hash(const std::string& str) const {
  uint32_t h = 2166136261;
  for (size_t i = 0; i < str.size(); i++) {
    h = h ^ uint32_t(str[i]);
    h = h * 16777619;
  }
  return h;
}
```

However the [gensim python implementation](https://github.com/RaRe-Technologies/gensim/blob/9021ea8b31871e760095086669651dd1f072fce0/gensim/models/utils_any2vec.py#L26) consider a ngram as a sequence of unicode characters:

```
        h = np.uint32(2166136261)
        for c in string:
            h = h ^ np.uint32(ord(c))
            h = h * np.uint32(16777619)
        return h

```

The two are not equivalent. Consider:
```
In [1]: import numpy as np

In [2]: def hash_unicode(s):
   ...:     h = np.uint32(2166136261)
   ...:     for c in s:
   ...:         h = h ^ np.uint32(ord(c))
   ...:         h = h * np.uint32(16777619)
   ...:     return h
   ...:
   ...:
   ...: def hash_bytes(s):
   ...:     h = np.uint32(2166136261)
   ...:     s = s.encode('utf-8')
   ...:     for c in s:
   ...:         h = h ^ np.uint32(c)
   ...:         h = h * np.uint32(16777619)
   ...:     return h
   ...:
   ...:

In [3]: hash_unicode("é")
Out[3]: 1812687940

In [4]: hash_bytes("é")
Out[4]: 513665217
```

I am not really familiar with gensims code, so I may have overlooked something.

I assume that the cython implementation treats the iteration over a `unicode` string also as an iteration over unicode characters, so the same consideration as above would apply, but I haven't verified this.

Edit: I checked the inputs to the hash function of the original fasttext implementation. There, eg. `<α>` is represented by `3c ffffffce ffffffb1 3e`

### LINKED ISSUE #2139: FastText incremental training fails
#### Description
Having successfully trained model (with 20 epochs), which has been saved and loaded back without any problems, I'm trying to continue training it for another 10 epochs - on the same data, with the same parameters - but it fails with an error: `TypeError: 'NoneType' object is not subscriptable` (for full traceback see below).

```python
from gensim.models.fasttext import FastText

# `train_data' is just a list of lists of strings (words), e.g.
# `[['w1', 'w2', 'w3', ...], ['w1', 'w4', 'w5', ...], ...]'.
model = FastText(
    train_data,
    sg=1,
    size=200,
    window=5,
    min_count=1,
    workers=16,
    negative=20,
    iter=20,
    min_n=3,
    max_n=5,
    word_ngrams=1,
    bucket=int(2e6)
)

# `model_file' is a string with the path to the file where model is being saved
model.save(model_file)

model = FastText.load(model_file)

# `train_data' here is exactly the same as before
model.train(train_data, epochs=10, total_examples=model.corpus_count)
```

#### Expected Results
Successfully trained model.

#### Actual Results
```python
[WARNING 2018-07-23 14:42:00,222] Effective 'alpha' higher than previous training cycles
[INFO 2018-07-23 14:42:00,222] training model with 16 workers on 15145 vocabulary and 200 features, using sg=1 hs=0 sample=0.001 negative=20 window=5
Exception in thread Thread-50:
  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.5/threading.py", line 862, in run
  File "/home/ubuntu/env/lib/python3.5/site-packages/gensim/models/base_any2vec.py", line 99, in _worker_loop
    tally, raw_tally = self._do_train_job(data_iterable, job_parameters, thread_private_mem)
  File "/home/ubuntu/env/lib/python3.5/site-packages/gensim/models/fasttext.py", line 454, in _do_train_job
    tally += train_batch_sg(self, sentences, alpha, work, neu1)
  File "gensim/models/fasttext_inner.pyx", line 319, in gensim.models.fasttext_inner.train_batch_sg
TypeError: 'NoneType' object is not subscriptable
```

#### Versions
Linux-4.15.0-1014-gcp-x86_64-with-Ubuntu-16.04-xenial
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609]
NumPy 1.14.5
SciPy 1.1.0
gensim 3.4.0
FAST_VERSION 1

### LINKED ISSUE #2160: [Feature request] Load full native fastText model to continue training on new data
Currently gensim cannot load and continue training native fastText model. According to the docs [[1]](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/fasttext.py#L685), this is because it only loads input-hidden matrix. However, fastText also saves hidden-output matrix [[2]](https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc#L176). 

Moreover, even the input-hidden matrix could support some sort of transfer learning, with hidden-output matrix inited randomly, similar to how `gensim.models.Word2Vec.intersect_word2vec_format()` works.

Please correct me if I'm wrong here, but I think there is no technical issue preventing loading and continue training fastText model. How about supporting this feature?

menshikh-iv:
The main goal - fix annoying FastText bugs

Should 
- [x] Fix #2139 (continue training after loading or training, without `build_vocab(update=True)` call
- [ ] [IN PROGRESS] Fix #2160 (allow to train loaded FB model)
- [ ] Fix #2059 (or just check, maybe it works correctly, I'm still unsure)

Note: PR will be updated in slow mode (hope to finish before next release).

### DIFF
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -886,6 +886,8 @@ class FastText(BaseWordEmbeddingsModel):
             .format(
                 self.wv.vectors_ngrams.shape, (self.trainables.bucket + len(self.wv.vocab), self.wv.vector_size)
             )
+        self.trainables.syn1neg = np.fromfile(file_handle, dtype=dtype, count=len(self.wv.vocab) * dim)
+        self.trainables.syn1neg = self.trainables.syn1neg.reshape((dim, len(self.wv.vocab)))
 
         self.trainables.init_ngrams_post_load(self.file_name, self.wv)
         self.trainables.vectors_vocab_lockf = ones((len(self.wv.vocab), self.wv.vector_size), dtype=REAL)
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -529,7 +529,7 @@ class FastText(BaseWordEmbeddingsModel):
         self.wv.vectors_norm = None
         self.wv.vectors_vocab_norm = None
         self.wv.vectors_ngrams_norm = None
-        self.wv.buckets_word = None
+        # self.wv.buckets_word = None
 
     def estimate_memory(self, vocab_size=None, report=None):
         vocab_size = vocab_size or len(self.wv.vocab)
@@ -926,7 +926,7 @@ class FastText(BaseWordEmbeddingsModel):
 
         """
         kwargs['ignore'] = kwargs.get(
-            'ignore', ['vectors_norm', 'vectors_vocab_norm', 'vectors_ngrams_norm', 'buckets_word'])
+            'ignore', ['vectors_norm', 'vectors_vocab_norm', 'vectors_ngrams_norm'])  # , 'buckets_word'])
         super(FastText, self).save(*args, **kwargs)
 
     @classmethod
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -836,7 +836,7 @@ class FastText(BaseWordEmbeddingsModel):
             word = word_bytes.decode(encoding)
             count, _ = self.struct_unpack(file_handle, '@qb')
 
-            self.wv.vocab[word] = Vocab(index=i, count=count)
+            self.wv.vocab[word] = Vocab(index=i, count=count, sample_int=i)
             self.wv.index2word.append(word)
 
         assert len(self.wv.vocab) == nwords, (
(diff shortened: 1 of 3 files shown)


aakash086:
Hi , 

Can someone please tell in which gensim fastext version would this fix be available .

menshikh-iv:
@aakash086 Hi, I hope that in `3.7.0` or `3.7.1`

menshikh-iv:
Continued in #2313

PR #2225, opened 2018-10-12
Remove `alpha="auto"` from `LdaMulticore` (not supported yet)

### LINKED ISSUE #2223: LdaMulticore parameter alpha: 'auto' not supported but documentation says it is
When using the value 'auto' for parameter "alpha" for LdaMulticore, the following exception is raised:
```
NotImplementedError: auto-tuning alpha not implemented in multicore LDA; use plain LdaModel.
```
(ldamulticore.py, line 146)

However the documentation string of the `__init__` states that "auto" is supported.

This may be a copy/paste error, but is there any chance to see auto get supported for the Multicore version as well? 

#### Versions
```
Linux-4.15.0-35-generic-x86_64-with-debian-buster-sid
Python 3.6.3 |Anaconda custom (64-bit)| (default, Oct 13 2017, 12:02:49) 
[GCC 7.2.0]
NumPy 1.14.2
SciPy 1.0.0
gensim 3.4.0
FAST_VERSION 1
```

However, the exception code is still present in the latest master as well, I think.

johann-petrak:
The 'auto' parameter value for parameter alpha is not supported
in LdaMulticore, so remove it from the documentation.

Fix #2223

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -371,7 +371,7 @@ class LdaModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
             Mapping from word IDs to words. It is used to determine the vocabulary size, as well as for
             debugging and topic printing.
         distributed : bool, optional
-            Whether distributed computing should be used to accerelate training.
+            Whether distributed computing should be used to accerelate training. 
         chunksize :  int, optional
             Number of documents to be used in each training chunk.
         passes : int, optional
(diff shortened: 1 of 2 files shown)


johann-petrak:
Commulative patch for both locations mentioned in #2223

menshikh-iv:
Thank you @johann-petrak, congratz with the first contribution :1st_place_medal:

Issue #2228, opened 2018-10-14
`pyemd` import check is misplaced


piskvorky:
Gensim only checks whether `pyemd` in installed once, on module load:
https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/keyedvectors.py#L755

All other calls to `wmdistance` will automatically fail, even if the `pyemd` module has been installed in the meanwhile.

This is unnecessary and annoying. Python is a dynamic language, we can import the module when needed (= where the check is actually happening, in `wmdistance`), not at the top.

Plus let's fix the `werein` typo in `At least one of the documents had no words that werein the vocabulary` and the non-Pythonic `len(document1) == 0` while at it. The code could use some style clean up.

kishan3:
Sir, please explain what is a pythonic way to equality. Also where to import `emd` if not at the beginning.

menshikh-iv:
@kishan3 I think @piskvorky mean "import pyemd when this needed", i.e. directly in `wmdistance` function.

kishan3:
@menshikh-iv ok got you.

Naba7:
Can I be assigned this issue?

Naba7:
@menshikh-iv I have opened a pr.

menshikh-iv:
@Naba7 of course, feel free to contribute (continue with your #2247)

menshikh-iv:
Fixed by #2240

PR #2229, opened 2018-10-15
Fix typo, pythonic equality, fix import for pyemd.

kishan3:
Fix for issue #2228

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -167,14 +167,6 @@ try:
 except ImportError:
     from Queue import Queue, Empty  # noqa:F401
 
-# If pyemd C extension is available, import it.
-# If pyemd is attempted to be used, but isn't installed, ImportError will be raised in wmdistance
-try:
-    from pyemd import emd
-    PYEMD_EXT = True
-except ImportError:
-    PYEMD_EXT = False
-
 from numpy import dot, float32 as REAL, empty, memmap as np_memmap, \
     double, array, zeros, vstack, sqrt, newaxis, integer, \
     ndarray, sum as np_sum, prod, argmax, divide as np_divide
@@ -764,9 +756,9 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
         if diff1 > 0 or diff2 > 0:
             logger.info('Removed %d and %d OOV words from document 1 and 2 (respectively).', diff1, diff2)
 
-        if not len(document1) or not len(document2):
+        if not document1 or not document2:
             logger.info(
-                "At least one of the documents had no words that wherein the vocabulary. "
+                "At least one of the documents had no words that were in the vocabulary. "
                 "Aborting (returning inf)."
             )
             return float('inf')


gensim/models/keyedvectors.py
```diff
@@ -765,9 +764,9 @@ def wmdistance(self, document1, document2):
         if diff1 > 0 or diff2 > 0:
             logger.info('Removed %d and %d OOV words from document 1 and 2 (respectively).', diff1, diff2)
 
-        if len(document1) == 0 or len(document2) == 0:
+        if not len(document1) or not len(document2):
             logger.info(
-                "At least one of the documents had no words that werein the vocabulary. "
+                "At least one of the documents had no words that wherein the vocabulary. "
```
piskvorky: That sentence doesn't make sense. I think the intent here was `were in`.

gensim/models/keyedvectors.py
```diff
@@ -765,9 +764,9 @@ def wmdistance(self, document1, document2):
         if diff1 > 0 or diff2 > 0:
             logger.info('Removed %d and %d OOV words from document 1 and 2 (respectively).', diff1, diff2)
 
-        if len(document1) == 0 or len(document2) == 0:
+        if not len(document1) or not len(document2):
```
piskvorky: `not x:` is how you test for emptiness in Python.

gensim/models/keyedvectors.py
```diff
@@ -752,8 +752,7 @@ def wmdistance(self, document1, document2):
             If `pyemd <https://pypi.org/project/pyemd/>`_  isn't installed.
 
         """
-        if not PYEMD_EXT:
-            raise ImportError("Please install pyemd Python package to compute WMD.")
+        from pyemd import emd
```
piskvorky: How does this affect existing code that relied on `PYEMD_EXT`? Do we still even need `PYEMD_EXT`, where is it used?

gensim/models/keyedvectors.py
kishan3: Ok, so are intending to use `is` keyword here? like `len(document1) is 0` ?

gensim/models/keyedvectors.py
kishan3: we can remote `PYEMD_EXT` from wherever it is used in the project with proper imports. Is that what you mean here?

menshikh-iv: Yes. As I remember, this also used in tests, but this isn't imported from here, i.e. filled directly in test (I don't think that we need the same change in the tests, only here).

gensim/models/keyedvectors.py
menshikh-iv: not `is`, by `x` @piskvorky mean right side, like `not document1`, etc.

kishan3: Ok got you! Thanks @menshikh-iv

gensim/models/keyedvectors.py
kishan3: Ok sir, so please let me know what else to do. I have removed the import from the start now.

PR #2240, opened 2018-10-23
Fixed pyemd import error, pythonic equality and typo #2228

ramprakash-94:
1. Explicitly imported pyemd inside wmdistance function
2. Updated document length check to a pythonic way
3. Fixed typo mentioned in the issue


gensim/models/keyedvectors.py
```diff
@@ -752,8 +745,13 @@ def wmdistance(self, document1, document2):
...
         """
-        if not PYEMD_EXT:
-            raise ImportError("Please install pyemd Python package to compute WMD.")
+
+        # If pyemd C extension is available, import it.
+        # If pyemd is attempted to be used, but isn't installed, ImportError will be raised in wmdistance
+        try:
+            from pyemd import emd
+        except ImportError:
+            logger.error('Pyemd not installed')
```
piskvorky: Why this change from an exception to logging an error and continuing?

ramprakash-94: My bad. I moved this piece of code from the top to this position because of a [failure](https://travis-ci.org/RaRe-Technologies/gensim/jobs/445388889) in the previous commit. I'll fix this.

ramprakash-94:
I think it should work now!

ramprakash-94:
@piskvorky Can you please take a look at this now? Thanks.

PR #2242, opened 2018-10-26
Add documentation about ranges to scoring functions for `Phrases`

### LINKED ISSUE #2111: Phrases documentation for threshold argument is misleading
https://github.com/RaRe-Technologies/gensim/blob/37e49971efa74310b300468a5b3cf531319c6536/gensim/models/phrases.py#L252-L255

It feels to me like this should have said the opposite: "**Heavily** depends on concrete scoring-function" rather than "**Hardly** depends on concrete socring-function".

For example, if you choose npmi instead of the default, the threshold has to between -1 and 1, which makes the default (10.0) make no sense. The current documentation suggests that 10.0 will be OK.

jenishah:
Added return type in the documentation of the  scoring function, fix #2111 .

### DIFF
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -676,6 +676,11 @@ def original_scorer(worda_count, wordb_count, bigram_count, len_vocab, min_count
     corpus_word_count : int
         Not used in this particular scoring technique.
 
+    Returns
+    -------
+    float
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Thank you @jenishah :+1:

PR #2243, opened 2018-10-27
 Fix `SvmLightCorpus.serialize` if `labels` instance of numpy.ndarray 

aquatiko:
Current version of serialization support only lists, but this
adds support for any type of sequence.

Closes: https://github.com/RaRe-Technologies/gensim/issues/2113


aquatiko:
@menshikh-iv travis-ci is failing at flake8. But it don't seem to be due to my changes.

PR #2245, opened 2018-10-31
Set `metadata=True` for `make_wikicorpus` script by default

### LINKED ISSUE #1322: wiki examples: sample code to get from tfidf doc to wikipedia title/uri and vice versa
I very much like the LSI and LDA [wiki examples](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/src/wiki.rst), but one aspect that i think is missing is: how to get from tf-idf doc vectors (or later LSI / LDA vecs) back to Wikipedia URIs (or titles if easier) and vice versa?

Am i missing something obvious, or do i have to run another pass over the wiki dump, as the titles aren't saved anywhere?

I'll happily make a PR to extend the examples with this...

Xinyi2016:
1. Set `metadata=True` in the `make_wikicorpus` script
2. Added sample codes to get the title, tfidf vector and topics distribution of a given article in the wiki example.

Fix #1322


gensim/scripts/make_wikicorpus.py
```diff
@@ -14,11 +14,13 @@
 
 This actually creates three files:
```
horpto: Now and before it creates more than 3 files.

menshikh-iv: nice catch, fixed

menshikh-iv:
thanks for PR @Xinyi2016, congratz with the first contribution :1st_place_medal:

PR #2249, opened 2018-11-01
put the error message in comments

### LINKED ISSUE #2228: `pyemd` import check is misplaced

Naba7:
fixed #2228

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -751,9 +751,10 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
         ImportError
             If `pyemd <https://pypi.org/project/pyemd/>`_  isn't installed.
 
-        """
+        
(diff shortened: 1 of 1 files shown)


gojomo:
Why on earth would you want to do this? This looks like it introduces a bug, hiding a necessary error message from users missing a required installed library?

PR #2250, opened 2018-11-01
fixed typo #2228

Naba7:
commented the error


gojomo:
This also appears, at heart, to be attempting the same fix as #2247, but mixed with a lot of unrelated material. Closing in favor of that issue.

PR #2251, opened 2018-11-01
Fix check for callbacks

allenyllee:
Use `any()` method to check if there are any metric in `self.metrics` is a instance of `(DiffMetric, ConvergenceMetric)`


menshikh-iv:
Thank you @allenyllee, nice catch, congratz with the first contribution :1st_place_medal:

PR #2254, opened 2018-11-01
Fix extra list creation in get_max_id

horpto:


menshikh-iv:
Good, thank you @horpto, just interest: how did you find that?

PR #2256, opened 2018-11-02
Fix flake8 warnings W605, W504

horpto:
Since recent release of flake8 (v3.6.0, 24.10.18) all PR-s have not been building due to new warnings W605 W504 (pycodestyle - dependence of flake8, version 2.4.0).

Now, by edited pep 8 you should break the line before binary operator, not after. And avoid slash with followed non-special character like 'i', '*' or others in non-raw string.

Also, here we pin `flake8-rst == 0.4.3` until bug with E402 ("glue" rst) will be fixed.


piskvorky:
Can you fill in the PR description please?

The PR name and the codes are too cryptic: What is this PR for, why (motivation), what was the main approach for the fix(es)?

menshikh-iv:
Nice, thank you @horpto :+1:

PR #2259, opened 2018-11-06
Fixes divided by zero when w_star_count==0

allenyllee:
If w_star_count==0, it will throw exception of divided by zero


menshikh-iv:
Thanks for PR @allenyllee :+1:

gensim/topic_coherence/direct_confirmation_measure.py
```diff
@@ -73,6 +73,9 @@ def log_conditional_probability(segmented_topics, accumulator, with_std=False, w
                 m_lc_i = np.log(((co_occur_count / num_docs) + EPSILON) / (w_star_count / num_docs))
             except KeyError:
                 m_lc_i = 0.0
+            except ZeroDivisionError:
+                # if w_star_count==0, it will throw exception of divided by zero
```
piskvorky: The same is true for `num_docs`.

Can you add a comment why `m_lc_i = 0.0` is the correct solution for these divisions by zero?

PR #2261, opened 2018-11-08
Remove unnecessary creations of lists at all

horpto:


gensim/models/ldamodel.py
```diff
@@ -567,15 +567,18 @@ def init_dir_prior(self, prior, name):
         if isinstance(prior, six.string_types):
             if prior == 'symmetric':
                 logger.info("using symmetric %s at %s", name, 1.0 / self.num_topics)
-                init_prior = np.asarray([1.0 / self.num_topics for i in xrange(prior_shape)], dtype=self.dtype)
+                init_prior = np.fromiter((1.0 / self.num_topics for i in xrange(prior_shape)),
+                                            dtype=self.dtype, count=prior_shape)
```
piskvorky: Hanging indent please, here and elsewhere.

We avoid vertical indent because it leads to inconsistent visual indenting (like here), especially when the lines are edited in the future.

gensim/test/test_fasttext.py
```diff
@@ -674,7 +674,6 @@ def online_sanity(self, model):
                 terro.append(l)
             else:
                 others.append(l)
-        self.assertTrue(all(['terrorism' not in l for l in others]))
```
piskvorky: Why this change?

menshikh-iv:
ping @horpto, please fix @piskvorky comments and I'll merge current PR (don't forget to merge fresh `develop` first)

gensim/test/test_fasttext.py
horpto: I thought that this line does not check anything worthwhile. By code it is clear that this check is always true. It's not critical for me, so I return this line back.

gensim/models/ldamodel.py
horpto: Done.

gensim/models/ldamodel.py
```diff
@@ -567,15 +567,18 @@ def init_dir_prior(self, prior, name):
         if isinstance(prior, six.string_types):
             if prior == 'symmetric':
                 logger.info("using symmetric %s at %s", name, 1.0 / self.num_topics)
-                init_prior = np.asarray([1.0 / self.num_topics for i in xrange(prior_shape)], dtype=self.dtype)
+                init_prior = np.fromiter((1.0 / self.num_topics for i in xrange(prior_shape)),
+                                         dtype=self.dtype, count=prior_shape)
```
piskvorky: This is still vertical indent.

In gensim, we use hanging indent always, consistently: https://www.python.org/dev/peps/pep-0008/#id3

PR #2263, opened 2018-11-09
Optimize `remove_unreachable_nodes` in `gensim.summarization`

horpto:


gensim/summarization/commons.py
```diff
@@ -63,5 +63,5 @@ def remove_unreachable_nodes(graph):
     """
 
     for node in graph.nodes():
-        if sum(graph.edge_weight((node, other)) for other in graph.neighbors(node)) == 0:
+        if all(graph.edge_weight((node, other)) == 0 for other in graph.neighbors(node)):
```
horpto: Can we also delete null-weighted edges for reachable nodes?
@piskvorky  @menshikh-iv

piskvorky: A question for @fedelopez77, who is the author of this module :)

fedelopez77: @horpto I think that it would be reasonable, but test it first.
For the corner case that all edges are zero, according to (https://github.com/RaRe-Technologies/gensim/blob/1e11073196701adb6fabb8ccd5f7689db01dc38c/gensim/summarization/summarizer.py#L362-L363) this is executed before: 
https://github.com/RaRe-Technologies/gensim/blob/1e11073196701adb6fabb8ccd5f7689db01dc38c/gensim/summarization/summarizer.py#L105-L106

menshikh-iv:
For merging this PR, @horpto please add tests for case suggested by @fedelopez77 (and merge fresh `develop` too).

gensim/summarization/commons.py
horpto: Ok,  actually we don't need to delete null-weighted edges, as:

- we can not add such edge at all
- in pagerank we are passing by all edges anyway

As another note, I can notice that gensim summarizer has a big serious memory problems ([this](https://stackoverflow.com/questions/50567108/gensim-summarizer-throws-memoryerror-any-solution) and [this](https://stackoverflow.com/questions/50516951/python-extracting-text-and-summarize-multiple-file-types-in-directory-with-tik)).

horpto:
@menshikh-iv review, please

menshikh-iv:
Thanks @horpto :+1:

PR #2264, opened 2018-11-11
Improve `six` usage (`xrange`, `map, `zip`)

horpto:
Now the code contains both range and xrange. It's not consistent. Since
Python2 support will end soon, it's better to make the code more like in
Python3.


horpto:
@menshikh-iv  review, please

menshikh-iv:
Thank you @horpto :fire:

PR #2267, opened 2018-11-16
Optimize `mz_entropy` from `gensim.summarization`

horpto:
- add cache for __analytic_entropy
- optimise time of creation the word_count remaking the method of word counting


menshikh-iv:
Nice catch @horpto :star:

PR #2269, opened 2018-11-16
Refactor `line2doc` methods of `LowCorpus` and `MalletCorpus`

horpto:


gensim/corpora/lowcorpus.py
```diff
@@ -159,25 +159,24 @@ def line2doc(self, line):
...
-            # get all distinct terms in this document, ignore unknown words
-            uniq_words = set(words).intersection(iterkeys(self.word2id))
-
             # the following creates a unique list of words *in the same order*
             # as they were in the input. when iterating over the documents,
             # the (word, count) pairs will appear in the same order as they
             # were in the input (bar duplicates), which looks better.
             # if this was not needed, we might as well have used useWords = set(words)
-            use_words, marker = [], set()
+            use_words, counts = [], Counter()
```
menshikh-iv: Please add a test for this method (to be fully sure that this produce exactly same output as before)

PR #2273, opened 2018-11-21
set normed vectors to None when model is trained

### LINKED ISSUE #2260: doc2vec/word2vec/fasttext models do not appear to improve if similarities checked mid-training epochs
#### Description
I am training a doc2vec model on a large corpus. I need to observe the model for more detailed statistics for my supervisor/boss.
The problem is similar to the problem below where I just slightly modified the [Doc2Vec Tutorial on the Lee Dataset](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/doc2vec-lee.ipynb). The model does not improve its recommendations for the` most_similar method`.

```python
import gensim
import os
import smart_open
import gensim.models.callbacks


# Set file names for train and test data
test_data_dir = '{}'.format(os.sep).join([gensim.__path__[0], 'test', 'test_data'])
lee_train_file = test_data_dir + os.sep + 'lee_background.cor'
lee_test_file = test_data_dir + os.sep + 'lee.cor'


def read_corpus(fname, tokens_only=False):
    with smart_open.smart_open(fname, encoding="iso-8859-1") as f:
        for i, line in enumerate(f):
            if tokens_only:
                yield gensim.utils.simple_preprocess(line)
            else:
                # For training data, add tags
                yield gensim.models.doc2vec.TaggedDocument(gensim.utils.simple_preprocess(line), [i])


train_corpus = list(read_corpus(lee_train_file))
test_corpus = list(read_corpus(lee_test_file, tokens_only=True))

results_new = {i: None for i, doc in enumerate(train_corpus)}
results_old = results_new.copy()


class TrainProgressEvaluation(gensim.models.callbacks.CallbackAny2Vec):

    def __init__(self, test_set, results_new, results_old):
        self.test_set = test_set
        self.results_new = results_new
        self.results_old = results_old
        self.epoch = 0

    def on_epoch_end(self, model):
        self.epoch += 1
        print(f"epoch {self.epoch} end")

    def on_batch_begin(self, model):
        for num, sample in enumerate(self.test_set):
            recs = model.docvecs.most_similar(num)
            # for the first call results_new[num] is None
            self.results_old[num] = results_new[num] or recs
            self.results_new[num] = recs
            for i in range(len(recs)):
                if not self.results_old[num][i][0] == self.results_new[num][i][0] or not self.results_old[num][i][1] == self.results_new[num][i][1]:
                    print(f"Sample {num} has changed.")
                    print(f"Old tag {self.results_old[num][i][0]}. New tag {self.results_new[num][i][0]}")
                    print(f"Old distance {self.results_old[num][i][1]}. New distance {self.results_new[num][i][1]}")


model = gensim.models.doc2vec.Doc2Vec(vector_size=50, min_count=2, epochs=40, workers=4)
model.build_vocab(train_corpus)
model.train(train_corpus, total_examples=model.corpus_count, epochs=model.epochs,
            callbacks=(TrainProgressEvaluation(train_corpus, results_new, results_old),))


```

#### Expected Results
I expect to see many improvements in either recommendation or distance.

#### Actual Results
Consol Output with four workers:
It surprises me that only the first sample in the training_corpus receives some updates. I don't understand it.

```
Sample 0 has changed.
/usr/local/lib/python3.7/site-packages/gensim/matutils.py:737: FutureWarning: Conversion of the second argument of issubdtype from `int` to `np.signedinteger` is deprecated. In future, it will be treated as `np.int64 == np.dtype(int).type`.
Old tag 116. New tag 30
  if np.issubdtype(vec.dtype, np.int):
Old distance 0.5557072162628174. New distance 0.4648822546005249
Sample 0 has changed.
Old tag 42. New tag 224
Old distance 0.48946425318717957. New distance 0.3621359169483185
Sample 0 has changed.
Old tag 51. New tag 96
Old distance 0.4082771837711334. New distance 0.31921446323394775
Sample 0 has changed.
Old tag 90. New tag 77
Old distance 0.3731566369533539. New distance 0.3184990882873535
Sample 0 has changed.
Old tag 128. New tag 45
Old distance 0.34601616859436035. New distance 0.30474674701690674Sample 0 has changed.
Old tag 30. New tag 116
Old distance 0.4648822546005249. New distance 0.5557072162628174
Sample 0 has changed.
Old tag 224. New tag 42
Old distance 0.3621359169483185. New distance 0.48946425318717957
Sample 0 has changed.
Old tag 96. New tag 51
Old distance 0.31921446323394775. New distance 0.4082771837711334
Sample 0 has changed.
Old tag 77. New tag 90
Sample 0 has changed.
Old tag 46. New tag 234
Old distance 0.3005654215812683. New distance 0.3441653251647949
Old distance 0.3184990882873535. New distance 0.3731566369533539
Sample 0 has changed.
Old tag 45. New tag 128
Old distance 0.30474674701690674. New distance 0.34601616859436035
Sample 0 has changed.
Old tag 46. New tag 234
Old distance 0.3005654215812683. New distance 0.3441653251647949
Sample 0 has changed.
Sample 0 has changed.
Old tag 111. New tag 76
Old tag 111. New tag 76
Old distance 0.280322402715683. New distance 0.32334667444229126
Old distance 0.280322402715683. New distance 0.32334667444229126
Sample 0 has changed.
Old tag 221. New tag 49

Old distance 0.2779023051261902. New distance 0.27320006489753723
Sample 0 has changed.
Old tag 52. New tag 4
Old distance 0.27472415566444397. New distance 0.27205419540405273
Sample 0 has changed.
Old tag 221. New tag 49
Old distance 0.2779023051261902. New distance 0.27320006489753723
Sample 0 has changed.
Sample 0 has changed.
Old tag 205. New tag 149
Old distance 0.26930660009384155. New distance 0.2699446976184845
Old tag 52. New tag 4
Old distance 0.27472415566444397. New distance 0.27205419540405273
Sample 0 has changed.
Old tag 205. New tag 149
Old distance 0.26930660009384155. New distance 0.2699446976184845
Sample 0 has changed.
Old tag 116. New tag 30
Old distance 0.5557072162628174. New distance 0.4648822546005249
Sample 0 has changed.
Old tag 42. New tag 224
Old distance 0.48946425318717957. New distance 0.3621359169483185
Sample 0 has changed.
Old tag 51. New tag 96
Old distance 0.4082771837711334. New distance 0.31921446323394775
Sample 0 has changed.
Old tag 90. New tag 77
Old distance 0.3731566369533539. New distance 0.3184990882873535
Sample 0 has changed.
Old tag 128. New tag 45
Old distance 0.34601616859436035. New distance 0.30474674701690674
Sample 0 has changed.
Old tag 234. New tag 46
Old distance 0.3441653251647949. New distance 0.3005654215812683
Sample 0 has changed.
Old tag 76. New tag 111
Old distance 0.32334667444229126. New distance 0.280322402715683
Sample 0 has changed.
Old tag 49. New tag 221
Old distance 0.27320006489753723. New distance 0.2779023051261902
Sample 0 has changed.
Old tag 4. New tag 52
Old distance 0.27205419540405273. New distance 0.27472415566444397
Sample 0 has changed.
Old tag 149. New tag 205
Old distance 0.2699446976184845. New distance 0.26930660009384155
epoch 1 end
epoch 1 end
epoch 2 end
epoch 3 end
epoch 4 end
....
```

So I debug the model and there are no improvements anymore:
```
/usr/local/lib/python3.7/site-packages/gensim/matutils.py:737: FutureWarning: Conversion of the second argument of issubdtype from `int` to `np.signedinteger` is deprecated. In future, it will be treated as `np.int64 == np.dtype(int).type`.
  if np.issubdtype(vec.dtype, np.int):
epoch 1 end
epoch 2 end
....
```

I try it with 1 worker only:
```
/usr/local/lib/python3.7/site-packages/gensim/matutils.py:737: FutureWarning: Conversion of the second argument of issubdtype from `int` to `np.signedinteger` is deprecated. In future, it will be treated as `np.int64 == np.dtype(int).type`.
  if np.issubdtype(vec.dtype, np.int):
epoch 1 end
epoch 2 end
epoch 3 end
....
```

What's happening here and how can I see during training how my doc2vec model improves? Because it is also not possible to see the training_error for doc2vec [#999](https://github.com/RaRe-Technologies/gensim/issues/999).
Further experimenting reveals that docvecs.vectors_docs are of course updated between each call of `batch_end`. But `most_similiar` always returns the same suggestion.

#### Versions
Darwin-17.5.0-x86_64-i386-64bit
Python 3.7.0 (default, Jun 29 2018, 20:13:13) 
[Clang 9.1.0 (clang-902.0.39.2)]
NumPy 1.15.0
SciPy 1.1.0
gensim 3.5.0
FAST_VERSION 0

timbicker:
Fixes #2260 

allows `most_similar()` to be called based on the newly trained vectors during multiple training iterations.


gojomo:
Makes sense to me. For completeness, the same thing should happen on Word2Vec & related `train()`s. 

To be an ideal PR, there'd also be a test that fails before the fix(es) are applied. 

I could also imagine people who've done an explicit `init_sims(replace=True)`, then trying to re-train, getting odd or bad results from that training (since they've discarded the original raw magnitudes). This sort of re-set could also catch such a condition – by testing if the normed and non-normed properties are identical objects – and issue a (non-failing because it still might kinda work) warning in such a case.

timbicker:
@gojomo thank you for your reply. I added two tests, one each for doc2vec and word2vec. I also wanted to add a test for fast text, but it turns out that there is some error when the model is trained more than once. There is also an issue #2139 for this. Should I still implement the test?

Further, word2vec overrides `_clear_post_train` and sets `self.wv.vectors_norm = None`. So the functionality is already included here. I therefore structured the resetting of the vectors in doc2vec analogously.

gensim/models/doc2vec.py
```diff
@@ -649,6 +649,7 @@ def clear_sims(self):
         """Resets the current word vectors. """
         self.wv.vectors_norm = None
         self.wv.vectors_docs_norm = None
+        self.docvecs.vectors_docs_norm = None
```
menshikh-iv: can you please check and verify that "similar" thing happens in w2v\fasttext models too?

gensim/test/test_doc2vec.py
```diff
@@ -687,6 +687,22 @@ def testTrainWarning(self, l):
...
+
+    def test_most_similiar_uses_newly_trained_vectors(self):
+        """Test if the doc2vec model deletes the normalized vectors that are used by most similar
+        when it is retrained"""
+        corpus = DocsLeeCorpus()
+        model = doc2vec.Doc2Vec(vector_size=10, min_count=1, workers=1)
+        model.build_vocab(corpus)
+
+        model.train(corpus, total_examples=model.corpus_count, epochs=1, start_alpha=0.1, end_alpha=0.1)
+        model.docvecs.most_similar(1)
```
menshikh-iv: it's better to test in the same manner (i.e. using callbacks) as mentioned in https://github.com/RaRe-Technologies/gensim/issues/2260#issue-377788511 (i.e. call most_similar "during training", not after). Same things for Word2Vec and FastText

menshikh-iv:
Abandoned, closing.

timbicker:
@menshikh-iv Hi sorry, I was busy in the last weeks. I will continue in the next days working on this PR, if it is still possible.

menshikh-iv:
@timbicker of course, feel free to re-open current PR and continue when you'll have time!

gensim/models/doc2vec.py
timbicker: Yes. If you look in `word2vec.py` you see that there are actually two functions `_clear_post_train` and `clear_sims` which do the same thing (one of the functions is therefore redundant): Setting `self.wv.vectors_norm = None`.
The same thing (setting `self.wv.vectors_norm = None` and more) also happens in `fasttext.py` in function `_clear_post_train`.
Does this solve your concerns or do you want a more detailed explanation?

gensim/test/test_doc2vec.py
timbicker: I see your point. The problem is that above mentioned function `_clear_post_train`, which is responsible to reset the previously trained docvecs, is only called after the whole training is done. This means that calls of `most_similar` inside a callback, won't be able to access the newly trained vectors.
One solution would be to add a `replace=False` to `most_similiar` which is further passed down to `init_sims`. The user then has to change `replace` to `True` in order to be able to evaluate the newly calculated dimensions during training.
Further, this modification can be done without changing the current behavior.

timbicker:
@menshikh-iv Hi, I just updated the PR but I can't reopen it. Can you reopen it for me?
Further, I added a `replace=False` to `most_similiar` which is further passed down to `init_sims`. The user then has to change `replace` to `True` in order to be able to evaluate the newly calculated vectors during training. This modification is imo necessary, because `_clear_post_train` is only called after all training is done, and therefore would not allow the user to use newly calculated keyedvectors in callbacks.

menshikh-iv:
@timbicker reopened, feel free to continue :+1:

gojomo:
Adding a `replace` to `most_similar()` seems unnecessarily complex, and inappropriate for a method that would otherwise have no side-effects on the model. 

Why not just clear the normed-vectors the moment any action that could change the raw vectors is initiated? (Pre-any-training-epoch, not post-any-training-epoch?) 

If some really-unique pattern of usage needs them cleared other times, that code can and should explicitly call `init_sims()` itself, rather than layering extra possibly-destructive side-effects into `most_similar()`.

timbicker:
Unfortunately `FastText` has problems with this implementation. 
```
  File "lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "lib/python3.5/threading.py", line 862, in run
  File "gensim/gensim/models/base_any2vec.py", line 211, in _worker_loop
    tally, raw_tally = self._do_train_job(data_iterable, job_parameters, thread_private_mem)
  File "gensim/gensim/models/fasttext.py", line 610, in _do_train_job
    tally += train_batch_cbow(self, sentences, alpha, work, neu1)
  File "gensim/models/fasttext_inner.pyx", line 431, in gensim.models.fasttext_inner.train_batch_cbow
TypeError: 'NoneType' object is not subscriptable
```
I don't know how to fix this problem, because I am not familiar with fasttext...
A workaround would be, to check if `self` is of class `FastText` in `train()` and then call `clear_keyed_vector_internals` (former `_clear_post_train`) accordingly.
Can you help me out here? What are your thoughts?

timbicker:
I solved the problem with two functions `_clear_epoch_start` and `_clear_post_train`. Both functions do nothing in their original parent implementation and can be overridden accordingly by the child functions. This allows `Doc2Vec` and `Word2Vec` to implement their functionality independent of `FastText`

gensim/models/doc2vec.py
timbicker: @menshikh-iv I am not sure, whether I should resolve the conversation or wait for a response by you. What is preferred here (or in general in such cases)?

gojomo:
Personally still see this as a little over-wrought for the purpose:

* it clears *both* on epoch-start and training-finish
* (new-method) `_clear_epoch_start()` only ever calls `clear_sims()` in subclasses, and is empty in superclass, adding more twisty indirection

I'd prefer the more-simple (even if slightly more error-prone) approach of: (1) always clear at the end of `train()`; (2) warn via docs the (fewer, particular, plausibly-expert) users of the advanced callbacks functionality that *if* they're doing the kinds of operations that cache model results, they need to `clear_sims()` to get updated results.

gensim/models/doc2vec.py
gojomo: Separately, the line above this (which existed prior to your edits) – `self.wv.vectors_docs_norm = None` – looks fishy. A normal `wv` set of word-vectors wouldn't be expected to even have a `vectors_docs_norm` property, unless some other code has mistakenly poked doc-vectors into there.

timbicker:
@gojomo thanks for your reply. Unfortunately, I won't have time anymore to give this PR the attention it deserves. I would appreciate it if you could set the flag to `help wanted` or something you think appropriate.
Maybe @Naba7 wants to finish?

m4nuC:
I am using FastText and I've stumbled on this issue during training. 
I've tried setting the norms to `None` but that doesn't seem to help.
 I've also tried  `model.wv.clear_keyed_vector_internals` or `model.clear_keyed_vector_internals`  ? But neither method seem to exist.

Are there any other workaround that I should use ?

gensim/models/fasttext.py
```diff
@@ -745,7 +741,10 @@ def clear_sims(self):
         You can recompute them later again using the :meth:`~gensim.models.fasttext.FastText.init_sims` method.
 
         """
-        self._clear_post_train()
+        self.wv.vectors_norm = None
+        self.wv.vectors_vocab_norm = None
+        self.wv.vectors_ngrams_norm = None
+        self.wv.buckets_word = None
```
gojomo: This is fishy: nothing about training (as opposed to other vocab-discovery steps) would change the `buckets_word` mapping of words to ngram-locations. But also: (1) `init_sims()` doesn't rebuild this `buckets_word` as implied by the comment; (2) unless it's a big performance win in some usual case, I'm not sure why `buckets_word` needs to exist at all – I thought known-words already had presummed vectors (requiring no ngram lookups), while for any OOV word, it's not very expensive to find the ngrams/ngram-bucket-indexes each time needed; (3) relatedly, it looks like the only use of `vectors_ngrams_norm` is nonsensical, so that property should be discarded entirely - but I've made a separate issue, #2667, for that.

mpenkov:
I think this is superseeded by #2698. Closing.

PR #2274, opened 2018-11-21
Reuse precalculated updated prior in `ldamodel.update_dir_prior`

horpto:


gensim/models/ldamodel.py
```diff
@@ -145,11 +145,11 @@ def update_dir_prior(prior, N, logphat, rho):
...
 
-    if all(rho * dprior + prior > 0):
-        prior += rho * dprior
+    updated_prior = rho * dprior + prior
+    if all(updated_prior > 0):
+        prior = updated_prior
     else:
-        logger.warning("updated prior not positive")
-
+        logger.warning("updated prior is not positive")
```
piskvorky: Unrelated to this PR, but: shouldn't this be an assert instead? What does such warning mean, what should the user do as a result?

This silent covering of a failure (`prior` is not actually updated, despite the function name) is unPythonic.

PR #2275, opened 2018-11-22
Refactor `BM25`

horpto:
- remove unnecessary attributes
- move calculation of average_idf in _initialize
- more readable names


gensim/summarization/bm25.py
```diff
@@ -80,33 +76,47 @@ def __init__(self, corpus):
...
+        # collect words with negative idf to set them a special epsilon value.
+        # idf can be negative if word is contained in more than half of documents
+        negative_idfs = []
+        for word, freq in iteritems(nd):
+            idf = math.log(self.corpus_size - freq + 0.5) - math.log(freq + 0.5)
+            self.idf[word] = idf
+            idf_sum += idf
+            if idf < 0:
+                negative_idfs.append(word)
+        self.average_idf = idf_sum / len(self.idf)
```
piskvorky: Use float (or explicitly cast to float). Both to avoid potential Python errors in integer vs float division, as well as to make the intent clear to readers of the code.

gensim/summarization/bm25.py
```diff
@@ -80,33 +76,47 @@ def __init__(self, corpus):
...
-        for word, freq in iteritems(self.df):
-            self.idf[word] = math.log(self.corpus_size - freq + 0.5) - math.log(freq + 0.5)
-
-    def get_score(self, document, index, average_idf):
+                if word not in nd:
+                    nd[word] = 0
+                nd[word] += 1
+
+        # collect idf sum to calculate an average idf for epsilon value
+        idf_sum = 0
```
piskvorky: Safer to use float (or explicitly cast to float below).

Both to avoid potential Python errors in integer vs float division in future refactorings, as well as (mainly) to make the intent clear to readers.

horpto: oh, yes. you're right. Damn python2.

PR #2102, opened 2018-06-25
restyle API Ref

piskvorky:
Render a nicer API reference page. Continues from #1809.

I had trouble merging `develop` into this PR -- there were files with trailing whitespace (`*_inner.c`), and even two files with Windows line endings: `lsi_dispatcher.py`, `lsi_worker.py`, which prevented me from committing the resolved conflicts.

I solved the conflicts by removing trailing whitespace, converting line endings to Unix and then finishing the merge commit.

I also fixed the content of the documentation of the distributed LSI and LDA files, since it used incorrect docstring style and introduced some minor documentation bugs, in PR #1912 (specifically commit 1611f3a477686bf9f462d0ed09b1daff2a58f09e).

PR #2281, opened 2018-11-30
Group and sort API reference

horpto:
Current API ref is awful. All is in one big heap and not sorted by anyhow. 
+ I've found that ftmodel is not specified in the API list.

### DIFF
--- a/docs/src/apiref.rst
+++ b/docs/src/apiref.rst
@@ -3,16 +3,26 @@
 API Reference
 =============
 
-Modules:
+.. comment: !!! keep TOC tree sorted
+
+Common modules:
 
 .. toctree::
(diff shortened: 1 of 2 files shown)


menshikh-iv:
@horpto 
1. fasttext wrapper shouldn't be in doc (because we have our own implementation, this wrapper is deprecated and will be removed in 4.0.0)
2. From PoV of "grouping" all stuff, we need to cleanup & finish #2102 (here we re-group all stuff, use plugins for auto-generate "rst boilerplates" and so on)

It will be pretty nice if you'll work on #2102 instead.

horpto:
I'm closing this PR  until issue #2102 will not be closed, then I'll reopen it.

PR #2282, opened 2018-12-03
EnsembleLda

sezanzeb:
**Tests**

~~We have some tests that we can also commit when this PR is of interest, I will polish them and embed them into the unittest module in that case.~~

They are commited.

**Materials**

As requested by @piskvorky, here are the materials that describe our module: 
[eLDA_algo_overview.pdf](https://github.com/RaRe-Technologies/gensim/files/2639134/eLDA_algo_overview.pdf)
[eLDA_motivation.pdf](https://github.com/RaRe-Technologies/gensim/files/2639135/eLDA_motivation.pdf)
[eLDA_when_to_use.pdf](https://github.com/RaRe-Technologies/gensim/files/2639136/eLDA_when_to_use.pdf)

~~**Code Review**~~

~~There are two places in "gensim/models/ensemblelda.py" for which I especially would like to have review:~~
- ~~lines 611 and 784: The error handling of multiprocessing~~
- ~~line 1124 and following: The way I provide functions crucial to gensims ldamodel api. Those functions just forward the calls to an internal object. So are they too redundant or do you think they make things easier? Is there a better way to do so?~~
- ~~And another one for the "tox -e flake8" test, which does not pass here:~~
```
gensim/corpora/opinosiscorpus.py:78:22: W605 invalid escape sequence '\w'
                    processed = [
```
- ~~would you prefer the implemention not to use pandas, since it has not been a dependency of gensim before?~~


menshikh-iv:
CC @piskvorky, wdyt about this feature (not PR itself), is it a good idea to add it to gensim?

menshikh-iv:
ping @piskvorky

piskvorky:
Definitely! Getting more robust LDA topics sounds like an awesome feature (and the accompanying descriptions/motivations are exemplary – we can point to this in future PRs).

I didn't have time to review the PR code, but yes, we prefer to avoid pandas (complex dependency, slow, unclear memory semantics). Cheers.

SophiaGoldberg:
Minor fix: when running `gensim/docs/notebooks/Opinosis.ipynb` we require that `gensim/gensim/corpora/__init__.py` contains the line of code: `from .opinosiscorpus import OpinosisCorpus`.

sezanzeb:
> Definitely! Getting more robust LDA topics sounds like an awesome feature (and the accompanying descriptions/motivations are exemplary – we can point to this in future PRs).
> 
> I didn't have time to review the PR code, but yes, we prefer to avoid pandas (complex dependency, unclear memory semantics, encourages bad engineering patterns). Can you use normal list/dict comprehensions instead? Cheers.

I'm working on it and will let you know when I'm done.

sezanzeb:
Tests work in the following environments:

4.19.32-1-MANJARO
- python 2.7.16
- python 3.7.2

Docker Image with Ubuntu 18.04
- python 2.7.15rc1
- python 3.6.7

I don't see enything in the CBDBSCAN class that would not be deterministic and that could be different in the online testing environment.

Still, results after clustering are different from the local machine.

If those tests work there is just one idea I had. Instead of using topic_model_kind="lda" you could import ldamodel and use topic_model_kind=ldamodel, maybe that adds more flexibility. But both can be realized at the same time basically.

sezanzeb:
@piskvorky @menshikh-iv 

the python 2.7 test of appveyor fails on this particular problem:

`backports/lzma/_lzmamodule.c(115) : fatal error C1083: Cannot open include file: 'lzma.h': No such file or directory` (https://ci.appveyor.com/project/piskvorky/gensim-431bq/builds/23839139/job/mxjwisoy50785fo2#L202)

Is this actually related to the package? Or is the environment broken?

Other than that all tests pass :). I am waiting for feedback now.

sezanzeb:
All tests pass now

piskvorky:
Thanks @sezanzeb . I'll try to review more in-depth and get back to you as soon as I can.

aloosley:
Thanks @piskvorky for reviewing our idea and @sezanzeb contribution to gensim.  

In case a audio-visual introduction of eLDA could help with any part of thinking about the code, I wanted to also share this link to a 22 minute eLDA overview presented this year at ML Prague:

https://slideslive.com/38913528/solving-the-text-labeling-challenge-with-ensemblelda-and-active-learning

lmcinnes:
@sezanzeb I am working on a similar idea using pLSA and a slightly different ensembling/clustering approach, but would very much like to reference this work as one of the inspirations. Do you have a citable reference for this, or is one potentially forthcoming? Thanks.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+                "ldamulticore": ldamulticore.LdaMulticore
+            }[topic_model_kind]
+        else:
+            self.topic_model_kind = topic_model_kind
+
+        # Store some of the parameters
+        self.num_models = num_models
+        self.gensim_kw_args = gensim_kw_args
+
+        # some settings affecting performance
```
mpenkov: As above. These comments may be more helpful in the docstring for the parameters.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+        self.tms = []
+        # initialize empty topic term distribution array
+        self.ttda = np.empty((0, len(gensim_kw_args["id2word"])))
+        self.asymmetric_distance_matrix_outdated = True
+
+        # in case the model will not train due to some
+        # parameters, stop here and don't train.
+        if num_models <= 0:
+            return
+        if ("corpus" not in gensim_kw_args):
```
mpenkov:
```suggestion
        if "corpus" not in gensim_kw_args:
```

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+            return
+        if gensim_kw_args["corpus"] is None:
```
mpenkov: You can combine this with the above conditional without losing any readibility.

```suggestion
        if gensim_kw_args.get("corpus") is None:
```

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+            return
+        if "iterations" in gensim_kw_args and gensim_kw_args["iterations"] <= 0:
```
mpenkov: Simpler:

```suggestion
        if gensim_kw_args.get("iterations", 0) <= 0:
```

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+            return
+        if "passes" in gensim_kw_args and gensim_kw_args["passes"] <= 0:
+            return
+
+        logger.info("Generating {} topic models...".format(num_models))
+        # training
```
mpenkov: Please avoid one-liner comments that add nothing to the code.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+        # create model that can provide the usual gensim api to the stable topics from the ensemble
+        self.generate_gensim_representation()
+
+    def convert_to_memory_friendly(self):
+        """removes the stored gensim models and only keeps their ttdas"""
+        self.tms = []
+        self.memory_friendly_ttda = True
+
+    def generate_gensim_representation(self):
+        """creates a gensim-model from the stable topics
```
mpenkov:
```suggestion
        """Creates a gensim model from the stable topics
```

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+        The prior for the gensim model, eta, Can be anything,
```
mpenkov:
```suggestion
        The prior for the gensim model, eta, can be anything,
```

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+    def asymmetric_distance_matrix_worker(self, worker_id, ttdas_sent, n_ttdas, pipe, threshold, method):
+        """ worker, that computes the distance to all other nodes
+        from a chunk of nodes. https://stackoverflow.com/a/1743350
+        """
+
+        logger.info("Spawned worker to generate {} rows of the asymmetric distance matrix".format(n_ttdas))
+        # the chunk of ttda that's going to be calculated:
+        ttda1 = self.ttda[ttdas_sent:ttdas_sent + n_ttdas]
+        distance_chunk = self.calculate_asymmetric_distance_matrix_chunk(ttda1=ttda1, ttda2=self.ttda,
```
mpenkov: This indentation is a bit ugly. Please do something like:

```python
kwargs = dict(ttda1=ttda1, ttda2=...)
distance_chunk = self.calculate_asymmetric_distance_matrix_chunk(**kwargs)
```

Also, this is a sign that your method lines are perhaps a little long. Consider shortening them.

Finally, think about scoping your methods. Are these methods public or _internal?

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+        self.asymmetric_distance_matrix_outdated = False
+
+        if threshold is None:
+            threshold = {"mass": 0.95, "rank": 0.11}[method]
+
+        logger.info("Generating a {} x {} asymmetric distance matrix...".format(len(self.ttda), len(self.ttda)))
+
+        # singlecore:
+        if workers is not None and workers <= 1:
+            self.asymmetric_distance_matrix = self.calculate_asymmetric_distance_matrix_chunk(ttda1=self.ttda,
```
mpenkov: Please fix this indentation as described above.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+                                                                                              ttda2=self.ttda,
+                                                                                              threshold=threshold,
+                                                                                              start_index=0,
+                                                                                              method=method)
+            return self.asymmetric_distance_matrix
+
+        # multicore:
```
mpenkov: Please expand this comment or remove it. In it's present shape, it's not helpful.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+            if self.memory_friendly_ttda:
+                # remember that this code is inside the worker processes memory,
+                # so self.ttda is the ttda of only a chunk of models
+                pipe.send(self.ttda)
+            else:
+                pipe.send(self.tms)
+
+            pipe.close()
+
+    def asymmetric_distance_matrix_worker(self, worker_id, ttdas_sent, n_ttdas, pipe, threshold, method):
```
mpenkov: Does this really need to be a method? It looks like the only time it uses `self` is to access `self.ttda`. You can just make this into a pure function and pass `ttda` directly. There are several benefits:

- Easier to read/write the code (fewer indentation levels)
- Easier to test the code. You don't need to create any objects (to call the method on), you just pass the parameters directly.
- Simpler classes

Also, consider naming your pure function more tersely. If this is an internal function (only used within the module), then you can be super terse, e.g. _matrix_worker.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+        create_mask = {"mass": mass_masking, "rank": rank_masking}[method]
+
+        # some help to find a better threshold by useful log messages
+        avg_mask_size = 0
+
+        # initialize the distance matrix. ndarray is faster than zeros
+        distances = np.ndarray((len(ttda1), len(ttda2)))
+
+        # now iterate over each topic
```
mpenkov: Please consider rewriting your code comments to be more helpful. Comments should explain **why** you're doing something, not **what**, because what you are doing should already be obvious from the code.

If you find yourself writing code where it's **not** obvious what you're doing, then instead of adding comments, please consider making the code easier to read by refactoring, using more helpful variable names, extracting code blocks to functions, etc.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+        for i in range(len(ttda1)):
+
+            # create mask from a, that removes noise from a and keeps the largest terms
+            a = ttda1[i]
```
mpenkov: Please use more helpful variable names (here and everywhere). What is `a`?

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+        create_mask = {"mass": mass_masking, "rank": rank_masking}[method]
+
+        # some help to find a better threshold by useful log messages
+        avg_mask_size = 0
+
+        # initialize the distance matrix. ndarray is faster than zeros
+        distances = np.ndarray((len(ttda1), len(ttda2)))
+
+        # now iterate over each topic
+        for i in range(len(ttda1)):
```
mpenkov: Please use built-in enumerate function:

```suggestion
        for i, something in enumerate(ttda1):
```

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+            # create mask from a, that removes noise from a and keeps the largest terms
+            a = ttda1[i]
+            mask = create_mask(a)
+            a_masked = a[mask]
+
+            avg_mask_size += mask.sum()
+
+            # now look at every possible pair for topic a:
+            for j in range(len(ttda2)):
```
mpenkov: Please look at the itertools module as an alternative to using nested for-loops.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+                    continue
+
+                # now mask b based on a, which will force the shape of a onto b
+                b_masked = ttda2[j][mask]
+
+                distance = 0
+                # is the masked b just some empty stuff? Then forget about it, no similarity, distance is 1
+                # (not 2 because that correspondsto negative values. The maximum distance is 1 here)
+                # don't normalize b_masked, otherwise the following threshold will never work:
+                if b_masked.sum() <= 0.05:
```
mpenkov: Please move this magic number to an _INTERNAL_CONSTANT.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+                    # if there is indeed some non-noise stuff in the masked topic b, look at
+                    # how similar the two topics are. note that normalizing is not needed for cosine distance,
+                    # as it only looks at the angle between vectors
+                    distance = cosine(a_masked, b_masked)
+
+                distances[i][j] = distance
+
+        avg_mask_size = avg_mask_size / ttda1.shape[0] / ttda1.shape[1]
+        percent = round(100 * avg_mask_size, 1)
+        if avg_mask_size > 0.75:
```
mpenkov: Please get rid of this magic number.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+        """see https://radimrehurek.com/gensim/models/ldamodel.html"""
+        self.has_gensim_representation()
+        return self.classic_model_representation.print_topics(*posargs, **kwArgs)
+
+    @property
+    def id2word(self):
+        return self.gensim_kw_args["id2word"]
+
+
+class CBDBSCAN():
```
mpenkov: Please add a docstring.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+        if min_samples is None:
+            min_samples = int(self.num_models / 2)
+
+        logger.info("Fitting the clustering model")
+
+        self.cluster_model = CBDBSCAN(eps=eps, min_samples=min_samples)
+        self.cluster_model.fit(self.asymmetric_distance_matrix)
+
+    def generate_stable_topics(self, min_cores=None):
```
mpenkov: This method is far too long. Please consider breaking it up somehow.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+            # 3. checking for "easy_valid"s
+            # checks if the core has at least min_cores of cores with the only label as itself
+            if cluster["amount_parent_labels"] >= 1:
+                if sum(map(lambda x: x == {label}, cluster["parent_labels"])) >= min_cores:
+                    cluster["is_valid"] = True
+
+        # Reaplying the rule 3
+        # this happens when we have a close relationship among 2 or more clusters
+        for cluster in [cluster for cluster in sorted_clusters if np.isnan(cluster["is_valid"])]:
```
mpenkov: Avoid unnecessarily overloading variable names:

```suggestion
        for cluster in [c for c in sorted_clusters if np.isnan(c["is_valid"])]:
```

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+                    if label in a["parent_labels"]:
+                        a["parent_labels"].remove(label)
+
+        # list of all the label numbers that are valid
+        valid_labels = np.array([cluster["label"] for cluster in sorted_clusters if cluster["is_valid"]])
+
+        for topic in results:
+            topic["valid_parents"] = {label for label in topic["parent_labels"] if label in valid_labels}
+
+        def validate_core(core):
```
mpenkov: Can this be a module-scope _internal function? If yes, then move it out. The outer method is already far too long.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+        if min_cores == 0:
+            min_cores = 1
+
+        if min_cores is None:
+            # min_cores is a number between 1 and 3, depending on the number of models
+            min_cores = min(3, max(1, int(self.num_models / 4 + 1)))
+
+        results = self.cluster_model.results
+
+        # first, group all the learned cores based on the label,
```
mpenkov: This code block could be a separate function: _group_by_label.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+                # counts how many different labels a core has as parents
+                topic["amount_parent_labels"] = len(topic["parent_labels"])
+
+                label = topic["label"]
+                if label not in grouped_by_labels:
+                    grouped_by_labels[label] = []
+                grouped_by_labels[label].append(topic)
+
+        # then aggregate their amount_parent_labels by maxing and parent_labels by concatenating
```
mpenkov: This code block could be another separate function: e.g. _aggregate_grouped_labels.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+        # APPLYING THE RULES 1 to 3
+        # iterate over the cluster labels, see which clusters/labels
+        # are valid to cause the creation of a stable topic
+        for i, cluster in enumerate(sorted_clusters):
+            label = cluster["label"]
+            # label is iterating over 0, 1, 3, ...
+
+            # 1. rule - remove if the cores in the cluster have no parents
+            if cluster["amount_parent_labels"] == 0:
+                # label is NOT VALID
```
mpenkov: Useless comment

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+            # 2. rule - remove if it has less than min_cores as parents
+            # (parents are always also cores)
+            if len(cluster["parent_labels"]) < min_cores:
+                cluster["is_valid"] = False
+                remove_from_all_sets(label)
+
+            # 3. checking for "easy_valid"s
+            # checks if the core has at least min_cores of cores with the only label as itself
+            if cluster["amount_parent_labels"] >= 1:
```
mpenkov: Shouldn't there be an else after this conditional?

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+        # APPLYING THE RULES 1 to 3
+        # iterate over the cluster labels, see which clusters/labels
+        # are valid to cause the creation of a stable topic
+        for i, cluster in enumerate(sorted_clusters):
+            label = cluster["label"]
+            # label is iterating over 0, 1, 3, ...
+
+            # 1. rule - remove if the cores in the cluster have no parents
+            if cluster["amount_parent_labels"] == 0:
```
mpenkov: Instead of these multiple conditionals nested inside the for loop, create the following function:

```python
def _is_valid_cluster(cluster, min_cores):
    """Document your three rules inside this function."""
    if cluster["amount_parent_labels"] == 0:
        return False
    elif len(cluster["parent_labels"]) < min_cores:
        return False
    elif cluster["amount_parent_labels"] < 1:  # Seems to overlap with the first condition
        return False
    return sum(map(lambda x: x == {label}, cluster["parent_labels"])) >= min_cores
```

Then you could simplify your loop as:

```python
for i, cluster in enumerate(sorted_clusters):  # totally untested code
    if not _is_valid_cluster(cluster, num_cores):
        cluster.update(is_valid=False)
        remove_from_all_sets(label)
```

See how things are simpler and more readable this way?

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+                    cluster["is_valid"] = True
+
+        # Reaplying the rule 3
+        # this happens when we have a close relationship among 2 or more clusters
+        for cluster in [cluster for cluster in sorted_clusters if np.isnan(cluster["is_valid"])]:
+
+            # this checks for parent_labels, which are also modified in this function
+            # hence order will influence the result. it starts with the most significant
+            # label (hence "sorted_clusters")
+            if sum(map(lambda x: x == {label}, cluster["parent_labels"])) >= min_cores:
```
mpenkov: OK, you're doing this sum in multiple places, perhaps it should be its own function?

```python
def _get_total_parent_labels(cluster):
    return sum(map(lambda x: x == {label}, cluster["parent_labels"]))
```

aloosley:
> @sezanzeb I am working on a similar idea using pLSA and a slightly different ensembling/clustering approach, but would very much like to reference this work as one of the inspirations. Do you have a citable reference for this, or is one potentially forthcoming? Thanks.

@lmcinnes , thanks so much for the suggestion!   

@sezanzeb , I suggest we use your thesis as the article to cite.  Can you double check with your university that it is ok and then we can put a citation note here?

sezanzeb:
> @sezanzeb , I suggest we use your thesis as the article to cite. Can you double check with your university that it is ok and then we can put a citation note here?

@aloosley, @lmcinnes There is a thesis and it's exciting that there is interest in it, thanks. Since its official completion there have been some changes though that need to be released first. Until when do you need it? I'll probably simply put it on my website as a downloadable pdf.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+from multiprocessing import Process, Pipe, ProcessError
+import numpy as np
+from scipy.spatial.distance import cosine
+from gensim import utils
+from gensim.models import ldamodel, ldamulticore, basemodel
+
+logger = logging.getLogger(__name__)
+
+# nps max random state of 2**32 - 1 is too large for windows:
+MAX_RANDOM_STATE = np.iinfo(np.int32).max
```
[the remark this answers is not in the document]
aloosley: Done, changed in next push.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+        distance_workers : number, optional
+            When distance_workers is None, it defaults to
+            os.cpu_count() for maximum performance.
+            Default is 1, which is not multiprocessed.
+            Set to > 1 to enable multiprocessing.
+        **gensim_kw_args
+            All the parameters for the gensim model,
+            that is used for each model in the ensemble,
+            can be provided as additional keyword arguments.
+            https://radimrehurek.com/gensim/models/ldamodel.html
```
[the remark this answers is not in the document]
aloosley: Done, changed in next push.  I like your usage of YMMV :-).

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+            gensim_kw_args["corpus"] = None
+
+        # dictionary. modified version of from gensim/models/ldamodel.py
+        # error messages are copied from the original gensim module [1]:
+        # will create a fake dict if no dict is provided.
+        if gensim_kw_args["id2word"] is None and not gensim_kw_args["corpus"] is None:
+            logger.warning("no word id mapping provided; initializing from corpus, assuming identity")
+            gensim_kw_args["id2word"] = utils.dict_from_corpus(gensim_kw_args["corpus"])
+        if gensim_kw_args["id2word"] is None and gensim_kw_args["corpus"] is None:
+            raise ValueError("at least one of corpus/id2word must be specified, to establish "
```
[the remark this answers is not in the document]
aloosley: Done, changed in next push.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+                             "input space dimensionality. Corpus should be provided using the "
+                             "keyword argument corpus.")
```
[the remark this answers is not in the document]
aloosley: Done, changed in next push.

gensim/models/ensemblelda.py
aloosley: Done, changed in next push.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+
+        if type(topic_model_kind) == str:
```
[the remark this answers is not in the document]
sezanzeb: done

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1268 @@
...
+            self.topic_model_kind = {
+                "lda": ldamodel.LdaModel,
+                "ldamulticore": ldamulticore.LdaMulticore
+            }[topic_model_kind]
+        else:
+            self.topic_model_kind = topic_model_kind
+
+        # Store some of the parameters
```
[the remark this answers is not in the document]
sezanzeb: done

gensim/models/ensemblelda.py
aloosley: Docstring here refactored to be more clear, including making this change.

gensim/models/ensemblelda.py
sezanzeb: done

gensim/models/ensemblelda.py
sezanzeb: done

gensim/models/ensemblelda.py
sezanzeb: done

gensim/models/ensemblelda.py
sezanzeb: done

sezanzeb: done

gensim/models/ensemblelda.py
sezanzeb: done

gensim/models/ensemblelda.py
sezanzeb: done

gensim/models/ensemblelda.py
sezanzeb: if hanging indents are also a valid option, this is done

gensim/models/ensemblelda.py
sezanzeb: done

gensim/models/ensemblelda.py
sezanzeb: thanks. It is indeed only used internally, should it still be outside of the EnsembleLda class? Added the _

gensim/models/ensemblelda.py
sezanzeb: thanks. done

gensim/models/ensemblelda.py
aloosley: Done.

gensim/models/ensemblelda.py
aloosley: Done.

lmcinnes:
@sezanzeb Mostly I just want to make sure that credit is given where possible. I am happy enough to wait until you have something available so I can point people toward something substantial.

gensim/models/ensemblelda.py
sezanzeb: For loops changed to enumerator loops for better code comprehension and overall cleaner code.

gensim/models/ensemblelda.py
sezanzeb: Done

aloosley:
> @sezanzeb Mostly I just want to make sure that credit is given where possible. I am happy enough to wait until you have something available so I can point people toward something substantial.

Thanks!  We're still working on getting a public release of @sezanzeb's full bachelor thesis on the topic.  In the meantime, we'll go the [pytorch-transformers](https://github.com/huggingface/pytorch-transformers) route by adding the following citation clause to the `ensemblelda.py`

Citation
--------
At the moment, there is no paper associated to ensemble LDA but we are working on publicly releasing Tobias Brigl's bachelor thesis on the topic.  In the meantime, please include a mention of us (Brigl, T.; Salles, A.; Loosley, A) and a link to this file (which will one day have the link: https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/ensemblelda.py) if you use this work in a published or open-source project.

gensim/models/ensemblelda.py
sezanzeb: no from what I can see but at this point the comment might be misaligned from the code

gensim/models/ensemblelda.py
aloosley: Added docstring, as well as full explanation for the CBDBSCAN algorithm (because we haven't published a proper public paper reference yet, I thought that would help).

gensim/models/ensemblelda.py
sezanzeb: this will break it, changed it back

sezanzeb:
@mpenkov 

this is how `tox -e compile,docs -vv` fails in circleci:

https://circleci.com/gh/RaRe-Technologies/gensim/2663
> RuntimeError: Python version >= 3.5 required

I don't see how this can be an issue caused by the changes in the PR. This is the change from which on it started to fail: https://github.com/RaRe-Technologies/gensim/commit/760304577eea337b28ff341dc1402b14cddd2188 which is merely an added blank line

For a different PR it doesn't fail: https://circleci.com/gh/RaRe-Technologies/gensim/2668

---

Btw the most recent circleci tests don't show any log output at all (https://circleci.com/gh/RaRe-Technologies/gensim/2671)

mpenkov:
Looks like numpy stopped supporting Py2.7.

@menshikh-iv @piskvorky I propose we stop running Py2.7 builds in CircleCI and Travis. This can be our first step in removing Py2.7 support for the next major release.

sezanzeb:
All reviews are resolved. Let's see if the tests pass

I will also rerun some of my Jupyter notebooks with the changed code to further look for potential problems and inform you when that's done.

I'll also review the thesis text once more soon and publish it on the website. The source url in ensemblelda.py will remain the same.

mpenkov:
OK, I will get back to you when I get these tests passing (see https://github.com/RaRe-Technologies/gensim/pull/2601).

gensim/models/ensemblelda.py
```diff
@@ -979,6 +979,11 @@ def _remove_from_all_sets(self, label, clusters):
             if label in a["parent_labels"]:
                 a["parent_labels"].remove(label)
 
+    def _is_easy_valid_cluster(self, label, cluster, min_cores):
+        """checks if the cluster has at least min_cores of cores with the only parent-label as itself,
```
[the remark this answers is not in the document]
sezanzeb: Thanks. I have resolved all problems that this tool mentioned: https://github.com/PyCQA/pydocstyle

However, there are some things I need to discuss with @aloosley first before continuing with the PR

mpenkov:
@sezanzeb @aloosley @piskvorky All the tests now pass. Is this ready to merge?

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1298 @@
...
+from multiprocessing import Process, Pipe, ProcessError
+import numpy as np
+from scipy.spatial.distance import cosine
+from gensim import utils
+from gensim.models import ldamodel, ldamulticore, basemodel
+
+logger = logging.getLogger(__name__)
+
+
+class EnsembleLda():
```
piskvorky: Inherit from `object` (for Python 2).

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1298 @@
...
+-----------
+.. The adjectives stable and reliable (topics) are used somewhat interchangeably throughout the doc strings and
+comments.
+
+"""
+
+import logging
+import os
+from multiprocessing import Process, Pipe, ProcessError
+import numpy as np
```
piskvorky: Blank line between built-in imports and 3rd party imports.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1298 @@
...
+from scipy.spatial.distance import cosine
+from gensim import utils
```
piskvorky: Blank line between 3rd party imports and internal imports.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1298 @@
...
+
+        if self.ttda.shape[1] != ttda.shape[1]:
+            raise ValueError(("target ttda dimensions do not match. Topics must be {} but was"
+                                "{} elements large").format(self.ttda.shape[-1], ttda.shape[-1]))
+        self.ttda = np.append(self.ttda, ttda, axis=0)
+
+        # tell recluster that the distance matrix needs to be regenerated
+        self.asymmetric_distance_matrix_outdated = True
+
+    def save(self, fname):
```
piskvorky: Inherit from `utils.SaveLoad`, to get these `save` and `load` methods for free. This "standard" Gensim serialization supports additional features, such as serializing large arrays (where plain pickle will fail).

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1298 @@
...
+            else:
+                ttda = answer
+            self.ttda = np.concatenate([self.ttda, ttda])
+
+        # end all processes
+        for p in processes:
+            p.terminate()
+
+    def _generate_topic_models(self, num_models, random_states=None, pipe=None):
+        """Train the topic models, that form the ensemble.
```
piskvorky:
```suggestion
        """Train the topic models that form the ensemble.
```

sezanzeb:
> @sezanzeb @aloosley @piskvorky All the tests now pass. Is this ready to merge?

not yet, there have been some experimental changes that @aloosley and I managed to discuss recently, but they are about to be reverted.

> @sezanzeb this model is sufficiently complex that few will dare touch it. Can we count on you to answer occasional user queries / fix bugs going forward? Thanks!

yes

also, there have been quite some simplifications that will be beneficial in understanding the code recently.

I will let you know when it's ready to merge, won't take long.

thanks!

gensim/models/ensemblelda.py
mpenkov: But we don't care about Python 2 anymore, right?

piskvorky: In that case, drop the `()`.

aloosley:
> > @sezanzeb @aloosley @piskvorky All the tests now pass. Is this ready to merge?
> 
> no, not yet, there have been some experimental changes that @aloosley and I managed to discuss recently, but they are about to be reverted.
> 
> > @sezanzeb this model is sufficiently complex that few will dare touch it. Can we count on you to answer occasional user queries / fix bugs going forward? Thanks!
> 
> yes
> 
> also, there have been quite some simplifications that will be beneficial in understanding the code recently. Right now I am running some experiments to make sure there is no impact on the results from those changes.
> 
> I will let you know when it's ready to merge, won't take long.
> 
> thanks!

After the ML Prague talk this year, some people were brave enough to try eLDA.  Despite the usual challenge of being pretty busy, we were mostly getting back to people on a 1-7 day basis.  

Once the merge is complete and we have this base for people to try it out, @sezanzeb and I will organize a few around the subject.  If those talks hit the right people and they get interested, then it'll also be our duty to answer their questions.  Then maybe, just maybe, that will be enough to form a small knowledge community around eLDA.

docs/notebooks/Opinosis.ipynb
```diff
@@ -0,0 +1,176 @@
+{
```
aloosley: Let's call this `ensemble_lda_with_opinosis`

gensim/models/ensemblelda.py
sezanzeb: It fails here when val is a class instead of an object:
https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/utils.py#L618

Since `hasattr(LdaModel, '_save_specials')` is True. The first argument to _save_specials will then be 'self', hence subname is missing as a positional argument.

```
TypeError: _save_specials() missing 1 required positional argument: 'subname'
```

I think it would work if it said something among
```
if hasattr(val, '_save_specials') and type(val) != type:
```
in line 609, do you see any downside to this? I'm going to commit it and see if gensims tests pass in the meantime. Or with isinstance, but apparently that causes other issues (see comment of line 609).

We need a class as one of the attributes in ensembleLda, because we need to store which models are going to be trained as children.

piskvorky: I don't follow. Can you show the full traceback?

sezanzeb: `elda.__dict__['topic_model_class']` is `gensim.models.ldamodel.LdaModel` or other classes like LdaMulticore

```
TypeError                          Traceback (most recent call last)
~/Code/ensemblelda/gensim/gensim/utils.py in save(self, fname_or_handle, separately, sep_limit, ignore, pickle_protocol)
    696         try:
--> 697             _pickle.dump(self, fname_or_handle, protocol=pickle_protocol)
    698             logger.info("saved %s object", self.__class__.__name__)

TypeError: file must have a 'write' attribute

TypeError                          Traceback (most recent call last)
<ipython-input-9-35ccc388ca18> in <module>
----> 1 elda.save("./bla")

~/Code/ensemblelda/gensim/gensim/utils.py in save(self, fname_or_handle, separately, sep_limit, ignore, pickle_protocol)
    698             logger.info("saved %s object", self.__class__.__name__)
    699         except TypeError:  # `fname_or_handle` does not have write attribute
--> 700             self._smart_save(fname_or_handle, separately, sep_limit, ignore, pickle_protocol=pickle_protocol)
    701 
    702 

~/Code/ensemblelda/gensim/gensim/utils.py in _smart_save(self, fname, separately, sep_limit, ignore, pickle_protocol)
    548 
    549         restores = self._save_specials(fname, separately, sep_limit, ignore, pickle_protocol,
--> 550                                        compress, subname)
    551         try:
    552             pickle(self, fname, protocol=pickle_protocol)

~/Code/ensemblelda/gensim/gensim/utils.py in _save_specials(self, fname, separately, sep_limit, ignore, pickle_protocol, compress, subname)
    612                 cfname = '.'.join((fname, attrib))
    613                 print(val, subname)
--> 614                 restores.extend(val._save_specials(cfname, None, sep_limit, ignore, pickle_protocol, compress, subname))
    615 
    616         try:

TypeError: _save_specials() missing 1 required positional argument: 'subname'
```

console output right before the line that fails:

`<class 'gensim.models.ldamodel.LdaModel'> <function SaveLoad._adapt_by_suffix.<locals>.<lambda> at 0x7f4dc8d40290>`

sezanzeb: the suggested change of `if hasattr(val, '_save_specials') and type(val) != type:` passes the CI tests

piskvorky: Thanks @sezanzeb. I'll look at this more closely, the existing Gensim code looks wild. CC @mpenkov.

piskvorky: OK, I think I see what you mean. Some object attribute is a class, and should not be saved recursively.

The `isinstance(val, SaveLoad)` mentioned in the comment looks better. That comment and code change was introduced [here](https://github.com/RaRe-Technologies/gensim/commit/9cac3843481688f79e3904e893f49163c6328789) by @gojomo, but looks like a regression to me.

@gojomo what are these "IPython reloads" that prompted this? Do we really need this?

piskvorky: OK, I think I deciphered that too :) If a user explicitly reloads the `utils` module, for example from an interactive Python prompt, `SaveLoad` will be a different object in the reloaded module vs the previously loaded modules.

And then the check `isinstance(obj, SaveLoad)` may return false, even though `obj` inherits from `SaveLoad` (but it's the other, non-reloaded `SaveLoad`).

I don't think this use-case is really worth it though. @mpenkov @gojomo thoughts? I'm +1 on returning back to plain `isinstance(obj, SaveLoad)`… assuming that solves this issue @sezanzeb identified here.

gojomo: Yes, the problem is that in a dynamic environment like IPython/Jupyter, where classes may be reloaded, some older objects' reference to `SaveLoad` may not be the identity-same class object – so it's more robust to test for the capability, using Python-style 'duck-typing', than the exact class/inheritance.  As that change has been working fine for 4.5 years, & also strikes me as generally more 'hygenic' & flexible for the common case of Jupyter-notebook work, a reversion seems misguided. I'm not sure why now, for the first time in years, this method would be called without a 'self' instance, so I'd tend to think whatever new usage is calling it in a new way should change. Or, if it needs a check against being inappropriately called on classes, adding the extra check @sezanzeb proposed makes more sense than re-introducing an older fragility.

piskvorky: That proposed check is not 100% though, because apart from "being a class", the object may also be a function or a module (among other things).

I can see how the reloading scenario can be useful. Although I'd expect you'll encounter other issues too if you do such runtime shenanigans, with different versions of the same module / class across your environment. I'm surprised hacking one `if` inside one method was enough.

More "hygenic" would be to go and check what we **really** want to be checking: whether the attribute is an object inheriting from SaveLoad, so we can save it separately.

Not to derail this PR, @sezanzeb did a great job. I'm mildly in favour of the more straightforward & correct `instance(obj, SaveLoad)`, but no strong preference. @mpenkov WDYT?

mpenkov: @piskvorky We already have an issue to author a tutorial for better discovery (https://github.com/RaRe-Technologies/gensim/issues/2673).

For the way forward, I suggest we go with what's already in this branch, and kick off a separate issue to deal with doing serialization "the proper way", whatever we decide that to be in the future.

piskvorky: The proper way is to use `SaveLoad`, there's no question about that. This discussion is on a subtle technical point: how to best test whether an object attribute is SaveLoad-serializable itself, recursively.

sezanzeb: for the moment maybe we can stick to `if hasattr(val, '_save_specials') and type(val) != type:` which makes things work right now without side effects. Later we can still change it to something more sophisticated.

piskvorky: @sezanzeb does `isinstance(obj, SaveLoad)` solve your issue?

gojomo: I still don't see `isinstance(obj, SaveLoad)` as `more straightforward & correct'. It's more fragile in a real usage scenario, Jupyter notebooks,  that's important for developers. 

I'm not sure why this new code is, for the 1st time, now passing actual class objects into this test. That no prior usage attempted this suggests to me it's a violation of the implicit contract of `save()` – save should be serializing instances, **not** full classes that might be recursively mentioned. 

So my hunch is the best fix would be to change the new code to avoid demanding very-different things from older code (like handling full-on class objects in `_smart_save()` special serialization). Maybe the `.topic_model_class` should be a specific string name, from a controlled-vocabulary of allowable options. Or, a fully-qualified class name string, to allow reference to any loadable class, but without having an actual class-object there. (Allowing actual class-objects in what you're trying to serialize is like having lambdas or functions or user-package callback objects – a major headache for later deserialization.) The explicit documentation of `SaveLoad` could be expanded to explain that properties with references to {classes, lambdas, functions, user-package code objects} are either unsupported or strongly not-recommended.

If it's the intent for `SaveLoad` to support saving references to class objects – classes which might know how to `_save_specials()` their instances, but not the class object – then `SaveLoad` could be upgraded to handle that case, and enforce the (very logical) idea that it is only instance-objects, and not class-objects, that get to specify their own custom serialization via `_save_specials()`. One way to do that is @sezanzeb's suggested fix, to tighten the test of an object's eligilbility for `_save_specials()`. (It might also be possible to check for different handling of class-objects earlier.)

sezanzeb: I will try if I can just pickle the name of the class as string and load it with https://docs.python.org/3/library/importlib.html#importlib.import_module afterwards.

---

```
import pickle
import gensim
with open('dump', 'wb') as f:
    pickle.dump(gensim.models.LdaModel, f)
```
```
!cat dump
```
```
�cgensim.models.ldamodel
LdaModel
q.
```

I think this just stores the name or something. Or what do you mean with "full classes"?

Restricting it on known topic model classes by using strings in the pickling would mean that new kinds of topic models cannot be clustered with eLDA.

---

I don't think I want to break stuff in jupyter notebooks by using isinstance

gojomo: It won't likely break any polished notebooks, that are designed to run right through. Rather, when doing related development, and interactively revising code & re-running cells with live instances, the `isinstance()` check will be more fragile. So it's not a big thing for average users, but it is something I've run into when extending/fixing gensim code. 

When supplying 'full classes' (class/type objects), yes, the serialization still just writes barely more than the class name to disk. But the behavior/expectations are different – it's "as if" a richer/typed object was written, and the pickle-machinery will try to restore that object, and fail if at load-time a class of the same name is unavailable, with a possibly-unhelpful error message & no load of the rest of the state (which might help debug the situation). 

If instead storing the fully-qualified class-name as string instead, then saving/loading should reliably succeed. Then, it's potentially later code, outside the deserialization routines, that changes that string into an instantiatable class when needed. If that step fails, it can perhaps give a more helpful error message, & allow a focused retry of just that step after the user has installed/imported the named class(es).

sezanzeb: https://github.com/RaRe-Technologies/gensim/blob/c559de6cd3366f51af2110758f578ac72d4ec768/gensim/models/ensemblelda.py#L271-L293

gojomo: This fits the gist of what I'd suggested, and so to the extent it avoids persisting a code object & passes tests, that's good. But I believe it could be even more robust & minimal if the class reference is *never* stored as a property, but rather *always* created on demand, at the moments that it is needed, from a fully-qualified string. Then, there's no need for a custom `save()`/`load()` at all. If there is a problem at the moment the class reference is created-on-demand, the exception naturally raised there may be enough to debug the problem – or it could be caught and described with an extra error message.

gensim/models/ensemblelda.py
```diff
@@ -98,6 +98,8 @@
 
 """
 
+print('asdf')
```
mpenkov: ?

sezanzeb: it was removed in 766f562a065f08b3dfdc0754e138ea2a15e38090

I added it to see if the correct module was imported during development

sezanzeb:
The docs compile now

sezanzeb:
whops close accidentally

gensim/models/ensemblelda.py
sezanzeb: @gojomo the custom load is now obsolete.

But I don't think I can make the save function obsolete as well. I could import topic_model_class every time I query the attribute, but I think this fails when topic_model_class is not a module that can be imported from a file. For example some custom class in a jupyter notebook.

sezanzeb:
@piskvorky @aloosley and me think this is ready to merge.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+                else:
+                    num_subprocess_models = int(num_models_unhandled / (workers - i))
+
+                # get the chunk from the random states that is meant to be for those models
+                random_states_for_worker = random_states[-num_models_unhandled:][:num_subprocess_models]
+
+                p = Process(target=self._generate_topic_models,
+                            args=(num_subprocess_models, random_states_for_worker, childConn))
+
+                processes += [p]
```
mpenkov: Please use the append function here and below.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+        # create worker processes:
+        # from what I know this is basically forking with a jump to a target function in each child
+        # so modifying the ensemble object will not modify the one in the parent because of no shared memory
+        processes = []
+        pipes = []
+        num_models_unhandled = num_models  # how many more models need to be trained by workers?
+
+        for i in range(workers):
+            try:
+                parentConn, childConn = Pipe()
```
mpenkov: Please reduce the scope of the try-catch.
Everything other than the pipe and process creation can go outside the block (can creating a pipe raise an exception?).

Please do this here and everywhere.

mpenkov: PEP8.

```suggestion
                parent_conn, child_conn = Pipe()
```

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+
+                processes += [p]
+                pipes += [(parentConn, childConn)]
+                p.start()
+
+                num_models_unhandled -= num_subprocess_models
+
+            except ProcessError:
+                logger.error("could not start process {}".format(i))
+                # close all pipes
```
mpenkov: Please get rid of these comments. They add no value.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+                            args=(num_subprocess_models, random_states_for_worker, childConn))
+
+                processes += [p]
+                pipes += [(parentConn, childConn)]
+                p.start()
+
+                num_models_unhandled -= num_subprocess_models
+
+            except ProcessError:
+                logger.error("could not start process {}".format(i))
```
mpenkov: The ordinal of the process that failed to create is helpful to know, but we can do better here. What was the original exception?

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+                for p in processes:
+                    if p.is_alive():
+                        p.terminate()
+                    del p
+                # stop
+                raise
+
+        # aggregate results
+        # will also block until workers are finished
+        for p in pipes:
```
mpenkov: Again, more helpful variable names, and you don't have to add comments to explain which of p[0] and p[1] is the parent and child.

```suggestion
        for (parent_conn, child_conn) in pipes:
```

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+        """
+        if pipe is not None:
+            logger.info("spawned worker to generate {} topic models".format(num_models))
+
+        if random_states is None:
+            random_states = [self.random_state.randint(self._MAX_RANDOM_STATE) for _ in range(num_models)]
+
+        assert len(random_states) == num_models
+
+        kwArgs = self.gensim_kw_args.copy()
```
mpenkov: Please conform to PEP8.

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+                p = Process(target=self._asymmetric_distance_matrix_worker,
+                            args=(i, ttdas_sent, n_ttdas, childConn, threshold, method))
+                ttdas_sent += n_ttdas
+
+                processes += [p]
+                pipes += [(parentConn, childConn)]
+                p.start()
+
+            except ProcessError:
+                logger.error("could not start process {}".format(i))
```
mpenkov: This duplicates previous code. Split this off into `_teardown(pipes, processes)` function.

gensim/models/ensemblelda.py
aloosley: Done

gensim/models/ensemblelda.py
aloosley: Done

gensim/models/ensemblelda.py
aloosley: Done

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+            else:
+                self.num_models += num_new_models
+
+        else:  # memory unfriendly ensembles
+
+            ttda = []
+
+            # 1. ttda array
+            if isinstance(target.dtype.type(), (np.number, float)):
+                raise ValueError('ttda arrays cannot be added to ensembles, for which memory_friendly_ttda=False, '
```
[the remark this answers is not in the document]
aloosley: Done

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+        classic_model_representation.state.sstats = sstats.astype(np.float32)
+        # fix expElogbeta.
+        classic_model_representation.sync_state()
+
+        self.classic_model_representation = classic_model_representation
+
+        return classic_model_representation
+
+    def add_model(self, target, num_new_models=None):
+        """Add the ttda of another model to the ensemble.
```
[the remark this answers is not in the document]
aloosley: Done

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+
+    def get_topic_model_class(self):
+        """Get the class that is used for :meth:`gensim.models.EnsembleLda.generate_gensim_representation`."""
+        if self.topic_model_class is None:
+            try:
+                module = importlib.import_module(self.topic_model_module_string)
+                self.topic_model_class = getattr(module, self.topic_model_class_string)
+                del self.topic_model_module_string
+                del self.topic_model_class_string
+            except ModuleNotFoundError:
```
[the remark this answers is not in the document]
sezanzeb: thanks. Can you explain why .format should not be used? Since python2 support was dropped it should be compatible with everybody who runs gensim, shouldn't it?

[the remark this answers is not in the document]
sezanzeb: the error messages are actually a tiny bit different

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+        """Add the ttda of another model to the ensemble.
+
+        This way, multiple topic models can be connected to an ensemble manually. Make sure that all the models use
+        the exact same dictionary/idword mapping.
+
+        In order to generate new stable topics afterwards, use
+            self._generate_asymmetric_distance_matrix()
+            self.recluster()
+
+        The ttda of another ensemble can also be used, in that case set num_new_models to the num_models parameter
```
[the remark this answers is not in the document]
aloosley: Done

gensim/models/ensemblelda.py
```diff
@@ -0,0 +1,1280 @@
...
+
+        return classic_model_representation
+
+    def add_model(self, target, num_new_models=None):
+        """Add the ttda of another model to the ensemble.
+
+        This way, multiple topic models can be connected to an ensemble manually. Make sure that all the models use
+        the exact same dictionary/idword mapping.
+
+        In order to generate new stable topics afterwards, use
```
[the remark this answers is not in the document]
aloosley: Fixed, with many more docstring polishes and RST formatting updates.

gensim/models/ensemblelda.py
aloosley: Done

sezanzeb:
Thanks for your efforts in reviewing our stuff!

haven't checked if opinosis still works after the changes, don't merge yet. Gonna check the following weekend or something

sezanzeb:
looks good

sezanzeb:
@piskvorky @mpenkov any updates?

piskvorky:
Sorry, the beginning of the year was super busy. We have a bunch of PRs lined up, we'll get to this with @mpenkov shortly! Thanks.

sezanzeb:
> Sorry, the beginning of the year was super busy. We have a bunch of PRs lined up, we'll get to this with @mpenkov shortly! Thanks.

Ah I see, don't worry at all!

sezanzeb:
I'm going to reopen this PR from my own fork, because I don't get access to the repo anymore, possibly due to organizational changes within that company.

PR #2290, opened 2018-12-09
Fix typos in `gensim/models/keyedvectors.py`

rsdel2007:


menshikh-iv:
Thank you @rsdel2007 :+1:

PR #2297, opened 2018-12-15
Update word2vec.py

Tusharmalik:


piskvorky:
Thanks, but it's really `mmap` :)

PR #2298, opened 2018-12-16
Reduce memory consumption of summarizer

horpto:
By a request of @menshikh-iv:

- I found a couple of questions ([this](https://stackoverflow.com/questions/50567108/gensim-summarizer-throws-memoryerror-any-solution) and [this](https://stackoverflow.com/questions/50516951/python-extracting-text-and-summarize-multiple-file-types-in-directory-with-tik) about memory of summarization when I was looking for an information related to my previous PR with graph nodes removing. 
- I've taken first part of the english translated War and Peace. It consists of 14426 unique docs (number of nodes). I've tested on computer with 16 GiB of memory, Intel i5, 64-bit Windows 10, Python 3.7. Original version have eaten 14 GiB of memory after 7 minutes of execution and failed with MemoryError. My version have eaten 2 GiB of memory and taken ~2min (108 sec), but failed to process full version War and Peace (~57000 unique docs) with MemoryError too (failed on the phase pagerank).


horpto:
I need https://github.com/RaRe-Technologies/gensim/pull/2263 to add some tests about graph.

gensim/summarization/pagerank_weighted.py
```diff
@@ -60,24 +62,28 @@ def pagerank_weighted(graph, damping=0.85):
         Nodes of `graph` as keys, its ranks as values.
 
     """
-    adjacency_matrix = build_adjacency_matrix(graph)
-    probability_matrix = build_probability_matrix(graph)
+    coeff_adjacency_matrix = build_adjacency_matrix(graph, coeff=damping)
+    probabilities = (1 - damping) / float(len(graph))
 
-    pagerank_matrix = damping * adjacency_matrix.todense() + (1 - damping) * probability_matrix
+    pagerank_matrix = coeff_adjacency_matrix.toarray()
```
horpto: Even for sparse graph we need O(len(graph)**2) memory =(
For example, graph with 50000 nodes it will be needed 50000*50000 *4 bytes = 9536 MiB.
That's not cool.

gensim/summarization/bm25.py
```diff
@@ -162,6 +162,48 @@ def get_scores(self, document):
...
+        Returns
+        -------
+        list of float
+            BM25 scores.
+
+        """
+        scores = []
+        for index in range(self.corpus_size):
+            score = self.get_score(document, index)
+            if score > 0:
```
menshikh-iv: In that case `len(scores) <= self.corpus_size`, why?

gensim/summarization/graph.py
```diff
@@ -303,31 +289,27 @@ def add_edge(self, edge, wt=1, label='', attrs=None):
...
-            attrs = []
+        if wt == 0.0:
+            # empty edge is similar to no edge at all or removing it
+            if self.has_edge(edge):
+                self.del_edge(edge)
+            return
         u, v = edge
         if v not in self.node_neighbors[u] and u not in self.node_neighbors[v]:
-            self.node_neighbors[u].append(v)
+            self.node_neighbors[u][v] = wt
```
menshikh-iv: How looks `node_neighbors` now, is it 
```python
{"node_1": {"node_25": 0.5, ...} ...}

```
?

gensim/summarization/graph.py
```diff
@@ -303,31 +289,27 @@ def add_edge(self, edge, wt=1, label='', attrs=None):
...
-                self.node_neighbors[v].append(u)
-
-            self.add_edge_attributes((u, v), attrs)
-            self.set_edge_properties((u, v), label=label, weight=wt)
+                self.node_neighbors[v][u] = wt
         else:
             raise ValueError("Edge (%s, %s) already in graph" % (u, v))
 
-    def add_node(self, node, attrs=None):
+    def add_node(self, node):
```
menshikh-iv: we need only weight always and for this reason, you get rid attrs, right?

gensim/summarization/pagerank_weighted.py
menshikh-iv: but why? matrix has a little number of edges.

gensim/summarization/pagerank_weighted.py
```diff
@@ -89,40 +95,45 @@ def build_adjacency_matrix(graph):
...
+        neighbors = graph.neighbors(current_node)
+        neighbors_sum = sum(graph.edge_weight((current_node, neighbor)) for neighbor in neighbors)
+        for neighbor in neighbors:
+            edge_weight = float(graph.edge_weight((current_node, neighbor)))
+            if edge_weight != 0.0:
                 row.append(i)
-                col.append(j)
-                data.append(edge_weight / neighbors_sum)
+                col.append(nodes2id[neighbor])
+                data.append(coeff * edge_weight / neighbors_sum)
```
menshikh-iv: what's a reason to move dumping coeff here?

gensim/summarization/summarizer.py
```diff
@@ -358,8 +352,11 @@ def summarize_corpus(corpus, ratio=0.2):
     if len(corpus) < INPUT_MIN_LENGTH:
         logger.warning("Input corpus is expected to have at least %d documents.", INPUT_MIN_LENGTH)
 
+    logger.info('Building graph')
     graph = _build_graph(hashable_corpus)
+    logger.info('Filing graph')
```
menshikh-iv: `filling` I guess

gensim/summarization/graph.py
horpto: Yes, you're right. But one nuance: nodes actually are bow-s. It's a documents from corpus.

gensim/summarization/summarizer.py
horpto: yeah. I'll fix it.

gensim/summarization/pagerank_weighted.py
horpto: -1 pass by array on len(edges) (and actually by matrix on `(len(graph), len(graph)` due to  transforming from sparse to dense)

gensim/summarization/pagerank_weighted.py
horpto: -1 pass by array on len(edges) (and actually by matrix on (len(graph), len(graph) due to transforming from sparse to dense)

gensim/summarization/pagerank_weighted.py
horpto: Because to adjacency_matrix are added probability of damping. So matrix becomes not sparse. I'm gonna try to find the other ways to calculate pagerank but in other PR I think.

gensim/summarization/bm25.py
horpto: Because it's actually quite sparse array, isn't it?
In `summarizer._set_graph_edge_weights` such documents with little weight will be dropped anyway, so there is no reason to waste extra memory. And what's more, if we are needed a dense array we can uncompactify this bow.

gensim/summarization/graph.py
horpto: Yes, even in original commit of @fbarrios this attrs was not used anywhere else besides Graph itself. Labels in original commits was used in one script that already deleted anyway. 
I don't like that I broke backward compatibility but on the other side gensim is not a library for the graphs and it's an inner class. I may like a compromise between compatibility and efficiency create something like a SimpleGraph and use it and deprecate ordinary Graph.

gensim/summarization/graph.py
menshikh-iv: that's fine, thanks for clarification

gensim/summarization/graph.py
menshikh-iv: That's fine I guess, this is some stuff for internal usage, in this case, we shouldn't get any problems with it

gensim/summarization/bm25.py
menshikh-iv: I don't get, how we understand ids of documents that have 0 scores in that case?

menshikh-iv:
> I've taken first part of the english translated War and Peace. It consists of 14426 unique docs (number of nodes). I've tested on computer with 16 GiB of memory, Intel i5, 64-bit Windows 10, Python 3.7. Original version have eaten 14 GiB of memory after 7 minutes of execution and failed with MemoryError. My version have eaten 2 GiB of memory and taken ~2min (108 sec), but failed to process full version War and Peace (~57000 unique docs) with MemoryError too (failed on the phase pagerank).


Awesome result, nice improvement for new release, thanks @horpto :+1:

PR #2300, opened 2018-12-18
Refactor ldamulticore to serialize less data

horpto:
Communication between processes is not free due to pickling/unpickling.
So to reduce overhead we are trying to send a less data.


gensim/models/ldamulticore.py
```diff
@@ -289,7 +289,7 @@ def process_result_queue(force=False):
                 chunk_put = False
                 while not chunk_put:
                     try:
-                        job_queue.put((chunk_no, chunk, self), block=False, timeout=0.1)
+                        job_queue.put((chunk_no, chunk, self.state), block=True, timeout=0.1)
```
menshikh-iv: why `block=True`?

horpto: because in this place we should or remove timeout or set `block` to `True`. As said in [docs](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Queue.put) at `Block=False` timeout is ignored. In case when queue is full this `put` can be repeated very often and eat CPU, so I just make it blocked.

horpto:
@menshikh-iv this one patch doesn't bring me a big improvement. It save me a roughly 3 minutes from 80 at all (one measurement). This patch is more about refactoring. I copy approach from distributed LDA (lda_worker) to multicore to make it both more similar. For a distributed LDA communication quality (size of sent/received data) is even more important than for a multicore, so I choose  this and not vice versa.

gensim/models/ldamulticore.py
```diff
@@ -289,7 +289,7 @@ def process_result_queue(force=False):
                 # put the chunk into the workers' input job queue
                 while True:
                     try:
-                        job_queue.put((chunk_no, chunk, self), block=False)
+                        job_queue.put((chunk_no, chunk, self.state), block=False)
```
mpenkov: I think this is a little bit confusing.

The pool workers receive `self` as a parameter. Why are we passing `self.state` via the job queue as well?

piskvorky:
Plus benchmark numbers please: how much was saved, and how does it affect overall timing / RAM?

gensim/models/ldamulticore.py
menshikh-iv: I guess the reason is simple: `RAM_USAGE(self) > RAM_USAGE(self.state)`

mpenkov: But why do we need access to self.state when we already have self?

menshikh-iv: good question, idk
@horpto ?

mpenkov: @horpto Ping.

mpenkov: @horpto Can you please comment on this?

When we're [setting up the pool](https://github.com/RaRe-Technologies/gensim/pull/2300/files#diff-68941d8c838f8d70a6648f06b9866889R280), we're calling it like this:

```python
pool = Pool(self.workers, worker_e_step, (job_queue, result_queue, self))
```

The first arg is the number of workers, the second is an initializer function, and the third is a set of parameters to pass to the initializer ([docs](https://docs.python.org/3.8/library/multiprocessing.html#multiprocessing.pool.Pool)). The workers will call this initializer with the args on start-up. 

> If initializer is not None then each worker process will call initializer(*initargs) when it starts.

So, as you can see, we are already passing `self` to the workers.

Why do we need to pass `self` again here?

mpenkov:
> Plus benchmark numbers please: how much serialized data was saved, and how does it affect overall timing / RAM?

@horpto Are you able to complete this PR? In particular, we need to see those benchmark results.

horpto:
@mpenkov, i think i can make banchmark today-tomorrow. sorry for such long responses.

gensim/models/ldamulticore.py
horpto: > Why do we need to pass self again here?

Well, it's `multiprocessing` and synchronization issue.  Worker and master processes will have different state. When `Pool`  is called it will create `self.workers` processes (via [_repopulate_pool_static method](https://github.com/python/cpython/blob/8bea91b5e9ea07ca93958e131b436024f0b1b1cf/Lib/multiprocessing/pool.py#L312)) with our initargs. Current process class creates system process (2-3 different ways and especial pain is win) and then sends data to children process with pickle and pipe (for example [spawn_win32](https://github.com/python/cpython/blob/8bea91b5e9ea07ca93958e131b436024f0b1b1cf/Lib/multiprocessing/popen_spawn_win32.py#L93).) Of course, there is a fork type process [(fork popen)](https://github.com/python/cpython/blob/master/Lib/multiprocessing/popen_fork.py) but it's not crossplatform. What's more [multiprocess.Queue](https://github.com/python/cpython/blob/master/Lib/multiprocessing/queues.py#L35) also uses pipe+pickle so has overhead.

mpenkov:
OK, looks like all that's missing is the benchmarks, and then we can merge this.

sezanzeb:
I'm currently trying to make some benchmarks for this that measure time taken on some small subset of the wiki corpus (enwiki-latest-pages-articles1.xml-p1p30303.bz2 from https://dumps.wikimedia.org/enwiki/latest/)

But I wonder how I can test how much ram is saved. After transmitting and extracting the results, the ram would be freed by the garbage collector, right? So it might be only about small peaks that are difficult to capture.

sezanzeb:
If I didn't screw up the test, here are my results.

data: `enwiki-latest-pages-articles12.xml-p3926862p5040436.bz2` from https://dumps.wikimedia.org/enwiki/latest/

each gauss is created from 200 samples.

source notebook: [Benchmark_serialize_less.ipynb.zip](https://github.com/RaRe-Technologies/gensim/files/5272333/Benchmark_serialize_less.ipynb.zip)

The x-axis is time in seconds. Y axis is the normalized density of timing results.

# 1. Model

`workers=11, num_topics=100, passes=1, chunksize=1000`

newplot

# 2. Model

`workers=11, num_topics=100, passes=1, chunksize=2000`

newplot-1

# 3. Model

`workers=11, num_topics=100, passes=1, chunksize=4000`

newplot-2

horpto:
Wow, thanks, @sezanzeb.  If I understand graphics correctly so my PR should have a better performance but only for small chanks. When I was working on it, I had no a great profit in my tests.

sezanzeb:
> Wow, thanks, @sezanzeb. If I understand graphics correctly so my PR should have a better performance but only for small chanks. When I was working on it, I had no a great profit in my tests.

Or when decreasing the gamma_threshold, or any other configuration that makes the worker take longer. Maybe somebody who applies this model to real world problems knows a more realistic configuration and corpus, my benchmark above is possibly rather unrealistic. But it shows that there is a benefit.

I wonder if this can be merged now

mpenkov (approved):
Looks good to me. Thank you for the work @horpto, and the benchmarks @sezanzeb. Great team effort, guys!!

PR #2303, opened 2018-12-19
Improve `filter_extremes` methods in `Dictionary` and `HashDictionary`

horpto:
- use search by set instead of list
- refine default value of dict.get() method
- inplace sort


menshikh-iv:
Thanks @horpto :star:

piskvorky:
Awesome! These are the kind of deep and small but **important** fixes that help us so much. 👍

PR #2307, opened 2018-12-23
Added Function relative_cosine_similarity in keyedvectors.py

### LINKED ISSUE #2175: Feature suggestion: relative cosine similarity for word2vec
Hi all,
Based on this paper, do you think it worths the effort to implement relative cosine similarity measure? 
https://ufal.mff.cuni.cz/pbml/105/art-leeuwenberg-et-al.pdf

Note that I'm not suggesting this as a potential contributor but as a grateful user.
Thank you,
Viktor

rsdel2007:
Fixes #2175.
I have implemented relative_cosine_similarity as function according to the [paper](https://ufal.mff.cuni.cz/pbml/105/art-leeuwenberg-et-al.pdf) and as @gojomo suggested in #2175 discussion. 
According to paper : 
  rcs(top-n) = cosine_similarity(wordA,wordB)/(sum of cosine_similarities of top-n similar words to wordA)
 For finding the top-n similar words I have used method `similar_by_word(word,topn)`.


gensim/models/keyedvectors.py
```diff
@@ -1384,7 +1384,42 @@ def init_sims(self, replace=False):
             else:
                 self.vectors_norm = (self.vectors / sqrt((self.vectors ** 2).sum(-1))[..., newaxis]).astype(REAL)
 
+    def relative_cosine_similarity(self, wa, wb, topn=10):
+        """Compute the relative cosine similarity between two words given top-n similar words,
+        proposed by Artuur Leeuwenberg,Mihaela Vela,Jon Dehdari,Josef van Genabith
```
gojomo: Spaces after commas.

gensim/models/keyedvectors.py
```diff
@@ -1384,7 +1384,42 @@ def init_sims(self, replace=False):
...
+        "A Minimally Supervised Approach for Synonym Extraction with Word Embeddings"
+        <https://ufal.mff.cuni.cz/pbml/105/art-leeuwenberg-et-al.pdf>.
+        To calculate relative cosine similarity between two words, equation (1) of the paper is used.
+        For WordNet synonyms, if rcs(topn=10) is greater than 0.10 than wa and wb are more similar than
```
gojomo: Second of three 'than's on this line should actually be 'then' (consequently) not 'than' (comparative).

gensim/models/keyedvectors.py
```diff
@@ -1384,7 +1384,42 @@ def init_sims(self, replace=False):
...
+            topn_words.append(result[i][0])
+            topn_cosine.append(result[i][1])
+
+        topn_cosine = np.array(topn_cosine)
+
+        norm = np.sum(topn_cosine)
+
+        rcs = (self.similarity(wa, wb)) / norm
 
+        return rcs
```
gojomo: Need blank line before next method.

gensim/models/keyedvectors.py
```diff
@@ -1384,7 +1384,42 @@ def init_sims(self, replace=False):
...
+        topn: int, optional
+         Number of top-n similar words to look with respect to wa.
+        Returns
+        -------
+        numpy.float64
+            relative cosine similarity between wa and wb.
+        """
+
+        result = self.similar_by_word(wa, topn)
+        topn_words = []
```
gojomo: `topn_words` is never needed to calculate results - so no good reason to create.

gensim/models/keyedvectors.py
```diff
@@ -1384,7 +1384,42 @@ def init_sims(self, replace=False):
...
+         word for which we evaluating relative cosine similarity with wa.
+        topn: int, optional
+         Number of top-n similar words to look with respect to wa.
+        Returns
+        -------
+        numpy.float64
+            relative cosine similarity between wa and wb.
+        """
+
+        result = self.similar_by_word(wa, topn)
```
gojomo: As this is a list of results, using a plural variable name would be slightly better. Also, it's common in the existing gensim code to call the list-of-most-similar-items `sims` (short for 'similars'), so I'd recommend that variable name here.

gensim/models/keyedvectors.py
```diff
@@ -1384,7 +1384,42 @@ def init_sims(self, replace=False):
...
+        topn_words = []
+        topn_cosine = []
+        for i in range(topn):
+            topn_words.append(result[i][0])
+            topn_cosine.append(result[i][1])
+
+        topn_cosine = np.array(topn_cosine)
+
+        norm = np.sum(topn_cosine)
```
gojomo: `norm` isn't a good name here, as it usually means something other than a sum. 

But, there's not really a need to loop-append, convert-to-np-array, or put the sum calculation in a local variable. The sum can be a short, idiomatic calculation at the place where it's needed as the denominator of the final return-value calculation, for example just: `sum(result[1] for result in sims)`.

gojomo:
I've added some line-by-line comments; there should also be a unit test method that confirms expected results in some minimal way.

Automatic are failing, but that appears unrelated to your changes - @menshikh-iv, it again looks related to some doc-building command.

rsdel2007:
Yeah, sure I will make the changes.

gensim/models/keyedvectors.py
rsdel2007: Done.

gensim/models/keyedvectors.py
rsdel2007: Yeah...Thanks:).

gensim/models/keyedvectors.py
rsdel2007: Done.

gensim/models/keyedvectors.py
rsdel2007: Okay...Done.

gensim/models/keyedvectors.py
rsdel2007: Done. `sims` is used as variable name.

gensim/models/keyedvectors.py
rsdel2007: Done.

rsdel2007:
@gojomo can you help me, how to write unit test?

gojomo:
It is almost always better to cut & paste actual error text than to use a screenshot - it ensures the text will appear in search results, etc.

Are you sure you're executing the tests inside the same active environment (finding the same working copy of gensim) as where you added `relative_cosine_similarity()`? I presume you've run the method separately in your own ad-hoc tests... how did you run it then? Can you still run it there?

horpto:
@rsdel2007 
Add your tests into PR even they are not completed successfully.

rsdel2007:
@horpto ,@gojomo I have added the test. Please me tell what changes should I do?

gojomo:
The test is not succeeding - see for example https://travis-ci.org/RaRe-Technologies/gensim/jobs/472632483

But, you absolutely need to figure out locally how to run the tests without the `has no attribute 'relative_cosine_similarity'`error appearing. Without being at your system, I don't have further ideas, except to say that the error is suggestive that the `python`/library-search-directories that you're running `test_keyedvectors.py` with is *not* your development python-env, but some other one, such as the system default. (You should be using separate envs, as are typically created by `virtualenv` or `venv`, to keep projects' libraries separate during development.) You could try looking up other unit-test tutorials or virtual-env tutorials that may be custom to the "visual studio" tool you appear to be using. 

The unit-test needs to pass and **make sense** according to the supposed benefits of this new calculation, ideally by matching the claims/examples of the paper in which it originated.

rsdel2007:
Thanks @gojomo I figure out the problem.
One problem which I am getting while preparing for the unit test is that results for `topn` similar words to a given word are not same for `most_similar` and `wordnet`. That's why I am getting some differences from the paper.
So according to you what should I do?

gojomo:
Can you fashion a test which probes/demonstrates the same advantages the paper claims for this measure, even if the unit-test environment only has access to much smaller corpuses/vector-sets?

rsdel2007:
@gojomo I have added the test.

rsdel2007:
@gojomo @menshikh-iv, Please take a look.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,40 @@ def init_sims(self, replace=False):
...
+         word for which we have to look top-n similar word.
+        wb: str
+         word for which we evaluating relative cosine similarity with wa.
+        topn: int, optional
+         Number of top-n similar words to look with respect to wa.
+        Returns
+        -------
+        numpy.float64
+            relative cosine similarity between wa and wb.
+        """
```
piskvorky: PEP 257: no blank line aftere doc string.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,40 @@ def init_sims(self, replace=False):
...
 
+    def relative_cosine_similarity(self, wa, wb, topn=10):
+        """Compute the relative cosine similarity between two words given top-n similar words,
+        proposed by Artuur Leeuwenberg, Mihaela Vela, Jon Dehdari, Josef van Genabith
+        "A Minimally Supervised Approach for Synonym Extraction with Word Embeddings"
+        <https://ufal.mff.cuni.cz/pbml/105/art-leeuwenberg-et-al.pdf>.
+        To calculate relative cosine similarity between two words, equation (1) of the paper is used.
+        For WordNet synonyms, if rcs(topn=10) is greater than 0.10 then wa and wb are more similar than
+        any arbitrary word pairs.
+        Parameters
```
piskvorky: Dtto.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,40 @@ def init_sims(self, replace=False):
...
+        ----------
+        wa: str
+         word for which we have to look top-n similar word.
+        wb: str
+         word for which we evaluating relative cosine similarity with wa.
+        topn: int, optional
+         Number of top-n similar words to look with respect to wa.
+        Returns
```
piskvorky: Dtto.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,40 @@ def init_sims(self, replace=False):
             else:
                 self.vectors_norm = (self.vectors / sqrt((self.vectors ** 2).sum(-1))[..., newaxis]).astype(REAL)
 
+    def relative_cosine_similarity(self, wa, wb, topn=10):
+        """Compute the relative cosine similarity between two words given top-n similar words,
+        proposed by Artuur Leeuwenberg, Mihaela Vela, Jon Dehdari, Josef van Genabith
+        "A Minimally Supervised Approach for Synonym Extraction with Word Embeddings"
+        <https://ufal.mff.cuni.cz/pbml/105/art-leeuwenberg-et-al.pdf>.
+        To calculate relative cosine similarity between two words, equation (1) of the paper is used.
```
piskvorky: Please use a blank line to separate blocks of comments / code.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,40 @@ def init_sims(self, replace=False):
...
+        For WordNet synonyms, if rcs(topn=10) is greater than 0.10 then wa and wb are more similar than
+        any arbitrary word pairs.
+        Parameters
+        ----------
+        wa: str
+         word for which we have to look top-n similar word.
```
piskvorky: Indentation should be 4 spaces (like the rest of the file).

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,40 @@ def init_sims(self, replace=False):
...
+        topn: int, optional
+         Number of top-n similar words to look with respect to wa.
+        Returns
+        -------
+        numpy.float64
+            relative cosine similarity between wa and wb.
+        """
+
+        sims = self.similar_by_word(wa, topn)
+        rcs = (self.similarity(wa, wb)) / (sum(result[1] for result in sims))
```
piskvorky: Is it guaranteed that `sims` is not empty? Otherwise this line will throw an exception.

gensim/models/keyedvectors.py
rsdel2007: Done

gensim/models/keyedvectors.py
rsdel2007: Done.

gensim/models/keyedvectors.py
rsdel2007: Done.

gensim/models/keyedvectors.py
rsdel2007: Done.

gensim/models/keyedvectors.py
rsdel2007: Done.

gensim/models/keyedvectors.py
rsdel2007: `sims` is not empty, since `relative_cosine_similarity` is called with arguments `wa,wb` and `topn` and 'wa' is used to compute `sims`.

piskvorky: I'm not sure what that means or how that follows, but if a non-obvious condition is known not to happen, it's a good practice to make it explicit: `assert sims`.

This makes the assumptions clearer to humans reading the code, as well as to stop errors silently propagation if the assumption is broken (perhaps due to some future refactoring).

gojomo: @rsdel I suspect there's a corner case – such as when the full vector set has only one word – where `sims` could be empty... but then the failure of `similarity(wa, wb)` should give a meaningful error.

gensim/test/test_keyedvectors.py
```diff
@@ -104,6 +104,27 @@ def test_most_similar_topn(self):
...
+        'skillful', 'skilful', 'dear', 'near', 'dependable', 'safe', 'secure', 'right', 'ripe', 'well',
+         'effective', 'in_effect', 'in_force', 'serious', 'sound', 'salutary', 'honest', 'undecomposed',
+          'unspoiled', 'unspoilt', 'thoroughly', 'soundly']   # synonyms for "good" as per wordnet
+        cos_sim = []
+        for i in range(len(wordnet_syn)):
+            if wordnet_syn[i] in self.vectors.vocab:
+                cos_sim.append(self.vectors.similarity("good", wordnet_syn[i]))
+        cos_sim = sorted(cos_sim, reverse=True)  # cosine_similarity of "good" with wordnet_syn in decreasing order
+        # computing relative_cosine_similarity of two similar words
+        rcs_wordnet = self.vectors.similarity("good", "nice") / sum(cos_sim[i] for i in range(10))
```
gojomo: I'm not sure what this is calculating. It's kind of like the `relactive_cosine_similarity()` formula, but now with only WordNet synonyms as contributors to the denominator. And, only those synonyms which happen to be in this vector-set. Are all those words in the `euclidean_vectors.bin` test vectors set? As a result, I'm not sure what the following asserts really test. Is this matching something in the paper?

gensim/test/test_keyedvectors.py
```diff
@@ -104,6 +104,27 @@ def test_most_similar_topn(self):
...
+        rcs = self.vectors.relative_cosine_similarity("good", "nice", 10)
+        self.assertTrue(rcs_wordnet >= rcs)
+        self.assertTrue(np.allclose(rcs_wordnet, rcs, 0, 0.125))
+        # computing relative_cosine_similarity for two non-similar words
+        rcs = self.vectors.relative_cosine_similarity("good", "worst", 10)
+        self.assertTrue(rcs < 0.10)
```
gojomo: Is `0.10` an important threshold from the paper, or just chosen because it works? Is this sort of contrast – between a word `good` and a near-antonym `worst` – the sort of thing RCS is supposed to be good for?

gensim/test/test_keyedvectors.py
rsdel2007: Actually, this is the problem which I found while making test.There is not any **claim or any perfect result** in the paper and I can't find any way to confirm on corpus other than `wordnet`, so I think best way will be to compare the `relative_cosine_similarity` of `wordnet` synonyms and `most_similar` ones under a threshold of `0.125`.

Let me give explain the insights of the section _relative_ cosine similarity of the paper_:
1) Construct a set of the top 10 most (cosine) similar words for w1 (called topn in the paper).
2) Calculate a normalized score for each of the words in the topn, by dividing by the sum of the topn cosine similarity scores.

They mostly wanted to know if the most similar word of w1 was a synonym or not, and not a synonym/hypernym etc. They expected that if the most (cosine) similar word is a lot more (cosine) similar than the other words in the topn it is more likely to be a synonym, than if it is only slightly more similar. So this is what the rcs takes into account.
So, they come to conclusion which is the only claim in the paper is that if a **word pair have a rcs greater 0.10 than it is more likely to be an arbitrary pair**.
`0.10` can be used threshold but this result is based on `wordnet` corpus. On a short corpus this result may be more lower. The threshold is nothing but the mean of the cosine_similarities of topn words. So on a short corpus it may be anything less than `0.10`. 

@gojomo ,Can you suggest some better way to test why looking at above description?

gensim/models/keyedvectors.py
rsdel2007: Okay, I will do that.

rsdel2007: I have made the changes.

piskvorky: The change seems unrelated to our discussion here. How are `len(self.vocab)` and `topn` connected to whether `sims` is empty or not?

piskvorky: If you want to test for empty `sims`, why test for some proxy instead of for empty `sims`?

This check may be useful for other purposes (I don't know), but it is too round-about as an invariant check for empty results / pending division by zero.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,42 @@ def init_sims(self, replace=False):
...
+            word for which we evaluating relative cosine similarity with wa.
+        topn: int, optional
+            Number of top-n similar words to look with respect to wa.
+        Returns
+        -------
+        numpy.float64
+            relative cosine similarity between wa and wb.
+        """
+        sims = self.similar_by_word(wa, topn)
+        assert sims, "Cannot generate similar words"
```
piskvorky: Thanks! Please make the message a little more concrete / verbose though.

As a user, if this message `Cannot generate similar words` jumped at me, I wouldn't know what that means.

How about `Failed code invariant: list of similar words must never be empty.` ?

This is of course assuming it truly is an invariant (can never happen, as opposed to "shouldn't happen on most inputs").

rsdel2007: Thanks. I have made the changes.

viplexke:
Hi, according to the paper, precision of rcs should go up when increasing
the similarity threshold, in contrast with plain cosine similarity, which
was the main motive behind rcs. So I thought maybe calculating such a
precision curve for one or two top10 cosine similarity sets could prove the
usefulness. Testing on a few individual pairs may work as well, using the
fact that rcs gives lower values for co-hyponyms and related words (false
positives) than plain cs. I hope that helps. I'll try to give test cases..


2019. jan. 2., Sze 0:28 dátummal Gordon Mohr <notifications@github.com> ezt
írta:

viplexke:
"""test cases in [word, synonym, antonym, co-hyponym, related word] form"""

Here are 5 dictionaries with 5 hierarchical category key each.

dic_mother={"word":"mother", "synonym":"mom", "antonym":"father",
"co-hyponym":"father", "related_word":"birth"}
dic_cautious={"word":"cautious", "synonym":"careful", "antonym":"careless",
"co-hyponym":"shrewd", "related_word":"danger"}
dic_guy={"word":"guy", "synonym":"dude", "antonym":"girl",
"co-hyponym":"bachelor", "related_word":"son"}
dic_white={"word":"white", "synonym":"milky", "antonym":"black",
"co-hyponym":"blue", "related_word":"blank"}
dic_bend={"word":"bend", "synonym":"curl", "antonym":"straighten",
"co-hyponym":"stretch", "related_word":"river"}

For each case, we can calculate 6 values:

(R)CS(x.word, x.synonym) / (R)CS(x.word, x.antonym)
(R)CS(x.word, x.synonym) / (R)CS(x.word, x.co-hyponym)
(R)CS(x.word, x.synonym) / (R)CS(x.word, x.related_word)

where (R)CS is the (relative) cosine similarity and each ratio with RCS
should be higher thatn the respective ratio with CS:

CS(x.word, x.synonym) / CS(x.word, x.antonym) < RCS(x.word, x.synonym) /
RCS(x.word, x.antonym)

I think this shows that it does what it design for, and that it's not
broken. If this fails significantly, then I was wrong :)

viplexke:
" each ratio with RCS should be higher thatn the respective ratio with CS"
- of course I don't mean the universality. The more the better.

rsdel2007:
@viplexke according to paper,
        `rcs(wa,wb,topn)= cs(wa,wb)/(sum of cosine_similarities of top-n similar words to wa)`.
So, `rcs(x.word, x.synonym) /rcs(x.word, x.antonym)` will be equal to `cs(x.word, x.synonym) / cs(x.word, x.antonym)`.
> CS(x.word, x.synonym) / CS(x.word, x.antonym) < RCS(x.word, x.synonym) / RCS(x.word, x.antonym) 
So how can this prove the functionality?

viplexke:
You right, I skipped the sum. The formula is
CS(x.word, x.synonym) / CS(x.word, x.antonym) < RCS(x.word, x.synonym,
dictionary) / RCS(x.word, x.antonym, dictionary)
where *dictionary *corresponds to topn

rsdel2007:
@viplexke Can you please check again. It seems same to me again.

viplexke:
> @viplexke Can you please check again. It seems same to me again.

Sorry, you're right. Looks like I'll have to take my time. 
I'll find some test and modify test_relative_cosine_similarity() accordingly. 
Should I create a pull request (never done such a thing in github) or just copy the code here when I'm ready?

rsdel2007:
@viplexke Since this feature is not merged, you post it here.
wdyt @gojomo ?

viplexke:
> @viplexke Can you please check again. It seems same to me again.

You're right, sorry about that.
I'll find another test and change the test function accordingly. I wouldn't use a fixed similarity threshold (e.g. 0.1) to define synonyms so I'll rather stick to this comparison of cs and rcs.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,42 @@ def init_sims(self, replace=False):
...
+        topn: int, optional
+            Number of top-n similar words to look with respect to wa.
+        Returns
+        -------
+        numpy.float64
+            relative cosine similarity between wa and wb.
+        """
+        sims = self.similar_by_word(wa, topn)
+        assert sims, "Failed code invariant: list of similar words must never be empty."
+        rcs = (self.similarity(wa, wb)) / (sum(result[1] for result in sims))
```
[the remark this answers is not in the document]
piskvorky: Actually, prepend `float` if this is meant to be a float division. Both to avoid potential errors due to integer operands in python2, and to make the intent clear.

Also, can you unpack `result` into named variables, instead of writing `result[1]`?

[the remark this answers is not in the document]
menshikh-iv:
```suggestion
        rcs = float(self.similarity(wa, wb)) / (sum(sim for _, sim in sims))
```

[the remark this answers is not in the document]
piskvorky: @menshikh-iv you have to teach me how to do that :)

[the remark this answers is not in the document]
menshikh-iv: @piskvorky 

image

gensim/models/keyedvectors.py
```diff
@@ -195,6 +195,7 @@ class Vocab(object):
     and for constructing binary trees (incl. both word leaves and inner nodes).
 
     """
+
```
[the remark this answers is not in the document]
rsdel2007: Done.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,42 @@ def init_sims(self, replace=False):
             else:
                 self.vectors_norm = (self.vectors / sqrt((self.vectors ** 2).sum(-1))[..., newaxis]).astype(REAL)
 
+    def relative_cosine_similarity(self, wa, wb, topn=10):
+        """Compute the relative cosine similarity between two words given top-n similar words,
+        proposed by Artuur Leeuwenberg, Mihaela Vela, Jon Dehdari, Josef van Genabith
```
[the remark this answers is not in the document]
rsdel2007: Okay..Done.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,42 @@ def init_sims(self, replace=False):
...
+        <https://ufal.mff.cuni.cz/pbml/105/art-leeuwenberg-et-al.pdf>.
+
+        To calculate relative cosine similarity between two words, equation (1) of the paper is used.
+        For WordNet synonyms, if rcs(topn=10) is greater than 0.10 then wa and wb are more similar than
+        any arbitrary word pairs.
+
+        Parameters
+        ----------
+        wa: str
+            word for which we have to look top-n similar word.
```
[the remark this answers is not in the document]
rsdel2007: Done.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,42 @@ def init_sims(self, replace=False):
...
+        wb: str
+            word for which we evaluating relative cosine similarity with wa.
+        topn: int, optional
+            Number of top-n similar words to look with respect to wa.
```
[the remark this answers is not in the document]
rsdel2007: Done.

gensim/models/keyedvectors.py
```diff
@@ -1384,12 +1387,42 @@ def init_sims(self, replace=False):
...
+        Returns
+        -------
+        numpy.float64
+            relative cosine similarity between wa and wb.
+        """
```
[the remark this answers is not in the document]
rsdel2007: Done.

gensim/models/keyedvectors.py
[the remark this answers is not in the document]
rsdel2007: Done.

gensim/test/test_keyedvectors.py
```diff
@@ -104,6 +104,27 @@ def test_most_similar_topn(self):
         predicted = self.vectors.most_similar('war', topn=None)
         self.assertEqual(len(predicted), len(self.vectors.vocab))
 
+    def test_relative_cosine_similarity(self):
+        """Test relative_cosine_similarity returns expected results with an input of a word pair and topn"""
+        wordnet_syn = ['good', 'goodness', 'commodity', 'trade_good', 'full', 'estimable', 'honorable',
+        'respectable', 'beneficial', 'just', 'upright', 'adept', 'expert', 'practiced', 'proficient',
```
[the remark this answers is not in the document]
rsdel2007: Done.

gensim/models/keyedvectors.py
```diff
@@ -1385,6 +1385,36 @@ def init_sims(self, replace=False):
             logger.info("precomputing L2-norms of word weight vectors")
             self.vectors_norm = _l2_norm(self.vectors, replace=replace)
 
+    def relative_cosine_similarity(self, wa, wb, topn=10):
+        """Compute the relative cosine similarity between two words given top-n similar words,
+        by Artuur Leeuwenberg, ... "A Minimally Supervised Approach for Synonym Extraction with Word Embeddings"
```
menshikh-iv: should be proper link rendered by sphinx, in this case

```
by `Artuur ..... <https://ufal.mff.cuni.cz/pbml/105/art-leeuwenberg-et-al.pdf>`_
```

menshikh-iv:
@gojomo how do you feel about this PR? Looks good for me, waiting approve from you (or notes about what's still should be fixed)

gensim/models/keyedvectors.py
rsdel2007: Done. :+1:

gensim/models/keyedvectors.py
```diff
@@ -1385,6 +1385,35 @@ def init_sims(self, replace=False):
             logger.info("precomputing L2-norms of word weight vectors")
             self.vectors_norm = _l2_norm(self.vectors, replace=replace)
 
+    def relative_cosine_similarity(self, wa, wb, topn=10):
+        """Compute the relative cosine similarity between two words given top-n similar words,
+        by Artuur ... <https://ufal.mff.cuni.cz/pbml/105/art-leeuwenberg-et-al.pdf>.
```
piskvorky: Use full reference please (not this "...").

gensim/models/keyedvectors.py
rsdel2007: Do I have to include another authors and paper name in < >?

menshikh-iv: just replace with this
```
`Artuur Leeuwenberga, Mihaela Velab , Jon Dehdaribc, Josef van Genabithbc "A Minimally Supervised Approach for Synonym Extraction
with Word Embeddings" <https://ufal.mff.cuni.cz/pbml/105/art-leeuwenberg-et-al.pdf>`_.
```

rsdel2007: Thanks :smile:

gensim/models/keyedvectors.py
rsdel2007: Done.

gojomo:
Though the difficulty in demonstrating/testing the value of this calculation has me more doubtful than initially of its value, I suspect the simple implementation is correct, and the test is OK/passing, and there's little risk of harm to users who don't use it, so merging is OK with me!

menshikh-iv:
congratz @rsdel2007 :+1:

Issue #2258, opened 2018-11-05
Integer overflow during `FastText` training with `corpus_file`


joelkuiper:
#### Description

`model = FastText(corpus_file="sentences_norm.txt.gz", workers=14, iter=5, size=200, sg=1, hs=1)`

with the following sizes 

```
2018-11-05 16:57:52,809 : INFO : collected 6532860 word types from a corpus of 4728738902 raw words and 238627116 sentences
2018-11-05 16:57:52,809 : INFO : Loading a fresh vocabulary
2018-11-05 16:58:00,788 : INFO : effective_min_count=5 retains 1887156 unique words (28% of original 6532860, drops 4645704)
2018-11-05 16:58:00,788 : INFO : effective_min_count=5 leaves 4721157112 word corpus (99% of original 4728738902, drops 7581790)
2018-11-05 16:58:07,437 : INFO : deleting the raw counts dictionary of 6532860 items
2018-11-05 16:58:07,615 : INFO : sample=0.001 downsamples 26 most-common words
2018-11-05 16:58:07,615 : INFO : downsampling leaves estimated 3749158657 word corpus (79.4% of prior 4721157112)
2018-11-05 16:58:11,281 : INFO : constructing a huffman tree from 1887156 words
2018-11-05 16:59:36,077 : INFO : built huffman tree with maximum node depth 30
2018-11-05 17:00:17,300 : INFO : estimated required memory for 1887156 words, 1929637 buckets and 200 dimensions: 7871448352 bytes
2018-11-05 17:00:17,398 : INFO : resetting layer weights
2018-11-05 17:01:43,333 : INFO : Total number of ngrams is 1929637
2018-11-05 17:02:11,990 : INFO : training model with 14 workers on 1887156 vocabulary and 200 features, using sg=1 hs=1 sample=0.001 negative=5 window=5
```

yields 

```
Exception in thread Thread-2120:
Traceback (most recent call last):
  File "/home/joelkuiper/anaconda3/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/joelkuiper/anaconda3/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/joelkuiper/anaconda3/lib/python3.6/site-packages/gensim/models/base_any2vec.py", line 175, in _worker_loop_corpusfile
    total_examples=total_examples, total_words=total_words, **kwargs)
[...]

PR #2239, opened 2018-10-22
Fix overflow error for `*Vec` corpusfile-based training

bm371613:
Fix #2258 

Corpus size is currently limited in terms of the total word count to what fits a 32-bit signed integer. This can result in a `OverflowError: value too large to convert to int` for a big corpus. This PR replaces `int` with a `long long` to allow for bigger corpora.

PR #2314, opened 2019-01-02
Fix integer overflow during `FastText` training with `corpus_file`

irinade:
(#2258)

* replace `int` by `long long`

### DIFF
--- a/gensim/models/fasttext_corpusfile.pyx
+++ b/gensim/models/fasttext_corpusfile.pyx
@@ -133,7 +133,7 @@ def train_epoch_sg(
 
     cdef int i, j, k
     cdef int effective_words = 0, effective_sentences = 0
-    cdef int total_effective_words = 0, total_sentences = 0, total_words = 0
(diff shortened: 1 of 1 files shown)


mpenkov:
Thank you for the PR.

@menshikh-iv I think there is an outstanding PR here that covers more cases: https://github.com/RaRe-Technologies/gensim/pull/2239. I think we should merge that PR in favor of this one. Thoughts?

irinade:
Hello !
Do you have an idea of what causes the errors ? Is it linked to the modification proposed ?

mpenkov:
@irinade The build errors on CircleCI look unrelated.

The modification itself looks good, but we already have a PR (#2239) that covers the particular problem and many potential others.

menshikh-iv:
@irinade thanks for PR, error really unrelated, I'm on it in https://github.com/RaRe-Technologies/gensim/tree/fix-build

I agree with @mpenkov, we should accept "wider" PR #2239, which overlaps the current, anyway, thanks for contribution :+1:

PR #2321, opened 2019-01-08
Add "3CosMul" and "3CosAVG" methods for solving analogies to `KeyedVectors`

ckarats:
3CosMul and 3CosAVG(set based vector offset) methods for solving analogies implemented.
Number of Topn similar words added as input in the evaluation of analogies.


ckarats:
3CosMul and 3CosAVG(set based vector offset) methods for solving analogies implemented.
Number of Topn similar words added as input in the evaluation of analogies. Particularly,

-most_similar_cosmul method: restrict_vocab has been added as input  similarly to most_similar method.

-evaluate_word_analogies: Topk has been added as input (by default has been set to one). It is the number of top similar words to be considered a possible successful prediction. It is given as input in most_similar or most_similar_cosmul methods.

-evaluate_word_analogies: Method (string) has been added as input. It is responsible for the way that the analogies will be solved (3CosMul or 3CosAdd by default)

-evaluate_word_analogies_setBased: New method added. It solves the analogies using the 3CosAvg method, which is a set based method as it is using an average offset.

tox.ini
```diff
@@ -6,7 +6,7 @@ platform = linux: linux
            win: win64
 
 [flake8]
-ignore = E12, W503
+ignore = E12, W503, E722, E226, E302, E41, E228, E225, W291, E501, F841, E303, W293, W391
```
menshikh-iv: bad idea: please revert this change and fix all PEP8 issues in code

gensim/models/keyedvectors.py
```diff
@@ -500,6 +500,7 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
             List of words that contribute negatively.
         topn : int, optional
             Number of top-N similar words to return.
+        last: Only for set-Based analogy. The word which belongs in the same pair as the expected word.
```
menshikh-iv: see doc for another parameters, should be
```
param_name : type
    Description
```

gensim/models/keyedvectors.py
```diff
@@ -544,7 +545,12 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
                     all_words.add(self.vocab[word].index)
         if not mean:
             raise ValueError("cannot compute similarity with no input")
-        mean = matutils.unitvec(array(mean).mean(axis=0)).astype(REAL)
+        if last is None:
+            mean = matutils.unitvec(array(mean).mean(axis=0)).astype(REAL)
+        else:
+            if (len(positive)+len(negative))%2 != 0:
+                raise ValueError("wrong input word analogies.. one or more words are missing..")            
+            mean = matutils.unitvec(array(mean).mean(axis=0)/2).astype(REAL)+self.word_vec(last, use_norm=True)
```
menshikh-iv: `mean` not unit length, looks like a bug

gensim/models/keyedvectors.py
```diff
@@ -557,6 +563,119 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
         # ignore (don't return) words from the input
         result = [(self.index2word[sim], float(dists[sim])) for sim in best if sim not in all_words]
         return result[:topn]
+        
+    def evaluate_word_analogies_setBased(self, analogies, restrict_vocab=300000, case_insensitive=True, dummy4unknown=False, topk=1):
```
menshikh-iv: don't use `camelCase` in method names

gensim/models/keyedvectors.py
```diff
@@ -557,6 +563,119 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+        """Compute performance of the model on an analogy test set.
+        This is modern variant of :meth:`~gensim.models.keyedvectors.WordEmbeddingsKeyedVectors.accuracy`, see
+        `discussion on GitHub #1935 <https://github.com/RaRe-Technologies/gensim/pull/1935>`_.
+        The accuracy is reported (printed to log and returned as a score) for each section separately,
```
menshikh-iv: Instead of copy-paster docstring form another method, need to 
- make a reference to the original method
- describe, why the current method different, what's new here, why is that useful

gensim/models/keyedvectors.py
menshikh-iv: btw, if `last` is missing in vocab then what happens?

gensim/models/keyedvectors.py
```diff
@@ -557,6 +563,119 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+            Number of the top similar words we can consider for a successful prediction.
+
+        Returns
+        -------
+        (float, list of dict of (str, (str, str, str))
+            Overall evaluation score and full lists of correct and incorrect predictions divided by sections.
+
+        """
+
+        ok_vocab = [(w, self.vocab[w]) for w in self.index2word[:restrict_vocab]]
```
menshikh-iv: too much copy-paste from the original method, is it possible to refactor both a bit (move "common" part to the hidden internal method)?

gensim/models/keyedvectors.py
```diff
@@ -818,7 +937,7 @@ def nbow(document):
         # Compute WMD.
         return emd(d1, d2, distance_matrix)
 
-    def most_similar_cosmul(self, positive=None, negative=None, topn=10):
+    def most_similar_cosmul(self, positive=None, negative=None, topn=10, restrict_vocab=None):
```
menshikh-iv: why `restrict_vocab` needed here?

menshikh-iv:
ping @ckarats , do you planning to finish PR?

ckarats:
> ping @ckarats , do you planning to finish PR?

Considering your comments i will have resolved every issue until Sunday 27/01

gensim/models/keyedvectors.py
```diff
@@ -640,7 +641,7 @@ def evaluate_word_analogies_setBased(self, analogies, restrict_vocab=300000, cas
                 predicted = None
                 # find the most likely prediction using 3CosAvg set based vector offset) method
                 # Implementation of the set-based method for solving analogies
-                sims = self.most_similar(positive=terms[0:len(terms)-2:2], negative=terms[1:len(terms)-2:2], last=terms[len(terms)-2], topn=topk, restrict_vocab=restrict_vocab)
+                sims = self.most_similar(positive=terms[1:len(terms)-2:2], negative=terms[0:len(terms)-2:2], last=terms[len(terms)-2], topn=topk, restrict_vocab=restrict_vocab)
```
piskvorky: Why start at `1`, why end at `len(terms) - 2`, go with step `2`?

This assumes `len(terms) >= 2`; why should that be?

gensim/models/keyedvectors.py
```diff
@@ -550,7 +551,7 @@ def most_similar(self, positive=None, negative=None, last=None, topn=10, restric
         else:
             if (len(positive)+len(negative))%2 != 0:
                 raise ValueError("wrong input word analogies.. one or more words are missing..")            
-            mean = matutils.unitvec(array(mean).mean(axis=0)/2).astype(REAL)+self.word_vec(last, use_norm=True)
+            mean = matutils.unitvec(2*(array(mean).mean(axis=0))).astype(REAL)+matutils.unitvec(self.word_vec(last, use_norm=True)).astype(REAL)
```
piskvorky: Why this change?

gensim/models/keyedvectors.py
```diff
@@ -1130,16 +1256,22 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
...
                 self.vocab = original_vocab
-                for element in sims:
+
+                for element in sims:                  
                     predicted = element[0].upper() if case_insensitive else element[0]
                     if predicted in ok_vocab and predicted not in ignore:
                         if predicted != expected:
                             logger.debug("%s: expected %s, predicted %s", line.strip(), expected, predicted)
-                        break
+                        if predicted == expected:
```
piskvorky: Why this change?

gensim/models/keyedvectors.py
```diff
@@ -1052,15 +1178,12 @@ def _log_evaluate_word_analogies(section):
             logger.info("%s: %.1f%% (%i/%i)", section['section'], 100.0 * score, correct, correct + incorrect)
             return score
 
-    def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensitive=True, dummy4unknown=False):
+    def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensitive=True, dummy4unknown=False, topk=1, method='3CosAdd'):
```
piskvorky: Is this default backward compatible?

gensim/models/keyedvectors.py
```diff
@@ -1052,15 +1178,12 @@ def _log_evaluate_word_analogies(section):
...
         """Compute performance of the model on an analogy test set.
-
```
piskvorky: -1: easier to read with the comment blocks separated by blank lines.

gensim/models/keyedvectors.py
ckarats: We have (a_i, b_i) word pairs in each row. For 11 pairs, we want to consider the average of the difference of the first ten pairs Σ(b_i - a_i)/10 and add it to the a11, and then using cosine similarity as in the 3CosAdd case try to find the b11 word. So the a_i words are the negative ones and the b_i are the positives. We are checking that we have complete pairs and there is not any word missing.

ckarats: len(terms) >= 2 needs to hold because otherwise it is a single word

gensim/models/keyedvectors.py
ckarats: Because it was mistaken. We are computing the average of all elements considering their weight, which means that if we have 10 pairs the sum will be divided by 20 and as we want the average of the ten differences we then should multiple it by 2 (so that is will be divided by 10)

gensim/models/keyedvectors.py
```diff
@@ -557,6 +564,119 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+                section = {'section': line.lstrip(': ').strip(), 'correct': [], 'incorrect': []}
+            else:
+                if not section:
+                    raise ValueError("Missing section header before line #%i in %s" % (line_no, analogies))
+                try:
+                    if case_insensitive:
+                        terms = [word.upper() for word in line.split()]
+                    else:
+                        terms = [word for word in line.split()]
+                except ValueError:
```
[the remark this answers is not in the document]
ckarats: Same as evaluate_word_analogies method. It is for invalid lines

gensim/models/keyedvectors.py
ckarats: It starts at one because we cosinder the second word in each pair as the positive.

gensim/models/keyedvectors.py
ckarats: break command changed because we want to consider also the top-5 similar words for each case.

gensim/models/keyedvectors.py
ckarats: Fixed

gensim/models/keyedvectors.py
ckarats: Same as the case of 3CosAdd. We would like to consider the 300,000 first words not every word of the vocabulary.

gensim/models/keyedvectors.py
```diff
@@ -557,6 +564,119 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+
+        Returns
+        -------
+        (float, list of dict of (str, (str, str, str))
+            Overall evaluation score and full lists of correct and incorrect predictions divided by sections.
+
+        """
+
+        ok_vocab = [(w, self.vocab[w]) for w in self.index2word[:restrict_vocab]]
+        ok_vocab = {w.upper(): v for w, v in reversed(ok_vocab)} if case_insensitive else dict(ok_vocab)
```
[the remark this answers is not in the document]
ckarats: Because of descending order. We are using restricted_vocab parameter.

gensim/models/keyedvectors.py
ckarats: Unit length used. Fixed. Thanks!

last is always considered to be for the set based case, the first word of the last pair.

gensim/models/keyedvectors.py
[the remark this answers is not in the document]
horpto: oh, I get it. It's due to `In case of multiple case variants of a single word, the vector for the first occurrence
            (also the most frequent if vocabulary is sorted) is taken.`

gensim/models/keyedvectors.py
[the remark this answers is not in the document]
horpto: It's something strange. `utils.to_unicode` must return string so it should have split() and upper

gensim/models/keyedvectors.py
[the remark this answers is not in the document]
ckarats: Exactly

gensim/models/keyedvectors.py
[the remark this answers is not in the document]
ckarats: It is exaclty the same as evaluate_word_analogies method (check lines 1237-1239) 

upper is used only for the case when words are case insensitive

tox.ini
ckarats: Fixed

gensim/models/keyedvectors.py
piskvorky: Thanks, but all of this is too hard to read. Too much effort for maintainters (current and future) to decipher.

Please use comments, named variables and clear invariants:
* If an array *must* contain at least two items by design, use `assert` to make that clear (now it's not clear to me what the code does if that precondition is not fulfilled). If it's a user error instead (not a programmer error), raise a `ValueError`. 
* If some parts of an array hold special meaning, give them a special name / variable. For example, `assert len(terms) > 2`, `target_positive, target_negative = terms[ : -2]`, `positive_words = terms[1 : : 2]  # the second word in each pair is the positive one` etc (just examples).

gensim/models/keyedvectors.py
piskvorky: Sorry, I don't follow. What option does the fall-through-without-break code path solve?

There are too many levels of indentation here, too complicated. Can you factor some of this out into a named function?

gensim/models/keyedvectors.py
```diff
@@ -1130,16 +1254,23 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
...
                 ignore = {a, b, c}  # input words to be ignored
                 predicted = None
-                # find the most likely prediction using 3CosAdd (vector offset) method
-                # TODO: implement 3CosMul and set-based methods for solving analogies
-                sims = self.most_similar(positive=[b, c], negative=[a], topn=5, restrict_vocab=restrict_vocab)
+                # find the most likely prediction using 3CosAdd (vector offset) method or 3CosMul
+
+                if method == '3CosAdd':
+                    sims = self.most_similar(positive=[b, c], negative=[a], topn=topk, restrict_vocab=restrict_vocab)
+                if method == '3CosMul':
```
piskvorky: `elif` here, plus add a default `else` which raises `ValueError` for unknown methods.

gensim/models/keyedvectors.py
ckarats: As it is implemented until now, when there is one of word among the top-k words similar words that belongs to the ok_vocab and not in "ignore" list, then we have a break command and we exit the for-loop. The latter means that we consider only this word and not all the top-k words in the comparison with the expected.

According to the change i ve made, we check if there is a match between the expected word and any of the k words from the top-k similar words list, which is different and in my opinion it is closer to the state of the art of the analogy questions used for the evaluation. I am looking forward to you comment. Thanks a lot!

gensim/models/keyedvectors.py
ckarats: Fixed

gensim/models/keyedvectors.py
ckarats: Fixed

gensim/models/keyedvectors.py
ckarats: Fixed

gensim/models/keyedvectors.py
```diff
@@ -1130,16 +1275,26 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
                 self.vocab = ok_vocab
                 ignore = {a, b, c}  # input words to be ignored
                 predicted = None
-                # find the most likely prediction using 3CosAdd (vector offset) method
-                # TODO: implement 3CosMul and set-based methods for solving analogies
-                sims = self.most_similar(positive=[b, c], negative=[a], topn=5, restrict_vocab=restrict_vocab)
+                # find the most likely prediction using 3CosAdd (vector offset) method or 3CosMul method
+
+                if method == '3CosAdd':
+                    sims = self.most_similar(positive=[b, c], negative=[a], topn=topk, restrict_vocab=restrict_vocab)
```
piskvorky: This doesn't look backward compatible. Whereas before we got **top 5** predictions, now we only get **top 1** prediction.

And if the top 1 prediction happens to be ingored (`ignore` etc), we fail, whereas before we continued on to the first acceptable word.

This seems fundamentally broken, how does this pass our unit tests? Maybe the unit tests need expanding, too.

gensim/models/keyedvectors.py
```diff
@@ -1130,16 +1275,26 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
...
+                elif method == '3CosMul':
+                    sims = self.most_similar_cosmul(positive=[b, c], negative=[a], topn=topk,
+                    restrict_vocab=restrict_vocab)
+                else:
+                    raise ValueError("Unknown method for the Evaluation on the Analogy test set...")
                 self.vocab = original_vocab
+
                 for element in sims:
```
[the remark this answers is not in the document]
ckarats: Fixed

gensim/models/keyedvectors.py
```diff
@@ -1130,16 +1275,26 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
...
                     predicted = element[0].upper() if case_insensitive else element[0]
                     if predicted in ok_vocab and predicted not in ignore:
                         if predicted != expected:
                             logger.debug("%s: expected %s, predicted %s", line.strip(), expected, predicted)
-                        break
+                        # check if the current predicted vector from top-k ones is equal to the expected, if yes break
```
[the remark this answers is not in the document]
ckarats: It is comparing words that correspond to vectors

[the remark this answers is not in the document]
ckarats: Sorry if i didn't define it properly

gensim/models/keyedvectors.py
ckarats: I changed topk to be equal to 5 by default. Now it should be compartilble. 

"The(TooMany)ProblemsofAnalogicalReasoningwithWordVectors, A.Rogers, A.Drozd et al."
"Methodical Evaluation of Arabic Word Embeddings, Mohammed Elrazzaz  et al."
In the papers mentioned above, it is written that e.g. the top-3 means that for a successful prediction the expected one should among the top-3 similar words. 
At your case top-3 and considering your implementation it means that if the first one is a, b or c the it checks the second etc..

ckarats: If you want to change the break part of the code as it used to be and keep only the implementation of 3CosMul and 3CosAvg methods just tell me.

gensim/models/keyedvectors.py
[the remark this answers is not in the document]
piskvorky: Yes, rephrase please. The comment is misleading now.

gensim/models/keyedvectors.py
piskvorky: The default logic for evaluation should stay the same.

Adding an option to "check more than one hit (after ignoring trivial hits of `a`, `b`, `c`)" is welcome, but must not change the existing logic, and must default to the previous default.

ckarats: Fixed.. The logic stays the same.

gensim/models/keyedvectors.py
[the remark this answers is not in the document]
ckarats: Fixed. Comment removed. Logic of the computation remains as it used to be

ckarats:
@piskvorky after the last changes please advice what is next to be done. Thanks!

ckarats:
@piskvorky @menshikh-iv kind reminder. Should I close the PR, otherwise please advice what is need to be done (more unit test etc..). Thanks!

piskvorky:
@ckarats sorry for the wait, we're reviewing internally.

@mpenkov this is a good example of a PR with new contributed functionality. As a "newcomer" to *2vec evaluation, would you understand what this PR brings, *who* should use it, *why* and *how*? Is the accompanying documentation clear?

(in addition to checking the actual newly introduced / modified code, which is a separate concern)

gensim/models/keyedvectors.py
```diff
@@ -1037,7 +1178,7 @@ def _log_evaluate_word_analogies(section):
 
         Parameters
         ----------
-        section : dict of (str, (str, str, str, str))
+        section : dict of (str, list(str, str, str, str...))
```
piskvorky: What does `…` mean? What does `Section given from evaluation.` actually mean?

A concrete example would help here.

gensim/models/keyedvectors.py
```diff
@@ -558,6 +564,134 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+        dummy4unknown : bool, optional
+            If True - produce zero accuracies for tuples with out-of-vocabulary words.
+            Otherwise, these tuples are skipped entirely and not used in the evaluation.
+        topk: int, optional
+            Number of the top similar words we can consider for a successful prediction.
+
+        Returns
+        -------
+        (float, list of dict of (str, (str, str, ...., str))
+            Overall evaluation score and full lists of correct and incorrect predictions divided by sections.
```
piskvorky: Can you clarify the formats? What is `dict of (str, (str, str, ...., str))`, specifically?

A concrete example here would help a lot IMO.

gensim/models/keyedvectors.py
```diff
@@ -814,17 +948,18 @@ def nbow(document):
         # Compute nBOW representation of documents.
         d1 = nbow(document1)
         d2 = nbow(document2)
-
```
piskvorky: Please keep the blank lines that separate code blocks (easier to read).

gensim/models/keyedvectors.py
```diff
@@ -839,6 +974,11 @@ def most_similar_cosmul(self, positive=None, negative=None, topn=10):
             List of words that contribute negatively.
         topn : int, optional
             Number of top-N similar words to return.
+        restrict_vocab : int, optional
+            Optional integer which limits the range of vectors which
+            are searched for most-similar values. For example, restrict_vocab=10000 would
+            only check the first 10000 word vectors in the vocabulary order. (This may be
+            meaningful if you've sorted the vocabulary by descending frequency.)
```
piskvorky: How would a user `sort the vocabulary by descending order`? Let's be more concrete.

gensim/models/keyedvectors.py
```diff
@@ -1052,15 +1193,16 @@ def _log_evaluate_word_analogies(section):
             logger.info("%s: %.1f%% (%i/%i)", section['section'], 100.0 * score, correct, correct + incorrect)
             return score
 
-    def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensitive=True, dummy4unknown=False):
+    def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensitive=True,
+    dummy4unknown=False, topk=5, method='3CosAdd'):
```
piskvorky: Start the params on a new line, with an extra indent level.

gensim/models/keyedvectors.py
```diff
@@ -1081,15 +1223,17 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
         dummy4unknown : bool, optional
             If True - produce zero accuracies for 4-tuples with out-of-vocabulary words.
             Otherwise, these tuples are skipped entirely and not used in the evaluation.
+        topk: int, optional
+            Number of the top similar words we can consider for a successful prediction.
```
piskvorky:
```suggestion
            Consider an analogy task a success if the target word appears among the `topk` most similar
            words (after ignoring any of the input words). Default is `topk=1`, meaning the task is a success
            only if the target words is the top most similar word.
```

gensim/models/keyedvectors.py
piskvorky: Why is `topk=5` the default? AFAIK, the analogy is a success only if it matches the top-1 word.

gensim/models/keyedvectors.py
```diff
@@ -1130,16 +1274,24 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
...
                 predicted = None
-                # find the most likely prediction using 3CosAdd (vector offset) method
-                # TODO: implement 3CosMul and set-based methods for solving analogies
-                sims = self.most_similar(positive=[b, c], negative=[a], topn=5, restrict_vocab=restrict_vocab)
+                # find the most likely prediction using 3CosAdd (vector offset) method or 3CosMul method
+
+                if method == '3CosAdd':
+                    sims = self.most_similar(positive=[b, c], negative=[a], topn=topk, restrict_vocab=restrict_vocab)
+                elif method == '3CosMul':
+                    sims = self.most_similar_cosmul(positive=[b, c], negative=[a], topn=topk,
```
piskvorky: Hanging indent (here and elsewhere): if the line is too long, start the parameters on a new line. See [PEP8 examples of hanging indent](https://www.python.org/dev/peps/pep-0008/#indentation).

gensim/models/keyedvectors.py
ckarats: At line 1081 of the existing code the given input in most_similar method is topn = 5. 
This is what i mean as the default value, which has been retained in my implementation.

gensim/models/keyedvectors.py
ckarats: Fixed

gensim/models/keyedvectors.py
ckarats: Fixed

gensim/models/keyedvectors.py
```diff
@@ -558,6 +564,134 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
         result = [(self.index2word[sim], float(dists[sim])) for sim in best if sim not in all_words]
         return result[:topn]
 
+    def evaluate_word_analogies_set_based(self, analogies, restrict_vocab=300000,
+    case_insensitive=True, dummy4unknown=False, topk=5):
```
mpenkov: Please fix indentation here

gensim/models/keyedvectors.py
```diff
@@ -814,17 +948,18 @@ def nbow(document):
...
 
-    def most_similar_cosmul(self, positive=None, negative=None, topn=10):
+    def most_similar_cosmul(self, positive=None, negative=None, topn=10, restrict_vocab=None):
         """Find the top-N most similar words, using the multiplicative combination objective,
         proposed by `Omer Levy and Yoav Goldberg "Linguistic Regularities in Sparse and Explicit Word Representations"
         <http://www.aclweb.org/anthology/W14-1618>`_. Positive words still contribute positively towards the similarity,
         negative words negatively, but with less susceptibility to one large distance dominating the calculation.
         In the common analogy-solving case, of two positive and one negative examples,
         this method is equivalent to the "3CosMul" objective (equation (4)) of Levy and Goldberg.
+        restrict_vocab has been added as an optional parameter,
```
mpenkov: I don't think it's relevant from the POV of the user that this parameter "has been added". The link to the pull request is helpful, but not ideal. The documentation needs to be self-contained, if there's anything relevant in the PR that belongs in the docs, it should be there directly. Otherwise, the link is unnecessary, in my opinion.

gensim/models/keyedvectors.py
mpenkov: An example would be really helpful here.

gensim/models/keyedvectors.py
```diff
@@ -1081,15 +1223,17 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
         dummy4unknown : bool, optional
             If True - produce zero accuracies for 4-tuples with out-of-vocabulary words.
             Otherwise, these tuples are skipped entirely and not used in the evaluation.
+        topk: int, optional
+            Number of the top similar words we can consider for a successful prediction.
+        method: str, by default equal to 3CosAdd.
```
```suggestion
        method: str, optional
```

You should also mention the default value (3CosAdd) as well as any other acceptable values below.

gensim/models/keyedvectors.py
```diff
@@ -1130,16 +1274,24 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
...
-                sims = self.most_similar(positive=[b, c], negative=[a], topn=5, restrict_vocab=restrict_vocab)
+                # find the most likely prediction using 3CosAdd (vector offset) method or 3CosMul method
+
+                if method == '3CosAdd':
+                    sims = self.most_similar(positive=[b, c], negative=[a], topn=topk, restrict_vocab=restrict_vocab)
+                elif method == '3CosMul':
+                    sims = self.most_similar_cosmul(positive=[b, c], negative=[a], topn=topk,
+                    restrict_vocab=restrict_vocab)
+                else:
+                    raise ValueError("Unknown method for the Evaluation on the Analogy test set...")
```
mpenkov: This exception could be more helpful:

```suggestion
                    raise ValueError("expected method %r to be one of %r" % method, _ACCEPTED_METHODS)
```

gensim/models/keyedvectors.py
```diff
@@ -558,6 +564,134 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+        the average of the difference between every given pair of words and then we add it
+        in the source word of the last pair trying to predict the target word of the last pair.
+        This method corresponds to the `compute-accuracy` script of the original C word2vec.
+        See also `3CosAvg <https://www.aclweb.org/anthology/C/C16/C16-1332.pdf>`_. For more
+        details see  on GitHub #2321 <https://github.com/RaRe-Technologies/gensim/pull/2321>`_.
+
+        Parameters
+        ----------
+        analogies : str
+            Path to file, where lines are tuples of words, split into sections by ": SECTION NAME" lines.
```
mpenkov: Is it possible to include a few example lines here? If yes, then we should do so, and save the reader some effort.

gensim/models/keyedvectors.py
```diff
@@ -558,6 +564,134 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+            See `gensim/test/test_data/questions-words.txt` as example.
+        restrict_vocab : int, optional
+            Ignore all tuples containing a word not in the first `restrict_vocab` words.
+            This may be meaningful if you've sorted the model vocabulary by descending frequency (which is standard
```
mpenkov: I think this is _only_ meaningful if the vocab is sorted, because otherwise, you're grabbing words in the order they were added.

gensim/models/keyedvectors.py
```diff
@@ -558,6 +564,134 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+            Overall evaluation score and full lists of correct and incorrect predictions divided by sections.
+
+        """
+        ok_vocab = [(w, self.vocab[w]) for w in self.index2word[:restrict_vocab]]
+        ok_vocab = {w.upper(): v for w, v in reversed(ok_vocab)} if case_insensitive else dict(ok_vocab)
+        oov = 0
+        logger.info("Evaluating word analogies for top %i words in the model on %s", restrict_vocab, analogies)
+        sections, section = [], None
+        tuples = 0
+        for line_no, line in enumerate(utils.smart_open(analogies)):
```
mpenkov: This method is doing too much. You have a parser for the analogies file integrated with logic for calculating analogies, performing case normalization, and other things.

This is bad for several reasons:

1. The code is not reusable
2. The code is difficult to test (BTW, where are the unit tests for this functionality? e.g. parsing the analogies file.
3. The function is long, nested and difficult to keep in reader's memory in one piece
4. Poor separation of concerns

My recommendation: refactor this. Start by splitting out the parser logic (e.g. splitting the file into sections).

mpenkov:
> @mpenkov this is a good example of a PR with new contributed functionality. As a "newcomer" to *2vec evaluation, would you understand what this PR brings, _who_ should use it, _why_ and _how_? Is the accompanying documentation clear?

@piskvorky Yes, I agree that this is a good example.

Unfortunately, no, most of what you mention isn't obvious to me. The title is descriptive, but the documentation in the commit could be better. For example, we're adding a new parameter to most_similar. We need to explain what "set-based Analogy" means. Examples would be good.

Another question for @piskvorky: do we really want to grow the KeyedVectors class any further by adding new methods like evaluate_word_analogies_set_based? That class (and module) is already bloated enough as it is. Could (should?) this new stuff go into a separate module? I'm not familiar with the functionality, so I'm just thinking out loud here.

Also, as a general comment for @ckarats, the current documentation reads like it was written from the POV of the author, e.g. feature X was added in PR Y to enable Z. This is natural, but is less helpful to the reader. Try to put yourself in the reader's shoes: what could they be trying to achieve with your function? What do they need to know? Remember, your audience is going to be *all* users of gensim, not just those who are used to the old functionality. So, spend less time explaining the difference with the old version, and focus on explaining what the code does as a whole.

> (in addition to checking the actual newly introduced / modified code, which is a separate concern)

I skimmed through the changes. They look OK to me, but we could do better with 1) improving code readability and refactoring and 2) improving documentation based on the points I mentioned above, and 3) implementing unit tests: I don't see any tests for the new functionality, am I missing something here?

ckarats:
@mpenkov i m going to add some unit test. Moreover, i m goint to proceed with the requested changes. Thanks!

gensim/models/keyedvectors.py
ckarats: @mpenkov the logic of the implementation is the same as for the quadruples analogy questions. Could you please clarify what do you mean by "splitting the file into sections" and explain what is the difference with the case of the quadruples (not set based computations). Thanks.

mpenkov:
@ckarats What is the status of this PR? Are you able to finish it?

ckarats:
@mpenkov i will finish it asap. It will be completed until next Sunday. Thanks!

PR #2326, opened 2019-01-10
Improve `KeyedVector.wmdistance`

horpto:
- bring out check on docset1 from inner cycle
- use symmetry of the metric to reduce calculations (actually in original code `distance_matrix` was not symmetrical sometimes although it should be by definition; `pyemd` forgives this)


menshikh-iv:
Thank you @horpto, you are top commiter in release :star:

PR #2331, opened 2019-01-13
Remove unimportant warning from `Phrases`

### LINKED ISSUE #2325: Unimportant 'Warning' in phrases.py
Upon 1st use of `PhrasesTransformation.__getitem__()`, there's a UserWarning issued "For a faster implementation, use the gensim.models.phrases.Phraser class". See:

https://github.com/RaRe-Technologies/gensim/blob/7fca36f3eb0df70e5f823e2fb4852193b4ad7393/gensim/models/phrases.py#L655

It's kind of excessive/alarming to display a "warning" to deliver a low-severity usage recommendation - thus leading to user inquiries like this at SO: https://stackoverflow.com/questions/54112456/warnings-in-gensim-phrasestransformer

I suggest no "warning" here, as there's nothing wrong or especially concerning. Subtle efficiency tips should go elsewhere.

rsdel2007:
Fixes #2325.
As @gojomo mentioned `warning` message is shown unnecessarily and it should be removed.

### DIFF
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -39,7 +39,6 @@ Examples
 import sys
 import os
 import logging
-import warnings
 from collections import defaultdict
 import functools as ft
 import itertools as it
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Thank you @rsdel2007, you are very active last month, awesome :star:

PR #2340, opened 2019-01-19
Optimize `FastText.load_fasttext_model`

### LINKED ISSUE #1261: Improve FastText loading times

mpenkov:
Should fix #1261


gensim/models/fasttext.py
```diff
@@ -754,7 +761,7 @@ def __contains__(self, word):
         return self.wv.__contains__(word)
 
     @classmethod
-    def load_fasttext_format(cls, model_file, encoding='utf8'):
+    def load_fasttext_format(cls, model_file, encoding='utf8', fast=False):
```
menshikh-iv: I guess this should be named a bit different: `full_model` for example, `fast` is very multivalued (here and everywhere)

gensim/models/doc2vec_inner.c
```diff
@@ -859,7 +859,7 @@ static const char *__pyx_f[] = {
 #endif
 
 
-/* "../../.virtualenvs/release/local/lib/python2.7/site-packages/Cython/Includes/numpy/__init__.pxd":776
+/* "../../envs/gensim/lib/python3.7/site-packages/Cython/Includes/numpy/__init__.pxd":776
```
menshikh-iv: revert back all unrelated `*.c` please (only `_utils_any2vec.c` needed in current PR)

gensim/models/_utils_any2vec.pyx
```diff
@@ -44,6 +50,31 @@ cpdef ft_hash(unicode string):
     return h
 
 
+cpdef ft_hash_bytes(bytes bytez):
```
menshikh-iv: why not `ft_hash` (just replace previous impl)?

gensim/test/test_fasttext.py
```diff
@@ -59,6 +59,21 @@ def setUp(self):
         self.test_model = FT_gensim.load_fasttext_format(self.test_model_file)
         self.test_new_model_file = datapath('lee_fasttext_new')
 
+    def test_native_partial_model(self):
```
menshikh-iv: need to properly test new hash/ngram functions too

gensim/test/test_utils.py
```diff
@@ -269,20 +269,26 @@ def hash_main(alg):
     import six
 
     assert six.PY3, 'this only works under Py3'
+    assert gensim.models.utils_any2vec.FAST_VERSION == 0, 'Cython extensions are broken'
```
menshikh-iv: 0 is fine, -1 is broken

gensim/models/utils_any2vec.py
```diff
@@ -88,44 +102,80 @@ def _ft_hash_py_broken(string):
...
+    #
+    # This is a trick to avoid encoding each ngram separately, which is
+    # computationally expensive.  It works for two reasons:
+    #
+    # 1) ngrams are guaranteed to _not_ have any spaces, because that is the
+    # character we use to split sentences into words.
+    # 2) space is an ASCII character, so it survives UTF-8 encoding unchanged.
+    #
+    encoded_ngrams = (" ".join(ngrams)).encode("utf-8").split()
+    hashes = [_ft_hash_bytes(en) % num_buckets for en in encoded_ngrams]
```
menshikh-iv: have no sense to combine both in single func in that case

gensim/test/test_utils.py
mpenkov: Yes, so if FAST_VERSION == 0, the assertion passes. If FAST_VERSION == -1, then the assertion fails with the specified message.

gensim/models/utils_any2vec.py
mpenkov: It helps to avoid code duplication. We perform these operations in around a dozen places, so it makes sense to extract them into a separate function instead of copy-pasting them.

gensim/models/_utils_any2vec.pyx
mpenkov: Replaced

gensim/test/test_utils.py
menshikh-iv: Fast version can be eq 1 for example, better to >= 0

gensim/models/utils_any2vec.py
```diff
@@ -88,44 +102,80 @@ def _ft_hash_py_broken(string):
...
+    ngrams = compute_ngrams(word, minn, maxn)
+    #
+    # This is a trick to avoid encoding each ngram separately, which is
+    # computationally expensive.  It works for two reasons:
+    #
+    # 1) ngrams are guaranteed to _not_ have any spaces, because that is the
+    # character we use to split sentences into words.
+    # 2) space is an ASCII character, so it survives UTF-8 encoding unchanged.
+    #
+    encoded_ngrams = (" ".join(ngrams)).encode("utf-8").split()
```
horpto: I'm not sure what's better:
- such one liner to encode in bytes or
- just call encode for each str in `_ft_hash`

mpenkov: The benchmarks suggest that the former is better, don't they?

https://github.com/RaRe-Technologies/gensim/issues/1261#issuecomment-455790163

gensim/test/test_utils.py
mpenkov: Should we change this everywhere? There are other modules that rely on FAST_VERSION.

If we change it, may as well make it a boolean, right?

gensim/models/utils_any2vec.py
horpto: yes, they do. I just wonder why it's there.

mpenkov: Why do you think encoding is not a bottleneck? In my opinion, that's exactly where the problem is.

The new version does one join, one encode, and one split per word.

The older version did an encode per ngram. There can be many more ngrams than words, especially for longer words. Furthermore, the ngrams overlap, meaning you're unnecessarily encoding the same character multiple times. This optimization wasn't my idea, but my limited understanding suggests that avoiding the unnecessary encode calls is where the win comes from.

@menshikh-iv Can you please comment?

menshikh-iv: 1. This line isn't a bottleneck according to profiling result
2. I'm only worried about "terminator" here (not sure, but some wrong stuff happens and `" "` potentially can be part of token (though shouldn't be), for this reason, I guess, better to generate ngrams from bytes (same as FB) + use hashing in same manner (if ngram found - calc hash now, don't collect all of this)

gensim/models/_utils_any2vec.pyx
```diff
@@ -7,6 +7,10 @@
 
 """General functions used for any2vec models."""
 
+cdef extern from "stdint_wrapper.h":
```
menshikh-iv: worth to comment (why we do that), I mean directly in code

gensim/models/_utils_any2vec.pyx
```diff
@@ -21,26 +25,28 @@ cdef _byte_to_int_py2(b):
...
     """
-    cdef unsigned int h = 2166136261
-    for c in string.encode("utf-8"):
-        h = np.uint32(h ^ np.uint32(np.int8(_byte_to_int(c))))
-        h = np.uint32(h * np.uint32(16777619))
+    cdef uint32_t h = 2166136261
+    cdef char b
+
+    for b in bytez:
+        h = h ^ <uint32_t>(<int8_t>b)  # FIXME I drop 'ord' from py2, not sure about correctenss
```
menshikh-iv: what's about `ord`?

gensim/models/fasttext.py
```diff
@@ -704,6 +704,13 @@ def train(self, sentences=None, corpus_file=None, total_examples=None, total_wor
             >>> model.train(sentences, total_examples=model.corpus_count, epochs=model.epochs)
 
         """
+        cant_train = hasattr(self.trainables, 'syn1neg') and self.trainables.syn1neg is None
+        if cant_train:
+            raise ValueError(
+                'this model cannot be trained any further, '
+                'if this is a native model, try loading it with full_model=True'
```
menshikh-iv: `with ....` -> `with FastText.load_fasttext_format(path, full_model=True)`

gensim/models/utils_any2vec.py
```diff
@@ -88,44 +102,80 @@ def _ft_hash_py_broken(string):
 
 try:
     from gensim.models._utils_any2vec import (
-        ft_hash as _ft_hash_cy,
+        compute_ngrams,
```
menshikh-iv: I guess we also need byte-based compute_ngrams (as "glued" func)

gensim/test/test_fasttext.py
```diff
@@ -59,6 +59,21 @@ def setUp(self):
...
+        """Can we skip loading the NN and still get a working model?"""
```
menshikh-iv: still missing needed tests (hash func, new ngram)

gensim/test/test_utils.py
piskvorky: > may as well make it a boolean
Definitely not. The particular version is important for debugging.

gensim/models/_utils_any2vec.pyx
menshikh-iv: ping @mpenkov

gensim/models/utils_any2vec.py
mpenkov: OK, I copied your port of the FB implementation.

gensim/models/_utils_any2vec.pyx
mpenkov: it doesn't affect the results of the unit tests, so I think we can forget about it.

gensim/models/_utils_any2vec.pyx
```diff
@@ -92,3 +92,58 @@ cpdef compute_ngrams(word, unsigned int min_n, unsigned int max_n):
...
+    `Original implementation <https://github.com/facebookresearch/fastText/blob/7842495a4d64c7a3bb4339d45d6e64321d002ed8/src/dictionary.cc#L172>`__
+
+    """
+    cdef bytes utf8_word = f'<{word}>'.encode("utf-8")
+    cdef int num_bytes = len(utf8_word)
+    cdef int j, i, n
+
+    ngrams = []
+    for i in range(num_bytes):
+        ngram = []
```
horpto: this is better to move down after check on `_MB_START`

gensim/models/_utils_any2vec.pyx
```diff
@@ -92,3 +92,58 @@ cpdef compute_ngrams(word, unsigned int min_n, unsigned int max_n):
...
+
+        j, n = i, 1
+        while j < num_bytes and n <= max_n:
+            ngram.append(utf8_word[j])
+            j += 1
+            while j < num_bytes and (utf8_word[j] & _MB_MASK) == _MB_START:
+                ngram.append(utf8_word[j])
+                j += 1
+            if n >= min_n and not (n == 1 and (i == 0 or j == num_bytes)):
+                ngrams.append(bytes(ngram))
```
horpto: can we use slice of `utf8_word` instead of list ngram?
i.e. something like `utf8_word[i:j]`

gensim/models/fasttext.py
```diff
@@ -704,6 +708,14 @@ def train(self, sentences=None, corpus_file=None, total_examples=None, total_wor
             >>> model.train(sentences, total_examples=model.corpus_count, epochs=model.epochs)
 
         """
+        cant_train = hasattr(self.trainables, 'syn1neg') and self.trainables.syn1neg is None
```
horpto: stupid question: what if self.trainables does not have `syn1neg` attr at all, so can model train ?

mpenkov: Yes. I don't see any other code that sets syn1neg to None. So, the new code uses that value to mean "cannot continue training".

If trainables does not have syn1neg at all, it is possible to _start_ training.

gensim/models/_utils_any2vec.pyx
```diff
@@ -92,3 +92,66 @@ cpdef compute_ngrams(word, unsigned int min_n, unsigned int max_n):
...
+            continue
+
+        j, n = i, 1
+        while j < num_bytes and n <= max_n:
+            j += 1
+            while j < num_bytes and _is_utf8_continue(utf8_word[j]):
+                j += 1
+            if n >= min_n and not (n == 1 and (i == 0 or j == num_bytes)):
+                ngram = bytes(utf8_word[i:j])
+                ngrams.append(bytes(ngram))
```
menshikh-iv: `bytes` twice (I guess need to remove `bytes` in current line)

gensim/models/_utils_any2vec.pyx
```diff
@@ -92,3 +92,66 @@ cpdef compute_ngrams(word, unsigned int min_n, unsigned int max_n):
...
+
+
+cdef _is_utf8_continue_py2(b):
+    return ord(b) & _MB_MASK == _MB_START
+
+
+_is_utf8_continue = _is_utf8_continue_py2 if PY2 else _is_utf8_continue_py3
+
+
+cpdef compute_ngrams_bytes(word, unsigned int min_n, unsigned int max_n):
```
horpto: I check this current code on benchmark from @menshikh-iv  and find that it's slower than code with `split+join` (3.7s vs 5.6s). In C-code there are quite a lot of callings of python functions.

```suggestion
cdef unsigned char _MB_MASK = 0xC0
cdef unsigned char _MB_START = 0x80


cpdef compute_ngrams_bytes_my(word, unsigned int min_n, unsigned int max_n):
    """Computes ngrams for a word.
    Ported from the original FB implementation.
    Parameters
    ----------
    word : str
        A unicode string.
    min_n : unsigned int
        The minimum ngram length.
    max_n : unsigned int
        The maximum ngram length.
    Returns:
    --------
    list of bytes
        A list of ngrams, where each ngram is a list of **bytes**.
    See Also
    --------
    `Original implementation <https://github.com/facebookresearch/fastText/blob/7842495a4d64c7a3bb4339d45d6e64321d002ed8/src/dictionary.cc#L172>`__
    """
    cdef bytes utf8_word = f'<{word}>'.encode("utf-8")
    cdef const unsigned char* utf8_word_ptr = utf8_word
    cdef size_t num_bytes = len(utf8_word)
    cdef size_t j, i, n

    ngrams = []
    for i in range(num_bytes):
        if utf8_word_ptr[i] & _MB_MASK == _MB_START:
            continue

        j, n = i, 1
        while j < num_bytes and n <= max_n:
            j += 1
            while j < num_bytes and (utf8_word_ptr[i] & _MB_MASK) == _MB_START:
                j += 1
            if n >= min_n and not (n == 1 and (i == 0 or j == num_bytes)):
                ngrams.append(utf8_word[i:j])
            n += 1
    return ngrams
```

I don't check it on python2, but think that it should works fine. It's better to recheck (bemchark) my code, current code and classic variant (with split+join of str) again.

gensim/test/test_utils.py
```diff
@@ -330,24 +335,148 @@ def setUp(self):
...
+
+    for line in sys.stdin:
+        word = line.rstrip('\n')
+        ngrams = fun(word, minn, maxn)
+        print("%r: %r," % (word, ngrams))
+
+
+class NgramsTest(unittest.TestCase):
+    def setUp(self):
+        self.expected_text = {
```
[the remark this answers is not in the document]
mpenkov: No, by the main function above.

I eyeballed them to check that they are right.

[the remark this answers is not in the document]
menshikh-iv: can you check FB output (just to be sure)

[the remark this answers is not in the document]
menshikh-iv: `./fasttext print-ngrams <model> <word>`

gensim/models/_utils_any2vec.pyx
mpenkov: There are two problems with the above code, both related to Py2:

- fstrings won't work in Python 2
- utf8_word_ptr[i] & _MB_MASK won't work in Python 2, because of str to int comparison. That's why you need to use byte_to_int function.

mpenkov: The benchmarking results look good: 

```
(gensim) misha@cabron:~/git/gensim$ ipython -i bench.py 
Python 3.7.1 (default, Oct 22 2018, 11:21:55) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from gensim.models._utils_any2vec import compute_ngrams_bytes_optimized

In [2]: from gensim.models._utils_any2vec import compute_ngrams_bytes

In [3]: %time benchmark(ngram_func=compute_ngrams_bytes)
CPU times: user 956 ms, sys: 2.66 ms, total: 958 ms
Wall time: 957 ms

In [4]: %time benchmark(ngram_func=compute_ngrams_bytes_optimized)
CPU times: user 614 ms, sys: 103 µs, total: 614 ms
Wall time: 612 ms
```

mpenkov: @horpto Thank you for helping with this!

gensim/test/test_utils.py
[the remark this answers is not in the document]
mpenkov: OK, I've added a test against output of that command.

gensim/models/_utils_any2vec.pyx
horpto: @mpenkov Ok, great!

But I wonder why `bytes(bytez[i:j])` is faster than original `utf8_word[i:j]` =(. `bytes(bytez[i:j])` allocates new memory and make memcpy unlike `utf8_word[i:j]`.
Original code generates `PySequence_GetSlice` call. Maybe it has a more overhead than I'd thouht. Cython has all the cards to generate a more concise call, not general PySequence.

mpenkov:
I benchmarked the model loading in this PR against 3.7.0 using:

```python
import logging

logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(filename)s:%(lineno)s - %(message)s')
m = FastText.load_fasttext_format("cc.ru.300.bin")
```

Before: 13 min
After: 2min 20s

fb-improv

[370.prof.gz](https://github.com/RaRe-Technologies/gensim/files/2790047/370.prof.gz)

[improv.prof.gz](https://github.com/RaRe-Technologies/gensim/files/2790049/improv.prof.gz)

menshikh-iv:
Awesome, thanks @mpenkov and @horpto :+1:

piskvorky:
@mpenkov how much of that time is the loading vs access? (in gensim and in fb)

menshikh-iv:
@piskvorky you are right, all of it is loading, retrieve a vector by word works fast.

Note: I guess reason in "retrieve vectors for vocab" (`adjust_vector` mostly) on loading, FB doesn't do that (construct all vectors on-the-fly, don't precompute vocab), but we do.

PR #2345, opened 2019-01-21
Minor corrections and improvements in `Any*Vec` docstrings

tobycheese:
Fixed some small things in docstrings and in an exception over which I stumbled while making myself familiar with gensim:

* Docstrings still mention 'sentences argument, while in code this is now called 'documents'
* Clarify that when not passing training data, the model is left uninitialized
* A generator is also an iterator. 'Sequence' might be a better term (forgotten to mention that one in my commit msg)


menshikh-iv:
Thanks @tobycheese, congratz with the first contribution :1st_place_medal:

PR #2347, opened 2019-01-21
Add warning when string is used as argument to infer_vector()

tobycheese:
Another one that I just stumbled over.

`infer_vector()` in doc2vec.py requires a list of strings (words). If one does not pay attention and passes a string instead, this does not fail (as strings are lists of characters).

`infer_vector()` then calculates the vector based on the single characters in the string, which most of the time is not what one wants I think. Imho this merits a warning.


gensim/models/doc2vec.py
```diff
@@ -906,6 +907,12 @@ def infer_vector(self, doc_words, alpha=None, min_alpha=None, epochs=None, steps
             The inferred paragraph vector for the new document.
 
         """
+        if isinstance(doc_words, string_types):
+            warnings.warn(
+                "The parameter `doc_words` takes a list of strings, while you passed a single string, "
+                "which will be accepted and treated as a list of characters. Is this what you wanted?"
```
piskvorky: It's never what users wanted. I'd prefer an exception here.

And if users really want to pass in a list of characters, they can pass in a list of characters (`list(string)`, not a string).

piskvorky:
This is a welcome sanity-check, thanks!

menshikh-iv:
@tobycheese thanks for PR, please
- merge latest `develop` into your branch (I merged #2345)
- raise a ValueError instead of warning if user pass string to infer_vector (instead of warning)

After, I'll merge your PR

gojomo:
Only reason for no check previously was that parameter-name/doc-comment were thought clear enough, and perhaps if Python loose typing wants to let a string behave just like a list-of-one-character-strings, why should we complain? (It's not inconcievable that someone would be trying to create a model based on single-char tokens.) But certainly this bites people, and those who need single-char words can just force their strings to be lists, so some sort of warning or error is a good idea.

gensim/models/doc2vec.py
```diff
@@ -908,10 +908,7 @@ def infer_vector(self, doc_words, alpha=None, min_alpha=None, epochs=None, steps
 
         """
         if isinstance(doc_words, string_types):
-            warnings.warn(
-                "The parameter `doc_words` takes a list of strings, while you passed a single string, "
-                "which will be accepted and treated as a list of characters. Is this what you wanted?"
-            )
+            raise ValueError("You must pass a list of strings, not a single string.")
```
gojomo: Perhaps this should be a `TypeError` - because all string-types will cause the error. And naming the method or parameter would make the error easier to interpret, without looking further up the stack dump (which occasionally by user preferences might not even be shown.) For example: "Parameter `doc_words` of `infer_vector()` must be a list of strings (not a single string)."

piskvorky:
Thanks @tobycheese ! A small step for code, one giant leap for user-friendliness :)

tobycheese:
Thanks for commiting, userfriendlyness ftw!

PR #2353, opened 2019-01-25
[WIP] Improve Fasttext documentation

mpenkov:


gensim/models/fasttext.py
```diff
@@ -31,6 +34,15 @@
...
     >>> model = FastText(common_texts, size=4, window=3, min_count=1, iter=10)
+    >>> sentences = [
+    ...     ['computer', 'artificial', 'intelligence'],
+    ...     ['artificial', 'trees'],
+    ...     ['human', 'intelligence'],
+    ...     ['artificial', 'graph'],
+    ...     ['intelligence'],
+    ...     ['artificial', 'intelligence', 'system']
+    ... ]
+    >>> model.train(sentences, total_examples=len(sentences), epochs=model.epochs)
```
piskvorky: Where is this `model.epochs` coming from? The model instantiation above shows no such variable.

gensim/models/fasttext.py
```diff
@@ -41,7 +53,49 @@
...
+
+    >>> new_sentences = [
+    ...     ['sweet', 'child', 'of', 'mine'],
+    ...     ['rocket', 'queen'],
+    ...     ['you', 'could', 'be', 'mine'],
+    ...     ['november', 'rain'],
+    ... ]
+    >>> 'rocket' in model.wv
+    False
+    >>> model.train(new_sentences, total_examples=len(sentences), epochs=model.epochs)
```
piskvorky: Why `total_example=len(sentences)`? Even if correct, it looks strange… my first question as a user.
Let's provide the answer here in the docs.

gensim/models/fasttext.py
```diff
@@ -41,7 +53,49 @@
...
+You may continue training them on new data:
+
+.. sourcecode:: pycon
+
+    >>> 'computer' in fb_full.wv.vocab  # New word, currently out of vocab
+    False
+    >>> 'rocket' in fb_full.wv.vocab
+    False
+    >>> fb_full.train(sentences, total_examples=len(sentences), epochs=model.epochs)
+    >>> fb_full.train(new_sentences, total_examples=len(new_sentences), epochs=model.epochs)
```
piskvorky: Is this `total_examples` optional? Or is it required even for objects that have `__len__`? (like the `list` here)

gensim/models/fasttext.py
```diff
@@ -85,6 +139,28 @@
...
 
+Implementation Notes
+--------------------
+
+These notes may help developers navigate our fastText implementation.
+Our FastText implementation is split across several submodules:
+
+- :py:mod:`gensim.models.fasttext`: This module.  Contains FastText-specific functionality only.
+- :py:mod:`gensim.models.keyedvectors`: Implements both generic and FastText-specific functionality.
+- :py:mod:`gensim.models.word2vec`:
```
piskvorky: Missing text?

gensim/models/fasttext.py
```diff
@@ -85,6 +139,28 @@
...
+- :py:mod:`gensim.models.base_any2vec`:
+- :py:mod:`gensim.models.utils_any2vec`: Wrapper over Cython extensions.
+
+Our implementation relies heavily on inheritance.
+It consists of several important classes:
+
+- :py:class:`FastTextVocab`: the vocabulary.  Redundant, simply wraps its superclass.
```
piskvorky: Doesn't sound like an "important class". Do we need it here?

gensim/models/fasttext.py
```diff
@@ -862,15 +939,44 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=None, case_inse
...
+    """Represents the inner shallow neural network used to train :class:`~gensim.models.fasttext.FastText`.
+
+    Mostly inherits from its parent (:py:class:`gensim.models.word2vec.Word2VecTrainables`).
+    Adds logic for calculating and maintaining ngram weights.
+
+    Attributes
+    ----------
+
+    hashfxn : function
+        Used for randomly initializing weights.  Defaults to the built-in hash()
```
piskvorky: Only one space after punctuation + end sentences in a full stop (here and elsewhere).

gensim/models/fasttext.py
```diff
@@ -862,15 +939,44 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=None, case_inse
...
+        A one-dimensional array with one element for each term in the vocab.  Set in reset_weights to an array of ones.
+    vectors_vocab_lockf : numpy.array
+        Similar to vectors_vocab_lockf, ones(len(model.trainables.vectors), dtype=REAL)
+    vectors_ngrams_lockf : numpy.array
+        np.ones((self.bucket, wv.vector_size), dtype=REAL)
+
+    Notes
+    -----
+
+    The lockf stuff looks like it gets used by the fast C implementation.
```
piskvorky: This sounds like an internal comment, not API docs :)

gensim/models/fasttext.py
```diff
@@ -759,7 +835,8 @@ def load_fasttext_format(cls, model_file, encoding='utf8'):
 
         Notes
         ------
-        Due to limitations in the FastText API, you cannot continue training with a model loaded this way.
+        This function effectively ignores `.vec` output file.
```
piskvorky: What is the `.vec` for, if it's ignored?

gensim/models/fasttext.py
mpenkov: Yes, the epochs parameter is optional, so not included during instantiation.

Unfortunately, there is also some confusion about its name: the FastText constructor uses iter to specify the number of epochs, whereas the superclass uses the proper name epochs.

gensim/models/fasttext.py
mpenkov: It's correct. I agree that is confusing. The docstring for the [train function](https://radimrehurek.com/gensim/models/fasttext.html#gensim.models.fasttext.FastText.train) attempts to clarify the situation.

Personally, I think if neither total_examples and total_words are specified, we should try to determine sensible defaults by looking at e.g. len(sentences). WDYT @menshikh-iv ?

gensim/models/fasttext.py
mpenkov: Currently, it is required.

gensim/models/fasttext.py
piskvorky: Hm. If it's optional, let's not use it in `train`. Or if we use it in `train`, let's instantiate it explicitly. This neither-here-nor-there example is confusing ("where is this value come from?").

Regarding `iter` / `epochs` -- can you please rename it to `epochs`, consistently? I remember some discussion around this (cc @menshikh-iv @gojomo ), but can't imagine why we'd want both. At most we could support `iter` for a while as a synonym, but with a clear deprecation warning.

This is a perfect opportunity to clean up some of the API mess, rather keep piling on.

gensim/models/fasttext.py
piskvorky: And why is it `len(sentences)` above, but `len(new_sentences)` here?

gensim/models/fasttext.py
piskvorky: Are you sure? I read the docs and still don't get why it's not `len(new_sentences)`.

Please include some top-level intuition here. A short sentence on why this parameter is mandatory, and what should be its value, because it looks really strange and superfluous.

gojomo: For `train()` to manage `alpha` correctly, and show meaningful progress estimates, it needs a good estimate of the size of the supplied-corpus - even when the corpus (as an iterable) may not self-report its length. In the typical case where the same corpus was just surveyed for its vocabulary, this value should be handy. (In the Word2Vec case, the count from the vocab-scan is cached inside the model for later consultation – unsure if the FT paths do this.) In the case where other/new data is being supplied, the caller should supply the right counts for the current data.

gensim/models/fasttext.py
mpenkov: It's illustrating training continuation.

The first call trains on sentences. The second call trains on new_sentences.

gensim/models/fasttext.py
mpenkov: I agree regarding the cleanup. My preference would be to leave epochs/iter out of the constructor. The model doesn't need that parameter until training time.

gensim/models/fasttext.py
gojomo: If it's able to self-report its length, in count of texts, then yes, that would work as `total_examples`. But because streaming a corpus from an iterable is the motivating case for this interface, relying on that inside the method seems inappropriate to me. Leaving it the responsibility of the caller seems fine to me – and if they're lucky enough to have a corpus that reports its own length, they can supply it easily.

mpenkov: Also, while we're talking about simplifying the API, what do you think about removing the sentences and corpus_file parameters from the constructor? Currently, we have an inconsistency: in the constructor, we just pass sentences/corpus_file **without** total_examples and total_words parameters.  In the train function, we include those additional parameters.

Instead of passing sentences in the constructor, the user can pass them in separately via the train function.

Pros:

- Simpler constructor.  There will be less parameters.
- Easier to understand and consistent API.

Cons:

- Getting a trained model now takes two steps instead of one (instantiate, train). Not sure how much of a con this is given that the train function is more powerful than the constructor anyway - it has parameters that the constructor doesn't.

@menshikh-iv @piskvorky @gojomo What do you think?

gensim/models/fasttext.py
mpenkov: OK, I added an explanation.

gensim/models/fasttext.py
mpenkov: >  But because streaming a corpus from an iterable is the motivating case for this interface, relying on that inside the method seems inappropriate to me.

@gojomo Why do you think it is inappropriate? We could do something like this:

```python
if total_examples or total_words:
    pass  # nothing to do here
elif sentences and hasattr(sentences, '__len__'):
    total_examples = len(sentences)
elif data_corpus and hasattr(data_corpus, '__len__'):
    total_examples = len(data_corpus)
else:
    raise ValueError(
        'unable to infer total_examples or total_words from the training source, '
        'please pass one of them explicitly'
    )
```

It looks ugly, but it allows to user to do something like:

````python
model.train(sentences)
````

instead of

```python
model.train(sentences, num_examples=len(sentences))
```

I feel the former is more Pythonic.

Finally, I think having two separate keyword parameters for the input is confusing for the user. In my opinion, it would look a lot simpler if we unified the two parameters, and dealt with untangling them in the implementation.

gensim/models/fasttext.py
mpenkov: Added

gensim/models/fasttext.py
mpenkov: OK, I finally understood what the inconsistency was, and fixed it. Thanks.

gensim/models/fasttext.py
mpenkov: Removed. There's a comment below that explains the same thing in more detail anyway

mpenkov:
> @mpenkov looks good, thanks! Do the docstring examples now cover all the confusion we saw from users (mailing list, github)? Any important workflows left, undocumented / under-documented?

It's hard for me to claim we've resolved _all_ the confusion, because I haven't thoroughly read all the feedback from the users. Nevertheless, this PR addresses (or begins to address, owing to the ongoing discussions above) several key problems that we identified in the task description:

1. Unnecessary parameters in the constructor (ongoing, plan to continue unti we reach consensus).
2. IO problems (key, vec, etc).
3. Lack of examples for training continuation.
4. Lack of design documentation.

@menshikh-iv Have I missed anything?

gensim/models/fasttext.py
mpenkov: OK, fixed.

gensim/models/fasttext.py
piskvorky: @mpenkov yes, I'd consider that "sensible" defaults. Thanks.

Agreed on unifying parameters. How would that work? Keep both as acceptable input (internally unify to one), no deprecations? Deprecate one? Which one?

gensim/models/fasttext.py
```diff
@@ -41,7 +53,51 @@
...
+    >>> import numpy as np
+    >>> old_computer = np.copy(model.wv['computer'])  # Grab the existing vector for this word
+    >>> new_sentences = [
+    ...     ['computers', 'expensive'],
+    ...     ['computer', 'chess', 'players', 'stronger', 'than', 'humans'],
+    ...     ['computers', 'are', 'everywhere'],
+    ... ]
+    >>> model.train(new_sentences, total_examples=len(new_sentences), epochs=model.epochs)
+    >>> new_computer = model.wv['computer']
+    >>> # FIXME: why is this True??
```
piskvorky: @mpenkov what's this about?

gensim/models/fasttext.py
piskvorky: Model in Gensim generally allow the `trained_model = Constructor(params_including_training_params)`. So breaking that could be confusing (and a big change to backward incompatibility).

I'm not totally opposed though, especially if we still allow ctr params for a while with "deprecated" warnings. The API needs a clean up, and now is a good time.

gensim/models/fasttext.py
mpenkov: @menshikh-iv and I are still trying to work out why the vectors still match after training. It's likely a broken a example.

gensim/models/fasttext.py
mpenkov: It's not just training parameters you need to include in the constructor. It's also parameters for vocabulary creation. So you're managing at least 3 sets of separate parameters, 2 of which are duplicated by other methods of the class.

gensim/models/fasttext.py
mpenkov: OK. I think it's worth dealing with API refactoring in a separate PR, for two reasons:

1. Such changes risk introducing regressions, and I'd rather not have them together with the relatively safe documentation changes.
2. We've already spent much effort refactoring FastText, and it may be prudent to wait for things to stabilize (e.g. fix introduced regressions) before proceeding.

gensim/models/fasttext.py
```diff
@@ -27,10 +30,87 @@
 
 .. sourcecode:: pycon
 
-    >>> from gensim.test.utils import common_texts
-    >>> from gensim.models import FastText
+    >>> # from gensim.models import FastText  # FIXME: why does Sphinx dislike this import?
```
menshikh-iv: what's a problem with it, can you explain, please?

gensim/models/fasttext.py
```diff
@@ -27,10 +30,87 @@
...
+    >>> from gensim.test.utils import common_texts  # some example sentences
+    >>> print(common_texts[0])
```
menshikh-iv: good idea to add an empty line (`>>> `) between imports and actual commands, here and everywhere

gensim/models/fasttext.py
```diff
@@ -27,10 +30,87 @@
...
+This is OK for smaller datasets, but for larger datasets, we recommend streaming the file,
+for example from disk or the network.
+In Gensim, we refer to such datasets as "corpora" (singular "corpus"), and keep them
+in the format described in :class:`~gensim.models.word2vec.LineSentence`.
+Passing a corpus is simple:
+
+.. sourcecode:: pycon
+
+    >>> from gensim.test.utils import datapath
+    >>> corpus_file = datapath('lee_background.cor')  # absolute path to corpus
```
menshikh-iv: make all examples more readable please (just split by `>>> ` blocks)

gensim/models/fasttext.py
```diff
@@ -771,7 +931,11 @@ def load_fasttext_format(cls, model_file, encoding='utf8', full_model=True):
 
         Notes
         ------
-        Due to limitations in the FastText API, you cannot continue training with a model loaded this way.
+        Facebook provides both `.vec` and `.bin` files with their modules.
```
menshikh-iv: this method deserves a good example
- load FB model + infer + train
- load FB without second matrix & infer

gensim/models/fasttext.py
```diff
@@ -877,15 +1041,43 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=None, case_inse
...
 class FastTextTrainables(Word2VecTrainables):
-    """Represents the inner shallow neural network used to train :class:`~gensim.models.fasttext.FastText`."""
+    """Represents the inner shallow neural network used to train :class:`~gensim.models.fasttext.FastText`.
+
+    Mostly inherits from its parent (:class:`~gensim.models.word2vec.Word2VecTrainables`).
+    Adds logic for calculating and maintaining ngram weights.
+
+    Attributes
+    ----------
+
```
menshikh-iv: no need empty line

gensim/models/fasttext.py
```diff
@@ -987,7 +1179,7 @@ def _load_fasttext_format(model_file, encoding='utf-8', full_model=True):
     encoding : str, optional
         Specifies the file encoding.
     full_model : boolean, optional
-        If False, skips loading the hidden output matrix.  This saves a fair bit
+        If False, skips loading the hidden output matrix. This saves a fair bit
         of CPU time and RAM, but prevents training continuation.
```
menshikh-iv: somehow highlight `prevents training continuation.` (make it bold maybe?)

gensim/models/fasttext.py
```diff
@@ -41,7 +121,63 @@
...
+    >>> model.build_vocab(new_sentences, update=True)  # Update the vocabulary
+    >>> model.train(new_sentences, total_examples=len(new_sentences), epochs=model.epochs)
+    >>> new_vector = model.wv['computation']
+    >>> np.allclose(old_vector, new_vector, atol=1e-4)  # Vector has changed, model has learnt something
+    False
+    >>> 'computation' in model.wv.vocab  # Word is still out of vocab
+    False
+
+.. Important::
+    Be sure to call the :meth:`~gensim.models.fasttext.FastText.build_vocab`
```
menshikh-iv: deserve a description, why this needed?

gensim/models/_fasttext_bin.py
```diff
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 """Load models from the native binary format released by Facebook.
 
+The main entry point is the :func:`~gensim.models._fasttext_bin.load` function.
```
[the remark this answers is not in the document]
mpenkov: Done

gensim/models/fasttext.py
mpenkov: Done

gensim/models/fasttext.py
mpenkov: Done

gensim/models/fasttext.py
mpenkov: Done

gensim/models/fasttext.py
mpenkov: OK, removed

gensim/models/fasttext.py
mpenkov: Bolded

gensim/models/fasttext.py
mpenkov: Added

gensim/models/fasttext.py
piskvorky: Yes, we should promote them as separate steps. Question is, do we deprecate (certainly not *remove*) them from ctr?

gensim/models/keyedvectors.py
```diff
@@ -1943,14 +1943,19 @@ class FastTextKeyedVectors(WordEmbeddingsKeyedVectors):
         If True, uses the Facebook-compatible hash function instead of the
         Gensim backwards-compatible hash function.
 
+    Some important attributes:
```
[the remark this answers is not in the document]
menshikh-iv: > The section doesn't actually show up in the docs, though.

yes (I guess reason in the custom template), but this is not important in the current case, because parameters + descriptions rendered correctly.

I guess just delete `Some ...` line and that's all

gensim/models/fasttext.py
menshikh-iv: aha, instead a current comment please add `# noqa: F401` (this will suppress a message)

gensim/models/fasttext.py
mpenkov: I think the first step should be to deprecate them. After a while, we can remove them, perhaps in time for a major release.

If we want a one-liner way to instantiate and train, we can always write a pure function and promote that.

gensim/models/fasttext.py
mpenkov: Unfortunately, that does not help. flake8 still errors out with the same message. It seems to be ignoring the noqa directive in the comments.

gensim/models/keyedvectors.py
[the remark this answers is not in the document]
mpenkov: OK, done.

gensim/models/fasttext.py
menshikh-iv: missing mention about `update_vocab=True`

menshikh-iv:
@piskvorky LGTM, wdyt?

gensim/models/fasttext.py
mpenkov: OK, added this too

gensim/models/fasttext.py
```diff
@@ -27,10 +30,88 @@
...
+
+You can also pass all the above parameters to the constructor to do everything
+in a single line:
+
+.. sourcecode:: pycon
+
+    >>> model2 = FastText(size=4, window=3, min_count=1, sentences=common_texts, iter=10)
+
+.. Important::
+    We intend to deprecate this second method of passing everything through the constructor.
```
piskvorky:
```suggestion
    This style of initialize-and-train in a single line is **deprecated**. It's included for backward compatibility only. 
    Please use the initialize-build_vocab`-`train` pattern above instead, including using `epochs` instead of `iter`.
```

gensim/models/fasttext.py
```diff
@@ -27,10 +30,88 @@
...
+    >>>
+    >>> corpus_file = datapath('lee_background.cor')  # absolute path to corpus
+    >>> model3 = FastText(size=4, window=3, min_count=1)
+    >>> model3.build_vocab(corpus_file=corpus_file)  # scan over corpus to build the vocabulary
+    >>>
+    >>> total_examples = model3.corpus_count  # number of sentences in the corpus
+    >>> total_words = model3.corpus_total_words  # number of words in the corpus
+    >>> model3.train(corpus_file=corpus_file, total_examples=total_examples, total_words=total_words, epochs=5)
+
+The model needs the `total_examples` and `total_words` parameters in order to
```
piskvorky: Does it really need both? IIRC, it only needs one of those.

gensim/models/fasttext.py
```diff
@@ -27,10 +30,88 @@
...
+Your iterator should yield a list of strings each time.
+Gensim will take care of the rest:
+
+.. sourcecode:: pycon
+
+    >>> class MyIter:
+    ...     def __iter__(self):
+    ...         with open(datapath('crime-and-punishment.txt')) as fin:
+    ...             for line in fin:
+    ...                 yield line.lower().strip().split(" ")
```
piskvorky: Just `split()` (or we'll be left with newlines and all sorts of unicode "whitespace").

gensim/models/fasttext.py
```diff
@@ -27,10 +30,88 @@
...
+You may calculate them by scanning over the corpus yourself, too.
+
+If you have a corpus in a different format, then you can use it by wrapping it
+in an `iterator <https://wiki.python.org/moin/Iterator>`_.
+Your iterator should yield a list of strings each time.
+Gensim will take care of the rest:
+
+.. sourcecode:: pycon
+
+    >>> class MyIter:
```
piskvorky: `MyIter(object)` (we still support Python 2).

piskvorky:
This looks great! I wish all PRs were this meticulous.

I have some things I'd want to add, but it's nitpicking at this point, probably best if we merge and I'll go over it myself here and there.

gensim/models/fasttext.py
mpenkov: How about we just do:

```python
yield 'one'
yield 'two'
yield 'three'
```

just to illustrate what's going on?

gensim/models/fasttext.py
mpenkov: You're right, I updated the example.

gensim/models/fasttext.py
mpenkov: OK, done.

gensim/models/fasttext.py
mpenkov: OK, I've improved the example with `gensim.utils.tokenize`

menshikh-iv:
Awesome works, great @mpenkov :fire:

PR #2358, opened 2019-01-28
Move out `process_result_queue` from cycle in LdaMulticore

horpto:
+ simplify inner check
+ add heuristic for case with large number of workers

### DIFF
--- a/gensim/models/ldamulticore.py
+++ b/gensim/models/ldamulticore.py
@@ -231,7 +231,8 @@ class LdaMulticore(LdaModel):
         else:
             updatetype = "batch"
             updateafter = lencorpus
-        evalafter = min(lencorpus, (self.eval_every or 0) * updateafter)
+        eval_every = self.eval_every or 0
+        evalafter = min(lencorpus, eval_every * updateafter)
 
         updates_per_pass = max(1, lencorpus / updateafter)
         logger.info(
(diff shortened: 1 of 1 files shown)


horpto:
It's the light version of my try to refactor LdaMulticore https://github.com/horpto/gensim/commit/16fa8fa4ab77c5f754f91be2c22652769a041bfb
I'm not sure should I to continue work on that branch. I've tried to extract work with the queues and pool in the separate entity and then use it as a base for others `*Multicore`.

gensim/models/ldamulticore.py
```diff
@@ -257,47 +258,45 @@ def update(self, corpus, chunks_as_numpy=False):
...
+            merged_new = False
+            while not result_queue.empty():
+                other.merge(result_queue.get())
+                queue_size[0] -= 1
+                merged_new = True
+
+            if (force and merged_new and queue_size[0] == 0) or (other.numdocs >= updateafter):
+                self.do_mstep(rho(), other, pass_ > 0)
+                other.reset()
+                if eval_every > 0 and (force or (self.num_updates / updateafter) % eval_every == 0):
```
horpto: if `force` is true then `queue_size[0] == 0` already checked.

gensim/models/ldamulticore.py
```diff
@@ -257,47 +258,45 @@ def update(self, corpus, chunks_as_numpy=False):
...
+            LDA model if necessary.
+
+            """
+            merged_new = False
+            while not result_queue.empty():
+                other.merge(result_queue.get())
+                queue_size[0] -= 1
+                merged_new = True
+
+            if (force and merged_new and queue_size[0] == 0) or (other.numdocs >= updateafter):
```
horpto: Check on `not self.batch` is redundant as if `self.batch` is truthy so `updateafter` is equal to `len(corpus)`. It means there will not be any temporary mstep, only after all processed docs or in `force` case. It's exactly mean what we want.

PR #2356, opened 2019-01-25
Fix WordEmbeddingsKeyedVectors.most_similar

Witiko:
This PR fixes two issues discovered in #2105:

- The `gensim.models.keyedvectors.WordEmbeddingsKeyedVectors.most_similar` method returns an unexpected result (a numeric array rather than an empty list) when `topn=0`.
- The example code for `gensim.similarities.SoftCosineSimilarity` needs fixing.

PR #2359, opened 2019-01-28
Fix the example code for SparseTermSimilarityMatrix

Witiko:
A follow-up to #2356:

- The example code for `gensim.similarities.SparseTermSimilarityMatrix` needs fixing.


menshikh-iv:
Thank you @Witiko

PR #2360, opened 2019-01-29
Fix light linting issues in `LdaSeqModel`

horpto:


gensim/models/ldaseqmodel.py
```diff
@@ -578,8 +575,7 @@ def doc_topics(self, doc_number):
             Probability for each topic in the mixture (essentially a point in the `self.num_topics - 1` simplex.
 
         """
-        doc_topic = np.copy(self.gammas)
-        doc_topic /= doc_topic.sum(axis=1)[:, np.newaxis]
+        doc_topic = self.gammas / self.gammas.sum(axis=1)[:, np.newaxis]
         return doc_topic[doc_number]
```
menshikh-iv: why not to add copy here?

gensim/models/ldaseqmodel.py
```diff
@@ -1091,10 +1089,8 @@ def update_obs(self, sstats, totals):
                     w_counts = np.zeros(len(w_counts))
 
                 # TODO: apply lambda function
-                for t in range(0, T):
-                    mean_deriv = mean_deriv_mtx[t]
-                    mean_deriv = self.compute_mean_deriv(w, t, mean_deriv)
-                    mean_deriv_mtx[t] = mean_deriv
+                for t in range(T):
+                    mean_deriv_mtx[t] = self.compute_mean_deriv(w, t, mean_deriv_mtx[t])
```
menshikh-iv: are you sure? is `compute_mean_deriv` don't mutate arguments?

gensim/models/ldaseqmodel.py
horpto: `numpy` returns new array as a result of division operation. However operation '/=' does not create new arrays.

menshikh-iv: I mean why we don't copy `doc_topic[doc_number]` (to avoid mutation from external code)?

gensim/models/ldaseqmodel.py
```diff
@@ -172,6 +167,8 @@ def __init__(self, corpus=None, time_slice=None, id2word=None, alphas=0.01, num_
 
         # if a corpus and time_slice is provided, depending on the user choice of initializing LDA, we start DTM.
         if corpus is not None and time_slice is not None:
+            self.max_doc_len = max(len(line) for line in corpus)
```
horpto: move down because of this and original code is not supported case `corpus=None `.

gensim/models/ldaseqmodel.py
horpto: It's good point. I've check it and `compute_mean_deriv` changes `mean_deriv_mtx[t]` and returns it =).

menshikh-iv: > I've check it and compute_mean_deriv changes mean_deriv_mtx[t] and returns it =).

Oh jesus ...

gensim/models/ldaseqmodel.py
```diff
@@ -896,9 +895,9 @@ def sslm_counts_init(self, obs_variance, chain_variance, sstats):
         T = self.num_time_slices
 
         log_norm_counts = np.copy(sstats)
-        log_norm_counts = log_norm_counts / sum(log_norm_counts)
-        log_norm_counts = log_norm_counts + 1.0 / W
-        log_norm_counts = log_norm_counts / sum(log_norm_counts)
+        log_norm_counts /= sum(log_norm_counts)
```
horpto: it's a quite suspicious place. `sum(log_norm_counts)` means `log_norm_counts.sum(axis=0)`, not number and it's not obvious. Better to avoid it.

gensim/models/ldaseqmodel.py
```diff
@@ -460,8 +457,8 @@ def make_lda_seq_slice(self, lda, time):
             The stationary model updated to reflect the passed time slice.
 
         """
-        for k in range(0, self.num_topics):
-            lda.topics[:, k] = np.copy(self.topic_chains[k].e_log_prob[:, time])
+        for k in range(self.num_topics):
+            lda.topics[:, k] = self.topic_chains[k].e_log_prob[:, time]
```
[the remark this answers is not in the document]
horpto: Cause it's just copy column from one matrix to another. All values from `e_log_prob[:, time]` matrix will copy in memory of `lda.topics[:, k]` matrix. Change of `e_log_prob[:, time]` will not cause changing in `lda.topics[:k]`. I've checked it on numpy arrays as both of type is numpy array.

menshikh-iv:
Thanks for fixed @horpto :1st_place_medal:

PR #2361, opened 2019-01-29
NMF optimization & documentation

anotherbugmaster:
Massive performance improvements and better docs.

Takes less memory and 4-5 times faster now. Also metrics such as perplexity works as expected.

TODO:

- [x] Fix `use_r` functionality (invalid, dramatically slows down training, while not improving quality too much)
- [x] ~Make the `load` method work with the old implementation~ (invalid)
- [ ] Revamp wikipedia benchmark and add sklearn implementation
- [x] Revamp `corpus` parameter docs, make it easier to grasp
- [x] Document or deprecate `use_r` parameter
- [x] More info in the tutorial notebook
- [x] Fix tests
- [ ] Add useful parameters ranges
- [x] Document minimum_probability
- [x] Remove unused parameters from the tutorial notebook
- [ ] Remove unused parameters from wiki notebook
- [x] Remove r from eval metrics
- [x] Change eval metric to relative error
- [ ] Add RAM estimation


gensim/models/nmf.py
```diff
@@ -45,25 +79,26 @@ def __init__(
...
 
         Parameters
         ----------
         corpus : iterable of list of (int, float), optional
-            Training corpus. If not given, model is left untrained.
+            Training corpus. Contains list of counts of words for every document.
         num_topics : int, optional
             Number of topics to extract.
-        id2word: gensim.corpora.Dictionary, optional
+        id2word: gensim.corpora.dictionary.Dictionary, optional
```
menshikh-iv:
```
:class:`~gensim.corpora.dictionary.Dictionary`
```

gensim/models/nmf.py
```diff
@@ -45,25 +79,26 @@ def __init__(
...
-        passes: int, optioanl
+        passes: int, optional
             Number of full passes over the training corpus.
-        lambda_ : float, optional
+        \lambda_ : float, optional
             Residuals regularizer coefficient. Increasing it helps prevent ovefitting. Has no effect if `use_r` is set
             to False.
         kappa : float, optional
-            Optimizer step coefficient. Increaing it makes model train faster, but adds a risk that it won't converge.
+            Gradient descent step size.
```
menshikh-iv: missing useful ranges for parameters (here and everywhere)

anotherbugmaster: You're right, added that to TODO

gensim/models/nmf.py
```diff
@@ -45,25 +79,26 @@ def __init__(
...
-        id2word: gensim.corpora.Dictionary, optional
+        id2word: gensim.corpora.dictionary.Dictionary, optional
             Mapping from token id to token. If not set words get replaced with word ids.
         chunksize: int, optional
             Number of documents to be used in each training chunk.
-        passes: int, optioanl
+        passes: int, optional
             Number of full passes over the training corpus.
-        lambda_ : float, optional
+        \lambda_ : float, optional
```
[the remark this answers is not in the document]
anotherbugmaster: Because otherwise sphinx shows `lambda_` as `lambda`

gensim/models/nmf.py
```diff
@@ -1,4 +1,38 @@
...
+
+            r = v - Wh
+            bound and smooth r
+
+        update A and B
+
+        update W:
+            do gradient descent for W using A and B values
+
+The NMF should be used whenever one needs faster topic extraction.
```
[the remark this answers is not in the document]
anotherbugmaster: Ok!

gensim/models/nmf.py
anotherbugmaster: Done

gensim/models/nmf.py
```diff
@@ -458,26 +494,18 @@ def _setup(self, corpus):
...
         ----------
-        corpus : iterable of list(int, float)
-            Training corpus.
+        corpus : iterable of list of (int, float), optional
+            Training corpus. Contains list of counts of words for every document.
         chunks_as_numpy : bool, optional
-            Whether each chunk passed to the inference step should be a numpy.ndarray or not. Numpy can in some settings
-            turn the term IDs into floats, these will be converted back into integers in inference, which incurs a
-            performance hit. For distributed computing it may be desirable to keep the chunks as `numpy.ndarray`.
+            Deprecated.
```
horpto: not better to delete it all?  yeah, it's not backward-compatible, but it's a fresh module.

anotherbugmaster: I agree. @menshikh-iv, what do you think? I bet there are not so many people who are already actively using current NMF in their projects

anotherbugmaster: I think it would be better to break compatibility now, than drag it till the 4.0.0

menshikh-iv: not sure, let's ask @piskvorky: we want to remove some parameters from constructor, is it OK to break backward compatibility now (instead of stay it as "dead" parameters forever)?

piskvorky: Current NMF wasn't promoted yet, so backward compatibility is no issue. I'd prefer the release (after docs & bug fixes) to be clean & clear.

anotherbugmaster:
@menshikh-iv, @piskvorky, @horpto, I've just updated [the tutorial notebook](https://github.com/RaRe-Technologies/gensim/blob/7458fa566aea57edcdc249b7feb93578867b4c6c/docs/notebooks/nmf_tutorial.ipynb), please review and share your thoughts.

Cool stuff:

image

image

gensim/models/nmf.py
```diff
@@ -146,9 +225,8 @@ def get_topics(self, normalize=None):
     def __getitem__(self, bow, eps=None):
         return self.get_document_topics(bow, eps)
 
-    def show_topics(self, num_topics=10, num_words=10, log=False,
-                    formatted=True, normalize=None):
-        """Get a representation for selected topics.
+    def show_topics(self, num_topics=10, num_words=10, log=False, formatted=True, normalize=None):
+        """Get the topics sorted by sparsity.
```
piskvorky: What does `sparsity` mean here? The code looks like it's computing a column-wise average (?).

gensim/models/nmf.py
```diff
@@ -217,8 +295,8 @@ def show_topic(self, topicid, topn=10, normalize=None):
             The ID of the topic to be returned
         topn : int, optional
             Number of the most significant words that are associated with the topic.
-        normalize : bool, optional
-            Whether to normalize an output vector.
+        normalize: bool, optional
+            Whether to normalize the result. Allows for estimation of perplexity, coherence, e.t.c.
```
piskvorky: dtto

gensim/models/nmf.py
```diff
@@ -346,18 +424,19 @@ def log_perplexity(self, corpus):
...
 
         Parameters
         ----------
         word_id : int
             The word for which the topic distribution will be computed.
         minimum_probability : float, optional
-            Topics with an assigned probability below this threshold will be discarded.
-        normalize : bool, optional
-            Whether to normalize an output vector.
+            If `normalize` is True, than only topics with larger probabilities than this are included in a result.
```
piskvorky: Same as above.

gensim/models/nmf_pgd.pyx
```diff
@@ -8,26 +8,23 @@
 
 from libc.math cimport sqrt, fabs, copysign
 from cython.parallel import prange
+import numpy as np
+cimport numpy as np
 
 cdef double fmin(double x, double y) nogil:
```
horpto: don't think you need this. cython has native support  of min/max. check it.
and as another option you can add `inline`  in function definition.

gensim/models/nmf.py
```diff
@@ -422,33 +503,37 @@ def get_document_topics(self, bow, minimum_probability=None,
...
+            Training corpus. Contains list of counts of words for every document.
 
         """
-        self._h, self._r = None, None
-        first_doc_it = itertools.tee(corpus, 1)
-        first_doc = next(first_doc_it[0])
-        first_doc = matutils.corpus2csc([first_doc], len(self.id2word))
+        self._h = None
+
+        if isinstance(corpus, scipy.sparse.csc.csc_matrix):
```
piskvorky: Is this some special undocumented case? Deserves a comment.

gensim/models/nmf_pgd.pyx
```diff
@@ -47,8 +44,9 @@ def solve_h(double[:, ::1] h, double[:, :] Wt_v_minus_r, double[:, ::1] WtW, dou
 
     for sample_idx in prange(n_samples, nogil=True):
```
horpto: Don't you should add some options in setup.py to enable this "magic" ?

gensim/models/nmf.py
```diff
@@ -422,33 +503,37 @@ def get_document_topics(self, bow, minimum_probability=None,
...
+            first_doc = corpus.getcol(0)
+        else:
+            first_doc_it = itertools.tee(corpus, 1)
+            first_doc = next(first_doc_it[0])
```
piskvorky: I think this will still advance the original generator (lose the first document).

Isn't there a mismatch between expected corpus docs and actually seen corpus docs during iteration, after `_setup`?

gensim/models/nmf_pgd.pyx
menshikh-iv: No, this is a mandatory piece: you'll be surprised, but windows compiler for py27 is very old and wired, have issues even with this simple thing, for this reason, we have it here.

gensim/models/nmf.py
```diff
@@ -487,53 +562,50 @@ def update(self, corpus, chunks_as_numpy=False):
...
-            ):
+            for chunk in utils.grouper(corpus, self.chunksize, as_numpy=False):
                 self.random_state.shuffle(chunk)
-                v = matutils.corpus2csc(chunk, len(self.id2word)).tocsr()
-                self._h, self._r = self._solveproj(
-                    v, self._W, r=self._r, h=self._h, v_max=self.v_max
-                )
-                h, r = self._h, self._r
+
+                if isinstance(corpus, scipy.sparse.csc.csc_matrix):
```
piskvorky: Hmm, how does this work in combination with `grouper` above? The grouper groups per rows, but CSC doesn't allow row-wise slicing?

gensim/models/nmf.py
```diff
@@ -612,45 +677,30 @@ def _solveproj(self, v, W, h=None, r=None, v_max=None):
...
-            r = scipy.sparse.csr_matrix(rshape)
+        Wt = W.T
+        WtW = Wt.dot(W)
 
-        WtW = W.T.dot(W)
+        h_error = None
 
-        _h_r_error = None
+        for iter_number in range(self._h_max_iter):
+            logger.debug("h_error: %s" % h_error)
```
piskvorky:
```suggestion
            logger.debug("h_error: %s", h_error)
```

gensim/models/nmf_pgd.pyx
horpto: =(, so just add inline

menshikh-iv: ping @anotherbugmaster

gensim/models/nmf_pgd.pyx
menshikh-iv: not mandatory, current "magic" works even without a compiler flags :) 
I guess it's a good idea to enable compiler options later for all `.pyx`, but not now.

gensim/models/nmf.py
anotherbugmaster: Corpus can be csc matrix now, I'll add it to the docstring

gensim/models/nmf_pgd.pyx
horpto: Of course, it will work. but will it work parallel ?

gensim/models/nmf.py
anotherbugmaster: Yes, it's a bug, I'll fix it

gensim/models/nmf.py
anotherbugmaster: No, that's something that I checked, `itertools.tee` copies generator and leaves the original one intact.

gensim/models/nmf.py
```diff
@@ -1,21 +1,110 @@
...
+
+    for v in batches:
+        infer h:
+            do coordinate gradient descent step to find h that minimizes (v - Wh) l2 norm
+            bound h so that it is non-negative
+
+        update A and B
+
+        update W:
+            do gradient descent for W using A and B values
```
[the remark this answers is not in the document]
anotherbugmaster: I'll clarify that.

gensim/models/nmf.py
```diff
@@ -31,79 +120,69 @@ def __init__(
...
 
         Parameters
         ----------
         corpus : iterable of list of (int, float), optional
-            Training corpus. If not given, model is left untrained.
+            Training corpus. Contains list of counts of words for every document.
         num_topics : int, optional
             Number of topics to extract.
-        id2word: gensim.corpora.Dictionary, optional
+        id2word: :class:`~gensim.corpora.dictionary.Dictionary`, optional
```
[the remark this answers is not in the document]
anotherbugmaster: Right, I'll fix this

gensim/models/nmf.py
```diff
@@ -31,79 +120,69 @@ def __init__(
...
             Number of full passes over the training corpus.
-        lambda_ : float, optional
-            Residuals regularizer coefficient. Increasing it helps prevent ovefitting. Has no effect if `use_r` is set
-            to False.
         kappa : float, optional
-            Optimizer step coefficient. Increaing it makes model train faster, but adds a risk that it won't converge.
+            Gradient descent step size.
+            Larger value makes the model train faster, but could lead to non-convergence if set too large.
+        minimum_probability:
+            If `normalize` is True, than only topics with larger probabilities than this are included in a result.
```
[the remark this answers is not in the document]
anotherbugmaster: Result is a return value of `.get_topics`, `.get_term_topics`, `.get_topic_terms`, e.t.c., I couldn't find a better term for that.

gensim/models/nmf.py
```diff
@@ -126,16 +205,16 @@ def get_topics(self, normalize=None):
 
         Parameters
         ----------
-        normalize : bool, optional
-            Whether to normalize an output vector.
+        normalize: bool, optional
+            Whether to normalize the result. Allows for estimation of perplexity, coherence, e.t.c.
```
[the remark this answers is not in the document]
anotherbugmaster: Yes, the default should be changed.

anotherbugmaster:
>     1. Does `trainset` / `testset` need to be a `np.array`, instead of `list`? Storing objects (`dict`s) as numpy array is inefficient and looks strange. If really necessary, deserves a comment "why".

Yes, this is not efficient, but I wanted to fix the random seed during permutation while not messing with the global random seed.

>     2. Can we introduce the notebook cells with a bit of context / motivation? For example, under the section **Coherence**, mention why this section is here in the notebook + link to an explanation what coherence is?

Sure, I'll explain it in more details.

>     3. The density definition as `def density(matrix):    return (matrix > 0).mean()` seems non-standard and deserves a comment (why density as "mean of above-zero elements"?)

The (matrix > 0) is a bool matrix, mean function sums up `True`'s for every cell in which condition stands, than divides it by number of all elements in matrix. Thus, it's the ordinary density, nothing fancy. :)

gensim/models/nmf.py
anotherbugmaster: image

Fixed, faces look much better now. :)

gensim/models/nmf_pgd.pyx
anotherbugmaster: Well, works parallel on my machine. :)

anotherbugmaster: I agree, we'll deal with this later

gensim/models/nmf.py
```diff
@@ -1,21 +1,110 @@
...
+.. sourcecode:: pycon
+
+    >>> nmf.update(other_corpus)
+    >>> vector = nmf[unseen_doc]
+
+A lot of parameters can be tuned to optimize training for your specific case
+
+.. sourcecode:: pycon
+
+    >>> nmf = Nmf(common_corpus, num_topics=50, kappa=0.1, eval_every=5)  # descrease training step size
```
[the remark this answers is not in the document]
anotherbugmaster: Fixed

gensim/models/nmf.py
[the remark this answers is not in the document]
anotherbugmaster: Fixed.

gensim/models/nmf.py
```diff
@@ -1,21 +1,110 @@
...
+
+    Initialize W, A and B matrices
+
+    Input corpus
+    Split corpus to batches
+
+    for v in batches:
+        infer h:
+            do coordinate gradient descent step to find h that minimizes (v - Wh) l2 norm
+            bound h so that it is non-negative
```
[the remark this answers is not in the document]
anotherbugmaster: Fixed.

gensim/models/nmf.py
```diff
@@ -31,79 +120,69 @@ def __init__(
...
         random_state=None,
     ):
-        """
+        r"""
 
         Parameters
         ----------
         corpus : iterable of list of (int, float), optional
-            Training corpus. If not given, model is left untrained.
+            Training corpus. Contains list of counts of words for every document.
```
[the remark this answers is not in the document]
anotherbugmaster: Fixed.

gensim/models/nmf.py
```diff
@@ -1,21 +1,110 @@
...
+
+    Input corpus
+    Split corpus to batches
+
+    for v in batches:
+        infer h:
+            do coordinate gradient descent step to find h that minimizes (v - Wh) l2 norm
+            bound h so that it is non-negative
+
+        update A and B
```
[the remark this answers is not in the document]
anotherbugmaster: Fixed

gensim/models/nmf.py
anotherbugmaster: Fixed.

gensim/models/nmf.py
anotherbugmaster: Fixed.

gensim/models/nmf.py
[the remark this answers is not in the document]
anotherbugmaster: Actually, I changed the type in the docstring, because this parameter overrides global `self.normalize` setting of the model if set.

gensim/models/nmf.py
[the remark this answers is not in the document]
anotherbugmaster: Fixed.

gensim/models/nmf.py
```diff
@@ -31,79 +120,69 @@ def __init__(
...
         num_topics : int, optional
             Number of topics to extract.
-        id2word: gensim.corpora.Dictionary, optional
+        id2word: :class:`~gensim.corpora.dictionary.Dictionary`, optional
             Mapping from token id to token. If not set words get replaced with word ids.
         chunksize: int, optional
             Number of documents to be used in each training chunk.
-        passes: int, optioanl
+        passes: int, optional
             Number of full passes over the training corpus.
```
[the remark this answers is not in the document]
anotherbugmaster: Fixed.

gensim/models/nmf.py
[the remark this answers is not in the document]
anotherbugmaster: Fixed.

gensim/models/nmf.py
```diff
@@ -31,79 +120,69 @@ def __init__(
...
-        lambda_ : float, optional
-            Residuals regularizer coefficient. Increasing it helps prevent ovefitting. Has no effect if `use_r` is set
-            to False.
         kappa : float, optional
-            Optimizer step coefficient. Increaing it makes model train faster, but adds a risk that it won't converge.
+            Gradient descent step size.
+            Larger value makes the model train faster, but could lead to non-convergence if set too large.
+        minimum_probability:
+            If `normalize` is True, than only topics with larger probabilities than this are included in a result.
+            If `normalize` is False, than only topics with larger factors than this are included in a result.
```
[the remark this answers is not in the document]
anotherbugmaster: Fixed.

gensim/models/nmf.py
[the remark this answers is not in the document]
anotherbugmaster: Fixed.

gensim/models/nmf.py
anotherbugmaster: Right, fixed that.

menshikh-iv:
@anotherbugmaster awesome :fire:

Release can't more wait, so, I merge PR right now.
Please don't forget to fix unchecked notes from first post.

gensim/models/nmf.py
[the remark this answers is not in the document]
piskvorky: I still see it on other places (not fully fixed).

gensim/models/nmf.py
```diff
@@ -177,7 +264,7 @@ def show_topics(self, num_topics=10, num_words=10, log=False,
         if normalize is None:
             normalize = self.normalize
 
-        sparsity = self._W.getnnz(axis=0)
+        sparsity = (self._W == 0).mean(axis=0)
```
piskvorky: Mean of zeros? What does this do?

gensim/models/nmf.py
piskvorky:
```python
corpus = (i for i in range(10))
first_doc_it = itertools.tee(corpus, 1)
first_doc = next(first_doc_it[0])
print("first doc", first_doc)
for doc_no, doc in enumerate(corpus):
    print(f"doc #{doc_no}: doc")

outputs:

first doc 0
doc #0: 1
doc #1: 2
doc #2: 3
doc #3: 4
doc #4: 5
doc #5: 6
doc #6: 7
doc #7: 8
doc #8: 9
```

gensim/models/nmf.py
piskvorky: I still see no comment about this code path. **Why** does this method accept two different input types? Is it some optimization?

gensim/models/nmf.py
piskvorky: OK, got it. So `sparsity` is an array of densities of the columns.

BTW this `sparse_matrix == 0` prints a warning in my scipy 0.19.0:

```
/Volumes/work/workspace/vew/pii3.6/lib/python3.6/site-packages/scipy/sparse/compressed.py:214: SparseEfficiencyWarning: Comparing a sparse matrix with 0 using == is inefficient, try using != instead.
  ", try using != instead.", SparseEfficiencyWarning)
```

gensim/models/nmf.py
piskvorky: Does NMF accept generators? Or does it require a "restartable" sequence in `corpus`?

piskvorky: Can you add those corner cases into unit tests, too? To prevent regressions. (generator input, empty docs, corpus shape not divisible by chunksize, etc)

gensim/models/nmf.py
```diff
@@ -458,90 +561,88 @@ def _setup(self, corpus):
...
-            for chunk in utils.grouper(
-                corpus, self.chunksize, as_numpy=chunks_as_numpy
-            ):
-                self.random_state.shuffle(chunk)
-                v = matutils.corpus2csc(chunk, len(self.id2word)).tocsr()
-                self._h, self._r = self._solveproj(
-                    v, self._W, r=self._r, h=self._h, v_max=self.v_max
+            if isinstance(corpus, scipy.sparse.csc.csc_matrix):
+                grouper = (
+                    corpus[:, col_idx:col_idx + self.chunksize]
```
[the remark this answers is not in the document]
anotherbugmaster: The corpus has shape (n_tokens, n_documents) in this case, so the `grouper` splits the input corpus in chunks by column (documents dimension) and then shuffles columns of every chunk (shuffles documents order in the chunk).

gensim/models/nmf.py
anotherbugmaster: NMF loses the first document of a non-restartable generator for now, though non-restartable sequence should work fine with `passes=1`.

I'll fix that and add unit tests.

gensim/models/nmf.py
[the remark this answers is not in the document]
anotherbugmaster: Yes, it's unusual, but that's the format that's used in the original paper. I'll specify the shape of the corpus in the module docstring.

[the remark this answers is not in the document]
piskvorky: This may be due to some changes in scipy, but your code results in

```
IndexError: index out of bounds: 0 <= 5 <= 10, 0 <= 100500 <= 10, 5 <= 100500
```

in scipy 0.19.0.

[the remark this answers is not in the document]
anotherbugmaster: Oh, I see. Seems like they added the support for that in the following versions, I'll do something like `corpus[:, col_idx:min(col_idx + self.chunksize, chunk.shape[1])]`

gensim/models/nmf.py
anotherbugmaster: Oops, I added that to the `__init__` docstring, but forgot to mention csc in the `update`. I'll fix it.

gensim/models/nmf.py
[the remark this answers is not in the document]
piskvorky: Thanks. Plus add a comment, so someone doesn't accidentally remove it again in the future. 

We may simplify it again once the older scipy versions become irrelevant in the future.

gensim/models/nmf.py
piskvorky: Hmm, so if I understand correctly, previously we were interpreting documents as features and vice versa, while shuffling the features?

How did that even work? Something must be wrong with the eval if that worked (enough) to pass all human & automated checks.

anotherbugmaster: Yes, but only in the case of csc in Olivietti faces dataset, other examples use iterable of word counts as input.

I think what happen is one bug interfered with another an it worked. :neutral_face: 

Specifically, in the notebook I transposed input faces matrix twice, thus passing dataset with shape `(n_faces, n_pixels)` to the model, where it was split to the lists of rows of shape `(chunksize, n_pixels)`, and [here](https://github.com/RaRe-Technologies/gensim/blob/7458fa566aea57edcdc249b7feb93578867b4c6c/gensim/models/nmf.py#L570) I took the first row in a batch and transposed it, thus getting `v` with shape `(n_pixels, 1)`, which is valid dimensions.

The effect of this was that NMF were training only on the first face of every batch, that's why the quality of faces were worse (only 1/chunksize of examples were actually used).

piskvorky: Beautiful! Such tricky double-bugs can be hard to find :)

Can we do anything to prevent similar bugs in the future? More unit tests / asserts?

anotherbugmaster: Sure! We'll just need to check shapes of internal matrices after an update.

piskvorky:
@anotherbugmaster let me know about the performance (RAM, speed, quality) in our target use-case of sparse input (text, not dense faces), especially in comparison to sklearn or other tools. Then we can publish this. Super cool feature!

anotherbugmaster:
@piskvorky Ok! I'll push the notebook with metrics and additional annotations later this weekend.

anotherbugmaster:
@piskvorky Ok! I'll push the notebook with metrics and additional annotations later this weekend.

anotherbugmaster:
Here are the current wikipedia results:

image

piskvorky:
So, 0.14 GB vs 10 GB RAM compared to sklearn… nice!

But the L2 norm is worrying, that's a huge difference. Is that expected?

anotherbugmaster:
> So, 0.17 GB vs 16 GB RAM compared to sklearn… nice! And 2x as fast racehorse
> 
> But the L2 norm is worrying, that's a big difference. Is that expected? Especially given the lower perplexity (?).

Checked the metrics functions again and found a bug in evaluation of l2 (I normalized the input corpus for sklearn, but not for NMF and LDA, hence the invalid metrics). I've fixed it, re-running wikipedia notebook now. I expect changes only in l2 norm of all models and in perplexity of sklearn NMF.

Here are updated metrics for the tutorial:

image

> Can you post the resulting topics too? For visual, manual comparison.

Sure, but I'm not sure how to present them. I think I'll take top-5 topics from each model to fit them in one scroll.

> Reordering the table to show model first, then train time, then RAM, then the quality metrics should make it easier to read and interpret. Thanks!

Ok!

anotherbugmaster:
@piskvorky, here are the updated metrics on wikipedia:

image

piskvorky:
Awesome. Please update the notebook (and reorder the table columns) and I'll "officially" announce our new model.

Great work @anotherbugmaster , this came together very nicely in the end.

piskvorky:
@anotherbugmaster since this PR is already merged, can you push these updated metrics / tables / notebook in a new PR?

anotherbugmaster:
@piskvorky Sure, [here it is](https://github.com/RaRe-Technologies/gensim/pull/2371)

gensim/models/nmf.py
anotherbugmaster: I just realized that `self._W` is actually dense, thus there will be no warning.

piskvorky: Careful about temporary copies and RAM then (each sub-expression like `1 - (D + 1))` can create an independent array copy).

piskvorky: Yes, that seems more efficient.

But remember to include a clear comment, otherwise we'll be scratching our heads in a few months, wondering what this construct is for.

gensim/models/nmf.py
```diff
@@ -1,21 +1,114 @@
-"""Online Non-Negative Matrix Factorization."""
+"""`Online Non-Negative Matrix Factorization. <https://arxiv.org/abs/1604.02634>`
```
piskvorky: This formatting is broken: RST links must end in `_`.

I'd suggest instead:
```python
Online Non-Negative Matrix Factorization. This uses the efficient sparse incremental algorithm of Renbo Zhao, Vincent Y. F. Tan et al. `[PDF] <https://arxiv.org/abs/1604.02634>`_.
```

PR #2363, opened 2019-01-30
Use complex dtype for distributed cases

horpto:
+ fix doc for grouper
+ add argument in chunksize

Now word ids will not converse to float and it will be stored and pickled as an integer in memory and so there is no backward conversion issues + speed + size of pickled data.

benchmark:

``` python
import numpy as np
from pickle import dumps
bow = [(np.random.randint(1000000), np.random.rand()*10e6) for i in range(10000)]

def test(b):
    print('TRY:', len(b))

    print('pure python')

    %timeit dumps(b)

    print('np array without args')

    %timeit dumps(np.array(b))

    print('np array with simple dtype')
    %timeit dumps(np.array(b, dtype=np.float32))

    complex_dt = np.dtype([('1', np.int32), ('2', np.float32)])
    print('np array with complex dtype')
    %timeit dumps(np.array(b, dtype=complex_dt))

    print('-'*50)

test(bow[:50])
test(bow[:150])
test(bow[:500])
test(bow[:1000])
test(bow[:2000])
test(bow[:5000])
test(bow)
```

results (anyone else should recheck it):

```
TRY: 50
pure python
30.2 µs ± 407 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
np array without args
138 µs ± 4.54 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array with simple dtype
141 µs ± 4.92 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array with complex dtype
148 µs ± 3.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
--------------------------------------------------
TRY: 150
pure python
89.1 µs ± 520 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
np array without args
260 µs ± 6.07 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array with simple dtype
260 µs ± 5.11 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array with complex dtype
202 µs ± 4.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
--------------------------------------------------
TRY: 500
pure python
336 µs ± 5.43 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array without args
699 µs ± 11.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array with simple dtype
684 µs ± 7.86 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array with complex dtype
392 µs ± 5.77 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
--------------------------------------------------
TRY: 1000
pure python
660 µs ± 8.38 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array without args
1.33 ms ± 14.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array with simple dtype
1.32 ms ± 4.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
np array with complex dtype
689 µs ± 5.77 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
--------------------------------------------------
TRY: 2000
pure python
1.7 ms ± 76.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np array without args
2.65 ms ± 61 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np array with simple dtype
2.62 ms ± 20.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np array with complex dtype
1.29 ms ± 11.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
--------------------------------------------------
TRY: 5000
pure python
4.11 ms ± 52.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np array without args
6.58 ms ± 69.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np array with simple dtype
6.56 ms ± 43.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np array with complex dtype
2.92 ms ± 23.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
--------------------------------------------------
TRY: 10000
pure python
12 ms ± 95.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np array without args
12.7 ms ± 35.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np array with simple dtype
12.6 ms ± 34.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np array with complex dtype
5.43 ms ± 43.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
--------------------------------------------------
```

### DIFF
--- a/gensim/models/ldamulticore.py
+++ b/gensim/models/ldamulticore.py
@@ -276,13 +276,16 @@ class LdaMulticore(LdaModel):
                 if eval_every > 0 and (force or (self.num_updates / updateafter) % eval_every == 0):
                     self.log_perplexity(chunk, total_docs=lencorpus)
 
+        # HACK np.int32 should be enough for word ids.
+        chunk_dtype = np.dtype([('1', np.int32), ('2', self.dtype)]) if chunks_as_numpy else self.dtype
+
         logger.info("training LDA model using %i processes", self.workers)
         pool = Pool(self.workers, worker_e_step, (job_queue, result_queue,))
         for pass_ in range(self.passes):
             queue_size, reallen = [0], 0
             other = LdaState(self.eta, self.state.sstats.shape)
 
-            chunk_stream = utils.grouper(corpus, self.chunksize, as_numpy=chunks_as_numpy)
+            chunk_stream = utils.grouper(corpus, self.chunksize, as_numpy=chunks_as_numpy, dtype=chunk_dtype)
             for chunk_no, chunk in enumerate(chunk_stream):
                 reallen += len(chunk)  # keep track of how many documents we've processed so far
 
--- a/gensim/models/ldamulticore.py
+++ b/gensim/models/ldamulticore.py
@@ -209,9 +209,8 @@ class LdaMulticore(LdaModel):
             Stream of document vectors or sparse matrix of shape (`num_terms`, `num_documents`) used to update the
             model.
         chunks_as_numpy : bool
-            Whether each chunk passed to the inference step should be a np.ndarray or not. Numpy can in some settings
-            turn the term IDs into floats, these will be converted back into integers in inference, which incurs a
-            performance hit. For distributed computing it may be desirable to keep the chunks as `numpy.ndarray`.
+            Whether each chunk passed to the inference step should be a np.ndarray or not.
+            For distributed computing it may be desirable to keep the chunks as `numpy.ndarray`.
 
         """
         try:
--- a/gensim/utils.py
+++ b/gensim/utils.py
@@ -1146,6 +1146,8 @@ def chunkize_serial(iterable, chunksize, as_numpy=False, dtype=np.float32):
         Split iterable into chunks of this size.
     as_numpy : bool, optional
         Yield chunks as `np.ndarray` instead of lists.
+    dtype : numpy.dtype, optional
+        Data type of `np.ndarray`.
 
     Yields
     ------
@@ -1184,7 +1186,7 @@ class InputQueue(multiprocessing.Process):
     so that workers that use the queue are not starved for input chunks.
 
     """
-    def __init__(self, q, corpus, chunksize, maxsize, as_numpy):
+    def __init__(self, q, corpus, chunksize, maxsize, as_numpy, dtype=np.float32):
         """
         Parameters
         ----------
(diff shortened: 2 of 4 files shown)


gensim/models/ldamulticore.py
```diff
@@ -276,13 +275,16 @@ def process_result_queue(force=False):
                 if eval_every > 0 and (force or (self.num_updates / updateafter) % eval_every == 0):
                     self.log_perplexity(chunk, total_docs=lencorpus)
 
+        # HACK np.int32 should be enough for word ids.
+        chunk_dtype = np.dtype([('1', np.int32), ('2', self.dtype)]) if chunks_as_numpy else self.dtype
```
piskvorky: What does this construction do? Why the 2-element list, with two different types inside (int32 and self.dtype)?

Too "clever" to be dropped without a comment.

gensim/utils.py
```diff
@@ -1146,6 +1146,8 @@ def chunkize_serial(iterable, chunksize, as_numpy=False, dtype=np.float32):
         Split iterable into chunks of this size.
     as_numpy : bool, optional
         Yield chunks as `np.ndarray` instead of lists.
+    dtype : numpy.dtype, optional
+        Data type of `np.ndarray`.
```
piskvorky: Which `np.ndarray` array?

gensim/utils.py
```diff
@@ -1212,8 +1216,8 @@ def run(self):
             if self.as_numpy:
                 # HACK XXX convert documents to numpy arrays, to save memory.
                 # This also gives a scipy warning at runtime:
-                # "UserWarning: indices array has non-integer dtype (float64)"
-                wrapped_chunk = [[np.asarray(doc) for doc in chunk]]
+                # "UserWarning: indices array has non-integer dtype (float32)"
+                wrapped_chunk = [[np.asarray(doc, dtype=self.dtype) for doc in chunk]]
```
piskvorky: This changes the existing logic from float64 to float32, right?

It's probably OK, but we have to make sure to mention it in the release notes, because this could lead to nasty hidden incompatibilities and bugs for existing users, who rely on double precision.

PR #2366, opened 2019-01-30
Don't override `Doc2Vec.infer_vector` defaults in `Doc2VecKeyedVectors.similarity_unseen_documents`

gojomo:
`similarity_unseen_documents()` is explicitly specifying the (old, crude) `infer_vector()` defaults, rather than letting `infer_vector()` apply its own default behavior - which nowadays is to re-use the alpha/epochs parameters that the model was configured to use.


menshikh-iv:
Thank you @gojomo, good idea!

PR #2367, opened 2019-01-31
Fix `stdint_wrapper.h` missing in sdist

cgohlke:


menshikh-iv:
Thank you @cgohlke, congratz with the first contribution :1st_place_medal:!

Just interested, what's OS & Python version you used because you downloaded sdist (instead of any available wheel)?

menshikh-iv:
wow, awesome @cgohlke, thank you very much for your work: this project helped me a lot when I was at university and used windows!

PR #2370, opened 2019-02-03
Undo the hash2index optimization

### LINKED ISSUE #2329: Roll back optimization that trims empty buckets
Our implementation diverges from Facebook's because we trim buckets that have no ngrams assigned to them, saving memory and training time. We should roll back this optimization because it introduces divergent behavior and makes the code more complex.

See https://github.com/RaRe-Technologies/gensim/pull/2313#issue-241664476 and https://github.com/RaRe-Technologies/gensim/pull/2313#issuecomment-453492981 for more details

mpenkov:
This optimization reduced the number of ngram buckets to include only ngrams that we have seen during training.

This seemed like a good idea at the time, because it saved CPU cycles and RAM, but turned out to be a bad idea, because it introduced divergent behavior when compared to the reference implementation. For example:

We were unable to calculate vectors for terms that were _completely_ out of the vocab (so the term and all its ngrams were unseen). This is bad because the original FB implementation **always** returns a vector. It may _seem_ useless because it's initialized to a random vector, but that's not entirely true, because that vector is random at _initialization_ time. When we're querying the ngram's vector, the vector is _deterministic_, so it is useful.

Another problem is that it complicated the implementation. We now needed an additional layer of indirection that mapped hashes to bucket indices. Without this optimization, this mapping is essentially the identifiy function: the hash N always maps to the Nth bucket.

This pull request removes the optimization, resolving the problems that it introduced.

Fixes #2329

### DIFF
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -696,7 +696,9 @@ class FastText(BaseWordEmbeddingsModel):
             raise RuntimeError(
                 "You cannot do an online vocabulary-update of a model which has no prior vocabulary. "
                 "First build the vocabulary of your model with a corpus "
-                "before doing an online update.")
+                "by calling the :meth:`gensim.models.fasttext.FastText.build_vocab` method "
+                "before doing an online update."
+            )
         else:
             self.vocabulary.old_vocab_len = len(self.wv.vocab)
 
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -696,7 +696,7 @@ class FastText(BaseWordEmbeddingsModel):
             raise RuntimeError(
                 "You cannot do an online vocabulary-update of a model which has no prior vocabulary. "
                 "First build the vocabulary of your model with a corpus "
-                "by calling the :meth:`gensim.models.fasttext.FastText.build_vocab` method "
+                "by calling the gensim.models.fasttext.FastText.build_vocab method "
                 "before doing an online update."
             )
         else:
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -690,18 +690,22 @@ class FastText(BaseWordEmbeddingsModel):
             >>> model.train(sentences_2, total_examples=model.corpus_count, epochs=model.epochs)
 
         """
-        if update:
-            if not len(self.wv.vocab):
-                raise RuntimeError(
-                    "You cannot do an online vocabulary-update of a model which has no prior vocabulary. "
-                    "First build the vocabulary of your model with a corpus "
-                    "before doing an online update.")
+        if not update:
+            self.wv.init_ngrams_weights(self.trainables.seed)
+        elif not len(self.wv.vocab):
+            raise RuntimeError(
+                "You cannot do an online vocabulary-update of a model which has no prior vocabulary. "
+                "First build the vocabulary of your model with a corpus "
+                "before doing an online update.")
+        else:
             self.vocabulary.old_vocab_len = len(self.wv.vocab)
-            self.trainables.old_hash2index_len = len(self.wv.hash2index)
 
-        return super(FastText, self).build_vocab(
+        retval = super(FastText, self).build_vocab(
             sentences=sentences, corpus_file=corpus_file, update=update, progress_per=progress_per,
             keep_raw_vocab=keep_raw_vocab, trim_rule=trim_rule, **kwargs)
+
+        if update:
+            self.wv.update_ngrams_weights(self.trainables.seed, self.vocabulary.old_vocab_len)
 
     def _set_train_params(self, **kwargs):
         #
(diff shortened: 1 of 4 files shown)


gensim/models/fasttext.py
```diff
@@ -690,31 +690,35 @@ def build_vocab(self, sentences=None, corpus_file=None, update=False, progress_p
...
-                raise RuntimeError(
-                    "You cannot do an online vocabulary-update of a model which has no prior vocabulary. "
-                    "First build the vocabulary of your model with a corpus "
-                    "before doing an online update.")
+        if not update:
+            self.wv.init_ngrams_weights(self.trainables.seed)
+        elif not len(self.wv.vocab):
+            raise RuntimeError(
+                "You cannot do an online vocabulary-update of a model which has no prior vocabulary. "
+                "First build the vocabulary of your model with a corpus "
```
piskvorky: Let's mention the necessary method, too (`build_vocab` or whatever).

gensim/models/keyedvectors.py
```diff
@@ -2349,3 +2352,127 @@ def _l2_norm(m, replace=False):
...
+def _rollback_optimization(kv):
+    """Undo the optimization that pruned buckets.
+
+    This unfortunate optimization saves memory and CPU cycles, but breaks
+    compatibility with Facebook's model by introducing divergent behavior
+    for OOV words.
+
+    """
+    logger.warning(
+        "This older model was trained with a premature optimization. "
```
piskvorky:
```suggestion
        "This saved FastText model was trained with an optimization we no longer support. "
        "The current Gensim version reverses automatically this optimization during loading. "
        "Save the loaded model to a new file and reload to suppress this message."
```

gensim/models/keyedvectors.py
```diff
@@ -2349,3 +2352,127 @@ def _l2_norm(m, replace=False):
...
+def _unpack_copy(m, num_rows, hash2index, seed=1):
+    """Same as _unpack, but makes a copy of the matrix.
+
+    Simpler implementation, but uses more RAM.
+
+    """
+    rows, columns = m.shape
+    if rows == num_rows:
+        #
+        # Nothing to do.
```
piskvorky: What's the advantage of the extra `#` lines around a comment?

mpenkov: It’s a personal preference.

It forces the writer/editor to think twice about putting/leaving the comment there, and nudges towards writing prose explanations instead of one-liners.

It also improves readability (the code/comment boundary is more obvious).

gensim/models/fasttext.py
```diff
@@ -690,31 +690,37 @@ def build_vocab(self, sentences=None, corpus_file=None, update=False, progress_p
...
+                "by calling the :meth:`gensim.models.fasttext.FastText.build_vocab` method "
```
piskvorky: No need for `:meth:`, this exception will appear in logs, not be rendered by Sphinx.

gensim/models/keyedvectors.py
```diff
@@ -2349,3 +2352,129 @@ def _l2_norm(m, replace=False):
...
+    compatibility with Facebook's model by introducing divergent behavior
+    for OOV words.
+
+    """
+    logger.warning(
+        "This saved FastText model was trained with an optimization we no longer support. "
+        "The current Gensim version automatically reverses this optimization during loading. "
+        "Save the loaded model to a new file and reload to suppress this message."
+        "The current Gensim version reverses this optimization during loading. "
+        "Save the model to a new file and reload to suppress this message."
```
piskvorky: Last two lines not needed now (superseded by above).

mpenkov:
@piskvorky When do you think would be the best time to release this?

We can sneak it into 3.7.2 (the next bugfix release) or wait until the next minor release (3.8.0). Technically, this isn't really a bugfix, it's a feature improvement (more precisely, the removal of a bad feature), so it can wait.

WDYT?

piskvorky:
I'd consider it a bug fix, and release as soon as possible. But #2361 NMF still blocking.

gojomo:
Correcting a deviation from the behavior of Facebook's reference FastText implementation definitely seems like a bug fix to me! 

And it would be good to include a release-note advisory about the changed behavior, now that gensim's FastText (post-fix) will always return a vector for OOV words (and never the "all ngrams for word X absent from model" error it was throwing previously).

mpenkov:
Already done in the changelog on develop :)

gojomo:
@mpenkov I see a note about `__contains__`, but it's a little confusing. Isn't `__contains__()` and the Python built-in `in` operator supposed to give identical results? That's actually specified by the [Python 3.7 documentation](https://docs.python.org/3/reference/datamodel.html#object.__contains__), but the current text of the changelog suggests gensim will do the opposite: always return `True` for `__contains__()`, but sometimes return `False` for `in`. 

I don't see a note that all `[]` lookups will now always return a vector, and never return the `KeyError` they sometimes did in gensim through 3.7.1.

mpenkov:
OK, please have a look at this [commit](https://github.com/RaRe-Technologies/gensim/commit/e69e112942179a24a3bf3589b184d7187d1e089b) and let me know if it is sufficient.

gojomo:
Put an extensive note on the [commit](https://github.com/RaRe-Technologies/gensim/commit/e69e112942179a24a3bf3589b184d7187d1e089b#commitcomment-32668390).

PR #2371, opened 2019-02-04
NMF metrics and wikipedia

anotherbugmaster:
Add clean up and fixes on top of #2361:

- [x] Fix l2 of all models
- [x] Fix perplexity of the sklearn NMF
- [x] Change loss function in the logs of the model
- [x] [Print topics in the wikipedia notebook](https://github.com/RaRe-Technologies/gensim/pull/2361#issuecomment-459901426)
- [x] [Reorder the benchmark tables](https://github.com/RaRe-Technologies/gensim/pull/2361#issuecomment-459901426)
- [x] [Fix an out-of-bounds error in grouper](https://github.com/RaRe-Technologies/gensim/pull/2361#pullrequestreview-198452850)
- [x] [Specify the shape of the corpus in the module docstring](https://github.com/RaRe-Technologies/gensim/pull/2361#pullrequestreview-198452711)
- [x] [Fix the generator issue](https://github.com/RaRe-Technologies/gensim/pull/2361#discussion_r252396076)
- [x] Add more unittests
- [x] More explanations in wikipedia
- [x] Fix flake8
- [x] Add comment about sparsity method
- [x] Mention pass parameter behavior for generators/iterators
- [x] Fix smart_open import
- [x] Fix tests


gensim/models/nmf.py
```diff
@@ -3,9 +3,11 @@
 Implements online non-negative matrix factorization algorithm, which allows for fast latent topic inference.
 This NMF implementation updates in a streaming fashion and works best with sparse corpora.
 
+The shape of the input corpus is ``(n_words, n_documents)``, which is the dimension order used in the original paper.
```
piskvorky: Whoa, that doesn't sound right. The input corpus must be a stream of documents, like the rest of Gensim.

I thought this only affected the (barely documented) CSC code path?

gensim/models/nmf.py
```diff
@@ -264,7 +266,12 @@ def show_topics(self, num_topics=10, num_words=10, log=False, formatted=True, no
         if normalize is None:
             normalize = self.normalize
 
-        sparsity = (self._W == 0).mean(axis=0)
+        sparsity = np.zeros(self._W.shape[1])
```
piskvorky: Needs clear code comments, as per my comment.

gensim/models/nmf.py
```diff
@@ -274,7 +281,7 @@ def show_topics(self, num_topics=10, num_words=10, log=False, formatted=True, no
 
             sorted_topics = list(matutils.argsort(sparsity))
             chosen_topics = (
-                sorted_topics[: num_topics // 2] + sorted_topics[-num_topics // 2:]
+                    sorted_topics[: num_topics // 2] + sorted_topics[-num_topics // 2:]
```
piskvorky: Wrong indent.

gensim/models/nmf.py
```diff
@@ -548,10 +555,15 @@ def _setup(self, corpus):
 
         if isinstance(corpus, scipy.sparse.csc.csc_matrix):
             first_doc = corpus.getcol(0)
-        else:
-            first_doc_it = itertools.tee(corpus, 1)
-            first_doc = next(first_doc_it[0])
+        elif hasattr(corpus, "__iter__"):
```
piskvorky: This is not the right way to test for an iterable in Python. Also, still destructive in case the corpus stream is a non-repeatable generator.

Check out `utils.is_corpus` instead.

gensim/models/nmf.py
anotherbugmaster: No, you got it right, it accepts streams too, I'll make it clearer.

gensim/models/nmf.py
anotherbugmaster: I test for an iterable here, the next line tests for an iterator.

Thanks, I'll use `utils` method.

piskvorky:
Hm, what's going on here, why is the commit diff so large? Most of this have already been merged in #2361, why does it appear here again?

anotherbugmaster:
I don't know, maybe that's because it's the same branch and github shows all commits since branching commit?

gensim/models/nmf.py
anotherbugmaster: I test for an iterable here, the next line tests for an iterator.

I need to distinguish between iterator and an ordinary iterable, hence can't use the utils.is_corpus method.

anotherbugmaster: Radim, NMF works correctly with repeatable iterables, non-repeatable iterables (iterators) and csc matrices. It does arbitrary number of passes in the first case, only one pass in the second case and arbitrary number of passes in the third case.

piskvorky: What do you mean by arbitrary? Corpus passes are expensive, we should know exactly how many passes there are.

anotherbugmaster: Pardon my english, I mean that you can make as much passes as you want.

gensim/models/nmf.py
piskvorky: Document streams are the primary interface. Other input format options can be mentioned too if supported, but not with the same strength.

We don't want people passing in CSC matrices, then complaining on our forum they're running out of memory…

gensim/models/nmf.py
piskvorky: I see. What happens when I specify 2 passes but supply a generator input?

anotherbugmaster: It will make 2 passes, as was intended.

So, if we can re-iterate over something - we can make multiple passes, if not - it will just do one pass and then stop.

gensim/models/nmf.py
anotherbugmaster: Yeah, I agree, just wanted to make it clear that the internal matrices are transposed in contrast to the ordinary LDA.

I'll just mention csc in the parameters section of the methods then.

gensim/models/nmf.py
anotherbugmaster: I'll mention that behavior in the description of the `pass` parameter

piskvorky: I cannot make 2 passes over a generator :)

I think the proper action when user specifies 2 passes, but only supplies a non-restartable sequence, is to raise a `ValueError`.

anotherbugmaster: I thought you meant generator function, like if we call `__iter__`, and it creates an iterator every time.

Anyway, I agree, I'll make it so that NMF will raise a `ValueError` in that case.

anotherbugmaster:
@piskvorky [Seems like github doesn't keep track of target branch commits in PR](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch), that's why we got all those old commits here.

anotherbugmaster:
@piskvorky Ok, I fixed everything, do you have any more suggestions?

I really want to release it already. :)

[Wikipeda benchmark](https://github.com/RaRe-Technologies/gensim/blob/f9dd1c5a2324b32496e23a81de638db085dd4f1c/docs/notebooks/nmf_wikipedia.ipynb)

Here's a benchmark table for the release notes:

<div>
<table border=\1\ class=\dataframe\>
  <thead>
    <tr style=\text-align: right;\>
      <th>model</th>
      <th>train_time</th>
      <th>mean_ram</th>
      <th>max_ram</th>
      <th>perplexity</th>
      <th>coherence</th>
      <th>l2_norm</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td><b>gensim_nmf</b></td>
      <td><b>00:26:09</b></td>
      <td><b>134 MB</b></td>
      <td>168 MB</td>
      <td><b>3817.239000</b></td>
      <td>-2.9077</td>
      <td><b>1986.291000</b></td>
    </tr>
    <tr>
      <td>lda</td>
      <td>01:27:46</td>
      <td>140 MB</td>
      <td><b>140 MB</b></td>
      <td>4701.976000</td>
      <td><b>-2.5286</b></td>
      <td>2273.643000</td>
    </tr>
    <tr>
      <td>sklearn_nmf</td>
      <td>00:50:30</td>
      <td>10753 MB</td>
      <td>16407 MB</td>
      <td>3943.036299</td>
      <td>-</td>
      <td>1987.280856</td>
    </tr>
  </tbody>
</table>
</div>

piskvorky:
@anotherbugmaster what's missing for the last "Fix tests" checkbox in this PR?

anotherbugmaster:
Oops, it's nothing, just forgot to mark it.

gensim/test/test_nmf.py
```diff
@@ -139,15 +160,15 @@ def testLargeMmapCompressed(self):
...
     def testDtypeBackwardCompatibility(self):
-        nmf_3_6_0_fname = datapath('nmf_3_6_0_model')
+        nmf_fname = datapath('nmf_model')
         test_doc = [(0, 1), (1, 1), (2, 1)]
-        expected_topics = [(0, 1.0)]
+        expected_topics = [(1, 1.0)]
 
         # save model to use in test
-        self.model.save(nmf_3_6_0_fname)
+        # self.model.save(nmf_fname)
```
piskvorky: Why is this test commented out?

gensim/models/nmf.py
```diff
@@ -548,10 +556,18 @@ def _setup(self, corpus):
...
+                    raise ValueError("Corpus is an iterator, only `passes=1` is possible")
+
+                first_doc_it, corpus = itertools.tee(corpus, 2)
+                first_doc = next(iter(first_doc_it))
+            else:
+                first_doc = next(iter(corpus))
+
             first_doc = matutils.corpus2csc([first_doc], len(self.id2word))
+
         self.w_std = np.sqrt(first_doc.mean() / (self.num_tokens * self.num_topics))
```
piskvorky: What happens when `hasattr(corpus, "__iter__")` is `False`? Isn't `first_doc` left uninitialized => error on this line?

gensim/test/test_nmf.py
anotherbugmaster: This is a compatibility test, one should uncomment this line in order to serialize the model once again

anotherbugmaster:
I'll try to fix it tonight, shouldn't take long

piskvorky:
> @piskvorky [Seems like github doesn't keep track of target branch commits in PR](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch), that's why we got all those old commits here.

I'll try to apply that recommended "fix" of changing the base branch to force a correct diff. Hold you hats!

piskvorky:
OK, that didn't help :(

@anotherbugmaster can you try the `rebase / push --force` trick? Or the "open a new PR" trick?

piskvorky:
Aaah, I think I see what's going on! @menshikh-iv squashed #2361 into a single commit when merging. So now all the individual commits are missing in git history => Github considers them new.

We have two options:
1) you rebase your changes on top of `develop` (instead of your old un-squashed branch)
2) we ignore the crazy diff for now.

In the future, I propose not squashing PRs when merging. CC @mpenkov thoughts?

anotherbugmaster:
I've tried to rebase the branch, but couldn't get it to work. :(

I've fixed the initialization, but due to different permutations of the dataset I should run wikipedia notebook once again, it will be ready tomorrow.

gensim/models/nmf.py
```diff
@@ -580,22 +561,25 @@ def _setup(self, corpus):
...
+        corpus : iterable of list of (int, float) or `csc_matrix` with the shape (n_tokens, n_documents)
             Training corpus.
             Can be either iterable of documents, which are lists of `(word_id, word_count)`,
             or a sparse csc matrix of BOWs for each document.
             If not specified, the model is left uninitialized (presumably, to be trained later with `self.train()`).
 
         """
-        if self._W is None:
-            corpus = self._setup(corpus)
+        if not (utils.is_corpus(corpus) or isinstance(corpus, scipy.sparse.csc.csc_matrix)):
```
piskvorky: `utils.is_corpus` is destructive, see its return value.

It would be better to check the "easier" condition first (csc_matrix), because Python does short-circuiting of boolean expressions. But in fact I don't think we need this check at all -- Python as a language relies on *duck typing*, not explicit type checks.

gensim/models/nmf.py
```diff
@@ -580,22 +561,25 @@ def _setup(self, corpus):
...
+            raise ValueError(
+                "Corpus type should be either `gensim.corpus.IndexedCorpus` or `scipy.sparse.csc.csc_matrix`"
```
piskvorky: That doesn't sound right -- we should accept any document stream, not just `IndexedCorpus`. (but see comment above -- I don't think we want this check at all)

gensim/models/nmf.py
```diff
@@ -620,6 +604,9 @@ def update(self, corpus):
                         num_terms=self.num_tokens,
                     )
 
+                if self._W is None:
+                    self._setup(v)
```
piskvorky: Nice solution!

How about a code comment, to make it clear what's going on:

```suggestion
                if self._W is None:
                    # This is the first document batch => initialize internal statistics from these initial documents.
                    # Such statistics are needed because FIXME why?
                    self._setup(v)
```

Btw what are the implications if these "first few documents" are not representative of the rest of the corpus?

gensim/models/nmf.py
```diff
@@ -569,22 +566,27 @@ def update(self, corpus):
...
         """
-        if self._W is None:
-            self._setup(corpus)
+        if not (utils.is_corpus(corpus) or isinstance(corpus, scipy.sparse.csc.csc_matrix)):
+            raise ValueError(
+                "Corpus type should be either `gensim.corpus.IndexedCorpus` or `scipy.sparse.csc.csc_matrix`"
+            )
+
+        if (hasattr(corpus, "__next__") or hasattr(corpus, "next")) and self.passes > 1:
+            raise ValueError("Corpus is an iterator, only `passes=1` is valid.")
```
piskvorky:
```suggestion
        if isinstance(corpus, collections.Iterator):
            raise ValueError("Corpus is an iterator, so it only supports `passes=1` but `passes=%s` was given." % self.passes)
```

gensim/models/nmf.py
```diff
@@ -569,22 +566,27 @@ def update(self, corpus):
...
 
         chunk_idx = 1
 
         for _ in range(self.passes):
             if isinstance(corpus, scipy.sparse.csc.csc_matrix):
                 grouper = (
-                    corpus[:, col_idx:col_idx + self.chunksize]
+                    corpus[:, col_idx:min(corpus.shape[1], col_idx + self.chunksize)]
```
piskvorky:
```suggestion
                    corpus[:, col_idx : min(corpus.shape[1], col_idx + self.chunksize)]  # older scipy (0.19 etc) throw an error when slicing beyond actual dimensions
```

gensim/models/nmf.py
anotherbugmaster: Then the `self._W` matrix might not be initialized as expected. I've changed the algorithm so that it uses the mean of the whole batch, it should be more stable.

anotherbugmaster: And if not - well, it will still work, but will take a little more time to converge.

gensim/models/nmf.py
anotherbugmaster: Done!

gensim/models/nmf.py
anotherbugmaster: Added the comment. :+1:

gensim/models/nmf.py
anotherbugmaster: Agreed, removed the check.

gensim/models/nmf.py
anotherbugmaster: Added the comment.

anotherbugmaster:
All done! Here's the updated wiki benchmark:

<div>
<table border=\1\ class=\dataframe\>
  <thead>
    <tr style=\text-align: right;\>
      <th>model</th>
      <th>train_time</th>
      <th>mean_ram</th>
      <th>max_ram</th>
      <th>perplexity</th>
      <th>coherence</th>
      <th>l2_norm</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td><b>gensim_nmf</b></td>
      <td><b>00:25:16</b></td>
      <td><b>130 MB</b></td>
      <td>165 MB</td>
      <td><b>3741.860600</b></td>
      <td>-2.9857</td>
      <td><b>1983.378700</b></td>
    </tr>
    <tr>
      <td>lda</td>
      <td>01:25:48</td>
      <td>140 MB</td>
      <td><b>140 MB</b></td>
      <td>4701.976000</td>
      <td><b>-2.5286</b></td>
      <td>2273.643000</td>
    </tr>
    <tr>
      <td>sklearn_nmf</td>
      <td>00:49:29</td>
      <td>10716 MB</td>
      <td>15961 MB</td>
      <td>3943.036299</td>
      <td>-</td>
      <td>1987.280856</td>
    </tr>
  </tbody>
</table>
</div>

anotherbugmaster:
> I missed the `nmf_wiki` notebook before. Can we merge this into `nmf_tutorial`?
> 
> I see no point in having two separate notebooks demonstrating the same model (it's just more maintenance, and people may miss one or the other).

The thing is that wiki notebook takes **a lot** of time to run, and nmf tutorial runs in just a couple of minutes. Not every user wants to re-run benchmark himself, but everybody wants to play with the model and see if it works. I think that it would be better to have 2 notebooks because of that.

piskvorky:
I think that's alright, as long as the wiki part is in its own section, with a clear warning that it takes a lot of time to run.

I still plan to go over docstrings + notebook and clean up a bit. I'll try to merge them.

piskvorky:
@anotherbugmaster I noticed the notebooks refer to `nmf._W` and `nmf._h`.

But the `_` underscore implies these are internal attributes that should not be accessed directly by users.

What is the correct way to access the actual trained `W` and `h` matrices? Why are they private?

piskvorky:
@anotherbugmaster I pushed my fixes for the NMF Tutorial in this branch https://github.com/RaRe-Technologies/gensim/tree/nmf_speedups.

I cannot push to yours, and Github doesn't allow me to create PR against your fork (some Github bug). So can you please merge that branch into your PR, so it appears here?

anotherbugmaster:
> @anotherbugmaster I noticed the notebooks refer to `nmf._W` and `nmf._h`.
> 
> But the `_` underscore implies these are internal attributes that should not be accessed directly by users.
> 
> What is the correct way to access the actual trained `W` and `h` matrices? Why are they private?

Well, my idea was that a user should interact with the model using methods such as `get_topics()`, `show_topics`, e.t.c., and not through direct access to the internal matrices. That's definitely negotiable though, I suppose we could make them public.

> I cannot push to your fork, and Github doesn't allow me to create PR against your fork (some Github bug, cannot select `base`). So can you please merge that branch into your PR, so the fix appears here?

@piskvorky, sure, I'll do it first thing tomorrow.

piskvorky:
The raw matrices are useful when building larger solutions. Imagine you want to input the word-topic matrix into another algorithm, say in PyTorch or Keras as np.array… Reconstructing it from `show_topics` will be a pain. Calling `model.W` is trivial.

anotherbugmaster:
> The raw matrices are useful when building larger solutions. Imagine you want to input the word-topic matrix into another algorithm, say in PyTorch or Keras as np.array… Reconstructing it from `show_topics` will be a pain. Calling `model.W` is trivial.

> I prefer to "hide" only implementation details… but I don't think `W` is going away in NMF :) (although the name is not great… `word_topics` looks better to me)

Actually, [get_topics()](https://github.com/RaRe-Technologies/gensim/blob/d712ad46bb923bb91bc8271e7104f7c7f5bec566/gensim/models/nmf.py#L212) method is just for that - getting W matrix, and optionally apply normalization. It works in the same way as LDA's `get_topics()`, I'd prefer to have a similar interface.

As for `self._h` matrix, I can't really think of any practical application for users, they're supposed to use `get_document_topics()` or `get_item()` to infer topic-document matrix of a certain corpus.

piskvorky:
Aha, you're right. I thought it returns the information in Gensim format (sparse tuples). No problem then.

piskvorky:
I made additional fixes -- @anotherbugmaster can you merge my branch into your PR?

Also, shouldn't the NMF training happen over TFIDF (not plain bag-of-words counts)?

The Wikipedia notebook needs a big rewrite (both code and language). I started in Section 4 (work-in-progress), planning to finally finish NMF this week.

piskvorky:
@gojomo could I ask you to give feedback on the [NMF tutorial notebook](https://github.com/RaRe-Technologies/gensim/blob/nmf_speedups/docs/notebooks/nmf_tutorial.ipynb)?

For the tutorial content/structure, but also as a native English speaker. Thanks!

anotherbugmaster:
> I pushed additional commits with tutorial fixes -- @anotherbugmaster can you merge my branch into your PR?

Sure, sorry for the delay, I'll do it ASAP.

> But shouldn't the NMF training happen over TFIDF (not plain bag-of-words counts)? Is pure BOW a good practice with NMF?

I don't know! Also, I'm not really sure how to measure perplexity if we'll use TF-IDF matrix. I could try implementing that in the next release, but now it seems unnecessary.

> And `Gensim's NMF implementation works as fast as Sklearn NMF` doesn't seem to be true -- I see 0.2s for sklearn, but 0.7s for gensim?

You're looking at the absolete version of the notebook, [here's the right one](https://github.com/RaRe-Technologies/gensim/blob/d712ad46bb923bb91bc8271e7104f7c7f5bec566/docs/notebooks/nmf_tutorial.ipynb).

Also, @gojomo, that's the right version of the NMF Tutorial.

image

> The Wikipedia notebook needs a rewrite, both code and language. I started it in Section 4 (work-in-progress), planning to finally finish NMF this week.

Nice! Tell me if I could help somehow.

piskvorky:
> You're looking at the absolete version of the notebook, [here's the right one](https://github.com/RaRe-Technologies/gensim/blob/d712ad46bb923bb91bc8271e7104f7c7f5bec566/docs/notebooks/nmf_tutorial.ipynb).

Nope! The "correct" version is the version I linked to, in my branch. That is the only version now (until you merge into your PR).

Abstain from making any changes to the notebook, or we'll have conflicts.

anotherbugmaster:
> > You're looking at the absolete version of the notebook, [here's the right one](https://github.com/RaRe-Technologies/gensim/blob/d712ad46bb923bb91bc8271e7104f7c7f5bec566/docs/notebooks/nmf_tutorial.ipynb).
> 
> Nope! The "correct" version is the version I linked to, in my branch. That is the only version now (until you merge into your PR).

Hm. I'll check it out, seems strange that time differs so much.

> Abstain from making any changes to the notebook, or we'll have conflicts.
> 
> > I don't know! Also, I'm not really sure how to measure perplexity if we'll use TF-IDF matrix. I could try implementing that in the next release, but now it seems unnecessary.
> 
> That's not good. We don't want to recommend bad practices in our tutorials.
> @anotherbugmaster can you check whether NMF should be run on top of regularized vectors such as TFIDF instead of bag-of-words counts (as I suspect it does, just like SVD)? If so, we'll have to fix this. People love to copy-paste from tutorials.

Ok, I'll check it, but, as I said, the perplexity metric won't be applicable in the case of TF-IDF.

anotherbugmaster:
I'm also not sure how to compare performance of LDA vs NMF in the case of TF-IDF matrix. Do you have any idea how could I do that?

piskvorky:
We have 2 options IMO:

1. include the time of TFIDF in the time of NMF (in case TFIDF is really required)
2. don't include TFIDF time (keep it as a separate "preprocessing" step)

anotherbugmaster:
Yes, but LDA would factorize BOW matrix while NMF would factorize TF-IDF, which means that I can't compare perplexities and l2 norms of the models. Which metrics should I use?

piskvorky:
Yes to which part? If perplexity doesn't apply to a model, we cannot use it. What does "perplexity" mean for NMF? And what does L2 mean for LDA?

Metrics that evaluate the topic quality directly (intrusion, coherence, upstream classification) should be applicable to all topic models.

anotherbugmaster:
I use [this method](http://qpleple.com/perplexity-to-evaluate-topic-models/) to measure perplexity. Of course, NMF is not probabilistic, but we can normalize its factors and get factors that take values from 0 to 1. It's not completely fair metric, of course.

When we use TF-IDF matrix, we lose information about BOW for each document, although we can reconstruct it.

L2 for LDA means that we multiply the learned word-topic matrix by the topic-document matrix of some validation corpus and then measure l2 norm of the difference between that and BOW of validation corpus.

Coherence and upstream classification can be used, but we'll need some labels to do that. What kind of task can we measure on Wikipedia corpus?

The intrusion metric would need human labeling, it's hard to reproduce.

Personally, I think that coherence and F1 on some classification task would be perfect.

anotherbugmaster:
Also, could I use TF-IDF corpus with LDA? I'm not sure what would be the effect of that.

piskvorky:
No, LDA should be trained without TFIDF (best LDA practice is to train on bag-of-words directly, that's how the algo was designed).

anotherbugmaster:
@piskvorky, just re-run the Olivetti faces part, got these results:

Here's my machine:

image

Here's RaRe hetzner:

image

I wonder what could cause such difference in time.

piskvorky:
Fast BLAS for dense operations? I'm on Macbook, which has good BLAS, and this is what I see:

screen shot 2019-02-15 at 15 39 37

piskvorky:
@anotherbugmaster I pushed some more notebook fixes, please merge again.

Could I ask you to resolve a few minor issues I found while running the Wiki notebook?

1. When a model starts training, can it log information about that's going on? I.e. that some training is starting, with what parameters, etc. Imagine it's your task to debug from a log -- what would help *you* understand what was going on?

screen shot 2019-02-16 at 08 10 01

2. Related: what are all the `INFO : Loss: 0.024172445101868792` log lines about? Can you make these (I assume progress?) lines more informative? What is the actual progress? Why are there so many, how do they differ from one another? Do we need to show the user 18 digits of precision?
   For some examples of more meaningful progress messages, have a look at e.g. LsiModel or LdaModel.

3. During NMF model saving, why aren't the internal large matrices being saved to separate files? This helps with memory-sharing models in production (`load(fname, mmap='r')`), and should happen automatically during `save()`. Can you please check what's going on with the NMF's internal matrix attributes that they're not being properly stored separately?

Thanks!

piskvorky:
Ping @anotherbugmaster -- what's the status on the points above please? Nr. 5 is the most critical (the metrics).

anotherbugmaster:
@piskvorky, sorry, I'm a bit busy this week.

Absolutely agree with 1-4, I'll see what I can do.

As for the metrics: The thing is, LDA model is intrinsically probabilistic while NMF is not. I think that the only two metrics we can use are:

- Coherence (I perceive the Gensim's CoherenceModel doesn't require a model to be probabilistic)
- F1 on some downstream task (Do we have some kind of labels for wikipedia articles?)

mpenkov:
@piskvorky I labeled this for 3.7.2 based on your comments in https://github.com/RaRe-Technologies/gensim/pull/2370#issuecomment-465940377

Are there any other PRs that should be included in 3.7.2? Could you please label them?

piskvorky:
I don't think we need a metric to use for all models (besides time & RAM).

But we do need the metrics we use to make sense -- otherwise what are we benchmarking?

I still don't understand what L2 is measuring right now, for LDA. Perplexity on NMF… maybe, but also unclear.

piskvorky:
@mpenkov I don't think so, we're waiting only on this PR for the release. It is critical though -- especially if we're comparing against other implementations (sklearn) and algorithms. Much potential for bad blood and embarrassment.

mpenkov:
OK, do you need my input on this PR, or do you have it under control?

piskvorky:
We identified the critical issues, now we need to resolve them :)

It's mostly on @anotherbugmaster (and I guess his mentor, Ivan), but of course your input is always welcome too.

When do you reckon we could release @mpenkov , is everything else in place?

anotherbugmaster:
I'm going to resolve the issues this weekend, is it ok?

piskvorky:
@anotherbugmaster sure. Let us know when ready for the final review.

piskvorky:
@anotherbugmaster what's the status?

anotherbugmaster:
@piskvorky working on it, plan to finish it by the end of the day.

anotherbugmaster:
Investigated the `.save()` issue. It's not an issue at all, all internal structures have fewer elements than the default threshold (10 * 1024**2), that's why they're not serialized separately.

anotherbugmaster:
The only quality metric I decided to use is coherence (because we don't have any labels for the Wikipedia dataset). I'll keep F1 in the 20-newsgroups section though.

piskvorky:
> Investigated the `.save()` issue. It's not an issue at all, internal structures have fewer elements than the default threshold (10 * 1024**2), that's why they're not serialized separately.

Aha. And this is on 20 newsgroups, right? The internal matrix built from Wiki surely is larger than 10 MB -- can you confirm please?

> The only quality metric I decided to use is coherence (because we don't have any labels for the Wikipedia dataset). I'll keep F1 in the 20-newsgroups section though.

That doesn't sound right. L2 is the most important metric for NMF, we definitely want that.

IMO we don't need to run LDA on Wikipedia. But we do want to compare against sklearn.

anotherbugmaster:
>Aha! And this is on 20 newsgroups, right? The internal matrix built from Wiki surely is larger than 10 MB, and should be stored separately -- can you confirm please?

It seems like the documentation for the `SaveLoad._save_specials()` is misleading: `sep_limit` is measured in number of elements, not in [bytes](https://github.com/RaRe-Technologies/gensim/blob/a3dbdcc59f3823b530db3ac38afdbb3df2761d6f/gensim/utils.py#L591). The largest matrix is `nmf._W`, which size is `200000 * 50 == 1e8 < 10 * 1024**2`.

>That doesn't sound right. L2 is the most important metric for NMF, we definitely want that (plus of course "RAM" and "time"). And if coherence works both on LDA and NMF, why not also use it also on the sklearn results?

Ok then, I'll keep L2 for the NMFs, but remove it for LDA (this metric in this case is not obvious and not entirely correct).

I'll do the wrapper for the sklearn NMF and `gensim.models.CoherenceModel`, of course.

>@anotherbugmaster What about the best practices when it comes to post-processing bag-of-words inputs for NMF, e.g. via TF-IDF? Does it help the algo achieve better results through normalization / narrower feature value distribution?

Oh, I forgot to mention, sorry! Yes, you're right, it improved the F1 on 20-newsgroups by 2%. I think I'll just pass the `TfidfModel(corpus)` to the NMF in the tutorials, is it ok? Or I can just embed this step right into the `nmf.update()`.

piskvorky:
> It seems like the documentation for the `SaveLoad._save_specials()` is misleading

You're right, this looks like a bug. Dense arrays are indeed measured in bytes (correct), but sparse arrays are measured in elements (incorrect). Let me raise a separate ticket for this.

> I think I'll just pass the `TfidfModel(corpus)` to the NMF in the tutorials, is it ok? Or I can just embed this step right into the `nmf.update()`.

`TfidfModel(dictionary)` should be more efficient (no need for an extra corpus pass, all the necessary stats are already collected in the `Dictionary`). And then working with (or storing to disk) `tfidf[bow_corpus]`, instead of just `bow_corpus`. Definitely no embedding inside `nmf.update`, that's not the right place for combining models :)

piskvorky:
@anotherbugmaster I looked at the code, and `sep_limit` isn't actually comparing the number of elements in sparse arrays. It's comparing the number of *non-zero elements*: #2396.

So if I got this right, the code should be storing separately **more sparse arrays** than necessary, not less.

anotherbugmaster:
>@anotherbugmaster I checked the code, and `sep_limit` isn't actually comparing the number of elements in sparse arrays. It's comparing the number of _non-zero elements_: #2396.

So if I got this right, the code should be storing separately **more sparse arrays** than necessary, not less.

Right, but our NMF doesn't have any sparse attributes now. :) The only thing that's sparse is a converted corpus batch, but it's not written in any internal structure. I also manually substituted the `nmf._W` matrix with the dense array of shape `(1024 * 10, 1025)`, serialized the model, and matrix got saved into the separate file.

piskvorky:
OK, great!

These are lesser concerns though. What we need for the release urgently are the fixed metrics (top priority) + correct best practices in tutorial (tfidf).

gensim/matutils.py
```diff
@@ -141,8 +141,7 @@ def corpus2csc(corpus, num_terms=None, dtype=np.float64, num_docs=None, num_nnz=
             if printprogress and docno % printprogress == 0:
                 logger.info("PROGRESS: at document #%i/%i", docno, num_docs)
             posnext = posnow + len(doc)
-            indices[posnow: posnext] = [feature_id for feature_id, _ in doc]
-            data[posnow: posnext] = [feature_weight for _, feature_weight in doc]
+            indices[posnow: posnext], data[posnow: posnext] = zip(*doc)
```
piskvorky: -1 on all the changes in this commit: `zip(*)` fails on empty docs, and is harder to read.

gensim/models/nmf.py
```diff
@@ -606,7 +606,7 @@ def update(self, corpus, chunksize=None, passes=None, eval_every=None):
                         chunk,
                         num_terms=self.num_tokens,
                         num_docs=len(chunk),
-                        num_nnz=sum(1 for _ in itertools.chain.from_iterable(chunk)),
+                        num_nnz=sum(len(doc) for doc in chunk),
```
piskvorky: Is `doc` always guaranteed to have `len()`?

gensim/matutils.py
anotherbugmaster: But it makes it twice as fast! I can add a comment and a check for the document length.

gensim/models/nmf.py
anotherbugmaster: Hm. Ok, we `doc` isn't required to have `len`. Explicit specification of the parameters speed up the conversion a bit, but I think it's not that much of a gain. I'll get back an implicit conversion then.

gensim/matutils.py
piskvorky: OK… let's see the cleaned up code with comments.

IMO this method should be cythonized for a proper speed up. Not sure if obfuscation is the way to go here. CC @mpenkov

piskvorky:
@anotherbugmaster instead of adding unrelated commits and optimizations (not urgent,  not asked for), can you finish the actual critical issues so we can release?

These unrelated tangents are worrying and never-ending.

anotherbugmaster:
The thing is that Wikipedia benchmark is a bit strange now: the large part of the training is just reading the file and converting the corpus to csc. I'm blocked by the building of the training corpus for a couple of hours anyway, so I'm just polishing things up.

piskvorky:
OK, makes sense. Preparing input and preprocessing is typically the largest part of training (also true for LSI, LDA, Word2vec etc), that's not surprising.

The important bit for this PR is that we show best practices, and make the processing comparable (i.e. metrics make sense, all models use the same preprocessing, etc).

gensim/matutils.py
```diff
@@ -141,7 +141,8 @@ def corpus2csc(corpus, num_terms=None, dtype=np.float64, num_docs=None, num_nnz=
             if printprogress and docno % printprogress == 0:
                 logger.info("PROGRESS: at document #%i/%i", docno, num_docs)
             posnext = posnow + len(doc)
-            indices[posnow: posnext], data[posnow: posnext] = zip(*doc)
+            # zip(*doc) transforms doc to (token_indices, token_counts]
+            indices[posnow: posnext], data[posnow: posnext] = zip(*doc) if len(doc) > 0 else ([], [])
```
piskvorky: `if doc` more Pythonic than `if len(doc) > 0`.

gensim/models/nmf.py
```diff
@@ -649,8 +648,8 @@ def _solve_w(self):
 
         def error(WA):
             return (
-                0.5 * np.einsum('ij,ij', WA, self._W, optimize=True)
-                - np.einsum('ij,ij', self._W, self.B, optimize=True)
+                0.5 * np.einsum('ij,ij', WA, self._W)
+                - np.einsum('ij,ij', self._W, self.B)
```
piskvorky: This expression could be on one line. But the function deserves a docstring: what is it calculating, what is `WA`?

gensim/models/nmf.py
```diff
@@ -558,7 +558,10 @@ def update(self, corpus, chunksize=None, passes=None, eval_every=None):
...
 
         try:
-            lencorpus = len(corpus)
+            if isinstance(corpus, scipy.sparse.csc.csc_matrix):
+                lencorpus = corpus.shape[1]
+            else:
+                lencorpus = len(corpus)
         except Exception:
             logger.warning("input corpus stream has no len(); counting documents")
             lencorpus = sum(1 for _ in corpus)
```
piskvorky: What happens when `corpus` is a generator?

anotherbugmaster:
> What happens when corpus is a stream (generator)?

It'll break. I think I'll just set lencorpus to `?` or something like that.

> Can you add unit tests for generator input please? Let's make sure there are no surprises there, that NMF is truly online.

Sure.

anotherbugmaster:
Ok, I've updated the notebook, it'll finish executing in about 3 hours, I'll re-upload it when it happens.

gensim/models/nmf.py
```diff
@@ -557,30 +557,35 @@ def update(self, corpus, chunksize=None, passes=None, eval_every=None):
...
             if isinstance(corpus, scipy.sparse.csc.csc_matrix):
                 lencorpus = corpus.shape[1]
             else:
                 lencorpus = len(corpus)
+
+            if chunksize is None:
+                chunksize = min(lencorpus, self.chunksize)
+
+            evalafter = min(lencorpus, (eval_every or 0) * chunksize)
         except Exception:
```
piskvorky: If this is a check only for `len(corpus)`, it is better make the try/catch around that code. And be more specific with `AttributeError` instead of generic `Exception`.

gensim/models/nmf.py
```diff
@@ -557,30 +557,35 @@ def update(self, corpus, chunksize=None, passes=None, eval_every=None):
...
-            logger.warning("input corpus stream has no len(); counting documents")
-            lencorpus = sum(1 for _ in corpus)
+            logger.warning("input corpus stream has no len()")
```
piskvorky: No `len()` is a perfectly valid input state, the warning is too severe. Maybe INFO?

anotherbugmaster: Alright

gensim/models/nmf.py
anotherbugmaster: Ok!

gensim/models/nmf.py
anotherbugmaster: Will do

anotherbugmaster:
Ok, the notebook is ready!

@piskvorky, @mpenkov, sorry for the delay, please review the [updated version](https://github.com/RaRe-Technologies/gensim/blob/063caa73205696ce13e3f7d360db176ba05b70d4/docs/notebooks/nmf_tutorial.ipynb).

@piskvorky I'll fix your objections today.

piskvorky:
@anotherbugmaster can you remove the long log outputs from the final version (cells 17, 35…)?

That is, keep the logging in the notebook (when people run the notebook themselves, they should see the log), but "Clear cell output" on these particular cells, commit and push. Thanks!

For NMF, what's up with all the `W error diff: 0.0`? What is happening in the training?

anotherbugmaster:
> @anotherbugmaster thanks! Review notes:
> 
>     1. Can you remove the long log outputs from the final version (cells 17, 35…)?
>        That is, keep the logging in the notebook (when people run the notebook themselves, they should see the log), but "Clear cell output" on these particular cells, commit and push. Thanks!

Of course!

>     2. What's up with all the `W error diff: 0.0`? What is happening in the training?

Error on the matrix W is too low before the update, so it's not updating.

>     3. The section on `Residuals matrix of the last batch of shape (words, batch)` seems unfinished.

Oops, removed that, thanks.
 
>     4. For coherence, is lower better or higher better? How to interpret these benchmark numbers?

Higher is better, I'll add that to the notebook.
 
> screen shot 2019-03-04 at 10 31 22
> 
>     1. How is `794 MB` 100x more RAM than `17940 MB`?

I haven't updated the numbers yet. :)

piskvorky:
>> 1. How is `794 MB` 100x more RAM than `17940 MB`?
> I haven't updated the numbers yet. :)

Aha. So our RAM footprint of NMF went up 5x, in the last commits? What happened?

anotherbugmaster:
```
@contextmanager
def measure_ram(output, tick=5):
    def _measure_ram(pid, output, tick=tick):
        py = psutil.Process(pid)
        with open(output, 'w') as outfile:
            while True:
                memory = py.memory_info().rss
                outfile.write("{}\n".format(memory))
                outfile.flush()
                time.sleep(tick)

    pid = os.getpid()
    p = Process(target=_measure_ram, args=(pid, output, tick))
    p.start()
    yield
    p.terminate()
```

I used to subtract `start_ram` from the `memory` variable, but it resulted in the negative values on your Macbook. Not sure what's happening here, but my guess is that the GC removes deletes some structure right after `start_ram` were measured, and it results in negative values.

piskvorky:
That's surprising. I wouldn't expect RAM during training to be **lower** than RAM right after Python process is started.

I'm thinking `psutil` may be at fault… memory management works very differently on OSX (Mach Kernel) than on Linux.

piskvorky:
@anotherbugmaster thanks for the final updates!

* In `l2_norm - L2 norm of v - Wh`, the `v` is undefined. It's the entire input corpus, right?
* For the wiki training: L² norm of `94.9842` on a 5M x 30k matrix implies an average element error of only 0.0002. That's pretty incredible for 50 topics! I don't see any obvious issue with the code, but is such a low number expected?

anotherbugmaster:
>     * In `l2_norm - L2 norm of v - Wh`, the `v` is undefined. It's the entire input corpus, right?

Yes, that's correct.

>     * For the wiki training: L² norm of `94.9842` on a 5M x 30k matrix implies an average element error of only 0.0002. That's pretty incredible for 50 topics! I don't see any obvious issue with the code, but is such a low number expected? How come L2 on 20newsgroups is only 2x less (`42.4759`), even though it's a ~10,000x smaller corpus?

Well, our corpus is not simple a BOW anymore, but a TF-IDF, that's why an l2 has become so small. It also explains why the difference between 20newsgroups and Wikipedia results is so small.

piskvorky:
Wait, 20newsgroups still uses plain BOW, instead of TFIDF?

That's not good -- let's be consistent in best practices. If we recommend TFIDF for text models, let's use TFIDF consistently everywhere. Can you fix that?

piskvorky:
@anotherbugmaster looking at the code, Sklearn uses `SklearnNmf(n_components=5, random_state=42).fit(train_dense_corpus_tfidf)`.

Is that really consistent with Gensim NMF? Why not supply sparse input, if Sklearn supports it?

anotherbugmaster:
> > Well, our corpus is not simple a BOW anymore, but a TF-IDF, that's why an l2 and the difference between the 20newsgroups and Wikipedia results have become so small.
> 
> Wait, 20newsgroups still uses plain BOW, instead of TFIDF? The reason for the relatively large L2 on 20newsgroups is BOW input, whereas low L2 on Wikipedia is due to TFIDF input?
> 
> That's not good -- let's be consistent in best practices. If we recommend TFIDF for text models, let's use TFIDF consistently everywhere (also for sklearn). Can you please fix that?

No, TF-IDF used in 20newsgroup too, but TF-IDF significantly lowers the l2 norm.

anotherbugmaster:
> @anotherbugmaster looking at the notebook code, for Sklearn I see `SklearnNmf(n_components=5, random_state=42).fit(train_dense_corpus_tfidf)`.
> 
> Is the dense/sparse input consistent with Gensim NMF? Why not supply sparse input, if Sklearn supports it?

Right, I forgot to change dense matrix to sparse here. Will do!

piskvorky:
> No, TF-IDF used in 20newsgroup too, but TF-IDF significantly lowers the l2 norm.

Yes, but then it lowers the norm for both 20newsgroups and wiki. So why the (only 2x) difference? What's the intuition behind the norm being so similar, for 2 vastly different corpora?

I'm not saying it's wrong, just very surprising, so I'd like to understand what's going on.

piskvorky:
@anotherbugmaster to ground the intuition -- what's the L2 norm for a "trivial" reconstruction = zero matrix? In other words, what's the Frobenious norm of the original input matrix?

piskvorky:
@mpenkov we may have to release without NMF.

piskvorky:
@mpenkov how about this: let's merge the current PR, to fix PR diff issues due to previously "squashed" PR and add the NMF bug fixes introduced since 2.4.1.

But let's wait with announcing / promoting NMF until the outstanding issues are cleared (sklearn density, strange logs full of zeros, strangely low metrics).

mpenkov:
But let's wait with announcing / promoting NMF until the outstanding issues are cleared (sklearn density, strange logs full of zeros, strangely low metrics).

@piskvorky We can't really merge this PR without announcing _something_, because we've changed some publicly visible behavior, e.g. removed log_perplexity method, added parameters to update method, etc.

Perhaps I'm misunderstanding what you mean by "announcing"?

piskvorky:
Announcing = telling the world Gensim has a cool new model: NMF. Sharing the links to tutorials, benchmarks.

The merge of NMF PRs in the previous releases was rushed and premature. It had some critical bugs, shouldn't have been merged. As a result, I didn't announce NMF yet. So while technically released, NMF is not really "public" yet, I consider it an internal work-in-progress.

anotherbugmaster:
@piskvorky, I computed the l2 norms:

20newsgroups, `W.dot(H), tf-idf`: 3.3029151214383066
Wikipedia, `W.dot(H), tf-idf`: 30.495559130718288

piskvorky:
@anotherbugmaster what is the Frobenius norm of the input matrix?

anotherbugmaster:
@piskvorky just updated the comment

anotherbugmaster:
As for logs full of zeros, I can just log the raw W error, but don't know if it'll be any more informative.

piskvorky:
Thanks. So our baseline is `l2_norm=100.0` (a decomposition that does nothing, all zeros), and sklearn produces `l2_norm=94.8459` and gensim `l2_norm=94.9842`. Ideal would be 0.0 (perfect reconstruction; impossible in practice due to limited topics and sparsity constraints). Correct?

Would you mind adding LSI (SVD) and reporting the L2 error there? It sounds OK, but just for comparison.

Thanks!

> As for logs full of zeros, I can just log the raw W error, but don't know if it'll be any more informative.

I'm not concerned about the logging itself, more what it tells us about the training process. How do you interpret the mostly-zero-errors during the training? What conclusions should users draw from that? Subsample their input? Increase precision? What is going on?

anotherbugmaster:
I've replace dense tf-idf to csc in the `20newsgroups` section. Sklearn is a lot faster now.

image

piskvorky:
@anotherbugmaster aha. Good we didn't publish those flawed benchmarks :-)

What is your explanation for sklearn being ~5x faster?

anotherbugmaster:
> Thanks. So our baseline is `l2_norm=100.0` (a decomposition that does nothing, all zeros), and sklearn produces `l2_norm=94.8459` and gensim `l2_norm=94.9842`. Ideal would be 0.0 (perfect reconstruction; impossible in practice due to limited topics and sparsity constraints). Correct?
> 
> Would you mind adding LSI (SVD) and reporting the L2 error there? It sounds OK, but just for comparison.
> 
> Thanks!
> 
> > As for logs full of zeros, I can just log the raw W error, but don't know if it'll be any more informative.
> 
> I'm not concerned about the logging itself, more what it tells us about the training process. How do you interpret the mostly-zero-errors during the training? What conclusions should users draw from that? Subsample their input? Increase precision? What is going on?

I can add a comment. If W error diff equals zero than it means that the model already reconstructs current batch as good as it gets and doesn't learn nothing new.

anotherbugmaster:
> Would you mind adding LSI (SVD) and reporting the L2 error there? It sounds OK, but just for comparison.

Sure, no problem, I'll re-run the notebook today.

piskvorky:
> I can add a comment. If W error diff equals zero than it means that the model already reconstructs current batch as good as it gets and doesn't learn nothing new.

I understand that, but what does this mean for the user? Subsample? Increase precision somewhere?

Surely "nothing is being trained" is not a desirable training progress.

anotherbugmaster:
>What is your explanation for sklearn being ~5x faster?

Obviously, it doesn't need to learn iteratively, and due to the small size of the dataset it became an advantage. Things work differently on the Wikipedia: large corpus is learned much faster with an iterative approach.

anotherbugmaster:
>I understand that, but what does this mean for the user? Subsample? Increase precision somewhere?

>Surely "nothing is being trained" is not a desirable training progress.

Oh, I see. Yes, user can either subsample the trainset or increase the precision of the model (`w_stop_iteration` and `h_stop_iteration` parameters).

piskvorky:
> Things work differently on the Wikipedia: large corpus is learned much faster with an iterative approach.

Yes, but see our other thread, about the model not really learning anything in reality. Is it really faster, on a corpus where it learns something? Wouldn't a subsample learn similar/better topics, in faster time (and 5x faster with sklearn)?

anotherbugmaster:
Hm, I see. I think that the most obvious advantage over the Sklearn is the huge difference in RAM footprint, but I'll definitely need to check the subsampling case.

mpenkov:
@piskvorky My understanding is that you'd like to quietly merge this branch without mentioning anything in the change log. Is that correct?

@anotherbugmaster Please let me know when this is ready to merge.

piskvorky:
@mpenkov correct. Let's merge and release 3.7.2 ASAP (and finish the rest of NMF later).

> without mentioning anything in the change log

We can mention these NMF fixes in the release log. It's not a channel many people actively follow, still counts as "silent" :) I'll wait with real promo until the tutorial is more compelling.

anotherbugmaster:
@mpenkov Sure!

Ok then, I'll check the possible caveats more thoroughly and report the results next week.

piskvorky:
@anotherbugmaster any progress on the SVD (LSI) stats, for comparison?

Plus investigating the "NMF not learning anything" (subsampling, precision). Cheers.

mpenkov:
@anotherbugmaster I'm going to merge this so we can go ahead with the next release.

Please continue working on this PR and pushing your changes. We'll merge again in the near future.

anotherbugmaster:
@piskvorky, sorry, I forgot to make an experiment :(

I think I'll report something on this weekend.

piskvorky:
@anotherbugmaster please start a new branch from the `develop` branch, to have a clean PR. A previously squashed merge is showing as hundreds of new commits in this PR; let's avoid that in the new PR. Cheers.

piskvorky:
@anotherbugmaster what's the status here? I'd love to get NMF wrapped up nicely, so we can promote it.

anotherbugmaster:
@piskvorky Sorry, Radim, still haven't finished research yet. I'll do it ASAP.

piskvorky:
@anotherbugmaster how is it going? Let's get this finished while there's still momentum (or remove it).

anotherbugmaster:
@piskvorky, I've changed the algorithm so that the model updates every batch even if an error is already low. Good news is that it didn't affect performance, here are the results:

image

I've also changed an error in logs, now it looks like this:

image

The lower the error, the better.

piskvorky:
@anotherbugmaster Why, what's your rationale / what's the advantage of updating the model even it makes no difference? What was the time + L2_norm before? (to easily compare to this latest `27:20` and `94.97`)

Can you also give the reconstruction error for LSI on the same data please, for anchoring. Cheers.

piskvorky:
@anotherbugmaster ping -- can we finish this please?

anotherbugmaster:
@piskvorky I'll run an LSI tomorrow.

anotherbugmaster:
Ok, here's Wikipedia comparison:

image

anotherbugmaster:
>I'm not sure how forcing near-zero updates answers them.

Model updates every batch now (it skipped a lot of them before, that wasn't correct behavior).

piskvorky:
Thanks @anotherbugmaster . Let me ping @mpenkov and @gojomo for a final review -- are the motivation, evaluation and usage instructions convincing? Anything crucial missing, before we push this out "for real"? Thanks!

piskvorky:
@anotherbugmaster can you open a new PR (starting from upstream `develop`), with your recent changes to the update process + the LSI comparison? Cheers.

mpenkov:
@anotherbugmaster There are some logging-related bugs. I suspect they are blocking the release of our conda [feedstock](https://github.com/conda-forge/gensim-feedstock/pull/26). I'll highlight the problem iines in a separate review.

gensim/models/nmf.py
```diff
@@ -602,44 +620,57 @@ def update(self, corpus):
                         num_terms=self.num_tokens,
                     )
 
+                    chunk_len = len(chunk)
+
+                logger.info(
+                    "PROGRESS: pass %i, at document #%i/%i",
+                    pass_, chunk_idx * chunksize + chunk_len, lencorpus
```
mpenkov: lencorpus may be a float (inf) here, but the formatstring requires it to be an integer

gensim/models/nmf.py
```diff
@@ -564,36 +528,90 @@ def _setup(self, corpus):
...
+            return
 
-        for _ in range(self.passes):
+        if isinstance(corpus, collections.Iterator) and self.passes > 1:
+            raise ValueError("Corpus is an iterator, only `passes=1` is valid.")
+
+        logger.info(
+            "running NMF training, %s topics, %i passes over the supplied corpus of %i documents, evaluating l2 norm "
+            "every %i documents",
+            self.num_topics, passes, lencorpus if lencorpus < np.inf else "?", evalafter,
```
mpenkov: The formatstr expects an integer, but you will end up passing a string if lencorpus is np.inf.

piskvorky:
> @anotherbugmaster can you open a new PR (starting from upstream develop), with your recent changes to the update process + the LSI comparison? Cheers.

@anotherbugmaster ping. Do you plan to finish NMF?

PR #2382, opened 2019-02-14
Clean up FastText Cython code, fix division by zero

### LINKED ISSUE #2377: FastText segfaults for some ngram ranges
hey,

I feel something is bad with my text data. I used common crawl for text data and even without any preprocesing (to dismiss as error source) the fasttext model stops during training and just exists without any messages. That is only for some ngram ranges like:

```
model=FastText(sentences,min_n=4)
print(model)
```
In this case no modell summary is printed and the model was not built but there is no message at all.

Giving as paramater instead just `min_n=2, max_n=4` the model is built and works fine.

I used my text data and the sentence iterator like here

```
>>> from gensim.utils import tokenize
>>> import smart_open
>>>
>>>
>>> class MyIter(object):
...     def __iter__(self):
...         path = datapath('crime-and-punishment.txt')
...         with smart_open.smart_open(path, 'r', encoding='utf-8') as fin:
...             for line in fin:
...                 yield list(tokenize(line))

```

with above fasttext statement. 

Might this comes from bad encoding like `\xad`? How to find out what is bad in my data?

mpenkov:
The main goal of this PR is to fix #2377 

I found the original Cython code hard to debug, so I improved it by moving long code blocks out to separate functions and introducing better variable names. Functionally, it's the same, but it's a bit easier to read, and it helped me uncover the zero division problem.


mpenkov:
@piskvorky We have several options of proceeding with this PR.

1. Include in the next bugfix release (3.7.2)
2. Wait until next minor release (3.8.0)
3. Split the bug fix from the Cython improvements, release the former in 3.7.2, and the latter in 3.8.0

WDYT? Any preferences?

piskvorky:
I don't understand the trade-offs. Can I leave this decision with you @mpenkov ?

mpenkov:
I'm gonna go with option 3). It fixes the immediate bug, and avoids the risk of introducing new ones.

Opened https://github.com/RaRe-Technologies/gensim/pull/2404

We'll merge the Cython improvements _after_ the bugfix release.

PR #2392, opened 2019-02-22
Fix the off-by-one bug in the TFIDF model.

### LINKED ISSUE #2375: Off-by-one counting error in TFIDF
Building a TFIDF model outputs this in the log:

```
INFO - 2019-02-06 16:09:54,670 - resulting dictionary: Dictionary(10000 unique tokens: [u'writings', u'foul', u'prefix', u'woods', u'hanging']...)
INFO - 2019-02-06 16:10:01,718 - collecting document frequencies
INFO - 2019-02-06 16:10:01,718 - PROGRESS: processing document #0
INFO - 2019-02-06 16:10:02,162 - calculating IDF weights for 1701 documents and 9999 features (2220559 matrix non-zeros)
```

Note the "9999 features", despite the dictionary having 10000 features. (any dictionary and any number of features -- the actual values don't matter, it's always off-by-one).

This seems to be a bug introduced in [this refactoring](https://github.com/RaRe-Technologies/gensim/commit/af01bc702f7ad48ed9ecacb256d9f6f7da6d3bc4#diff-896942304f9417b1b8842823b19d7431R128).

Not sure if this is an isolated problem or what other places are affected by similar "optimizations".

AMR-KELEG:
Fixes #2375.
Use len to compute the number of features.
Since the ids are zero-indexed, Using max causes an off-by-one bug.

### DIFF
--- a/gensim/models/tfidfmodel.py
+++ b/gensim/models/tfidfmodel.py
@@ -392,7 +392,7 @@ class TfidfModel(interfaces.TransformationABC):
         # and finally compute the idf weights
         logger.info(
             "calculating IDF weights for %i documents and %i features (%i matrix non-zeros)",
-            self.num_docs, 1 + max([-1] + list(dfs.keys())), self.num_nnz
+            self.num_docs, max(dfs.keys()) + 1 if dfs else 0, self.num_nnz
         )
         self.idfs = precompute_idfs(self.wglobal, self.dfs, self.num_docs)
 
--- a/gensim/models/tfidfmodel.py
+++ b/gensim/models/tfidfmodel.py
@@ -390,10 +390,9 @@ class TfidfModel(interfaces.TransformationABC):
         self.num_nnz = numnnz
         self.dfs = dfs
         # and finally compute the idf weights
-        n_features = len(dfs) if dfs else 0
         logger.info(
             "calculating IDF weights for %i documents and %i features (%i matrix non-zeros)",
-            self.num_docs, n_features, self.num_nnz
+            self.num_docs, 1 + max([-1] + list(dfs.keys())), self.num_nnz
         )
         self.idfs = precompute_idfs(self.wglobal, self.dfs, self.num_docs)
 
--- a/gensim/models/tfidfmodel.py
+++ b/gensim/models/tfidfmodel.py
@@ -390,7 +390,7 @@ class TfidfModel(interfaces.TransformationABC):
         self.num_nnz = numnnz
         self.dfs = dfs
         # and finally compute the idf weights
-        n_features = max(dfs) if dfs else 0
+        n_features = len(dfs) if dfs else 0
         logger.info(
             "calculating IDF weights for %i documents and %i features (%i matrix non-zeros)",
             self.num_docs, n_features, self.num_nnz


gensim/models/tfidfmodel.py
```diff
@@ -390,7 +390,7 @@ def initialize(self, corpus):
         self.num_nnz = numnnz
         self.dfs = dfs
         # and finally compute the idf weights
-        n_features = max(dfs) if dfs else 0
+        n_features = len(dfs) if dfs else 0
```
mpenkov: dfs is never None, so we can use len without the ternary operator.

```suggestion
        n_features = len(dfs)
```

Furthermore, I think n_features is a useless variable here. We're defining it here, only to use it once in the logging statement below.

gensim/models/tfidfmodel.py
```diff
@@ -390,7 +390,7 @@ def initialize(self, corpus):
...
         logger.info(
             "calculating IDF weights for %i documents and %i features (%i matrix non-zeros)",
             self.num_docs, n_features, self.num_nnz
```
mpenkov:
```suggestion
            self.num_docs, len(dfs), self.num_nnz
```

gensim/models/tfidfmodel.py
piskvorky: What will this do when `dfs = {1: 123, 10000: 124}`?

See the original line (before someone introduced this off-by-one bug in some refactoring) for the correct solution.

AMR-KELEG: Well for the case `dfs = {1: 123, 10000: 124}`, I thought that 2 should be the correct number of features since IDs are just numbers that are attached to the tokens.

AMR-KELEG: I agree.

AMR-KELEG: @piskvorky Using the original line `1 + max([-1] + dfs.keys())` results in a TypeError:
`TypeError: can only concatenate list (not "dict_keys") to list`
Should I just cast the keys to a python list?

piskvorky: I see what you mean, but following such logic would lead to segfaults. Gensim relies on the fact that *all* features in a given feature range are valid, even if not present in a particular training corpus. They maybe be still present in a test / prediction-time corpus.

I'm strongly -1 on redefining the meaning of  "number of features" here, even if it's just for logging. It would lead to inconsistencies and ultimately bugs.

AMR-KELEG: I agree that this change might be confusing if other parts of gensim assume that missing token IDs are implicitly assigned to features yet to be defined.

piskvorky: They aren't yet-to-be-defined. They are defined, but didn't appear in that particular corpus.

piskvorky:
Looks good, thanks @AMR-KELEG ! Can you check whether the same issues is present somewhere else, too? (e.g. from that same refactoring)

gensim/models/tfidfmodel.py
```diff
@@ -390,10 +390,9 @@ def initialize(self, corpus):
         self.num_nnz = numnnz
         self.dfs = dfs
         # and finally compute the idf weights
-        n_features = max(dfs) if dfs else 0
         logger.info(
             "calculating IDF weights for %i documents and %i features (%i matrix non-zeros)",
-            self.num_docs, n_features, self.num_nnz
+            self.num_docs, 1 + max([-1] + list(dfs.keys())), self.num_nnz
```
piskvorky:
```suggestion
            self.num_docs, max(dfs.keys()) if dfs else 0, self.num_nnz
```

AMR-KELEG:
> Looks good, thanks @AMR-KELEG ! Can you check whether the same issues is present somewhere else, too? (e.g. from that same refactoring)

I have checked the commit and I don't think this edit was introduced in any other files.
The only edit that might look similar is this one (https://github.com/RaRe-Technologies/gensim/commit/af01bc702f7ad48ed9ecacb256d9f6f7da6d3bc4#diff-327b58fcebe80ddff670af4d52bdde1cR232)
However, I believe this edit won't case errors.

I have also searched the whole project for this line `n_features = max(dfs) if dfs else 0` and it was only present in the tfidf model script.

PR #2397, opened 2019-02-27
Make termsim matrix positive definite even with negative similarities

Witiko:
When a `SparseTermSimilarityMatrix` object is constructed with non-default parameters that satisfy the following constraints: `symmetric and positive_definite and threshold < 0`, the matrix embedded in the constructed object may not be positive definite. This PR hardens the code against negative term similarities, making sure that the term similarity matrix is [always positive definite](https://en.wikipedia.org/wiki/Diagonally_dominant_matrix).


Witiko:
Note that this set of parameters is unusual, since it makes sense to ground the similarities to zero (that is why we are using a sparse matrix afterall), so although this bugfix changes the produced matrices, relatively few users should be affected.

mpenkov:
Thank you for this PR. Unfortunately, I'm not very familiar with the underlying code, so it's difficult for me to comment on the substance of your PR. I do notice that it appears to be a bugfix. So, I have a few questions:

- Is it possible to add a unit test? Is it worth it?
- How does the bug manifest itself? In other words, what can happen if we _don't_ fix this?

Witiko:
> Is it possible to add a unit test? Is it worth it?

It is possible to unit-test this specific bug, which may be useful if someone were to revert the fix in the future. If we wanted to unit-test that *any* matrix produced by the `SparseTermSimilarityMatrix` class constructor is positive definite, we would need to randomly generate a number of matrices, which makes the test non-reproducible and difficult to debug if it fails.

> How does the bug manifest itself? In other words, what can happen if we don't fix this?

The matrix produced by the `SparseTermSimilarityMatrix` is not positive definite, violating the API contract. Although Gensim itself does not depend on the positive definiteness of a matrix at the moment, the user may want to derive word embeddings from the matrix using [Cholesky factorization](https://en.m.wikipedia.org/wiki/Cholesky_decomposition), which will fail for matrices that are not symmetric and positive definite.

mpenkov:
Ok, could you please add a test to this PR?

Witiko:
I have just added the unit test.

PR #2192, opened 2018-09-21
Add `flake8-rst` for docstring code examples

kataev:
Hello, i'm created new tool [flake8-rst](https://pypi.org/project/flake8-rst/).
Idea is simple, crop with regex code snippets and put it in flake8.
PR splitted into logical commits for better review. First commit, add `sourcecode:: pycon`, next fix found issues.

PR #2399, opened 2019-03-02
Fix regression bug in tutorial

piskvorky:
There is a minor bug in the [Corpora and vector spaces](https://radimrehurek.com/gensim/tut1.html): The import of `corpora` got removed by in #2192 [here](https://github.com/RaRe-Technologies/gensim/commit/4543646d3fe3496e11bc935e72cbf9b18504442e#diff-4c6ad40ae122a450e380ca08bda00468L23) (I assume by accident).

Identified by a confused user on our [mailing list](https://groups.google.com/forum/#!topic/gensim/1krhNlwGwDY).

This PR reconstitutes the import, plus fixes the code style since I was at it.

### DIFF
--- a/docs/src/tut1.rst
+++ b/docs/src/tut1.rst
@@ -50,8 +50,10 @@ as well as words that only appear once in the corpus:
     >>>
     >>> # remove common words and tokenize
     >>> stoplist = set('for a of the and to in'.split())
-    >>> texts = [[word for word in document.lower().split() if word not in stoplist]
-    >>>          for document in documents]
+    >>> texts = [
+    >>>     [word for word in document.lower().split() if word not in stoplist]
+    >>>     for document in documents
+    >>> ]
     >>>
     >>> # remove words that appear only once
     >>> frequency = defaultdict(int)
@@ -59,8 +61,10 @@ as well as words that only appear once in the corpus:
     >>>     for token in text:
     >>>         frequency[token] += 1
     >>>
-    >>> texts = [[token for token in text if frequency[token] > 1]
-    >>>          for text in texts]
+    >>> texts = [
+    >>>     [token for token in text if frequency[token] > 1]
+    >>>     for text in texts
+    >>> ]
     >>>
     >>> pprint(texts)
     [['human', 'interface', 'computer'],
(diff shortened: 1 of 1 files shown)


docs/src/tut1.rst
```diff
@@ -50,17 +50,21 @@ as well as words that only appear once in the corpus:
     >>>
     >>> # remove common words and tokenize
     >>> stoplist = set('for a of the and to in'.split())
-    >>> texts = [[word for word in document.lower().split() if word not in stoplist]
-    >>>          for document in documents]
+    >>> texts = [
+    >>>     [word for word in document.lower().split() if word not in stoplist]
```
mpenkov: Do we want to demonstrate the best practice for string tokenization here?

docs/src/tut1.rst
```diff
@@ -50,17 +50,21 @@ as well as words that only appear once in the corpus:
...
+    >>>     for document in documents
+    >>> ]
     >>>
     >>> # remove words that appear only once
     >>> frequency = defaultdict(int)
```
mpenkov: I know this isn't part of the PR, but it may be better to simplify this section as:

```python
# untested code but "should" work
import collections
counter = collections.Counter()
for text in texts:
    counter.update(text)
```

docs/src/tut1.rst
```diff
@@ -207,8 +214,11 @@ Similarly, to construct the dictionary without loading all texts into memory:
     >>> # collect statistics about all tokens
     >>> dictionary = corpora.Dictionary(line.lower().split() for line in open('mycorpus.txt'))
```
mpenkov: Can we use a real corpus here?

That way, we can doctest this file.

docs/src/tut1.rst
piskvorky: The best practice is to use a dedicated library (e.g. NLTK)… do we want to complicate the tutorial?

Maybe we should.

But if we don't, this should be something really stupid (like it is now), so people don't accidentally copy-paste thinking it's a good idea.

docs/src/tut1.rst
piskvorky: Fine with me.

docs/src/tut1.rst
mpenkov: I was thinking more along the lines of https://radimrehurek.com/gensim/utils.html#gensim.utils.tokenize

WDYT?

piskvorky: OK, that's an option (but `utils.simple_preprocess` better),

mpenkov:
There's a ton of improvements we could make to the tutorials, but they're out of scope for this PR. I opened #2424 instead.

PR #2406, opened 2019-03-07
improve implementation of doc2vec's infer_vector

chie8842:
In current doc2vec's infer_vector implementation, it judges same content at `if` statement step by step at L926.
But I think it will be faster when it is judged at first and then train.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -98,7 +98,7 @@ except ImportError:
     # failed... fall back to plain numpy (20-80x slower training than the above)
     FAST_VERSION = -1
 
-    def train_document_dbow(model, doc_words, doctag_indexes, alpha, work=None,
+    def train_document_dbow(model, doc_words, doctag_indexes, alpha, work=None, neu1=None,
                             train_words=False, learn_doctags=True, learn_words=True, learn_hidden=True,
                             word_vectors=None, word_locks=None, doctag_vectors=None, doctag_locks=None):
         """Update distributed bag of words model ("PV-DBOW") by training on a single document.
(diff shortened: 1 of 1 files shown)


gensim/models/doc2vec.py
```diff
@@ -917,27 +917,24 @@ def infer_vector(self, doc_words, alpha=None, min_alpha=None, epochs=None, steps
         doctag_vectors, doctag_locks = self.trainables.get_doctag_trainables(doc_words, self.docvecs.vector_size)
         doctag_indexes = [0]
         work = zeros(self.trainables.layer1_size, dtype=REAL)
+        neu1 = None
         if not self.sg:
             neu1 = matutils.zeros_aligned(self.trainables.layer1_size, dtype=REAL)
 
         alpha_delta = (alpha - min_alpha) / max(epochs - 1, 1)
 
+        if self.sg:
```
mpenkov: Cosmetically, these changes look good, but they break the build. Can you please make the unit tests pass?

gensim/models/doc2vec.py
```diff
@@ -917,27 +917,24 @@ def infer_vector(self, doc_words, alpha=None, min_alpha=None, epochs=None, steps
         doctag_vectors, doctag_locks = self.trainables.get_doctag_trainables(doc_words, self.docvecs.vector_size)
         doctag_indexes = [0]
         work = zeros(self.trainables.layer1_size, dtype=REAL)
+        neu1 = None
```
mpenkov: Why did you add `neu1` keyword parameter? Why are you ignoring its specified value and setting it to None?

gojomo:
I doubt this makes a noticeable performance difference, but could be convinced otherwise with benchmarking. 

While giving all the functions the same signature has a nice symmetry, and enables the use of a function pointer to configure the loop, it also leads to the weirdness @mpenkov notes - adding an unneeded/ignorable parameter to one of the function-signatures. Rather than assigning `None`, it may be more appropriate to just note in a comment that it's intentionally ignored in the `_dbow` case. 

If there's actually a performance win here, moving the loop into the three `if` branches would get the same benefit, albeit with a bit of duplication, without changing function signatures or using a function-pointer. Whether to prefer one approach or the other is a matter of style I don't have a strong opinion about.

mpenkov:
+1 on the benchmarks. IMO if there are no significant improvements, then the changes in this PR are purely cosmetic and not really worth it.

mpenkov:
@chie8842 Are you able to continue this PR? In order to proceed, we need to see benchmark results that illustrate the merit of your changes.

以上よろしくお願いします。

PR #2411, opened 2019-03-10
avoid collisions when decoding bad unicode

### LINKED ISSUE #2402: AssertionError: unexpected number of vectors when loading Korean FB model
Hi,

I downloaded pretrained word vector file (.bin) from facebook 
(https://fasttext.cc/docs/en/crawl-vectors.html)
However, when I tried to use this model it happens to make error.

    fasttext_model = FastText.load_fasttext_format('cc.ko.300.bin', encoding='utf8')

    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xed in position 0: invalid continuation byte

But weird thing is that it operates well when I use old version bin file (https://fasttext.cc/docs/en/pretrained-vectors.html)

So I tried to find the solution and found that this problem had happend and solved.

The issue was made through facebook fasttext issue
(https://github.com/facebookresearch/fastText/issues/715)
And they fixed it by
(https://github.com/facebookresearch/fastText/commit/e13484bcb261cda51d33c4940ab5e207aba3ee79)

So, I think gensim load_fasttext_format function had this UnicodeDecodeError because of above problem.

Can you help me to find and solve this problem?

------------------------------------------

I tried changing
word = word_bytes.decode(encoding, errors='replace')
word = word_bytes.decode(encoding, errors='ignore')
in gensim\models\_fasttext_bin.py , line 177
But both made same error
    File "C:\Users\User\PycharmProjects\ksenticnet\venv\lib\site-packages\gensim\models\keyedvectors.py", line 2207, in init_post_load
    assert vectors.shape[0] == vocab_words + self.bucket, 'unexpected number of vectors'
AssertionError: unexpected number of vectors

#### Versions
Windows-10-10.0.17134-SP0
Python 3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 17:00:18) [MSC v.1900 64 bit (AMD64)]
NumPy 1.16.2
SciPy 1.2.1
gensim 3.7.1
FAST_VERSION 0

mpenkov:
Fix #2402


gensim/models/_fasttext_bin.py
```diff
@@ -160,6 +161,16 @@ def _load_vocab(fin, new_format, encoding='utf-8'):
     """
     vocab_size, nwords, nlabels = _struct_unpack(fin, '@3i')
 
+    #
+    # We must use backslashreplace instead of replace or ignore here, because
+    # we must avoid collisions in the decoded word, e.g.
+    # https://github.com/RaRe-Technologies/gensim/issues/2402
+    #
+    # Unfortunately, backslashreplace is only available on Py3.  On Py2, we
+    # can't really do anything to avoid collisions.
```
piskvorky: Can we calculate the number of collisions, and adjust the `assert` accordingly?  What functionality hinges on the difference between "actual vectors after loading" vs "declared vectors in file before loading"?

mpenkov: I've rewritten the code to avoid collisions.

mpenkov:
@piskvorky Please have a look at this when you have the chance. This is the last PR we need to merge before the next release (sorry I missed it in my earlier email).

gensim/models/_fasttext_bin.py
```diff
@@ -197,6 +194,41 @@ def _load_vocab(fin, new_format, encoding='utf-8'):
...
+    # output is highly likely to be garbage, but its the only thing that:
+    #
+    # - does not cause collisions (see https://github.com/RaRe-Technologies/gensim/issues/2402)
+    # - works under Py2
+    #
+    del encoding
+    word = word_bytes.decode('latin1')
+    logger.error(
+        'failed to decode invalid unicode bytes %r; falling back to latin1, using %r; '
+        'consider upgrading to Python 3 for better Unicode handling (and so much more)',
```
piskvorky: Seems unrelated. The problem is with badly encoded strings in the model (FB's fault), or possibly our choice to accept such inputs instead of failing explicitly.

There's nothing wrong with Python 2, and Python 3 won't fix the problem (just mask it in a different way).

gensim/models/_fasttext_bin.py
```diff
@@ -197,6 +194,41 @@ def _load_vocab(fin, new_format, encoding='utf-8'):
     return raw_vocab, vocab_size, nwords
 
 
+def _decode_bad_unicode_py2(word_bytes, encoding):
```
piskvorky: A docstring would be welcome; magic happening here.

gensim/models/_fasttext_bin.py
```diff
@@ -197,6 +194,41 @@ def _load_vocab(fin, new_format, encoding='utf-8'):
...
+    word = word_bytes.decode('latin1')
+    logger.error(
+        'failed to decode invalid unicode bytes %r; falling back to latin1, using %r; '
+        'consider upgrading to Python 3 for better Unicode handling (and so much more)',
+        word_bytes, word
+    )
+    return word
+
+
+def _decode_bad_unicode_py3(word_bytes, encoding):
```
piskvorky: Docstring please.

gensim/models/_fasttext_bin.py
mpenkov: Is it really necessary? My opinion is no, because:

1. It’s an internal function. The docstring will not appear in the generated documentation. According to PEP8, docstrings are not necessary in this case (https://stackoverflow.com/questions/7687407/are-docstrings-for-internal-functions-python-necessary#7687430).
2. The function body is shorter than what a docstring would be.
3. The docstring would not add any new value: it would merely duplicate the function name and body and existing comment.
4. The function behavior is fully captured by unit tests.

gensim/models/_fasttext_bin.py
mpenkov: Under Py2, you get a string of complete garbage. The term is essentially lost from the vocabulary. Under Py3, you get a mostly OK string with the broken characters escaped. The term remains in the vocabulary, and can contribute vectors to the model. See the unit tests for an example.

So under Py2, our handling of this edge case is worse than under Py3.

mpenkov:
FB seems to operate on raw bytes, so they are able to ignore the problem altogether.

gensim/models/_fasttext_bin.py
piskvorky: Both terms are garbage and lost from the vocabulary, I see no difference. No user query or input text will ever match them, whether it's latin1-garbage or backslash-escaped-garbage.

The only useful property might be *substrings* that still happen to be decoded correctly within that term (probably ~ASCII subset; true both for latin1 and backslash-escaped).

IMO the correct solution would be to "refuse to guess" and either:
1. fail the model loading with an exception, in case valid unicode (via utf8?) is the contract of the FT format, aka "not our problem"; or
2. work with bytes instead of text, in case there's no decoding contract (which I hope is not the case, because that's a lot of work, fixing all the interfaces that expect text!).

Which approach does the Facebook's reference implementation use?

Either way, if we want to go down the road of silently modifying the models, the difference of py2 vs py3 is moot. Anything you can do with unicode in python3, you can also do in python2. We'd just implement `backslashreplace` (looks easy enough) and be consistent, instead of muddying the water by logging arcane py2/3 upgrade recommendations.

gensim/models/_fasttext_bin.py
piskvorky: I'm still +1 on a docstring. Not for the code (which is clean and simple enough, no problem reading that), but for the purpose and motivation behind having this magic in the first place.

If I saw code like this in a new code base, my first reaction would be "WTF, why shouldn't I remove this?"

How about this:

```python
def _decode_bad_unicode_py2(word_bytes, encoding):
    """
    Decode bytes that failed to cleanly decode using utf8, using latin1 instead.

    This is a hack, but it's needed because we're keep *some* representation even of invalid
    (badly encoded) terms. Such invalid terms appear in models produced by other tools, such
    as Facebook's own fastText models (ha!).

    Note the latin1-decoded output is unlikely to be of much use: the characters will be garbage, and
    never actually match with real words. Still, some (ASCII) ngram substrings may be useful, plus
    we avoid term collisions that would otherwise happen with `errors="replace"`, or vocabulary count
    mismatches that would happen if we ignored this invalid term altogether (see also
    https://github.com/RaRe-Technologies/gensim/issues/2402).

    """
    word = word_bytes.decode('latin1')
    logger.error(
        'failed to decode invalid utf8 bytes %r; falling back to latin1, using %r instead',
        word_bytes, word,
    )
    return word
```

gensim/models/_fasttext_bin.py
piskvorky: >  We'd just implement backslashreplace (looks relatively easy) 

[Googling](https://stackoverflow.com/questions/42860186/exact-equivalent-of-b-decodeutf-8-backslashreplace-in-python-2) turned out this, as a possible `backslashreplace` backport:

```python
def backslashescape_backport(ex):
    # The error handler receives the UnicodeDecodeError, which contains arguments of the
    # string and start/end indexes of the bad portion.
    bstr, start, end = ex.object, ex.start, ex.end

    # The return value is a tuple of Unicode string and the index to continue conversion.
    # Note: iterating byte strings returns int on 3.x but str on 2.x
    return u''.join('\\x{:02x}'.format(c if isinstance(c, int) else ord(c)) for c in bstr[start:end]), end

codecs.register_error("backslashescape_backport", backslashescape_backport)
bad_term = u"Žluťoučký koníček".encode("latin2")  # create invalid UTF8
print(repr(bad_term.decode("utf-8", "backslashescape_backport")))
u'\\xaelu\\xbbou\\xe8k\\xfd kon\\xed\\xe8ek'
```

gensim/models/_fasttext_bin.py
mpenkov: I ended up reverting this in favor of the backport.

gensim/models/_fasttext_bin.py
mpenkov: Reverted in favor of backslashreplace backport.

PR #2417, opened 2019-03-15
Add nmslib indexer

masa3141:
Hi, I added [nmslib](https://github.com/nmslib/nmslib) indexer.

Some research shows nmslib is better than annoy indexer.
https://erikbern.com/2018/06/17/new-approximate-nearest-neighbor-benchmarks.html
https://www.benfrederickson.com/approximate-nearest-neighbours-for-recommender-systems/

This is the first time to contribute to gensim. If I miss something, please let me know.


piskvorky:
Thanks @masa3141 !

Ping @searchivarius – would you have capacity to review this? (choice of NMSLIB APIs, parameters, etc). Cheers.

searchivarius:
@piskvorky yes, will do soon. thanks letting me know!

masa3141:
Ping @searchivarius 
Hi could you review the code? Thanks!

searchivarius:
Please, bear with me. There was recently reported what seems to be a serious bug. I need to fix it first.

masa3141:
I see. Thank you!

masa3141:
@searchivarius 
Sorry to bother you again. 
Could you review the code? 
Thanks!

searchivarius:
@masa3141 sorry for the delay, I still haven't resolved the issue with the recently reported NMSLIB bug. In fact, knnQuery might not work, but I don't know yet why, knnQueryBatch seems to be working fine.
BTW, do I get it correctly that you use **only the cosine distance**?

masa3141:
Thank you for your support in your busy time. Please review at your convenience. 
Yes, I only use cosine distance because annoy indexer in gensim only supports cosine distance.

searchivarius:
Ok, I have started working on this. Should be ready soon!

searchivarius:
Hi @masa3141 there seems to be, indeed, an issue with knnQuery. Could you slightly change your code so it uses knnQueryBatch instead? I am going to fix knnQuery very soon, but it will take time and the older NMSLIB versions will still be broken.

Please, make query matrix 2-dim numpy, example is here:
https://github.com/nmslib/nmslib/blob/master/python_bindings/notebooks/search_vector_dense_optim.ipynb

Thank you!

masa3141:
Hi @searchivarius, thank you for the review!
I changed the code to use knnQueryBatch instead of knnQuery.
Thanks!

masa3141:
Did the test fail due to Memory Error?

searchivarius:
@masa3141 no it's related to the using wrong memory layout for numpy array data. I am quite surprised it wasn't caught before.

masa3141:
How do I fix it? Does it happen only in python2.7?

searchivarius:
@masa3141 ohhh I missed that you changed to knnQueryBatch already. I will review shortly.

searchivarius:
Hi @masa3141 and @piskvorky I have reviewed the querying part. It does look fine to me, many thanks! It would be nice to double check how all is working though when things are merged.

masa3141:
@searchivarius Thank you for the review!
@piskvorky It seems to fail to install nmslib in ci container in Python2.7. How do I fix this?

piskvorky:
No idea. CC @mpenkov .

masa3141:
I changed to install nmslib into CI only when python version is over 3.0.
It seems to work.
(The test on python2.7 windows is failing. This failure happens in other pull requests.)

masa3141:
@piskvorky @mpenkov 
Hi, do I need to do anything?
Thanks.

mpenkov:
@masa3141 Not in the immediate future. Let us work out why Appveyor is failing. I'll get back to you once that's done.

masa3141:
@mpenkov Thanks for fixing Appveyor issue! When will this feature be merged?

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,217 @@
+#!/usr/bin/env python
```
mpenkov: Why do you need this? I think this module does not need to be executable.

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,217 @@
...
+            >>> _, temp_fn = mkstemp()
+            >>> indexer.save(temp_fn)
+            >>>
+            >>> new_indexer = NmslibIndexer()
+            >>> new_indexer.load(temp_fn)
+            >>> new_indexer.model = model
+
+        """
+        fname_dict = fname + '.d'
+        if not (os.path.exists(fname) and os.path.exists(fname_dict)):
```
mpenkov: Why are you checking local paths only?

We're using smart_open, it can open things over the network, etc. I think instead of explicitly checking that the files exist, you should just open them with smart_open and see if it works. If it doesn't work, then it's the user's fault for providing bad fname.

https://stackoverflow.com/questions/12265451/ask-forgiveness-not-permission-explain

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,217 @@
...
+How it works
+------------
+Searching in generic non-metric space.
+
+More information about Nmslib: `github repository <https://github.com/nmslib/nmslib>`_.
+
+"""
+import os
+
+from smart_open import smart_open
```
mpenkov: Please use smart_open's open function instead.

See https://github.com/RaRe-Technologies/smart_open#migrating-to-the-new-open-function

masa3141: Thanks. Changed to use smart_open's open function instead.

gensim/similarities/nmslib.py
masa3141: I see. Changed the code.

gensim/similarities/nmslib.py
masa3141: Deleted the code. (I referred to index.py that implemented annoy indexer. index.py has this line.)

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+            >>> indexer.save(temp_fn)
+            >>>
+            >>> new_indexer = NmslibIndexer()
+            >>> new_indexer.load(temp_fn)
+            >>> new_indexer.model = model
+
+        """
+        fname_dict = fname + '.d'
+        with open(fname_dict, 'rb') as f:
+            d = _pickle.loads(f.read())
```
mpenkov: Doesn't pickle support reading directly from files?

```suggestion
            d = _pickle.load(f)
```

masa3141: Thanks. Changed to use pickle.load instead of pickle.loads.

mpenkov:
@masa3141 Thank you for your changes. I think this feature is ready to merge.

@piskvorky How do you want to handle documentation for this feature? I think right now is the best opportunity to add some docs describing how to use the new functionality. The PR author (and to some extent, myself) have this PR fresh in memory, so we could proceed without a context switch.

Yes, we could "do it later", but we risk letting it slip, and including new functionality without telling people how to use it.

In summary, do you think we should handle documentation (e.g. a tutorial or howto guide) as part of this PR? Or should we deal with it separately?

piskvorky:
Absolutely, no new functionality merged without (top-level) motivational docs.

We can always add more description to API params, correct typos and whatnot, but if the core workflow and "why" is unclear, we won't even know what to describe. Plus users will never use it if the context / motivation is buried.

Where the motivation docs should live is another question. Maybe in the new docs structure you're fleshing out now? Or as a Jupyter notebook that you'll covert later? WDYT?

masa3141:
@mpenkov @piskvorky 
Thanks. I'm thinking to add a notebook "nmslibtutorial.ipynb" under docs/notebooks/, similar to annoytutorial.ipynb. 
Is this enough? Or should I add other docs?

mpenkov:
Yes. Please give me a few days to author documentation requirements (what we expect, best practices, etc)

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+    :class:`~gensim.models.fasttext.FastText` and :class:`~gensim.models.keyedvectors.Word2VecKeyedVectors` classes.
+
+    """
+
+    def __init__(self, model=None, index_params=None, query_time_params=None):
+        """
+        Parameters
+        ----------
+        model : :class:`~gensim.models.base_any2vec.BaseWordEmbeddingsModel`, optional
+            Model, that will be used as source for index.
```
mpenkov: What happens if model is None? It may be worth including an example showing this use case, if it is valid.

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+        index_params : dict, optional
+            index_params for Nmslib indexer.
+        query_time_params : dict, optional
+            query_time_params for Nmslib indexer.
+
+        Examples
```
mpenkov: This example really belongs in the class docstring (or even better, the top-level module docstring) instead of the initializer method docstring because it demonstrates more than just the initializer method.

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+
+        if model:
+            if isinstance(self.model, Doc2Vec):
+                self.build_from_doc2vec()
+            elif isinstance(self.model, (Word2Vec, FastText)):
+                self.build_from_word2vec()
+            elif isinstance(self.model, (WordEmbeddingsKeyedVectors, KeyedVectors)):
+                self.build_from_keyedvectors()
+            else:
+                raise ValueError("Only a Word2Vec, Doc2Vec, FastText or KeyedVectors instance can be used")
```
mpenkov: Avoid use of passive language where possible

```suggestion
                raise ValueError("model must be a Word2Vec, Doc2Vec, FastText or KeyedVectors instance")
```

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def save(self, fname, protocol=2):
+        """Save NmslibIndexer instance.
```
mpenkov:
```suggestion
        """Save this NmslibIndexer instance to a file.
```

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+
+        Parameters
+        ----------
+        fname : str
+            Path to output file,
```
mpenkov:
```suggestion
            Path to the output file,
```

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+            will produce 2 files: `fname` - parameters and `fname`.d - :class:`~nmslib.NmslibIndex`.
+        protocol : int, optional
+            Protocol for pickle.
+
+        Notes
+        -----
+        This method save **only** index (**model isn't preserved**).
```
mpenkov: I think it'd be more helpful to the user to mention:

```suggestion
        This method saves **only** the index (**the model isn't preserved**).
        If you want to save the model, do that explicitly by calling the model.save method.
```

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+
+        """
+        fname_dict = fname + '.d'
+        self.index.saveIndex(fname)
+        d = {'index_params': self.index_params, 'query_time_params': self.query_time_params, 'labels': self.labels}
+        with open(fname_dict, 'wb') as fout:
+            _pickle.dump(d, fout, protocol=protocol)
+
+    def load(self, fname):
+        """Load NmslibIndexer instance
```
mpenkov:
```suggestion
        """Load a NmslibIndexer instance from a file
```

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+
+        Parameters
+        ----------
+        fname : str
+            Path to dump with NmslibIndexer.
+
+        Examples
```
mpenkov: I would put this example in the module docstring, because it covers more than just one method (load and save).

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+        This method save **only** index (**model isn't preserved**).
+
+        """
+        fname_dict = fname + '.d'
+        self.index.saveIndex(fname)
+        d = {'index_params': self.index_params, 'query_time_params': self.query_time_params, 'labels': self.labels}
+        with open(fname_dict, 'wb') as fout:
+            _pickle.dump(d, fout, protocol=protocol)
+
+    def load(self, fname):
```
mpenkov: Gensim follows the pattern of using class method for loading, e.g. 

```python
model = FastText.load('/path/to/file')
```

instead of

```
model = FastText()
model.load('/path/to/file')
```

Please adjust your load function and its documentation to be a class method.

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,210 @@
...
+        fname_dict = fname + '.d'
+        with open(fname_dict, 'rb') as f:
+            d = _pickle.load(f)
+        self.index_params = d['index_params']
+        self.query_time_params = d['query_time_params']
+        self.index = nmslib.init()
+        self.index.loadIndex(fname)
+        self.labels = d['labels']
+
+    def build_from_word2vec(self):
```
mpenkov: Please make this method internal (same for other build methods) - it appears that users of this class won't be calling this method directly.

Internal methods start with an underscore (e.g. _build_from_word2vec).

mpenkov:
OK, I authored a short guide on how to write documentation: https://github.com/RaRe-Technologies/gensim/wiki/How-to-author-documentation

Please have a read through and get back to me about the kind of documentation we can write for this new module. I'll give you some ideas below.

First, have a look at the module docstring. I can see some potential improvements:

- Mention what the main entry points to the module are (classes, functions). In this case, it looks like the NmslibIndexer class is the main (and only) one. 
- The intro mentions which classes your feature integrates with. This is helpful, but a user would ask: **how** do I integrate this feature with those classes? This kind of information would be helpful at the top level (I can see you have an example in the docstring initializer method of the NmslibIndexer class, but that's a bit too far to look).
- The "how it works" section is not very helpful, because it doesn't actually explain how nmslib works. You don't **have** to explain how the library works (linking to it is sufficient) but in this case the heading is misleading.
- Style: the docstring is very short. You don't really have to include headings: write simple prose.

Next, consider writing a tutorial or how-to guide for your module. One question that I have is: how is the NmslibIndexer better than its alternatives? Is it possible to demonstrate a measurable benefit using a concrete example?

mpenkov:
Also, while re-reading your docstring, I found some minor points and added suggestions. Normally, I'd commit these suggestions myself (they are minor and not worth bothering you as the author), but I can't do it in your case, because of lack of permissions. Next time you make a PR, please consider making your branch accept commits from reviewers (it's a github option).

gensim/similarities/nmslib.py
masa3141: If the model is None, index and labels are not initialized. In that case, a user should load or init the index and labels by themselves. I add this information to doc string.
Also this is used by load function with model=None.

gensim/similarities/nmslib.py
masa3141: Thanks. I moved this to the module docstring.

gensim/similarities/nmslib.py
masa3141: Thanks. Changed.

gensim/similarities/nmslib.py
masa3141: Thanks. Changed.

gensim/similarities/nmslib.py
masa3141: Thanks. Changed.

gensim/similarities/nmslib.py
masa3141: Thanks. Changed.

gensim/similarities/nmslib.py
masa3141: Thanks. Changed.

gensim/similarities/nmslib.py
masa3141: Thanks. I moved this to the module docstring.

gensim/similarities/nmslib.py
masa3141: I changed this to use the class method.
BTW, annoy indexer doesn't use the class method for loading. So should we fix this in the future?

gensim/similarities/nmslib.py
masa3141: I changed this.

masa3141:
@mpenkov 
Thanks for reviewing and feedbacks. I fixed the code. 

I added a jupyter notebook for a tutorial. I followed annoytutorial.ipynb. (I copied some sentences from the notebook. If it is not good, please let me know. I will rewrite the document in the notebook.)
In the notebook, I compared the annoy and nmslib indexer. This shows nmslib is faster and more accurate.

I checked "Allow edits from maintainers.", which says "If checked, users with write access to RaRe-Technologies/gensim can add new commits to your feature/nmslib branch. You can always change this setting later."
So please check this? And if there are minor points, could you push those to this branch?

gensim/similarities/nmslib.py
```diff
@@ -58,28 +95,16 @@ def __init__(self, model=None, index_params=None, query_time_params=None):
         ----------
         model : :class:`~gensim.models.base_any2vec.BaseWordEmbeddingsModel`, optional
             Model, that will be used as source for index.
+            If the model is None, index and labels are not initialized.
+            In that case please load or init the index and labels by yourself.
```
mpenkov: This is better than before, but still not very helpful. The main questions: How is the user supposed to load/init the index? Which functions to call?

Next, does it make sense to pass model=None from an API perspective? Does it allow the user to do something new? If it just means the user has to do more work using other functions, and does not have any advantages, then I think we should force them to pass a non-None model, because this aligns with the Zen of Python: there should only be one obvious way to do things.

gensim/similarities/nmslib.py
```diff
@@ -10,20 +10,57 @@
 This module contains integration Nmslib with :class:`~gensim.models.word2vec.Word2Vec`,
 :class:`~gensim.models.doc2vec.Doc2Vec`, :class:`~gensim.models.fasttext.FastText` and
 :class:`~gensim.models.keyedvectors.KeyedVectors`.
-
+An instance of NmslibIndexer needs to be created in order to use Nmslib in gensim.
```
mpenkov: Please avoid passive language where possible. It's a poor choice in English (personally, I avoid it in general regardless of language), because it's verbose and ambiguous. Consider the alternative:

> To use nmslib, instantiate a :class:`~gensim.similarities.nmslib.NmslibIndexer` class and pass the instance as the ``indexer`` parameter to your model's ``most_similar`` method (e.g. :py:func:`~gensim.models.doc2vec.most_similar`).

So how it effectively says the same thing, but directly, tersely, and unambiguously?

searchivarius:
@mpenkov in all fairness, there is usually nothing wrong about passive voice. One doesn't have to avoid passive voice at all costs. Technical language has a lot of passive voice.
https://writingcenter.unc.edu/tips-and-tools/passive-voice/

mpenkov:
> in all fairness, there is usually nothing wrong about passive voice.

Do you really think "An instance of NmslibIndexer needs to be created in order to use Nmslib in gensim." is a well-written English sentence?

> One doesn't have to avoid passive voice at all costs.

I completely agree with you there. On the other hand, one doesn't have to go out of one's way to use passive voice where it's unnecessary (see example above).

> Technical language has a lot of passive voice.

Yes, and a lot of technical language is written poorly.

searchivarius:
@mpenkov I don't think it's ungrammatical. It's a bit terse. Maybe I wouldn't use passive voice in this specific sentence, b/c we mention a concept for the first time.

mpenkov:
> An instance of NmslibIndexer needs to be created in order to use Nmslib in gensim.

While the above sentence is grammatically correct, it's still poorly written. It's not "terse", either: it grabs the reader's attention for over a dozen mostly needless words to convey little useful information. I believe the technical term to describe this style of writing is "drivel".

If you still disagree, then I think it's best we agree to disagree on this.

gensim/similarities/nmslib.py
mpenkov: Sure, but let's do that in a separate PR.

gensim/similarities/nmslib.py
mpenkov: I just realized I can push to your branch, so I've applied the above change myself.

masa3141: Thanks for your change!

gensim/similarities/nmslib.py
masa3141: In the load function, we call this init function with model=None.
```
    @classmethod
    def load(cls, fname):
        """Load a NmslibIndexer instance from a file

        Parameters
        ----------
        fname : str
            Path to dump with NmslibIndexer.

        """
        fname_dict = fname + '.d'
        with open(fname_dict, 'rb') as f:
            d = _pickle.load(f)
        index_params = d['index_params']
        query_time_params = d['query_time_params']
        nmslib_instance = cls(index_params=index_params, query_time_params=query_time_params)
        index = nmslib.init()
        index.loadIndex(fname)
        nmslib_instance.index = index
        nmslib_instance.labels = d['labels']
        return nmslib_instance
```
Do you think this init function should not allow model=None? In that case, this load function should have a model argument? (i.e, load(cls, fname, model))

mpenkov: OK, so we have two separate use cases for the initializer.

1. Construct a new class, passing a model (called by user)
2. Construct a new class, don't pass a model (load function, internal use case)

So it sounds like for the user, they will **always** have to pass a model. So the model should not be a keyword parameter. Is that correct? If so, let's update the initializer parameters and docstring to reflect that. Next, to support the second use case, we should make the load function pass None as the model explicitly.

Next, I think the second use case above isn't clearly documented. Look at the save/load example:

```
    >>> indexer = NmslibIndexer(model)
    >>> _, temp_fn = mkstemp()
    >>> indexer.save(temp_fn)
    >>>
    >>> new_indexer = NmslibIndexer.load(temp_fn)
    >>> model.most_similar("cat", topn=2, indexer=new_indexer)
    [('cat', 1.0), ('meow', 0.5595494508743286)]
```

When `new_indexer` is created, it knows nothing about the model. The docstring for the initializer says this:

```
            If the model is None, index and labels are not initialized.
            In that case please load or init the index and labels by yourself.
```

Why is the code not loading/initializing index and labels?

masa3141: I see. I will change the initializer's parameters and docstring so that the model should not be a keyword parameter. And I will make the load function pass None as the model explicitly.

In the load function, we create an indexer with model=None. After that, we assign index and labels.
```
        nmslib_instance = cls(index_params=index_params, query_time_params=query_time_params)
        index = nmslib.init()
        index.loadIndex(fname)
        nmslib_instance.index = index
        nmslib_instance.labels = d['labels']
        return nmslib_instance
```
So when we use load function to create a indexer, the indexer has index and labels.

mpenkov: Yes. We should also move these lines:

```
            If the model is None, index and labels are not initialized.
            In that case please load or init the index and labels by yourself.
```

from the docstring to somehere around [here](https://github.com/RaRe-Technologies/gensim/blob/7e2d07fcf13f1162b5157a331ee5d385419f7c13/gensim/similarities/nmslib.py#L117). This sort of information is purely an implementation detail, so should not be visible to users.

gensim/similarities/nmslib.py
```diff
@@ -0,0 +1,225 @@
...
+        index.loadIndex(fname)
+        nmslib_instance.index = index
+        nmslib_instance.labels = d['labels']
+        return nmslib_instance
+
+    def _build_from_word2vec(self):
+        """Build an Nmslib index using word vectors from a Word2Vec model."""
+
+        self.model.init_sims()
+        return self._build_from_model(self.model.wv.vectors_norm, self.model.wv.index2word)
```
mpenkov: It doesn't look like you're using this return value anywhere (same for the other build functions). Furthermore, it looks like this value will always be None.

I think it'd be clearer if you didn't return anything.

gensim/similarities/nmslib.py
masa3141: OK, thanks. I will move it.

gensim/similarities/nmslib.py
masa3141: I see. I will delete "return".

gensim/similarities/nmslib.py
masa3141: Fixed those.

gensim/similarities/nmslib.py
masa3141: Deleted "return".

masa3141:
@mpenkov 
py27-win test fails. Do you know the reason and solution?
Thanks.

mpenkov:
Hmm, looks like a dependency problem:

```
ERROR:   py27-win: could not install deps [numpy==1.11.3, scipy==0.18.1, .[test-win]]; v = InvocationError(u"'C:\\projects\\gensim-431bq\\.tox\\py27-win\\Scripts\\pip.EXE' install --timeout=60 --trusted-host 28daf2247a33ed269873-7b1aad3fab3cc330e1fd9d109892382a.r6.cf2.rackcdn.com --find-links http://28daf2247a33ed269873-7b1aad3fab3cc330e1fd9d109892382a.r6.cf2.rackcdn.com/ numpy==1.11.3 scipy==0.18.1 '.[test-win]'", 2)
```

masa3141:
Interesting. It didn't happen before. From today this happens. 
I have no idea to fix this. Could you look at this problem? I think this will happen in other pull requests.

mpenkov:
OK, leave it with me. Once I fix this, I will ping you.

gensim/test/test_similarities.py
```diff
@@ -696,6 +696,156 @@ def testSaveLoad(self):
...
+    def setUp(self):
+        try:
+            import nmslib  # noqa:F401
+        except ImportError:
+            raise unittest.SkipTest("Nmslib library is not available")
+
+        from gensim.similarities.nmslib import NmslibIndexer
+        self.indexer = NmslibIndexer
+
+    def testWord2Vec(self):
```
mpenkov:
```suggestion
    def test_word2vec(self):
```

gensim/test/test_similarities.py
```diff
@@ -696,6 +696,156 @@ def testSaveLoad(self):
...
+        model = word2vec.Word2Vec(texts, min_count=1)
+        model.init_sims()
+        index = self.indexer(model)
+
+        self.assertVectorIsSimilarToItself(model.wv, index)
+        self.assertApproxNeighborsMatchExact(model, model.wv, index)
+        self.assertIndexSaved(index)
+        self.assertLoadedIndexEqual(index, model)
+
+    def testFastText(self):
```
mpenkov:
```suggestion
    def test_fasttext(self):
```

gensim/test/test_similarities.py
```diff
@@ -696,6 +696,156 @@ def testSaveLoad(self):
...
+        model = FastText(LeeReader(datapath('lee.cor')))
+        model.init_sims()
+        index = self.indexer(model)
+
+        self.assertVectorIsSimilarToItself(model.wv, index)
+        self.assertApproxNeighborsMatchExact(model, model.wv, index)
+        self.assertIndexSaved(index)
+        self.assertLoadedIndexEqual(index, model)
+
+    def testNmslibIndexingOfKeyedVectors(self):
```
mpenkov:
```suggestion
    def test_indexing_keyedvectors(self):
```

gensim/test/test_similarities.py
```diff
@@ -696,6 +696,156 @@ def testSaveLoad(self):
...
+        from gensim.similarities.nmslib import NmslibIndexer
+        keyVectors_file = datapath('lee_fasttext.vec')
+        model = KeyedVectors.load_word2vec_format(keyVectors_file)
+        index = NmslibIndexer(model)
+
+        self.assertVectorIsSimilarToItself(model, index)
+        self.assertApproxNeighborsMatchExact(model, model, index)
+
+    def testLoadMissingRaisesError(self):
```
mpenkov:
```suggestion
    def test_load_missing_raises_error(self):
```

masa3141:
I found there are many camel cases in test method names. (Ex, test_nmf.py, test_sklearn_api.py)
Should we change to use underscores in test_similarities.py?

mpenkov:
Yes, we should, but those files are out of scope for this PR.

Let's handle the nmslib test cases only for now, and leave the rest for another PR.

masa3141:
I see. I changed to use underscores.

mpenkov:
Opened a new ticket with support: https://help.appveyor.com/discussions/problems/24208-unable-to-use-the-latest-version-of-pip-for-my-tox-builds

masa3141:
@mpenkov 
Is there anything I should do?

mpenkov:
@masa3141 Finally merged. Thank you for your contribution and your patience!

masa3141:
Thanks a lot!

searchivarius:
Hurray, thanks everybody!

masa3141:
@mpenkov Hi, 3.8.0 release doesn't contain this nmslib feature? The release note doesn't mention this feature. Thanks

mpenkov:
@masa3141 Thank you for pointing it out. Looks like our changelog generation script missed the nmslib PR. The indexer itself is definitely there: https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/similarities/nmslib.py

I will update the change log.

PR #2419, opened 2019-03-16
Make matutils.unitvec always return float norm when requested

Witiko:
Contrary to the documentation, `matutils.unitvec(…, return_norm=True)` function does not always return a float norm:

1. Zero vectors stored as numpy arrays will return an integer norm.
2. Zero vectors stored as lists of tuples (the Gensim BOW format) will return no norm.

This PR applies minimum changes to fix these two issues.

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -733,7 +733,7 @@ def unitvec(vec, norm='l2', return_norm=False):
         if norm == 'l1':
             veclen = np.sum(np.abs(vec))
         if norm == 'l2':
-            veclen = blas_nrm2(vec)
+            veclen = blas_nrm2(vec) if len(vec) else 0.0
         if veclen > 0.0:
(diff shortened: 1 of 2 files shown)


gensim/matutils.py
```diff
@@ -725,15 +725,15 @@ def unitvec(vec, norm='l2', return_norm=False):
...
+                return vec, 1.0
             else:
                 return vec
 
     if isinstance(vec, np.ndarray):
         if norm == 'l1':
             veclen = np.sum(np.abs(vec))
         if norm == 'l2':
-            veclen = blas_nrm2(vec)
+            veclen = blas_nrm2(vec) if len(vec) else 0.0
```
piskvorky: `len` on numpy arrays is ambiguous… better use `shape`?

Witiko: I replaced `if len(vec)` with `if 0 in vec.shape`.

PR #2420, opened 2019-03-17
Complete the implementation of SMART

Witiko:
This PR continues #1791 by completing the implementation of Salton's [SMART Information Retrieval System][smart] in `models.tfidf`, and follows up on the [discussion in the Gensim Google group][topic] from February. See the list of changes:

 [smart]: https://en.wikipedia.org/wiki/SMART_Information_Retrieval_System (SMART Information Retrieval System)
 [topic]: https://groups.google.com/forum/#!searchin/gensim/tf-idf|sort:date/gensim/15OMa3-1fdE/y9Dx9yKtGwAJ

- Make `t` an alias for the `n` term frequency method.
- Make `x` an alias for the `n` document frequency method.
- Implement the `t` document frequency method, and rename the existing `t` document frequency method to `f`.
- Make `x` an alias for the `n` document length normalization method.
- Implement the `u` and `b` pivoted document length normalization methods.
- Produce a helpful error message when a SMART scheme in the `ddd.qqq` format is requested:

```python
>>> from gensim.models import TfidfModel
>>> TfidfModel(smartirs='Lnu.nnn')
ValueError: The notation Lnu.nnn specifies two term-weighting schemes,
            one for collection documents (Lnu) and one for queries (nnn).
            You must train two separate tf-idf models.
```


Witiko:
Pivoted characted length normalization (i.e. the `b` document length normalization method) was called the state-of-the-art document length normalization method by [Salton in 2001][salton01] (see Table 1). I have not implement pivoted characted length normalization in this PR (yet) due to the technical difficulties outlined below.

 [salton01]: http://singhal.info/ieee2001.pdf (Modern information retrieval: A brief overview)

Pivoted characted length normalization requires word lengths and collection frequencies. The `corpora.dictionary.Dictionary` only stores document frequencies in the `dfs` attribute (number of documents containing a word), not collection frequencies (total number of word occurences). Implementing pivoted characted length normalization would require storing collection frequencies in `Dictionary`, making `models.tfidf.smartirs_normalize` a method of `models.tfidf.TfidfModel`, and disabling the pivoted characted length normalization when `TfidfModel` is instantiated from a corpus rather than a dictionary, since a corpus contains no information about the original words and their lengths.

gensim/corpora/dictionary.py
```diff
@@ -36,6 +36,8 @@ class Dictionary(utils.SaveLoad, Mapping):
         token -> tokenId.
     id2token : dict of (int, str)
         Reverse mapping for token2id, initialized in a lazy manner to save memory (not created until needed).
+    cfs : dict of (int, int)
```
piskvorky: What are the memory implications of this change?

Witiko: Instead of storing V integers in the `dfs` attribute, where V is the vocabulary size, we are now storing 2V integers in the `dfs` and `cfs` attributes. [Empirically](https://en.wikipedia.org/wiki/Heaps%27_law), V is the square root of the corpus size.

piskvorky: How much is this extra `dict` (in MBs of RAM), for let's say 100k vocab?

piskvorky: Actually, never mind. I thought it's storing string keys, but it's actually storing ints, I'm blind. Probably no issue then 👍

Witiko:
I am almost done implementing the `b` document length normalization method, but it seems that an interface would need to change. Specifically, `Tfidf` accepts the `normalize` parameter, which is either a boolean or a callable that receives a locally and globally weighted document vector as its only positional argument. Character length pivoted normalization requires the unweighted (bag of words) document vector as well, so that the character length of a document can be determined. However, calling `normalize` with an additional positional parameter breaks an established interface.

Witiko:
This PR should be ready for a review.

markroxor:
@Witiko thanks for the changes, I wanted to add them since a long time. :)

gensim/matutils.py
```diff
@@ -715,6 +715,8 @@ def unitvec(vec, norm='l2', return_norm=False):
             veclen = np.sum(np.abs(vec.data))
         if norm == 'l2':
             veclen = np.sqrt(np.sum(vec.data ** 2))
+        if norm == 'unique':
```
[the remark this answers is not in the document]
Witiko: I agree. However, lines 714–718 are like this, so the suggested change is inconsistent, unless I rewrite more of the existing code.

gensim/matutils.py
```diff
@@ -734,6 +736,8 @@ def unitvec(vec, norm='l2', return_norm=False):
             veclen = np.sum(np.abs(vec))
         if norm == 'l2':
             veclen = blas_nrm2(vec)
+        if norm == 'unique':
```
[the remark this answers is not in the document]
Witiko: See my above comment.

gensim/matutils.py
```diff
@@ -757,6 +761,8 @@ def unitvec(vec, norm='l2', return_norm=False):
             length = float(sum(abs(val) for _, val in vec))
         if norm == 'l2':
             length = 1.0 * math.sqrt(sum(val ** 2 for _, val in vec))
+        if norm == 'unique':
```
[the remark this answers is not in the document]
Witiko: See my above comment.

gensim/matutils.py
```diff
@@ -706,7 +706,7 @@ def unitvec(vec, norm='l2', return_norm=False):
     Zero-vector will be unchanged.
 
     """
-    if norm not in ('l1', 'l2'):
+    if norm not in ('l1', 'l2', 'unique'):
```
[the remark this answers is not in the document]
Witiko: The use of `%r` is inconsistent with the rest of the module. I added a more conservative implementation of your suggestion in ee0ed45d.

gensim/models/tfidfmodel.py
```diff
@@ -39,24 +41,28 @@ def resolve_weights(smartirs):
...
     Returns
     -------
-    3-tuple (local_letter, global_letter, normalization_letter)
+    str of (local_letter, global_letter, normalization_letter) or None
 
     local_letter : str
         Term frequency weighing, one of:
-            * `n` - natural,
-            * `l` - logarithm,
+            * `b` - binary,
```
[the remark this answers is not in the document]
Witiko: In general, I would agree. However, these letters are standard, see [SMART Information Retrieval System](https://en.wikipedia.org/wiki/SMART_Information_Retrieval_System).

gensim/models/tfidfmodel.py
```diff
@@ -310,21 +345,16 @@ def __init__(self, corpus=None, id2word=None, dictionary=None, wlocal=utils.iden
...
         self.wlocal, self.wglobal, self.normalize = wlocal, wglobal, normalize
         self.num_docs, self.num_nnz, self.idfs = None, None, None
-        self.smartirs = smartirs
+        self.smartirs = resolve_weights(smartirs)
         self.slope = slope
         self.pivot = pivot
         self.eps = 1e-12
 
-        # If smartirs is not None, override wlocal, wglobal and normalize
+        # If smartirs is not None, override wlocal and wglobal
```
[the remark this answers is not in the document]
Witiko: Agreed. I removed the comment in 7f4e9da2.

gensim/models/tfidfmodel.py
```diff
@@ -346,6 +378,28 @@ def __init__(self, corpus=None, id2word=None, dictionary=None, wlocal=utils.iden
             # be initialized in some other way
             pass
 
+        # If smartirs is not None, override pivot and normalize
+        if smartirs is not None:
+            if self.pivot is None:
+                if n_n == "u":
+                    if dictionary is not None or corpus is not None:
+                        if callable(self.normalize):
+                            logger.warning("constructor received smartirs; ignoring normalize")
```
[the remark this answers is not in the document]
Witiko: These warnings inform the user about the interplay between the named arguments. The `self.normalize` callable is only used (or ignored) in `__getitem__`, which is too late to emit a warning.

gensim/models/tfidfmodel.py
```diff
@@ -346,6 +378,28 @@ def __init__(self, corpus=None, id2word=None, dictionary=None, wlocal=utils.iden
             # be initialized in some other way
             pass
 
+        # If smartirs is not None, override pivot and normalize
+        if smartirs is not None:
```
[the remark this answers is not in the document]
Witiko: I implemented your suggestions in 623d785e and 69208675.

Witiko:
As discussed in #2420, I improved the documentation and added a unit test for `wlocal` in 18d30cb1d0edde517ddedcba434ac35bd5e05ddf.

gensim/models/tfidfmodel.py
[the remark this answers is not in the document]
mpenkov: I think the link to the Wiki page from our documentation would be helpful, but I still think replacing the magic strings with symbolic constants is worth it. My main motivation is to decrease the cognitive load on the reader and maintainer of the code. Even if the current magic strings are standard, you're expecting the reader/maintainer to remember what each one of them means in order to understand the code. This may not seem like much if you're very familiar with SMART, but it adds an unnecessary barrier to less enlightened people like myself.

Is there a compelling reason not to add symbolic constants? If you're worried about polluting the namespace, you could prefix each of them with `SMART_` to make it obvious what they are.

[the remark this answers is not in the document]
Witiko: > I think the link to the Wiki page from our documentation would be helpful

It is already part of the documentation.

> Is there a compelling reason not to add symbolic constants?

I see several reasons for opposing the addition of symbolic constants:

- The individual letters are clearly documented in the class docstring. No information is added by duplicating this information in the suggested form.
- Succint expressions such as `w_tf not in 'btnaldL'` become less readable in the suggested form.
- PR #1791 introduced the SMART letters. The PR was reviewed by two senior maintainers and found readable.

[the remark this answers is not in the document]
Witiko: > I think the link to the Wiki page from our documentation would be helpful

It is already part of the documentation.

> Is there a compelling reason not to add symbolic constants?

I see several reasons for opposing the addition of symbolic constants:

- The individual letters are clearly documented in the class docstring. No information is added by duplicating this information.
- Succint expressions such as `w_tf not in 'btnaldL'` become less readable in the suggested form.
- PR #1791 introduced the SMART letters. The PR was reviewed by two senior maintainers (Cc @piskvorky, @menshikh-iv) and found readable.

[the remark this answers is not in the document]
mpenkov: @piskvorky and @menshikh-iv What do you think?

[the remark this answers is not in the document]
piskvorky: I'm OK with direct string-like representation, as long as it's contained to the SMART module/functionality, and doesn't "magically" appear in other documentation / workflows.

That is, users who work explicitly with SMART can be expected to know the SMART nomenclature. But we shouldn't bother anyone else with it (users of plain TfIdf, Dictionary etc). As a user, I don't want to be studying some arcane lettering system just to use standard TFIDF.

I'll try to review the actual PR to see what's the case here, but that's my opinion in general.

gensim/models/tfidfmodel.py
```diff
@@ -29,7 +31,7 @@ def resolve_weights(smartirs):
 
     Parameters
     ----------
-    smartirs : str
+    smartirs : str or None
```
piskvorky: What does `None` do?

gensim/models/tfidfmodel.py
```diff
@@ -39,24 +41,28 @@ def resolve_weights(smartirs):
 
     Returns
     -------
-    3-tuple (local_letter, global_letter, normalization_letter)
+    str of (local_letter, global_letter, normalization_letter) or None
```
piskvorky: What does `None` result mean?

gensim/models/tfidfmodel.py
```diff
@@ -65,21 +71,43 @@ def resolve_weights(smartirs):
         doesn't fit the list of permissible values.
 
     """
+
```
piskvorky: PEP8 / PEP257: no blank line between docstring and function body.

gensim/models/tfidfmodel.py
```diff
@@ -65,21 +71,43 @@ def resolve_weights(smartirs):
...
+    if w_tf not in 'btnaldL':
+        raise ValueError("Expected term frequency weight to be one of 'btnaldL', except got {}".format(w_tf))
+
+    if w_df not in 'xnftp':
+        raise ValueError("Expected inverse document frequency weight to be one of 'xnftp', except got {}".format(w_df))
 
-    if w_df not in 'ntp':
-        raise ValueError("Expected inverse document frequency weight to be one of 'ntp', except got {}".format(w_df))
+    if w_n not in 'xncub':
+        raise ValueError("Expected normalization weight to be one of 'xncub', except got {}".format(w_n))
```
piskvorky:
```suggestion
        raise ValueError("Expected normalization weight to be one of 'xncub', got {}".format(w_n))
```

piskvorky: Lang fix: Please drop the word `except` from all these messages.

gensim/models/tfidfmodel.py
```diff
@@ -176,36 +206,33 @@ def smartirs_wglobal(docfreq, totaldocs, global_scheme):
...
-        Input array
+        The tf-idf vector.
     norm_scheme : {'n', 'c'}
-        Normalizing function to use:
-        `n`: no normalization
-        `c`: unit L2 norm (scale `x` to unit euclidean length)
+        Document length normalization scheme.
     return_norm : bool, optional
         Return the length of `x` as well?
-
```
piskvorky: Does this render correctly? I remember missing blank lines in between documentation sections can mess up the rending (as well as make the code harder to read).

I'd prefer to retain the section separator blanks, for readability reasons.

gensim/models/tfidfmodel.py
Witiko: Confuse the reader. I removed the polymorphism in 35f0f9db.

gensim/models/tfidfmodel.py
Witiko: Confuse the reader. I removed the polymorphism in 35f0f9db.

gensim/models/tfidfmodel.py
Witiko: Fixed in 5d1213ad.

gensim/models/tfidfmodel.py
Witiko: Fixed in 13081f76.

gensim/models/tfidfmodel.py
Witiko: This is definitely an error, which must have happened during my editing. Fixed in 78746020.

Witiko:
Thank you, @mpenkov and @piskvorky, for the helpful reviews. All change requests seem resolved.

gensim/models/tfidfmodel.py
```diff
@@ -274,20 +299,24 @@ def __init__(self, corpus=None, id2word=None, dictionary=None, wlocal=utils.iden
             for example 'ntc', 'bpn' and so on, where the letters represents the term weighting of the document vector.
 
             Term frequency weighing:
-                * `n` - natural,
-                * `l` - logarithm,
+                * `b` - binary,
```
mpenkov: Do we really need to duplicate SMART documentation here?

resolve_weights is a public function, so it will be visible to everyone via the docs. Would it be sufficient to hyperlink to that function instead?

> For valid values, see the :func:`~gensim.models.tfidfmodel.resolve_weights` function.

gensim/sklearn_api/tfidf.py
```diff
@@ -60,31 +60,37 @@ def __init__(self, id2word=None, dictionary=None, wlocal=gensim.utils.identity,
...
-
-            Document frequency weighting:
-                * `n` - none,
-                * `t` - idf,
-                * `p` - prob idf.
-
-            Document normalization:
-                * `n` - none,
-                * `c` - cosine.
+            local_letter : str
```
mpenkov: Can we hyperlink to resolve_weights instead?

gensim/models/tfidfmodel.py
Witiko: I am in favor of keeping the duplication, because the SMART letters are standard (we will not be adding new letters any time soon), and removing them from the `TfidfModel` class docstring forces the reader to jump through hoops to find the relevant documentation. Therefore, it is only a minor improvement for the maintainer and a net negative for the reader.

gensim/sklearn_api/tfidf.py
Witiko: See above.

gensim/models/tfidfmodel.py
Witiko: I am in favor of adding a `See Also` docstring to `resolve_weights` and `TfIdfTransformer.__init__` to make the maintainer's life easier.

mpenkov: Alright, fair enough.

Witiko:
Are there any other changes that you'd like to see, @piskvorky?

Witiko:
What I'd like to see is changing the default SMART scheme of `TfidfModel` from `nfc` to something more recent, such as `dtb` suggested by [Singhal](http://singhal.info/ieee2001.pdf), see our [previous discussion](https://groups.google.com/d/msg/gensim/15OMa3-1fdE/9KLRCYRWHQAJ). This would help eliminate [weak baselines](https://sigir.org/wp-content/uploads/2019/01/p040.pdf), but it would also break user code.

piskvorky:
We can add a **fat warning / recommendation box** into the documentation, so people know that (hopefully) superior alternatives exist. But I'd be wary of changing defaults, especially for well-established methods like TFIDF.

Although the "pivoted normalization" project in particular was a massive flop in my estimation. I'd need to see a much more coherent discussion of its merits and modes of use, along with benchmarks, before doing anything with it. To be honest, I'm more leaning toward ripping it out of Gensim, because both the code and its documentation (tutorial) were seriously weak IIRC.

Witiko:
> We can add a **fat warning / recommendation box** into the documentation, so people know that (hopefully) superior alternatives exist. But I'd be wary of changing defaults, especially for well-established methods like TFIDF.

TFIDF is a blanket term, which covers various weighting schemes including pivoted normalization. I agree that the `TfidfModel` class is established and changing its default weighting scheme may lead to confusion.

> Although the "pivoted normalization" project in particular was a massive flop in my estimation. I'd need to see a much more coherent discussion of its merits and modes of use, along with benchmarks, before doing anything with it. To be honest, I'm more leaning toward ripping it out of Gensim, because both the code and its documentation (tutorial) were seriously weak IIRC.

Pivoted normalization improves the performance of TFIDF, as shown in the TREC SMART papers, see for example the [TREC 8 results](https://trec.nist.gov/pubs/trec8/papers/sabir8.ps):

Spectacle J32602

If the published results are too old / not persuasive, we can run our own benchmarks.

piskvorky:
Thanks for the detailed follow-up @Witiko.

Yes, I always take academic results with a grain of salt, because of "publish-or-perish". SOTA tables typically don't factor in complexity and algorithmic robustness, which is critical for real-world inputs (and that's when they don't directly overfit to the SOTA dataset, which they often do). Simplicity and sanity >> a few percent lift in accuracy.

But I believe the main issue with the original project was its lack of documentation and workflow motivation, more than the code. IIRC the pivoted normalization required the user to supply some barely-documented parameter. IIRC I asked the author about this and they had no answer, rendering the whole project utterly useless and academic. I doubt anyone's used that implementation in Gensim since.

If you could fix that (the docs, possibly code), that'd be great. The other option is removing it.

Witiko:
Note that this pull request is a large step, because now the pivoted normalization is hidden behind the SMART API. Just specifying `smart='dtb'` uses pivoted normalization behind the scenes, although the user is free to tweak the `slope` parameter.

piskvorky:
OK, great. I'll have to re-read your updated docs, because our [current docs](https://radimrehurek.com/gensim/models/tfidfmodel.html#gensim.models.tfidfmodel.TfidfModel):

Screen Shot 2019-05-08 at 21 01 02

are utterly impotent. I wouldn't know what to set these parameters to myself => useless mode. Just looking at it, I'm angry we ever merged that 😠

Witiko:
Currently, the docs only specify that these parameters are overriden by SMART. There is definitely some room for improvement.

Witiko:
@piskvorky I suggest the following text:

> ### pivot : float, optional
> In pivoted document length normalization, the effective norm of a document is the weighted average of the old norm and a pivot: latex_b37b39905b07667424ea9842348f0ca1.
> 
> When `pivot` is None, and `smartirs` specifies either the unique (`u`) or the character-length (`b`) pivoted document length normalization scheme, and either `corpus` or `dictionary` are specified, then the pivot will be determined automatically.
> 
> See also the blog post at https://rare-technologies.com/pivoted-document-length-normalisation/.
>
> ### slope : float, optional
> In pivoted document length normalization, the effective norm of a document is the weighted average of the old norm and a pivot: latex_b37b39905b07667424ea9842348f0ca1.
> 
> Setting `slope` to 1.0 disables pivoted document length normalization, and setting `slope` to 0.0 uses only the pivot as the norm. Singhal suggests setting `slope` between 0.2 and 0.3 for best results.  Default is 0.25.
> 
> See also the blog post at https://rare-technologies.com/pivoted-document-length-normalisation/.

piskvorky:
Thanks @Witiko ! Much better. Can you please expand two things:

1. What is *pivoted length normalization* (not only the formula, but what purpose does it serve, where does it fit conceptually in the processing pipeline, who should care & why – a short sentence or two, upfront, for context).

2. This paragraph:

> When `pivot` is None, `smartirs` specifies either the unique (`u`) or the character-length (`b`) pivoted document length normalization scheme, and either `corpus` or `dictionary` are specified, then the pivot will be determined automatically. Otherwise when `pivot` is None, pivoted document length normalization will be disabled. Default is None.

is hard to parse. I think I got what you mean; how about this instead:

> Pivot can be either specified directly as a float, or left to be determined automatically.
>
> To set `pivot` automatically, do all of the following three steps:
> 1. Leave `pivot=None` (default value)
> 2. Set either `u` or `b` mode in the `smartirs` parameter.
> 3. Set either `corpus` or `dictionary` parameter (the `pivot` value will be determined automatically from the properties of that particular `corpus` or `dictionary`). 
>
> If `pivot=None` (default value) but steps 2. and 3. above are not met, pivot normalization is turned off (no pivot normalization).

If my reading of your explanation here is correct (is it?), the API seems unfortunate. `pivot=None` meaning both "ignore pivot" and "pivot is active and determined automatically", based on values of some other parameters, is not good API design. Can you think of a way to make that cleaner?

> I can also write a follow-up article to rare-technologies.com/pivoted-document-length-normalisation, so that users see how to produce a strong baseline using the SMART API.

That'd be great indeed! Then I'd link the two through, and promote this "blog series", once it offers more actionable clarity to readers.

Witiko:
@piskvorky:
> 1. What is _pivoted length normalization_ (not only the formula, but what purpose does it serve, where does it fit conceptually in the processing pipeline, who should care & why – a short sentence or two max, upfront, for context).

We can add the following short explanation: “In information retrieval, TF-IDF is biased against long documents. Pivoted document length normalization solves this problem by changing the norm of a document as follows: `slope * old_norm + (1.0 - slope) * pivot`.”

> If my reading of your explanation here is correct (is it?), the API seems unfortunate. `pivot=None` meaning both "ignore pivot" and "pivot is active and determined automatically", based on values of some other parameters, is not good API design. Can you think of a way to make that cleaner?

We can make `smartirs` always override `pivot`, i.e. the value of `pivot` will be ignored if `smartirs` is specified. This is much more intuitive, since `smartirs` is a higher-level API, it is consistent, because `smartirs` already overrides `normalize`, and it allows us to skip step 1 in the documentation:

1. Set either `u` or `b` document normalization mode in the `smartirs` parameter.
2. Set either `corpus` or `dictionary` parameter (the `pivot` will be determined automatically from the properties of that particular `corpus` or `dictionary`).

> Is setting `pivot=0.25` enough to active the "manual" pivoted normalizaton, or do I still have to do step 2?

Yes, you only need `corpus` or `dictionary` to automatically determine `pivot`. When you specify `pivot` manually, `corpus` or `dictionary` is not needed. This is also the current behavior of Gensim, nothing changes here.

This is the current suggested text of the documentation:

> ### pivot *(float, optional)*
> In information retrieval, TF-IDF is biased against long documents [1]. Pivoted document length normalization solves this problem by changing the norm of a document as follows: `slope * old_norm + (1.0 - slope) * pivot`.
> 
> You can either specify the `pivot` directly as a float, or leave it to be determined automatically. To set `pivot` automatically, do all of the following three steps:
>
> 1. Set either `u` or `b` document normalization mode in the `smartirs` parameter.
> 2. Set either `corpus` or `dictionary` parameter (the `pivot` will be determined automatically from the properties of that particular `corpus` or `dictionary`).
>
> If `pivot` is None (default value) and steps 2. and 3. are not met, pivoted document length normalization will be disabled.
> 
> See also the blog post at https://rare-technologies.com/pivoted-document-length-normalisation/.
>
> ### slope *(float, optional)*
> In information retrieval, TF-IDF is biased against long documents [1]. Pivoted document length normalization solves this problem by changing the norm of a document as follows: `slope * old_norm + (1.0 - slope) * pivot`.
> 
> Setting `slope` to 0.0 uses only the `pivot` as the norm, and setting `slope` to 1.0 disables pivoted document length normalization. Singhal [2] suggests setting `slope` between 0.2 and 0.3 for best results.  Default is 0.25.
> 
> See also the blog post at https://rare-technologies.com/pivoted-document-length-normalisation/.
>
> ## References
> 1. Singhal, A., Buckley, C., & Mitra, M. (1996). Pivoted Document Length Normalization. *SIGIR Forum*, 51, 176-184.
> 2. Singhal, A. (2001). Modern information retrieval: A brief overview. *IEEE Data Eng. Bull.*, 24(4), 35-43.

piskvorky:
@mpenkov WDYT? Does this documentation upgrade make more sense?

Witiko:
I think the new text strikes a nice balance: it is brief and informative. If you have no other suggestions, I will commit the changes, so that the PR is ready for the 3.8 release.

@piskvorky Please, let me know when you'd need the article, and what markup language I should use for the submission.

> That'd be great indeed! Then I'd link the two through, and promote this "blog series", once it offers clearer value to readers (ideally with some actionable, "obviously useful" example).

SMART was developed for information retrieval, but text classification is easier to discuss, so I have yet to decide if I will use one or the other (or both) for the examples. The idea is to start with BOW and work our way through various SMART schemes, ending up with a much stronger vector space model.

mpenkov:
Looks good to me.

Trying to work out why the Appveyor builds are failing. Once that's sorted, we can merge.

Witiko:
@mpenkov Thank you for the feedback. We discussed the AppVeyor builds in #2497.

Witiko:
@piskvorky @mpenkov I just pushed the documentation changes. The rendered documentation looks as follows. With the AppVeyor build fixed, this PR should be ready for merge.

scrot

mpenkov:
OK, finally merged. @Witiko Thank you for your contribution!

PR #2423, opened 2019-03-19
Update word2vec.ipynb

asyabo:
This code sample doesn't work with 'rb' mode. The error would be "TypeError: can't concat str to bytes".

### DIFF
--- a/docs/notebooks/word2vec.ipynb
+++ b/docs/notebooks/word2vec.ipynb
@@ -116,7 +116,7 @@
     " \n",
     "    def __iter__(self):\n",
     "        for fname in os.listdir(self.dirname):\n",
-    "            for line in smart_open(os.path.join(self.dirname, fname), 'rb'):\n",
+    "            for line in smart_open(os.path.join(self.dirname, fname), 'r'):\n",
     "                yield line.split()"
    ]
   },


mpenkov:
Good catch! Congrats on your first contribution to gensim :+1:

docs/notebooks/word2vec.ipynb
```diff
@@ -116,7 +116,7 @@
     " \n",
     "    def __iter__(self):\n",
     "        for fname in os.listdir(self.dirname):\n",
-    "            for line in smart_open(os.path.join(self.dirname, fname), 'rb'):\n",
+    "            for line in smart_open(os.path.join(self.dirname, fname), 'r'):\n",
```
piskvorky: Please use an explicit encoding. Opening in "text mode", without specifying the encoding, should always be considered an error.

There are two ways to do this: either open in binary mode and then decode; or open in text mode and provide `encoding` parameter. The latter is only available in Python 3, IIRC.

mpenkov: > Opening a file in "text mode", without specifying the encoding, should always be considered an error.

Why should it be considered an error? There is a default encoding that works the vast majority of the time.

> The latter is only available in Python 3, IIRC.

We're using smart_open here, so it doesn't matter which Python version we're running.

piskvorky: Oh, we back-ported the `encoding` parameter into python2, in `smart_open`? That's cool, I didn't know (or forgot).

> Why should it be considered an error? There is a default encoding that works the vast majority of the time.

It's an error because it consistently leads to errors and subtle bugs (in addition to being unnecessarily about the code intent). We never want to rely on any "default encodings", nor encourage such engineering patterns. If the code is expecting `utf8` (or `ascii` or …), it should say so.

mpenkov: There are dozens of places where we invoke smart_open without specifying the encoding to open text. Should we go through the docs and explicitly specify the encoding in each case?

piskvorky: Definitely! Same goes for Gensim or any other of our code or tutorials. I'm not sure when / how these bugs slipped through, but if you find any, please squash them with mercy!

Is it possible that they're somehow related to the previous behaviour of smart_open as "binary by default", which was changed recently, to match built-in `open`?

mpenkov: No, because we never updated the documentation to use the new `open` function.

Issue #2455, opened 2019-04-20
accuracy function for analogy test gives back empty lists as result


vlfgns5:
#### Problem description
Analogy test for accuracy function returns empty list due to typo. (topn = False)

#### Solution
There are 2 solutions regarding this matter in my opinion
: First one is the change  keyedvectors.py script in gensim/models on line 1187

Line : 1187 in keyedvectors.py
(https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/keyedvectors.py)

sims = most_similar(self, positive=[b, c], negative=[a], topn=False, restrict_vocab=restrict_vocab)
==>
sims = most_similar(self, positive=[b, c], negative=[a], topn=None, restrict_vocab=restrict_vocab)

Second solution
: change keyedvectors.py script in gensim/models on line 514, 556

if topn is not None and topn < 1:
==>
if not topn and topn<1:

&&

if topn is None:
==>
if not topn:

Well...first one is better i think

piskvorky:
Thanks for reporting @vlfgns5.

The problem seems to have been introduced by #2356. @Witiko can you have a look? Are there other places similarly affected by that change?

Witiko:
@piskvorky You are correct. Where previously topn was only cast to a boolean, #2356 gave `topn=0` different semantics (return an empty iterable) than `topn=None` (return all similarities, previously undocumented behavior). However, I missed callers that would use `topn=False`.

According to `grep topn=False`, there are no callers other than `EuclideanKeyedVectors.accuracy` that use `topn=False`. I can either fix the called by changing `topn=False` to `topn=None`, or I can make `most_similar` accept both `False` and `None`.

Witiko:
Digging deeper into this, properly supporting `most_similar(…, topn=None)` needs more work. When `indexer` is `None`, then `topn=None` (and `topn=False`) returns the expected result. However, when an `indexer` is provided (such as `Annoyindexer`), then the call fails, because Annoy only supports integer `topn`. Therefore, we need an additional fix that disables the indexer when `topn` is None:

```diff
< if indexer is not None:
> if indexer is not None and isinstance(topn, int):
```

PR #2461, opened 2019-04-23
 Fix WordEmbeddingsKeyedVectors.most_similar

Witiko:
This PR continues #2356 in reaction to #2455. The PR accomplishes the following:
- Regain support for `most_similar(…, topn=x)`, where `bool(x)` is `False`.
- Document `most_similar(…, topn=None)`.
- Support `topn=None` in `most_similar_cosmul`, `similar_by_word`, and `similar_by_vector`.
- Disable `indexer` in `most_similar` when `topn` is `None`, since indexers don't support `topn=None`.


gensim/models/keyedvectors.py
```diff
@@ -507,11 +508,12 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
 
         Returns
         -------
-        list of (str, float)
-            Sequence of (word, similarity).
+        {list of (str, float), numpy.array}
```
mpenkov: I can see that by `{A, B}` you mean `either A or B`. Is this a standard form of expression? If not, then perhaps it may be easier to explicitly say `A or B` in order to avoid confusing users. For example, upon seeing this first, I thought the return is some kind of set or dict.

gensim/models/keyedvectors.py
```diff
@@ -507,11 +508,12 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+            Sequence of (word, similarity) when `topn` is int. When `topn` is None, then
```
```suggestion
            When `topn` is an int, sequence of `topn` elements in the form of (word, similarity). When `topn` is None, then
```

gensim/models/keyedvectors.py
```diff
@@ -507,11 +508,12 @@ def most_similar(self, positive=None, negative=None, topn=10, restrict_vocab=Non
...
+            similarities for all words are returned.
```
mpenkov: I think mentioning the dimensionality of the array is helpful here.

```suggestion
            similarities for all words are returned as a numpy.array.
            The array will have N rows and M columns, where N is ... and M is ....
```

gensim/models/keyedvectors.py
Witiko: Thank you for spotting this. The numpy conventions use this syntax for enumerator types:

```python
order : {'C', 'F', 'A'}
    Description of `order`.
```

However, `or` is the correct operator for a union of types. Fixed in 7d017430.

gensim/models/keyedvectors.py
Witiko: Fixed in e1ea69c3.

gensim/models/keyedvectors.py
Witiko: Added in 7594c9ec.

Witiko:
> What do you think? Is it possible to test this new fix?

Thank you for the review. I added two unit tests in fbfe2d4 and 2f16efe.

PR #2466, opened 2019-04-28
Nmf bugfix

mpenkov:
The logging bugs are blocking our gensim feedstock release on conda.

The divide-by-zero doesn't cause any test failures, but we should fix it anyway.


piskvorky:
CC @anotherbugmaster.

mpenkov:
I'm gonna go ahead and merge this in the interest of time.

PR #2469, opened 2019-04-29
Adding type check for corpus_file argument

### LINKED ISSUE #2460: Passing nonsense (int) for `corpus_file` generates confusing OSError in get_offsets_and_start_doctags_for_corpusfile
#### Problem description

I am trying to train a doc2vec model over 1.35 million docs. Each line of the opened file corresponds to a doc and I am creating a TaggedDocument object from that. I am getting these errors which I cannot resolve. Note that the 1351747 below corresponds to the number of docs.
```
OSError: [Errno 9] Bad file descriptor: 1351747
```

This is my code:
```python
taggeddoc = []
f = open('EnglishT.txt')
for line in f:
    td = TaggedDocument(gensim.utils.to_unicode(str.encode(line)).split(),[str(index)])
    taggeddoc.append(td)
    index += 1
f.close()
print(len(taggeddoc))
model = gensim.models.doc2vec.Doc2Vec(vector_size=50, dm=0, dbow_words=0, workers=30, alpha=0.025, min_alpha=0.025)

model.build_vocab(taggeddoc)
# start training
for epoch in range(500):
    model.train(taggeddoc, model.corpus_count, epochs=model.iter)
    model.alpha -= 0.002  
    model.min_alpha = model.alpha 
```

Using Python3:
```
  File "en_d2v.py", line 25, in <module>
    model.train(taggeddoc, model.corpus_count, epochs=model.iter)
  File "/home/junlp/.local/lib/python3.6/site-packages/gensim/models/doc2vec.py", line 799, in train
    offsets, start_doctags = self._get_offsets_and_start_doctags_for_corpusfile(corpus_file, self.workers)
  File "/home/junlp/.local/lib/python3.6/site-packages/gensim/models/doc2vec.py", line 828, in _get_offsets_and_start_doctags_for_corpusfile
    corpus_file_size = os.path.getsize(corpus_file)
  File "/usr/lib/python3.6/genericpath.py", line 50, in getsize
    return os.stat(filename).st_size
OSError: [Errno 9] Bad file descriptor: 1351747
```

Using Python2
```
  File "en_d2v.py", line 32, in <module>
    model.train(taggeddoc, model.corpus_count, epochs=model.iter)
  File "/home/junlp/.local/lib/python2.7/site-packages/gensim/models/doc2vec.py", line 799, in train
    offsets, start_doctags = self._get_offsets_and_start_doctags_for_corpusfile(corpus_file, self.workers)
  File "/home/junlp/.local/lib/python2.7/site-packages/gensim/models/doc2vec.py", line 828, in _get_offsets_and_start_doctags_for_corpusfile
    corpus_file_size = os.path.getsize(corpus_file)
  File "/usr/lib/python2.7/genericpath.py", line 57, in getsize
    return os.stat(filename).st_size
TypeError: coercing to Unicode: need string or buffer, int found
```

#### Versions

Linux-4.15.0-46-generic-x86_64-with-Ubuntu-18.04-bionic
('Python', '2.7.15rc1 (default, Nov 12 2018, 14:31:15) \n[GCC 7.3.0]') 
('NumPy', '1.16.2')
('SciPy', '1.2.1')
No handlers could be found for logger "smart_open.ssh"
('gensim', '3.7.2')
('FAST_VERSION', 1)

For Python3,
Python 3.6.7 (default, Oct 22 2018, 11:32:17)

saraswatmks:
Add a type check for `corpus_file` argument and raise type error if the type of `corpus_file` is not a string. Fixes #2460.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -795,19 +795,15 @@ class Doc2Vec(BaseWordEmbeddingsModel):
         """
         kwargs = {}
 
-        # Check if both documents and corpus_file are None
         if corpus_file is None and documents is None:
             raise TypeError("Either one of corpus_file or documents value must be provided")
 
-        # Check if both documents and corpus_file are not None
         if corpus_file is not None and documents is not None:
             raise TypeError("Both corpus_file and documents must not be provided at the same time")
 
-        # Check if corpus_file is string type
         if documents is None and not os.path.isfile(corpus_file):
             raise TypeError("Parameter corpus_file must be a valid path to a file, got %r instead" % corpus_file)
 
-        # Check if documents is iterable
         if documents is not None and not isinstance(documents, Iterable):
             raise TypeError("documents must be an iterable of list, got %r instead" % documents)
 
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -807,9 +807,9 @@ class Doc2Vec(BaseWordEmbeddingsModel):
         if not isinstance(corpus_file, string_types) and documents is None:
             raise TypeError("Parameter corpus_file must be a valid path to a file, got %s instead." % corpus_file)
 
-        # Check if documents is iterable and not string type
-        if isinstance(documents, Iterable) and isinstance(documents, string_types):
-            raise TypeError("documents must be an iterable of list and not a string type.")
+        # Check if documents is iterable
+        if not isinstance(documents, Iterable):
+            raise TypeError("documents must be an iterable of list, got %s instead." % documents)
 
         if corpus_file is not None:
             # Calculate offsets for each worker along with initial doctags (doctag ~ document/line number in a file)
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -797,19 +797,19 @@ class Doc2Vec(BaseWordEmbeddingsModel):
 
         # Check if both documents and corpus_file are None
         if corpus_file is None and documents is None:
-            raise TypeError("Either one of corpus_file or documents value must be provided.")
+            raise TypeError("Either one of corpus_file or documents value must be provided")
 
         # Check if both documents and corpus_file are not None
         if corpus_file is not None and documents is not None:
-            raise TypeError("Instead provide value to either of corpus_file or documents parameter but not both.")
+            raise TypeError("Both corpus_file and documents must not be provided at the same time")
 
         # Check if corpus_file is string type
         if documents is None and not os.path.isfile(corpus_file):
-            raise TypeError("Parameter corpus_file must be a valid path to a file, got %s instead." % corpus_file)
+            raise TypeError("Parameter corpus_file must be a valid path to a file, got %r instead" % corpus_file)
 
         # Check if documents is iterable
         if documents is not None and not isinstance(documents, Iterable):
-            raise TypeError("documents must be an iterable of list, got %s instead." % documents)
+            raise TypeError("documents must be an iterable of list, got %r instead" % documents)
 
         if corpus_file is not None:
             # Calculate offsets for each worker along with initial doctags (doctag ~ document/line number in a file)
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -800,12 +800,12 @@ class Doc2Vec(BaseWordEmbeddingsModel):
             raise TypeError("Either one of corpus_file or documents value must be provided.")
 
         # Check if both documents and corpus_file are not None
-        if corpus_file is not None and documents is not None:
+        if not ((corpus_file is None) ^ (documents is None)):
             raise TypeError("Instead provide value to either of corpus_file or documents parameter but not both.")
 
         # Check if documents is not None and not string type either
         if documents is not None and isinstance(documents, string_types):
-            raise TypeError("Documents must be an iterable of list and not a string type.")
+            raise TypeError("documents must be an iterable of list and not a string type.")
 
         # Check the type of corpus_file
         if not isinstance(corpus_file, string_types) and documents is None:
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -800,11 +800,11 @@ class Doc2Vec(BaseWordEmbeddingsModel):
             raise TypeError("Either one of corpus_file or documents value must be provided.")
 
         # Check if both documents and corpus_file are not None
-        if not ((corpus_file is None) ^ (documents is None)):
+        if corpus_file is not None and documents is not None:
             raise TypeError("Instead provide value to either of corpus_file or documents parameter but not both.")
 
         # Check if corpus_file is string type
-        if documents is None and not isinstance(corpus_file, string_types):
+        if documents is None and not os.path.isfile(corpus_file):
             raise TypeError("Parameter corpus_file must be a valid path to a file, got %s instead." % corpus_file)
 
         # Check if documents is iterable
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -70,7 +70,7 @@ try:
 except ImportError:
     from Queue import Queue  # noqa:F401
 
-from collections import namedtuple, defaultdict, Iterable
+from collections import namedtuple, defaultdict
 from timeit import default_timer
 
 from numpy import zeros, float32 as REAL, empty, ones, \
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -809,8 +809,8 @@ class Doc2Vec(BaseWordEmbeddingsModel):
 
         # Check the type of corpus_file
         if corpus_file is not None and not isinstance(corpus_file, string_types):
-            raise TypeError(f"""Parameter corpus_file of train() must be a
-                             string (path to a file) got {corpus_file} instead.""")
+            raise TypeError("""Parameter corpus_file of train() must be a
+                             string (path to an existing file) got %s instead.""" % corpus_file)
 
         if corpus_file is not None:
             # Calculate offsets for each worker along with initial doctags (doctag ~ document/line number in a file)
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -800,16 +800,17 @@ class Doc2Vec(BaseWordEmbeddingsModel):
             raise TypeError("Either one of corpus_file or documents value must be provided.")
 
         # Check if both documents and corpus_file are not None
-        if not corpus_file is None and not documents is None:
+        if corpus_file is not None and documents is not None:
             raise TypeError("Instead provide value to either of corpus_file or documents parameter but not both.")
 
         # Check if documents is not None and iterable but not string type
-        if not documents is None and isinstance(documents, Iterable) and isinstance(documents, string_types):
+        if documents is not None and isinstance(documents, Iterable) and isinstance(documents, string_types):
             raise TypeError("Documents must be an iterable of list and not a string type.")
 
         # Check the type of corpus_file
         if corpus_file is not None and not isinstance(corpus_file, string_types):
-            raise TypeError(f"Parameter corpus_file of train() must be a string (path to a file) got {corpus_file} instead.")
+            raise TypeError(f"""Parameter corpus_file of train() must be a
+                             string (path to a file) got {corpus_file} instead.""")
 
         if corpus_file is not None:
             # Calculate offsets for each worker along with initial doctags (doctag ~ document/line number in a file)
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -803,14 +803,13 @@ class Doc2Vec(BaseWordEmbeddingsModel):
         if corpus_file is not None and documents is not None:
             raise TypeError("Instead provide value to either of corpus_file or documents parameter but not both.")
 
-        # Check if documents is not None and iterable but not string type
-        if documents is not None and isinstance(documents, Iterable) and isinstance(documents, string_types):
+        # Check if documents is not None and not string type either
+        if documents is not None and isinstance(documents, string_types):
             raise TypeError("Documents must be an iterable of list and not a string type.")
 
         # Check the type of corpus_file
-        if corpus_file is not None and not isinstance(corpus_file, string_types):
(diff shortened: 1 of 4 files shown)


gensim/models/doc2vec.py
```diff
@@ -794,6 +794,11 @@ def train(self, documents=None, corpus_file=None, total_examples=None, total_wor
 
         """
         kwargs = {}
+
+        # Check the type of corpus_file
+        if not isinstance(corpus_file, string_types):
+            raise TypeError("Parameter corpus_file of train() must be a string (path to a file).")
```
piskvorky: Can you also show what the received parameter is instead?

Having concrete error messages helps avoid confusion.

gensim/models/doc2vec.py
```diff
@@ -794,6 +794,11 @@ def train(self, documents=None, corpus_file=None, total_examples=None, total_wor
 
         """
         kwargs = {}
+
+        # Check the type of corpus_file
+        if not isinstance(corpus_file, string_types):
```
mpenkov: `corpus_file` may legitimately be None, when `documents` is not None. This is the reason why some of the unit tests fail. Please have a look at Travis CI.

So, if you go ahead with your proposed check, you need something like:

```suggestion
        if corpus_file is not None and not isinstance(corpus_file, string_types):
```

Also, please add a unit test that stresses your new functionality (pass in a non-string `corpus_file`, expect a `TypeError` raised).

Next, what is the benefit of raising TypeError here? What happens with the existing code if we do _not_ raise TypeError? What kind of exception does the user see?

I think if we go ahead with this kind of parameter checking, we should do it properly:

1. Ensure that one of `documents` or `corpus_file` is None (they cannot both be non-None)
2. Ensure that one of `documents` or `corpus_file` is not None (they cannot both be None)
3. If `documents` is not None, then it must be an iterable
4. If `corpus_file` is not None, then it must be a string

My question is: is it worth it? @piskvorky

piskvorky: Agreed on the tests. I edited the PR description for context (link to original issue).

saraswatmks: @piskvorky @mpenkov Thanks for the detailed feedback. Working on it.

gojomo: Also I'd say: while checking whether it's a string may help a bit, if there's checking, it'd make sense to ensure things like illegal or missing paths also generate meaningful error messages. It might be possible to just leverage some existing path-checking/file-existence-checking method, or wrap the failing method(s) in a handler that catches errors and shows a good-enough error message like "Problem with `corpus_file` value X".

gensim/models/doc2vec.py
```diff
@@ -794,6 +794,24 @@ def train(self, documents=None, corpus_file=None, total_examples=None, total_wor
...
         kwargs = {}
+
+        # Check if both documents and corpus_file are None
+        if corpus_file is None and documents is None:
+            raise TypeError("Either one of corpus_file or documents value must be provided.")
+
+        # Check if both documents and corpus_file are not None
+        if corpus_file is not None and documents is not None:
+            raise TypeError("Instead provide value to either of corpus_file or documents parameter but not both.")
+
```
gojomo: I'd be tempted to combine these two cases with an XOR: `if (corpus_file is None) ^ (documents is None):`.

gensim/models/doc2vec.py
```diff
@@ -794,6 +794,24 @@ def train(self, documents=None, corpus_file=None, total_examples=None, total_wor
...
+        # Check if documents is not None and iterable but not string type
+        if documents is not None and isinstance(documents, Iterable) and isinstance(documents, string_types):
+            raise TypeError("Documents must be an iterable of list and not a string type.")
```
gojomo: I believe the test of is-`Iterable` and is-`string_types` is likely redundant, as all strings are iterable.

gensim/models/doc2vec.py
```diff
@@ -794,6 +794,24 @@ def train(self, documents=None, corpus_file=None, total_examples=None, total_wor
...
+
+        # Check the type of corpus_file
+        if corpus_file is not None and not isinstance(corpus_file, string_types):
+            raise TypeError("""Parameter corpus_file of train() must be a
+                             string (path to an existing file) got %s instead.""" % corpus_file)
```
gojomo: Since `None` will already fail the is-`string_types` test, the check `is not None` is redundant.

gensim/models/doc2vec.py
mpenkov: I think it's also important to not go overboard with the checking. For example, you're making sure it's not a string here. Why not take that further and make sure it's not an int? A float? A file buffer? And so on.

gensim/models/doc2vec.py
mpenkov: The use of a multiline string is poor here. Your exception message contains two lines, and a lot of unnecessary spaces.

I think you also don't need to mention the function name: it'll be obvious from the stack trace.

gensim/models/doc2vec.py
saraswatmks: @gojomo I was not sure how to go about with handling this. `documents` should be an iterable 
but not a string since like you said strings are iterable. I think I got it now, so instead of checking for both, checking only for `string` type would do.

gensim/models/doc2vec.py
saraswatmks: I did it this way for readability. Never mind, I'll make the changes.

gensim/models/doc2vec.py
saraswatmks: @gojomo so should I check `corpus_file` for being a valid file path instead of checking for `string_type` ?

gensim/models/doc2vec.py
saraswatmks: @mpenkov noob question: should I resolve the conversations or the person who reviews the PR does it ?

gensim/models/doc2vec.py
gojomo: Supplying a (path) string as `documents` might sufficiently reveal itself to be an error in other ways - like training that completes almost instantly, with little to no vocabulary. Separately, I'm not sure that all objects that would work because they are effectively "iterable" will actually test positive as being `isinstance()` `Iterable`.

gensim/models/doc2vec.py
gojomo: Yes, it'd actually need to be `if not ((corpus_file is None) ^ (documents is None)):` - but then essentially one error message, roughly "supply one or the other but not both", would be fine for either mistake.

gensim/models/doc2vec.py
gojomo: I'm not sure, what covers the most cases with the most straightforward code? A path-test, with the right catches/error-message, might "catch N birds with one stone", pointing the user in the right direction to understand their parameter error with fewer conditionals/lines-of-tests. Or it might complicate things compared to the simple proper-type test. But the motivating case for this change - that users following slightly-older examples would get a confusing error message, due to a new `corpus_file` parameter` – will actually already have been handled by the simple "one or the other but not both" test. So, even the string test isn't strictly required to address the motivating case. My pref would be: do the simplest thing that resolves the motivating case, for sure. If there's a easy/clear bit of extra checking that makes also makes sense, consider it as well.

gensim/models/doc2vec.py
mpenkov: I think it makes sense for the original reviewer to do it, once they're happy their comments have been addressed.

saraswatmks:
@mpenkov @gojomo Kindly check my latest commit. I've made the changes as suggested. For now, we test four things before training would begin (as suggested):
1. Check if both `corpus_file` and `documents` are None.
2. Check if both `corpus_file` and `documents` are provided.
3. Check if `corpus_file` is not a string type.
4. Check if `documents` is not Iterable.

Explanation point 4: I check only for being iterable to handle cases when someone would pass integer, float, bool value to `documents`. In case someone passes string, like @gojomo mentioned above, the training will finish super fast with almost no vocabulary.

I think these four cases cover most plausible situations which users might encounter.

gensim/models/doc2vec.py
piskvorky: Strong -1 on this: unreadable. Please stick to simpler (and easier to maintain) constructs .

gensim/models/doc2vec.py
```diff
@@ -794,6 +794,23 @@ def train(self, documents=None, corpus_file=None, total_examples=None, total_wor
...
+        # Check if both documents and corpus_file are None
+        if corpus_file is None and documents is None:
+            raise TypeError("Either one of corpus_file or documents value must be provided.")
+
+        # Check if both documents and corpus_file are not None
+        if not ((corpus_file is None) ^ (documents is None)):
+            raise TypeError("Instead provide value to either of corpus_file or documents parameter but not both.")
+
+        # Check if corpus_file is string type
+        if documents is None and not isinstance(corpus_file, string_types):
```
mpenkov: If `corpus_file` must be a valid path to a file, then why not explicitly check for that, instead of checking the type?

```suggestion
        if documents is None and not os.path.isfile(corpus_file):
```

piskvorky: Wouldn't that throw an exception for incorrect (non-string) `corpus_file`? It's an interesting idea for a try-catch block though.

mpenkov: We could wrap it in a _is_valid_corpus(corpus_file) function for malformed input.

Given that this input validation will be happening in mulitple places, it's probably not a bad idea.

saraswatmks: @piskvorky @mpenkov I've kept it simple. I used `os.path.isfile` as suggested. It returns `False` at all type of invalid inputs including non-string / float or whatever trash value a user would pass. 

Also, to do this for fasttext, I think it will be better:
1. To create a separate PR
2. Wait until the this PR is finalised so that I can follow the same code approach there are well.

gensim/models/doc2vec.py
gojomo: The clearest approach would be not to have a method with such delicate XOR one-but-not-the-other positional parameters - use separate methods with distinct names! But failing that, a one-line XOR, whose next line is an error message with a plain-english description of the constraint, seems pretty simple to me - just accurately reflecting the design choice already made.

gensim/models/doc2vec.py
mpenkov: I'm -1 on creating separate PRs for this, for several reasons:

1. The changes are tiny. They span several lines of code.
2. The changes will be identical for each model (e.g. doc2vec, fasttext, etc)
3. PR overhead (review, merge, release, document) is not worth the effort.
4. It's salami-slicing.

For your second point, sure, I think it's a good idea to wait until people come to a consensus before expanding your approach to the other models.

saraswatmks: @mpenkov Could you please review the latest changes and let me know is we are good to go or something else needs to be done ?

gensim/models/doc2vec.py
```diff
@@ -794,6 +794,23 @@ def train(self, documents=None, corpus_file=None, total_examples=None, total_wor
...
         """
         kwargs = {}
+
+        # Check if both documents and corpus_file are None
+        if corpus_file is None and documents is None:
+            raise TypeError("Either one of corpus_file or documents value must be provided.")
+
+        # Check if both documents and corpus_file are not None
+        if corpus_file is not None and documents is not None:
+            raise TypeError("Instead provide value to either of corpus_file or documents parameter but not both.")
```
mpenkov:
```suggestion
            raise TypeError("Provide either corpus_file or documents parameter, but not both.")
```

mpenkov: For consistency with the above.

```suggestion
            raise TypeError("Both corpus_file and documents may not be provided at the same time")
```

gensim/models/doc2vec.py
```diff
@@ -794,6 +794,23 @@ def train(self, documents=None, corpus_file=None, total_examples=None, total_wor
 
         """
         kwargs = {}
+
+        # Check if both documents and corpus_file are None
+        if corpus_file is None and documents is None:
+            raise TypeError("Either one of corpus_file or documents value must be provided.")
```
mpenkov:
```suggestion
            raise TypeError("Either one of corpus_file or documents value must be provided")
```

I tried committing these nitpicks to your branch myself, but I don't have permissions.

gensim/models/doc2vec.py
```diff
@@ -794,6 +794,23 @@ def train(self, documents=None, corpus_file=None, total_examples=None, total_wor
...
+
+        # Check if both documents and corpus_file are not None
+        if corpus_file is not None and documents is not None:
+            raise TypeError("Instead provide value to either of corpus_file or documents parameter but not both.")
+
+        # Check if corpus_file is string type
+        if documents is None and not os.path.isfile(corpus_file):
+            raise TypeError("Parameter corpus_file must be a valid path to a file, got %s instead." % corpus_file)
```
mpenkov:
```suggestion
            raise TypeError("Parameter corpus_file must be a valid path to a file, got %r instead." % corpus_file)
```

Better than %s, because corpus_file may contain spaces.

gensim/models/doc2vec.py
mpenkov: Sure. Left you some nitpicky comments.

In the future, please consider ticking "allow commits from maintainers" when creating a PR on github. That allows the reviewers of your PR to push changes to your branch, potentially shortcutting some of the back-and-forth for minor issues like typos, formatting, etc.

saraswatmks: @mpenkov thanks for the suggestion. I believe we are solid now on this PR. I will now make these changes in fasttext module as well.

mpenkov: Yes, please go ahead.

saraswatmks: @mpenkov I've done the changes in fasttext module as well. Please have a look.

gensim/models/fasttext.py
```diff
@@ -901,6 +903,23 @@ def train(self, sentences=None, corpus_file=None, total_examples=None, total_wor
             >>> model.train(sentences, total_examples=model.corpus_count, epochs=model.epochs)
 
         """
+
+        # Check if both sentences and corpus_file are None
```
mpenkov: Please remove these comments. They add no value to the source code.

mpenkov:
It looks like all comments have been addressed, so I'm merging this in.

@saraswatmks Congrats on your first contribution to gensim :1st_place_medal: Thank you!

PR #2472, opened 2019-05-01
Fix misleading Doc2Vec.docvecs comment

gojomo:
Existing doc-comment was confused & misleading, implying `Doc2Vec` handles word-senses by giving single word tokens different word-vectors in different contexts. (See <https://stackoverflow.com/questions/55939511/word-vectors-from-a-whole-doc2vec-model-vs-word-vectors-from-a-particular-docum/55941468#55941468> for an example confused user.) `Doc2Vec` doesn't do that, so this changes the comment to be matter-of-fact about accessing vectors via `.docvecs`.


mpenkov:
Looks like there's a trailing whitespace somewhere, and it's causing flake8 to choke. Can you please have a look and fix?

mpenkov:
There are still problems with spaces that fail the flake8 linter. Please have a look [here](https://travis-ci.org/RaRe-Technologies/gensim/jobs/527432869) for details.

gojomo:
I think it's fixed now.

Issue #2459, opened 2019-04-21
potential Doc2Vec feature: reverse inference, to synthesize doc/summary words


gojomo:
Motivated by the SO question: https://stackoverflow.com/questions/55768598/interpret-the-doc2vec-vectors-clusters-representation/55779049#55779049

`Doc2Vec` could plausibly have a function that's reverse-inference: take a doc-vector, return a (ranked) list of words most-predicted by that input vector. It'd work highly analogously to `Word2Vec.predict_output_word()`. Such a list of words *might* be useful as a sort-of summary or label for a doc-vector.

saraswatmks:
Hi @gojomo Is it yet to be done ? can I take it up?

gojomo:
@saraswatmks Yes, and further, you don't have to ask permission: a good PR submission will be reviewed/welcomed even without having declared your interest first. It's more important to "show working code" than "declare interest".

saraswatmks:
@gojomo Just to be sure, this function will do the following:
**Option 1:**
1. Takes a doc-vector (I will check for vector length as a sanity check)
2. Returns a list of tuples of words and scores which are most similar to the doc-vector
3. After doing dot product, the words will be returned from the whole corpus.

**Option 2:**
1. Instead of input vector, I ask for doc-id
2. Get the vector for that doc-id
3. Filter all the words which occurred in that doc-id and get vector of all those words
4. Do the dot product only on the filtered word vectors. This way we ensure that we return only those words which actually occurred in that doc-idd.

What do you think ?

gojomo:
Re: Option 1 – It's not "most similar" words which are needed here. Rather, it's "most predicted". The logic and behavior should be highly analogous to `Word2Vec.predict_output_word()`, except taking a single vector rather than a list-of-words.

(Option 2 is impossible, as the model includes no record of the words which occurred in that doc-id.)

gojomo:
See also similar #2152. 

Simply refactoring `predict_output_word` to the part that looks-up words, and the part that turns a raw-context-vector-into-ranked-predictions, might essentially create this doc2vec feature (supply the doc-vec as the raw-context-vector) almost effortlessly.

PR #2478, opened 2019-05-06
potential Doc2Vec feature: reverse inference in doc2vec 

saraswatmks:
This feature is in accordance to this issue: [2459](https://github.com/RaRe-Technologies/gensim/issues/2459)

It does the following:
1. `predict_output_word` function take a document vector and returns a list of tuples having word and probability score.
2. The reference for this implementation has been taken from `Word2Vec.predict_output_word()` function.

### DIFF
--- a/gensim/models/base_any2vec.py
+++ b/gensim/models/base_any2vec.py
@@ -39,7 +39,7 @@ import threading
 from six.moves import range
 from six import itervalues, string_types
 from gensim import matutils
-from numpy import float32 as REAL, ones, random, dtype, zeros
+from numpy import float32 as REAL, ones, random, dtype, zeros, sum as np_sum, exp, dot
 from types import GeneratorType
 from gensim.utils import deprecated
 import warnings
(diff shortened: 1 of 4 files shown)


gojomo:
Thanks, looks promising! To distinguish from `predict_output_word()` (in both function, and to prevent overiding it), I'd suggest a slightly different name. Perhaps `predict_words()`? (Similarly, the two methods might be able to share some logic.)

Have you been able to "eyeball" some results on test datasets, to get an idea if (as conjectured in the feature-request) they work as a sort of `summary`? For example:

* do the top-n words for a known-document doc-vector seem related to the text? 
* if using a `Doc2Vec` model with trained word-vectors (like `dm=1` or `dm=0, dbow_words=1` mode), do these predicted-words, or the most-similar-words-to-the-doc-vector, better overlap with the actual words of the text?
* similarly, are the top-N most-predicted or the top-N most-similar closers via a measure like "word' mover's distance" to the actual text?

saraswatmks:
@gojomo I appreciate your quick response. I haven't checked the sanity of results yet. I will do that now.

saraswatmks:
@gojomo I've renamed the function to `predict_words`. Also, the common logic I moved to parent class `BaseWordEmbeddingsModel`. I did the "eyeball" testing of results. They look fine. The similar words come from the same document whose document vector is provided, so yes there is significant overlap. The words overlap varies with different parameters `dm=1 or dm=0, dbow_words=1`. 

**P.S** - The pipeline is failing because scikit-learn no longer supports build on python 2.7

gensim/models/base_any2vec.py
```diff
@@ -1470,3 +1470,58 @@ def evaluate_word_pairs(self, pairs, delimiter='\t', restrict_vocab=300000,
 
         """
         return self.wv.evaluate_word_pairs(pairs, delimiter, restrict_vocab, case_insensitive, dummy4unknown)
+
+    def _find_similar_words(self, doc_vector=None, context_words_list=None, topn=None):
```
gojomo: These aren't necessarily "similar" words (as in the various other "similarity" methods), but *predicted* words, and the method name should reflect that. Separately, it's smewhat weird factoring to have a `doc_vector` input for a method in a Base class, and also weird to have two alternate parameters that are (not usually/sanely?) used together – that's a hint this *isn't* the truly shared forward-prop functionality – which would start with the shared path the method below does with `use_vector`.

gojomo:
As a speculative/unproven feature – even though I think it's a promising idea, or wouldn't have suggested it – I think there needs to be some published example(s) of its use/value to justify its integration. Ideally, this would be either a demo notebook (or extension of some existing notebook to show this functionality).

saraswatmks:
@gojomo  I agree with you. I will make the requested changes along with a demo notebook.

mpenkov:
@saraswatmks How's it going? Could you please give us an estimate of when you can finish this PR? Otherwise, we will have to close it due to lack of activity.

PR #2479, opened 2019-05-06
generator support in bm25

### LINKED ISSUE #2434: BM25 does not support generator as corpus
``__init__`` method in BM25 class takes a "list of list of str" as corpus instead of a generator. More precisely, this is what it looks like right now:

```
def __init__(self, corpus):
        """
        Parameters
        ----------
        corpus : list of list of str
            Given corpus.
        """
        self.corpus_size = len(corpus)
        self.avgdl = 0
        self.doc_freqs = []
        self.idf = {}
        self.doc_len = []
        self._initialize(corpus)
```
As we know, considering a generator instead of a list would be great to handle large collections of documents that do not fit in memory.

saraswatmks:
Fixes #2434.

This PR does the following:
1. Make bm25 accept generators as inputs.

### DIFF
--- a/gensim/summarization/bm25.py
+++ b/gensim/summarization/bm25.py
@@ -74,7 +74,7 @@ class BM25(object):
             Given corpus.
 
         """
-        self.corpus_size = len(corpus)
+        self.corpus_size = 0
         self.avgdl = 0
         self.doc_freqs = []
         self.idf = {}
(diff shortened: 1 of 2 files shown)


saraswatmks:
> Looks good to me

I discovered this is approved but not merged. Anything else to be done ?

mpenkov:
No, that's it. I've merged your changes. Thank you for reminding me about this PR, and for your contribution!

PR #2482, opened 2019-05-08
fix CHANGELOG.md

mpenkov:
- Update links to point to html_url instead of url
- Update update_changelog.py to use correct links and bold author name


mpenkov:
I'm going to merge this in the interest of time. @piskvorky if you find any more problems, please let me know.

PR #2491, opened 2019-05-11
Documentation changes for #2489

### LINKED ISSUE #2489: Document how to get complete document distribution from LDA
Document how to to get full document distribution from LDA, that would help in those cases where you are implementing your own distance algorithm and you need full  distribution. Personally I set  minimum_probability=0.0 but there might be other ways.

Antupis:
Documentation how I did get full topic distribution from LDA fixes #2489.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -82,6 +82,12 @@ A lot of parameters can be tuned to optimize training for your specific case
 
     >>> lda = LdaModel(common_corpus, num_topics=50, alpha='auto', eval_every=5)  # learn asymmetric alpha from data
 
+LDA returns complete topic probability distribution when you set minimum_probability to 0.0
+
+.. sourcecode:: pycon
+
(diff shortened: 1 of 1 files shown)


gensim/models/ldamodel.py
```diff
@@ -82,6 +82,12 @@
 
     >>> lda = LdaModel(common_corpus, num_topics=50, alpha='auto', eval_every=5)  # learn asymmetric alpha from data
 
+LDA returns complete topic probability distribution when you set minimum_probability to 0.0
```
piskvorky: This is too vague, and IMO more confusing than illuminating.

What is "complete"?
Why would you set `minimum_probability` manually? Or indeed *not* set it and keep the default?

Note that Gensim only returns *sparse* bow tuples: no explicit zeros, no dense vectors.

mpenkov:
Hi @Antupis , do you plan to finish this PR?

If you do, please respond to the comments above, and merge upstream/develop into your branch. Otherwise, we will have to close your PR.

PR #2493, opened 2019-05-15
Develop2

charmoniumQ:
Added option to specify Pyro nameserver


gensim/models/lsimodel.py
```diff
@@ -427,15 +430,20 @@ def __init__(self, corpus=None, num_topics=200, id2word=None, chunksize=20000,
...
-                )
-                self.dispatcher = dispatcher
-                self.numworkers = len(dispatcher.getworkers())
-                logger.info("using distributed version with %i workers", self.numworkers)
+                if ns_conf is None:
+                    ns_conf = {}
+
+                with utils.getNS(**ns_conf) as ns:
+                    from gensim.models.lsi_dispatcher import LSI_DISPATCHER_PREFIX
+                    dispatcher = Pyro4.Proxy(ns.list(prefix=LSI_DISPATCHER_PREFIX)[LSI_DISPATCHER_PREFIX])
```
mpenkov: Does this need to be a local variable? Is there any reason to not assign directly to self.dispatcher?

gensim/models/lsimodel.py
```diff
@@ -427,15 +430,20 @@ def __init__(self, corpus=None, num_topics=200, id2word=None, chunksize=20000,
...
-                    power_iters=self.power_iters, extra_samples=self.extra_samples, distributed=False, onepass=onepass
-                )
-                self.dispatcher = dispatcher
-                self.numworkers = len(dispatcher.getworkers())
-                logger.info("using distributed version with %i workers", self.numworkers)
+                if ns_conf is None:
+                    ns_conf = {}
+
+                with utils.getNS(**ns_conf) as ns:
+                    from gensim.models.lsi_dispatcher import LSI_DISPATCHER_PREFIX
```
mpenkov: Is there a reason to keep this import here in a try-catch? If not, please move it to the module scope.

gensim/test/test_ldamodel_distributed.py
```diff
@@ -0,0 +1,527 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2010 Radim Rehurek <radimrehurek@seznam.cz>
```
mpenkov: Is this new source? If yes, then the year should be 2019. If no, then please explain where this is coming from.

gensim/test/test_ldamodel_distributed.py
```diff
@@ -0,0 +1,527 @@
...
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+
+"""
+Automated tests for checking transformation algorithms (the models package).
```
mpenkov: Is this an accurate docstring? I think you're testing something else in this file.

gensim/test/test_ldamodel_distributed.py
```diff
@@ -0,0 +1,527 @@
...
+
+    def testAlphaAuto(self):
+        model1 = self.class_(corpus, id2word=dictionary, alpha='symmetric', passes=10)
+        modelauto = self.class_(corpus, id2word=dictionary, alpha='auto', passes=10)
+
+        # did we learn something?
+        self.assertFalse(all(np.equal(model1.alpha, modelauto.alpha)))
+
+    def testAlpha(self):
+        kwargs = dict(
```
mpenkov: This can fit one line. Please look for other instances in your changes where there are unnecessary newlines, and fix them too.

gensim/test/test_ldamodel_distributed.py
```diff
@@ -0,0 +1,527 @@
...
+            id2word=dictionary,
+            num_topics=2,
+            alpha=None
+        )
+        expected_shape = (2,)
+
+        # should not raise anything
+        self.class_(**kwargs)
+
+        kwargs['alpha'] = 'symmetric'
```
mpenkov: Are these blocks independent tests?

If they are, then you should break this test method up into multiple methods, here and everywhere.

gensim/test/test_ldamodel_distributed.py
```diff
@@ -0,0 +1,527 @@
...
+            for test_rhot in test_rhots:
+                model.update(self.corpus)
+
+                msg = ", ".join(str(x) for x in [passes, model.num_updates, model.state.numdocs])
+                self.assertAlmostEqual(final_rhot(model), test_rhot, msg=msg)
+
+            self.assertEqual(model.state.numdocs, len(corpus) * len(test_rhots))
+            self.assertEqual(model.num_updates, len(corpus) * len(test_rhots))
+
+    # def testTopicSeeding(self):
```
mpenkov: What is this commented out test?

gensim/test/test_ldamodel_distributed.py
```diff
@@ -0,0 +1,527 @@
...
+        # self.model.save(lda_3_0_1_fname)
+
+        # load a model saved using a 3.0.1 version of Gensim
+        model = self.class_.load(lda_3_0_1_fname)
+
+        # and test it on a predefined document
+        topics = model[test_doc]
+        self.assertTrue(np.allclose(expected_topics, topics))
+
+# endclass TestLdaModel
```
mpenkov: Please get rid of these endclass comments, we don't use them in gensim.

gensim/test/test_ldamodel_distributed.py
```diff
@@ -0,0 +1,527 @@
...
+                self.assertTrue(np.issubdtype(v, np.floating))
+
+    def testGetTopicTerms(self):
+        topic_terms = self.model.get_topic_terms(1)
+
+        for k, v in topic_terms:
+            self.assertTrue(isinstance(k, numbers.Integral))
+            self.assertTrue(np.issubdtype(v, np.floating))
+
+    def testGetDocumentTopics(self):
```
mpenkov: There's a lot of logic in this test method. That's not a good thing, because logic is prone to bugs, and you risk introducing bugs into actual test code (which itself does **not** get tested by anything).

For this reason, test code should be as simple as possible.

Furthermore, this method is far too long for a single test. There are many places where the test can fail. This is bad from a maintainer's (mine) perspective, because when a test fails, I'll have to work out what exactly it's doing and why it's failing. There's a lot to work out here.

Some suggestions:

- Break this test up into multiple smaller tests, if possible
- Extract common funcitonality into pure functions, and reuse them. For instance, much of your code here and in other tests performs type checking, e.g.
  1. Is the variable a list?
  2. Is each element of the list a tuple of some type?

You should also avoid loops in test code, if possible. Instead of iterating through a dictionary and checking items one by one, create a dictionary of expected values, and compare that to actual values in one assertion.

gensim/test/test_ldamodel_distributed.py
```diff
@@ -0,0 +1,527 @@
...
+        }
+        self.ns_process = Subprocess.Popen([sys.executable, '-m', 'Pyro4.naming', '--host=' + host, '--port=' + str(port)], env=env)
+        self.worker_processes = [
+            Subprocess.Popen([sys.executable, '-m', 'gensim.models.lsi_worker', '--host', host, '--port', str(port)], env=env)
+            for _ in range(2)
+        ]
+        self.dispatcher_process = Subprocess.Popen([sys.executable, '-m', 'gensim.models.lsi_dispatcher', '--host', host, '--port', str(port)], env=env)
+        self.model = self.class_(corpus, id2word=dictionary, num_topics=2, passes=100, distributed=True, ns_conf=dict(host=self.host, port=self.port))
+
+    def testTransform(self):
```
mpenkov: Use PEP8 method names, here and everywhere.

```suggestion
    def test_transform(self):
```

gensim/test/test_ldamodel_distributed.py
```diff
@@ -0,0 +1,527 @@
...
+        self.assertRaises(AssertionError, self.class_, **kwargs)
+
+        kwargs['eta'] = "gensim is cool"
+        self.assertRaises(ValueError, self.class_, **kwargs)
+
+        kwargs['eta'] = "asymmetric"
+        self.assertRaises(ValueError, self.class_, **kwargs)
+
+    def testTopTopics(self):
+        top_topics = self.model.top_topics(self.corpus)
```
mpenkov: Why are you checking the type of top_top topics and its keys, instead of checking the type and actual value? Why not do something like:

```
expected_top_topics = [
    ('foo', 'bar'),
    ...,
]
self.assertEqual(top_topics, expected_top_topics)
```

That would be a more robust and easy-to-read test, in my opinion.

The same thing applies to other tests as well.

mpenkov:
Hi @charmoniumQ , do you plan to finish this PR?

charmoniumQ:
Yes, but not in the near future.

PR #2505, opened 2019-05-29
Add build_vocab to poincare model

### LINKED ISSUE #2514: Threre is no support for online learning in poincare model
gensim supports online learning via `build_vocab` in fasttex, word2vec or other models.
But there is not `build_vocab` in poincare model.
Don't we need this mehotd or is it impossible to implement this feature ?

koiizukag:
This PR makes the following changes: to fix #2514 
- Add build_vocab to poincare model for online learning.
- add unit test.


koiizukag:
Add issue #2514 😃

koiizukag:
This is my first-time to commit this project.
If there is some problem in this PR, please teach me 👍

koiizukag:
@piskvorky Hi, please assign some reviewer

piskvorky:
@mpenkov will get to your PR eventually. We're busy finishing up a "Gensim docs" project, so this may take a while. Thanks for your patience @koiizukag .

gensim/models/poincare.py
```diff
@@ -167,47 +171,48 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
...
         """Load relations from the train data and build vocab."""
-        vocab = {}
-        index2word = []
-        all_relations = []  # List of all relation pairs
-        node_relations = defaultdict(set)  # Mapping from node index to its related node indices
+        old_index2word_len = len(self.kv.index2word)
 
         logger.info("loading relations from train data..")
-        for relation in self.train_data:
+        for relation in relations:
```
mpenkov: I think we risk tripping over a TypeError here. You're iterating over something that can be None.

Please double-check and add a unit test to cover this case, if necessary.

gensim/models/poincare.py
```diff
@@ -167,47 +171,48 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
         self._np_random = np_random.RandomState(seed)
         self.init_range = init_range
         self._loss_grad = None
-        self._load_relations()
-        self._init_embeddings()
+        self.build_vocab(train_data)
 
-    def _load_relations(self):
+    def build_vocab(self, relations=None, update=False):
```
mpenkov: You've made this a public method, so it needs a good docstring. Please look at other docstrings in the code for examples on how to write a good one.

gensim/models/poincare.py
```diff
@@ -167,47 +171,48 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
...
 
         logger.info("loading relations from train data..")
-        for relation in self.train_data:
+        for relation in relations:
             if len(relation) != 2:
                 raise ValueError('Relation pair "%s" should have exactly two items' % repr(relation))
             for item in relation:
-                if item in vocab:
-                    vocab[item].count += 1
+                if item in self.kv.vocab:
```
mpenkov: Ask for forgiveness instead of permission - it's more Pythonic.

```python
try:
   dict[item] += 1
except KeyError:
   dict[item] = 0
```

https://stackoverflow.com/questions/12265451/ask-forgiveness-not-permission-explain

piskvorky: -1: The original code (current version) looks clearer and better to me. Obfuscating the logic with unnecessary exception handling would be a regression IMO.

mpenkov: OK, up to you.

gensim/models/poincare.py
```diff
@@ -173,8 +173,34 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
...
+    def build_vocab(self, relations, update=False):
+        """Build vocabulary from a relations.
+        Each relations must be a tuples of unicode strings.
+
+        Parameters
+        ----------
+        relations : list of tuples
+            List of tuples of positive examples of the form (node_1_index, node_2_index).
+        update : bool
+            If true, the new nodes in `relations` will be added to model's vocab.
```
piskvorky: But `relations` doesn't contain nodes. Its description above says it contains "node indexes" (btw where does the user find those?).

Also, what happens if `False` (the default)?

gensim/models/poincare.py
koiizukag: Thank you for the nice review.
Because we must input some relations (not None) in build_vocab, I removed default None value here 👍

gensim/models/poincare.py
koiizukag: I modified the description of relations using __init__ description.
> But relations doesn't contain nodes. Its description above says it contains "node indexes" (btw where does the user find those?).

If `update=False`, the embeddings are initialized by random values.
(It means that the trained embeddings are cleaned.)
> Also, what happens if False (the default)?

piskvorky: OK, I don't really understand that, but that information should appear in the documentation.

koiizukag: I added update=False description in build_vocab 👍

gensim/models/poincare.py
```diff
@@ -168,47 +172,77 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
         self._np_random = np_random.RandomState(seed)
         self.init_range = init_range
         self._loss_grad = None
-        self._load_relations()
-        self._init_embeddings()
+        self.build_vocab(train_data)
+
+    def build_vocab(self, relations, update=False):
+        """Build vocabulary from relations.
```
mpenkov: More helpful:

```suggestion
        """Build the model's vocabulary from known relations.
```

gensim/models/poincare.py
```diff
@@ -168,47 +172,77 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
...
-        all_relations = []  # List of all relation pairs
-        node_relations = defaultdict(set)  # Mapping from node index to its related node indices
+        Parameters
+        ----------
+        relations : {iterable of (str, str), :class:`gensim.models.poincare.PoincareRelations`}
+            Iterable of relations, e.g. a list of tuples, or a :class:`gensim.models.poincare.PoincareRelations`
+            instance streaming from a file. Note that the relations are treated as ordered pairs,
+            i.e. a relation (a, b) does not imply the opposite relation (b, a). In case the relations are symmetric,
+            the data should contain both relations (a, b) and (b, a).
+        update : bool
```
mpenkov:
```suggestion
        update : bool, optional
```

gensim/models/poincare.py
```diff
@@ -168,47 +172,77 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
...
+            If true, new nodes's embeddings are initialized.
```
mpenkov:
```suggestion
            If true, only new nodes's embeddings are initialized.
            Use this when the model already has an existing vocabulary and you want to update it.
```

gensim/models/poincare.py
```diff
@@ -168,47 +172,77 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
...
+            If false, all node's embeddings are initialized.
```
mpenkov:
```suggestion
            If false, all node's embeddings are initialized.
            Use this when you're creating a new vocabulary from scratch.
```

gensim/models/poincare.py
```diff
@@ -168,47 +172,77 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
...
+        --------
+        Train a model and update vocab for online training:
+
+        .. sourcecode:: pycon
+
+            >>> from gensim.models.poincare import PoincareModel
+            >>> relations_1 = [('kangaroo', 'marsupial'), ('kangaroo', 'mammal')]
+            >>> relations_2 = [('striped_skunk', 'mammal')]
+            >>>
+            >>> model = PoincareModel(relations_1, negative=1)
```
mpenkov: Let's make it **really** obvious what we're actually doing:

```suggestion
            >>> # train a new model from initial data
            >>> model = PoincareModel(initial_relations, negative=1)
```

gensim/models/poincare.py
```diff
@@ -168,47 +172,77 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
...
+            >>> model.train(epochs=50)
+            >>>
+            >>> model.build_vocab(relations_2, update=True)
```
mpenkov: It's more helpful to move the definition of online_relations here, because it's used immediately after its defined.

```suggestion
            >>> # online training: update the vocabulary and continue training
            >>> online_relations = [('striped_skunk', 'mammal')]
            >>> model.build_vocab(online_relations, update=True)
```

gensim/models/poincare.py
```diff
@@ -168,47 +172,77 @@ def __init__(self, train_data, size=50, alpha=0.1, negative=10, workers=1, epsil
...
+            If false, all node's embeddings are initialized.
+
+        Examples
+        --------
+        Train a model and update vocab for online training:
+
+        .. sourcecode:: pycon
+
+            >>> from gensim.models.poincare import PoincareModel
+            >>> relations_1 = [('kangaroo', 'marsupial'), ('kangaroo', 'mammal')]
```
mpenkov: Let's use more helpful variable names:

```suggestion
            >>> initial_relations = [('kangaroo', 'marsupial'), ('kangaroo', 'mammal')]
```

koiizukag:
Thanks for your modification. LGTM

koiizukag:
@mpenkov Is there anything to be done next❓

koiizukag:
@piskvorky I think this PR is ready to merge. Please check here 🙏

piskvorky:
@mpenkov will be with you shortly, there's a lot going on. Thanks for your patience.

mpenkov:
Merged. @koiizukag Congratulations on your first contribution to gensim :1st_place_medal: Thank you!

PR #2513, opened 2019-05-30
Correct "Market Matrix" to "Matrix Market".

Shooter23:
Small change only to documentation.


piskvorky:
Nicely spotted! Thanks.

PR #2516, opened 2019-05-31
Expose max_final_vocab parameter in FastText constructor

scribu:
I think it was just an oversight in #1915.

### DIFF
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -461,7 +461,7 @@ class FastText(BaseWordEmbeddingsModel):
 
     """
     def __init__(self, sentences=None, corpus_file=None, sg=0, hs=0, size=100, alpha=0.025, window=5, min_count=5,
-                 max_vocab_size=None, word_ngrams=1, sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
+                 max_vocab_size=None, max_final_vocab=None, word_ngrams=1, sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
                  negative=5, ns_exponent=0.75, cbow_mean=1, hashfxn=hash, iter=5, null_word=0, min_n=3, max_n=6,
                  sorted_vocab=1, bucket=2000000, trim_rule=None, batch_words=MAX_WORDS_IN_BATCH, callbacks=(),
                  compatible_hash=True):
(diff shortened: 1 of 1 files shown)


gensim/models/fasttext.py
```diff
@@ -460,11 +460,38 @@ class FastText(BaseWordEmbeddingsModel):
...
+            sentences=None,
+            corpus_file=None,
+            sg=0,
+            hs=0,
+            size=100,
+            alpha=0.025,
+            window=5,
+            min_count=5,
+            max_vocab_size=None,
+            max_final_vocab=None,
```
mpenkov: Move this parameter to the end of the list. We don't want to break backward compatibility for people who insist on passing keyword arguments as unnamed.

Also, please avoid mixing in innocuous formatting changes with actual functionality, as it makes your code more difficult to review. Roll back this formatting change in the constructor.

gensim/models/fasttext.py
```diff
@@ -507,6 +534,10 @@ def __init__(self, sentences=None, corpus_file=None, sg=0, hs=0, size=100, alpha
             Limits the RAM during vocabulary building; if there are more unique
             words than this, then prune the infrequent ones. Every 10 million word types need about 1GB of RAM.
             Set to `None` for no limit.
+        max_final_vocab : int, optional
```
mpenkov: I think it's worth explaining the difference between this and `max_vocab_size`. How do they interact together?

gensim/models/fasttext.py
```diff
@@ -589,8 +620,14 @@ def __init__(self, sentences=None, corpus_file=None, sg=0, hs=0, size=100, alpha
...
         self.vocabulary = FastTextVocab(
-            max_vocab_size=max_vocab_size, min_count=min_count, sample=sample,
-            sorted_vocab=bool(sorted_vocab), null_word=null_word, ns_exponent=ns_exponent)
+            max_vocab_size=max_vocab_size,
+            max_final_vocab=max_final_vocab,
+            min_count=min_count,
+            sample=sample,
+            sorted_vocab=bool(sorted_vocab),
+            null_word=null_word,
+            ns_exponent=ns_exponent
```
mpenkov: Future-proof your code. If we want to add more parameters to end of this list, we won't have to change the last line.

```suggestion
            ns_exponent=ns_exponent,
```

mpenkov:
Continued in #2867 (I don't have permissions to push changes to this branch).

PR #2518, opened 2019-06-03
Typo fix

Guitaricet:
Minor typo in the docs

`load_facebook_vectors` is not a method of FastText class

### DIFF
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -190,7 +190,7 @@ You may continue training them on new data:
     True
 
 If you do not intend to continue training the model, consider using the
-:func:`gensim.models.FastText.load_facebook_vectors` function instead.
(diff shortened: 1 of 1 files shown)


gensim/models/fasttext.py
```diff
@@ -190,7 +190,7 @@
     True
 
 If you do not intend to continue training the model, consider using the
-:func:`gensim.models.FastText.load_facebook_vectors` function instead.
+:func:`gensim.models.fastText.load_facebook_vectors` function instead.
```
mpenkov: Should it be

```suggestion
:func:`gensim.models.fasttext.load_facebook_vectors` function instead.
```

instead?

Guitaricet: RIght

mpenkov:
Thanks for fixing this up. Congrats on your first contribution :)

PR #2528, opened 2019-06-14
Fix smart_open deprecation warning in index loading

itayB:

PR #2530, opened 2019-06-15
Fix smart_open deprecation warning globally

### LINKED ISSUE #2531: MmWriter raises UserWarning on __init__
#### Problem description

When using MmCorpus, the MmWriter class raises a UserWarning on `__init__` because it is using a deprecated function of smart_open.

```python
MmCorpus.serialize('corpus.mm', some_corpus)

[2019-06-17 13:20:19,741] [INFO] {save_corpus:122} storing corpus in Matrix Market format to corpus.mm
/home/rob/venv/lib/python3.6/site-packages/smart_open/smart_open_lib.py:398: UserWarning: This function is deprecated, use smart_open.open instead. See the migration notes for details: https://github.com/RaRe-Technologies/smart_open/blob/master/README.rst#migrating-to-the-new-open-function
  'See the migration notes for details: %s' % _MIGRATION_NOTES_URL
```

#### Versions

```
Linux-4.15.0-51-generic-x86_64-with-Ubuntu-18.04-bionic
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0]
NumPy 1.16.3
SciPy 1.3.0
gensim 3.7.3
FAST_VERSION 1
```

itayB:
Continued from #2528.

Fixes #2531, closes #2528.


gensim/corpora/_mmreader.c
```diff
@@ -4155,7 +4155,7 @@ static PyObject *__pyx_pf_6gensim_7corpora_9_mmreader_8MmReader_11docbyoffset(st
  *         if offset == -1:
  *             return []             # <<<<<<<<<<<<<<
  *         if isinstance(self.input, string_types):
- *             fin, close_fin = utils.smart_open(self.input), True
+ *             fin, close_fin = utils.open(self.input), True
```
piskvorky: You should never modify auto-generated files.

(here and everywhere else)

gensim/corpora/_mmreader.pyx
```diff
@@ -188,7 +188,7 @@ cdef class MmReader(object):
         if offset == -1:
             return []
         if isinstance(self.input, string_types):
-            fin, close_fin = utils.smart_open(self.input), True
+            fin, close_fin = utils.open(self.input), True
```
piskvorky: This doesn't seem compatible because the old default mode was `rb`. See the [Migration guide](https://github.com/RaRe-Technologies/smart_open#migrating-to-the-new-open-function).

(here and everywhere else)

gensim/utils.py
```diff
@@ -113,7 +113,7 @@ def _synchronizer(self, *args, **kwargs):
 
 
 def file_or_filename(input):
-    """Open a filename for reading with `smart_open`, or seek to the beginning if `input` is an already open file.
```
piskvorky: This change is also wrong.

mpenkov:
I agree with @piskvorky that blindly find-replacing is not the way to go here. Let's come up with a more intelligent approach.

You'll see most code accesses smart_open indirectly via the utils module. Perhaps we can contain the changes to that module? That will reduce the risk and make things easier to test.

itayB:
I can create `smart_open` function in utils module. This function will call to the new `open` so we won't have to change in so much places. WDYT?
I will try to do later on (I'm out now)

piskvorky:
Nah, I'm for replacing it everywhere, not only in `utils`. Just take more care with the migration: not every `smart_open` string should become `open`.

itayB:
OK, I'll fix it later on and keep you posted

gensim/corpora/_mmreader.pyx
```diff
@@ -188,7 +188,7 @@ cdef class MmReader(object):
         if offset == -1:
             return []
         if isinstance(self.input, string_types):
-            fin, close_fin = utils.smart_open(self.input), True
+            fin, close_fin = utils.open(self.input, 'rb'), True
```
itayB: @piskvorky Hi!
I guess that this change should update `_mmreader.c`  /  `_mmreader.so` right?
How do I trigger this compilation?
I tried few things with no lack, `c` file changed but the `so` didn't, and I'm still getting:
```
  File "gensim/corpora/_mmreader.pyx", line 191, in gensim.corpora._mmreader.MmReader.docbyoffset
AttributeError: module 'gensim.utils' has no attribute 'smart_open'
```

piskvorky: CC @mpenkov what's the process regenerating the compiled modules? Force `pip` re-installation of `gensim`?

mpenkov: Running cython on the affected files will also work, if gensim in set up in develop mode.

itayB: Thanks guys, I generated the file and the tests passed. Can you review again, please?

mpenkov: Thank you. I'll have a look closer to the weekend.

gensim/utils.py
```diff
@@ -1348,7 +1348,7 @@ def smart_extension(fname, ext):
 
 
 def pickle(obj, fname, protocol=2):
-    """Pickle object `obj` to file `fname`, using smart_open so that `fname` can be on S3, HDFS, compressed etc.
+    """Pickle object `obj` to file `fname`, using open so that `fname` can be on S3, HDFS, compressed etc.
```
piskvorky: Wrong replacement.

itayB: I changed the usage in documented function, do you want me to leave the comments with `smart_open`? same for the `unpickle` comment below?

piskvorky: Yeah, the replacements need to make sense.

`Pickle X using open so that it can be on S3, HDFS, compressed etc.` is a totally confusing comment. `open` cannot do anything about S3, HDFS, etc (the `smart_open` library can).

itayB: ok, no problem, reverting those comments to original

itayB: @piskvorky fixed

gensim/utils.py
```diff
@@ -1378,8 +1378,8 @@ def unpickle(fname):
         Python object loaded from `fname`.
 
     """
-    with smart_open(fname, 'rb') as f:
-        # Because of loading from S3 load can't be used (missing readline in smart_open)
+    with open(fname, 'rb') as f:
+        # Because of loading from S3 load can't be used (missing readline in open)
```
piskvorky: This comment also incorrect.

Unrelated: @mpenkov is this comment true? Is `smart_open` missing some methods? I don't understand this code block… what's up with the `latin1`?

piskvorky: OK, looking at the [Python pickle docs](https://docs.python.org/3/library/pickle.html#pickle.load), this seems related to compatibility for loading Python 2 pickles in Python 3.

```
Optional keyword arguments are fix_imports, encoding and errors, which are used to control compatibility support for pickle stream generated by Python 2. If fix_imports is true, pickle will try to map the old Python 2 names to the new names used in Python 3. The encoding and errors tell pickle how to decode 8-bit string instances pickled by Python 2; these default to ‘ASCII’ and ‘strict’, respectively. The encoding can be ‘bytes’ to read these 8-bit string instances as bytes objects. Using encoding='latin1' is required for unpickling NumPy arrays and instances of datetime, date and time pickled by Python 2.
```

Still not sure what this comment line is trying to say about S3 and smart_open though.

itayB: fixed

gensim/models/deprecated/old_saveload.py
```diff
@@ -367,7 +367,7 @@ def unpickle(fname):
         Python object loaded from `fname`.
 
     """
-    with smart_open(fname, 'rb') as f:
+    with open(fname, 'rb') as f:
```
mpenkov: smart_open **does** have readline now, so the below comment is inaccurate.

Looks like the only reason to read the whole thing into memory is that replace call.

gensim/models/poincare.py
```diff
@@ -53,7 +53,7 @@
 from scipy.stats import spearmanr
 from six import string_types
 from six.moves import zip, range
-from smart_open import smart_open
+from smart_open import open
```
mpenkov: Why are we using two separate methods to open files?

1. smart_open.open
2. utils.open

?

piskvorky: AFAICS `utils.open` is an alias of `smart_open.open` from that module.

I agree there's no point using that: just use the local `from smart_open import open` version.

@itayB can you please fix that too, while we're cleaning up?

gensim/models/deprecated/old_saveload.py
itayB: removed this comment

gensim/models/poincare.py
itayB: fixed - removed this import and change other usages to `utils.open`

piskvorky: IMO `better use the local "from smart_open import open" version, consistently`.
But the other way around works too, I guess :)

itayB:
@mpenkov @piskvorky is there anything else you want me to do in this PR?

piskvorky:
Thanks a lot @itayB , let's wait for @mpenkov 's review.

gensim/models/deprecated/old_saveload.py
```diff
@@ -31,7 +31,7 @@
 
 from six import iteritems
 
-from smart_open import smart_open
+from smart_open import open
```
mpenkov: I thought we decided [here](https://github.com/RaRe-Technologies/gensim/pull/2530#discussion_r296324334) to use utils.open consistently. Why are we still using this import?

gensim/models/doc2vec.py
```diff
@@ -1505,7 +1505,7 @@ def __iter__(self):
             fname = os.path.join(self.dirname, fname)
             if not os.path.isfile(fname):
                 continue
-            for item_no, line in enumerate(utils.smart_open(fname)):
+            for item_no, line in enumerate(utils.open(fname, 'rb')):
```
mpenkov: Let's be consistent:

```python
# importing
from gensim import utils

# opening
with utils.open(...) as fin:
    ...
```

Here you've done "from gensim.utils import open" but instead of using it, you've used "utils.open" instead. That's a bit confusing.

itayB: no problem, I'll change all occurrences and update shortly

gensim/scripts/segment_wiki.py
```diff
@@ -63,7 +63,7 @@
 from functools import partial
 
 from gensim.corpora.wikicorpus import IGNORED_NAMESPACES, WikiCorpus, filter_wiki, find_interlinks, get_namespace, utils
```
itayB: @mpenkov here there is already import of other `utils`. I'll use `from gensim.utils import open` here - is that ok?

gensim/models/doc2vec.py
itayB: @mpenkov done. Can you please review again?

gensim/models/poincare.py
itayB: @piskvorky I changed everything to work with `utils.open` - should be consistented now

gensim/models/deprecated/doc2vec.py
```diff
@@ -992,7 +992,7 @@ def __iter__(self):
             fname = os.path.join(self.dirname, fname)
             if not os.path.isfile(fname):
                 continue
-            for item_no, line in enumerate(utils.smart_open(fname)):
+            for item_no, line in enumerate(utils.open(fname, 'rb')):
```
mpenkov: Can we put this inside a context manager? We're opening a file pointer and leaving it dangling here, it's poor practice.

gensim/models/deprecated/keyedvectors.py
```diff
@@ -934,7 +934,7 @@ def accuracy(self, questions, restrict_vocab=30000, most_similar=most_similar, c
         ok_vocab = {w.upper(): v for w, v in reversed(ok_vocab)} if case_insensitive else dict(ok_vocab)
 
         sections, section = [], None
-        for line_no, line in enumerate(utils.smart_open(questions)):
+        for line_no, line in enumerate(utils.open(questions, 'rb')):
```
mpenkov: Context manager

gensim/models/deprecated/keyedvectors.py
```diff
@@ -1030,7 +1030,7 @@ def evaluate_word_pairs(self, pairs, delimiter='\t', restrict_vocab=300000,
         original_vocab = self.vocab
         self.vocab = ok_vocab
 
-        for line_no, line in enumerate(utils.smart_open(pairs)):
+        for line_no, line in enumerate(utils.open(pairs, 'rb')):
```
mpenkov: Context manager

gensim/models/deprecated/word2vec.py
```diff
@@ -1699,7 +1699,7 @@ def __iter__(self):
             fname = os.path.join(self.dirname, fname)
             if not os.path.isfile(fname):
                 continue
-            for line in utils.smart_open(fname):
+            for line in utils.open(fname, 'rb'):
```
mpenkov: Context manager

gensim/models/doc2vec.py
mpenkov: Looking good. Can you please wrap the open in a context manager?

gensim/models/keyedvectors.py
```diff
@@ -1062,7 +1062,7 @@ def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensi
         logger.info("Evaluating word analogies for top %i words in the model on %s", restrict_vocab, analogies)
         sections, section = [], None
         quadruplets_no = 0
-        for line_no, line in enumerate(utils.smart_open(analogies)):
+        for line_no, line in enumerate(utils.open(analogies, 'rb')):
```
mpenkov: Context manager

gensim/scripts/segment_wiki.py
mpenkov: Let's use namespaces:

```
import gensim.utils
with gensim.utils.open(...) as fin:
    ...
```

gensim/test/test_word2vec.py
```diff
@@ -1069,30 +1069,30 @@ def testLineSentenceWorksWithFilename(self):
...
                 sentences = word2vec.LineSentence(fin)
                 for words in sentences:
                     self.assertEqual(words, utils.to_unicode(orig.readline()).split())
 
     def testPathLineSentences(self):
         """Does PathLineSentences work with a path argument?"""
-        with utils.smart_open(os.path.join(datapath('PathLineSentences'), '1.txt')) as orig1,\
-        utils.smart_open(os.path.join(datapath('PathLineSentences'), '2.txt.bz2')) as orig2:
+        with utils.open(os.path.join(datapath('PathLineSentences'), '1.txt'), 'rb') as orig1,\
+        utils.open(os.path.join(datapath('PathLineSentences'), '2.txt.bz2'), 'rb') as orig2:
```
mpenkov: I think in this case it makes more sense to nest:

```python
with utils.open(...) as orig1:
    with utils.open(...) as orig2:
        ...
```

instead of messing with line breaks.

gensim/scripts/segment_wiki.py
itayB: done

gensim/models/doc2vec.py
itayB: done

gensim/models/deprecated/doc2vec.py
itayB: done

gensim/models/deprecated/keyedvectors.py
itayB: done

gensim/models/deprecated/word2vec.py
itayB: done

gensim/test/test_word2vec.py
itayB: done

gensim/models/keyedvectors.py
itayB: done, changed almost everywhere (except 2 places that handle the closing explicitly)

mpenkov:
In the future, please avoid squashing everything together and force-pushing during a review. It makes it difficult for the reviewer to manage the change set. Right now, it's impossible to distinguish code that I've reviewed already, and code that you've pushed since the previous review.

itayB:
> In the future, please avoid squashing everything together and force-pushing during a review. It makes it difficult for the reviewer to manage the change set. Right now, it's impossible to distinguish code that I've reviewed already, and new code that you've pushed since the previous review.

No problem, sorry about that

mpenkov:
OK, I think this is ready to merge. @piskvorky Do you have any objections?

piskvorky:
No objections, nice work @itayB ! This Gensim cleanup was long overdue.

itayB:
> No objections, nice work @itayB ! This Gensim cleanup was long overdue.

thanks @piskvorky !
I think that there's still 1 change requested that is waiting for your approval :-)

itayB:
Hi @mpenkov @piskvorky, I just want to make sure; Is there anything that you expect me to do or should I just wait for you to merge this?

mpenkov:
Please wait for us to merge. Thank you for your contribution and your patience.

mpenkov:
OK, merged. @itayB Thank you for taking care of this!

PR #2533, opened 2019-06-18
Update self.vector_norms after adding new vectors to gensim.models.keyedvectors.WordEmbeddingKeyedVectors

seungwooson:
The original function of `init_sim` is checking if there are no objects in `self.vectors_norm`, then it is calculated and saved by L2-normalizing `self.vectors`. 

However, if there are any newly added vectors in `self.vectors` that have not been pre-computed into L2-normalized vectors, we should pre-compute them then add to the `self.vectors_norm`.

This is extremely useful, because although KeyedVectors have ability to add new vectors (such as inferred vectors) and their tags, they cannot execute `self.most_similar()` method for newly added vectors because the method requires lookup of `self.vectors_norm`.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -1348,7 +1348,7 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
         if getattr(self, 'vectors_norm', None) is None or replace:
             logger.info("precomputing L2-norms of word weight vectors")
             self.vectors_norm = _l2_norm(self.vectors, replace=replace)
-        elif (len(self.vectors_norm) == len(self.vectors)):  # if all of the vectors are precomputed
+        elif len(self.vectors_norm) == len(self.vectors):  # all vectors are precomputed
             pass
         else:  # when some newly added vectors in self.vectors are not precomputed
             logger.info("adding L2-norm vectors for new documents")


gensim/models/keyedvectors.py
```diff
@@ -1348,6 +1348,12 @@ def init_sims(self, replace=False):
         if getattr(self, 'vectors_norm', None) is None or replace:
             logger.info("precomputing L2-norms of word weight vectors")
             self.vectors_norm = _l2_norm(self.vectors, replace=replace)
+        elif (len(self.vectors_norm) == len(self.vectors)):  # if all of the vectors are precomputed
```
mpenkov: Get rid of unnecessary parentheses.

```suggestion
        elif len(self.vectors_norm) == len(self.vectors):  # if all of the vectors are precomputed
```

seungwooson:
> Thank you for your contribution! It looks good, but please add some tests that demonstrate why your changes are necessary/useful.
> 
> These tests should fail without your changes, and pass once your changes are made.
> 
> Thank you!

Hi @mpenkov, thanks for your feedback! Where would you like me to add the test cases and their outputs? I can demo using a jupyter notebook but not sure where I should attach it.

mpenkov:
Is it possible to write unit tests to cover the new funcitonality?

mpenkov:
@seungwooson Hi, just thought I'd check in about this PR. Are you able to finish it?

mpenkov:
@seungwooson Ping. Are you able to finish this PR?

seungwooson:
> @seungwooson Ping. Are you able to finish this PR?

I'll finish the PR in the upcoming week.

mpenkov:
@seungwooson How's it going? Are you able to finish this PR?

gensim/models/keyedvectors.py
```diff
@@ -1348,6 +1348,12 @@ def init_sims(self, replace=False):
         if getattr(self, 'vectors_norm', None) is None or replace:
             logger.info("precomputing L2-norms of word weight vectors")
             self.vectors_norm = _l2_norm(self.vectors, replace=replace)
+        elif len(self.vectors_norm) == len(self.vectors):  # all vectors are precomputed
+            pass
+        else:  # when some newly added vectors in self.vectors are not precomputed
+            logger.info("adding L2-norm vectors for new documents")
+            diff = len(self.vectors) - len(self.vectors_norm)
```
mpenkov, line 1355: Can `diff` be negative here?

gojomo:
WordEmbeddingsKeyedVectors, as a class, goes away in #2698, and so does the separate cached `vectors_norms` property. The ways there for getting norms (of a vector or all vectors) will probably 'just work' for any purely-functional unit test created (though a test that reaches into the implementation particulars might break). So I suggest just pulling out the top-level motivation for this, and verifying that it's met by #2698, so this can be set aside.

slavaGanzin:
Awesome work guys! 1.5 years to review 4 lines of code.

piskvorky:
Doesn't seem like @seungwooson is able to verify as per @gojomo 's suggestion. So let me close this as obsolete / not needed, the whole concept of `vectors_norm` has become irrelevant.

PR #2547, opened 2019-07-02
fix python version check

charsyam:
python version check is wrong now.
it uses sys.version_info[:1] but it returns tuples not Integer.
so it always fails.

I just fixed it

Thanks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ import warnings
 from setuptools import setup, find_packages, Extension
 from setuptools.command.build_ext import build_ext
 
-if sys.version_info[:2] < (2, 7) or (sys.version_info[:1] == 3 and sys.version_info[:2] < (3, 5)):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -17,7 +17,7 @@
 from setuptools import setup, find_packages, Extension
 from setuptools.command.build_ext import build_ext
 
-if sys.version_info[:2] < (2, 7) or (sys.version_info[:1] == 3 and sys.version_info[:2] < (3, 5)):
+if sys.version_info[:2] < (2, 7) or (sys.version_info[:1][0] == 3 and sys.version_info[:2] < (3, 5)):
```
mpenkov: Isn't this simpler and easier to read?

```suggestion
if sys.version_info[:2] < (2, 7) or ((3, 0) <= sys.version_info[:2] < (3, 5)):
```

charsyam:
@mpenkov Thanks, It is better to read

mpenkov:
@charsyam OK, merged. Congrats on your first contribution to gensim :1st_place_medal: Thank you!

PR #2552, opened 2019-07-04
Fix bug where a module import set up logging, pin doctools for Py2

piskvorky:
- Module imports must never issue logging events, because logging is not set up in the main app yet. The logging event triggered a default (unwanted) config init.
- `warnings.warn` better, but in this case, we can more simply re-use the existing functionality from `gensim.utils`

The bug manifested itself as an unexpected log output on Gensim import:

```
 2019-07-04 23:22:48,902 [MainProcess/98263] [INFO] gensim.summarization.textcleaner:37: 'pattern' package not found; tag filters are not available for English
```

This may be a deeper issue with the `summarization` subpackage doing weird stuff. This PR is only a surface stop-gap, I didn't look any deeper.


mpenkov:
Looks like the tests are failing. Did you forget to check some changes in? Or is this a work in progress?

piskvorky:
I didn't run it, just edited the faulty file on Github, to get rid of the logging setup.

Edited some more now, hopefully that fixes it.

mpenkov:
Tests are still failing, but the cause seems to be some Py2 bs with one of our dependencies.

piskvorky:
According to https://github.com/pywbem/pywbem/issues/1788, pinning `docutils` to 0.14 helps… not sure where to do that though. Is `docutils` some internal dependency of Sphinx?

mpenkov:
Finally got around to looking at this. Pinned doctools in setup.py, let's see if it helps.

PR #2589, opened 2019-09-03
Optimize Poincare model training

koiizukag:
Reduce computational complexity O(N^2) => O(N).
This part is the bottleneck of the overall training phase.

### DIFF
--- a/gensim/models/poincare.py
+++ b/gensim/models/poincare.py
@@ -568,7 +568,8 @@ class PoincareModel(utils.SaveLoad):
         """
         counts = Counter(node_indices)
         node_dict = defaultdict(list)
-        [node_dict[node_index].append(i) for i, node_index in enumerate(node_indices)]
+        for i, node_index in enumerate(node_indices):
+            node_dict[node_index].append(i)
         for node_index, count in counts.items():
             if count == 1:
                 continue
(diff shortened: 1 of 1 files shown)


gensim/models/poincare.py
```diff
@@ -567,10 +567,12 @@ def _handle_duplicates(vector_updates, node_indices):
 
         """
         counts = Counter(node_indices)
+        node_dict = defaultdict(list)
+        [node_dict[node_index].append(i) for i, node_index in enumerate(node_indices)]
```
piskvorky: Rewrite as loop (unpythonic construct).

koiizukag:
Thanks @piskvorky . 

I found this insufficient implementation via python profiler.
- https://github.com/rkern/line_profiler
- https://github.com/what-studio/profiling

If we use multiprocessing in independent for loop, we can make this model faster.
ex: https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/poincare.py#L545
Also, it is better to use Cython like other gensim models.

mpenkov:
Thank you for your contribution @koiizukag.

@piskvorky This looks good to merge to me. Is there anything left on this PR that you want to do before we merge this?

piskvorky:
LGTM. My only question was around fixing other similar issues in Poincare, since we (@koiizukag ) are at it. But nothing blocking.

PR #2601, opened 2019-09-10
Disable Py2.7 builds under Travis, CircleCI and AppVeyor

### LINKED ISSUE #2548: Drop Py2 support
An increasing number of our dependencies are dropping support for Py2. Numpy and pytest are the more visible examples, but there are others. Dealing with this requires identifying such packages (usually, through failed builds) and working around the problem. This hits us hard on two fronts:

1. It consumes precious maintainer time (e.g. https://github.com/RaRe-Technologies/gensim/pull/2546)
2. It blocks other work: we cannot reliably merge pull requests while our builds are failing

I recommend that we prioritize removing Py2 support. To get there:

1. Get builds working again (https://github.com/RaRe-Technologies/gensim/pull/2546 should finally work, after several days of poking around)
2. Merge any PRs that are ready to merge
3. Pause work on the remaining PRs
4. Release 3.9.0, the next minor version, and the last version to support Py2
5. Aggressively remove Py2 support
6. Release 4.0.0, explicitly mention that it's Py3 only

I think the above plan is a reasonable compromise:

- People who are stuck with Py2 for whatever reason can still continue to use 3.9.0 (or any previous version)
- Development and maintenance will continue with Py3 only

@piskvorky @menshikh-iv Thoughts?

mpenkov:
- Fix #2548 
- Make `annoy` an optional dependency (keep it out of setup.py)
- Introduce `ci/install_wheels.py`


.travis.yml
```diff
@@ -12,15 +12,9 @@ language: python
 
 matrix:
   include:
-    - python: '2.7'
-      env: TOXENV="flake8,flake8-docs"
-
     - python: '3.6'
       env: TOXENV="flake8,flake8-docs"
 
-    - python: '2.7'
```
menshikh-iv: requires similar changes in https://github.com/MacPython/gensim-wheels/blob/master/.travis.yml and https://github.com/MacPython/gensim-wheels/blob/master/appveyor.yml (btw, don't forget to update `multibuild` first too)

setup.py
```diff
@@ -282,8 +282,8 @@ def finalize_options(self):
     'sphinx',
     'sphinxcontrib-napoleon',
     'plotly',
-    'pattern <= 2.6',
     'sphinxcontrib.programoutput',
+    'annoy',
```
menshikh-iv: why? If I remember correctly, this doesn't work on Mac (or already fixed?)

setup.py
```diff
@@ -282,8 +282,8 @@ def finalize_options(self):
     'sphinx',
     'sphinxcontrib-napoleon',
     'plotly',
-    'pattern <= 2.6',
```
menshikh-iv: why?

tox.ini
```diff
@@ -30,8 +30,6 @@ deps =
     py37: numpy==1.14.5
     py37: scipy==1.1.0
 
-    py27: numpy==1.11.3
-    py27: scipy==0.18.1
```
menshikh-iv: need also to remove ` --trusted-host 28daf2247a33e...` I guess too

tox.ini
```diff
@@ -73,15 +72,15 @@ commands = flake8-rst gensim/ docs/ {posargs}
 
 
 [testenv:compile]
-basepython = python2
+basepython = python3.7
```
menshikh-iv: are you sure about base version? Maybe just `python3`? `python3.7` not really frequently available on machine

tox.ini
```diff
@@ -73,15 +72,15 @@ commands = flake8-rst gensim/ docs/ {posargs}
...
 recreate = True
 
-deps = numpy==1.11.3
+deps = numpy==1.14.5
```
menshikh-iv: maybe update numpy/scipy versions everywhere? (incl. deps above and `setup.py`)

This will allow installing numpy from pypi (without any `--trusted-host`)

btw, don't forget to mirror this change to `gensim-wheels` and `gensim-forge`)

setup.py
mpenkov: Because of this: https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/similarities/index.py#L50

setup.py
mpenkov: I can't find this in PyPI anywhere.

https://pypi.org/search/?q=pattern

tox.ini
mpenkov: What is that stuff? I'm not particularly familiar with them

setup.py
menshikh-iv: hmm, please check that on gensim-wheels (osx), I guess this can be an issue (and probably better to move that into `__init__` function of indexer, instead of module level

setup.py
menshikh-iv: btw, I'm +1 for fully drop it from gensim at all

tox.ini
menshikh-iv: Let's discuss that on telegram

tox.ini
mpenkov: I moved this to install_wheels.py script

PR #2605, opened 2019-09-19
Fix usage of base_dir instead of BASE_DIR in _load_info in downloader.

movb:
Setting `BASE_DIR` in downloader module does not affect `base_dir`, used in `_load_info` function.


mpenkov:
Merged. Congrats on your first contribution!

PR #2606, opened 2019-09-20
Refactor bm25 to include model parametrization

### LINKED ISSUE #2597: It is not possible to set values for BM25 model hyperparameters (K1, B, and EPSILON)
#### Problem description

I'm using the BM25 implementation and there is currently no way of defining its hyperparameters.

It would be nice to tune them according to the corpus being used.

Thank you in advance

marcelo-dalmeida:
Fixes #2597 part 1/2

In this first part, I intended to make easier to use BM25 model, using model parametrization instead of setting global variables.

Now one can set k1, b, and epsilon parameters directly

### DIFF
--- a/gensim/summarization/bm25.py
+++ b/gensim/summarization/bm25.py
@@ -68,47 +68,27 @@ class BM25(object):
 
     def __init__(self, corpus, k1=PARAM_K1, b=PARAM_B, epsilon=EPSILON):
         """
-
-        'k1 is a variable which helps determine term frequency saturation characteristics.
-        That is, it limits how much a single query term can affect the score of a given document.
-        A higher/lower k1 value means that the slope of “tf() of BM25” curve changes.
-        This has the effect of changing how “terms occurring extra times add extra score.”
-        An interpretation of k1 is that for documents of the average length, it is the value of the term frequency that
-        gives a score of half the maximum score for the considered term. The curve of the impact of tf on the score
-        grows quickly when tf() ≤ k1 and slower and slower when tf() > k1.'
-
-        Shane Connelly (2018). Practical BM25 - Part 2: The BM25 Algorithm and its Variables
-        https://www.elastic.co/pt/blog/practical-bm25-part-2-the-bm25-algorithm-and-its-variables
-
-
-        'If b is bigger, the effects of the length of the document compared to the average length are more amplified.
-        To see this, you can imagine if you set b to 0, the effect of the length ratio would be completely nullified
-        and the length of the document would have no bearing on the score'
-
-        Shane Connelly (2018). Practical BM25 - Part 2: The BM25 Algorithm and its Variables
-        https://www.elastic.co/pt/blog/practical-bm25-part-2-the-bm25-algorithm-and-its-variables
-
-
-        'A significant number of such experiments have been done, and suggest that in general values
-        such as 0.5 < b < 0.8 and 1.2 < k1 < 2 are reasonably good in many circumstances.
-        However, there is also evidence that optimal values do depend on other factors
-        (such as the type of documents or queries).'
-
-        Robertson, Stephen; Zaragoza, Hugo (2009). The Probabilistic Relevance Framework: BM25 and Beyond,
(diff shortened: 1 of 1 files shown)


Witiko:
Thank you for the submission. I suggest that we don't delete the constants, so that we don't break an established interface. The constants can serve as default parameter values for the class and the functions.

Witiko:
To fix the failing checks, see https://github.com/RaRe-Technologies/gensim/issues/2597#issuecomment-533884662.

gensim/summarization/bm25.py
```diff
@@ -66,14 +62,56 @@ class BM25(object):
         List of document lengths.
     """
 
-    def __init__(self, corpus):
+    def __init__(self, corpus, k1=1.5, b=0.75, epsilon=0.25):
         """
+
+        'k1 is a variable which helps determine term frequency saturation characteristics.
```
Witiko: First of all, the documentation is going to be displayed [on the Gensim website](https://radimrehurek.com/gensim/summarization/bm25.html). Therefore, you should use the [reStructuredText markup](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) for links, inline code, and citations.

Second of all, the documentation should follow the [NumPy docstring standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard). The class docstring should begin with a one-line summary, followed by one or more paragraphs of class description. Discussion of k1 and b should definitely be abbreviated and perhaps moved to the corresponding parameter descriptions. Multiple paragraphs are supported in parameter descriptions, but shorter is often better.

gensim/summarization/bm25.py
```diff
@@ -264,7 +268,7 @@ def _get_scores(bm25, document):
     return bm25.get_scores(document)
 
 
-def iter_bm25_bow(corpus, k1=1.5, b=0.75, epsilon=0.25, n_jobs=1):
+def iter_bm25_bow(corpus, k1=PARAM_K1, b=PARAM_B, epsilon=EPSILON, n_jobs=1):
```
Witiko: It may be better to move k1, b, and epsilon to the end of the keyword arguments. If someone calls `get_bm25_weights(corpus, n_jobs)`, then your change would silently break their code. The same applies to function `get_bm25_weights`.

mpenkov:
@Witiko Is this ready to merge?

Witiko:
@mpenkov No, there are still some open issues from a month ago. @marcelo-dalmeida, would you like to tackle them, or do you want me to take over?

marcelo-dalmeida:
Hi @Witiko 

I think I missed the change suggestion notification and the notion of time passing by

Sorry for hindering progress. If there is a desired 'deadline', please let me know.  I would not like to be an obstacle

I am going to fit these changes in my weekend compromise list

Witiko:
Hello @marcelo-dalmeida, that's a question for @mpenkov. If we are currently in a merge window for a release, then there may be a soft deadline. Otherwise, there should be no deadline, although the sooner the better.

mpenkov:
There is no deadline. If you're not going to be working on this for a while, could you please make it obvious what remains to be done? TODO lists work really well.

- [ ] try
- [ ] doing
- [x] this

That way it'll be obvious what's going on (without reading through the entire thread) and you'll remember what you were working on when you come back to the PR.

Witiko:
Looks good to me.

mpenkov:
@Witiko Is it easy (or worth it) to add tests for this new functionality?

Witiko:
Yes, I think this will need some tests in `gensim/test/test_BM25.py`.

mpenkov:
Ping @marcelo-dalmeida @Witiko 

How's it going on this PR?

Witiko:
The PR seems abandoned. Should I take over?

mpenkov:
If you don't hear from the author in the next few days, yes, please do.

marcelo-dalmeida:
Yes, please

Witiko:
Ok, I'll pull @marcelo-dalmeida's branch, add the unit tests and create a separate pull request.

PR #2607, opened 2019-09-23
Doc-comment class Doc2Vec only via __init__ method

gojomo:
For consideration/discussion: having both a class doc-comment, and an `__init__` method doc-comment, means both are stitched together in HTML docs (https://radimrehurek.com/gensim/models/doc2vec.html#gensim.models.doc2vec.Doc2Vec), **and** the parameter-details that every user needs to consult wind up **below** the confusing mess of internal-attributes that hardly anyone cares about. Google's Python doc-comment conventions suggest that a class and its parameters can be documented either under the class declaration, **or** under the `__init__` method, but not split between both. (See, eg: <https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html>). Here I've moved the notes about the class under `__init__()`, as that seems to get the most important info – init parameters - to where both code- and doc- readers will most easily find it.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -552,10 +552,10 @@ class Doc2Vec(BaseWordEmbeddingsModel):
 
         trainables : :class:`~gensim.models.doc2vec.Doc2VecTrainables`
             This object represents the inner shallow neural network used to train the embeddings. The semantics of the
-            network differ slightly in the two available training modes (CBOW or SG) but you can think of it as a NN 
-            with a single projection and hidden layer which we train on the corpus. The weights are then used as our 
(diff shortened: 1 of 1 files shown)


gojomo:
Thoughts @piskvorky @mpenkov ?

piskvorky:
Makes sense to me. Are other modules affected? (word2vec, lsi, etc)

gojomo:
I've noticed `Word2Vec` could benefit from the same moving-up of parameters (above internal attributes). Haven't looked everywhere, but if "either class comment or __init__ but not both" becomes project-style, someone could do a full pass-over.

mpenkov:
+1 this makes sense to me

gojomo:
Superceded by #2698 & 7cf2853cea9008bf0c3f9a982de09cb6777bfe13.

Issue #2600, opened 2019-09-09
Missing wheels for Windows


piskvorky:
Lately, there's been a stream of issues from Windows users, e.g. [here](https://groups.google.com/forum/#!topic/gensim/PbiMySYNPjs).

Looking at PyPI, it seems binary wheels are completely missing from the last few distributions (incl. the latest 3.8.0):
https://pypi.org/project/gensim/#files

This means Windows users don't get to install the optimized Gensim easily, and unless they have a compiler properly set up, they'll install the slow (uncompiled) version of word2vec, fastttext etc.

We definitely want those binary wheels on PyPI, **especially** for Windows. Windows users are the least tech-savvy, least likely to have a C compiler, and most likely to raise confused support tickets.

piskvorky:
The last distribution with Windows wheels seems to be 3.7.1, from January 2019:
https://pypi.org/project/gensim/3.7.1/#files

mpenkov:
@piskvorky I had a look, the problem is caused by failing AppVeyor builds (we use Travis CI for Linux and Mac, and AppVeyor for Windows). During the handover, I spent a lot of time dealing with problems in Travis CI, and neglected to check AppVeyor. Since the builds were failing, the Windows wheels weren't being built.

I'll investigate why the builds are failing - it seems to be a MemoryError in FastText code (https://ci.appveyor.com/project/piskvorky/gensim-wheels/builds/25836814/job/8yiy55wlf43soy01). Since the tests pass everywhere else, I suspect the AppVeyor machine is genuinely running out of memory, but I will need to double-check this.

Once that's fixed, I'll add some failsafes to prevent this from happening again:

1. Update documentation to explicitly focus on both Travis CI and AppVeyor (right now, it just mentions CI vaguely)
2. Add automatic checks for wheels for all platforms before finalizing the release

piskvorky:
OK, thanks. This is important, definitely warrants a new "bugfix release".

piskvorky:
@mpenkov do you have an ETA on those wheels?

Also needed here https://groups.google.com/d/msg/gensim/3ms1iH_Miuc/Wej0Go7EAQAJ

Having an internal automated process is nice, but not as urgent as having the wheels actually available for Windows users.

mpenkov:
[...]

PR #2610, opened 2019-09-26
added release/check_wheels.py

mpenkov:
This is a script for checking that all wheels are present in the dist subdirectory before publishing the release to PyPI.

The idea is that you run it in between `tox -e download_wheels` and `tox -e upload_wheels`. It will prevent us making an incomplete release (wheels missing for some platforms).

Part of the work-around for #2600.

### DIFF
--- a/release/check_wheels.py
+++ b/release/check_wheels.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+#
+# Authors: Michael Penkov <m@penkov.dev>
+# Copyright (C) 2019 RaRe Technologies s.r.o.
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
 """Check that our wheels are all there."""
 import os
 import os.path
--- a/release/check_wheels.py
+++ b/release/check_wheels.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Authors: Michael Penkov <m@penkov.dev>
-# Copyright (C) 2019 RaRe Technologies s.r.o.
+# Copyright (C) 2019 Radim Rehurek <me@radimrehurek.com>
 # Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
 """Check that our wheels are all there."""
 import os
--- a/release/check_wheels.py
+++ b/release/check_wheels.py
@@ -6,11 +6,14 @@
 """Check that our wheels are all there."""
 import os
 import os.path
+import re
+import sys
 
 #
 # We expect this to be set as part of the release process.
 #
 release = os.environ['RELEASE']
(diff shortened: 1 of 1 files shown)


release/check_wheels.py
```diff
@@ -0,0 +1,56 @@
+"""Check that our wheels are all there."""
```
piskvorky: Add standard preamble please (licence, author etc).

release/check_wheels.py
```diff
@@ -0,0 +1,61 @@
+# -*- coding: utf-8 -*-
+#
+# Authors: Michael Penkov <m@penkov.dev>
+# Copyright (C) 2019 RaRe Technologies s.r.o.
```
piskvorky:
```suggestion
# Copyright (C) 2019 Radim Rehurek <radimrehurek@seznam.cz>
```

mpenkov:
@piskvorky I applied your suggestion.

Should we apply the same changes to remaining source? More specifically:

- Make sure there is a preamble
- Make sure the preamble contains the author
- Make sure the preamble lists you personally as the copyright owner

There are many files that either have no preamble at all, or list the copyright owner as RaRe Tech (I copied the preamble from this file from another file in the gensim repo).

release/check_wheels.py
menshikh-iv: Any reasons for "copyright header" in each file? I just want to understand why this needed

release/check_wheels.py
```diff
@@ -0,0 +1,61 @@
...
+
+fail = False
+for f in expected:
+    wheel_path = os.path.join(dist_path, f % dict(release=release))
+    if not os.path.isfile(wheel_path):
+        print('FAIL: %s' % wheel_path)
+        fail = True
+
+if not fail:
+    print('OK')
```
menshikh-iv: maybe use retcode of script too (more "linux" way)?

release/check_wheels.py
```diff
@@ -0,0 +1,61 @@
...
+# Copyright (C) 2019 Radim Rehurek <me@radimrehurek.com>
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+"""Check that our wheels are all there."""
+import os
+import os.path
+
+#
+# We expect this to be set as part of the release process.
+#
+release = os.environ['RELEASE']
```
menshikh-iv: any basic validation for RELEASE var?

release/check_wheels.py
mpenkov: Good idea

release/check_wheels.py
mpenkov: OK, done

piskvorky:
@mpenkov ideally yes. I *think* we were cleaning the headers up in a recent refactoring (recent meaning ~1-2 years ago). So hopefully it's not too wild.

PR #2621, opened 2019-10-05
Changed the default value of blocksize in mz_keywords #2523

### LINKED ISSUE #2523: bug in gensim.summarization.mz_entropy.mz_keywords
**Problem statement:**

It seems to be a bug if the text is too short and number of words is lower than blocksize. In my case the values were: `n_words (232.0) ` and `blocksize (1024)`.

**Log:**

```
gensim\summarization\mz_entropy.py:127: RuntimeWarning: invalid value encountered in double_scalars
  - __log_combinations(n_words, blocksize)
```

**Dirty solution:**

Override `blocksize` value from the default `1024` to something lower:

`mz_keywords(text, blocksize=128)`

csaranbalaji:
fixes #2523

### DIFF
--- a/gensim/summarization/mz_entropy.py
+++ b/gensim/summarization/mz_entropy.py
@@ -10,7 +10,7 @@ import numpy as np
 import scipy
 
 
-def mz_keywords(text, blocksize=1024, scores=False, split=False, weighted=True, threshold=0.0):
+def mz_keywords(text, blocksize=128, scores=False, split=False, weighted=True, threshold=0.0):
     """Extract keywords from text using the Montemurro and Zanette entropy algorithm. [1]_
 
     Parameters
--- a/gensim/test/test_summarization.py
+++ b/gensim/test/test_summarization.py
@@ -270,8 +270,8 @@ class TestSummarizationTest(unittest.TestCase):
 
         with utils.open(os.path.join(pre_path, "head500.noblanks.cor"), 'rb') as f:
             text = utils.to_unicode(f.read())
-        text = u' '.join(text.split()[:1280])
-        kwds = mz_keywords(text)
+        text = u' '.join(text.split()[:10240])
+        kwds = mz_keywords(text, 1024)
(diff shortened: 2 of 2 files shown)


piskvorky:
That doesn't sound like a proper fix. Why should 128 be better than 1024? Why not 32, or 5, any other value?

csaranbalaji:
I'm not familiar with gensim, just implemented the dirty solution mentioned in https://github.com/RaRe-Technologies/gensim/issues/2523 assuming it would be right.

piskvorky:
Sorry, I'm -1 on that. It probably just sweeps the problem under a different rug, not really fixing anything.

csaranbalaji:
@piskvorky Debugging it I found that the function `gensim.summarization.mz_entropy.__log_combinations(n_words, blocksize)` returns `-inf` when the `n_words` is less than `blocksize`. As you said this is not a proper fix as it would require n_words as 128 to work properly. The error mentioned in https://github.com/RaRe-Technologies/gensim/issues/2523 occurs only when `__log_combinations(n_words, blocksize)` is called from `marginal_prob` function. I couldn't figure out when this function is called as for a very small sentence it is not called whereas for a quite bigger sentence it is called. 

As I'm not familiar with the codebase I'm not sure if the blocksize could be reduced. Sorry about that. It would be great if you could tell me what should be done to find the root cause of this issue. Thanks.

mpenkov:
I think this is a good opportunity to dive in and figure out what's going on.

- What is that function trying to achieve?
- What are the consequences of your proposed fix? e.g. who relies on the return value of that function not being -inf.
- What are non-"dirty" ways to fix this?

The code seems to link to an academic paper. It may be worth having a peek at it, if you swing that way.

I know that you're unfamiliar with the code base, but so is pretty much everyone else (that code is two years old, contributed by a non-core contribution AFAICT), so if you're up to the challenge, give it a dig :) If not, then document your findings as well as possible (either here or in the codebase) and we'll take it from there.

PR #2622, opened 2019-10-06
Merging a bug fix #2561. 

### LINKED ISSUE #2561: WikiCorpus.filter_wiki/remove_markup don't remove headings
#### Problem description

I am trying to get clean wiki texts. But still getting headings markup.

Just create WikiCorpus and call `get_texts`. Some texts will contain `==headings==` (different number of `=` and different headings, of course).
Simple test-case:
```
>>> import gensim.corpora.wikicorpus
>>> print(gensim.corpora.wikicorpus.filter_wiki('===heading==='))
===heading===
```

#### Versions

```Linux-4.9.125-linuxkit-x86_64-with-debian-10.0
Python 3.7.4 (default, Jul 13 2019, 14:04:11) 
[GCC 8.3.0]
NumPy 1.16.4
SciPy 1.3.0
gensim 3.8.0
FAST_VERSION 1
```

marat-s:
[Fix of 2561](https://github.com/RaRe-Technologies/gensim/issues/2561)

### DIFF
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -84,6 +84,8 @@ RE_P17 = re.compile(
     r'(^.{0,2}((bgcolor)|(\d{0,1}[ ]?colspan)|(rowspan)|(style=)|(class=)|(align=))(.*))',
     re.UNICODE
 )
+"""Remove headings """
+RE_P18 = re.compile(r'(#{1,6}) +|^\s*=*|\=*$|^\s*-*|\-*$', re.UNICODE)
 """Table markup"""
 IGNORED_NAMESPACES = [
     'Wikipedia', 'Category', 'File', 'Portal', 'Template',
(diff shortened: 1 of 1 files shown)


gojomo:
The link to the issue in the PR description is good - but ideally the title will also describe the matters affected. 

But also: see my comment on #2561 – despite the original reporter's preference that headings be stripped, I doubt that's the right default for everyone.

mpenkov:
I've taken the liberty of adjusting the title.

@gojomo Perhaps we can include a `strip_headers` keyword parameter (defaults to False) to handle both sides of the argument?

gojomo:
@mpenkov A parameter-controlled option would certainly support the needs of the reporter, @0x0badc0de, and address my concern about inconveniencing others. 

But this need might be narrow enough that the right advice, to those who need it, is to run the ~1-liner to strip headers yourself if you need it. (Roughly just a `s/^(=+).*\1//` regex replacement on the output text.)

gojomo:
See the further discussion on #2561 for clarification on what kind of fix the original reporter was expecting, and would be best.

gensim/corpora/wikicorpus.py
```diff
@@ -256,6 +262,9 @@ def remove_markup(text, promote_remaining=True, simplify_links=True):
         text = re.sub(RE_P13, '\n', text)  # leave only cell content
         text = re.sub(RE_P17, '\n', text)  # remove formatting lines
 
+        if not retain_heading_markup:
+            text = re.sub(RE_P18, '', text)  # remove headings
```
gojomo: Does this retain the words of the heading? (I don't think so.)

mpenkov:
Please add unit tests for your new functionaity.

PR #2626, opened 2019-10-07
Properly install Pattern library for documentation build

### LINKED ISSUE #2534: "Pattern library is not installed" ... but it is (perhaps because it is called pattern3)
#### Problem description

I am trying to do a gensim tutorial project. I am attempting to lemmatize and getting this error:
ImportError: Pattern library is not installed. Pattern library is needed in order to use lemmatize function

After searching the web, I installed pattern3, I still get the same error. I tried to do 'import pattern3 as pattern' but that doesn't help. Obviously I'm on python 3 (3.6)

 import gensim
 from gensim.utils import simple_preprocess, lemmatize
 import pattern3

lemmatize("hello")

#### Versions

Please provide the output of:

```python
```

Darwin-15.6.0-x86_64-i386-64bit
Python 3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
NumPy 1.16.4
SciPy 1.2.1
gensim 3.7.3
FAST_VERSION 1

#### Additional info

`from pattern3.web import Twitter`
this works fine

`from pattern.web import Twitter`
this does not (even after I do `import pattern3 as pattern`)

Hiyorimi:
Fixed dependency in the only place I could find. 

Fix #2534

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -282,7 +282,7 @@ docs_testenv = linux_testenv + distributed_env + [
     'sphinx',
     'sphinxcontrib-napoleon',
     'plotly',
-    'pattern >= 3.6',
+    'Pattern >= 3.6',
     'sphinxcontrib.programoutput',
 ]
 #
--- a/setup.py
+++ b/setup.py
@@ -282,6 +282,7 @@ docs_testenv = linux_testenv + distributed_env + [
     'sphinx',
     'sphinxcontrib-napoleon',
     'plotly',
+# Pattern's version is specified to install Pattern 3.6, which adds python3 support
     'Pattern >= 3.6',
     'sphinxcontrib.programoutput',
 ]
--- a/setup.py
+++ b/setup.py
@@ -276,13 +276,13 @@ if sys.version_info < (3, 7):
     ])
 
 if (3, 0) < sys.version_info < (3, 7):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -276,13 +276,13 @@ def finalize_options(self):
...
-    linux_testenv.extend(['nmslib'])    
-    
+    linux_testenv.extend(['nmslib'])
+
 docs_testenv = linux_testenv + distributed_env + [
     'sphinx',
     'sphinxcontrib-napoleon',
     'plotly',
-    'pattern <= 2.6',
+    'Pattern >= 3.6',
```
mpenkov: Is there a reason why we need to specify the version here?

Hiyorimi: To get version with python3 support.

mpenkov: Could you please mention that in a code comment? Someone is likely to ask the same question later on.

Hiyorimi: Pushed

mpenkov:
Looks good to me. Congrats on your first contribution to gensim :1st_place_medal: !

Hiyorimi:
Thx!

PR #2630, opened 2019-10-12
Remove native Python implementations of Cython extensions

### LINKED ISSUE #2511: Get rid of Python implementation of fastText, word2vec, doc2vec
We have Cython implementations for the above models, so that we fall back to Python when Cython is not available.

This is more hassle than it's worth, because:

1. The Python versions are too slow to be practical
2. We have to expend additional effort to test the Python versions (leading to bugs like this: https://github.com/RaRe-Technologies/gensim/issues/2508#issuecomment-497182316)
3. The Cython implementations are available most of the time. For people who don't have a compiler, we offer wheels.

mpenkov:
Fix #2511


gensim/models/deprecated/word2vec.py
```diff
@@ -162,9 +162,6 @@
 
 logger = logging.getLogger(__name__)
 
-
-# failed... fall back to plain numpy (20-80x slower training than the above)
-FAST_VERSION = -1
```
piskvorky: We may want to keep `FAST_VERSION`: it's in all tutorials and instructions, all around the web.

Also, its value is useful in debugging. Even with optimizations (`FAST_VERSION >= 0`), it's sometimes good to know if FAST_VERSION is 0, 1, or 2.

Although I'm not sure what this `deprecated/word2vec.py` is.

gensim/models/fasttext.py
```diff
@@ -301,132 +301,14 @@
 
 try:
     from gensim.models.fasttext_inner import train_batch_sg, train_batch_cbow
-    from gensim.models.fasttext_inner import FAST_VERSION, MAX_WORDS_IN_BATCH
```
piskvorky: Dtto: the actual value of `FAST_VERSION` is useful in debugging and support.

gensim/models/word2vec.py
```diff
@@ -151,201 +151,20 @@
...
 from six.moves import range
 
+import gensim.utils
+
 logger = logging.getLogger(__name__)
 
 try:
     from gensim.models.word2vec_inner import train_batch_sg, train_batch_cbow
     from gensim.models.word2vec_inner import score_sentence_sg, score_sentence_cbow
-    from gensim.models.word2vec_inner import FAST_VERSION, MAX_WORDS_IN_BATCH
```
piskvorky: Ditto.

gensim/utils.py
```diff
@@ -55,6 +55,29 @@
 PAT_ALPHABETIC = re.compile(r'(((?![\d])\w)+)', re.UNICODE)
 RE_HTML_ENTITY = re.compile(r'&(#?)([xX]?)(\w{1,8});', re.UNICODE)
 
+NO_CYTHON = RuntimeError(
+    "Cython extensions are unavailable. "
+    "Without them, this gensim functionality is disabled. "
+    "If you've installed from a package, ask the package maintainer to include Cython extensions. "
+    "If you're building gensim from source yourself, run `python setup.py build_ext --inplace` "
+    "and retry. "
+    "Alternatively, install an older version of gensim (earlier than 4.0.0). "
```
piskvorky: I wouldn't leave this out, not recommend older versions.

gensim/matutils.py
```diff
@@ -1424,183 +1424,6 @@ def close(self):
 
 try:
     # try to load fast, cythonized code if possible
-    from gensim.corpora._mmreader import MmReader
+    from gensim.corpora._mmreader import MmReader  # noqa
```
menshikh-iv: why? this must be always available, no more try/exc wth `FAST_VERSION`/`NO_CYTHON`/etc

gensim/models/doc2vec.py
```diff
@@ -74,13 +74,12 @@
...
-    memmap as np_memmap, vstack, integer, dtype, sum as np_sum, add as np_add, repeat as np_repeat, concatenate
-
+    memmap as np_memmap, vstack, integer, dtype
 
 from gensim.utils import call_on_class_only, deprecated
 from gensim import utils, matutils  # utility fnc for pickling, common scipy operations etc
-from gensim.models.word2vec import Word2VecKeyedVectors, Word2VecVocab, Word2VecTrainables, train_cbow_pair,\
-    train_sg_pair, train_batch_sg
+from gensim.models.word2vec import Word2VecKeyedVectors, Word2VecVocab, Word2VecTrainables
+from gensim.models.word2vec import train_cbow_pair, train_sg_pair, train_batch_sg  # noqa
```
menshikh-iv: `noqa` for what?

gensim/models/doc2vec.py
```diff
@@ -91,265 +90,12 @@
...
-                       * np_repeat(e_locks, model.vector_size).reshape(-1, model.vector_size))
-
-            if learn_doctags:
-                np_add.at(doctag_vectors, doctag_indexes, neu1e_r[:doctag_len])
-            if learn_words:
-                np_add.at(word_vectors, word_context_indexes, neu1e_r[doctag_len:])
-
-        return len(padded_document_indexes) - pre_pad_count - post_pad_count
+    #
+    # We could raise utils.NO_CYTHON right here, but let's delay it until
```
menshikh-iv: you no need to catch/raise anything here if extension available, if not - the user should use wheel / compile self.

gensim/test/test_fasttext.py
```diff
@@ -42,6 +41,17 @@ def __iter__(self):
                 yield utils.simple_preprocess(line)
 
 
+#
+# Not sure why these tests are hanging in my development environment.
+# They seem to be passing on CI.
+#
+SKIP_CORPUS_FILE = (
+    os.environ.get('DISABLE_CORPUS_FILE_TRAINING'),
```
menshikh-iv: default value if `DISABLE_CORPUS_FILE_TRAINING` not set?

tox.ini
```diff
@@ -44,9 +44,7 @@ commands =
     python --version
     pip --version
     python ci/install_wheels.py
-    python -c "from gensim.models.word2vec import FAST_VERSION; print(FAST_VERSION)"
```
menshikh-iv: Need to add explicit cython compilation in `setup.py`

mpenkov: How do we do this? I had a look through http://docs.cython.org/en/latest/src/quickstart/build.html, it looks like we're already including `ext_modules`. What else can we do?

gensim/test/test_fasttext.py
mpenkov: I'll get rid of this, we don't need it anymore.

tox.ini
menshikh-iv: 1. Add `cython` to install requirements
2. Remove "python fallback"
3. Cleanup all `.so` from folder
4. Run `python setup.py install` and make sure than all `.so` generated automatically, if not - ping me, I'll have a look myself

setup.py
```diff
@@ -395,6 +335,7 @@ def finalize_options(self):
         'scipy >= 0.18.1',
         'six >= 1.5.0',
         'smart_open >= 1.8.1',
+        'cython',
```
piskvorky: Why is this necessary?

If a user installs from a binary wheels, they don't need Cython for anything.

And if they install from source, they should be using the precompiled sources (`.c`), so they still don't need Cython?

mpenkov: Yes, they don't need it, but it's not difficult to install, so there is no harm in including it.

What is the benefit of including *.c sources, when they can be regenerated from Cython anyway?

piskvorky: Distributing generated `.c` is the official recommended way:
https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules

But I actually don't know why. We could easily pin the Cython version, to ensure repeatable cythonization. Is Cython really so painless to install? CC @menshikh-iv @gojomo do you remember why we do it this way?

Btw if users install from a binary wheel, these dependencies would still kick in, right? We definitely don't want to force users to install a compiler (Windows etc).

mpenkov: First, to answer your questions: yeah, with wheels, the binaries will already be in place, so Cython or compilation will happen.

Reading through your link, I can see two reasons why distributing .c and .cpp is the recommended way:

1. The installed Cython version may be incompatible with the source, and fail to compile correctly
2. Cython may not be installed all together, forcing the user to install it
3. Even if the correct Cython version is installed, the user "probably does not want to use it just to compile your module"

They don't seem to be particularly compelling. 1) and 2) are kind of weak, because installing Cython seems trivial (`pip install cython==1.2.3`) if we decide to pin the Cython version. 3) just sounds like bullshit to me: if people are building things from source, it's reasonable for them to expect to use a compiler at some stage.

That doesn't leave much justification for the "strong recommendation", in my opinion. Perhaps there are other reasons we are not aware of?

My main motivations for **not** including the .c and .cpp sources are:

1. Much simpler setup.py
2. Less maintenance work because we don't have to keep the sources in version control

@menshikh-iv WDYT?

menshikh-iv: @piskvorky 

> is Cython really so painless to install?

From my experience - yes, I never get this kind of problem.


>  do you remember why we do it this way?

No, this was done before I came to gensim, hope than @gojomo remember.


@mpenkov 

I'm +1 with pinning Cython version (to avoid any potential incompatibilities).

> if people are building things from source, it's reasonable for them to expect to use a compiler at some stage.


I agree, sometimes this really needed (in case if you want to make a contribution).
I also supports both of your points (setup.py & less maintenance). I see no practical reasons to store "auto-generated" files in a git repo.

piskvorky: > First, to answer your questions: yeah, with wheels, the binaries will already be in place, so Cython or compilation will happen.

You mean "will **not** happen", right?

If that's the case = Windows users can continue to enjoy Gensim without having to install compilers, then I'm +1 on requiring Cython (=compiler) if installing from source (as opposed to binary wheel).

I'm thinking that maybe we had to make Cython optional before because of the pure-Python code paths. But that's no longer an issue.

mpenkov: Yes, sorry, I misspoke. Compilation will **not** happen.

gojomo: `cython` has not previously been a declared prerequisite, and was only necessary for developers who were modifying the `.pyx` (etc) sources. Before wheels, every install required a local compilation of the `.c` files. That setup predates any of my contributions. (Although the wheels clearly help on Windows, I'm not sure they're actually a better experience everywhere else. I liked the clarity of having all source that was possibly running on your machine locally; the wheels introduce the risk that a corrupted/subverted build process sends users binaries that don't match public source.) Personally I wouldn't add `cython` as a declared requirement - why trigger installers to install something, with the storage costs and even executable-code risks, that isn't really needed?

mpenkov: I've amended setup.py to only install cython when it is needed (e.g. if the C/C++ files are missing).

mpenkov:
TODO:

[ ] Build wheel locally
[ ] Install wheel, confirm that Cython is not required

setup.py
```diff
@@ -276,11 +231,10 @@ def make_cpp_extensions():
             language='c++',
             extra_compile_args=extra_args,
             extra_link_args=extra_args,
-            include_dirs=include_dirs,
         )
 
 
-ext_modules = list(make_c_extensions()) + list(make_cpp_extensions())
+ext_modules = list(itertools.chain(make_c_extensions(), make_cpp_extensions()))
```
piskvorky: Why? Both longer *and* more complicated.

Issue #2556, opened 2019-07-14
D2VTransformer raises if passed a Pandas series without index key 0


rpetchler:
`D2VTransformer` raises if passed a Pandas series with an index that does not contain the key 0:

```python
import pandas as pd
from gensim.sklearn_api import D2VTransformer
from gensim.test.utils import common_texts

series = pd.Series(common_texts)
series.index += 1  # Increment the index so that it does not contain the key 0

transformer = D2VTransformer(min_count=1, size=5)
transformer.fit(series)
```

Output:

```python
﻿Traceback (most recent call last):
  File "main.py", line 9, in <module>
    transformer.fit(series)
  File "venv/lib/python3.7/site-packages/gensim/sklearn_api/d2vmodel.py", line 162, in fit
    if isinstance(X[0], doc2vec.TaggedDocument):
  File "venv/lib/python3.7/site-packages/pandas/core/series.py", line 868, in __getitem__
    result = self.index.get_value(self, key)
  File "venv/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 4375, in get_value
    tz=getattr(series.dtype, 'tz', None))
  File "pandas/_libs/index.pyx", line 81, in pandas._libs.index.IndexEngine.get_value
  File "pandas/_libs/index.pyx", line 89, in pandas._libs.index.IndexEngine.get_value
  File "pandas/_libs/index.pyx", line 132, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 987, in pandas._libs.hashtable.Int64HashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 993, in pandas._libs.hashtable.Int64HashTable.get_item
KeyError: 0
```

This occurs because the [`fit`](https://github.com/RaRe-Technologies/gensim/blob/4543646d3fe3496e11bc935e72cbf9b18504442e/gensim/sklearn_api/d2vmodel.py#L162) and [`transform`](https://github.com/RaRe-Technologies/gensim/blob/4543646d3fe3496e11bc935e72cbf9b18504442e/gensim/sklearn_api/d2vmodel.py#L198) methods of `D2VTransformer` require `__getitem__` on the passed iterable not to raise an exception for key 0.

Versions:

Darwin-18.6.0-x86_64-i386-64bit
[...]

PR #2632, opened 2019-10-14
Fixing d2vtransformer index error #2556

Hiyorimi:


gensim/sklearn_api/d2vmodel.py
```diff
@@ -145,6 +145,11 @@ def __init__(self, dm_mean=None, dm=1, dbow_words=0, dm_concat=0, dm_tag_count=1
         self.sorted_vocab = sorted_vocab
         self.batch_words = batch_words
 
+    def _get_first(some_iterable):
+        """Returns first element of some iterable."""
```
piskvorky: PEP257: `Return X`, not `Returns X` (imperative mode).

gensim/sklearn_api/d2vmodel.py
```diff
@@ -159,7 +164,8 @@ def fit(self, X, y=None):
             The trained model.
 
         """
-        if isinstance([i for i in X[:1]][0], doc2vec.TaggedDocument):
+
```
piskvorky: Extra blank line.

gensim/sklearn_api/d2vmodel.py
```diff
@@ -159,7 +164,8 @@ def fit(self, X, y=None):
             The trained model.
 
         """
-        if isinstance([i for i in X[:1]][0], doc2vec.TaggedDocument):
+        
+        if isinstance(self._get_first(X), doc2vec.TaggedDocument):
```
piskvorky: What happened to the `next(iter(x))` pattern? Is it not suitable here?

gensim/sklearn_api/d2vmodel.py
Hiyorimi: Fixed

gensim/sklearn_api/d2vmodel.py
Hiyorimi: It doesn't work on Pandas dataframe:


```python
>>> a = pd.DataFrame([[1,2], [2,4]], columns=['1', '2'])
>>> a.index += 1
>>> _get_first = lambda X: next(iter(X))
>>> _get_first(a)
'1'```

File: gensim/sklearn_api/d2vmodel.py (hunk 2 above)
Comment (Hiyorimi) [reply gensim/sklearn_api/d2vmodel.py]: fixed

**Comment by Hiyorimi** [comment]
@piskvorky done

File: gensim/sklearn_api/d2vmodel.py (hunk 3 above)
Comment (piskvorky) [reply gensim/sklearn_api/d2vmodel.py]: That's weird. So `next(iter(x))` returns something else than `list(X)[0]`, on non-empty inputs?

If so, I bet other things will be broken too.

Comment (Hiyorimi) [reply gensim/sklearn_api/d2vmodel.py]: Pandas has special .iterrows() method for that, I guess it is there for reasons)

**Comment by Hiyorimi** [comment]
@piskvorky I can move method into .fit() method, but I doubt there is something else we can do :)

**Comment by Hiyorimi** [comment]
@piskvorky how can I resolve build issues?

**Comment by piskvorky** [comment]
If you click on the "Details" below, next to Travis CI, it will take you to the log of build issues. They seem like genuine errors, e.g. you call a method with the wrong number of parameters:

Screen Shot 2019-10-16 at 09 32 14

@mpenkov is traveling to Europe now, offline for a while.

File: gensim/sklearn_api/d2vmodel.py [hunk 4]
```diff
@@ -145,6 +145,11 @@ def __init__(self, dm_mean=None, dm=1, dbow_words=0, dm_concat=0, dm_tag_count=1
         self.sorted_vocab = sorted_vocab
         self.batch_words = batch_words
 
+    def _get_first(some_iterable):
```
mpenkov: Methods should accept `self` as the first parameter.

Why is this a method? It can be a pure function, either at the module level or internal.

gensim/sklearn_api/d2vmodel.py
```diff
@@ -159,7 +164,7 @@ def fit(self, X, y=None):
             The trained model.
 
         """
-        if isinstance([i for i in X[:1]][0], doc2vec.TaggedDocument):
+        if isinstance(self._get_first(X), doc2vec.TaggedDocument):
```
mpenkov: Don't you want this instead?

```suggestion
        if isinstance(self._get_first(X[:1]), doc2vec.TaggedDocument):
```

Hiyorimi: Done

gensim/sklearn_api/d2vmodel.py
Hiyorimi: done

gensim/sklearn_api/d2vmodel.py
```diff
@@ -164,7 +159,11 @@ def fit(self, X, y=None):
             The trained model.
 
         """
-        if isinstance(self._get_first(X), doc2vec.TaggedDocument):
+        def _get_first(some_iterable):
+            """Return first element of some iterable."""
+            for elem in some_iterable:
+                return elem
+        if isinstance(_get_first(X[:1]), doc2vec.TaggedDocument):
```
piskvorky: Isn't this the same as `_get_first(_get_first(X))`?

I'm still confused by what `X` actually is, and what Pandas is doing. At the very least, this line deserves a thorough comment (though a clear code is always preferable).

Hiyorimi: I thought that @mpenkov suggested `[:1]` part to improve performance. Since it part is confusing, I will remove it, because to me it seems redundant.

mpenkov: I suggested it because it seemed correct, not because of performance.

If you look at the source before the change, it's:

```python
if isinstance([i for i in X[:1]][0], doc2vec.TaggedDocument):
    ...
```

My understanding of what that source is doing:

```python
tmp = X[:1]  # it'd be helpful to know what is X (and what is tmp) here
first = _get_first(tmp)  # first = [i for i in tmp][0]
if isinstance(first, doc2vec.TaggedDocument):
    ...
```

Is my understanding incorrect?

Hiyorimi: Wow, I missed that. It should indeed be `_get_first(_get_first(X))`

mpenkov: Can we somehow make it more obvious what X is in this case? It's the elephant in the room.

What data type is it?

mpenkov: Finally, we also want a unit test to cover this code path, in order to prevent regressions.

brienna:
I'm struggling with making D2VTransformer accept an iterator. It would be nice to see a fix. The only alternative solutions are to load all documents into memory, or to write my own wrapper class. 

```
<ipython-input-98-3b56b3a565e2> in <module>
      1 d = D2VTransformer()
      2 X = AbstractIterator(x_test, 'doc2vec')
----> 3 d.fit(X)

~/anaconda3/envs/thesis-env/lib/python3.7/site-packages/gensim/sklearn_api/d2vmodel.py in fit(self, X, y)
    160 
    161         """
--> 162         if isinstance(X[0], doc2vec.TaggedDocument):
    163             d2v_sentences = X
    164         else:

TypeError: 'AbstractIterator' object is not subscriptable
```

piskvorky:
Hi @brienna , sorry, we've decommissioned the sklearn wrappers. They'll all disappear in Gensim 4.0.0: [release notes](https://github.com/RaRe-Technologies/gensim/releases/tag/4.0.0beta), for reasons you mention (unclear design and documentation).

If you find them useful, you could fork and adapt the current (deprecated) code, or – perhaps more easily and flexibly – write your own wrapper, optimized for the task at hand.

mpenkov:
@Hiyorimi Thank you for your contribution, but we have since completely removed the sklearn wrappers, so this PR is no longer relevant to gensim.

PR #2639, opened 2019-10-17
Account for empty corpus in HDPModel. Fixes #2563

### LINKED ISSUE #2563: HdpModel(corpus=[], id2word=dic, alpha=1, T=500) will never ends
#### Problem description

Hi ~ I find that when I give a empty array as corpus to the constructor of corpora.Dictionary , it would never ends,the program got an endless loop.

I know I really should check the corpus to make sure it cannot be an empty array,but perhaps the HdpModel constructor should raise an Exception at once instead of endless-loop without any message(which perhaps make some new learner confused )

You can repeat the problem I found before
```python

# _*_ coding=utf-8 _*_
from gensim import corpora
from gensim.models import HdpModel

if __name__ == '__main__':
    dic = corpora.Dictionary([])
    print(1)
   m = HdpModel(corpus=[], id2word=dic, alpha=1, T=500)
    print('this will never be print')

```

in actually, when the corpus is [](an empty list),the HdpModel's constructor call the
```
 if corpus is not None:
            self.update(corpus)
```
and in the update function, when the corpus is empty array, the "while True" loop will never ends .It just because of that "utils.grouper(corpus, self.chunksize)" in "while True" is an empty generator too, " for chunk in utils.grouper" will never happen， caused the while-True loop never ends.

That's very nice of you to see the issue, I am not a native English speaker，so please forgive my pool English .If you have any question, just communicate with me .

dm20:
There are a couple of ways to go about preventing the infinite loop mentioned in the Issue from occurring. I thought maybe to user logger.error, or to skip the while loop if the corpus is empty. ButI saw in the _inference_ method that a RunTimeError is thrown when lda_alpha and lda_beta are not initialized. So I thought this seemed consistent.

### DIFF
--- a/gensim/models/hdpmodel.py
+++ b/gensim/models/hdpmodel.py
@@ -339,6 +339,8 @@ class HdpModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
             corresponding lda model from :meth:`~gensim.models.hdpmodel.HdpModel.suggested_lda_model`
 
         """
+        if len(corpus) == 0:
+            raise RuntimeError("cannot initialize model with empty corpus")
         self.corpus = corpus
         self.id2word = id2word
         self.chunksize = chunksize


gensim/models/hdpmodel.py
```diff
@@ -339,6 +339,8 @@ def __init__(self, corpus, id2word, max_chunks=None, max_time=None,
             corresponding lda model from :meth:`~gensim.models.hdpmodel.HdpModel.suggested_lda_model`
 
         """
+        if len(corpus) == 0:
```
piskvorky: Is `corpus` guaranteed to support `len()` in HdpModel? What if it's just a generator?

piskvorky: Most definitely not. The `len` exception is still there, plus now you're serializing into entire corpus into RAM! If it's a generator, it will be gone.

Does HdpModel support generators? Or are we guaranteed to have a repeatable iterable with `len`?

dm20: ah ok got it, well the comments in the file just say 
```
corpus : iterable of list of (int, float)
            Corpus in BoW format.
```
so I wouldn't call this a guarantee but it does sound like list is the intended type. Should there maybe be type checking to guarantee that list is used?

mpenkov: It's not a list. It's an _iterable_ of something.

That something is a list of tuples, but that still doesn't change the fact that you're dealing with an iterable, and you won't necessarily be able to call `len()` on it.

dm20: oh I see, I understand now. it looks like `len()` is already being called on the corpus in another place in the constructor
```
if corpus:
        self.m_D = len(corpus)
```
leading me to believe that the corpus does in fact need to be a list..or should this be changed as well?

dm20: would it maybe be easiest to use the existing utils.is_corpus method in the constructor? since it covers the empty corpus case and works for generators

mpenkov: Yeah, I'd say calling `len(corpus)` is probably a bug, because it risks consuming the iterable.

piskvorky: It's possible the HDP model does not support iterables. It is possible it supports `list` only :( (unlike other models in Gensim)

That's why I ask about the contract of HdpModel – whether it explicitly needs its input data in RAM (~list), or if it supports streamed inputs (~any iterable).

dm20: @piskvorky I'll look over the code and also look around for usage of the model with a non-list iterable. if it turns out that only list is supported, would it be helpful to add support for all iterables? or not worth the effort since the model may be deprecated soon

piskvorky: > I'll look over the code and also look around for usage of the model with a non-list iterable.

That will be great, thanks.

> or not worth the effort since the model may be deprecated soon

I'd say "not worth the effort" right now. It isprobably possible to clean up and optimize HDP, but we lack a clear motivating use-case. So, before any technical optimizations, we'd need a **good strong tutorial / how-to document**, showing off *why* HDP is worth it, how it's better than other models, when to use it.

The performance considerations are secondary to my mind.

PR #2640, opened 2019-10-17
Increase max supported numpy version to < 1.17.0 for py2

### LINKED ISSUE #2634: numpy requirement is too strict/outdated
#### Problem description

Since `gensim 3.8.0`,  `numpy` is restricted to be `numpy >= 1.11.3, <= 1.16.1`.
That was first introduced in #2546 for py2 since `numpy 1.16.1` **was** the last version supporting py2. However, as of now, the latest version supporting py2 is [`numpy 1.16.5`](https://github.com/numpy/numpy/releases/tag/v1.16.5).

Should restrict numpy to `numpy >= 1.11.3, < 1.17.0`, since `1.17.0` is the first version removing python2 support.

#### Versions

```
$ python
Python 2.7.12 (default, Jan  9 2017, 12:16:27)
[GCC 4.8.4] on linux2
Linux-4.9.184-linuxkit-x86_64-with-Ubuntu-14.04-trusty
('Python', '2.7.12 (default, Jan  9 2017, 12:16:27) \n[GCC 4.8.4]')
('NumPy', '1.16.1')
('SciPy', '0.18.1')
('gensim', '3.8.1')
('FAST_VERSION', 1)
```

dm20:
Fix #2634

As reported in the issue, the last version of numpy that supported python2 was 1.16.5, so this requirement can be increased slightly from 1.16.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -233,7 +233,7 @@ Copyright (c) 2009-now Radim Rehurek
 # https://docs.pytest.org/en/latest/py27-py34-deprecation.html
 #
 if PY2:
-    NUMPY_STR = 'numpy >= 1.11.3, <= 1.16.1'
+    NUMPY_STR = 'numpy >= 1.11.3, < 1.17.0'
     PYTEST_STR = 'pytest == 4.6.4'
 else:
     NUMPY_STR = 'numpy >= 1.11.3'


mpenkov:
Thank you for your contribution.

We dropped Py2 support, so do we even need this restriction anymore?

allanbenW:
This is for 3.8.x though, which still supports py2.7

mpenkov:
No, technically, it's for 4.0.0 (Python2 support is already removed from the current develop branch).

We just haven't bumped the version on the develop branch - that happens later, as part of our release process.

dm20:
are there old versions that are still supported where this would be beneficial?

mpenkov:
I don't think so.

We don't maintain the older versions: once they're released, we move on to the next version. I can see us making an exception to fix critical defects, but this numpy dependency isn't one of them.

piskvorky:
@dm20 do you still use Python 2? Or how did you find this issue / PR?

dm20:
@piskvorky I am using Python 3 with gensim - I found this issue when looking through the project for quick/easy Hacktoberfest issues. Beyond Hacktoberfest, gensim has been very helpful to me and my colleagues at work so I want to help in some way if I can!

mpenkov:
I think unless there is a significant reason for pinning the version (e.g. incompatibility), then we can simply remove the numpy pin altogether.

@menshikh-iv What do you think?

mpenkov:
@allanbenW We ended up accepting your proposed change for the 3.8.3 release (7125731fef5e9f162c2b7188ec3a2b82e5d098df). Thanks again!

Issue #2642, opened 2019-10-21
Speed up word2vec / fasttext model loading


piskvorky:
Loading a large word2vec model with `load_word_format(binary=True)` is slow. Users complain that loading the "standard" models published by Facebook is too slow, plus it also affects the speed of our own tests and tutorial autogeneration.

Some numbers:

```python
time gensim.models.keyedvectors.KeyedVectors.load_word2vec_format('./word2vec-google-news-300.gz', binary=True)
2019-10-21 22:24:08,326 : INFO : loading projection weights from ./word2vec-google-news-300.gz
2019-10-21 22:26:54,620 : INFO : loaded (3000000, 300) matrix from ./word2vec-google-news-300.gz
CPU times: user 2min 42s, sys: 3.64 s, total: 2min 46s
Wall time: 2min 46s
```

The I/O part itself = only loading the bytes from the file without any interpretation, takes about 30 seconds:

```python
time full = io.BytesIO(smart_open.open('./word2vec-google-news-300.gz', 'rb').read())
CPU times: user 20.9 s, sys: 8.13 s, total: 29.1 s
Wall time: 31.9 s
```

…which means our parsing code is taking up the majority of the `load_word2vec_format` time. Ideally, we shouldn't need much more than the 30 seconds (= the raw I/O speed) for the full `load_word_format(binary=True)`. Nearly 3 minutes is too much.

**Task**: Optimize `load_word2vec_format`, especially the `binary=True` branch. The code seems to live here:
https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/utils_any2vec.py#L369

piyush-kgp:
Hi, I am interested in solving this issue. However I was unable to reproduce what you got. Specifically, the following code block (from what you linked):
```
from gensim.models.keyedvectors import Vocab
from gensim import utils
import numpy as np
import io
import smart_open
import time

tit_file_read = time.time()
fin = utils.open("~/Downloads/GoogleNews-vectors-negative300.bin", 'rb')
tat_file_read = time.time()

header = utils.to_unicode(fin.readline(), 'utf-8')
vocab_size, vector_size = (int(x) for x in header.split())  # throws for invalid file format
binary_len = np.dtype(np.float32).itemsize * vector_size

word2vec = {}
word=[]

tit_for = time.time()
for i in range(vocab_size):
    # mixed text and binary: read text first, then binary
    word = []
    while True:
[...]

PR #2643, opened 2019-10-21
Optimizing load_word2vec_format for binary equal to true

maniis:
#2642  Changed numpy.fromstring() to split() because it is slow in comparision to split().


maniis:
Testing

PR #2654, opened 2019-10-26
replacing deleted notebooks with placeholders

mpenkov:
#2591 ported some Jupyter notebooks to Sphinx gallery and then removed them. That turned out to be undesirable because it left many broken links that we cannot control via our Web server.

This PR replaces the previously removed notebooks with a placeholder document that points to the new location.

### DIFF
--- a/docs/notebooks/Corpora_and_Vector_Spaces.ipynb
+++ b/docs/notebooks/Corpora_and_Vector_Spaces.ipynb
@@ -1,0 +1,34 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This tutorial has moved.\n",
+    "\n",
+    "Please see https://radimrehurek.com/gensim/auto_examples/."
+   ]
+  }
+ ],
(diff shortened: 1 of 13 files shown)


piskvorky:
Looks good, except some of these tutorials were *removed*, rather than *moved*. So people won't find them at that new suggested location.

qinpeiming:
good

gojomo:
While better than a 404-not-found, I was hoping for placeholders that point directly to the analogous new tutorial, where there is one. (If there isn't one, because the old tutorial was so bad it's better deleted, then a 404 might be fine. I'm not sure which ones have been deleted-without-an-equivalent-replacement.)

mpenkov:
OK. I'm going to merge this in the interest of our users, and come back to tweaking things later (probably around next week or so).

rice-cake:
in Similarity Queries notebook:

sims = sorted(enumerate(sims), key=lambda item: -item[1])
for i, s in enumerate(sims): print(s, documents[i]) <-- incorrect

for s in sims: print(s, documents[s[0]]) <-- correct

Issue #2535, opened 2019-06-21
streamlining most_similar_cosmul


eyaler:
this is a feature request to make most_similar_cosmul() more similar (pun intended) to most_similar() by the following changes:

1. add restrict_vocab argument
2. allow positive and negative to be strings instead of list-of-strings in case of single words
3. allow usage in evaluate_word_analogies

EinKaiser:
Hi, can I have a go at this?

mpenkov:
Yes.

n3hrox:
Hey @eyaler, @mpenkov - since there was no open PRs for this I decided to give it a go, let me know what you think

PR #2656, opened 2019-10-28
streamlining most_similar_cosmul and evaluate_word_analogies

n3hrox:
Closes: https://github.com/RaRe-Technologies/gensim/issues/2535

This is my first PR for gensim so all comments are welcome. 
To be honest I have no idea how to test `restrict_vocab` for `most_similar_cosmul` or `most_similar` for `evaluate_word_analogies`. I wanted to write something similar to already existing tests for these keywords but did not find any (nor tests for `restrict_vocab` keyword in case of `most_similar` function and nor `most_similar` keyword in case of `accuracy` function)

Summary:

- Added new `restrict_vocab` parameter to `most_similar_cosmul`
- Improved `most_similar_cosmul` shorthand to handle both positive and negative cases
- Parameterized similarity function in `evaluate_word_analogies`


gojomo:
+1 (having looked over code, but not tested functionality)

gensim/models/keyedvectors.py
```diff
@@ -818,10 +825,13 @@ def most_similar_cosmul(self, positive=None, negative=None, topn=10):
 
         self.init_sims()
 
-        if isinstance(positive, string_types) and not negative:
+        if isinstance(positive, string_types):
             # allow calls like most_similar_cosmul('dog'), as a shorthand for most_similar_cosmul(['dog'])
             positive = [positive]
 
+        if isinstance(negative, string_types):
```
mpenkov: If I understand correctly, this enables behavior like:

    most_similar_cosmul('dog', 'cat')

where `dog` is positive and `cat` is negative. That's helpful shorthand, but without documentation, people won't find out about it.

Can you please add a paragraph to the docstring explaining the above shorthand?

gensim/models/keyedvectors.py
```diff
@@ -1017,7 +1028,9 @@ def _log_evaluate_word_analogies(section):
             logger.info("%s: %.1f%% (%i/%i)", section['section'], 100.0 * score, correct, correct + incorrect)
             return score
 
-    def evaluate_word_analogies(self, analogies, restrict_vocab=300000, case_insensitive=True, dummy4unknown=False):
+    def evaluate_word_analogies(
```
mpenkov: Please add some unit tests to cover the newly added functionality.

gensim/models/keyedvectors.py
gojomo: Note that this just makes the special type-testing treatment of `negative` match that of `positive` (in both `most_similar()` and `most_similar_cosmul()`) – but that special treatment, while used extensively in examples, isn't currently documented even in the `most_similar()` case! I'd suggest that treating `negative` symmetrically with `positive` is a good idea, and should also be done in `most_similar()` for consistency, and both of their doc-comments should be improved/harmonized to explain this behavior.

mpenkov:
@n3hrox Ping! Are you able to finish this PR?

n3hrox:
@mpenkov I will try to come back to this during weekend. I waited really long for this to be reviewed, started new job and had completely no time during Dec/Jan

gensim/models/keyedvectors.py
n3hrox: Done

gensim/models/keyedvectors.py
n3hrox: @mpenkov added shorthand in docstring for clarity, is it okay or is there some other place for docs as well that this should be added?

n3hrox:
@mpenkov I adjusted PR accordingly, please re-review

piskvorky:
@mpenkov what happened here? PR was marked Stale, but it looks like @n3hrox did respond (2 years ago…).

mpenkov:
Yeah, looks like they responded right after we marked it as stale, and then we didn't follow up.

mpenkov:
I think the correct action is to reopen and push this over the line ourselves.

@n3hrox Sorry for the delay. This fell off our radar.

mpenkov (approved):

PR #2661, opened 2019-10-31
Document accessing model's vocabulary

### LINKED ISSUE #2660: Document accessing the vocabulary of a *2vec model
@andreamoro like this:

```python
from gensim.test.utils import common_texts

model = Word2Vec(common_texts, size=100, window=5, min_count=1, workers=4)
word_vectors = model.wv
print(list(word_vectors.vocab.keys()))

['minors', 'graph', 'system', 'trees', 'eps', 'computer', 'survey', 'user', 'human', 'time', 'interface', 'response']
```

[Documentation of `KeyedVectors`](https://radimrehurek.com/gensim/models/keyedvectors.html) = the class holding the trained word vectors.

@mpenkov listing the model vocab is a reasonable task, but I couldn't find it in our documentation either. I had to look at the source code. Maybe we can add it somewhere? API ref? Tutorial? @andreamoro where would you expect / look for this information?

_Originally posted by @piskvorky in https://github.com/RaRe-Technologies/gensim/issues/2422#issuecomment-547780433_

mpenkov:
- Fix #2660


docs/src/auto_examples/tutorials/run_word2vec.py
```diff
@@ -145,7 +152,7 @@
 # out the FastText model.
 #
 try:
-    vec_weapon = wv['cameroon']
+    vec_cameroon = wv['cameroon']
```
piskvorky: Nice catch 👍

PR #2671, opened 2019-11-07
Speed up word2vec model loading

### LINKED ISSUE #2642: Speed up word2vec / fasttext model loading

lopusz:
Fixes #2642 .

### DIFF
--- a/gensim/models/utils_any2vec.py
+++ b/gensim/models/utils_any2vec.py
@@ -177,7 +177,7 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
         (Experimental) Can coerce dimensions to a non-default float type (such as `np.float16`) to save memory.
         Such types may result in much slower bulk operations or incompatibility with optimized routines.)
     binary_chunk_size : int, optional
-        Size of chunk in which binary files are read. Used mostly for testing. Defalut value 100 kB.
+        Read input file in chunks of this many bytes for performance reasons.
 
     Returns
     -------
@@ -204,12 +204,6 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
         result.vectors[word_id] = weights
         result.index2word.append(word)
 
-    def __remove_initial_new_line(s):
-        i = 0
-        while i < len(s) and s[i] == '\n':
-            i += 1
-        return s[i:]
-
     def __add_words_from_binary_chunk_to_result(result, counts, max_words, chunk, vector_size, datatype):
         start = 0
         n = len(chunk)
@@ -222,7 +216,7 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
             if i_space != -1 and (n - i_vector) >= n_bytes_per_vector:
                 word = chunk[start:i_space].decode("utf-8", errors=unicode_errors)
                 # Some binary files are reported to have obsolete new line in the beginning of word, remove it
-                word = __remove_initial_new_line(word)
+                word = word.lstrip('\n')
                 vector = frombuffer(chunk, offset=i_vector, count=vector_size, dtype=REAL).astype(datatype)
                 __add_word_to_result(result, counts, word, vector)
                 start = i_vector + n_bytes_per_vector
@@ -242,6 +236,7 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
             for line in fin:
                 word, count = utils.to_unicode(line, errors=unicode_errors).strip().split()
                 counts[word] = int(count)
+
     logger.info("loading projection weights from %s", fname)
     with utils.open(fname, 'rb') as fin:
         header = utils.to_unicode(fin.readline(), encoding=encoding)
@@ -259,8 +254,8 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
                 new_chunk = fin.read(binary_chunk_size)
                 chunk += new_chunk
                 max_words = vocab_size - len(result.vocab)
-                processed_words, chunk = __add_words_from_binary_chunk_to_result(result, counts, max_words,
-                                                                chunk, vector_size, datatype)
+                processed_words, chunk = __add_words_from_binary_chunk_to_result(
+                    result, counts, max_words, chunk, vector_size, datatype)
                 tot_processed_words += processed_words
                 if len(new_chunk) < binary_chunk_size:
                     break
--- a/gensim/models/utils_any2vec.py
+++ b/gensim/models/utils_any2vec.py
@@ -28,7 +28,7 @@ current Facebook implementation, use :py:func:`compute_ngrams_bytes` and
 import logging
 from gensim import utils
 
-from numpy import zeros, dtype, float32 as REAL, ascontiguousarray, fromstring
+from numpy import zeros, dtype, float32 as REAL, ascontiguousarray, frombuffer
 
 from six.moves import range
 from six import iteritems, PY2
@@ -147,7 +147,7 @@ def _save_word2vec_format(fname, vocab, vectors, fvocab=None, binary=False, tota
 
 
 def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8', unicode_errors='strict',
-                          limit=None, datatype=REAL):
+                          limit=None, datatype=REAL, binary_chunk_size=100 * 1024):
     """Load the input-hidden weight matrix from the original C word2vec-tool format.
 
     Note that the information stored in the file is incomplete (the binary tree is missing),
@@ -176,6 +176,8 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
     datatype : type, optional
         (Experimental) Can coerce dimensions to a non-default float type (such as `np.float16`) to save memory.
         Such types may result in much slower bulk operations or incompatibility with optimized routines.)
+    binary_chunk_size : int, optional
+        Size of chunk in which binary files are read. Used mostly for testing. Defalut value 100 kB.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


gensim/models/utils_any2vec.py
```diff
@@ -176,14 +176,64 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
     datatype : type, optional
         (Experimental) Can coerce dimensions to a non-default float type (such as `np.float16`) to save memory.
         Such types may result in much slower bulk operations or incompatibility with optimized routines.)
+    binary_chunk_size : int, optional
+        Size of chunk in which binary files are read. Used mostly for testing. Defalut value 100 kB.
```
piskvorky:
```suggestion
        Read input file in chunks of this many bytes, for performance reasons.
```

Does this need to be an API parameter (as opposed to a constant)? Under what conditions would a user want to change this parameter?

gensim/models/utils_any2vec.py
```diff
@@ -176,14 +176,64 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
+            # use count from the vocab file
+            result.vocab[word] = Vocab(index=word_id, count=counts[word])
+        else:
+            # vocab file given, but word is missing -- set count to None (TODO: or raise?)
+            logger.warning("vocabulary file is incomplete: '%s' is missing", word)
+            result.vocab[word] = Vocab(index=word_id, count=None)
+        result.vectors[word_id] = weights
+        result.index2word.append(word)
+
+    def __remove_initial_new_line(s):
```
piskvorky: Difference to `s.lstrip('\n')`?

gensim/models/utils_any2vec.py
```diff
@@ -192,7 +242,6 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
             for line in fin:
                 word, count = utils.to_unicode(line, errors=unicode_errors).strip().split()
                 counts[word] = int(count)
-
```
piskvorky: Keep the blank line please. It helps separate code blocks for readability.

gensim/models/utils_any2vec.py
```diff
@@ -202,43 +251,21 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
-                add_word(word, weights)
+            chunk = b''
+            tot_processed_words = 0
+
+            while tot_processed_words < vocab_size:
+                new_chunk = fin.read(binary_chunk_size)
+                chunk += new_chunk
+                max_words = vocab_size - len(result.vocab)
+                processed_words, chunk = __add_words_from_binary_chunk_to_result(result, counts, max_words,
+                                                                chunk, vector_size, datatype)
```
piskvorky: Hanging indent please (here and everywhere). No vertical indent in Gensim.

gensim/models/utils_any2vec.py
```diff
@@ -202,43 +251,21 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
+                tot_processed_words += processed_words
+                if len(new_chunk) < binary_chunk_size:
+                    break
+            if tot_processed_words != vocab_size:
+                raise EOFError("unexpected end of input; is count incorrect or file otherwise damaged?")
```
piskvorky: This seems too harsh.

If the user requests top 30k vocab, but the input file has only 20k word vectors in it, we shouldn't fail with an exception. They have no way of knowing how many words are inside a-priori.

We want to act like Python `[: n]` slicing, like in the original code: return the vocab prefix that exists.

piskvorky: Scratch that - I mixed up `limit` with `vocab_size`. The code still acts like `[: limit]`, right?

gensim/models/utils_any2vec.py
```diff
@@ -202,43 +251,21 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
-                # use count from the vocab file
-                result.vocab[word] = Vocab(index=word_id, count=counts[word])
-            else:
-                # vocab file given, but word is missing -- set count to None (TODO: or raise?)
-                logger.warning("vocabulary file is incomplete: '%s' is missing", word)
-                result.vocab[word] = Vocab(index=word_id, count=None)
-            result.vectors[word_id] = weights
-            result.index2word.append(word)
-
         if binary:
```
mpenkov: While we're doing this, I think it's worth to split out these things into separate functions, for readability:

```
if binary:
    _word2vec_read_binary(fin, ...)
else:
    _word2vec_read_text(fin, ...)
```

What do you guys think?

gensim/test/test_utils_any2vec.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2017 Radim Rehurek <me@radimrehurek.com>
```
mpenkov:
```suggestion
# Copyright (C) 2019 Radim Rehurek <me@radimrehurek.com>
```

gensim/models/utils_any2vec.py
```diff
@@ -176,14 +176,64 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
         Returns the loaded model as an instance of :class:`cls`.
 
     """
+
+    def __add_word_to_result(result, counts, word, weights):
+        word_id = len(result.vocab)
+        if word in result.vocab:
+            logger.warning("duplicate word '%s' in %s, ignoring all but first", word, fname)
+            return
+        if counts is None:
```
mpenkov: You're making the reader do a bit more reading then necessary. You could save your future readers some eyeball work by doing:

```python
if counts is None:
    # most common scenario: no vocab file given. just make up some bogus counts, in descending order
    word_count = vocab_size - word_id
elif word in counts:
    # use count from the vocab file
    word_count = counts[word]
else:
    logger.warning("vocabulary file is incomplete: '%s' is missing", word)
    word_count = None

result.vocab[word] = Vocab(index=word_id, count=word_count)
```

If it does not affect performance, you could also replace the elif-else with a try-catch, as that'd be more Pythonic.

gensim/models/utils_any2vec.py
```diff
@@ -176,14 +176,64 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
+        i = 0
+        while i < len(s) and s[i] == '\n':
+            i += 1
+        return s[i:]
+
+    def __add_words_from_binary_chunk_to_result(result, counts, max_words, chunk, vector_size, datatype):
+        start = 0
+        n = len(chunk)
+        processed_words = 0
+        n_bytes_per_vector = vector_size * dtype(REAL).itemsize
```
mpenkov: Shorter name, same meaning.

```suggestion
        bytes_per_vector = vector_size * dtype(REAL).itemsize
```

gensim/models/utils_any2vec.py
```diff
@@ -176,14 +176,64 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
+
+    def __remove_initial_new_line(s):
+        i = 0
+        while i < len(s) and s[i] == '\n':
+            i += 1
+        return s[i:]
+
+    def __add_words_from_binary_chunk_to_result(result, counts, max_words, chunk, vector_size, datatype):
+        start = 0
+        n = len(chunk)
```
mpenkov: I think `n` is a bit too cryptic here (it also clashes a bit with `n_bytes_per_vector` (is this intentional?).

```suggestion
        n = len(chunk)
```

You could call it `len_chunk` but I'm not sure if there's much benefit to using that over just `len(chunk)`.

gensim/models/utils_any2vec.py
```diff
@@ -176,14 +176,64 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
+        processed_words = 0
+        n_bytes_per_vector = vector_size * dtype(REAL).itemsize
+
+        for _ in range(0, max_words):
+            i_space = chunk.find(b' ', start)
+            i_vector = i_space + 1
+            if i_space != -1 and (n - i_vector) >= n_bytes_per_vector:
```
mpenkov: I think it'd be better for readability if you did:

```
if termination_condition:
    break

word = ...
vector = frombuffer(...)
# and so on
```

gensim/models/utils_any2vec.py
lopusz: This parameter was useful for testing. I used it in unit tests e.g. to make sure that code is correct in exotic cases like chunk being equal to word+vector block size, etc.

I did not surface it to more "user facing" functions such as `keyedvectors.KeyedVectors.load_word2vec_format`

gensim/models/utils_any2vec.py
lopusz: Good catch! For some queer reason i forgot that except strip there is also lstrip. Corrected.

gensim/models/utils_any2vec.py
lopusz: Corrected.

gensim/models/utils_any2vec.py
lopusz: Corrected!

gensim/models/utils_any2vec.py
lopusz: Yes. The code "glues" `limit` and `vocab_size` doing  
```
vocab_size = min(vocab_size, limit)
```
(This was the original behaviour)

`vocab_size` is known from header, so file should contain at least as `vocab_size` vectors (even if user specifies strange limit).
 
There is a test  in `gensim` which check if this function throws an exception in the too short file case

https://github.com/RaRe-Technologies/gensim/blob/3d6596112f8f1fc0e839a32c5a00ef3d7365c264/gensim/test/test_word2vec.py#L426

gensim/models/utils_any2vec.py
```diff
@@ -176,14 +176,64 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
+            i += 1
+        return s[i:]
+
+    def __add_words_from_binary_chunk_to_result(result, counts, max_words, chunk, vector_size, datatype):
+        start = 0
+        n = len(chunk)
+        processed_words = 0
+        n_bytes_per_vector = vector_size * dtype(REAL).itemsize
+
+        for _ in range(0, max_words):
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/models/utils_any2vec.py
```diff
@@ -176,14 +176,64 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
+        result.vectors[word_id] = weights
+        result.index2word.append(word)
+
+    def __remove_initial_new_line(s):
+        i = 0
+        while i < len(s) and s[i] == '\n':
+            i += 1
+        return s[i:]
+
+    def __add_words_from_binary_chunk_to_result(result, counts, max_words, chunk, vector_size, datatype):
```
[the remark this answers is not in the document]
lopusz: Simplified.

gensim/models/utils_any2vec.py
lopusz: I think this is a good change. Refactored binary and text parts to a separate functions.

gensim/test/test_utils_any2vec.py
lopusz: Corrected.

gensim/models/utils_any2vec.py
lopusz: Corrected.

gensim/models/utils_any2vec.py
lopusz: Good catch. Changed to `len(chunk)` call.

gensim/models/utils_any2vec.py
lopusz: Definitely agree. Corrected.

gensim/models/utils_any2vec.py
lopusz: This was unchanged by my patches, however, your version of the code definitely reads better.
Included.

lopusz:
Hi Radim, Hi Michael,

thank you for very quick & detailed review and + a lot of  suggestions.
I cleaned up the code according to your hints.

Do you think this is now mergeable?

Best regards,
Michał

piskvorky:
I wish all our PRs were this nice and clean! 🌟

PR #2676, opened 2019-11-10
fixed get_keras_embedding, now accepts word mapping

Hamekoded:
This is my first PR, feedback is welcome.

This fixes the Keras integration code, by allowing the get_keras_embedding function to accept word index parameter. Also fixes the example notebook.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -1412,7 +1412,7 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
 
         Warnings
         --------
-        Current method work only if `Keras <https://pypi.org/project/Keras/>`_ installed.
+        Current method works only if `Keras <https://pypi.org/project/Keras/>`_ installed.
 
         """
         try:
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -1394,7 +1394,7 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
 
         word_index : {str : int}
             A mapping from tokens to their indices the way they will be provided in the input to the embedding layer.
-            The embedding of each token will be placed in the corresponding index in the embedding matrix.
+            The embedding of each token will be placed at the corresponding index in the returned matrix.
             Tokens not in the index are ignored and not placed in the returned layer embedding matrix.
             This is useful when the token indices are produced by a process that is not coupled with the embedding
             model, e.x. an Keras Tokenizer object.
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -1428,7 +1428,6 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
                 if word in self.vocab:
                     weights[index] = self.get_vector(word)
 
-        # set `trainable` as `False` to use the pretrained word embedding
         layer = Embedding(
             input_dim=weights.shape[0], output_dim=weights.shape[1],
             weights=[weights], trainable=train_embeddings
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -1383,7 +1383,7 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
 
         return rcs
 
-    def get_keras_embedding(self, train_embeddings=False, word_index = None):
+    def get_keras_embedding(self, train_embeddings=False, word_index=None):
         """Get a Keras 'Embedding' layer with weights set as the Word2Vec model's learned word embeddings.
 
         Parameters
@@ -1434,7 +1434,6 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
             weights=[weights], trainable=train_embeddings
         )
         return layer
-
 
 
 class WordEmbeddingSimilarityIndex(TermSimilarityIndex):
@@ -1549,7 +1548,6 @@ class Word2VecKeyedVectors(WordEmbeddingsKeyedVectors):
         return _load_word2vec_format(
             cls, fname, fvocab=fvocab, binary=binary, encoding=encoding, unicode_errors=unicode_errors,
             limit=limit, datatype=datatype)
-
 
     @classmethod
     def load(cls, fname_or_handle, **kwargs):
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -1383,6 +1383,59 @@ class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
 
         return rcs
 
+    def get_keras_embedding(self, train_embeddings=False, word_index = None):
+        """Get a Keras 'Embedding' layer with weights set as the Word2Vec model's learned word embeddings.
+
+        Parameters
+        ----------
+        train_embeddings : bool
+            If False, the weights are frozen and stopped from being updated.
+            If True, the weights can/will be further trained/updated.
+
+        word_index : {str : int}
+            A mapping from tokens to their indices the way they will be provided in the input to the embedding layer.
+            The embedding of each token will be placed in the corresponding index in the embedding matrix.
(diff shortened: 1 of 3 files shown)


mpenkov:
Thank you for your contribution.

Can you please make it more obvious what you're trying to achieve here? Is there an existing issue that you're solving?

Also, it looks like the tests fail. Please investigate and make them pass. Give me a yell if you can't figure it out after a couple of hours of effort.

Hamekoded:
This solves the issue described in this thread: https://groups.google.com/forum/#!topic/gensim/24tfamYOzW4
Though this is not strictly a bug, the current design makes it very un-intuitive to use it properly. It requires the user to numerate the tokens according to the vocabulary of the Word2Vec object, which is not really obvious and also requires boilerplate. The example notebook also showed the incorrect use of the get_keras_embedding function.

I think test fail because they try to import keras and fail. The existing test case skips the tests when it fails to import keras, but that means that the keras integration is not tested during ci. So I'm not sure what to do here

mpenkov:
You have two options to get the tests to pass:

1. Disable the tests when keras is not installed (using `unittest.skipIf(not KERAS_IS_INSTALLED)` or something similar. This was the previous behavior.
2. Install keras into the test environment (see setup.py and tox.ini)

Ideally, we'd go with 2), unless there's a compelling reason not to install keras during CI runs. Is there such a reason?

Hamekoded:
I don't have a compelling reason for not importing keras. However, I see this piece of code in setup.py
``` 
if sys.version_info < (3, 7):
    linux_testenv.extend([
        'tensorflow <= 1.3.0',
        'keras >= 2.0.4, <= 2.1.4',
    ])
```

so I imagine there's a reason this is added for only specific versions and just for linux env.

gensim/test/test_keyedvectors.py
```diff
@@ -400,32 +400,42 @@ def test_load_model_and_vocab_file_ignore(self):
...
     @patch('numpy.random.normal')
     def test_get_keras_embedding_word_index_passed_with_oov_word(self, normal_func):
         normal_func.return_value = np.zeros((3, self.vectors.vectors.shape[1]))
         word_index = {'is': 1, 'not_a_real_word': 2}
         embedding_layer = self.vectors.get_keras_embedding(word_index=word_index)
-        self.assertEqual( embedding_layer._initial_weights[0].shape, (3, self.vectors.vectors.shape[1]))
+        self.assertEqual(embedding_layer._initial_weights[0].shape, (3, self.vectors.vectors.shape[1]))
         self.assertTrue(np.array_equal(embedding_layer._initial_weights[0][2, :],
-                        np.zeros(self.vectors.vectors.shape[1])))
+                                       np.zeros(self.vectors.vectors.shape[1])))
```
piskvorky: Hanging indent please, here and everywhere. No vertical indent in Gensim.

gensim/test/test_keyedvectors.py
```diff
@@ -400,6 +400,45 @@ def test_load_model_and_vocab_file_ignore(self):
...
+except ImportError:
+    keras_installed = False
+
+
+class WordEmbeddingsKeyedVectorsTest(unittest.TestCase):
+    def setUp(self):
+        self.vectors = EuclideanKeyedVectors.load_word2vec_format(
+            datapath('euclidean_vectors.bin'), binary=True, datatype=np.float64)
+
+    @unittest.skipIf(not keras_installed, 'keras needs to be installed for this test')
```
mpenkov: Please move this decorator from the method to the class. If keras is not installed, none of the tests under the WordEmbeddingsKeyedVectorsTest should run, right?

gensim/test/test_keyedvectors.py
```diff
@@ -400,6 +400,43 @@ def test_load_model_and_vocab_file_ignore(self):
             model.get_vector(u'どういたしまして'), np.array([.1, .2, .3], dtype=np.float32)))
 
 
+try:
+    import keras  # noqa: F401
+
+    keras_installed = True
```
mpenkov:
```suggestion
    KERAS_INSTALLED = True
```

gensim/models/keyedvectors.py
```diff
@@ -1383,6 +1383,58 @@ def relative_cosine_similarity(self, wa, wb, topn=10):
...
+        if word_index is None:
+            weights = self.vectors
+        else:
+            max_index = max(word_index.values())
+            weights = np.random.normal(size=(max_index + 1, self.vectors.shape[1]))
+            for word, index in word_index.items():
+                if word in self.vocab:
+                    weights[index] = self.get_vector(word)
+
+        # set `trainable` as `False` to use the pretrained word embedding
```
mpenkov: Who is this comment for?

gensim/models/keyedvectors.py
```diff
@@ -1383,6 +1383,58 @@ def relative_cosine_similarity(self, wa, wb, topn=10):
...
+
+        Parameters
+        ----------
+        train_embeddings : bool
+            If False, the weights are frozen and stopped from being updated.
+            If True, the weights can/will be further trained/updated.
+
+        word_index : {str : int}
+            A mapping from tokens to their indices the way they will be provided in the input to the embedding layer.
+            The embedding of each token will be placed in the corresponding index in the embedding matrix.
```
```suggestion
            The embedding of each token will be placed at the corresponding index in the returned matrix.
```

gensim/models/keyedvectors.py
```diff
@@ -1383,6 +1383,58 @@ def relative_cosine_similarity(self, wa, wb, topn=10):
...
+            Tokens not in the index are ignored and not placed in the returned layer embedding matrix.
```
```suggestion
            Tokens not in the index are ignored.
```

gensim/models/keyedvectors.py
```diff
@@ -1383,6 +1383,58 @@ def relative_cosine_similarity(self, wa, wb, topn=10):
...
+            Embedding layer.
+
+        Raises
+        ------
+        ImportError
+            If `Keras <https://pypi.org/project/Keras/>`_ not installed.
+
+        Warnings
+        --------
+        Current method work only if `Keras <https://pypi.org/project/Keras/>`_ installed.
```
```suggestion
        Current method works only if `Keras <https://pypi.org/project/Keras/>`_ installed.
```

gensim/models/keyedvectors.py
Hamekoded: This was just copied from the original code, I just moved it to a new file.

gensim/test/test_keyedvectors.py
Hamekoded: I'm fine with the uppercase but this need to be changed in the rest of the file. Can this be done with this "commit suggestion" feature?

mpenkov: No, unfortunately that part has to be done manually. Are you able to do it?

Hamekoded: Sure, I'll change it

piskvorky:
@Hamekoded the tests are failing now, we have a red build badge:
FAILED gensim/test/test_keyedvectors.py::WordEmbeddingsKeyedVectorsTest::test_get_keras_embedding_word_index_none
FAILED gensim/test/test_keyedvectors.py::WordEmbeddingsKeyedVectorsTest::test_get_keras_embedding_word_index_passed
FAILED gensim/test/test_keyedvectors.py::WordEmbeddingsKeyedVectorsTest::test_get_keras_embedding_word_index_passed_with_oov_word

Can you fix this please?

Hamekoded:
it looks like the import statement alias changed in this PR (https://github.com/RaRe-Technologies/gensim/pull/2761/files). @mpenkov what was the reason for dropping the alias?

mpenkov:
@Hamekoded see [this discussion](https://github.com/RaRe-Technologies/gensim/pull/2761#discussion_r386065188).

Hamekoded:
So is there a way to just change the name of the class to the one without the alias, without a new PR? It would take me some time to get a new PR in place

mpenkov:
Addressed in https://github.com/RaRe-Technologies/gensim/pull/2774

PR #2678, opened 2019-11-12
Update word2vec2tensor.py

kaddynator:
Just a grammatical mistake.


mpenkov:
Good catch, thank you!

And congratulations on your first contribution to gensim :1st_place_medal: !

PR #2682, opened 2019-11-20
Fix local import degrading the performance of word2vec model loading …

lopusz:
Fix local import degrading the performance of word2vec model loading (#2671)

### DIFF
--- a/gensim/models/utils_any2vec.py
+++ b/gensim/models/utils_any2vec.py
@@ -27,6 +27,7 @@ current Facebook implementation, use :py:func:`compute_ngrams_bytes` and
 
 import logging
 from gensim import utils
+import gensim.models.keyedvectors
 
 from numpy import zeros, dtype, float32 as REAL, ascontiguousarray, frombuffer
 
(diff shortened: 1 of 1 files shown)


lopusz:
Hi @piskvorky , do you think we can merge this tiny update?

All cheks pass and the update reduces CO2 footprint of word2vec loading... ;)

piskvorky:
In that case let's leave it open! CO2 is vital for life. 8)

Leaving for @mpenkov to review & merge if ready, LGTM.

mpenkov:
I wonder why the original author chose a local import there... and also whether there are other places where we suffer from a similar problem.

Just grepping around, we seem to be importing things outside of module scope (or at least beyond the root level of indentation) quite frequently:

```
$ grep -R "^    " gensim | grep " import " | grep -v ">>>" | cut -d: -f1 | sort | uniq | wc -l
49
```

So, there are 49 modules we may want to review for imports. WDYT @piskvorky ?

piskvorky:
I tracked that import down to the dreadful #1777 by @manneshiva and @janpom . So, how knows.

I expect in many cases these indented imports will be due to `try: except` blocks, or to avoid costly imports or 3rd party (optional) dependencies. But in cases like the one here, I'm +1 on fixing this globally of course.

Issue #2684, opened 2019-11-23
Unreasonable Query Result


JiaqiLiu:
#### Problem description

      **The query result seems not correct. The code is self-explained. Thank you!**

#### Steps/code/corpus to reproduce

Include full tracebacks, logs and datasets if necessary. Please keep the examples minimal ("minimal reproducible example").

```
from gensim.summarization.bm25 import BM25, get_bm25_weights


text1 = "A constellation is a group of stars that are considered to form imaginary outlines or meaningful patterns on the celestial sphere."
text2 = "The 88 modern constellations are formally defined regions of the sky together covering the entire celestial sphere."
text = [text1, text2]

corpus = [text1.split(" "), text2.split(" ")]
print(f'corpus: {corpus}')

query = text2.split(" ")

bm25 = BM25(corpus)
scores = bm25.get_scores(query)
scores = [(s, i) for i, s in enumerate(scores)]
scores.sort(key=lambda t: t[0], reverse=True)
print(f'scores:         {scores}')

for s, idx in scores:
  print(f'{s}\t{idx}: {text[idx]}')
```

Output:

```
-0.3601521710456333         0: A constellation is a group of stars that are considered to form imaginary outlines or meaningful patterns on the celestial sphere.
-0.44989406787023367     1: The 88 modern constellations are formally defined regions of the sky together covering the entire celestial sphere.
```

#### Versions

Please provide the output of:

```python
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import gensim; print("gensim", gensim.__version__)
from gensim.models import word2vec;print("FAST_VERSION", word2vec.FAST_VERSION)
```

Output:
```
macOS-10.14.6-x86_64-i386-64bit
Python 3.8.0 (default, Nov  6 2019, 15:49:01)
[Clang 4.0.1 (tags/RELEASE_401/final)]
NumPy 1.17.4
SciPy 1.3.3
gensim 3.8.1
FAST_VERSION 0
```

piskvorky:
@Witiko could you please have a look?

Witiko:
This issue seems to be due to the inverse document frequency term in the BM25 formula.
Since the corpus is so small, most words have zero or negative inverse document frequency.

Witiko:
[...]

PR #2687, opened 2019-11-26
Warn when BM25.average_idf < 0

Witiko:
When the BM25 model is trained on a corpus that is too small or artificially created from non-textual data, many words can occur in more than a half of the entire corpus, making the average inverse document frequency (IDF) negative. This breaks the assumptions of BM25 and may produce confusing results, see #2684. This PR adds a warning that informs the user that a corpus breaks the assumptions of BM25:
```python
>> import logging
>>
>> logging.basicConfig(level=logging.INFO)
>>
>> from gensim.summarization.bm25 import BM25
>>
>> text1 = \
>>     'A constellation is a group of stars that are considered to form imaginary outlines'
>>     ' or meaningful patterns on the celestial sphere.'
>> text2 = \
>>     'The 88 modern constellations are formally defined regions of the sky together'
>>     ' covering the entire celestial sphere.'
>> corpus = [text1.split(' '), text2.split(' ')]
>> bm25 = BM25(corpus)
```
    WARNING:gensim.summarization.bm25:Average inverse document frequency is less than zero.
    Your corpus of 2 documents is either too small or it does not originate from natural text.
    BM25 will likely produce unintuitive results.

### DIFF
--- a/gensim/summarization/bm25.py
+++ b/gensim/summarization/bm25.py
@@ -37,6 +37,7 @@ Data:
 """
 
 
+import logging
 import math
 from six import iteritems
 from six.moves import range
@@ -47,6 +48,8 @@ from ..utils import effective_n_jobs
 PARAM_K1 = 1.5
 PARAM_B = 0.75
 EPSILON = 0.25
+
+logger = logging.getLogger(__name__)
 
 
 class BM25(object):
@@ -115,6 +118,13 @@ class BM25(object):
             if idf < 0:
                 negative_idfs.append(word)
         self.average_idf = float(idf_sum) / len(self.idf)
+
+        if self.average_idf < 0:
+            logger.warning(
+                'Average inverse document frequency is less than zero. Your corpus of {} documents'
+                ' is either too small or it does not originate from natural text. BM25 may produce'
+                ' unintuitive results.'.format(self.corpus_size)
+            )
 
         eps = EPSILON * self.average_idf
         for word in negative_idfs:


gensim/summarization/bm25.py
```diff
@@ -116,6 +119,13 @@ def _initialize(self, corpus):
                 negative_idfs.append(word)
         self.average_idf = float(idf_sum) / len(self.idf)
 
+        if self.average_idf < 0:
+            logger.warning(
+                'Average inverse document frequency is less than zero. Your corpus of {} documents'
+                ' is either too small or it does not originate from actual text documents. BM25'
+                ' will likely produce "wrong" results.'.format(self.corpus_size)
```
piskvorky: This `"wrong"` sounds almost mystical :) Any way we can make this more concrete, avoid the scare quotes?

gensim/summarization/bm25.py
```diff
@@ -116,6 +119,13 @@ def _initialize(self, corpus):
                 negative_idfs.append(word)
         self.average_idf = float(idf_sum) / len(self.idf)
 
+        if self.average_idf < 0:
```
piskvorky: What does IDF == 0 mean, is that edge case OK?

Witiko: Average IDF < 0 iff the product of (N - n_w) / n_w is less than one, where N is the corpus size and n_w is the number of documents containing word w. Due to the Zipf's empirical law, this should not happen for large enough corpora (few words are very common, most words occur rarely).

Words w with IDF_w < 0 will contribute negatively to the BM25 score, i.e. documents with more words w in common with the query will receive smaller BM25 score. This seems “wrong”, but it is actually what we want when we are dealing with stopwords.

gensim/summarization/bm25.py
Witiko: Rigor does not always improve clarity. See my other comment and let me know if you have an idea how to better rephrase this.

gensim/summarization/bm25.py
piskvorky: > Average IDF < 0 iff the product of (N - n_w) / n_w is less than one

You mean average over all such products = averaged over all `_w`?

Anyway, my question was around the edge case `== 0` (as opposed to `< 0`). It's probably not important, just wondering whether we can capture more "weird malformed inputs" by extending the inequality.

gensim/summarization/bm25.py
piskvorky:
```suggestion
                ' will likely produce unintuitive results.'.format(self.corpus_size)
```

gensim/summarization/bm25.py
Witiko: > > Average IDF < 0 iff the product of (N - n_w) / n_w is less than one
> 
> You mean average over all such products = averaged over all `_w`?

Average_w IDF_w < 0
iff Average_w log (N - n_w) / n_w < 0
iff Product_w (N - n_w) / n_w < 1.

> Anyway, my question was around the edge case == 0 (as opposed to < 0). It's probably not
> important, just wondering whether we can capture more "weird malformed inputs" by extending
> the inequality. "Zeros" are often interesting edge cases (empty inputs, empty dictionaries etc).

IDF_w = 0 for words w that occur precisely in one half of the corpus. There are a lot of such words in the example from the original post:
```python
>>> bm25.average_idf
-0.2514746738178282
>>> bm25.idf
{'A': 0.0, 'constellation': 0.0, 'is': 0.0, 'a': 0.0, 'group': 0.0, 'of': -0.06286866845445704, 'stars': 0.0, 'that': 0.0, 'are': -0.06286866845445704, 'considered': 0.0, 'to': 0.0, 'form': 0.0, 'imaginary': 0.0, 'outlines': 0.0, 'or': 0.0, 'meaningful': 0.0, 'patterns': 0.0, 'on': 0.0, 'the': -0.06286866845445704, 'celestial': -0.06286866845445704, 'sphere.': -0.06286866845445704, 'The': 0.0, '88': 0.0, 'modern': 0.0, 'constellations': 0.0, 'formally': 0.0, 'defined': 0.0, 'regions': 0.0, 'sky': 0.0, 'together': 0.0, 'covering': 0.0, 'entire': 0.0}
```
You may notice that the mean of `bm25.idf` is not `bm25.average_idf`. That's because BM25 floors IDF_w by setting all negative IDF_w to EPSILON times average IDF, where EPSILON is 0.25.

Witiko: On a related note, some of us would appreciate support for inline math in GitHub comments.

piskvorky: Great, thank you for the detailed analysis @Witiko

Witiko:
GitHub says Travis is pending, but when you go to Travis, thhe CI build successfully completed. Go figure.

Issue #2574, opened 2019-08-09
Dictionary gensim.corpora should update the collection frequencies cfs after filtering with filter_extremes


aless80:
#### Problem description

I am using the Dictionary class gensim.corpora.dictionary.Dictionary , in particular the filter_extremes method and the cfs property (returning a collection frequencies dictionary mapping token_id to tokenfrequency). 
It appears that after using  filter_extremes, the cfs still shows unfiltered token_id's can be absent from the filtered Dictionary instance. This can lead to KeyErrors. 

#### Steps to reproduce

```python
from gensim.corpora import Dictionary
corpus = [['common','single'],['common']]
dct = Dictionary(corpus)
len(dct)  #2
dct.cfs   #{0: 2, 1: 1}
dct.filter_extremes(no_below=1, no_above=0.5, keep_n=1)
len(dct)  #1
dct.cfs   #{1: 1}
dct[1]
```

The last line fails with 'KeyError: 1', as the filtered Dictionary has only one element. I suspect the cfs dictionary should be updated once the Dictionary instance dct gets filtered by filter_extremes. The documentation has this warning for the filter_extremes method, but I do not think it is relevant for this case: "After the pruning, resulting gaps in word ids are shrunk. Due to this gap shrinking, the same word may have a different word id before and after the call to this function!"

#### Versions

Please provide the output of:

```python
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import gensim; print("gensim", gensim.__version__)
from gensim.models import word2vec;print("FAST_VERSION", word2vec.FAST_VERSION)

Linux-4.15.0-50-generic-x86_64-with-LinuxMint-19-tara
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0]
NumPy 1.16.3
SciPy 0.19.1
gensim 3.8.0
FAST_VERSION 1
```

mpenkov:
Thank you for reporting. Looks like a bug to me. Are you able to make a PR?

aless80:
Not at the moment

EinKaiser:
Hi, is this issue still open?

mpenkov:
Yes, it is.

PR #2695, opened 2019-12-04
Update issue2574

EinKaiser:
Could you please explain how this is related to #2574?

What is the motivation behind this PR? What are you trying to achieve?

### DIFF
--- a/test_corpora_dictionary_2574.py
+++ b/test_corpora_dictionary_2574.py
@@ -1,0 +1,375 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+
+"""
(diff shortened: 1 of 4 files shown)


test_corpora_dictionary_2574.py
```diff
@@ -0,0 +1,375 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
```
mpenkov: @piskvorky What should we do about the license header in contributed files?

piskvorky: Keep our standard header, everywhere.

And if a contributor doesn't agree with the copyright assignment, we cannot accept the contribution (too much headache).

EinKaiser: Sorry about the license header mix-up. Ill close this PR.

Issue #2578, opened 2019-08-14
"OverflowError: value too large to convert to int" when training word2vec on a large corpus


miweru:
Hi,
i am trying to train word2vec on on a large corpus (>500GB) and the newest gensim version (3.8.0) and an error occurs on every Thread:

`Exception in thread Thread-10:
Traceback (most recent call last):
  File "/home/michael/miniconda3/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/home/michael/miniconda3/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/home/michael/miniconda3/lib/python3.7/site-packages/gensim/models/base_any2vec.py", line 175, in _worker_loop_corpusfile
    total_examples=total_examples, total_words=total_words, **kwargs)
  File "/home/michael/miniconda3/lib/python3.7/site-packages/gensim/models/word2vec.py", line 794, in _do_train_epoch
    total_examples, total_words, work, neu1, self.compute_loss)
  File "gensim/models/word2vec_corpusfile.pyx", line 379, in gensim.models.word2vec_corpusfile.train_epoch_cbow
OverflowError: value too large to convert to int`

My command for training the Model is:
`Word2Vec(corpus_file="encc_tokenized", size = 1024, window = 8, workers = 16)`

Thank you for help

piskvorky:
CC @persiyanov – is this expected?

piskvorky:
Looking at the code, it's indeed using `int` for all sort of counter that should be large: `expected_examples`, `effective_words`, `effective_sentences`, `total_sentences`, indexes… This will need fixing :(

Or was there some reason to using only `int`s @persiyanov ?

miweru:
The same problem occurs also with FastText:
`Exception in thread Thread-8:
Traceback (most recent call last):
  File "/home/michael/miniconda3/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/home/michael/miniconda3/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/home/michael/miniconda3/lib/python3.7/site-packages/gensim/models/base_any2vec.py", line 175, in _worker_loop_corpusfile
    total_examples=total_examples, total_words=total_words, **kwargs)
[...]

PR #2700, opened 2019-12-10
Fix overflow error for `*Vec` corpusfile-based training

persiyanov:
Fix for https://github.com/RaRe-Technologies/gensim/issues/2578

Previous fix (https://github.com/RaRe-Technologies/gensim/pull/2239/files) solved the problem for datasets that have number of `words` exceeding max int. However, there were some users who used far bigger datasets which exceeded max int even in terms of `examples`.


persiyanov:
cc @piskvorky @mpenkov Sorry for the delay guys, I was bogged down with other stuff. But the fix for https://github.com/RaRe-Technologies/gensim/issues/2578 is finally here :)

mpenkov:
@persiyanov Thank you for looking into this. Can you investigate why the AppVeyor tests are failing?

persiyanov:
@mpenkov For python 3.5, the error is ` ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject`

For python 3.6, there are other runtime errors like `_configtest.obj : error LNK2001: unresolved external symbol acosl`.

I presume it is related to how I generated new C/C++ files with cython. I used python 3.7.4 and usual cython command like `cython --cplus gensim/models/fasttext_corpusfile.pyx`. As of now, what is the most proper way to regenerate C/C++ files? I think that command worked few years ago.

mpenkov:
I expect #2706 to fix this.

mpenkov:
@persiyanov Please merge gensim/develop into your branch to fix the appveyor issues.

persiyanov:
@mpenkov Done!

gojomo:
This shouldn't have added the `*.cpp` files into git, should it? (They should be fully deterministic byproducts of the `.pyx` files, and our builds/releases had been working for quite a while without any `.cpp`/`.c` files from developers' individual builds in the source tree.) @persiyanov @mpenkov

PR #2706, opened 2019-12-21
Fix appveyor builds

mpenkov:
We were installing older versions of numpy/scipy because we wanted wheels for all of our Python versions and platforms, which originally included 2.7. This was causing some recent AppVeyor builds to fail, e.g. https://github.com/RaRe-Technologies/gensim/pull/2700.

Now that we've dropped support for 2.7 in future builds, we can enjoy the use of newer numpy/scipy versions, as they all target 3.5 and above with wheels for all platforms.

The `install_wheels.py` script is mostly redundant now, since it is no longer Python version-dependent, but we may want to keep it around for a little while longer in case the dependencies become tricky. The new `check_wheels.py` script is helpful for identifying versions with wheels.


continuous_integration/check_wheels.py
```diff
@@ -0,0 +1,33 @@
+"""Print available wheels for a particular Python package."""
```
piskvorky: Missing file header. Especially if script taken / adapted from elsewhere (must have a licence compatible with LGPL).

continuous_integration/check_wheels.py
```diff
@@ -0,0 +1,33 @@
...
+import re
+import sys
+
+import requests
```
piskvorky: Where does `requests` come from?

continuous_integration/install_wheels.py
```diff
@@ -0,0 +1,52 @@
+"""Install wheels for numpy and scipy.
```
piskvorky: Missing file header.

continuous_integration/appveyor/requirements.txt
```diff
@@ -1,18 +1,9 @@
...
-# wheel for windows we ca delete this --find-links line.
---find-links http://28daf2247a33ed269873-7b1aad3fab3cc330e1fd9d109892382a.r6.cf2.rackcdn.com/
-
-# fix the versions of numpy to force the use of numpy and scipy to use the whl
-# of the rackspace folder instead of trying to install from more recent
-# source tarball published on PyPI
-numpy==1.11.3
-scipy==0.18.1
+#
+# /continuous_integration/install_wheels.py takes care of installing numpy and scipy.
```
menshikh-iv: are you sure than this still required? because you get rid py27 (if I remember correctly), I guess it should work without "hacks"

continuous_integration/install_wheels.py
```diff
@@ -0,0 +1,52 @@
...
+the current Python version.
+
+We use this when building/testing gensim in a CI environment (Travis, AppVeyor,
+etc).
+
+The versions we currently target are:
+
+    - 3.5 (AppVeyor, TravisCI)
+    - 3.6 (AppVeyor, TravisCI)
+    - 3.7 (AppVeyor, TravisCI, CircleCI)
```
menshikh-iv: what's about adding 3.8?

continuous_integration/install_wheels.py
```diff
@@ -0,0 +1,52 @@
...
+import subprocess
+import sys
+
+
+def main():
+    #
+    # We don't support Py2 anymore, so the most recent versions of both
+    # numpy and scipy have what we need.
+    #
+    packages = ['numpy==1.17.4', 'scipy==1.4.1']
```
menshikh-iv: why these versions? Maybe use minimal based on `requirements.txt`?

gojomo:
#2715 overlaps a bit with this - going further in removing py3.5, `install_wheels.py`, `appveyor/requirements.txt` (which all work fine), and in adding py3.8.  

(But maybe this `check_wheels.py` can help diagnose why Appveyor-py3.8 is insisting on building cython/numpy?)

continuous_integration/check_wheels.py
mpenkov: It's an indirect dependency of gensim (via `smart_open`).

We don't explicitly list it anywhere like setup.py because this is a utility script.

continuous_integration/install_wheels.py
mpenkov: Let's deal with that in a separate PR. I just want these builds to start working again. Currently, the failing builds are blocking work from other people from being merged in.

continuous_integration/appveyor/requirements.txt
mpenkov: You're right. I've removed this install_wheels bollocks altogether. If it works, we can "just" build with the latest versions of scipy and numpy, and expect them to have wheels available. Let's see if it builds.

continuous_integration/install_wheels.py
mpenkov: I'm not specifying the versions at all anymore. `numpy` and `scipy` get installed like all the other packages do (with `setup.py`).

mpenkov:
@piskvorky I'd like to merge this ASAP so we can start merging other PRs. They are currently blocked because the appveyor builds are failing.

Is this OK to merge?

continuous_integration/check_wheels.py
```diff
@@ -0,0 +1,39 @@
+#
```
piskvorky:
```suggestion
#!/usr/bin/env python
```

continuous_integration/check_wheels.py
```diff
@@ -0,0 +1,39 @@
...
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 RaRe Technologies s.r.o.
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+#
```
piskvorky:
```suggestion

```

piskvorky:
@mpenkov I added two minor header fixes (I cannot push to this PR). Otherwise looks good!

PR #2712, opened 2019-12-27
Implement saving to Facebook format

### LINKED ISSUE #2611: Implement saving to Facebook format
We currently support reading FastText models from Facebook's format. The [gensim.models._fasttext_bin](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/_fasttext_bin.py) does this.

This enables people to use gensim with a model that was trained using Facebook's binaries.

Sometimes, people want things to work the other way: they start with gensim, train a model, and then want to save it to Facebook's format.

For this ticket, you will implement a `save(model, fout)` function that accepts a FastText object and saves it to a file stream in a Facebook-compatible format. It will essentially reverse the effects of the [load](https://github.com/RaRe-Technologies/gensim/blob/f89808d52d0250e4e4bbab2293980f8f4d3989b9/gensim/models/_fasttext_bin.py#L291) function.

lopusz:
Fix #2611

### DIFF
--- a/gensim/models/_fasttext_bin.py
+++ b/gensim/models/_fasttext_bin.py
@@ -81,13 +81,13 @@ _OLD_HEADER_FORMAT = [
     ('epoch', 'i'),
     ('min_count', 'i'),
     ('neg', 'i'),
-    ('_', 'i'),
+    ('word_ngrams', 'i'), # Unused in loading
     ('loss', 'i'),
     ('model', 'i'),
     ('bucket', 'i'),
     ('minn', 'i'),
     ('maxn', 'i'),
-    ('_', 'i'),
+    ('lr_update_rate', 'i'), # Unused in loading
     ('t', 'd'),
 ]
 
@@ -185,6 +185,7 @@ def _load_vocab(fin, new_format, encoding='utf-8'):
         The loaded vocabulary.  Keys are words, values are counts.
         The vocabulary size.
         The number of words.
+        The numnber of tokens.
     """
     vocab_size, nwords, nlabels = _struct_unpack(fin, '@3i')
 
@@ -334,7 +335,6 @@ def load(fin, encoding='utf-8', full_model=True):
     header_spec = _NEW_HEADER_FORMAT if new_format else _OLD_HEADER_FORMAT
     model = {name: _struct_unpack(fin, fmt)[0] for (name, fmt) in header_spec}
 
-
     if not new_format:
         model.update(dim=magic, ws=version)
 
@@ -392,19 +392,19 @@ def _sign_model(fout):
     fout.write(_FASTTEXT_VERSION.tobytes())
 
 
-def _conv_field_to_bytes(value, field_type):
+def _conv_field_to_bytes(field_value, field_type):
     if field_type == 'i':
-        return (np.int32(value).tobytes())
+        return (np.int32(field_value).tobytes())
     elif field_type == 'd':
-        return (np.float64(value).tobytes())
+        return (np.float64(field_value).tobytes())
     else:
-        raise NotImplementedError('Currently conversion to "%s" type is not implemmented.' % type)
+        raise NotImplementedError('Currently conversion to "%s" type is not implemmented.' % field_type)
 
 
-def _get_field(model, field, field_type):
+def _get_field_from_model(model, field):
 
     if field == 'bucket':
-        res = model.bucket
+        res = model.trainables.bucket
     elif field == 'dim':
         res = model.vector_size
     elif field == 'epoch':
@@ -423,7 +423,7 @@ def _get_field(model, field, field_type):
             # TODO Is this right ?!?
             res = 2
     elif field == 'maxn':
-        res = model.wv.min_n
+        res = model.wv.max_n
     elif field == 'minn':
         res = model.wv.min_n
     elif field == 'min_count':
@@ -449,15 +449,19 @@ def _get_field(model, field, field_type):
     else:
         raise NotImplementedError('Extraction of header field "%s" from Gensim FastText object not implemmented.' % field)
 
-    return _conv_field_to_bytes(res, field_type)
+    return res
 
 
-def _args_save(fout, model):
+def _args_save(fout, model, fb_fasttext_parameters):
 
     # Reimplementation of the [Args::save](https://github.com/facebookresearch/fastText/blob/master/src/args.cc)
 
     for field, field_type in _NEW_HEADER_FORMAT:
-        fout.write(_get_field(model, field, field_type))
+        if field in fb_fasttext_parameters:
+            field_value = fb_fasttext_parameters[field]
+        else:
+            field_value = _get_field_from_model(model, field)
+        fout.write(_conv_field_to_bytes(field_value, field_type))
 
 
 def _dict_save(fout, model, encoding):
@@ -494,13 +498,14 @@ def _dict_save(fout, model, encoding):
     # prunedidx_size_=-1, -1 value denotes no prunning index (prunning is only supported in supervised mode)
     fout.write(np.int64(-1))
 
-    for word, vocab_entry in model.wv.vocab.items():
+    for word in model.wv.index2word:
+        word_count = model.wv.vocab["the"].count
         fout.write(word.encode(encoding))
         fout.write(_END_OF_WORD_MARKER)
-        fout.write(np.int64(vocab_entry.count).tobytes())
+        fout.write(np.int64(word_count).tobytes())
         fout.write(_DICT_WORD_ENTRY_TYPE_MARKER)
 
-    # We are in unsupervised case, therefore pruned_idx is empty. so we do not need to write anything else
+    # We are in unsupervised case, therefore pruned_idx is empty, so we do not need to write anything else
 
 
 def _input_save(fout, model):
@@ -531,7 +536,7 @@ def _output_save(fout, model):
     fout.write(hidden_output.tobytes())
 
 
-def _save(fout, model, encoding):
+def _save(fout, model, fb_fasttext_parameters, encoding):
 
     # Unfortunatelly there is no documentation of the FB binary format
     # This is just reimplementation of
@@ -550,87 +555,24 @@ def _save(fout, model, encoding):
     #   ```
 
     _sign_model(fout)
-    _args_save(fout, model)
+    _args_save(fout, model, fb_fasttext_parameters)
     _dict_save(fout, model, encoding)
-    fout.write(struct.pack('@?', False))  # TODO Check if quantization works for unsupervised models
+    fout.write(struct.pack('@?', False))  # Save 'quant_', which is False for unsupervisded models
 
     # Save words and ngrams vectors
     _input_save(fout, model)
-    fout.write(struct.pack('@?', False))  # TODO Check if quantization works for unsupervised models
+    fout.write(struct.pack('@?', False))  # Save 'quot_', which is False for unsupervisded models
 
     # Save output layers of the model
     _output_save(fout, model)
 
 
-def save(fout, model, encoding='utf-8'):
+def save(model, fout, fb_fasttext_parameters, encoding):
     if isinstance(fout, str):
         with open(fout, "wb") as fout_stream:
-            _save(fout_stream, model, encoding)
+            _save(fout_stream, model, fb_fasttext_parameters, encoding)
     else:
-        _save(fout, model)
-
-
-# COMPARING FUNCTIONALITY
-
-
-def _sign_load(fin):
-    keys = ['fileformat_magic', 'version']
-    vals = _struct_unpack(fin, '@2i')
-    return dict(zip(keys, vals))
-
-
-def _load_key_fmt_list_to_dict(fin, key_fmt_list):
-    res = {}
-    for key, fmt in key_fmt_list:
-        res[key] = _struct_unpack(fin, fmt)[0]
-    return res
-
-
-def _args_load(fin):
-    return _load_key_fmt_list_to_dict(fin, _NEW_HEADER_FORMAT)
-
-
-def _dict_header_load(fin):
-    DICT_HEADER_FORMAT = [('size', 'i'),
-                          ('nwords', 'i'),
-                          ('nlabels', 'i'),
-                          ('ntokens', 'i'),
-                          ('pruneidx_size', '@q')]
-    return _load_key_fmt_list_to_dict(fin, DICT_HEADER_FORMAT)
-
-
-def _yield_differing_keys(d1, d2):
-    assert set(d1.keys()) == set(d2.keys())
-
-    for k in d1.keys():
-        v1, v2 = d1[k], d2[k]
-
-        if v1 != v2:
-            yield k, v1, v2
-
-    return None
-
-
-def _print_differences_between_dicts(d1, d2):
-    if d1 != d2:
-        for k, v1, v2 in _yield_differing_keys(d1, d2):
-            print('Key "%s" differs -> %s != %s' % (k, str(v1), str(v2)))
-
-
-def compare_fasttext_files(fname1, fname2):
-
-    with open(fname1, 'rb') as fin1, open(fname2, 'rb') as fin2:
-        sign1 = _sign_load(fin1)
-        sign2 = _sign_load(fin2)
-        _print_differences_between_dicts(sign1, sign2)
-
-        args1 = _args_load(fin1)
-        args2 = _args_load(fin2)
-        _print_differences_between_dicts(args1, args2)
-
-        dict_header1 = _dict_header_load(fin1)
-        dict_header2 = _dict_header_load(fin2)
-        _print_differences_between_dicts(dict_header1, dict_header2)
+        _save(fout, model, fb_fasttext_parameters, encoding)
 
 
 if six.PY2:
--- a/gensim/models/_fasttext_bin.py
+++ b/gensim/models/_fasttext_bin.py
@@ -501,7 +501,7 @@ def _dict_save(fout, model, encoding):
     fout.write(np.int64(-1))
 
     for word in model.wv.index2word:
-        word_count = model.wv.vocab["the"].count
+        word_count = model.wv.vocab[word].count
         fout.write(word.encode(encoding))
         fout.write(_END_OF_WORD_MARKER)
         fout.write(np.int64(word_count).tobytes())
--- a/gensim/models/_fasttext_bin.py
+++ b/gensim/models/_fasttext_bin.py
@@ -420,7 +420,6 @@ def _get_field_from_model(model, field):
         elif model.hs == 0 and model.negative > 0:
             res = 1
         elif model.hs == 0 and model.negative == 0:
-            # TODO Is this right ?!?
             res = 2
     elif field == 'maxn':
         res = model.wv.max_n
@@ -522,10 +521,6 @@ def _input_save(fout, model):
 
 
 def _output_save(fout, model):
-
-    # TODO Can model.hs and model.negative be both False?
-    # TODO Can model.hs and model.negative be both True?
-
     if model.hs:
         hidden_output = model.trainables.syn1
     if model.negative:
--- a/gensim/models/_fasttext_bin.py
+++ b/gensim/models/_fasttext_bin.py
@@ -469,25 +469,6 @@ def _dict_save(fout, model, encoding):
     # Reimplementation of the
     # [Dictionary::save](https://github.com/facebookresearch/fastText/blob/master/src/dictionary.cc)
 
-    # out.write((char*)&size_, sizeof(int32_t));
-    # out.write((char*)&nwords_, sizeof(int32_t));
-    # out.write((char*)&nlabels_, sizeof(int32_t));
-    # out.write((char*)&ntokens_, sizeof(int64_t));
-    # out.write((char*)&pruneidx_size_, sizeof(int64_t));
-    # for (int32_t i = 0; i < size_; i++) {
-    #   entry e = words_[i];
-    #   out.write(e.word.data(), e.word.size() * sizeof(char));
-    #   out.put(0);
-    #   out.write((char*)&(e.count), sizeof(int64_t));
-    #   out.write((char*)&(e.type), sizeof(entry_type));
-    # }
-    # for (const auto pair : pruneidx_) {
-    #   out.write((char*)&(pair.first), sizeof(int32_t));
-    #       out.write((char*)&(pair.second), sizeof(int32_t));
-    # }
-
-    # TODO Check what is the difference between `size` and `nwords`
-
     fout.write(np.int32(len(model.wv.vocab)).tobytes())
 
     fout.write(np.int32(len(model.wv.vocab)).tobytes())
@@ -536,21 +517,14 @@ def _output_save(fout, model):
 
 def _save(fout, model, fb_fasttext_parameters, encoding):
 
-    # Unfortunatelly there is no documentation of the FB binary format
+    # Unfortunatelly, there is no documentation of the FB binary format
     # This is just reimplementation of
+
     # [FastText::saveModel](https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc)
 
-    #   As of writing this (12.2019) the C++ code looks as follows
-    #
-    #   ```
-    #   signModel(ofs);
-    #   args_->save(ofs);
-    #   dict_->save(ofs);
-    #   ofs.write((char*)&(quant_), sizeof(bool));
-    #   input_->save(ofs);
-    #   ofs.write((char*)&(args_->qout), sizeof(bool));
-    #   output_->save(ofs);
-    #   ```
+    # Based on v0.9.1, more precisely commit da2745fcccb848c7a225a7d558218ee4c64d5333
+
+    # Code follows original C++ naming
 
     _sign_model(fout)
     _args_save(fout, model, fb_fasttext_parameters)
--- a/gensim/models/_fasttext_bin.py
+++ b/gensim/models/_fasttext_bin.py
@@ -43,7 +43,15 @@ _END_OF_WORD_MARKER = b'\x00'
 
 logger = logging.getLogger(__name__)
 
-_FASTTEXT_FILEFORMAT_MAGIC = 793712314
+# Constants for FastText vesrion and FastText file format magic (both int32)
+# https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc#L25
+
+_FASTTEXT_VERSION = np.int32(12)
+_FASTTEXT_FILEFORMAT_MAGIC = np.int32(793712314)
+
+
+# _NEW_HEADER_FORMAT is constructed on the basis of args::save method, see
+# https://github.com/facebookresearch/fastText/blob/master/src/args.cc
 
 _NEW_HEADER_FORMAT = [
     ('dim', 'i'),
@@ -51,13 +59,13 @@ _NEW_HEADER_FORMAT = [
     ('epoch', 'i'),
     ('min_count', 'i'),
     ('neg', 'i'),
-    ('_', 'i'),
+    ('word_ngrams', 'i'),   # Unused in loading
     ('loss', 'i'),
     ('model', 'i'),
     ('bucket', 'i'),
     ('minn', 'i'),
     ('maxn', 'i'),
-    ('_', 'i'),
+    ('lr_update_rate', 'i'),   # Unused in loading
     ('t', 'd'),
 ]
 
@@ -311,6 +319,7 @@ def load(fin, encoding='utf-8', full_model=True):
         fin = open(fin, 'rb')
 
     magic, version = _struct_unpack(fin, '@2i')
+    print(version)
     new_format = magic == _FASTTEXT_FILEFORMAT_MAGIC
 
     header_spec = _NEW_HEADER_FORMAT if new_format else _OLD_HEADER_FORMAT
@@ -364,6 +373,131 @@ def _backslashreplace_backport(ex):
     bstr, start, end = ex.object, ex.start, ex.end
     text = u''.join('\\x{:02x}'.format(ord(c)) for c in bstr[start:end])
     return text, end
+
+
+def _sign_model(fout):
+    # Reimplementation of the FastText::signModel function, see
+    # https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc
+    fout.write(_FASTTEXT_FILEFORMAT_MAGIC.tobytes())
+    fout.write(_FASTTEXT_VERSION.tobytes())
+
+
+def _conv_field_to_bytes(value, field_type):
+    if field_type == 'i':
+        return (np.int32(value).tobytes())
+    elif field_type == 'd':
+        return (np.float64(value).tobytes())
+    else:
+        raise NotImplementedError('Currently conversion to "%s" type is not implemmented.' % type)
+
+
+def _get_field(model, field, field_type):
+
+    if field == 'bucket':
+        res = model.bucket
+    elif field == 'dim':
+        res = model.vector_size
+    elif field == 'epoch':
+        res = model.epochs
+    elif field == 'loss':
+        # `loss` => ns: 1, hs: 2, softmax: 3, one-vs-all: 4
+        # ns = negative sampling loss (default)
+        # hs = hierarchical softmax loss
+        # softmax =  softmax loss
+        # one-vs-all = one vs all loss (supervised)
+        if model.hs == 1:
+            res = 2
+        elif model.hs == 0 and model.negative > 0:
+            res = 1
+        elif model.hs == 0 and model.negative == 0:
+            # TODO Is this right ?!?
+            res = 2
+    elif field == 'maxn':
+        res = model.wv.min_n
+    elif field == 'minn':
+        res = model.wv.min_n
+    elif field == 'min_count':
+        res = model.vocabulary.min_count
+    elif field == 'model':
+        # `model` => cbow:1, sg:2, sup:3
+        # cbow = continous bag of words (default)
+        # sg = skip-gram
+        # sup = supervised
+        res = 1 if model.sg == 1 else 2
+    elif field == 'neg':
+        res = model.negative
+    elif field == 't':
+        res = model.vocabulary.sample
+    elif field == 'word_ngrams':
+        # This is skipped in gensim loading settig, using the default from FB C++ code
+        res = 1
+    elif field == 'ws':
+        res = model.window
+    elif field == 'lr_update_rate':
+        # This is skipped in gensim loading settig, using the default from FB C++ code
+        res = 100
+    else:
+        raise NotImplementedError('Extraction of header field "%s" from Gensim FastText object not implemmented.' % field)
+
+    return _conv_field_to_bytes(res, field_type)
+
+
+def _args_save(fout, model):
+
+    # Reimplementation of the Args::save method, see
+    # https://github.com/facebookresearch/fastText/blob/master/src/args.cc
+
+    for field, field_type in _NEW_HEADER_FORMAT:
+        fout.write(_get_field(model, field, field_type))
+
+
+def _dict_save(fout, model):
+    pass
+
+
+def _save_vocab(fout, model):
+    pass
+
+
+def _save_vector_ngrams(fout, model):
+    pass
+
+
+def _save_hidden_outputs(fout, model):
+    pass
+
+
+def _save(fout, model):
+
+    # Unfortunatelly there is no documentation of the FB binary format
+    # This is just reimplementation of FastText::saveModel method
+    # See https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc
+
+    #   As of writing this (12.2019) the C++ code looks as follows
+    #
+    #   ```
+    #   signModel(ofs);
+    #   args_->save(ofs);
+    #   dict_->save(ofs);
+    #   ofs.write((char*)&(quant_), sizeof(bool));
+    #   input_->save(ofs);
+    #   ofs.write((char*)&(args_->qout), sizeof(bool));
+    #   output_->save(ofs);
+    #   ```
+
+    _sign_model(fout)
+    _args_save(fout, model)
+    _dict_save(fout, model)
+    _save_vector_ngrams(fout, model)
+    _save_hidden_outputs(fout, model)
+
+
+def save(fout, model):
+    if isinstance(fout, str):
+        with open(fout, "wb") as fout_stream:
+            _save(fout_stream, model)
+    else:
+        _save(fout, model)
 
 
 if six.PY2:
--- a/gensim/models/_fasttext_bin.py
+++ b/gensim/models/_fasttext_bin.py
@@ -447,11 +447,11 @@ def _get_field_from_model(model, field):
         # softmax =  softmax loss
         # one-vs-all = one vs all loss (supervised)
         if model.hs == 1:
-            res = 2
-        elif model.hs == 0 and model.negative > 0:
             res = 1
-        elif model.hs == 0 and model.negative == 0:
+        elif model.hs == 0:
             res = 2
+        elif model.hs == 0 and model.negative == 0:
+            res = 1
     elif field == 'maxn':
         res = model.wv.max_n
     elif field == 'minn':
--- a/gensim/models/_fasttext_bin.py
+++ b/gensim/models/_fasttext_bin.py
@@ -441,7 +441,7 @@ def _get_field_from_model(model, field):
     elif field == 'epoch':
         res = model.epochs
     elif field == 'loss':
-        # `loss` => ns: 1, hs: 2, softmax: 3, one-vs-all: 4
+        # `loss` => hs: 1, ns: 2, softmax: 3, ova-vs-all: 4
         # ns = negative sampling loss (default)
         # hs = hierarchical softmax loss
         # softmax =  softmax loss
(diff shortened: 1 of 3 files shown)


gensim/models/_fasttext_bin.py
```diff
@@ -168,6 +185,7 @@ def _load_vocab(fin, new_format, encoding='utf-8'):
         The loaded vocabulary.  Keys are words, values are counts.
         The vocabulary size.
         The number of words.
+        The numnber of tokens.
```
piskvorky:
```suggestion
        The number of tokens.
```

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,166 @@ def _backslashreplace_backport(ex):
     return text, end
 
 
+def _sign_model(fout):
```
piskvorky: Docstring(s) please. If I understood correctly, something simple as:

```python
def _sign_model(fout):
    """Write version info into the binary output stream `fout`.

    Reimplementation of `FastText::signModel <https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc>`_.
    """
```

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,166 @@ def _backslashreplace_backport(ex):
...
+    if field_type == 'i':
+        return (np.int32(field_value).tobytes())
+    elif field_type == 'd':
+        return (np.float64(field_value).tobytes())
+    else:
+        raise NotImplementedError('Currently conversion to "%s" type is not implemmented.' % field_type)
+
+
+def _get_field_from_model(model, field):
+
```
piskvorky: Dtto: docstrings instead of the blank lines please.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,166 @@ def _backslashreplace_backport(ex):
...
+
+def _save(fout, model, fb_fasttext_parameters, encoding):
+
+    # Unfortunatelly, there is no documentation of the FB binary format
+    # This is just reimplementation of
+
+    # [FastText::saveModel](https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc)
+
+    # Based on v0.9.1, more precisely commit da2745fcccb848c7a225a7d558218ee4c64d5333
```
piskvorky: +1 on these clear comments, that's great. Super useful for future maintenance.

gensim/models/fasttext.py
```diff
@@ -1216,15 +1216,15 @@ def _load_fasttext_format(model_file, encoding='utf-8', full_model=True):
         window=m.ws,
         iter=m.epoch,
         negative=m.neg,
-        hs=(m.loss == 1),
-        sg=(m.model == 2),
+        hs=int(m.loss == 2),
```
piskvorky: Why this change from `1` to `2`? Was there a bug?

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1305,299 @@ def _run(self, fin):
...
+            model = self._create_and_save_test_model(fpath, model_params)
+            wv = self._get_wordvectors_from_fb_fastttext(fasttext_cmd, fpath, model.wv.index2word)
+
+            for i, w in enumerate(model.wv.index2word):
+                diff = calc_max_diff(wv[i, :], model.wv[w])
+                # Because fasttext command line prints vectors with limited accuracy
+                self.assertLess(diff, 1.0e-4)
+
+    def test_load_fasttext_format_cbow(self):
+        model_params = {"size": 10, "min_count": 1, "hs": 1, "sg": 0,
```
piskvorky: Hanging indent please, here and everywhere.

We don't use vertical indent in gensim at all (hard to maintain).

gensim/models/_fasttext_bin.py
lopusz: :)

gensim/models/_fasttext_bin.py
lopusz: Corrected (also in the next functions).

gensim/test/test_fasttext.py
lopusz: Corrected, everywhere where I could spot it.

lopusz: PS. Have you considered using `black` formatter?

Perhaps it could alleviate such problems once and forever ;) (e.g. spacy I believe uses it).

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/fasttext.py
lopusz: No, my mistake I took the sequence of "enums" from help instead of cpp code.

Good catch. Corrected!

gensim/models/_fasttext_bin.py
lopusz: Corrected.

As far as motivation goes. I did not have very specific use case in mind.
Improving interoperability of gensim with original fastText seemed like a nice feature.

One of the loose ideas I had, was to try fine tuning gensim model using fastText. However FB, still does not support incremental learning, 

https://github.com/facebookresearch/fastText/pull/423

so I believe this is not yet possible.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,266 @@ def _backslashreplace_backport(ex):
...
+        res = 1
+    elif field == 'ws':
+        res = model.window
+    elif field == 'lr_update_rate':
+        # This is skipped in gensim loading settig, using the default from FB C++ code
+        res = 100
+    else:
+        msg = 'Extraction of header field "' + field + '" from Gensim FastText object not implemmented.'
+        raise NotImplementedError(msg)
+    return res
```
mpenkov: Is there a reason why we're not utilizing early returns here?

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1305,334 @@ def _run(self, fin):
...
+
+    def test_roundtrip_file_file_cbow(self):
+        if not os.environ.get("FT_HOME", None):
+            self.skipTest("FT_HOME env variable not set")
+        else:
+            model_params = {"size": 10, "sg": 0, "seed": 42}
+            self._check_roundtrip_file_file(model_params)
+
+
+class SaveFacebookFormatReadingTest(SaveFacebookFormatTest):
```
mpenkov: Way too much logic for a unit test.

gensim/test/test_fasttext.py
piskvorky: Yes, there was a discussion around using `black` a while back (1-2 years ago).  IIRC there was some issue, so we only implemented a PEP8 check.

But it's a good idea in general, we should revisit this.

lopusz:
Hi @mpenkov, thank you for the quick & thorough review! Reading your comments was very insightful. Contributing to gensim is a sheer pleasure.

You are right. The testing part is clumsy. I will give it more love & care along the lines that you suggested.

 I will tick the individual comments as I go and I will let you know when I am done.

Best regards,
Michał

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,266 @@ def _backslashreplace_backport(ex):
...
+        # cbow = continous bag of words (default)
+        # sg = skip-gram
+        # sup = supervised
+        res = 2 if model.sg == 1 else 1
+    elif field == 'neg':
+        res = model.negative
+    elif field == 't':
+        res = model.vocabulary.sample
+    elif field == 'word_ngrams':
+        # This is skipped in gensim loading settig, using the default from FB C++ code
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,266 @@ def _backslashreplace_backport(ex):
...
+        res = 1
+    elif field == 'ws':
+        res = model.window
+    elif field == 'lr_update_rate':
+        # This is skipped in gensim loading settig, using the default from FB C++ code
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,266 @@ def _backslashreplace_backport(ex):
...
+        saved model
+    fb_fasttext_parameters: dictionary
+        dictionary contain parameters containing `lr_update_rate`, `word_ngrams`
+        unused by gensim implementation, so they have to be provided externally
+    encoding: str
+        encoding used in the output file
+
+    Notes
+    -----
+    Unfortunatelly, there is no documentation of the Facebook's native fasttext `.bin` format
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,266 @@ def _backslashreplace_backport(ex):
...
+
+    def _save(model, fout, fb_fasttext_parameters, encoding):
+        _sign_model(fout)
+        _args_save(fout, model, fb_fasttext_parameters)
+        _dict_save(fout, model, encoding)
+        fout.write(struct.pack('@?', False))  # Save 'quant_', which is False for unsupervisded models
+
+        # Save words and ngrams vectors
+        _input_save(fout, model)
+        fout.write(struct.pack('@?', False))  # Save 'quot_', which is False for unsupervisded models
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,266 @@ def _backslashreplace_backport(ex):
...
+
+    This is just reimplementation of
+    [FastText::saveModel](https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc)
+
+    Based on v0.9.1, more precisely commit da2745fcccb848c7a225a7d558218ee4c64d5333
+
+    Code follows the original C++ code naming.
+    """
+
+    def _save(model, fout, fb_fasttext_parameters, encoding):
```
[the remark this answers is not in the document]
lopusz: Inner function moved to the top level.

gensim/models/_fasttext_bin.py
```diff
@@ -176,7 +194,8 @@ def _load_vocab(fin, new_format, encoding='utf-8'):
         raise NotImplementedError("Supervised fastText models are not supported")
     logger.info("loading %s words for fastText model from %s", vocab_size, fin.name)
 
-    _struct_unpack(fin, '@1q')  # number of tokens
+    ntokens, = _struct_unpack(fin, '@q')  # number of tokens
```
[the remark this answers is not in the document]
lopusz: Tricky one. `struct.unpack` always returns tuple, because it can produce multiple values from a given format. So in this case a call produces one-uple `(ntokens,)`, coma converts it to raw `ntokes`.

Changed to much more readable `struct.unpack(fin, '@q')[0]`

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,266 @@ def _backslashreplace_backport(ex):
...
+        hidden_output = model.trainables.syn1
+    if model.negative:
+        hidden_output = model.trainables.syn1neg
+
+    hidden_n, hidden_dim = hidden_output.shape
+    fout.write(struct.pack('@2q', hidden_n, hidden_dim))
+    fout.write(hidden_output.tobytes())
+
+
+def save(model, fout, fb_fasttext_parameters, encoding):
```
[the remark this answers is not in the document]
lopusz: My understanding is as follows  `_fasttext_bin.py` is internal module intended to use in `fasttext.py`.

For loading we have a bunch of  `_internal` functions (`_batched_generator`, `_load_matrix`, etc) and top-level `load` used in `fasttext.py`

I tried to copy the same for saving i.e. bunch `_internal` functions (`_sign_model`, `_args_save`, etc) with top-level `save` used in `fasttext.py`.

If this does not make sense, let's change it.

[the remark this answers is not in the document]
mpenkov: OK, if you intend for this function to be called from outside the module, then the name is fine. However, in that case, you must include a full docstring, including the parameters and their types. This is important information for people who will be using your module - they won't necessarily be familiar with the module's implementation details, so make the documentation as helpful as possible.

Also, if possible, assign default values to optional parameters, and save users the labor of working out what they mean. E.g. the encoding is usually utf8, and there's probably already a constant for that in that module somewhere.

I see that this function mostly wraps the internal `_save_to_stream` function, so perhaps move the docstring from there?

mpenkov:
@lopusz Please merge gensim/develop into your branch to fix the appveyor issues.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+
+    def _check_roundtrip_model_model(self, model_params):
+
+        with temporary_file("roundtrip_model_to_model.bin") as fpath:
+
+            model_trained = _create_and_save_test_model(fpath, model_params)
+
+            model_loaded = gensim.models.fasttext.load_facebook_model(fpath)
+
+            self.assertEqual(model_trained.vector_size, model_loaded.vector_size)
```
mpenkov: These assertions belong outside of the context manager because they do not depend on the temporary file. The models are already loaded into memory, aren't they?

Please check your context manager blocks and ensure their scope is as tight as possible.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+
+def calc_max_diff(v1, v2):
+    return np.max(np.abs(v1 - v2))
+
+
+class SaveFacebookFormatModelTest(unittest.TestCase):
+    """
+    This class containts tests that check the following scenario:
+
+    + create fastText model1
```
mpenkov: This comment describes _what_ the code does. It isn't really helpful, because that's already obvious from the code itself.

If it's not obvious, then the source code needs to be clearer. In this particular case, I think the code is OK, so you just need to get rid of the comment.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+    + save a model1 to Facebook binary file
+    + load the file as model2
+    + check if  model1 == model2
+    """
+
+    def _check_roundtrip_model_model(self, model_params):
```
mpenkov: Why does model appear in the method name twice? It doesn't really add any value to the name.

I think `_check_roundtrip` is sufficient here.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+    def test_round_trip_model_model_skipgram(self):
+        model_params = {
+            "size": 10,
+            "min_count": 1,
+            "hs": 1,
+            "sg": 1,
+            "negative": 5,
+            "seed": 42,
+            "workers": 1}
+        self._check_roundtrip_model_model(model_params)
```
mpenkov: Wouldn't it be simpler to unwrap model_params and do:

```suggestion
        self._check_roundtrip_model_model(sg=0)
```

and define the rest of the keyword arguments in the check-roundtrip function.

That will make it obvious what you're actually testing, because it appears the `sg` parameter is the only one that's changing here. It will also make the code a bit more readable as a side-effect.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+    def test_roundtrip_file_file_skipgram(self):
+        model_params = {
+            "size": 10,
+            "min_count": 1,
+            "hs": 1,
+            "sg": 1,
+            "negative": 0,
+            "seed": 42,
+            "workers": 1}
+        self._check_roundtrip_file_file(model_params)
```
mpenkov: Unwrap model params here, as above, and everywhere else, please.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+            "min_count": 1,
+            "hs": 1,
+            "sg": 0,
+            "negative": 0,
+            "seed": 42,
+            "workers": 1}
+        self._check_roundtrip_file_file(model_params)
+
+
+def _save_test_model(out_base_fname, model_params, fasttext_cmd):
```
mpenkov: I think the name is misleading here, because you're not just saving, you're _training_ a model here.

But really, I think this function may be unnecessary. Just train the two models locally, and add the binaries to the git tree. How large could they be?

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+    + create binary fastText file model1.bin using facebook_binary (FT)
+    + load file model1.bin to variable `model`
+    + save `model` to model2.bin using gensim
+    + check if files model1.bin and model2.bin are byte-identical
+
+    Requires env. variable FT_HOME to point to location of Facebook fastText binary
+    """
+
+    def _check_roundtrip_file_file(self, model_params):
+        ft_home = os.environ.get("FT_HOME", None)
```
mpenkov: You're doing this check in three separate places (here, and in each of the two tests). Move it to module scope:

```python
FT_HOME = os.environ.get("FT_HOME")
FT_CMD = ...
```

and use the result in its stead.

You can also skip the entire suite if FT_HOME is missing, instead of skipping individual tests.

Also, you don't need to pass None to `get()`, as that is the default behavior.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+            self.assertEqual(model_trained.vocabulary.sample, model_loaded.vocabulary.sample)
+            self.assertEqual(set(model_trained.wv.index2word), set(model_loaded.wv.index2word))
+
+            for w in model_trained.wv.index2word:
+                v_orig = model_trained.wv[w]
+                v_loaded = model_loaded.wv[w]
+
+                try:
+                    self.assertLess(calc_max_diff(v_orig, v_loaded), MAX_WORDVEC_COMPONENT_DIFFERENCE)
+                except AssertionError as e:
```
mpenkov: Why is this necessary? Is the default exception message insufficient?

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+                    s = e.args[0]
+                    s += " (max difference for components in wordvector of \"%s\"" % w
+                    s += " larger than the thershold %f" % MAX_WORDVEC_COMPONENT_DIFFERENCE
+                    e.args = (s,)
+                    raise e
+
+    def test_round_trip_model_model_skipgram(self):
```
mpenkov: Please use concise test names, here and everywhere. You're not risking any collisions here. The roundtrip isi implied from the test suite itself, and model/model is redundant.

```suggestion
    def test_skipgram(self):
```

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
         self.assertTrue(np.allclose(_ARRAY, array))
 
 
+def _create_and_save_test_model(fname, model_params):
```
mpenkov: I think it's relevant to put `fb` in the function name to make it obvious which format you're saving in. `test` isn't particuarly relevant here.

```suggestion
def _create_and_save_fb_model(fname, model_params):
```

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+        self._check_roundtrip_model_model(model_params)
+
+
+def _read_binary_file(fname):
+    with open(fname, "rb") as f:
+        data = f.read()
+    return data
+
+
+class SaveFacebookFormatFileGensimTest(unittest.TestCase):
```
mpenkov: There are really four things you need to know about this class:

0. It's a test
1. It's for the save function
2. It compares results byte to byte
4. It tests pure gensim (unlike it's FB cousin)

```suggestion
class SaveGensimByteIdentityTest(unittest.TestCase):
```

Everything else, e.g.

- FacebookFormat
- File

isn't really relevant enough to warrant putting it in the very name.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+    model_type = "cbow" if model_params["sg"] == 0 else "skipgram"
+    size = str(model_params["size"])
+    seed = str(model_params["seed"])
+
+    cmd = fasttext_cmd + " " + model_type + " -input " + inp_fname + \
+            " -output " + out_base_fname + " -dim " + size + " -seed " + seed
+    subprocess.run(cmd, shell=True)
+
+
+class SaveFacebookFormatFileFastTextTest(unittest.TestCase):
```
mpenkov:
```suggestion
class SaveFacebookByteIdentityTest(unittest.TestCase):
```

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+    """
+
+    def _check_roundtrip_file_file(self, model_params):
+        ft_home = os.environ.get("FT_HOME", None)
+        fasttext_cmd = os.path.join(ft_home, "fasttext")
+
+        # fasttext tool creates both *vec and *bin files, so we have to remove both, even thought *vec is unused
+
+        with temporary_file("roundtrip_file_to_file1.bin") as fpath1bin, \
+            temporary_file("roundtrip_file_to_file2.bin") as fpath2bin, \
```
mpenkov: I think if you indent by one more level, flake8 won't complain anymore.

But, even better, if you use shorter names, you won't need to span multiple lines with the context manager.

```python
with temp_file("model1") as m1, temp_file("model2"), temp_file("vector1") as v1:
```

After all, who are those long filenames for? Those files get nuked when the test exits, regardless of success or failure, so putting too much effort into naming them is not productive.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+def _parse_wordvectors(text):
+    def _conv_line_to_array(line):
+        return np.array([float(s) for s in line.split()[1:]], dtype=np.float32)
+
+    return np.array([_conv_line_to_array(l) for l in text.splitlines()], dtype=np.float32)
+
+
+def _read_wordvectors_using_fasttext(fasttext_cmd, fasttext_fname, words):
+    cmd = fasttext_cmd + " print-word-vectors " + fasttext_fname
+    process = subprocess.Popen(
```
mpenkov: This can fit on one line. You don't need to specify `err`, because you're not using it.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+
+def _parse_wordvectors(text):
+    def _conv_line_to_array(line):
+        return np.array([float(s) for s in line.split()[1:]], dtype=np.float32)
+
+    return np.array([_conv_line_to_array(l) for l in text.splitlines()], dtype=np.float32)
+
+
+def _read_wordvectors_using_fasttext(fasttext_cmd, fasttext_fname, words):
+    cmd = fasttext_cmd + " print-word-vectors " + fasttext_fname
```
mpenkov: This is the [recommended way](https://docs.python.org/3.5/library/subprocess.html#popen-constructor) to pass the command:

```suggestion
    cmd = [fasttext_cmd, "print-word-vectors", fasttext_fname]
```

If you do it this way, then you don't need `shell=True` (unless you have another reason to use it, in that case, please let me know).

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1306,253 @@ def _run(self, fin):
...
+    process = subprocess.Popen(
+        cmd, stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        shell=True)
+    words_str = '\n'.join(words)
+    out, err = process.communicate(input=words_str.encode("utf-8"))
+    return _parse_wordvectors(out.decode("utf-8"))
```
mpenkov: You don't really need _parse_wordvectors to be a separate function here. You're only using it once.

```python
return np.array([_line_to_array(l) for l in out.decode('utf-8').splitlines()], dtype=np.float32)
```

gensim/test/test_fasttext.py
mpenkov: Personally, I'm against Black. My reasons:

1. I consider code formatting to be a form of expression (like art). I'd prefer to not have a computer program stomp all over it. It feels too authoritarian. This is the main reason.
2. I don't Black's style. Yes, our code base will be consistently formatted, but it will look like crap.

lopusz:
> Looks like you've been busy :)
> The tests are looking much better +1 . I did another round of reviews focusing on the tests only.
> Please have a look and let me know. 

@mpenkov Yeah, thank you for the appreciation :)  Indeed, I gave tests a lot of attention lately.
I was just about to write you, to have a look now!

I will have a go, over your remarks. I think now it will be quick clean-up, and we can merge that ;)

gensim/test/test_fasttext.py
lopusz: Corrected.

gensim/test/test_fasttext.py
lopusz: Corrected.

gensim/test/test_fasttext.py
lopusz: Corrected.

gensim/test/test_fasttext.py
lopusz: Corrected.

gensim/test/test_fasttext.py
lopusz: Corrected everywhere.

gensim/test/test_fasttext.py
lopusz: This is a nice one. A lot of simplification.

Corrected.

gensim/test/test_fasttext.py
lopusz: I wanted to have an information for which word the vectors actually differ (once this is the case).

But, yeah, let us go for the dead simple version.

Corrected.

gensim/test/test_fasttext.py
lopusz: Good catch. Corrected everywhere.

gensim/test/test_fasttext.py
lopusz: Corrected.

gensim/test/test_fasttext.py
lopusz: Corrected.

gensim/test/test_fasttext.py
lopusz: Corrected.

gensim/test/test_fasttext.py
lopusz: Corrected.

gensim/test/test_fasttext.py
lopusz: Good point. Corrected.

gensim/test/test_fasttext.py
lopusz: Definitely simpler. Corrected,

gensim/test/test_fasttext.py
lopusz: Creating the model explicitly can be useful. 

Imagine Facebook changes the format.  Once we provide the latest FT_CMD this will cause test failure. Once we correct our implementation tests will pass again.  

If we include binary files we live in ignorance, and we have to regenerate the model file for the test to work again.

lopusz:
Hi @mpenkov, I think I addressed all your comments. 

Please, have a look if that is mergeable.

Best regards,
Michał

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
+    fout.write(model.wv.vectors_ngrams.tobytes())
+
+
+def _output_save(fout, model):
+    """
+    Saves output layer of `model` to the binary stream `fout` containing a model in
+    the Facebook's native fastText `.bin` format.
+
+    Corresponding C++ fastText code:
+    DenseMatrix::save[DenseMatrix::save](https://github.com/facebookresearch/fastText/blob/master/src/densematrix.cc)
```
mpenkov: Fix formatting.

```suggestion
    [DenseMatrix::save](https://github.com/facebookresearch/fastText/blob/master/src/densematrix.cc)
```

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
+    # We are in unsupervised case, therefore pruned_idx is empty, so we do not need to write anything else
+
+
+def _input_save(fout, model):
+    """
+    Saves word and ngram vectors from `model` to the binary stream `fout` containing a model in
+    the Facebook's native fastText `.bin` format.
+
+    Corresponding C++ fastText code:
+    DenseMatrix::save[DenseMatrix::save](https://github.com/facebookresearch/fastText/blob/master/src/densematrix.cc)
```
mpenkov: Fix the link:

```suggestion
    [DenseMatrix::save](https://github.com/facebookresearch/fastText/blob/master/src/densematrix.cc)
```

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
+
+    Parameters
+    ----------
+    fout: writeable binary stream
+        stream to which model is saved
```
mpenkov:
```suggestion
        stream to which the vectors are saved
```

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
+    model: gensim.models.fasttext.FastText
+        saved model
```
mpenkov:
```suggestion
        the model that contains the vectors to save
```

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
```
mpenkov:
```suggestion
        the model that contains the output layer to save
```

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
+    Saves output layer of `model` to the binary stream `fout` containing a model in
+    the Facebook's native fastText `.bin` format.
+
+    Corresponding C++ fastText code:
+    DenseMatrix::save[DenseMatrix::save](https://github.com/facebookresearch/fastText/blob/master/src/densematrix.cc)
+
+    Parameters
+    ----------
+    fout: writeable binary stream
+        stream to which model is saved
```
mpenkov:
```suggestion
        stream to which the output layer is saved
```

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
+
+
+def _save_to_stream(model, fout, fb_fasttext_parameters, encoding):
+    """
+    Saves word embeddings to binary stream `fout` using the Facebook's native fasttext `.bin` format.
+
+    Parameters
+    ----------
+    fout: file name or writeable binary stream
+        stream to which model is saved
```
mpenkov:
```suggestion
        stream to which the word embeddings are saved
```

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
+    model: gensim.models.fasttext.FastText
+        saved model
```
mpenkov:
```suggestion
        the model that contains the word embeddings to save
```

gensim/models/fasttext.py
```diff
@@ -1282,3 +1287,40 @@ def _check_model(m):
...
+
+def save_facebook_model(model, path, encoding="utf-8", lr_update_rate=100, word_ngrams=1):
+    """Saves word embeddings to the Facebook's native fasttext `.bin` format.
+
+    Notes
+    ------
+    Facebook provides both `.vec` and `.bin` files with their modules.
+    The former contains human-readable vectors.
+    The latter contains machine-readable vectors along with other model parameters.
+    This function saves only the .bin file**.
```
mpenkov: Fix RST formatting.

```suggestion
    **This function saves only the .bin file**.
```

gensim/models/fasttext.py
```diff
@@ -1282,3 +1287,40 @@ def _check_model(m):
...
+    ----------
+    model : gensim.models.fasttext.FastText
+        FastText model to be saved.
+    path : str
+        Output path and filename (including `.bin` extension)
+    encoding : str, optional
+        Specifies the file encoding. Defaults to utf-8.
+
+    lr_update_rate : int
+        This parameter is used by Facebook faasttext tool, unused by Gensim.
```
mpenkov: Typo.

```suggestion
        This parameter is used by Facebook fasttext tool, unused by Gensim.
```

Please check elsewhere, too.

lopusz:
Hi Michael, thanks for another batch of remarks. I was mostly off line for a week, hence the delay.

I will give it final touches today/tomorrow,  so we can close that :) Stay tuned.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1302,194 @@ def _run(self, fin):
...
+            "sg": sg,
+            "size": 10,
+            "min_count": 1,
+            "hs": 1,
+            "negative": 5,
+            "seed": 42,
+            "workers": 1}
+
+        with temporary_file("roundtrip_model_to_model.bin") as fpath:
+
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1302,194 @@ def _run(self, fin):
...
+            "negative": 0,
+            "seed": 42,
+            "workers": 1}
+
+        with temporary_file("roundtrip_file_to_file1.bin") as fpath1, \
+            temporary_file("roundtrip_file_to_file2.bin") as fpath2:
+            _create_and_save_fb_model(fpath1, model_params)
+            model = gensim.models.fasttext.load_facebook_model(fpath1)
+            gensim.models.fasttext.save_facebook_model(model, fpath2)
+            self.assertEqual(_read_binary_file(fpath1), _read_binary_file(fpath2))
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1302,194 @@ def _run(self, fin):
...
+    + save `model` to model2.bin using gensim
+    + check if files model1.bin and model2.bin are byte-identical
+    """
+
+    def _check_roundtrip_file_file(self, sg):
+        model_params = {"size": 10, "sg": sg, "seed": 42}
+
+        # fasttext tool creates both *vec and *bin files, so we have to remove both, even thought *vec is unused
+
+        with temporary_file("m1.bin") as m1, temporary_file("m2.bin") as m2, temporary_file("m1.vec") as _:
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1302,194 @@ def _run(self, fin):
...
+
+            m1_basename = m1[:-4]
+            _save_test_model(m1_basename, model_params)
+            model = gensim.models.fasttext.load_facebook_model(m1)
+            gensim.models.fasttext.save_facebook_model(model, m2)
+            self.assertEqual(_read_binary_file(m1), _read_binary_file(m2))
```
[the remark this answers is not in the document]
lopusz: Corrected

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1302,194 @@ def _run(self, fin):
...
+
+    def test_skipgram(self):
+        self._check_roundtrip_file_file(sg=1)
+
+    def test_cbow(self):
+        self._check_roundtrip_file_file(sg=0)
+
+
+def _read_wordvectors_using_fasttext(fasttext_fname, words):
+    def _conv_line_to_array(line):
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
+        if field in fb_fasttext_parameters:
+            field_value = fb_fasttext_parameters[field]
+        else:
+            field_value = _get_field_from_model(model, field)
+        fout.write(_conv_field_to_bytes(field_value, field_type))
+
+
+def _dict_save(fout, model, encoding):
+    """
+    Saves vocabulary from `model` to the to the binary stream `fout` containing a model in the Facebook's
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,282 @@ def _backslashreplace_backport(ex):
...
+
+    Name mimics the original C++ implementation
+    [Dictionary::save](https://github.com/facebookresearch/fastText/blob/master/src/dictionary.cc)
+
+    Parameters
+    ----------
+    fout: writeable binary stream
+        stream to which model is saved
+    model: gensim.models.fasttext.FastText
+        saved model
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/_fasttext_bin.py
lopusz: Corrected.

gensim/models/_fasttext_bin.py
```diff
@@ -366,5 +386,266 @@ def _backslashreplace_backport(ex):
...
+    encoding: str
+        string encoding used in the output
+    """
+    fout.write(np.int32(len(model.wv.vocab)).tobytes())
+
+    fout.write(np.int32(len(model.wv.vocab)).tobytes())
```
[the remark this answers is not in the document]
lopusz: Comment added.

gensim/models/fasttext.py
lopusz: Corrected.

gensim/models/fasttext.py
lopusz: Corrected here and in one more place.

gensim/test/test_fasttext.py
```diff
@@ -811,13 +815,10 @@ def compare_with_wrapper(self, model_gensim, model_wrapper):
         # this limit can be increased when using Cython code
         self.assertGreaterEqual(overlap_count, 2)
 
+    @unittest.skipIf(not FT_HOME, "FT_HOME env variable not set, skipping test")
```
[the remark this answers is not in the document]
lopusz: Please note, that this is a correction to a TestFastTextModel test case, which checks the  FastText model implementation. I changed the tests for the model part to also make use of FT_HOME global, as test of FB saving do.

To my understanding moving the decorator to a class level will not do what we want. The result will be that you skip **all the tests** TestFastTextModel if the FT_CMD is not installed,  
whereas FT_CMD is required only for two tests.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1302,194 @@ def _run(self, fin):
...
+
+
+def _save_test_model(out_base_fname, model_params):
+    inp_fname = datapath('lee_background.cor')
+
+    model_type = "cbow" if model_params["sg"] == 0 else "skipgram"
+    size = str(model_params["size"])
+    seed = str(model_params["seed"])
+
+    cmd = FT_CMD + " " + model_type + " -input " + inp_fname + \
```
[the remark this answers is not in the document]
lopusz: Corrected.

gensim/test/test_fasttext.py
```diff
@@ -1300,6 +1302,194 @@ def _run(self, fin):
...
+        " -output " + out_base_fname + " -dim " + size + " -seed " + seed
+    subprocess.run(cmd, shell=True)
```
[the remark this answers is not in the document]
lopusz: Corrected.

lopusz:
@mpenkov Huh, I got slowed down on this last stretch, but I think I am done with the last batch of corrections. Please, have a look!

mpenkov:
Awesome work. Thank you for your effort and your patience!

lopusz:
@mpenkov @piskvorky thank you for your insightful comments and lightning fast feedback. Contributing to gensim is a pleasure :smile:

Issue #2713, opened 2019-12-27
Proposal: drop support for Python 3.5; add Python 3.8 to builds


gojomo:
I recommend the next major gensim release (4.x) also drop official Python 3.5 testing/support. Rationale:

* it's close to full end-of-life (under 10 months to 2020-09-13) – currently receiving only security fixes
* major Python-using cloud services (including Heroku, AWS Lambda, Google Cloud/Colab, etc) already only support Python 3.6+
* some useful features (like the variable annotations needed for dataclasses) only arrive in 3.6
* for most users, moving forward to 3.6.x+ isn't hard; those absolutely stuck on 3.5 can use an earlier gensim version, or try their luck in an unsupported configuration (some modules might still work in Python 3.5)

Meanwhile, Python 3.8 has been out a few months (with support at Heroku & AWS Lambda) and deserves added build/test support. So we could replace the 3.5 builds/test in our CI setup with a new 3.8 target.

piskvorky:
+1: I've always been in favour of 3.6+, but was overruled previously in https://github.com/RaRe-Technologies/gensim/issues/2548.

The numbers also support it: 3.6 is the second most popular Python (after 2.7).

gojomo:
PR #2714 drops 3.5 and adds 3.8, and strips out or updates a bunch of pinning or pre-installation-of-wheels-from-nonstandard-sources – that appeared possibly obsolete. 

It works pretty well at our CI/build services – except the 3.8-on-Appveyor (Windows) build. I've poked at it a little & think the problem is mainly outside our control and may self-resolve as Appveyor or other projects improve their 3.8 coverage. 

May need reconciliation with #2706's appveyor changes.

menshikh-iv:
@piskvorky yes, you are right (according to https://www.jetbrains.com/research/python-developers-survey-2018/)

PR #2714, opened 2019-12-27
Update/streamline CI builds (drop py3.5, add py3.8, remove/update old pins/workarounds)

gojomo:
PR to test changes to build recipes - adjusting Python & other-library versions. See also #2713.


gojomo:
replaced with pr #2715 from personal branch (to eliminate redundant `/pr` and `/branch`/`/push` builds).

PR #2720, opened 2020-01-03
Change similarity strategy when finding n best

### LINKED ISSUE #2584: gensim.similarities.Similarity merges results from shards incorrectly (LSI model)
If "num_best" is used, `gensim.similarities.Similarity` runs the query against each of the shards (MatrixSimilarity objects) and then merges the results.

MatrixSimilarity uses `matutils.full2sparse_clipped()` to pick "num_best" results which sorts by the absolute value.

gensim.similarities.Similarity on the other hand, just uses `heapq.nlargest` (in `__getitem__`) to merge the results from each of the shards. So negative sims are either pushed down the list or cut off completely.

svinkapeppa:
Fixes #2584

### DIFF
--- a/gensim/similarities/docsim.py
+++ b/gensim/similarities/docsim.py
@@ -233,6 +233,30 @@ def query_shard(args):
     return result
 
 
+def _nlargest(n, iterable):
+    """Helper for extracting n documents with maximum similarity.
+
+    Parameters
+    ----------
+    n : int
+        Number of elements to be extracted
+    iterable : iterable of list of (int, float)
+        Iterable containing documents with computed similarities
+
+    Returns
+    -------
+    :class:`list`
+        List with the n largest elements from the dataset defined by iterable.
+
+    Notes
+    -----
+    Elements are compared by the absolute value of similarity, because negative value of similarity
+    does not mean some form of dissimilarity.
+
+    """
+    return heapq.nlargest(n, itertools.chain(*iterable), key=lambda item: abs(item[1]))
+
+
 class Similarity(interfaces.SimilarityABC):
     """Compute cosine similarity of a dynamic query against a corpus of documents ('the index').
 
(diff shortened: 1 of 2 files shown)


gensim/similarities/docsim.py
```diff
@@ -539,7 +539,7 @@ def convert(shard_no, doc):
             if not is_corpus:
                 # user asked for num_best most similar and query is a single doc
                 results = (convert(shard_no, result) for shard_no, result in enumerate(shard_results))
-                result = heapq.nlargest(self.num_best, itertools.chain(*results), key=lambda item: item[1])
+                result = heapq.nlargest(self.num_best, itertools.chain(*results), key=lambda item: abs(item[1]))
```
mpenkov: This is getting a little bit too hairy. We're multiple indentation levels in, and we're duplicating a bit of code. Let's split out a function:

```python
def _nlargest(n, results):
    return heapq.nlargest(n, itertools.chain(*results), key=lambda item: abs(item[1]))
```

and then use it inside as:

```suggestion
                result = _nlargest(self.num_best, results)
```

piskvorky: Can `lambda` accept a tuple argument? If so, `key=lambda (feature, weight): abs(weight)` more readable.

svinkapeppa: @piskvorky Unfortunately it was removed from python3 (you can check it [here](https://www.python.org/dev/peps/pep-3113/))

@mpenkov I guess `_nlargest()` can be tested, whereas creating test for the whole `Similarity` object is complicated (I guess it involves changing the corpus, that is used for testing and fixing all other test OR maybe adding explicit test and corpus for `Similarity` that checks, that right objects are extracted with `_nlargest`). Which is more preferable?

mpenkov: Write tests for nlargest for now.

svinkapeppa:
I'm not sure if it's an actual issue but sometimes docstrings are not 100% correct: for example it is said that `Similarity.__getitem__` returns numpy or scipy arrays, but it can actually return regular python lists (e.g., when `num_best` was specified). What should contributors do in these cases?

mpenkov:
@svinkapeppa Thank you for your contribution!

I've merged it.

PR #2722, opened 2020-01-06
Refactor bm25 to include model parametrization (cont.)

### LINKED ISSUE #2597: It is not possible to set values for BM25 model hyperparameters (K1, B, and EPSILON)

Witiko:
This pull request adds unit tests to #2606 and merges the latest changes in the develop branch, which could not be automerged and interfered with the changes implemented in #2606.

The brunt of the work was done by @marcelo-dalmeida.

This pull request closes #2597 and #2606.


gensim/summarization/bm25.py
```diff
@@ -151,8 +173,8 @@ def get_score(self, document, index):
         for word in document:
             if word not in doc_freqs:
                 continue
-            score += (self.idf[word] * doc_freqs[word] * (PARAM_K1 + 1)
-                      / (doc_freqs[word] + PARAM_K1 * (1 - PARAM_B + PARAM_B * self.doc_len[index] / self.avgdl)))
+            score += (self.idf[word] * doc_freqs[word] * (self.k1 + 1)
```
mpenkov: Now that we have some tests, I think this PR is a good opportunity to make this more readable.

How about this inside the loop:

```python
# Maybe pick a better name for this? Looks like we can totally precompute it.
numerator_pad = self.k1 + 1
denominator_pad = self.k1 * (1 - self.b + self.b * self.doc_len[index] / self.avgd1)
for word in document:
    idf = self.idf[word]
    try:
        df = self.doc_freqs[index][word]
    except KeyError:
        # ask for forgiveness instead of permission
        continue
    score +=  (idf * df * numerator_pad) / (df + denominator_pad)
return score
```

WDYT?

gensim/test/test_BM25.py
```diff
@@ -62,6 +62,76 @@ def test_multiprocessing(self):
...
+        """ Changing the k1 parameter should give consistent results """
+        corpus = common_texts
+        index = 0
+        doc = corpus[index]
+        first_k1 = 1.0
+        second_k1 = 2.0
+
+        first_bm25 = BM25(corpus, k1=first_k1)
+        second_bm25 = BM25(corpus, k1=second_k1)
+        self.assertTrue(first_bm25.get_score(doc, index) < second_bm25.get_score(doc, index))
```
mpenkov: Simplify the assertion:

```suggestion
        first = first_bm25.get_score(doc, index)
        second = second_bm25.get_score(doc, index)
        self.assertLessThan(first, second)
```

This will give a more helpful message upon failure.

gensim/summarization/bm25.py
piskvorky: `word not in doc_freqs:` much more readable than catching exceptions, I prefer the original.

In general, is the division guaranteed not to raise division by zero?

mpenkov: OK, I updated my suggestion.

Yeah, I think we're OK with the div-by-zero because of the denominator_pad. @Witiko Will that always be non-zero?

piskvorky: "Not equal to `-df`" is what we need.

gensim/test/test_BM25.py
Witiko: I simplified the assertions in 4a61276.

gensim/summarization/bm25.py
Witiko: I am skimming the code, but, as long as `b` is within [0; 1] and `k1` is non-negative, `denominator_pad` is non-negative and `df` is positive, so no division by zero will occur.

If `corpus` is empty, then `self.avgdl` is zero, which would seem to lead to a division by zero, but the user will receive an `IndexError` before that.

Should I break down the BM25 formula into subcomponents like @mpenkov suggests?

Witiko: I refactored the inner loop in 13845d12.

Witiko:
@mpenkov Thanks for the review. Please, let me know if you have any other suggestions before merge.

PR #2729, opened 2020-01-16
Update get_keras_embedding for tf2

themrzmaster:
importing direct from keras raises error 
> 
AttributeError: module 'tensorflow' has no attribute 'get_default_graph'

as recommended by the tf community, we should import keras as tensorflow.keras

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -1522,7 +1522,7 @@ class Word2VecKeyedVectors(WordEmbeddingsKeyedVectors):
 
         """
         try:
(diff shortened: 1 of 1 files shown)


mpenkov:
@themrzmaster Thank you for your contribution.

Please:

1. Link to the recommendation from the TF community. It is unclear why the proposed method is an improvement.
2. Make the CI tests under Travis pass. They fail with an ImportError.

piskvorky:
Fixed in #2937, by moving the (trivial) `get_keras_embedding` to Wiki.

PR #2715, opened 2019-12-28
[RFM] Update/streamline CI builds (drop py3.5, add py3.8, remove/update old pins/workarounds)

gojomo:
PR to test changes to build recipes - adjusting Python & other-library versions. See discussion #2713.

(Replaces #2714)

PR #2730, opened 2020-01-22
Use time.time instead of time.clock in gensim/models/hdpmodel.py

tarohi24:
`time.clock` has been deprecated since in Python 3.3, and it is no longer available in Python 3.8 (https://github.com/python/cpython/pull/13270).

### DIFF
--- a/gensim/models/hdpmodel.py
+++ b/gensim/models/hdpmodel.py
@@ -464,7 +464,10 @@ class HdpModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
         """
         save_freq = max(1, int(10000 / self.chunksize))  # save every 10k docs, roughly
         chunks_processed = 0
-        start_time = time.clock()
+        try:
+            start_time = time.process_time()
+        except AttributeError:
+            start_time = time.clock()
 
         while True:
             for chunk in utils.grouper(corpus, self.chunksize):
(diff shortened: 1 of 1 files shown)


gensim/models/hdpmodel.py
```diff
@@ -464,7 +464,10 @@ def update(self, corpus):
         """
         save_freq = max(1, int(10000 / self.chunksize))  # save every 10k docs, roughly
         chunks_processed = 0
-        start_time = time.clock()
+        try:
+            start_time = time.process_time()
+        except AttributeError:
+            start_time = time.clock()
```
piskvorky: Both seem wrong to me. Why should user case about the process time? Who needs such runtime constraints?

What people care about is the wallclock time, so perhaps `time.time()`.

tarohi24: Thank you for the review!
As you pointed out, I should have used `time.time()`.
I'll fix it.

tarohi24:
I replaced `time.process_time()` with `time.time()`.

gensim/models/hdpmodel.py
```diff
@@ -464,7 +464,10 @@ def update(self, corpus):
         """
         save_freq = max(1, int(10000 / self.chunksize))  # save every 10k docs, roughly
         chunks_processed = 0
-        start_time = time.clock()
+        try:
+            start_time = time.time()
+        except AttributeError:
+            start_time = time.clock()
```
piskvorky: What's with the `clock()`? Just use `time.time()`, it's there always. No need for such `try: except`.

mpenkov: Whoops. Looks like my merge was premature. @tarohi24 Are you able to make another PR to patch this up? If not, then I will take care of it.

tarohi24: OK! I'llbfix it.

gensim/models/hdpmodel.py
```diff
@@ -508,12 +511,16 @@ def update_finished(self, start_time, chunks_processed, docs_processed):
...
         return (
             # chunk limit reached
             (self.max_chunks and chunks_processed == self.max_chunks)
 
             # time limit reached
-            or (self.max_time and time.clock() - start_time > self.max_time)
+            or (self.max_time and start_time - start_time > self.max_time)
```
gojomo: `start_time - start_time` makes no sense (and the above `AttributeError`-complicated fallback to `.clock()` seems fishy too, especially since the deprecation of `.clock()` is exactly the reason for this PR). But, I have a different fix incoming in #2715.

PR #2731, opened 2020-01-22
Speed up Soft Cosine Measure

eliadl:
Use selection matrices to speed up matrix slicing.

**Motivation:**
Sometimes `softcossim` needs to be calculated O(N^2) times, e.g. for a distance matrix.

**Results:**
At least 5x speedup of `softcossim` was observed on my machine (110 µs vs 550 µs).

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -880,7 +880,10 @@ def softcossim(vec1, vec2, similarity_matrix):
     dtype = similarity_matrix.dtype
     vec1 = np.fromiter((vec1[i] if i in vec1 else 0 for i in word_indices), dtype=dtype, count=len(word_indices))
     vec2 = np.fromiter((vec2[i] if i in vec2 else 0 for i in word_indices), dtype=dtype, count=len(word_indices))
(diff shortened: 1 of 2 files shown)


piskvorky:
@Witiko WDYT?

Witiko:
Looks good at first glance, but I won't be able to properly test the speed and correctness until next week. @eliadl, can you please explain the issue with the original code and what the source of the speed-up is?

eliadl:
@Witiko I went ahead and tested with larger matrices, and the performance **decreased**.
Before, I thought that numpy's lower-level implementation would compensate for the larger complexity of matrix operations.
But that doesn't seem to be the case.

To be safe I'll close this PR, but leave the branch for you to inspect if you'd like.

Sorry and thanks for your time and review.

eliadl:
@Witiko On second thought, perhaps it's worth considering my version only up to a certain size of the similarity matrix, as it does work faster for smaller matrices.

So the code would have an `if size > threshold else`.

Finding that `threshold` would require to profile the performance of different sizes.
Though I'm not sure different machines, OS's, python versions, etc. would have the same threshold.

PR #2738, opened 2020-01-28
Fix out of range issue in gensim.summarization.keywords

### LINKED ISSUE #2598: keywords.py gives `IndexError: list index out of range` when `words` parameter is provided.

carterols:
Fix #2598 

The issue arises when the parameter `words`, which is defined as _The maximum amount of words to return_, is greater than the number of `lemmas` or keywords found. This caused an index out of range error from line 306 because it wasn't handling the case where `words > len(lemmas)`.

**Before the fix:** 
`length = len(lemmas) * ratio if words is None else words`
`return [(scores[lemmas[i]], lemmas[i],) for i in range(int(length))]`
As one can see, if `words > len(lemmas)`, the return statement will throw an index out of range error because the for loop will continue past the length of the lemmas array.

**Example of reproducing issue:**
`lemmas = ["test", "issue", "reproduction"]`
`words = 5`

Before Fix:
Index out of range in line 306 because length = 5 and len(lemmas) = 3.

After Fix:
No issue because length = len(lemmas) = 3.

### DIFF
--- a/gensim/summarization/keywords.py
+++ b/gensim/summarization/keywords.py
@@ -302,7 +302,7 @@ def _extract_tokens(lemmas, scores, ratio, words):
 
     """
     lemmas.sort(key=lambda s: scores[s], reverse=True)
-    length = len(lemmas) * ratio if words is None else words
+    length = len(lemmas) * ratio if words is None else words if words <= len(lemmas) else len(lemmas)
     return [(scores[lemmas[i]], lemmas[i],) for i in range(int(length))]
 
 
--- a/gensim/summarization/keywords.py
+++ b/gensim/summarization/keywords.py
@@ -302,7 +302,7 @@ def _extract_tokens(lemmas, scores, ratio, words):
 
     """
     lemmas.sort(key=lambda s: scores[s], reverse=True)
-    length = len(lemmas) * ratio if words is None else words if words <= len(lemmas) else len(lemmas)
+    length = len(lemmas) * ratio if words is None else min(words, len(lemmas))
     return [(scores[lemmas[i]], lemmas[i],) for i in range(int(length))]
 
 
--- a/gensim/test/test_keywords.py
+++ b/gensim/test/test_keywords.py
@@ -106,6 +106,7 @@ class TestKeywordsTest(unittest.TestCase):
         text = 'Test string small length'
         kwds = keywords(text, words=5, split=True)
         self.assertIsNotNone(kwds)
+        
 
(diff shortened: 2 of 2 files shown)


gensim/summarization/keywords.py
```diff
@@ -302,7 +302,7 @@ def _extract_tokens(lemmas, scores, ratio, words):
 
     """
     lemmas.sort(key=lambda s: scores[s], reverse=True)
-    length = len(lemmas) * ratio if words is None else words
+    length = len(lemmas) * ratio if words is None else words if words <= len(lemmas) else len(lemmas)
```
piskvorky:
```suggestion
    length = len(lemmas) * ratio if words is None else min(words, len(lemmas))
```

More readable.

carterols: Applied change in latest commit

carterols:
@mpenkov I agree. I can write some tests for this and include it in the PR.

PR #2751, opened 2020-02-08
remove mention of py27

mattf:
on 25 oct 2019, setup.py was updated to require python 3.5. this change removes the suggestion of testing against py27.


mpenkov:
Thank you for your contribution @mattf !

PR #2761, opened 2020-02-29
Fix KeyedVectors.add matrix type

menshikh-iv:
## Problem

I created empty KV (for fill them later using `add` method) and after this operation I don't expect than datatype of KV internal matrix will be `np.float64`, instead of the type of added vectors.

## Code example

```python
import numpy as np


kv = KeyedVectors(2)
words, vectors = ["a"], np.array([1., 1.], dtype=np.float32).reshape(1, -1)


kv.add(words, vectors)

assert vectors.dtype == np.float32  # correct
assert kv.vectors.dtype == np.float32  # failed here
```

## Solution

Cast internal `KeyedVectors.vectors` attribute to dtype of `add(..., vectors)`

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -308,6 +308,7 @@ class BaseKeyedVectors(utils.SaveLoad):
             self.index2entity.append(entity)
 
         # add vectors for new entities
+        self.vectors = self.vectors.astype(weights.dtype)  # cast existing vectors to 'weights' type
         self.vectors = vstack((self.vectors, weights[~in_vocab_mask]))
 
         # change vectors for in_vocab entities if `replace` flag is specified
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -214,8 +214,8 @@ class Vocab(object):
 
 class BaseKeyedVectors(utils.SaveLoad):
     """Abstract base class / interface for various types of word vectors."""
-    def __init__(self, vector_size, dtype=REAL):
-        self.vectors = zeros((0, vector_size), dtype=dtype)
+    def __init__(self, vector_size):
+        self.vectors = zeros((0, vector_size), dtype=REAL)
         self.vocab = {}
         self.vector_size = vector_size
         self.index2entity = []
@@ -376,8 +376,8 @@ class BaseKeyedVectors(utils.SaveLoad):
 
 class WordEmbeddingsKeyedVectors(BaseKeyedVectors):
     """Class containing common methods for operations over word vectors."""
-    def __init__(self, vector_size, dtype=REAL):
-        super(WordEmbeddingsKeyedVectors, self).__init__(vector_size=vector_size, dtype=REAL)
+    def __init__(self, vector_size):
+        super(WordEmbeddingsKeyedVectors, self).__init__(vector_size=vector_size)
         self.vectors_norm = None
         self.index2word = []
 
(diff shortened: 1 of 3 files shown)


gensim/models/keyedvectors.py
```diff
@@ -308,6 +308,7 @@ def add(self, entities, weights, replace=False):
             self.index2entity.append(entity)
 
         # add vectors for new entities
+        self.vectors = self.vectors.astype(weights.dtype)  # cast existing vectors to 'weights' type
```
gojomo: If I understand `astype()` correctly, this will always create a new copy of the underlying array (possibly quite wasteful for a large `vectors` or one that's already the proper type). Also, for consistency, shouldn't the type always be `float32`, rather than matching `weights`? So, wouldn't it be better to: (1) ensure even the empty initialization of 'vectors' just specifies the right `float32` type; & (2) always cast the passed-in `weights` (if necessary) to the proper type?

gensim/test/test_keyedvectors.py
```diff
@@ -255,6 +255,15 @@ def test_add_multiple(self):
         for ent, vector in zip(entities, vectors):
             self.assertTrue(np.allclose(kv[ent], vector))
 
+    def test_add_type(self):
+        kv = EuclideanKeyedVectors(2)
```
gojomo: `EuclideanKeyedVectors` is a nonsensical alias to `KeyedVectors` based on some deprecated misfactoring, so shouldn't be used.

gensim/test/test_keyedvectors.py
```diff
@@ -255,6 +255,15 @@ def test_add_multiple(self):
...
+        words, vectors = ["a"], np.array([1., 1.], dtype=np.float32).reshape(1, -1)
+
+        assert kv.vectors.dtype == np.float64  # default dtype of empty KV
```
gojomo: Per comment on `keyedvectors.py`, it'd be better to banish any `float64` arrays from appearing here, rather than checking/adapting to it.

gensim/models/keyedvectors.py
menshikh-iv: > shouldn't the type always be float32, rather than matching weights? 

good question, to be honest, I'm not sure.

> (1) ensure even the empty initialization of 'vectors' in KeyedVectors.__init__() just specifies the right float32 type from the start;

I like that idea, maybe I'll add `dtype` argument to init function with the default `float32`, wdyt?

> (2) always cast the usually-smaller passed-in weights (as necessary) to the proper type

if we implement `(1)` - I agree

gensim/test/test_keyedvectors.py
menshikh-iv: Fixed 38daff5 (just get rid alias from import)

gensim/test/test_keyedvectors.py
menshikh-iv: Fixed e031fc4

gensim/models/keyedvectors.py
menshikh-iv: I added datatype as parameter (with f32 default) - e031fc4

gojomo: Defaulting to `float32` is definitely good; I think there's a chance that letting it ever be anything else would cause problems, perhaps SEGFAULT/crashing problems, in any Cython code that might wind up operating on such not-as-expected vectors.

menshikh-iv: Agree, in that case, I'll remove parametrization and just "pin" f32 everywhere (to avoid potential issues with Cython)

menshikh-iv: Fixed 7c7ce19

gensim/models/keyedvectors.py
```diff
@@ -215,7 +215,7 @@ def __str__(self):
 class BaseKeyedVectors(utils.SaveLoad):
     """Abstract base class / interface for various types of word vectors."""
     def __init__(self, vector_size):
-        self.vectors = zeros((0, vector_size))
+        self.vectors = zeros((0, vector_size), dtype=REAL)
```
mpenkov: Is there a reason why we don't explicitly say `float32` here? What's the benefit of using `REAL` as an alias?

menshikh-iv: idk, just "mimic" to existing code

PR #2698, opened 2019-12-05
KeyedVectors  & *2Vec API streamlining, consistency

gojomo:
Removing some of the layers of unnecessary complexity from #1777 & related wrong-turns. Stripping out deprecations that were marked as to-be-removed in gensim-4.0.0. 

High-level summary for release notes:

"Unify KeyedVectors into one consistently re-used & more-capable class. Improve its memory-efficiency. Correct memory inefficiencies in FastText. Re-integrate excessive abstract superclasses & helper classes into fewer single-model classes. Rationalize method & property names, removing previous-deprecations & other inconsistent/outdated options."

PR #2764, opened 2020-03-03
Fix fastText word_vec() for OOV words with use_norm=True

### LINKED ISSUE #2763: FastText OOV word embedding are calculated incorrectly when passing `use_norm=True`
#### Problem description

FastText OOV word embedding are calculated incorrectly when passing `use_norm=True` (particularly, when looking for most similar words).  The library first normalizes n-gram vectors, then averages them. But it should average them first; otherwise, results are inconsistent.

What happens: cosine similarities used for neighbor retrieval are different from similarities calculated directly from word vectors. 

Why it happens:
* usually when calculating vectors for OOV words fasttext calculates average of n-gram vectors
* but if we pass `use_norm=True`, then fasttext calculates average of *L2-normalized* n-gram vectors ([code](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/keyedvectors.py#L2090)). And it is wrong!
* when we lookup for most similar words, we use just this option, `use_norm=True` ([code](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/keyedvectors.py#L831)), how unfortunate!
* why averaging normalized vectors is wrong: because it was never done when model was trained, and is normally never done when the model is applied, so such vectors are most probably meaningless.
* how to do it right: *first* average n-gram vectors, and *then* normalize them. 

#### Call to action: 
Rewrite `word_vec` method for FastTextKeyedVectors to apply normalization and averaging in the rigth order. 

```
word = 'some_oov_word'
pairs = model.most_similar(word)
top_neighbor, top_simil = pairs[0]
print(top_simil)
print(model.cosine_similarities(model[word], model[top_neighbor].reshape(1, -1))[0])
```
these two prints are expected to produce identical numbers (similarity between the oov words and its closest neighbor), but in fact the numbers are different.

This [notebook](https://gist.github.com/avidale/c6b1d13b32a36f19750cd01148560561) reproduces the problem with a particular model for Russian language, but it is relevant for any language. 

#### Versions

The output of 
```python
```
is
```
Linux-4.14.137+-x86_64-with-Ubuntu-18.04-bionic
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
NumPy 1.17.5
SciPy 1.4.1
gensim 3.8.1
FAST_VERSION 1
```

avidale:
Fixes #2763

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -2114,10 +2114,7 @@ class FastTextKeyedVectors(WordEmbeddingsKeyedVectors):
             raise KeyError('cannot calculate vector for OOV word without ngrams')
         else:
             word_vec = np.zeros(self.vectors_ngrams.shape[1], dtype=np.float32)
-            if use_norm:
-                ngram_weights = self.vectors_ngrams_norm
-            else:
-                ngram_weights = self.vectors_ngrams
+            ngram_weights = self.vectors_ngrams
             ngram_hashes = ft_ngram_hashes(word, self.min_n, self.max_n, self.bucket, self.compatible_hash)
             if len(ngram_hashes) == 0:
                 #
@@ -2132,7 +2129,10 @@ class FastTextKeyedVectors(WordEmbeddingsKeyedVectors):
                 return word_vec
             for nh in ngram_hashes:
                 word_vec += ngram_weights[nh]
-            return word_vec / len(ngram_hashes)
+            result = word_vec / len(ngram_hashes)
+            if use_norm:
+                result /= sqrt(sum(result ** 2))
+            return result
 
     def init_sims(self, replace=False):
         """Precompute L2-normalized vectors.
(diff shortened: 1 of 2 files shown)


gojomo:
At the moment (through #e7f55c), seems like this only includes a test (good!), but not yet a fix?

avidale:
> At the moment (through [e7f555c](https://github.com/RaRe-Technologies/gensim/commit/e7f555c956839f24c021d5754aca643b1f86c63c)), seems like this only includes a test (good!), but not yet a fix?

That's right. I'll remove the WIP tag after the fix is included.

gojomo:
Fix looks good - though my slightly-different (using `np.linalg.norm`) approach in #2698 will likely supercede. Test is valuable for self-consistency of similarity-paths (one explicitly passing through `use_norm=True`, one not). 

+1 for 'squash & merge'!

mpenkov:
@avidale Congrats on your first PR to gensim! :1st_place_medal:

Issue #2105, opened 2018-06-26
AssertionError: sparse documents must not contain any explicit zero entries and the similarity matrix S must satisfy x^T * S * x > 0 for any nonzero bag-of-words vector x.


DennisCologne:
Hello there,

Maybe you can help me out with this real quick. I cannot run any of your examples. Not the one from https://radimrehurek.com/gensim/similarities/docsim.html, nor the one from this repo. All of them give me the following Assertion. 

```
AssertionError: sparse documents must not contain any explicit zero entries and the similarity matrix S must satisfy x^T * S * x > 0 for any nonzero bag-of-words vector x.
```

This is not working (other similaritiy measures of this module work fine):

```python
from gensim.test.utils import common_texts
from gensim.corpora import Dictionary
from gensim.models import Word2Vec
from gensim.similarities import SoftCosineSimilarity

model = Word2Vec(common_texts, size=20, min_count=1)  # train word-vectors
dictionary = Dictionary(common_texts)
bow_corpus = [dictionary.doc2bow(document) for document in common_texts]

similarity_matrix = model.wv.similarity_matrix(dictionary)  # construct similarity matrix
index = SoftCosineSimilarity(bow_corpus, similarity_matrix, num_best=10)

# Make a query.
query = 'graph trees computer'.split()
# calculate similarity between query and each doc from bow_corpus
sims = index[dictionary.doc2bow(query)]
```

Neither is this from the repo (I followed all previous steps):

```python
similarity = softcossim(sentence_obama, sentence_orange, similarity_matrix)
print('similarity = %.4f' % similarity)
```

Thanks in advance. I am trying to run this for two days now but nothing works.

Best,
Dennis

piskvorky:
@Witiko can you have a look?

DennisCologne:
Hi @Witiko,

thank you for your answer. 

Actually, it is Python 2.7.14 with Gensim 3.4.0... after further investigation, the matrix-vector multiplication returns a negative value even though all of the values in both are positive.

But you are right, I just tried it on my Python 3.6 environment and there it works fine. 
I guess I will use this environment than. But this problem might still be interesting for you.

Thanks again for the quick reply.

Best, 
Dennis

Witiko:
Hey @DennisCologne,

[...]

PR #2783, opened 2020-04-04
Reduce memory use of the term similarity matrix constructor, deprecate the positive_definite parameter, and extend normalization capabilities of the inner_product method

Witiko:
## Reduce memory use of the term similarity matrix constructor
Previously, the term similarity matrix constructor would use the dictionary of keys (DOK) sparse matrix format to incrementally build a sparse matrix. Since DOK uses a dictionary of python objects to keep track of assigned array indices, this was memory-consuming for large term similarity matrices.

This PR implements one of the future work items from #1827 and #2016. In the term similarity matrix constructor, DOK is replaced with three [native arrays](https://docs.python.org/3/library/array.html), as described in [a blog post][kula15] by @maciejkula. When constructing a symmetric term similarity matrix, an additional set of indices is used.

We achieve:
- 8× memory use reduction when constructing asymmetric term similarity matrices and
- 2× memory use reduction when constructing symmetric term-similarity matrices.

 [kula15]: https://maciejkula.github.io/2015/02/22/incremental-construction-of-sparse-matrices/

## Deprecate the `positive_definite` parameter
The term similarity matrix `positive_definite` parameter has a misleading name, because it does not ensure the [positive definiteness][] of the term similarity matrix but only [strong diagonal dominance][diagonal dominance]. Strong diagonal dominance together with [symmetry][] are then sufficient for positive definiteness. Positive definiteness is required for a unique [Cholesky factorization][] of the term similarity matrix into word embeddings, but it seems to improve the accuracy of the soft cosine similarity in general [[1][]].

 [diagonal dominance]: https://en.wikipedia.org/wiki/Diagonally_dominant_matrix
 [cholesky factorization]: https://en.wikipedia.org/wiki/Cholesky_decomposition
 [positive definiteness]: https://en.wikipedia.org/wiki/Definiteness_of_a_matrix
 [symmetry]: https://en.wikipedia.org/wiki/Symmetric_matrix

This PR renames `positive_definite` to `dominant` and updates the documentation to make the effect of the parameter clear. An example of converting a term similarity matrix into word embeddings using Cholesky factorization has also been added. For backwards compatibility, a new keyword argument `positive_definite` has been added. This keyword argument acts as an alias for `dominant` and emits a deprecation warning when used.

## Extend normalization capabilities of the `inner_product` method

Currently, the inner_product Xᵀ · S · Y (where X is a query, Y is a document, and S = E · Eᵀ is a term similarity matrix) supports the following vector normalization schemes:

1. no normalization (X), the inner product corresponds to query expansion, and
2. ℓ₂-normalization (X / sqrt[Xᵀ · S · X]), the inner product corresponds to the [soft cosine measure][].

The ℓ₂-normalization (2) is natural in [TF-IDF][] models with cosine normalization (SMART `??c`) but likely harmful in others. With no normalization (1), Xᵀ · S may easily become large when S is dense.

This PR implements a third vector normalization scheme:

3. maintain ℓ₂ norm (X / sqrt[{Xᵀ · S · X} / {Xᵀ · X}]), the inner product corresponds to query expansion with fractional membership. 

This scheme maintains X's ℓ₂ norm during the change of basis E without enforcing unit length, and should be better suited to TF-IDF models without cosine normalization (SMART `??n`, `??b`, and `??u`).

Below are examples of the schemes for X = [1, 0.5, 1], Y = [0.5, 1, 1], S = [[1, 1, 0], [1, 1, 0], [0, 0, 1]]:

1. [1, 0.5, 1]ᵀ · S · [0.5, 1, 1] = [1.5, 1, 1]ᵀ · [0.5, 1, 1] = 2.63
2. [0.6, 0.3, 0.6]ᵀ · S · [0.3, 0.6, 0.6] = [0.75, 0.6, 0.6]ᵀ · [0.3, 0.6, 0.6] = 0.95
3. [0.9, 0.45, 0.9]ᵀ · S · [0.45, 0.9, 0.9] = [1.13, 0.9, 0.9]ᵀ · [0.45, 0.9, 0.9] = 2.15

Currently, a vector normalization scheme applies to both queries and documents. However, when queries and documents use different TF-IDF models, they may benefit from using different schemes. This PR implements separate vector normalization schemes for queries (X) and documents (Y).

## References
1. Vít Novotný, Eniafe Festus Ayetiran, Michal Štefánik, Petr Sojka. *Text classification with word embedding regularization and soft similarity measure* (preprint), 2020. ([link to PDF][1], [link to HTML][2])

 [1]: https://arxiv.org/pdf/2003.05019 (Text classification with word embedding regularization and soft similarity measure)
 [2]: https://www.fi.muni.cz/~xnovot32/regularized-embeddings-distill/public/ (Text classification with word embedding regularization & soft similarity measure)
 [soft cosine measure]: https://en.wikipedia.org/wiki/Cosine_similarity#Soft_cosine_measure (Soft cosine measure)
 [tf-idf]: https://radimrehurek.com/gensim/models/tfidfmodel.html (models.tfidfmodel – TF-IDF model)


Witiko:
Commits 7bae2863bcd935eed33abf446be17b23d278b4cc through 7bae2863bcd935eed33abf446be17b23d278b4cc triggers the following error in circleci, see [the log](https://circleci.com/gh/RaRe-Technologies/gensim/3231):

> ModuleNotFoundError: No module named 'matplotlib'

The check fail due to a missing matplotlib dependency of sphinx and can be fixed as follows:

```
diff --git a/requirements_docs.txt b/requirements_docs.txt
index f3dc318f..d8f76294 100644
--- a/requirements_docs.txt
+++ b/requirements_docs.txt
@@ -12,3 +12,4 @@ testfixtures==6.10.0
 statsmodels==0.10.1
 pyemd==0.5.1
 pandas==0.25.2
+matplotlib==3.2.1
diff --git a/setup.py b/setup.py
index 5422a32a..c250e20b 100644
--- a/setup.py
+++ b/setup.py
@@ -304,6 +304,7 @@ docs_testenv = linux_testenv + distributed_env + [
     'statsmodels',
     'pyemd',
     'pandas',
+    'matplotlib',
 ]
 
 if sys.version_info < (3, 7):
```

Witiko:
The commits d317c9a51cee3b65796e822d8eca7aad5892230e and d71535fb5598c62998c8e3a21c1c0ffb641b32e6 add examples to the documentation of `SparseTermSimilarityMatrix` and triggers another error in circleci, see [the log](https://circleci.com/gh/RaRe-Technologies/gensim/3230):

> docstring of gensim.corpora._mmreader.MmReader.num_docs:1:duplicate object description of gensim.corpora._mmreader.MmReader.num_docs, other instance in corpora/_mmreader, use > :noindex: for one of them

I have not been able to pinpoint the cause of this failure.

Witiko:
Neither of the above two failures seem related to my changes, i.e. they seem to be due to the current state of the `develop` branch. Would you like me to fix the issues, @mpenkov, or should I wait for develop to stabilize before a release?

Witiko:
@mpenkov, @piskvorky: I added another related patch (*Reduce memory use of term similarity matrix constructor*) to this PR, since it builds on this PR and the improvements it brings seemed too great to wait until this PR is merged.

Witiko:
@mpenkov Anything I can do to help move this PR along?

Witiko:
@mpenkov, @piskvorky: I added a related patch (Extend normalization in the `inner_product` method) to this PR.

gensim/similarities/termsim.py
```diff
@@ -149,44 +161,53 @@ class SparseTermSimilarityMatrix(SaveLoad):
...
     ----------
     matrix : :class:`scipy.sparse.csc_matrix`
         The encapsulated sparse term similarity matrix.
-    """
-    PROGRESS_MESSAGE_PERIOD = 1000  # how many columns are processed between progress messages
 
-    def __init__(self, source, dictionary=None, tfidf=None, symmetric=True, positive_definite=False, nonzero_limit=100,
-                 dtype=np.float32):
+    """
+    def __init__(self, source, dictionary=None, tfidf=None, symmetric=True, dominant=False,
```
mpenkov, line 198: This initializer does a lot of work... I think it would be helpful to reduce it. On top of that, the whole pattern of re-calling the initializer from itself is ugly. I think this is a good opportunity to rework it.

Start with this:

```python
def _create_source(index, dictionary=None, tfidf=None, symmetric=True, ...):
    # do the heavy lifting here
    return matrix 
```

and then simplify the initializer to:

```python
# deal with the old positive_definite arg here and then forget about it

if not sparse.issparse(source):
    source = _create_source(source, dictionary=dictionary, symmetric=symmetric, ...)
self.matrix = source.tocsc()
```

gensim/similarities/termsim.py
```diff
@@ -207,24 +228,41 @@ def __init__(self, source, dictionary=None, tfidf=None, symmetric=True, positive
...
+        nonzero_counter_dtype = _shortest_uint_dtype(nonzero_limit)
+
+        column_nonzero = np.array([0] * matrix_order, dtype=nonzero_counter_dtype)
+        if dominant:
+            column_sum = np.zeros(matrix_order, dtype=dtype)
+        if symmetric:
+            assigned_cells = set()
+        row_buffer = array('Q')
+        column_buffer = array('Q')
+        data_buffer = array(similarity_type_code)
```
mpenkov, line 251: Here's an idea:

```python
# buffer population functions

def populate_buffer_symmetric(t1_index, t2_index, similarity):
    assigned_cells.add((t1_index, t2_index))
    column_buffer.append(t1_index)
    row_buffer.append(t2_index)
    data_buffer.append(similarity)

    assigned_cells.add((t2_index, t1_index))
    column_buffer.append(t2_index)
    row_buffer.append(t1_index)
    data_buffer.append(similarity)

def populate_buffer_assymetric(t1_index, t2_index, similarity):
    column_buffer.append(t1_index)
    row_buffer.append(t2_index)
    data_buffer.append(similarity)
```

Then you can do a little bit less work on those nested for-blocks.

gensim/similarities/termsim.py
```diff
@@ -293,17 +353,43 @@ def inner_product(self, X, Y, normalized=False):
...
+
+        if normalized_X not in valid_normalized_values:
+            raise ValueError('{} is not a valid value of normalize'.format(normalized_X))
+        if normalized_Y not in valid_normalized_values:
+            raise ValueError('{} is not a valid value of normalize'.format(normalized_Y))
+
         is_corpus_X, X = is_corpus(X)
         is_corpus_Y, Y = is_corpus(Y)
 
+        non_negative_norm_assertion_message = u"sparse documents must not contain any explicit " \
```
mpenkov, line 389: This can be a internal (module-level) constant outside of the function.

gensim/similarities/termsim.py
```diff
@@ -315,22 +401,37 @@ def inner_product(self, X, Y, normalized=False):
...
 
-                result /= sqrt(X_norm) * sqrt(Y_norm)
+            if normalized_X is True and normalized_Y is True:
                 result = np.clip(result, -1.0, 1.0)
 
             return result[0, 0]
         elif not is_corpus_X or not is_corpus_Y:
             if is_corpus_X and not is_corpus_Y:
-                is_corpus_X, X, is_corpus_Y, Y = is_corpus_Y, Y, is_corpus_X, X  # make Y the corpus
+                is_corpus_X, X, is_corpus_Y, Y, normalized_X, normalized_Y = \
```
mpenkov, line 433: That line break is ugly, I think multi-line assignments would be more readable here.

gensim/similarities/termsim.py
```diff
@@ -315,22 +401,37 @@ def inner_product(self, X, Y, normalized=False):
 
             result = X.T.dot(matrix).dot(Y)
 
-            if normalized:
+            norm = 1.0
+
+            if normalized_X:
```
mpenkov, line 406: If we're doing the same thing twice here, then it's better to make it a function and re-use it.

Usually, the rule is "three or more" but in this case the code is sufficiently complicated (at least to me as the maintainer) to warrant simplifying it.

gensim/similarities/termsim.py
```diff
@@ -344,23 +445,30 @@ def inner_product(self, X, Y, normalized=False):
...
-                assert \
-                    X_norm.min() > 0.0 and Y_norm.min() >= 0.0, \
-                    u"sparse documents must not contain any explicit zero entries and the similarity matrix S " \
-                    u"must satisfy x^T * S * x > 0 for any nonzero bag-of-words vector x."
+            # use the following equality: np.diag(A.T.dot(B).dot(A)) == A.T.dot(B).multiply(A.T).sum(axis=1).T
 
-                X = np.multiply(X, 1 / np.sqrt(X_norm)).T
-                Y = np.multiply(Y, 1 / np.sqrt(Y_norm))
-                Y = np.nan_to_num(Y)  # Account for division by zero when Y_norm.min() == 0.0
+            if normalized_X:
```
mpenkov, line 451: As above.

gensim/similarities/termsim.py
```diff
@@ -373,23 +481,29 @@ def inner_product(self, X, Y, normalized=False):
             Y = corpus2csc(Y if is_corpus_Y else [Y], num_terms=self.matrix.shape[0], dtype=dtype)
             matrix = self.matrix
 
-            if normalized:
-                # use the following equality: np.diag(A.T.dot(B).dot(A)) == A.T.dot(B).multiply(A.T).sum(axis=1).T
+            # use the following equality: np.diag(A.T.dot(B).dot(A)) == A.T.dot(B).multiply(A.T).sum(axis=1).T
+
+            if normalized_X:
```
mpenkov, line 486: As above.

gensim/similarities/termsim.py
```diff
@@ -52,6 +55,16 @@ def most_similar(self, term, topn=10):
         """
         raise NotImplementedError
 
+    def __str__(self):
+        return "%s(%s)" % (
```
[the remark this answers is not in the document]
Witiko: Fixed in ab413465.

gensim/similarities/termsim.py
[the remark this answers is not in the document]
Witiko: That would break backwards compatibility. If a user calls `SparseTermSimilarityMatrix(source, None, None, True, False)`, we want the final `False`, which previously specified the `positive_definite` parameter to specify the `dominant` parameter.

gensim/similarities/termsim.py
```diff
@@ -117,28 +130,27 @@ class SparseTermSimilarityMatrix(SaveLoad):
...
-    Future work should switch to building arrays of rows, columns, and non-zero elements and
-    directly passing these arrays to the CSC matrix constructor without copying.
-
     Examples
     --------
     >>> from gensim.test.utils import common_texts
     >>> from gensim.corpora import Dictionary
     >>> from gensim.models import Word2Vec, WordEmbeddingSimilarityIndex
     >>> from gensim.similarities import SoftCosineSimilarity, SparseTermSimilarityMatrix
+    >>> from gensim.similarities.index import AnnoyIndexer
```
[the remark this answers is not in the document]
Witiko: I prodded flake8 with 6557b849732b314570ea9d5132f1731d964e2fe6 to find out. [This makes kills Travis](https://travis-ci.org/github/RaRe-Technologies/gensim/jobs/710852563) with the following error:

```
gensim/similarities/termsim.py:565:9: E402 module level import not at top of file
ERROR: InvocationError for command /home/travis/build/RaRe-Technologies/gensim/.tox/flake8-docs/bin/flake8-rst gensim/ docs/ (exited with code 1)
```

gensim/similarities/termsim.py
```diff
@@ -207,24 +228,41 @@ def __init__(self, source, dictionary=None, tfidf=None, symmetric=True, positive
...
-                    "PROGRESS: at %.02f%% columns (%d / %d, %.06f%% density, "
-                    "%.06f%% projected density)",
-                    100.0 * (column_number + 1) / matrix_order, column_number + 1, matrix_order,
-                    100.0 * matrix.getnnz() / matrix_order**2,
-                    100.0 * np.clip(
-                        (1.0 * (matrix.getnnz() - matrix_order) / matrix_order**2)
-                        * (1.0 * matrix_order / (column_number + 1))
-                        + (1.0 / matrix_order),  # add density correspoding to the main diagonal
-                        0.0, 1.0))
+                    key=lambda x: (lambda term_index, term_idf: (term_idf, -term_index))(*x),
```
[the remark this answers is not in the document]
Witiko: Fixed in 03a536c493c023eb4007e6aeb18d3ec0d3dcae41.

gensim/similarities/termsim.py
```diff
@@ -240,34 +278,55 @@ def __init__(self, source, dictionary=None, tfidf=None, symmetric=True, positive
             else:
                 rows = sorted(
                     most_similar,
-                    key=lambda x: (lambda term_index, _: (tfidf.idfs[term_index], -term_index))(*x), reverse=True)
+                    key=lambda x: (lambda term_index, _: (tfidf.idfs[term_index], -term_index))(*x),
```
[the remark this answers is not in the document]
Witiko: Fixed in 03a536c493c023eb4007e6aeb18d3ec0d3dcae41.

gensim/similarities/termsim.py
```diff
@@ -293,17 +353,43 @@ def inner_product(self, X, Y, normalized=False):
...
+
         """
         if not X or not Y:
             return self.matrix.dtype.type(0.0)
 
+        if normalized in (True, False):
+            warnings.warn(
+                'Boolean parameter normalized will be removed in 4.0.0, use '
+                'normalized=({normalized}, {normalized}) instead of '
+                'normalized={normalized}'.format(normalized=normalized),
```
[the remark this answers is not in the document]
Witiko: Sorry, fixed in be6810e2.

gensim/test/test_similarities.py
```diff
@@ -1045,12 +1098,70 @@ def test_inner_product(self):
         result = matrix.inner_product(vec1, vec2)
         self.assertAlmostEqual(expected_result, result, places=5)
 
+        vec1 = self.dictionary.doc2bow([u"government", u"government", u"denied"])
```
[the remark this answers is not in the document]
Witiko: The issue here is that the problem is two-dimensional. We want to test the `inner_product` method for all combinations of vector types and normalization schemes. In 06540c5b, I sliced `test_inner_product` along the first axis into `test_inner_product_vector_zerovector`, `test_inner_product_vector_vector`, `test_inner_product_vector_corpus`, `test_inner_product_corpus_vector`, `test_inner_product_corpus_corpus`. Should I slice the unit tests along the second axis as well?

gensim/similarities/termsim.py
```diff
@@ -207,24 +228,41 @@ def __init__(self, source, dictionary=None, tfidf=None, symmetric=True, positive
...
-                        * (1.0 * matrix_order / (column_number + 1))
-                        + (1.0 / matrix_order),  # add density correspoding to the main diagonal
-                        0.0, 1.0))
+                    key=lambda x: (lambda term_index, term_idf: (term_idf, -term_index))(*x),
+                    reverse=True,
+                )
+            ]
+
+        if dtype is np.float16 or dtype is np.float32:
+            similarity_type_code = 'f'
```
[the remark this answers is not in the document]
Witiko: Fixed in 0f57f0a7.

gensim/similarities/termsim.py
```diff
@@ -240,34 +278,55 @@ def __init__(self, source, dictionary=None, tfidf=None, symmetric=True, positive
...
             for row_number, (t2_index, similarity) in zip(range(num_rows), rows):
-                if positive_definite and column_sum[t1_index] + abs(similarity) >= 1.0:
+                if dominant and column_sum[t1_index] + abs(similarity) >= 1.0:
                     break
                 if symmetric:
                     if column_nonzero[t2_index] < nonzero_limit \
-                            and (not positive_definite or column_sum[t2_index] + abs(similarity) < 1.0) \
-                            and not (t1_index, t2_index) in matrix:
-                        matrix[t1_index, t2_index] = similarity
+                            and (not dominant or column_sum[t2_index] + abs(similarity) < 1.0) \
```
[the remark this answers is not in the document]
Witiko: Fixed in 0a7c6542.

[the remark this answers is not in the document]
Witiko: Fixed in 0a7c6542.

gensim/similarities/termsim.py
Witiko: In fact, we can simplify this much further, see 0a7c6542.

gensim/similarities/termsim.py
Witiko: Fixed in 621b2525631b4146f4a301f1f84e47500869fdf0.

gensim/similarities/termsim.py
Witiko: Fixed in dc26e56ad4b3f707c94d3b542aecb64996fc170b.

gensim/similarities/termsim.py
Witiko: Fixed in 5d68e131.

gensim/similarities/termsim.py
Witiko: Fixed in c42b7cc8.

gensim/similarities/termsim.py
Witiko: Fixed in c42b7cc8, as above.

gensim/similarities/termsim.py
Witiko: Fixed in c42b7cc8, as above.

Witiko:
> Hi @Witiko , thank you for such a well-documented PR. I'm sorry it took me so long to get around to reviewing it. I got DOS'ed by real life. Anyway, I left you some comments, please have a look.

Hello, @mpenkov; it's no problem, I can relate. I responded to your review and merged the current develop. Let me know if you're happy with the code, or if there are other things you'd like to see polished.

Witiko:
@mpenkov I just added b10006e that fixes an issue mentioned in https://github.com/RaRe-Technologies/gensim/issues/2105#issuecomment-665205201 by checking that `dictionary` is not empty an raising a `ValueError` otherwise. This is a small change, and is documented and tested.

gensim/similarities/termsim.py
```diff
@@ -156,32 +140,288 @@ def most_similar(self, t1, topn=10):
...
+    matrix : :class:`scipy.sparse.coo_matrix`
+        The sparse term similarity matrix.
+
+    """
+    assert isinstance(index, TermSimilarityIndex)
+    assert dictionary is not None
+    matrix_order = len(dictionary)
+
+    if matrix_order == 0:
+        raise ValueError('The provided dictionary is empty')
```
piskvorky, line 184: That's more a statement of fact than an error :) What's the actual problem?

Ideally, we want to show user the error (what's wrong), plus also suggest a step to overcome the problem (not sure that applies here, just in general).

gensim/similarities/termsim.py
```diff
@@ -156,32 +140,288 @@ def most_similar(self, t1, topn=10):
...
+
+    logger.info("constructing a sparse term similarity matrix using %s", index)
+
+    if nonzero_limit is None:
+        nonzero_limit = matrix_order
+
+    def tfidf_sort_key(term_index):
+        if isinstance(term_index, tuple):
+            term_index, *_ = term_index
```
piskvorky, line 193: I'm not familiar with this `*_` syntax – what does it do?

Witiko: It's [PEP 3132](https://www.python.org/dev/peps/pep-3132) (Extended Iterable Unpacking), here synonymous with `term_index = term_index[0]`.  
What's curious is that although this syntax is not available until after Python 3.0, the CI is passing.

gensim/similarities/termsim.py
Witiko: I am not sure we have enough information here to say anything more meaningful. We may want to prohibit the creation of empty dictionaries or at least raise a warning if we worry that this may break existing code bases.

piskvorky: Even then, a message like `Dictionary in SparseTermSimilarity must not be empty` (or similar) is better.

Witiko: I updated the message in 8584880.

gensim/similarities/termsim.py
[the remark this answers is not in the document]
Witiko: Actually, can we discuss this? Gensim 3.8.3 supports Python 2.7 and Python 3.5+ and Gensim 4.0 only supports Python 3.5+. The `string.format` is available in Python 2.7 and since Python 3 as per [PEP 3101](https://www.python.org/dev/peps/pep-3101/), i.e. its usage should be non-problematic in both Gensim 3.8.3 and Gensim 4.0.

Witiko:
Continuous integration is failing on the unrelated `gensim.test.test_keras_integration` module.

piskvorky:
@Witiko merge `develop` into your PR.

Witiko:
Thank you, now the checks are passing.

gensim/similarities/termsim.py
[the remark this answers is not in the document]
mpenkov: It's a code style issue, not a compatibility issue. My understanding is @piskvorky dislikes .format, likes f-strings and is OK with the old-school percent formatting. This is my preference as well.

mpenkov:
@Witiko Thank you for this PR and your patience. Merged.

gensim/similarities/termsim.py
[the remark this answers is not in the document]
Witiko: I see. Aren't f-strings only available in Python 3.6+, whereas Gensim supports Python 3.5+, though? [[PEP 498]]

 [PEP 498]: https://www.python.org/dev/peps/pep-0498/

PR #2789, opened 2020-04-11
Fix word2vec doc-comment example of KeyedVectors usage, issue 2669

### LINKED ISSUE #2669: word2vec doc-comment example of KeyedVectors usage broken
The usage example in the word2vec.py doc-comment regarding `KeyedVectors` uses inconsistent paths and thus doesn't work. 

https://github.com/RaRe-Technologies/gensim/blob/e859c11f6f57bf3c883a718a9ab7067ac0c2d4cf/gensim/models/word2vec.py#L73

https://github.com/RaRe-Technologies/gensim/blob/e859c11f6f57bf3c883a718a9ab7067ac0c2d4cf/gensim/models/word2vec.py#L76

If vectors were saved to a tmpfile-path based on the filename `'wordvectors.kv'`, they need to loaded from that same path, not some other local-directory file named 'model.wv'.

(Also, in my opinion the use of `get_tmpfile()` adds unnecessary extra complexity to this example. People usually **don't** want their models in a "temp" directory, which some systems will occasionally delete, so the examples might as well do the simplest possible thing: store in the current working directory with simple string filenames. The example code above this is also confused, because it creates a temp-file path, but then doesn't actually use it, choosing to do the simple & right thing with a local file instead.)

bvinesh:
Fixes #2669 
Proposal is to remove use of get_tmpfile and (optionally) change to a temp directory instead. Let me know if you prefer a different approach and I will change it. Thank you.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -36,10 +36,13 @@ Initialize a model with e.g.:
 
 .. sourcecode:: pycon
 
-    >>> from gensim.test.utils import common_texts, get_tmpfile
+    >>> from gensim.test.utils import common_texts
     >>> from gensim.models import Word2Vec
+    >>> from tempfile import mkdtemp
+    >>> from os import chdir
     >>>
-    >>> path = get_tmpfile("word2vec.model")
+    >>> path = mkdtemp() # make a note of this path if you need it for later
+    >>> chdir(path)
     >>>
     >>> model = Word2Vec(common_texts, size=100, window=5, min_count=1, workers=4)
     >>> model.save("word2vec.model")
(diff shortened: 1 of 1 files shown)


bvinesh:
@mpenkov , much appreciated if this could be reviewed. This is my first PR. Please let me know if any changes are required. Thank you.

mpenkov:
Note to self: resolve conflict

mpenkov:
@bvinesh Yes, let's get rid of the temporary subdirectory. Are you able to finish this PR?

piskvorky:
@bvinesh we're planning a new release of Gensim soon. We could include this PR if you finish it – let us know (or close it).

PR #2797, opened 2020-04-16
Prevent smart_open from reconfiguring the logging subsystem

mpenkov:
Work around for:

- https://github.com/RaRe-Technologies/smart_open/issues/474                                                                                                    
- https://github.com/RaRe-Technologies/smart_open/issues/475

### DIFF
--- a/gensim/__init__.py
+++ b/gensim/__init__.py
@@ -2,12 +2,26 @@
 of documents.
 """
 
-from gensim import parsing, corpora, matutils, interfaces, models, similarities, summarization, utils  # noqa:F401
 import logging
 
 __version__ = '3.8.1'
 
-
 logger = logging.getLogger('gensim')
 if len(logger.handlers) == 0:  # To ensure reload() doesn't add another one
     logger.addHandler(logging.NullHandler())
+
+#
+# Older versions of smart_open lack a NullHandler, meaning they can
+# inadvertedly call logging.basicConfig upon import.  This has been fixed in
+# newer versions.  The below NullHandler is necessary because gensim is pinned
+# to an older smart_open version for Python 2.7 users.
+#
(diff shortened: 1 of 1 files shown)


gensim/__init__.py
```diff
@@ -2,12 +2,27 @@
...
 __version__ = '3.8.1'
 
-
 logger = logging.getLogger('gensim')
 if len(logger.handlers) == 0:  # To ensure reload() doesn't add another one
     logger.addHandler(logging.NullHandler())
+
+#
+# Older versions of smart_open lack a NullHandler, meaning they can
+# inadvertedly call logging.basicConfig upon import.  This has been fixed in
```
piskvorky, line 15: I'm not sure this is the case – see my investigation at https://github.com/RaRe-Technologies/smart_open/pull/476#issuecomment-614493217

gojomo:
Why give `gensim` a workaround, rather than fix `smart_open`?

mpenkov:
@gojomo We are fixing smart_open as well. However, because we have pinned smart_open in gensim, Py2.7 users won't get the fix. This work-around is for them.

gojomo:
That makes some sense, but if there are other projects in the same boat as `gensim` – not wanting to yet force Py3, but still reliant on `smart_open` – the nice thing to do for them is to make a fixed `smart_open` 1.x. Otherwise they'll all have to do this same kludgey workaround, *in addition to* pinning/advising-their-users-to-force an older `smart_open`. 

I'm not sure there *are* any other such projects, but if there's a choice between multiple projects all inserting this same bit of workaround-code at the top-level `__init__.py` level to use `smart_open`, or `smart_open` just fixing the issue in one place and pointing to some version-spec as "last Py 2.7 version of smart_open", the latter is the more globally-efficient practice.

PR #2799, opened 2020-04-17
Fix typo in comments

Chenxin-Guo:
The rows of the corpus are actually documents, fix the comment to reduce confusion.


piskvorky:
Thanks! ⭐

PR #2772, opened 2020-03-22
Add "downloads" badge to README

piskvorky:
Trying the idea from https://github.com/RaRe-Technologies/smart_open/issues/440.

#### Motivation

In general, seeing how much others use a package improves confidence in its robustness and maintenance. A kind of social proof.

I see no downsides, except if the service that provides this badge (https://pepy.tech/) proves unstable or unreliable.

Badges for gitter and wheels seem useless, so I removed them.

#### Checklist

Before you create the PR, please make sure you have:

- [x] Picked a concise, informative and complete title
- [x] Clearly explained the motivation behind the PR
- [x] Linked to any existing issues that your PR will be solving
- [x] Included tests for any new functionality
- [x] Checked that all unit tests pass

PR #2806, opened 2020-04-23
Cache README badges

### LINKED ISSUE #2805: README badges fail to render
I see this often in many Github repos (not just Gensim):

Screen Shot 2020-04-23 at 15 35 54

According to https://github.com/badges/shields/issues/1568, this is an issue with Github's non-caching proxy (`https://camo.githubusercontent.com`) timing out when querying the original image (hosted at `https://img.shields.io`).

The proxy timeout limit seems to be 4 seconds. So if `img.shields.io` takes longer than 4 seconds to return the image, Github will show a broken image instead.

Github injects its proxy automatically to all `` links and I found no way to disable it.

A solution suggested [here](https://github.com/badges/shields/issues/1568#issuecomment-407860059) is to use *another* proxy, one that caches the img.shields.io response so that github's proxy doesn't time out. For example Google's caching proxy at `https://images1-focus-opensocial.googleusercontent.com/gadgets/proxy`.

Example of a cached "shields => google's cache => github's proxy => browser" badge:
`https://images1-focus-opensocial.googleusercontent.com/gadgets/proxy?container=focus&refresh=3600&url=https%3A%2F%2Fimg.shields.io%2Fpypi%2Fdm%2Fgensim%3Fcolor%3Dblue` example

piskvorky:
Follow-up from #2772. Use google's caching proxy for all `img.shields.io` badges in our README.

Fixes #2805.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,15 @@
 gensim – Topic Modelling in Python
 ==================================
 
-[![Build Status](https://travis-ci.org/RaRe-Technologies/gensim.svg?branch=develop)](https://travis-ci.org/RaRe-Technologies/gensim)
-[![GitHub release](https://img.shields.io/github/release/rare-technologies/gensim.svg?maxAge=3600)](https://github.com/RaRe-Technologies/gensim/releases)
-[![Downloads](https://img.shields.io/pypi/dm/gensim?color=blue)](https://pepy.tech/project/gensim/month)
-[![DOI](https://zenodo.org/badge/DOI/10.13140/2.1.2393.1847.svg)](https://doi.org/10.13140/2.1.2393.1847)
-[![Mailing List](https://img.shields.io/badge/-Mailing%20List-blue.svg)](https://groups.google.com/forum/#!forum/gensim)
-[![Follow](https://img.shields.io/twitter/follow/gensim_py.svg?style=social&style=flat&logo=twitter&label=Follow&color=blue)](https://twitter.com/gensim_py)
+[comment]: # (The following image URLs are obfuscated = proxied and cached through Google because of:
+              https://github.com/RaRe-Technologies/gensim/issues/2805 )
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,12 +1,18 @@
...
-[![DOI](https://zenodo.org/badge/DOI/10.13140/2.1.2393.1847.svg)](https://doi.org/10.13140/2.1.2393.1847)
-[![Mailing List](https://img.shields.io/badge/-Mailing%20List-blue.svg)](https://groups.google.com/forum/#!forum/gensim)
-[![Follow](https://img.shields.io/twitter/follow/gensim_py.svg?style=social&style=flat&logo=twitter&label=Follow&color=blue)](https://twitter.com/gensim_py)
+<!--
+The following image URLs are obfuscated = proxied and cached through
+Google because of Github's proxying issues. See:
+https://github.com/RaRe-Technologies/gensim/issues/2805
+-->
+
+[![Build Status](https://images1-focus-opensocial.googleusercontent.com/gadgets/proxy?container=focus&refresh=3600&url=https%3A%2F%2Ftravis-ci.org%2FRaRe-Technologies%2Fgensim.svg%3Fbranch%3Ddevelop)](https://travis-ci.org/RaRe-Technologies/gensim)
```
mpenkov, line 10: Do you reckon it would helpful to have a util script that obfuscates/deobfuscates these badges? It'd be helpful if we need to add more badges or deobfuscate later. It'll also be more obvious what's going on.

piskvorky: This is possible but updating badges is so infrequent, and so easy to do when it does happen = a few minutes, that I don't think scripts are a good ROI. More stuff to maintain, extra cognitive load.

Do you think the comment above is insufficient?

mpenkov: No, I guess it's fine. If it becomes a pain, we can add scripts later.

PR #2809, opened 2020-04-26
Update the doc2vec-wikipedia notebook tutorial

aneeshpanoli:


piskvorky:
What was this about? Can you add some motivation / traceback please?

piskvorky:
Okay, I think I see. I'll try to fix this.

piskvorky:
I checked and that notebook seems severely outdated. For example `Doc2Vec.scale_vocab` does not exist any more – @gojomo any hope of salvaging it? Or shall we just delete this notebook to avoid user confusion?

gojomo:
A working example, *somewhere*, of applying `Doc2Vec` to a Wikipedia dump should remain - that's a very popular & understandable exercise, on a realistically-sizable corpus. Whether it's inside the project bundled files or elsewhere doesn't matter as much as that it's easily findable & kept working. 

WRT to this notebook's clumsy use of `scale_vocab()` to find a good `min_count` to reach a desired vocabulary size - the optional parameter `max_final_vocab` can now specify the user's preference here, & automatically discard just the right number of rarer words.

piskvorky:
@gojomo I updated the notebook in 490676cc34d909b8a361fa1ae1e835263a13673b & cleaned it up a bit. Can you please review?

In particular, I'm worried about that `WARNING : Effective 'alpha' higher than previous training cycles` – where is that coming from?

gojomo:
I will only be able to take a closer look later, but in general, that warning should only show if someone is calling `.train()` more than once on the same model, and with a higher starting `alpha` than the prior .`train()` final `alpha`. That's occasionally justifiable for expert purposes, but quite often unwise or a mistake.

It doesn't look like your notebook would've intentionally done that, in a straightforward top-to-bottom runthrough. Is there a chance that cell was manually run more than once, creating that cached output? I notice the relevant cell *doesn't* have the usual order-of-execution number in its `[ ]` prefix - and the earlier cells that *do* have such order-of-execution numbers jump around a bit. Ideally, I'd sugges the canonical/reviewed/reference form of a notebook, checked-in, should (if it displays cached output from prior runs at all) show the results of an automatic, straight-through run. (And if for some reason the notebook wants to demonstrate the effects of out-of-order, or repeated, execution, that's made really clear in the surrounding text.) I conjecture that if re-run cleanly from the start in order, that same warning won't appear – but if it does, it might be a bug in the code generating the warning (or some other weird cross-state-contamination I'm just not thinking of at the moment).

piskvorky:
Yeah, the notebook is meant for straight top-to-bottom execution, although I jump around a little as I fix and reorganize things.

Re-running is a chore because some of the cells take so long, but I'll do it before the final merge. That's also why I ask for review feedback now, to minimize such re-runs :)

Thanks for reviewing.

piskvorky:
PR superseded by #3307, closing here.

PR #2812, opened 2020-04-28
get tests to pass on Py27

### LINKED ISSUE #2811: Investigate Py2.7 support
[This PR](https://github.com/RaRe-Technologies/gensim/pull/2630) silently removed Py2.7 support as a side-effect. There is no mention of it in the PR description, commit messages or discussion with the reviewers.

I'm the one who made the change, but it was several months ago, and I myself don't remember what the original intention was.

We need Python 2.7 support for the 3.8.3 release, because a significant number of users (not sure how many) still use that version of Python.

- [x] Get gensim building on Py2.7 again. Try working of the current develop branch, if not, consider alternatives.
- [x] Start a new branch to bring back Py2.7 builds in the gensim-wheels repo (see https://github.com/MacPython/gensim-wheels/pull/24).

mpenkov:
Fix #2811

At first  glance, the removal of Py27 from gensim appears to have been mostly superficial. Getting it to build and run on a local Py27 install required minimal changes. I think if we can get CI to pass, and wheels to build, then we can release 3.8.3 off develop HEAD as planned.

@piskvorky @gojomo @menshikh-iv What do you think?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,8 +20,9 @@ import sys
 from setuptools import setup, find_packages, Extension
 from setuptools.command.build_ext import build_ext
 
-if sys.version_info[:2] < (3, 5):
-    raise Exception('This version of gensim needs 3.5 or later.')
+PY2 = sys.version_info[:2] == (2, 7)
+if sys.version_info[:2] < (3, 5) and not PY2:
+    raise Exception('This version of gensim requires Py2.7, or Py3.5 or greater')
 
 c_extensions = {
     'gensim.models.word2vec_inner': 'gensim/models/word2vec_inner.c',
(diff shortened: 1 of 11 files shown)


tox.ini
```diff
@@ -17,7 +17,7 @@ ignore = F821  ; TODO remove me when all examples in docstrings will be executab
 exclude=.venv, .git, .tox, dist, doc, build, gensim/models/deprecated
 
 [pytest]
-addopts = -rfxEXs --durations=20 --showlocals --reruns 3 --reruns-delay 1
```
menshikh-iv, line 20: why?

setup.py
```diff
@@ -20,8 +20,9 @@
 from setuptools import setup, find_packages, Extension
 from setuptools.command.build_ext import build_ext
 
-if sys.version_info[:2] < (3, 5):
-    raise Exception('This version of gensim needs 3.5 or later.')
+PY2 = sys.version_info[:2] == (2, 7)
+if sys.version_info[:2] < (3, 5) and not PY2:
```
menshikh-iv, line 24: https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires, please :)

tox.ini
mpenkov: I couldn't get this to install under Py2.

Why do we need to retry on failure? Do we have any tests where this is a benefit? I've never had a test fail the first time and succeed on subsequent attempts.

setup.py
mpenkov: Should we do that for 3.8.3, or save it for a non-bugfix release?

menshikh-iv: up to you :)

tox.ini
menshikh-iv: We have "blinking" tests, this trying to re-run them to fix

piskvorky:
That's great :)

PR #2814, opened 2020-04-28
[MRG] Update README instructions + clean up testing

### LINKED ISSUE #2788: Do test instructions still work?
In our [README](https://github.com/RaRe-Technologies/gensim#installation), we tell users to run `python setup.py test`.

One user [reported on our mailing list](https://groups.google.com/forum/#!topic/gensim/JfeV2phNTjk) that this doesn't actually work.

We should fix the docs or the tests – whichever is at fault here.

piskvorky:
I updated our instructions around testing and Python versions in the README.

I also updated setup.py, because some of the same instruction info is duplicated there (for PyPI, I presume). While at setup.py, I also cleaned up the testing dependencies for Linux and Windows.

Fixes #2788 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,11 @@ matrix:
     - python: '3.7'
       env:
         - TOXENV="py37-linux"
-        - BOTO_CONFIG="/dev/null"  # XXX: why is this here?
+      # The following two lines used to be necessary because Travis left files lying around in ~/.aws/,
+      # messing up our tests. Now fixed since https://github.com/travis-ci/travis-ci/issues/7940
+        # - BOTO_CONFIG="/dev/null"
+      #sudo: true
       dist: xenial
-      sudo: true  # XXX: why is this here vs. others?
 
     - python: '3.6'
       env: TOXENV="py36-linux"
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,9 @@ matrix:
     - python: '3.7'
       env:
         - TOXENV="py37-linux"
-        - BOTO_CONFIG="/dev/null"
+        - BOTO_CONFIG="/dev/null"  # XXX: why is this here?
       dist: xenial
-      sudo: true
+      sudo: true  # XXX: why is this here vs. others?
 
     - python: '3.6'
       env: TOXENV="py36-linux"
@@ -35,11 +35,11 @@ matrix:
 
 install:
   - pip install tox
-  - sudo apt-get install -y gdb  # install gdb
+  - sudo apt-get install -y gdb
 
 
 before_script:
-  - ulimit -c unlimited -S       # enable core dumps
+  - ulimit -c unlimited -S  # enable core dumps
 
 
 script: tox -vv
@@ -47,5 +47,5 @@ script: tox -vv
 
 after_failure:
   - pwd
-  - COREFILE=$(find . -maxdepth 1 -name "core*" | head -n 1) # find core file
+  - COREFILE=$(find . -maxdepth 1 -name "core*" | head -n 1)
   - if [[ -f "$COREFILE" ]]; then EXECFILE=$(gdb -c "$COREFILE" -batch | grep "Core was generated" | tr -d "\`" | cut -d' ' -f5); file "$COREFILE"; gdb -c "$COREFILE" "$EXECFILE" -x continuous_integration/debug.gdb -batch; fi
--- a/README.md
+++ b/README.md
@@ -69,23 +69,23 @@ NumPy. This is optional, but using an optimized BLAS such as [ATLAS] or
 magnitude. On OS X, NumPy picks up the BLAS that comes with it
 automatically, so you don’t need to do anything special.
 
-The simple way to install gensim is:
+Install the latest version of gensim:
 
-    pip install -U gensim
+```bash
+    pip install --upgrade gensim
+```
 
 Or, if you have instead downloaded and unzipped the [source tar.gz]
-package, you’d run:
+package:
 
-    python setup.py test
+```bash
     python setup.py install
+```
 
-For alternative modes of installation (without root privileges,
-development installation, optional install features), see the
-[documentation].
+For alternative modes of installation, see the [documentation].
 
-This version has been tested under Python 2.7, 3.5 and 3.6. Gensim’s github repo is hooked
-against [Travis CI for automated testing] on every commit push and pull
-request. Support for Python 2.6, 3.3 and 3.4 was dropped in gensim 1.0.0. Install gensim 0.13.4 if you *must* use Python 2.6, 3.3 or 3.4. Support for Python 2.5 was dropped in gensim 0.10.0; install gensim 0.9.1 if you *must* use Python 2.5). 
+Gensim is being `continuously tested <https://travis-ci.org/RaRe-Technologies/gensim>`_ under Python 3.5, 3.6, 3.7 and 3.8.
+Support for Python 2.7 was dropped in gensim 4.0.0 – install gensim 3.8.3 if you must use Python 2.7.
 
 How come gensim is so fast and memory efficient? Isn’t it pure Python, and isn’t Python slow and greedy?
 --------------------------------------------------------------------------------------------------------
@@ -113,14 +113,14 @@ Documentation
   [Tutorials]: https://radimrehurek.com/gensim/auto_examples/
   [Official Documentation and Walkthrough]: http://radimrehurek.com/gensim/
   [Official API Documentation]: http://radimrehurek.com/gensim/apiref.html
-  
+
 ---------
 
 Adopters
 --------
 
 | Company | Logo | Industry | Use of Gensim |
-|---------|------|----------|---------------|                          
+|---------|------|----------|---------------|
 | [RARE Technologies](http://rare-technologies.com) | ![rare](docs/src/readme_images/rare.png) | ML & NLP consulting | Creators of Gensim – this is us! |
 | [Amazon](http://www.amazon.com/) |  ![amazon](docs/src/readme_images/amazon.png) | Retail |  Document similarity. |
 | [National Institutes of Health](https://github.com/NIHOPA/pipeline_word2vec) | ![nih](docs/src/readme_images/nih.png) | Health | Processing grants and publications with word2vec. |
@@ -169,8 +169,8 @@ BibTeX entry:
   [Talentpair]: https://avatars3.githubusercontent.com/u/8418395?v=3&s=100
   [citing gensim in academic papers and theses]: https://scholar.google.cz/citations?view_op=view_citation&hl=en&user=9vG_kV0AAAAJ&citation_for_view=9vG_kV0AAAAJ:u-x6o8ySG0sC
 
-  
-  
+
+
   [documentation and Jupyter Notebook tutorials]: https://github.com/RaRe-Technologies/gensim/#documentation
   [Vector Space Model]: http://en.wikipedia.org/wiki/Vector_space_model
   [unsupervised document analysis]: http://en.wikipedia.org/wiki/Latent_semantic_indexing
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ package:
 
 For alternative modes of installation, see the [documentation].
 
-Gensim is being [continuously tested](https://travis-ci.org/RaRe-Technologies/gensim) under Python 3.5, 3.6, 3.7 and 3.8.
+Gensim is being [continuously tested](https://travis-ci.org/RaRe-Technologies/gensim) under Python 3.6, 3.7 and 3.8.
 Support for Python 2.7 was dropped in gensim 4.0.0 – install gensim 3.8.3 if you must use Python 2.7.
 
 How come gensim is so fast and memory efficient? Isn’t it pure Python, and isn’t Python slow and greedy?
--- a/README.md
+++ b/README.md
@@ -49,13 +49,6 @@ If this feature list left you scratching your head, you can first read
 more about the [Vector Space Model] and [unsupervised document analysis]
 on Wikipedia.
 
-Support
-------------
-
-Ask open-ended or research questions on the [Gensim Mailing List](https://groups.google.com/forum/#!forum/gensim).
-
-Raise bugs on [Github](https://github.com/RaRe-Technologies/gensim/blob/develop/CONTRIBUTING.md) but **make sure you follow the [issue template](https://github.com/RaRe-Technologies/gensim/blob/develop/ISSUE_TEMPLATE.md)**. Issues that are not bugs or fail to follow the issue template will be closed without inspection.
-
 Installation
 ------------
 
@@ -84,7 +77,7 @@ package:
 
 For alternative modes of installation, see the [documentation].
 
-Gensim is being `continuously tested <https://travis-ci.org/RaRe-Technologies/gensim>`_ under Python 3.5, 3.6, 3.7 and 3.8.
+Gensim is being [continuously tested](https://travis-ci.org/RaRe-Technologies/gensim) under Python 3.5, 3.6, 3.7 and 3.8.
 Support for Python 2.7 was dropped in gensim 4.0.0 – install gensim 3.8.3 if you must use Python 2.7.
 
 How come gensim is so fast and memory efficient? Isn’t it pure Python, and isn’t Python slow and greedy?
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -262,7 +261,7 @@ def run(self):
     'pytest-rerunfailures',
     'mock',
     'cython',
-    # temporarily remove pyemd to work around appveyor issues
+    # Temporarily remove pyemd to work around appveyor issues. XXX done?
```
piskvorky, line 264: @mpenkov @menshikh-iv I'd appreciate comments on the three `XXX` in this PR. Why is that code there? Is it still needed?

I'll add your comments directly to the source.

menshikh-iv: If I remember correctly, `pyemd` has broken installation on windows

menshikh-iv: About `nmslib`, ping @masa3141 https://github.com/RaRe-Technologies/gensim/pull/2417/commits/acff16674a5d1bbebfb37149a4d0dfd84333d8cd, can you explain please this one
```python
if (3, 0) < sys.version_info < (3, 7):
    linux_testenv.extend(['nmslib'])
```
why this needed?

menshikh-iv: About `tensorflow/keras`: this was done in https://github.com/RaRe-Technologies/gensim/pull/2211 and needed to avoid some testing issues (not sure how relevant this after 2 years)

piskvorky:
@mpenkov I'm cleaning up, still work-in-progress.

But why is Python 3.5 missing from Travis? Is that on purpose?

piskvorky:
It appears Travis can test (and build wheels for) Linux, OSX and Windows. Do we really need so many services (AppVeyor, MacPython…)?

.travis.yml
```diff
@@ -25,27 +25,27 @@ matrix:
     - python: '3.7'
       env:
         - TOXENV="py37-linux"
-        - BOTO_CONFIG="/dev/null"
+        - BOTO_CONFIG="/dev/null"  # XXX: why is this here?
```
piskvorky, line 28: @menshikh-iv can you comment here as well?

.travis.yml
```diff
@@ -25,27 +25,27 @@ matrix:
...
       dist: xenial
-      sudo: true
+      sudo: true  # XXX: why is this here vs. others?
```
piskvorky, line 30: @menshikh-iv can you comment here as well?

piskvorky:
@mpenkov Done here except `test_doc2vec.py` segfaults on py3.6. It passes on other Pythons.

Screen Shot 2020-04-29 at 23 32 12

This PR made no changes to doc2vec, so not sure what that is about. @gojomo Any idea, have you seen that before?

The only piece in [Travis logs](https://travis-ci.org/github/RaRe-Technologies/gensim/jobs/681203489) that seems vaguely relevant is this:

```
gensim/models/doc2vec_corpusfile.cpp: In function ‘PyObject* _ZL72__pyx_pf_6gensim_6models_18doc2vec_corpusfile_4d2v_train_epoch_dm_concatP7_objectS0_S0_S0_S0_S0_S0_S0_S0_S0_S0_S0_S0_S0_S0_S0_S0_S0_S0_.isra.64(PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*)’:
gensim/models/doc2vec_corpusfile.cpp:5757:168: warning: ‘__pyx_v_m’ may be used uninitialized in this function [-Wmaybe-uninitialized]
               __pyx_v_6gensim_6models_14word2vec_inner_our_saxpy((&__pyx_v_c.vector_size), (&(__pyx_v_c.doctag_locks[__pyx_v__doc_tag])), (&(__pyx_v_c.work[(__pyx_v_m * __pyx_v_c.vector_size)])), (&__pyx_v_6gensim_6models_18doc2vec_corpusfile_ONE), (&(__pyx_v_c.doctag_vectors[(__pyx_v__doc_tag * __pyx_v_c.vector_size)])), (&__pyx_v_6gensim_6models_18doc2vec_corpusfile_ONE));
```

.travis.yml
mpenkov: I think this is one of my creations, this ensures that boto3 does _not_ read any config files lying around in `~/.aws/`.

mpenkov:
> But why is Python 3.5 missing from Travis and AppVeyor in `develop`? Is that on purpose?

Yes, we removed them in https://github.com/RaRe-Technologies/gensim/pull/2715

mpenkov:
> It appears Travis can test (and build wheels for) Linux, OSX and Windows. Do we really need the additional services, like AppVeyor or MacPython?
> 
> They complicate and slow down our testing process considerably.

I think there are two aspects here, each with their own positives and negatives.

1. Do we need the MacPython/gensim-wheels repository to build our own wheels?
2. Do we need AppVeyor?

The gensim-wheels repository owned by the MacPython org gives us additional quota for building our wheels. In theory, we could manage this repository ourselves (or even move it into the gensim repository, in theory), but then we'd have less CI quota. I don't know _how_ much less, perhaps @menshikh-iv can chip in here.

AppVeyor is optional, in theory. We could invest some time and effort and experiment move our Windows builds to Travis CI. I think it's worth gauging our gains and losses, here, though.

Yes, we'd simplify our builds, and that's a plus. Everything would happen in travis.yml, and we wouldn't need appveyor.yml at all.

On the minus side, we'd increase our Travis footprint. There's a good chance that builds would actually take _longer_, because we'd be doing _everything_ on Travis, where we have a fixed quota. Currently, the Travis and AppVeyor builds run in parallel, independent of each other.

If you really want to go down this path, then I suggest a gradual change:

1. Move gensim Windows builds from Appveyor to Travis
2. Move gensim-wheels Windows builds from Appveyor to Travis
3. Move gensim-wheels repo to RaRe-Technologies control

but personally, I don't think we should do this, at least not while we have other, more important things  to do.

.travis.yml
menshikh-iv: Related to https://github.com/travis-ci/travis-ci/issues/7940, can be removed now I guess

.travis.yml
menshikh-iv: same as https://github.com/RaRe-Technologies/gensim/pull/2814#discussion_r417767513

menshikh-iv:
> It appears Travis can test (and build wheels for) Linux, OSX and Windows. Do we really need the additional services, like AppVeyor or MacPython?

I'm surprised with Travis+Windows, earlier Travis didn't support that.
I guess it's possible to move `AppVeyor` -> `Travis` in `gensim` repository, but I see no reasons (our existing approach well-studied and already works, what's a reason to change something here?)

We have separate `gensim-wheels` to avoid wheel building for each push because
- this takes much longer than CI in `gensim`
- this repository has a different "testing scheme" (in `gensim-wheels` we testing exactly compiled wheels for each platform, this isn't the same as tests here)

> On the minus side, we'd increase our Travis footprint. There's a good chance that builds would actually take longer, because we'd be doing everything on Travis, where we have a fixed quota. Currently, the Travis and AppVeyor builds run in parallel, independent of each other.

That's true. In that case, if we want to speedup current build, better move to
- non-travis 
- platform that allows us concurrent builds (`appveyor` doesn't), like `azure` or `circle-ci`


BTW generally, for speedup testing, the best solution is speedup our tests (not changing some CI services).

README.md
```diff
@@ -49,13 +49,6 @@ If this feature list left you scratching your head, you can first read
 more about the [Vector Space Model] and [unsupervised document analysis]
 on Wikipedia.
 
-Support
```
menshikh-iv, line 52: Why? Where users should place their questions?

README.md
```diff
@@ -69,23 +62,23 @@ NumPy. This is optional, but using an optimized BLAS such as [ATLAS] or
...
 
-For alternative modes of installation (without root privileges,
-development installation, optional install features), see the
-[documentation].
+For alternative modes of installation, see the [documentation].
 
-This version has been tested under Python 2.7, 3.5 and 3.6. Gensim’s github repo is hooked
-against [Travis CI for automated testing] on every commit push and pull
-request. Support for Python 2.6, 3.3 and 3.4 was dropped in gensim 1.0.0. Install gensim 0.13.4 if you *must* use Python 2.6, 3.3 or 3.4. Support for Python 2.5 was dropped in gensim 0.10.0; install gensim 0.9.1 if you *must* use Python 2.5). 
+Gensim is being [continuously tested](https://travis-ci.org/RaRe-Technologies/gensim) under Python 3.5, 3.6, 3.7 and 3.8.
```
menshikh-iv, line 80: `+ appveyor here`, `+ gensim-wheels here`

setup.py
```diff
@@ -262,18 +261,20 @@ def run(self):
     'pytest-rerunfailures',
     'mock',
     'cython',
-    # temporarily remove pyemd to work around appveyor issues
-    # 'pyemd',
+    'nmslib',
```
menshikh-iv, line 264: in case if you change dependencies here, you should "mirror" this changes in `gensim-wheels`

setup.py
```diff
@@ -21,7 +21,7 @@
 from setuptools.command.build_ext import build_ext
 
 if sys.version_info[:2] < (3, 5):
-    raise Exception('This version of gensim needs 3.5 or later.')
+    raise Exception('This version of gensim needs Python 3.5 or later.')
```
menshikh-iv, line 24: please use `setup(...., python_requires='>=3.5', ...)`, instead of raising exceptions in `setup.py`

setup.py
piskvorky: Why? How is this related to wheels?

README.md
piskvorky: Section still there, I just moved it.

README.md
piskvorky: Note to self: remove 3.5, we don't actually test that. https://github.com/RaRe-Technologies/gensim/pull/2814#issuecomment-621562196

setup.py
piskvorky: Yeah, I can change that too, while at it.

setup.py
menshikh-iv: though testing of course, if you change test deps here, you should use "almost same" in wheel testing too

setup.py
mpenkov: Maybe just a little bit :)

piskvorky: Actually, since we already dropped testing of py3.5, should I make this `>=3.6`?

setup.py
piskvorky: I don't know how that works – can I leave `gensim-wheels` with you?

piskvorky:
@menshikh-iv @mpenkov I'll try poking around the py3.6 segfault a little, but otherwise ready to merge. Anything else from you (esp. for release 3.8.3)?

> BTW generally, for speedup testing, the best solution is speedup our tests (not changing some CI services).

What I observed was ~16 minutes for Travis (all tests), and ~50 minutes for AppVeyor (only 2 tests). So the greatest speed up would be dropping AppVeyor.

Overall I saw a lot of hard-won but under-documented complexity in the testing pipeline. Being more verbose / explicit in these files, documenting the workaround and gotchas, should help us significantly with future maintenance.

menshikh-iv:
@piskvorky setup https://github.com/marketplace/azure-pipelines for gensim repository, I'll look into on weekend (looks like the best replacement to appveyour according to the experience of sklearn/pandas projects)

piskvorky:
Alright. @mpenkov clean up complete – feel to merge this at will.

menshikh-iv:
@piskvorky basic config is ready https://github.com/menshikh-iv/gensim/pull/2, waiting when you activate `azure` for gensim (please make sure than pipeline runs in separate PR, after that I'll continue)

gojomo:
> @mpenkov Mostly done with the cleanup now, except `test_doc2vec.py` segfaults on Travis py3.6. It passes on other Pythons.
> 
> Screen Shot 2020-04-29 at 23 32 12
> 
> This PR made no changes to doc2vec, so not sure what that is about. @gojomo Any idea, have you seen that before?
> 
> The only piece in [Travis logs](https://travis-ci.org/github/RaRe-Technologies/gensim/jobs/681203489) that seems vaguely relevant:
> 
> > gensim/models/doc2vec_corpusfile.cpp:5757:168: warning: ‘__pyx_v_m’ may be used uninitialized in this function [-Wmaybe-uninitialized]
> > __pyx_v_6gensim_6models_14word2vec_inner_our_saxpy((&__pyx_v_c.vector_size), (&(__pyx_v_c.doctag_locks[__pyx_v__doc_tag])), (&(__pyx_v_c.work[(__pyx_v_m * __pyx_v_c.vector_size)])), (&__pyx_v_6gensim_6models_18doc2vec_corpusfile_ONE), (&(__pyx_v_c.doctag_vectors[(__pyx_v__doc_tag * __pyx_v_c.vector_size)])), (&__pyx_v_6gensim_6models_18doc2vec_corpusfile_ONE));
> 
> (but the warning's also there in py3.7 and py3.8, which both pass, so likely a red herring)

I've wrestled with core dumps before, including some frustrating ones that only happened on Travis – and that's why there's some new post-failure `gdb`-based output in the logs: see the folded output under <https://travis-ci.org/github/RaRe-Technologies/gensim/jobs/681203489#L996>. 

But, those weren't around this file, and didn't have this particular (`double free or corruption (!prev)`) error. And, I'm not seeing anything in this PR that looks fishy for memory-corruption issues. But there do seem some builds, earlier in this PR, that succeeded – so I'd first check (1) is this a reliable or intermittent crash, when rerun with the exact same code?; (2) does undoing some of the innocent-looking things that were in the commits that immediately preceded crashes (like say the `pyemd` import) have any effect?

piskvorky:
@gojomo yes, the crashes are reliable. I narrowed it down to having tensorflow installed:

* tensorflow installed => Travis py3.6 crashes
* tensorflow not installed => Travis all tests pass, including 3.6

So my "fix" was https://github.com/RaRe-Technologies/gensim/pull/2814/commits/8cd68b2a5ea537f2953bdd592b1a1973b148431d but it's really a crude hack. Still a mystery to me what's happening, there may be a deeper issue lurking there.

setup.py
piskvorky: LOL I guess. Is the process documented somewhere?

And most importantly: is it worth it? How many actual bugs did that catch, versus how much upkeep & maintenance effort.

mpenkov: Created #2816 to deal with the issue outside of this PR.

menshikh-iv: >  Is the process documented somewhere? 

https://github.com/matthew-brett/multibuild

> is it worth it? 

Generally yes, because of this catch compilation issues / wired behavior on all versions of all OS (what we don't do in current repo)

> How many actual bugs did that catch over the years, versus how much upkeep & maintenance effort.

Depends on "actual bug" definition, maintenance effort correlated with `gensim` changes (like new test dependencies, new cython/c++ deps, new python version, etc). It burned a lot of time, that's true. I want to believe it's not in vain.

piskvorky: What's an example of a bug this has caught? First / most important that comes to mind.

menshikh-iv: Compilation issues on OSX & Windows (esp. windows + py27). This appears multiple times in cases when we add new python version support or new `pyx`/`c`/`cpp`

piskvorky: I'd expect compilation issues to be caught by CI.

Missing files in wheels could be a problem.

@mpenkov I'll look into that `multibuild` thing (or maybe the simpler `cibuildwheel` linked above) eventually, but not blocking for us now.

I'll aim to drop and simplify as much as possible. If the complexity – combined with the severe lack of documentation – is such that it leads to contributor frustration ("constantly repairing your wheels, to be honest, I'm sick of it") then that's not sustainable for a volunteer open source project.

menshikh-iv: > I'd expect compilation issues to be caught by CI.

Not really, because
- fully missing OSX
- missing some of win+py combinations

In wheel building, we checked almost all stuff

menshikh-iv: btw still waiting https://github.com/RaRe-Technologies/gensim/pull/2814#issuecomment-621950805

menshikh-iv:
Ping @piskvorky, I'm still waiting for https://github.com/RaRe-Technologies/gensim/pull/2814#issuecomment-621950805 or give me same permissions as for `gensim-wheels` and I'll set up that myself.

piskvorky:
Yes, still on my TODO list to check what that's about. Hopefully today.

PR #2830, opened 2020-05-07
Fixed KeyError in coherence model

### LINKED ISSUE #2711: Coherence score on new data Key Error
I want to compare different models (LDA, Mallet, etc.) with a Cross Validation. I train the model with training data and want to calculate the coherence score (c_v) with the test data. I do something like this:

```python
    dictionary = gensim.corpora.Dictionary(test)
    corpus = [dictionary.doc2bow(text) for text in test]

    cm = CoherenceModel(topics=topics, 
                        corpus=corpus, 
                        texts=test,
                        dictionary=dictionary, 
                        coherence="c_v")
    
    cm.get_coherence()
```

When a word in a topic found on the training data is not present in the test data this raises an key error in the dictionary at some point. Does someone know something about this? is this a bug or how do i solve this issue?

grafik

### LINKED ISSUE #3111: Coherence key error on held out set
Hi, I am trying to train a LDA model on a train set (80% of the data):

```
lda_model = gensim.models.ldamulticore.LdaMulticore(corpus=corpus_train,
                                                                id2word=id2word_train,
                                                                 num_topics=n_topics, 
                                                                 random_state=100,
                                                                 chunksize=200,
                                                                 workers=20,
                                                                 iterations=100)
```

And get the coherence on a test set (20% held out test):

```
cv_coherence_model_lda = CoherenceModel(model=lda_model, texts=texts_test, dictionary=id2word_test, coherence='c_v')
cv_coherence_total = cv_coherence_model_lda.get_coherence()
```

I run this for n_topics from 1 to 30. 
However, it keeps crashing on random n_topics, with the error outlined below. The Key Error number changes all the time. Any help would be appreciated.

```
<ipython-input-8-8b4757057a4f> in find_optimal_topics_gensim(data, splits, test_size)
     87             #coherence models
     88             cv_coherence_model_lda = CoherenceModel(model=lda_model, texts=texts_test, dictionary=id2word_test, coherence='c_v')
---> 89             cv_coherence_total = cv_coherence_model_lda.get_coherence()
     90 

/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gensim/models/coherencemodel.py in get_coherence(self)
    610 
    611         """
--> 612         confirmed_measures = self.get_coherence_per_topic()
    613         return self.aggregate_measures(confirmed_measures)
    614 

/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gensim/models/coherencemodel.py in get_coherence_per_topic(self, segmented_topics, with_std, with_support)
    570             segmented_topics = measure.seg(self.topics)
    571         if self._accumulator is None:
--> 572             self.estimate_probabilities(segmented_topics)
    573 
    574         kwargs = dict(with_std=with_std, with_support=with_support)

/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gensim/models/coherencemodel.py in estimate_probabilities(self, segmented_topics)
    542                 kwargs['model'] = self.keyed_vectors
    543 
--> 544             self._accumulator = self.measure.prob(**kwargs)
    545 
    546         return self._accumulator

/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gensim/topic_coherence/probability_estimation.py in p_boolean_sliding_window(texts, segmented_topics, dictionary, window_size, processes)
    152         accumulator = WordOccurrenceAccumulator(top_ids, dictionary)
    153     else:
--> 154         accumulator = ParallelWordOccurrenceAccumulator(processes, top_ids, dictionary)
    155     logger.info("using %s to estimate probabilities from sliding windows", accumulator)
    156     return accumulator.accumulate(texts, window_size)

/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gensim/topic_coherence/text_analysis.py in __init__(self, processes, *args, **kwargs)
    429 
    430     def __init__(self, processes, *args, **kwargs):
--> 431         super(ParallelWordOccurrenceAccumulator, self).__init__(*args)
    432         if processes < 2:
    433             raise ValueError(

/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gensim/topic_coherence/text_analysis.py in __init__(self, relevant_ids, dictionary)
    285 
    286         """
--> 287         super(WindowedTextsAnalyzer, self).__init__(relevant_ids, dictionary)
    288         self._none_token = self._vocab_size  # see _iter_texts for use of none token
    289 

/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gensim/topic_coherence/text_analysis.py in __init__(self, relevant_ids, dictionary)
    188         """
    189         super(UsesDictionary, self).__init__(relevant_ids)
--> 190         self.relevant_words = _ids_to_words(self.relevant_ids, dictionary)
    191         self.dictionary = dictionary
    192         self.token2id = dictionary.token2id

/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gensim/topic_coherence/text_analysis.py in _ids_to_words(ids, dictionary)
     59     top_words = set()
     60     for word_id in ids:
---> 61         word = dictionary.id2token[word_id]
     62         if isinstance(word, set):
     63             top_words = top_words.union(word)

KeyError: 1094
```

Versions used

#### Versions
```

Darwin-20.3.0-x86_64-i386-64bit
Python 3.7.9 (v3.7.9:13c94747c7, Aug 15 2020, 01:31:08) 
[Clang 6.0 (clang-600.0.57)]
Bits 64
NumPy 1.19.1
SciPy 1.5.2
gensim 4.0.1
FAST_VERSION 0
```

pietrotrope:
When the token list of a topic contains a token which is not present in the dictionary token2id the framework raised an error instead of excluding the token and continue to compute the metric score.

Excluding the tokens of the topics which are not present in the dictionary allows to compute the metric on different texts that do not have to contain every representative word of the topics.

- Fix #2711
- Fix #3111


piskvorky:
Thanks. Can you add a test too?

Are there other places in the coherence with the same issue? (poor input validation / preprocessing) Let's fix them all at once.

pietrotrope:
I ran some test and i fail when passing topics id's instead of tokens.
I need to solve this issue, after that we can proceed to search other places with the same issue

pietrotrope:
The idea behind the issue is to allow valuation of topics with new tokens and/or ID's that are not present in the texts by excluding them from the computation.

To explain i will refer to the issue example with a simple test.

Given:
```
topics = [  ['human', 'computer', 'system', 'interface'],
            ['graph', 'minors', 'trees', 'eps']]

test_WITH_human = [  ['human', 'computer', 'system', 'interface'],
                     ['graph', 'minors', 'trees', 'eps']]

test_WITHOUT_human = [  [ 'computer', 'system', 'interface'],
                        ['graph', 'minors', 'trees', 'eps']]
```

The following code
```
dictionary = gensim.corpora.Dictionary(test_WITHOUT_human)
corpus = [dictionary.doc2bow(text) for text in test_WITHOUT_human]

cm = CoherenceModel(topics=topics, 
                    corpus=corpus, 
                    texts=test_WITHOUT_human,
                    dictionary=dictionary, 
                    coherence="c_npmi")

cm.get_coherence()
```

Returns
```

--> 445             return np.array([self.dictionary.token2id[token] for token in topic])
    446         except KeyError:  # might be a list of token ids already, but let's verify all in dict

KeyError: 'human'
```
Instead of:
`1.000000000005771`

which is the metric score.


With the last update i get the right result from this test, but it raises some other issues due to the fact that in the initial version, token or id formatting was retrieved by handling an exception.

Original code:
```
def _ensure_elements_are_ids(self, topic):
        try:
            return np.array([self.dictionary.token2id[token] for token in topic])
        except KeyError:  # might be a list of token ids already, but let's verify all in dict
            topic = (self.dictionary.id2token[_id] for _id in topic)
            return np.array([self.dictionary.token2id[token] for token in topic])
```
To find out if topics are in token or ID form, this code consider an exception as an hint that topics lists are expressed in ID form and proceed to retrieve them from their ID's

But if we have a topic with a token not present in the texts the code raise the exception and proceed to compute topics as if they're in ID form.
Topics are not in ID form and we got a runtime error.

Let's try to add a control to exclude tokens which are not present in the texts:

```
def _ensure_elements_are_ids(self, topic):
        try:
            return np.array([self.dictionary.token2id[token] for token in topic if token in self.dictionary.token2id])
        except KeyError:  # might be a list of token ids already, but let's verify all in dict
            topic = (self.dictionary.id2token[_id] for _id in topic if _id in self.dictionary.id2token)
            return np.array([self.dictionary.token2id[token] for token in topic])
```

The test in the example with tokens is now covered, but if the input is made of ID's and not tokens, the code will not raise an exception because every id is not a key in the token2id dictionary and it will be excluded, resulting in a list of 0 topics instead of an excpetion which trigger the right code.

To handle the problem i proceeded to try this code:

 ```
def _ensure_elements_are_ids(self, topic):
            elements_are_tokens = np.array([self.dictionary.token2id[token] for token in topic if token in self.dictionary.token2id])
            topic_tokens_from_id = (self.dictionary.id2token[_id] for _id in topic if _id in self.dictionary.id2token)
            elements_are_ids = np.array([self.dictionary.token2id[token] for token in topic_tokens_from_id])
            if elements_are_tokens.size > elements_are_ids.size:
                return elements_are_tokens
            elif elements_are_ids.size > elements_are_tokens.size:
                return elements_are_ids
            else:
                raise Exception("Topic list is not a list of lists of tokens or ids")
```

With this code, we assume that the right formatting is the one with the higher number of matches with the token2id or id2token dictionaries.
When we can't say if the right formatting is tokens lists or ID's lists we raise an exception.
This example seems to pass more tests but it makes an assumption on the data formatting instead of finding out the real one.

If anyone can help me to solve the problem i will gladly try to work on it again and solve the problem once and for all.

File: gensim/models/coherencemodel.py [hunk 1]
```diff
@@ -441,11 +442,16 @@ def topics(self, topics):
         self._topics = new_topics
 
     def _ensure_elements_are_ids(self, topic):
-        try:
-            return np.array([self.dictionary.token2id[token] for token in topic])
-        except KeyError:  # might be a list of token ids already, but let's verify all in dict
-            topic = (self.dictionary.id2token[_id] for _id in topic)
-            return np.array([self.dictionary.token2id[token] for token in topic])
+        elements_are_tokens = np.array([self.dictionary.token2id[token]
```
mpenkov, line 445: Hanging indent please. Alternatively, break the list comprehension up to be more readable:

```python
tokens = [t for t in topic if token in self.dictionary.token2id]
elements_are_tokens = np.array([self.dictionary.token2id[t] for t in tokens])
...
```

gensim/models/coherencemodel.py
```diff
@@ -120,6 +120,7 @@ class CoherenceModel(interfaces.TransformationABC):
         >>> coherence = cm.get_coherence()  # get coherence value
 
     """
+
```
mpenkov, line 123: Why did you add this blank line?

mpenkov:
> To explain i will refer to the issue example with a simple test.

I think this would make a good unit test. It would demonstrate the usefulness of your fix and prevent future regressions.

piskvorky:
Looks like we dropped the ball on this PR. People still keep tripping over the same issue.

@pietrotrope @mpenkov do you think we could finish this & merge? Thanks.

pietrotrope:
hi, I tried to update the code by modifying it according to the requests.
I point out that the solution that I have proposed in recent months (and that I am trying to submit with this commit) excludes from the calculation of the coherence the words that are not present in the used dictionary.

piskvorky:
Thanks @pietrotrope .

>  I point out that the solution that I have proposed in recent months (and that I am trying to submit with this commit) excludes from the calculation of the coherence the words that are not present in the used dictionary.

That's fine – better than an exception at any rate. Is this fact clearly documented?

@mpenkov can you please review & merge?

T0admomo:
pietrotrope - just wanted to say thank you for taking the time to propose this fix! New to gensim and I quickly encountered this problem, this fix will cut down on a lot of preprocessing! following closely !

gensim/models/coherencemodel.py
```diff
@@ -441,12 +441,16 @@ def topics(self, topics):
         self._topics = new_topics
 
     def _ensure_elements_are_ids(self, topic):
-        try:
-            return np.array([self.dictionary.token2id[token] for token in topic])
-        except KeyError:  # might be a list of token ids already, but let's verify all in dict
-            topic = (self.dictionary.id2token[_id] for _id in topic)
-            return np.array([self.dictionary.token2id[token] for token in topic])
-
+        tokens = [t for t in topic if t in self.dictionary.token2id]
```
mpenkov, line 444: I think the naming here is very confusing, and it's possible to do better.

```python
def _ensure_elements_are_ids(self, topic):
  topic_is_tokens = [self.dictionary.token2id[t] for t in topic if t in self.dictionary.token2id]
  topic_is_ids = [i for i in topic if i in self.dictionary.id2token]
  if len(topic_is_tokens) > len(topic_is_ids):
    return np.array(topic_is_tokens)
  elif len(topic_is_ids) > len(topic_is_tokens):
    return np.array(topic_is_ids)
  else:
    raise ValueError('unable to interpret token as either a list of tokens or a list of ids')
```

gensim/test/test_coherencemodel.py
```diff
@@ -44,6 +44,11 @@ def setUp(self):
             ['user', 'graph', 'minors', 'system'],
             ['time', 'graph', 'survey', 'minors']
         ]
+        self.topics3 = [
+            ['human', 'computer', 'system', 'interface'],
+            ['graph', 'minors', 'trees', 'eps']
+        ]
```
mpenkov, line 50: I think you need at least 2 more tests to cover your functionality:

- Topics are IDs
- Topics are neither IDs nor tokens (check that ValueError is raised)

mpenkov (changes requested):
Left some comments, please have a look. Sorry it took me so long to review this.

gensim/models/coherencemodel.py
pietrotrope: Solved also this

gensim/test/test_coherencemodel.py
pietrotrope: And added the 2 new tests (also updated the one using self.topics3 because I forget to change at least one token with respect to self.topics1 in the last update

pietrotrope:
Hi! I updated the code,
as i said in the comments in the revision i did some refactoring (following your suggestions) and added the new tests (also corrected one)

gensim/test/test_coherencemodel.py
mpenkov: Is there a reason why this isn't a separate test (or set of tests)? What's the benefit of keeping everything in a single test method?

mpenkov:
Can you merge origin/develop branch into your PR's branch? Looks like there's some kind of conflict.

mpenkov:
Note to self: resolve conflicts, check tests.

mpenkov:
Merged. Thank you @pietrotrope !

PR #2836, opened 2020-05-11
delete .gitattributes

gojomo:
Per concern at RaRe-Technologies#2819 (comment) , detailed at https://stackoverflow.com/questions/61720528/simple-git-clone-mangles-many-files-with-phantom-un-resettable-modified-state?noredirect=1#comment109174230_61720528 , this new git config setting is causing problems on Linux & MacOS, at least.


menshikh-iv:
Hm, so, you can try to specify exact filenames that makes a problem on azure, like `gensim\\test\\test_data\\testcorpus.svmlight` and so on based on pytest output https://dev.azure.com/rare-technologies/f95569a5-add4-42fb-a6a2-f248676d3639/_apis/build/builds/42/logs/24

Or this way also contains same issue?

gojomo:
> Hm, so, you can try to specify exact filenames that makes a problem on azure, like `gensim\\test\\test_data\\testcorpus.svmlight` and so on based on pytest output https://dev.azure.com/rare-technologies/f95569a5-add4-42fb-a6a2-f248676d3639/_apis/build/builds/42/logs/24
> 
> Or this way also contains same issue?

I'm not sure what you mean. I don't have access to a Windows/Azure setup, and the git errors I'm seeing have no resemblance to those Azure path-name and IO-related errors. 

Rather, my working-directory git believes many files are 'modified'. (Further oddness: yesterday, it was about 145 files via a git-protocol clone, & I thought just as many via an https clone. But today, it looks like 159 via a git-protocol clone, and only a dozen or so via an https clone.) They can't be `reset`, but actually working with these 'changed' files **does** result in changes if committed back to a repo -- see #2837 for an example. 

Try it on Linux or MacOS - it should take < 60 seconds to reproduce with a fresh, totally-separate git clone.

 I'm not sure if the working files themselves are binary-different from the versions in a `pip install`

piskvorky:
Seems critical – if anyone starts a PR right now (with `.gitattributes`), they run a chance of messing up our repo. I just hope they haven't already.

@mpenkov can you please have a look as a priority? This wound is completely self-inflicted. It doesn't affect users (yet) but the longer it stays the bigger the trouble.

mpenkov:
From memory, these changes are a workaround for newline strangeness on Azure. Here's the link:

https://docs.microsoft.com/en-us/azure/devops/pipelines/troubleshooting?view=azure-devops#line-endings-for-multiple-platforms

@gojomo If we simply remove .gitattributes, then we break our Azure pipelines. We need to either:

1. Find another workaround.
2. Go back to Appveyor.

I think (2) is terrible, so I'd prefer (1). @menshikh-iv WDYT?

and include in this PR.

gojomo:
As a non-Windows user, I don't know what the Azure benefits are over Appveyor. (Because of this error, & four giant Azure 'checks' appearing in the Github build-summaries before the builds that are relevant to me, & clicking through on the Azure lines giving less-clear info about progress/etc than I'm used-to, the Azure switchover is 100% time-wasting annoyances so far, from my perspective, but I trust there must be some benefit in cost/reliability/etc if you say so.) And I don't have the resources to debug Azure issues. 

But if this 1-line workaround for Azure causes developer pain on the main contributor OSes (MacOS & Linux), and maybe even Windows (has anyone tried a fresh clone?), and the code/builds worked fine everywhere else including Appveyor beforehand, then this seems like a problem that should be fixed in an Azure-specific way that doesn't negatively impact anywhere/anyone else. Maybe, just the Azure-build-checkout needs to apply an Azure-specific hammer?

mpenkov:
> If nothing else, add a line that inserts this `.gitattributes` there, locally, after clone, and never commits-it-back-in?

I like this idea. @menshikh-iv Thoughts?

menshikh-iv:
I have 2 ideas
- apply `.gitattribute` only for the concrete subset of files that fail azure (replace `*` with paths). This will definitely fix azure, but I'm not sure does this avoid effect with "modified files" that Gordon described
- "re-checkout" repo in azure with generated `.gitattribute` in runtime, this definitely avoid Gordon issue, but not sure will azure work with it (`.gitattributes` works in a pretty strange way for me)

mpenkov:
My plan:

1. Disable the affected tests on Azure
2. Merge this PR, so that regular development can continue without hindrance
3. Re-enable the affected tests in a separate PR (@menshikh-iv can you do this?)

menshikh-iv:
@mpenkov got it, will dig into on weekend

PR #2846, opened 2020-05-21
Fix for Python 3.9/3.10: remove xml.etree.cElementTree

hugovk:
`xml.etree.cElementTree` has been deprecated since Python 3.3 and will be removed in Python 3.9 or Python 3.10 (https://github.com/python/cpython/pull/19921).

Since Python 3.3, `xml.etree.ElementTree` will use a fast implementation whenever available, so there's no need for the `try`/`except ImportError`.

* https://docs.python.org/3/library/xml.etree.elementtree.html

Also update the Trove classifiers to reflect Python 2.7 is no longer supported.


gensim/scripts/segment_wiki.py
```diff
@@ -61,10 +61,7 @@
 import multiprocessing
 import re
 import sys
-try:
-    from xml.etree import cElementTree as ET
-except ImportError:
-    from xml.etree import ElementTree as ET
+from xml.etree import ElementTree as ET
```
piskvorky, line 64: Not related to your PR, but I'm not a fan of such "shortcuts".

Can you please change this to use the full `ElementTree` or `xml.etree. ElementTree`?

hugovk: Do you mean like this?

```python
import xml.etree.ElementTree as ET
...
elem = ET.fromstring(page_xml)
```

Or like this?

```python
import xml.etree.ElementTree
...
elem = ElementTree.fromstring(page_xml)
```

---

I'll note the former is used in the docs and quite common:

```python
import xml.etree.ElementTree as ET
tree = ET.parse('country_data.xml')
root = tree.getroot()
```

But happy to change to whichever you wish.

piskvorky: `from xml.etree import ElementTree`. I don't want our code sprinkled with 2-letter acronym puzzles like `ET`.

They may be an obvious shorthand to the module's author, but not anybody else reading the code.

hugovk: Updated!

Issue #2537, opened 2019-06-25
FastText inconsistent normalization in adjust_vectors method vs word_vec


vackosar:
#### Problem description

There is inconsistency in normalization in ```FastTextKeyedVectors#adjust_vectors``` vs ```WordEmbeddingsKeyedVectors#word_vec```

#### Steps/code/corpus to reproduce

adjust_vectors:
```
            word_vec /= len(ngram_hashes) + 1
```

word_vec method:
```
            return word_vec / len(ngram_hashes)
```

#### Versions

```
Linux-4.18.0-24-generic-x86_64-with-Ubuntu-18.04-bionic
Python 3.7.1 (default, Oct 22 2018, 11:21:55) 
[GCC 8.2.0]
NumPy 1.16.3
SciPy 1.2.1
gensim 3.7.3
FAST_VERSION 1
```

vackosar:
The ```adjustment_vectors``` method adds subword representations to the words themselves, where ```+1``` seems to represent original learned vector of the whole word.

piskvorky:
@vackosar if that's the case, can you add a code comment to the relevant places? It might trip up others too, in the future, so let's be explicit.

mpenkov:
I agree that the two implementations appear to be inconsistent. I think this is a bug.

This is what I rewrote when I worked on fastText a few months ago:

```
b452a5b59 (Michael Penkov       2019-01-12 00:07:40 +0900 2289)             word_vec = np.copy(self.vectors_vocab[v.index])
411f54663 (Michael Penkov       2019-01-24 13:51:40 +0900 2290)             ngram_hashes = ft_ngram_hashes(w, self.min_n, self.max_n, self.bucket, self.compatible_hash)
411f54663 (Michael Penkov       2019-01-24 13:51:40 +0900 2291)             for nh in ngram_hashes:
b1850d948 (Michael Penkov       2019-03-07 15:44:31 +0900 2292)                 word_vec += self.vectors_ngrams[nh]
411f54663 (Michael Penkov       2019-01-24 13:51:40 +0900 2293)             word_vec /= len(ngram_hashes) + 1
b452a5b59 (Michael Penkov       2019-01-12 00:07:40 +0900 2294)             self.vectors[v.index] = word_vec
```

This is the old version of the FastText code:

```
9021ea8b3 (Johannes Baiter      2018-03-01 09:17:43 +0100 2001)             ngrams_found = 0
[...]

PR #2861, opened 2020-06-18
Warning in word2vec when batch_words > MAX_WORDS_IN_BATCH

### LINKED ISSUE #2801: Warning when batch_words > MAX_WORDS_IN_BATCH in word2vec
The doc says

         batch_words : int, optional
            Target size (in words) for batches of examples passed to worker threads (and
            thus cython routines).(Larger batches will be passed if individual
            texts are longer than 10000 words, but the standard cython code truncates to that maximum.)

I think a larger value should not be accepted.

dsandeep0138:
Also added a test case to verify if the warning is thrown when batch_words is more than 10000.

Fixes #2801

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -595,8 +595,9 @@ class Word2Vec(BaseWordEmbeddingsModel):
         self.trainables = Word2VecTrainables(seed=seed, vector_size=size, hashfxn=hashfxn)
 
         if batch_words > MAX_WORDS_IN_BATCH:
-            logger.warning("batch_words should be less than or equal to %d. Standard cython code "
-                           "truncates to the maximum if the individual texts are longer than %d words",
+            logger.warning("batch_words should be less than or equal to %d. "
+                           "Standard cython code truncates to the maximum if "
+                           "the individual texts are longer than %d words",
                            MAX_WORDS_IN_BATCH, MAX_WORDS_IN_BATCH)
 
         super(Word2Vec, self).__init__(
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -595,10 +595,13 @@ class Word2Vec(BaseWordEmbeddingsModel):
         self.trainables = Word2VecTrainables(seed=seed, vector_size=size, hashfxn=hashfxn)
 
         if batch_words > MAX_WORDS_IN_BATCH:
-            logger.warning("batch_words should be less than or equal to %d. "
-                           "Standard cython code truncates to the maximum if "
-                           "the individual texts are longer than %d words",
-                           MAX_WORDS_IN_BATCH, MAX_WORDS_IN_BATCH)
+            logger.warning(
+                "batch_words %d should be less than or equal to %d. "
+                "Cython code will truncate texts that are longer than %d words.",
+                batch_words,
+                MAX_WORDS_IN_BATCH,
+                MAX_WORDS_IN_BATCH
+            )
 
         super(Word2Vec, self).__init__(
             sentences=sentences, corpus_file=corpus_file, workers=workers, vector_size=size, epochs=iter,
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -563,9 +563,9 @@ class Word2Vec(BaseWordEmbeddingsModel):
             If 1, sort the vocabulary by descending frequency before assigning word indexes.
             See :meth:`~gensim.models.word2vec.Word2VecVocab.sort_vocab()`.
         batch_words : int, optional
-            Target size (in words) for batches of examples passed to worker threads (and
-            thus cython routines).(Larger batches will be passed if individual
-            texts are longer than 10000 words, but the standard cython code truncates to that maximum.)
+            Target size (in words) for batches of examples passed to worker threads, and
+            thus cython routines. Larger batches will be passed if individual texts are longer
+            than 10000 words, but the standard cython code truncates to that maximum.
         compute_loss: bool, optional
             If True, computes and stores loss value which can be retrieved using
             :meth:`~gensim.models.word2vec.Word2Vec.get_latest_training_loss`.
@@ -593,6 +593,11 @@ class Word2Vec(BaseWordEmbeddingsModel):
             max_vocab_size=max_vocab_size, min_count=min_count, sample=sample, sorted_vocab=bool(sorted_vocab),
             null_word=null_word, max_final_vocab=max_final_vocab, ns_exponent=ns_exponent)
         self.trainables = Word2VecTrainables(seed=seed, vector_size=size, hashfxn=hashfxn)
+
+        if batch_words > MAX_WORDS_IN_BATCH:
+            logger.warning("batch_words should be less than or equal to %d. Standard cython code "
+                           "truncates to the maximum if the individual texts are longer than %d words",
+                           MAX_WORDS_IN_BATCH, MAX_WORDS_IN_BATCH)
 
         super(Word2Vec, self).__init__(
             sentences=sentences, corpus_file=corpus_file, workers=workers, vector_size=size, epochs=iter,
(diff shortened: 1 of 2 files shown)


gensim/models/word2vec.py
```diff
@@ -594,6 +594,12 @@ def __init__(self, sentences=None, corpus_file=None, size=100, alpha=0.025, wind
             null_word=null_word, max_final_vocab=max_final_vocab, ns_exponent=ns_exponent)
         self.trainables = Word2VecTrainables(seed=seed, vector_size=size, hashfxn=hashfxn)
 
+        if batch_words > MAX_WORDS_IN_BATCH:
+            logger.warning("batch_words should be less than or equal to %d. "
```
piskvorky, line 598: Hanging indent please. We don't use vertical indent in gensim at all.

piskvorky, line 598:
```suggestion
            logger.warning("batch_words %d should be less than or equal to %d. "
```

gensim/models/word2vec.py
```diff
@@ -594,6 +594,12 @@ def __init__(self, sentences=None, corpus_file=None, size=100, alpha=0.025, wind
...
+                           "Standard cython code truncates to the maximum if "
```
piskvorky, line 599:
```suggestion
                           "Cython code will truncate texts that are longer than %s words."
```

gensim/test/test_word2vec.py
```diff
@@ -960,6 +961,13 @@ def testBuildVocabWarning(self, line):
         warning = "Each 'sentences' item should be a list of words (usually unicode strings)."
         self.assertTrue(warning in str(line))
 
+    @log_capture()
+    def testHighBatchWordsWarning(self, line):
+        """Test if warning is raised on using batch_words greater than the maximum allowed value"""
+        word2vec.Word2Vec(batch_words=20000)
+        warning = "batch_words should be less than or equal to " + str(MAX_WORDS_IN_BATCH)
```
piskvorky, line 968: Change to match suggestion above.

gensim/models/word2vec.py
```diff
@@ -595,10 +595,13 @@ def __init__(self, sentences=None, corpus_file=None, size=100, alpha=0.025, wind
...
 
         if batch_words > MAX_WORDS_IN_BATCH:
-            logger.warning("batch_words should be less than or equal to %d. "
-                           "Standard cython code truncates to the maximum if "
-                           "the individual texts are longer than %d words",
-                           MAX_WORDS_IN_BATCH, MAX_WORDS_IN_BATCH)
+            logger.warning(
+                "batch_words %d should be less than or equal to %d. "
+                "Cython code will truncate texts that are longer than %d words.",
+                batch_words,
```
piskvorky, line 601:
```suggestion
                batch_words, MAX_WORDS_IN_BATCH, MAX_WORDS_IN_BATCH,
```

dsandeep0138: Done

dsandeep0138:
@piskvorky Thanks for the review. I have made the changes based on your feedback.

@mpenkov keras related tests are failing. Can you please help with these? Thanks.

gojomo:
if the hope is for #2698 cleanup to be in 4.0 (and that's my hope!), these smaller tweaks to related classes should either (a) be held until after that's merged; or (b) be based off the massive changes there. Otherwise, they make its integration/progress harder. 

The related FastText & Doc2Vec classes would have the same issue, so any fix that only touches word2vec* files is incomplete. 

From my view, this doesn't fix the deeper problems with the Cython limit:

* submitted texts are silently truncated to the 10000 word limit without warning, which users may not realize
* modern users with far more memory (and larger texts) still face this limit, and can only change via the onerous expert-level approach of editing cython & recompiling, or extra pre-splitting of their input texts

A proper & better fix would be to essentially eliminate this limit entirely. One way would be to auto-split the users' input texts - somewhat rough in its handling of the users' data, but less so than silently truncating it! Another would be to use the `alloca()` c-function in the Cython code, which could stack-allocate whatever's needed for the user batch without recourse to a compiled-in constant. (Alas, `alloca()` isn't technically part of the POSIX standard, but I believe in practice it's everywhere we care about. Alternatively [variable-length-arrays](https://gcc.gnu.org/onlinedocs/gcc/Variable-Length.html) might do the trick, are part of ISO C99, but apparently not supported by Cython through 3.0a5.) 

Perhaps best-of-both-worlds would be to do both, use `alloca()` to handle any-sized text the user chooses to send, but also include an optional configurable 'split-if-over-this-size' cap, to help prevent them from shooting themselves in the foot with giant stack-overflowing texts. I ran experiments improving this in 2015, viewable in old branch `batching` (https://github.com/RaRe-Technologies/gensim/compare/develop...gojomo:batching), that could be adapted forward, if `alloca()` is considered OK to use. 

Given that the default for `batch_words` is already set to match the Cython limit – so only those rare users who choose to deviate from the default could run into any confusion – and that the best fix would be removing this parameter/complication entirely, I'd rather not even add such warnings-for-rare-cases. Better to just prioritize the better fix, or remove the ability to configure Python-side `batch_words` at all. (I think the only reason for setting a *smaller* value might be to simulate actual job-batching in tiny unit tests, which is also of minuscule importance.)

piskvorky:
We'd very much prioritize a better fix (your PR). We cannot stall all development until that happens though – especially since its ETA is unknown and the PR is not open for review yet ("early WIP").

Would it be possible to split up your PR (@mpenkov 's idea)? Get some parts of the changes in now, rest later?

The changes will be essential I believe. So basically anything to speed up the process and avoid a behemoth PR that blocks all development.

gojomo:
There's no current fix for this in #2698, nor is one planned.  

It's more that this is an incomplete fix in the wrong direction, for a tiny theoretical issue that there's not even any evidence anyone has ever hit. The bug it's linked-to is just, "I think a larger value should not be accepted". But this patch doesn't actually even reject larger values, it just issues a warning – which from experience many users simply ignore. And it only does so for one of the 3 classes where there could be a concern. 

So in my opinion all the discussion/effort so far around this bug/PR has been wasted on misguided minutiae. (Similarly: #2537/#2551 - a bunch of go-round about a non-issue.)

There is a closely-related problem - that texts are silently truncated in the Cython code – but that deserves a totally different better fix that would further obsolete this make-work. 

But even if theoretically this change was a **good** 'housekeeping' fix, a slight new usability comment/warning. I believe it would still makes sense for that to either (1) wait for later application, as such tiny changes should be lower-priority and wouldn't drive a big number release, anyway; or (2) build on the #2698 in progress, meaning less conflict-resolution work by other authors or myself. No one need be 'blocked', they just shouldn't be putting a fresh coat of paint onto walls already marked to be knocked-out. 

I don't want my work-in-progress to be holding anything else up, but so far I've not heard of any priority features/fixes that need be blocked that way. I'll post more on the status of #2698 there.

mpenkov:
Finally found the time to catch up this.

I agree with @gojomo that this PR is far from a perfect solution and it may make merging #2698 more difficult. At the same time, I think the changes here are an improvement over the existing code, and are relatively innocuous. I think it is best to merge this PR.

@piskvorky WDYT?

gojomo:
Disagree with 'improvement' and 'innocuous' characterizations very strongly: it's adding lines-of-code for a merely-theoretical issue, while creating extra merge work for others (mostly me), when the real fix for the related underlying limitation will be something completely different which will cause this code to be thrown away (which will be yet more work for whoever tackles that, which is likely me).

piskvorky:
Yeah the warning seems useless. My review was superficial, I didn't realize this would help approximately no one.

My (incorrect) reading was that the user will see a warning when texts are truncated (good), but that's not the case with this PR. Longer texts are still truncated, the warning is only shown if the user explicitly overrides an obscure parameter (vanishingly rare).

mpenkov:
OK, so in that case, the most appropriate action is to close this PR.

@dsandeep0138 Thank you for your contribution, but as you can see, the situation around the problem is quite difficult, and we cannot accept this PR. I hope this does not discourage you from future contributions to gensim.

dsandeep0138:
> OK, so in that case, the most appropriate action is to close this PR.
> 
> @dsandeep0138 Thank you for your contribution, but as you can see, the situation around the problem is quite difficult, and we cannot accept this PR. I hope this does not discourage you from future contributions to gensim.

Thank you @mpenkov for the encouraging words! Will surely try and contribute!

PR #2864, opened 2020-06-22
Replace numpy.random.RandomState with SFC64 - for speed

### LINKED ISSUE #2782: random.RandomState with different versions of numpy has vastly different performance
the performance of random.RandomState in word2vec.py (version 3.8.0)

```
def seeded_vector(self, seed_string, vector_size):
         once = random.RandomState(self.hashfxn(seed_string) & 0xffffffff)
         return (once.rand(vector_size) - 0.5) / vector_size
```

seemingly depends greatly on the version of numpy installed. With numpy = 1.14.3, the following code 

```
from  numpy.random import RandomState as Ran
from time import time
t1 = time()
for i in range(100000):
    temp = Ran(hash((i)) & 0xffffffff)
t2 = time()
t2-t1 
```

produced 
`0.28105926513671875`
exactly the same code with numpy= 1.18.1 produced 

`18.590345859527588`

I noticed this because I was training a model with millions of words as vocabulary, and after updating numpy unwittingly (via a anaconda update), I noticed that the time for build_vocab was significantly longer, and after some debugging, I nailed it down to random.RandomState in the `seeded_vector` function. 
I know this is indeed a numpy issue, but even they mentioned it that RandomState is legacy (https://docs.scipy.org/doc/numpy/reference/random/performance.html). Therefore I wonder if you have some plans to upgrade randomstate? Thanks!

zygm0nt:
This PR is related to #2782 

It changes the randomness generator from obsolete `RandomState` to the fastest alternative currently `SFC64`.

Closes #2782

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1691,8 +1691,8 @@ class Word2VecTrainables(utils.SaveLoad):
     def seeded_vector(self, seed_string, vector_size):
         """Get a random vector (but deterministic by seed_string)."""
         # Note: built-in hash() may vary by Python version or even (in Py3.x) per launch
-        once = random.RandomState(self.hashfxn(seed_string) & 0xffffffff)
-        return (once.rand(vector_size) - 0.5) / vector_size
+        once = random.Generator(random.SFC64(self.hashfxn(seed_string) & 0xffffffff))
+        return (once.random(vector_size) - 0.5) / vector_size
 
     def reset_weights(self, hs, negative, wv):
         """Reset all projection weights to an initial (untrained) state, but keep the existing vocabulary."""


piskvorky:
Needs a significant and detailed description (currently empty). Can you please extend it?

* What is the motivation?
* Who is affected and why?
* Effects of the PR: benchmarks, potential pitfalls, backward compatibility?

Plus unit tests. Cheers.

zygm0nt:
@piskvorky I've added a short description, because it's just a PR for the mentioned issue. 

I've observed the same problem as mentioned in the issue and found the fix helps.

Would gladly create a test for that, but can't think of a way to check a) speed, b) randomness - in such a test. Any suggestions?

piskvorky:
Thanks! Can you include some benchmark, on a realistic (non-trivial) dataset? E.g. `text9`, `wikipedia`.

gojomo:
This code doesn't yet even allow its single modified method to complete, because the `SFC64` instance does not implement the `.rand()` method called on the next line. All unit tests that use this method are failing. 

The `SFC64` must be wrapped in a `Generator`, and then its method `random()` should be called instead or `rand()`. 

But, once that's fixed, just a single before/after comparison on a run that takes a few minutes (text8 or text9) should be enough, to be sure the change is not having the opposite of the intended effect. 

It's just a 2-line fix, to match the Numpy docs'  [performance recommendations](https://numpy.org/doc/stable/reference/random/performance.html), which:

* no longer recommend the old MT19973 generator, noting that it both "fails some statistical tests and is not especially fast compared to modern PRNGs". 

* recommend either `PCG64`, in general, or `SFC64`, if we want speed & don't need 'jumpability' – which seems exactly our need.

piskvorky:
Oh wow. @zygm0nt you haven't run this code at all?

zygm0nt:
Sorry for that. Will fix the errors, than let you know.

zygm0nt:
@piskvorky @gojomo I've fixed the issue, now almost all tests are passing. Travis reports errors, but don't see the connection here with changes from this PR.

As advised I'll prepare a test run with standard dataset. How should I provide it? Log from a run would suffice?

gojomo:
The one test that's failing appears unrelated to this change.

(It's a type assertion in `test_keras_integration.py`, which hasn't changed since 2018, and only failing in the Python 3.8 tests, passing in 3.7 and 3.6... whereas all of 3.8, 3.7, & 3.6 passed on `develop` just 8 days ago. I wouldn't hold back this change to investigate; if it persists, it can be looked into separately.)

Just ensuring a nontrivial example runs before and after, and checking that the logged timings indicate no worsening of the relevant 'resetting model weights' step (and excerpting that report here or in the issue) would be enough for me, given all other relevant tests passing.

mpenkov:
> The one test that's failing appears unrelated to this change.

It appears unrelated, but the same test is passing on the master branch. If we move forward with this PR, we should:

1) Skip the test under Py3.8 to make sure everything stays green
2) Open up a separate issue to investigate, fix the underlying issue and re-enable the test

@zygm0nt Please do (1). I'll do (2).

gensim/models/word2vec.py
```diff
@@ -1691,8 +1691,10 @@ def prepare_weights(self, hs, negative, wv, update=False, vocabulary=None):
     def seeded_vector(self, seed_string, vector_size):
         """Get a random vector (but deterministic by seed_string)."""
         # Note: built-in hash() may vary by Python version or even (in Py3.x) per launch
-        once = random.RandomState(self.hashfxn(seed_string) & 0xffffffff)
-        return (once.rand(vector_size) - 0.5) / vector_size
+        once = random.Generator(
```
mpenkov, line 1694: This can all fit on one line, surely?

gojomo:
Why would we want 'everything green' by checking in version-specific test-skips - cruft that we'll have to remember to undo later? 

If it's a legit test that failing, it should keep failing until it's fixed. If the changes here didn't cause the failure – but rather revealed something else that broke elsewhere recently – that shouldn't hold up this fix. 

(I'm not also sure it's accurate to say everything's currently "passing" on the master branch - only that it passed on the last commit ~9 days ago. A simple retry of the build might reveal the same problem there.)

mpenkov:
> Why would we want 'everything green' by checking in version-specific test-skips - cruft that we'll have to remember to undo later?

Because that's how we know at a glance that everything still works. Without the all-green, we will have to check the failed CI runs for each build (did it fail because we expected it to fail, or was there another, unexpected problem?).

We won't have to remember to remove anything. We'll open a ticket for it and deal with it when we have the bandwidth.

> If it's a legit test that failing, it should keep failing until it's fixed.

First, we don't know if the test is failing because there's a problem with the code, its dependencies, or the test itself (or some combination of the above).

Second, if keep failing tests around, they'll fail for *every build* until we fix them. This means that for every build, we'll have to dig through CI logs and determine if the failure is expected or not. This is significant overhead for a maintainer. I'm strongly against this - I'd rather spend my time more productively, plus there's no guarantee that manual checks are 100% correct - we'll always get the call wrong once in a while.

> If the changes here didn't cause the failure – but rather revealed something else that broke elsewhere recently – that shouldn't hold up this fix.

There are two ways forward:

1) Merge this PR and deal with the problem later
2) Deal with the problem first and then merge this PR

We're not in a terrible rush to merge this, so my preference is (2). The amount of effort required for dealing with the problem in the immediate sense is low (disabling a test is easy), anyway.

gojomo:
But why would we want to "know at a glance that everything still works" when in fact, everything still *doesn't* work – a test is broken. Special-casing a test to be off for one Python version (if in fact that's even the real pattern-of-failure here – unclear from a single observation) just creates a dishonest green-light. It's like removing the battery from a smoke-detector instead of investigating "what's that burning smell?" 

And it contaminates this PR/commit with a totally unrelated bit of kludge code that's destined to be removed ASAP. 

Again, it's quite possible the trunk is *already broken* by the same mysterious (or intermittent) factor that caused this. (If you have a button that tries a rebuild, please try it. This discussion about protecting the green may be moot.) Every attempt to build on that *should* see this problem until fixed, and thus I do agree with your (2) "Deal with the problem first and then merge this PR" - to the extent that I wouldn't want anyone to do any temporary test-disablements until the problem is better understood.

mpenkov:
@gojomo Let's continue our discussion in #2866.

zygm0nt:
Didn't have time to setup the environment with mentioned datasets (wikipedia, text8). But I've repeated the training with my target dataset. And must say that this change resulted in dramatic reduce in time of resetting weights.

First log is from original training:

```
2020-06-21 21:53:43,103 [12113] INFO     gensim.models.word2vec: Loading a fresh vocabulary
2020-06-21 21:55:30,397 [12113] INFO     gensim.models.word2vec: effective_min_count=3 retains 19115982 unique words (53% of original 35448710, drops 16332728)
2020-06-21 21:55:30,397 [12113] INFO     gensim.models.word2vec: effective_min_count=3 leaves 950533618 word corpus (97% of original 972371594, drops 21837976)
2020-06-21 21:56:38,747 [12113] INFO     gensim.models.word2vec: deleting the raw counts dictionary of 35448710 items
2020-06-21 21:56:39,623 [12113] INFO     gensim.models.word2vec: sample=1e-05 downsamples 3133 most-common words
2020-06-21 21:56:39,623 [12113] INFO     gensim.models.word2vec: downsampling leaves estimated 641405759 word corpus (67.5% of prior 950533618)
2020-06-21 21:58:03,009 [12113] INFO     gensim.models.base_any2vec: estimated required memory for 19115982 words and 100 dimensions: 24850776600 bytes
2020-06-21 21:58:03,010 [12113] INFO     gensim.models.word2vec: resetting layer weights
2020-06-21 23:04:38,790 [12113] INFO     gensim.models.base_any2vec: training model with 64 workers on 19115982 vocabulary and 100 features, using sg=1 hs=0 sample=1e-05 negative=5 window=15 
```

The current newest version (3.8.3) takes ~1h15 minutes to reset weights.

This log is with the fixed version:

```2020-06-26 11:53:44,527 [4680] INFO     gensim.models.word2vec: collected 35448710 word types from a corpus of 972371594 raw words and 65740684 sentences
2020-06-26 11:53:44,527 [4680] INFO     gensim.models.word2vec: Loading a fresh vocabulary
2020-06-26 11:55:34,153 [4680] INFO     gensim.models.word2vec: effective_min_count=3 retains 19115982 unique words (53% of original 35448710, drops 16332728)
2020-06-26 11:55:34,153 [4680] INFO     gensim.models.word2vec: effective_min_count=3 leaves 950533618 word corpus (97% of original 972371594, drops 21837976)
2020-06-26 11:56:44,745 [4680] INFO     gensim.models.word2vec: deleting the raw counts dictionary of 35448710 items
2020-06-26 11:56:45,636 [4680] INFO     gensim.models.word2vec: sample=1e-05 downsamples 3133 most-common words
2020-06-26 11:56:45,636 [4680] INFO     gensim.models.word2vec: downsampling leaves estimated 641405759 word corpus (67.5% of prior 950533618)
2020-06-26 11:58:12,694 [4680] INFO     gensim.models.base_any2vec: estimated required memory for 19115982 words and 100 dimensions: 24850776600 bytes
2020-06-26 11:58:12,694 [4680] INFO     gensim.models.word2vec: resetting layer weights
2020-06-26 12:13:09,172 [4680] INFO     gensim.models.base_any2vec: training model with 64 workers on 19115982 vocabulary and 100 features, using sg=1 hs=0 sample=1e-05 negative=5 window=15```

So, the time difference between `resetting layer weights` log and the next log is ~14minutes, with fix.


Would you like me to present also the results for wikipedia/text8?

gojomo:
Personally, I think this is enough to apply the quick fix: a 5x speedup, and matching the `numpy` recommendations.

Stylistically I'd agree with @mpenkov's suggestion that this Generator+algorithm instantiation can fit on 1-line. (That seems the idiomatic usage in numpy example code, too. 

It is likely we could get a significant further speedup by letting `numpy` fill the whole array in one call, but we'd lose the current slight nod to determinism, where each vector's initial value is seeded by its key. (Thus, subsequent runs with the same model `seed` – but even with different vocabularies via a changed `min_count` etc – start each word at the same random place. Yet, given all the other things pulling against reproducible runs, especially multithreading, & my belief users should generally be discouraged from seeking reproducibility with such stochastic algorithms, this seeding trick probably isn't worth the performance cost.)

mpenkov:
@zygm0nt Please merge upstream/develop to fix the problem with keras.

zygm0nt:
Done, rebased and fixed stylistic remarks.

piskvorky:
Probably too late now, but avoid force-pushing to shared branches. It makes reviewing harder and loses history.

mpenkov:
@zygm0nt We've merged this. Thank you for your contribution!

zygm0nt:
Thanks!

Issue #2865, opened 2020-06-26
Investigate and fix Keras problem under Python 3.8


mpenkov:
One of our unit tests fails on Travis under Py3.8.

```
=================================== FAILURES ===================================

_____________ TestKerasWord2VecWrapper.testEmbeddingLayerCosineSim _____________

self = <gensim.test.test_keras_integration.TestKerasWord2VecWrapper testMethod=testEmbeddingLayerCosineSim>

    def testEmbeddingLayerCosineSim(self):

        """

        Test Keras 'Embedding' layer returned by 'get_embedding_layer' function for a simple word similarity task.

        """

        keras_w2v_model = self.model_cos_sim

        keras_w2v_model_wv = keras_w2v_model.wv

    

        embedding_layer = keras_w2v_model_wv.get_keras_embedding()

    

        input_a = Input(shape=(1,), dtype='int32', name='input_a')

        input_b = Input(shape=(1,), dtype='int32', name='input_b')

        embedding_a = embedding_layer(input_a)

        embedding_b = embedding_layer(input_b)

        similarity = dot([embedding_a, embedding_b], axes=2, normalize=True)

    

>       model = Model(input=[input_a, input_b], output=similarity)

embedding_a = <tf.Tensor 'embedding_4/Identity:0' shape=(None, 1, 100) dtype=float32>

embedding_b = <tf.Tensor 'embedding_4_1/Identity:0' shape=(None, 1, 100) dtype=float32>

embedding_layer = <tensorflow.python.keras.layers.embeddings.Embedding object at 0x7f603df0d130>

input_a    = <tf.Tensor 'input_a_3:0' shape=(None, 1) dtype=int32>

input_b    = <tf.Tensor 'input_b_3:0' shape=(None, 1) dtype=int32>

keras_w2v_model = <gensim.models.word2vec.Word2Vec object at 0x7f603df0d760>

keras_w2v_model_wv = <gensim.models.keyedvectors.Word2VecKeyedVectors object at 0x7f603df0d250>

self       = <gensim.test.test_keras_integration.TestKerasWord2VecWrapper testMethod=testEmbeddingLayerCosineSim>

similarity = <tf.Tensor 'dot_3/Identity:0' shape=(None, 1, 1) dtype=float32>

gensim/test/test_keras_integration.py:62: 

[...]

PR #2866, opened 2020-06-26
temporarily disable failing unit test

mpenkov:
The test was passing 10 days ago, but now it fails. The failure affects #2864, #2867 and possibly other in-flight PRs.

My plan is:

1. Merge this PR and disable the test
2. Investigate #2865 when I have time (or when someone else steps up to do it)
3. Re-enable the test

### DIFF
--- a/gensim/test/test_keras_integration.py
+++ b/gensim/test/test_keras_integration.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import unittest
 import numpy as np
@@ -46,7 +47,7 @@ class TestKerasWord2VecWrapper(unittest.TestCase):
         self.assertEqual(sims, sims2)
 
     @unittest.skipIf(
-        sys.version_info[:2] == (3, 8),
+        os.environ['TRAVIS'] == 'true' and sys.version_info[:2] == (3, 8),
         'see https://github.com/RaRe-Technologies/gensim/issues/2865'
     )
     def testEmbeddingLayerCosineSim(self):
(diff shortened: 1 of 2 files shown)


mpenkov:
@gojomo continued from #2864:

> But why would we want to "know at a glance that everything still works" when in fact, everything still doesn't work – a test is broken.

In this case, "everything" really means "everything, except for stuff that we know is broken". Remember, at any one time gensim literally has [dozens of bugs](https://github.com/RaRe-Technologies/gensim/labels/bug). Can you imagine what life for the maintainer would be like if each of those bugs caused a test to fail on each build?

> Special-casing a test to be off for one Python version (if in fact that's even the real pattern-of-failure here – unclear from a single observation) just creates a dishonest green-light. It's like removing the battery from a smoke-detector instead of investigating "what's that burning smell?"

No, that's a false analogy. Let me adjust it so it reflects reality a little better.

We're sitting in a room working on an [important time-sensitive project](https://github.com/RaRe-Technologies/gensim/projects/8). Time is short, because everybody has a life outside of gensim. Suddenly, one of the many smoke detectors in the room starts beeping (remember, we have multiple tests that cover that functionality, and only one of them fails: Py3.8 under Travis). Our options:

1. Work through the beeping. This is analogous to working with failing CI builds.
2. Stop work and thoroughly investigate the problem. Bring out the smoke detector manual. Call customer support, etc. This is analogous to solving the problem properly.
3. Disable the smoke detector, schedule a proper investigation once the important stuff is out of the way.

(1) is a PITA. (2) violates our priorities. (3) is this PR.

mpenkov:
Travis does our Linux builds. Azure is Windows.

mpenkov:
Closing in favor of #2868

PR #2867, opened 2020-06-26
Expose max_final_vocab parameter in FastText constructor

mpenkov:
Continued from #2516


mpenkov:
@scribu I finally merged your original contribution. Thank you!

PR #2868, opened 2020-06-26
[trial] pin keras=2.3.1 because 2.4.3 causes KerasWord2VecWrappper test failure in Py 3.8

gojomo:
May serve as work-around for #2865.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -314,7 +314,7 @@ if sys.version_info >= (3, 7):
     # See https://github.com/RaRe-Technologies/gensim/pull/2814#issuecomment-621477948
     linux_testenv += [
         'tensorflow',
-        'keras',
+        'keras==2.3.1',
     ]
 
 NUMPY_STR = 'numpy >= 1.11.3'


piskvorky:
Quick googling of the error message `Keyword argument not understood` suggests that Keras expects `inputs` and `outputs`, where we pass it `input` and `output`.

Probably a result of the new 2.4.0 Keras release from 9 days ago: https://github.com/keras-team/keras/releases

gojomo:
I believe the failure in the PRNG/SFC64 PR was in the same test class, but slightly different - an assertion failure. And if it's strictly a keras 2.3.x to 2.4.x change, it's curious that py3.6 & py3.7 pass. (A quick glance at the py3.7 log shows that yes, keras 2.4.3 was installed & passed all tests there.) 

So some more investigation may still be necessary, but I wanted to "try the simplest thing possible" before doing something like disabling a test that's otherwise there, and failing, for a real informational reason. 

Given that it is likely a keras change, and others may be bitten by it and get them to ensure py3.8 works the same as earlier Python in due time, I'd put full resolution of #2865 into a "wait and see if somebody else fixes the issue" bin. By the time gensim needs to do a release, we could (1) see if by then, simply removing the pin no longer causes any problems; (2) leave the pin & add a brief release note "stay with keras<3.4 if on Python 3.8" (which might be something users on Py3.8 would need to do anyway); or finally and least-desirably: (3) actually dig into the keras change & fix on gensim's end.

mpenkov:
Looks like the work-around is a success. I like it more than disabling the test.

Thank you @gojomo!

PR #2871, opened 2020-07-03
Clarify that license is LGPL-2.1

pombredanne:
Use only the license attribute with an SPDX license id for V2.1 of the
LGPL.

The setup.py classifier was inconsistent with the license attribute as 
one pointed to LGPL-2.0-or-later and one to LGPL-2.1-only which seems 
to be otherwise the correct license based on header comments and
documentation. Classifiers do not support the LGPL v2.1 and are 
eventually being subsumed by the license field [1]

[1] https://discuss.python.org/t/improving-license-clarity-with-better-package-metadata/2154

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -354,7 +354,7 @@ setup(
     url='http://radimrehurek.com/gensim',
     download_url='http://pypi.python.org/pypi/gensim',
 
-    license='LGPLv2.1',
+    license='LGPL-2.1-only',
 
     keywords='Singular Value Decomposition, SVD, Latent Semantic Indexing, '
         'LSA, LSI, Latent Dirichlet Allocation, LDA, '
@@ -369,7 +369,6 @@ setup(
         'Development Status :: 5 - Production/Stable',
         'Environment :: Console',
         'Intended Audience :: Science/Research',
-        'License :: OSI Approved :: GNU Lesser General Public License v2 or later (LGPLv2+)',
         'Operating System :: OS Independent',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',


piskvorky:
What's SPDX, why should we use their ids (`LGPL-2.1-only`)?

I'll have to look up the implications of "2.1 only" vs "2.1 or later", I don't remember. Or do you have a TL;DR somewhere handy @pombredanne ?

pombredanne:
@piskvorky SPDX is a project at the Linux Foundation to help streamline how we eXchange Data about Software Products, and in particular licensing information. See https://spdx.dev/
It maintains the de-facto reference list of license short codes https://spdx.org/licenses/  that are now use in many places such as the Linux Kernel, Rubygems and npm and should be soon be adopted for Python packages as I am working on a PEP.

The  "2.1 only" vs "2.1 or later" and the ids of `LGPL-2.1-only` and `LGPL-2.1-or-later` come from the Free Software Foundation, the author of these licenses. :)

The only vs. or later is simple enough:
- with "only" only that version of the LGPL will apply to your code.
- with "or later", that version and any other versions published by the FSF can be selected by a user as the license that applies to your code. So in practice it means today: LGPL 2.1 OR LGPL 3.0 and whenever there are new versions, these become automatically a choice that a user can make.

Just a side note: since you have 350+ contributors that contributed some bits of code under you current license, you may not be able to change the license without getting some approval first? unless they explicitly assigned their copyrights to you?

piskvorky:
Eh, I know what "only" and "later" means :) My question was for TL;DR implications.

Yeah we ask all contributors to assign copyright to me.

pombredanne:
> Eh, I know what the word "only" and "later" means 

Sorry for that!

The primary implications are whether or not someone can mix your code with various L/GPL licenses where the FSFS states that some relicensing may be needed in some cases.
See https://www.gnu.org/licenses/quick-guide-gplv3.html#new-compatible-licenses  for a somewhat convoluted discussion.

Another implication is whether you trust the FSF with future versions of the licenses of not: with or later, this assume that you trust them.

piskvorky:
Thanks – little difference to me. And since this never came up in 10 years of Gensim, I guess little difference to others too.

@mpenkov are you OK with this PR?

mpenkov:
Yeah, I don't think it makes a huge difference either way. Let's merge.

PR #2874, opened 2020-07-07
Fix broken documentation link to mycorpus.txt

### LINKED ISSUE #2872: Broken file link in `run_corpora_and_vector_spaces` tutorial
#### Problem description

The `run_corpora_and_vector_spaces.ipynb` tutorial depends on a file on the web, and that file is missing.

See https://groups.google.com/g/gensim/c/nX4lc8j0ZO0

#### Versions

Please provide the output of:

```python
```

Unknown (probably any).

benbroks:
Fixes #2872 

Motivation: Old notebook was dependent on a URL for a text file. Wanted to make sure this was able to run locally. No has to download to run the notebook.

Fixes: Creates mycorpus.txt inline. Fixed an image reference at the end of the same notebook.

### DIFF
--- a/docs/src/auto_examples/core/run_corpora_and_vector_spaces.ipynb
+++ b/docs/src/auto_examples/core/run_corpora_and_vector_spaces.ipynb
@@ -1,439 +1,838 @@
 {
-  "cells": [
-    {
-      "cell_type": "code",
-      "execution_count": null,
-      "metadata": {
-        "collapsed": false
-      },
-      "outputs": [],
-      "source": [
-        "%matplotlib inline"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {},
-      "source": [
-        "\nCorpora and Vector Spaces\n=========================\n\nDemonstrates transforming text into a vector space representation.\n\nAlso introduces corpus streaming and persistence to disk in various formats.\n"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": null,
-      "metadata": {
-        "collapsed": false
-      },
-      "outputs": [],
-      "source": [
-        "import logging\nlogging.basicConfig(format='%(asctime)s : %(levelname)s : %(message)s', level=logging.INFO)"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {},
-      "source": [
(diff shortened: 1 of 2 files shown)


piskvorky:
@mpenkov aren't these files auto-generated? What was the process for updating the docs?

gojomo:
Given that the file is only ~19 lines and ~69 words, I'm surprised it was ever set up to be downloaded from a remote URL!

piskvorky:
It was a demonstration of "remote input streaming" IIRC. Not worth the trouble.

mpenkov:
Yes, the .ipynb is an autogenerated file. We should not be changing it directly. If we want to go ahead with this update, we have to [update the .py file](https://github.com/RaRe-Technologies/gensim/tree/develop/docs/src/gallery#subdirectories) in docs/src/gallery. 

Personally, I'm not sure of what the benefit of ensuring that notebook runs without a network connection out of the box, but if we _do_ ahead with it, then:

1. We should update the .py file and generate the .ipynb from it
2. We don't need to import smart_open anymore.
3. We don't need to write the corpus to a file.  The generator can just yield the strings as necessary.

benbroks:
Apologies if this isn't along the lines of what you were looking for. Made the same changes in the .py file.

mpenkov:
@benbroks Sorry for the delay in reviewing this.

It looks like the documentation build is failing. Is it working for you locally?

If yes, then it's likely you haven't included all autogenerated files in this PR. At a glance, `gensim/docs/src/auto_examples/core/run_corpora_and_vector_spaces.rst` should also be included.

If no, then please make sure the docs build locally, and then update this PR.

piskvorky:
Ping @mpenkov this issue keeps coming back, people following the tutorial still get tripped up.

I placed the file under `https://radimrehurek.com/mycorpus.txt`, so it doesn't get overwritten / deleted with each release. Do we just update the location in the tutorial(s)?

I'm not a fan of the approach from this PR – not portable (`/tmp`), adds too much distracting setup (read/write files, irrelevant).

mpenkov:
We don't need to do anything, this is already resolved by https://github.com/RaRe-Technologies/gensim/pull/3148/files

@benbroks Thank you for your contribution, but the problem has already been resolved by another PR.

PR #2878, opened 2020-07-09
Update utils.py

benJman247:
most recent version of smart_open does not have 'open' method. Instead it has 'smart_open'. Have changed in my utils file and am offering my change to the community.


piskvorky:
I don't think that's correct – you should use `smart_open.open`.

benJman247:
> I don't think that's correct – you should use `smart_open.open`.

You're right. It works now. I have no idea why smart_open.open is working when it hasn't for days...

Issue #2879, opened 2020-07-12
Proposal: Replace word2vec-specific implementation w/ constrained subclass of FastText


gojomo:
I believe that everything `Word2Vec` does can also already be done equivalently via `FastText`, with constrained options (like turning off char-ngrams). So we could potentially eliminate a lot of duplicated algorithm code & future maintenance overhead by recharacterizing `FastText` as the root of our 2Vec hierarchy, instead of the original `Word2Vec` code. 

We'd want to ensure `FastText` gracefully handles char-ngram-disablement (by not making allocations/class-choices only required when ngrams are enabled), and make `Word2Vec` a subclass of `FastText` which hides some options, rather than the other way around with `FastText` as a subclass that adds new options. 

User-visible APIs might not change at all. Some new ngram-enabled `Doc2Vec` options might become possible if `Doc2Vec` derives from `FastText` (like an inference that works at least a little, in some modes, even with all OOV words). 

(Relates to: #2852)

piskvorky:
Makes sense to me. I also can't think of anything – beyond performance and branding – that word2vec has over fasttext.

gojomo:
A quick check of `Word2Vec` (all defaults) & `FastText` (all defaults except n-grams disabled) has already shown a few issues that will need addressing:

* it's a bit muddled how ngrams should be turned off - different comments/code imply either (1) `max_n` less than `min_n`; (2) `char_ngrams=0`; (3) `char_ngrams==0` *and* `max_n==0` (in `FastText.__init__()`).
* the `train()` step that takes 55s in plain `Word2Vec` take 84s in `FastText` - a lot of overhead for a disabled feature
* an `evaluate_word_analogies()` w/ `question_words.txt` fails *everything* on the FT-without-ngrams model, rather than roughly matching the should-be-equivalent `Word2Vec` - unclear why
* even without ngrams, the `FastText` model still maintains its two separate sets of per-word vectors: the raw whole-word vector, and the whole-word-plus-ngram-enrichment calculation – so RAM usage & on-disk storage is more expansive than would be best

Most of these probably have straightforward fixes, but the existing gensim `FastText` implementation isn't yet a simple drop-in replacement for `Word2Vec` as one might have hoped. (Potentially, rough parity in ngrams-disabled performance and quality-evaluations could have been part of the FT code's original review/testing.)

piskvorky:
Thanks for that check, that's a great start. Are the conclusions above still true after #2891?

[...]

PR #2891, opened 2020-07-22
[MRG] Fix FastText word-vectors w/ ngrams/buckets off

gojomo:
In looking at the potential for #2879, I'm finding bugs & suboptimalities in `FastText`. This quickie lets FastText work like Word2Vec, when ngrams are disabled - something that should've worked earlier, but seems to have never been tested.


gojomo:
(last commit 8f64b48 doesn't belong here: force-reverting)

gensim/models/fasttext.py
```diff
@@ -1401,6 +1401,7 @@ def adjust_vectors(self):
 
         """
         if self.bucket == 0:
+            self.vectors = self.vectors_vocab  # no ngrams influence
```
piskvorky, line 1404: Is our serialization smart enough to realize these attributes reference the same object, and avoid serializing both (into two separate files)?

gensim/test/test_fasttext.py
```diff
@@ -1371,6 +1372,7 @@ def _read_fb(fin):
 
 
 class ZeroBucketTest(unittest.TestCase):
+    """Tests FastText with no buckets/no-ngrams (essentially FastText-as-Word2Vec"""
```
piskvorky, line 1375:
```suggestion
    """Test FastText with no buckets / no-ngrams: essentially FastText-as-Word2Vec."""
```

gensim/test/test_fasttext.py
```diff
@@ -1379,6 +1381,15 @@ def test_out_of_vocab(self):
         model = train_gensim(bucket=0)
         self.assertRaises(KeyError, model.wv.word_vec, 'streamtrain')
 
+    def test_cbow_neg(self):
+        """See gensim.test.test_word2vec.TestWord2VecModel.test_cbow_neg"""
```
piskvorky, line 1385:
```suggestion
        """See `gensim.test.test_word2vec.TestWord2VecModel.test_cbow_neg`."""
```

gensim/models/fasttext.py
gojomo: It isn't (except in rare case array is so small it goes through normal single=-file pickling), but next-up PR #2892 simply explicitly avoids saving `.vectors` at all, from `FastTextKeyedVectors`, since even where it's different from `.vectors_vocab`, it's still just a  re-calculable cache of full-word-plus-ngram vectors derived from `.vectors_vocab` & `.vectors_ngram`.

gojomo:
Low risk, removes flaw, has test, @piskvorky-reviewed - pushing merge button.

PR #2896, opened 2020-07-24
Make docs clearer on `alpha` parameter in LDA model

xh2:
### Summary
Modified docstring on `alpha` in LDA model around 'symmetric' and 'asymmetric' options

### Motivation
- When I first read the doc, I saw by default `alpha='symmetric'`, but in the docs, only 'asymmetric' and 'auto' are listed as acceptable strings, so my first instinct was it was a typo. To be clearer, especially to new users, I think we should spell out all possible options, including the default 'symmetric' in the doc
- The current 'asymmetric' description doesn't seem to fit the actual formula in the code

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -538,7 +538,7 @@ class LdaModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
 
                 * an 1D array of length equal to the number of expected topics,
                 * 'symmetric': Uses a fixed symmetric prior per topic,
-                * 'asymmetric': Uses a fixed normalized asymmetric prior of `1.0 / (topic_index + sqrt(topic_no))`,
+                * 'asymmetric': Uses a fixed normalized asymmetric prior of `1.0 / (topic_index + sqrt(num_topics))`,
                 * 'auto': Learns an asymmetric prior from the corpus.
         name : {'alpha', 'eta'}
             Whether the `prior` is parameterized by the alpha vector (1 parameter per topic)
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -378,7 +378,7 @@ class LdaModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
             Can be set to an 1D array of length equal to the number of expected topics that expresses
             our a-priori belief for the each topics' probability.
             Alternatively default prior selecting strategies can be employed by supplying a string:
-                
+
                 * 'symmetric': Default; uses a fixed symmetric prior per topic,
                 * 'asymmetric': Uses a fixed normalized asymmetric prior of `1.0 / (topic_index + sqrt(topic_no))`,
                 * 'auto': Learns an asymmetric prior from the corpus (not available if `distributed==True`).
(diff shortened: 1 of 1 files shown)


piskvorky:
You're right, thanks for the fix!

> The current 'asymmetric' description doesn't seem to fit the actual formula in the code

Which is correct – the code or the documentation? I don't remember the motivation or original paper for the `asymmetric` prior any more, unfortunately.

xh2:
> Which is correct – the code or the documentation? I don't remember the motivation or original paper for the `asymmetric` prior any more, unfortunately.

The paper by Hoffman only used symmetric priors, while citing [Wallach, Hanna & Mimno, David & Mccallum, Andrew. (2009). Rethinking LDA: Why priors matter. NIPS. 23. 1973-1981. ](http://dirichlet.net/pdf/wallach09rethinking.pdf) for asymmetric priors. But I couldn't find either formula in that paper either, so not really sure what it is supposed to be.

The existing doc on asymmetric alpha, however, is actually suggesting a symmetric distribution instead? No? That was also what confused me and made me think it was a typo.

piskvorky:
I don't think so – with `asymmetric`, different topics get a different alpha. With `symmatric`, all topics get the same alpha.

xh2:
yes agree! but isn't `1.0 / topicno` (current docstring description for `asymmetric`) uniform? (the actual code is indeed asymmetric). unless here `topicno` means index not number of topics?

piskvorky:
Yes. For sure that's meant as "index of the topic", not "total number of topics". Another piece of documentation that could use a fix!

gensim/models/ldamodel.py
```diff
@@ -379,7 +379,8 @@ def __init__(self, corpus=None, num_topics=100, id2word=None,
             our a-priori belief for the each topics' probability.
             Alternatively default prior selecting strategies can be employed by supplying a string:
 
-                * 'asymmetric': Uses a fixed normalized asymmetric prior of `1.0 / topicno`.
+                * 'symmetric': Default; uses a fixed symmetric prior per topic,
+                * 'asymmetric': Uses a fixed normalized asymmetric prior of `1.0 / (topic_index + sqrt(topic_no))`,
```
piskvorky, line 383:
```suggestion
                * 'asymmetric': Uses a fixed normalized asymmetric prior of `1.0 / (topic_index + sqrt(num_topics))`,
```

gensim/models/ldamodel.py
```diff
@@ -536,7 +537,8 @@ def init_dir_prior(self, prior, name):
             If `name` == 'alpha', then the prior can be:
 
                 * an 1D array of length equal to the number of expected topics,
-                * 'asymmetric': Uses a fixed normalized asymmetric prior of `1.0 / topicno`.
+                * 'symmetric': Uses a fixed symmetric prior per topic,
+                * 'asymmetric': Uses a fixed normalized asymmetric prior of `1.0 / (topic_index + sqrt(topic_no))`,
```
piskvorky, line 541:
```suggestion
                * 'asymmetric': Uses a fixed normalized asymmetric prior of `1.0 / (topic_index + sqrt(num_topics))`,
```

PR #2821, opened 2020-05-03
[MRG] Fix similarity bug in NMSLIB indexer

piskvorky:
* Minor docstring clean up in the `similarities.nmslib` module. Follow up from https://github.com/RaRe-Technologies/gensim/issues/2820#issuecomment-623074467.
* Added `similarities.nmslib` to our API documentation – it was missing before.
* Fixed a bug where `similarities.nmslib`'s `most_similar()` returned correct neighbours but with incorrect similarities.
* Fixed deprecation warnings in Annoy: `The default argument for metric will be removed in future version of Annoy. Please pass metric='angular' explicitly.`
* Went over our API reference page https://radimrehurek.com/gensim/apiref.html (visually, as users actually see it) and fixed a bunch of misformattings and errors.

Issue #2820, opened 2020-05-02
Prepare gensim 3.8.3


mpenkov:
OK guys, looks like we're getting close to releasing this thing. I've just updated the CHANGELOG - @piskvorky please have a look and make any changes as necessary. Each update will require a re-run of the CI and a rebuild of the wheels, so please keep that in mind.

Some other relevant things to check:

- [Release checklist](https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#making-a-new-release)
- [Release milestone](https://github.com/RaRe-Technologies/gensim/milestone/2?closed=1)
- [Diff with current develop HEAD](https://github.com/RaRe-Technologies/gensim/compare/develop...release-3.8.3?expand=1)

I've gone through the above myself and think like we're ready to release. @piskvorky @menshikh-iv Please let me know if you feel the same and we'll get this thing out the door.

menshikh-iv:
@mpenkov I merged https://github.com/RaRe-Technologies/gensim-wheels/pull/1, so, after build done (~6h, thanks to Appveyour and `wheelhouse_uploader` bug), we will be ready to release.

I think we are ready

piskvorky:
I went through the tickets yesterday, all good.

Except I saw this new comment in the diff:
<img width="930" alt="Screen Shot 2020-05-03 at 09 15 51" src="https://user-images.githubusercontent.com/610412/80908285-ef353b80-8d1e-11ea-9ca3-f4ea8b46b9ef.png">

and went to look at the `nmslib` module and that's another rabbit hole (lots of fixes, probably buggy). I'll open a PR soon, but probably not critical for 3.8.3.

piskvorky:
I also replace all references to MacPython on the [Release checklist](https://github.com/RaRe-Technologies/gensim/wiki/Developer-page without) with our new mirror.

menshikh-iv:
@piskvorky about nmslib: this stuff NEVER works on `py2` as I investigate, but works on python3.
Current impoirt are the simplest workaround for this problem (and will be no needed when we drop py2)

piskvorky:
Nothing major in the end. I opened the PR: #2821.

@mpenkov up to you whether we include that in 3.8.3 or not, WDYT?

mpenkov:
Let's leave it out, it's not critical.

mpenkov:
OK, we're done!

- https://github.com/RaRe-Technologies/gensim/releases/tag/3.8.3
[...]

PR #2892, opened 2020-07-23
[WIP] 2Vec SaveLoad improvements

gojomo:
The 2Vec classes were doing serialization fixups in save()/load() that should really, by `utils.SaveLoad` design, be in `_save_specials()`/`_load_specials()` - to better handle cases when one `SaveLoad` contains another.

Also looks like there may be no existing functional tests of loading of `gensim-3.8.3` `FastText` model, so it's possible that functionality is broken-with-an-unhelpful-failure. If so, I'll add a test & fix for that here, if possible.

(This builds atop #2891, which can & should be applied asap.)

PR #2899, opened 2020-07-26
[MRG] Fix similarity bug in NMSLIB indexer + documentation fixes

### LINKED ISSUE #2882: Working from trunk branch 'develop', gensim.__version__ reports as 3.8.1
If working in a checkout/branch from `develop`, `gensim.__version__` reports as `'3.8.1`, per https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/__init__.py#L8

Ideally, as part of the release process, this would be updated to reflect a precise release number in release tags/packaging, but then immediately thereafter changed to indicate something non-released - perhaps based on some implied next-release number, like `3.8.4-dev` or `4.0.0-dev`.

piskvorky:
Supersedes #2821.

- [x] Fixed a bug where `similarities.nmslib`'s `most_similar()` returned correct neighbours but with incorrect similarities.
- [x] Docstring clean up in the `similarities.nmslib` module. Follow up from https://github.com/RaRe-Technologies/gensim/issues/2820#issuecomment-623074467.
- [x] Added `similarities.nmslib` to our API documentation – it was missing before.
- [x] Renamed overly broad `similarities.index` to the more appropriate `similarities.annoy`.
- [x] Fixed deprecation warnings in Annoy integration: `The default argument for metric will be removed in future version of Annoy. Please pass metric='angular' explicitly.`
- [x] Went over our API reference page https://radimrehurek.com/gensim/apiref.html (visually, as users actually see it) and fixed a bunch of misformattings and errors. The summarizer subpackage in particular makes me sad every time I look at it.
- [x] Various fixes to code style, language, import order etc.
- [x] Got rid of `six` in places I touched (re. "removal of python2 constructs").
- [x] Changed version from `3.8.1` to `4.0.0.dev0`; fixes #2882.
- [x] Rebuilt Sphinx documentation

This PR also presented the opportunity to test migrating existing code to post #2698 code. Notes (CC @gojomo):

1. Changed failing param `iter` => `epochs`
2. Changed failing param `size` => `vector_size`
3. Changed failing `len(model.wv.vocab)` => `len(model.wv)` in Annoy.
   - @gojomo is this valid even for fastText? What's the current contract on fastText's "size of vectors" – should `len()` return the total number of admissible keys (=infinity because fastText supports OOV), or count with ngrams, or just count of "full words"…?
4. Failed with `Word2Vec' object has no attribute 'init_sims`.
   - Fixed by re-introducing `Word2Vec.init_sims()`. That method still exists in doc2vec, fasttext, keyedvectors etc – not sure why missing from word2vec? @gojomo is this an omission or done on purpose?
5. Failed with `AttributeError: module 'numpy.random' has no attribute 'default_rng'`
   - Fixed by installing a newer numpy, 1.14.2 => 1.19.1. Our `setup.py` says >=1.11.3 ought to be enough – CC @mpenkov is this expected? I thought we're testing against the **oldest** supported version of our dependencies.
   - After upgrading numpy, got another error from scipy, `No module named 'numpy.testing.decorators'`. Fixed by upgrading scipy too, to `1.5.2`. 

We should cover (at least) those in our migration guide.

The list is likely not complete; it's only what I needed to successfully run our `run_annoy.py` tutorial & rebuild the sphinx docs.


docs/src/gallery/tutorials/run_annoy.py
```diff
@@ -126,19 +128,16 @@
...
 def avg_query_time(annoy_index=None, queries=1000):
-    """
-    Average query time of a most_similar method over 1000 random queries,
-    uses annoy if given an indexer
-    """
+    """Average query time of a most_similar method over 1000 random queries."""
     total_time = 0
     for _ in range(queries):
-        rand_vec = model.wv.vectors_norm[np.random.randint(0, len(model.wv.vocab))]
+        rand_vec = model.wv.vectors_norm[np.random.randint(0, len(model.wv))]
```
gojomo, line 134: Since `vectors_norm` is no longer a persistent cache of all unit-normed vectors, but rather (for backward-compatibility) re-calculated every time it's requested, this approach will now involve a lot of redundant recalculation of the whole array. Options: (a) grab it once before the loop; (b) adapt to use `get_vector(random_index, use_norm=True)`.

gensim/models/doc2vec.py
```diff
@@ -63,25 +67,19 @@
...
 from collections import namedtuple, defaultdict
 from collections.abc import Iterable
 from timeit import default_timer
-from dataclasses import dataclass
 
+from dataclasses import dataclass
 from numpy import zeros, float32 as REAL, vstack, integer, dtype
 import numpy as np
+from six.moves import range
+from six import string_types, integer_types, itervalues
```
gojomo, line 78: These `six` imports could easily be dropped with a few other tiny changes, now that only Py3 is supported

gensim/models/fasttext.py
```diff
@@ -277,21 +277,15 @@
 
 import logging
 import os
+from collections.abc import Iterable
```
gojomo, line 280: Curious: is there a tool which reorders/regroups imports per your preferences?

gensim/models/word2vec.py
```diff
@@ -804,6 +807,23 @@ def update_weights(self):
         # do not suppress learning for already learned words
         self.wv.vectors_lockf = np.ones(1, dtype=REAL)  # 0.0 values suppress word-backprop-updates; 1.0 allows
 
+    def init_sims(self, replace=False):
```
gojomo, line 810: This method is obsolete & shouldn't exist; code that thinks it needs it (1) probably doesn't anymore; (2) should do more explicitly & specifically what it needs elsewhere or more directly, as for example on KeyedVectors, perhaps using the `unit_normalize_all()` method. 

If it's retained for backward compatibility, it should be marked deprecated & just pass-through to the similarly deprecated method on KeyedVectors, without having odd side-effects like the deletion of `syn1` here (which inexplicably has no matching deletion of `syn1neg` for negative-sampling models?). If some caller really needs those layers deleted, it'd be better for the caller to do that explicitly.

setup.py
```diff
@@ -352,7 +352,7 @@ def run(self):
 
 setup(
     name='gensim',
-    version='3.8.1',
+    version='4.0.0.dev0',
```
gojomo, line 355: Thanks!

gensim/models/word2vec.py
gojomo: Re: the question in comments - IMO `init_sims()` also shouldn't exist on `Doc2Vec` & `FastText`; its persistence in those places is just incomplete removal of obsolete stuff. Keeping it on `KeyedVectors`, with both a deprecation warning & louder warning if used with `replace=True`, is just a guidepost to help those dealing with its more-general obsolescence.

gojomo:
> 3. Changed failing `len(model.wv.vocab)` => `len(model.wv.vocab)` in Annoy.
>    
>    * @gojomo is this valid even for fastText? What's the current contract on fastText's "size of vectors" – should `len()` return the total number of admissible keys (=infinity because fastText supports OOV), or count with ngrams, or just count of "full words"…?

The reported `len()` of a `FastTextKeyedVectors` will be that of its known vocabulary, which is inherited from the more generic `KeyedVectors`, and seems a reasonable value (compared to any alternative number), as it there's a plausible case for it & it will retain the ability to use the FTKV objects as drop-ins where the KV objects will also work.

`.vocab` should not be used: it's obsolete & the only remaining accessor throws an error: https://github.com/RaRe-Technologies/gensim/blob/344c4abe26f96ef1a7299e82719b82c8a05c67a0/gensim/models/keyedvectors.py#L560 (Simulating backward-compatible behavior would be very hard & inefficient, & after the bewildering mess of different things all called 'vocab' in the prior refactoring, I preferred as little reuse of that general term as possible.) 

> 4. Failed with `Word2Vec' object has no attribute 'init_sims`.
>    
>    * Fixed by re-introducing `Word2Vec.init_sims()`. That method still exists in doc2vec, fasttext, keyedvectors etc – not sure why missing from word2vec? @gojomo is this an omission or done on purpose?

Per my [source-line-targeted-comments](https://github.com/RaRe-Technologies/gensim/issues/1623#issuecomment-363561043), I think `init_sims()` should only remain on `KeyedVectors` as a (deprecated) hint to those adapting older code - the versions on `Word2Vec`, `Doc2Vec`, `FastText` should be gone, where they persist it's just remaining obsolete-code cleanup to occur. 

> 5. Failed with `AttributeError: module 'numpy.random' has no attribute 'default_rng'`
>    
>    * Fixed by installing a newer numpy, 1.14.2 => 1.19.1. Our `setup.py` says >=1.11.3 ought to be enough – CC @mpenkov is this expected? I thought we're testing against the **oldest** supported version of our dependencies.
>    * After upgrading numpy, got another error from scipy, `No module named 'numpy.testing.decorators'`. Fixed by upgrading scipy too, to `1.5.2`.

The numpy-performance-regression in the legacy PRNG led me to update code to use their new `default_rng` API, but I didn't update prerequisites to reflect this (and as you note, testing didn't show any errors). It looks like this `default_rng` API has been there since numpy 1.17.0 (released a year ago).

I personally feel it's OK (and even beneficial) for fresh gensim releases to require fairly-recent releases of those libraries that are more 'central'/large/many-contributor, like `numpy`, unless there's a really strong/specific reason to hold back. Lagging projects that may need older `numpy`/etc should also use older `gensim`! (And, if there were a strong economic reason some project needs a challenging configuration – such as bleeding-edge gensim but years-old `numpy` – the cost of creating/maintaining that idiosyncratic-configuration should fall to that project.)

gojomo:
An aside on approximate-nearest-neighbor libraries: there's a recent new option from Google Research – [ScaNN](https://github.com/google-research/google-research/tree/master/scann) – which they show as far outperforming ANNOY & FB Faiss.

docs/src/gallery/tutorials/run_annoy.py
piskvorky: Oh that's sneaky. Isn't it better to fail outright, rather than do an expensive operation silently, where existing code expects a cheap dict lookup?

piskvorky: Or, we want to keep backward comp, norm just one vector… I haven't checked the code but why would we recalc the whole array for a single `__getitem__`?

gensim/models/doc2vec.py
piskvorky: Right. I though I did it, an omission. I'll push a fix.

gensim/models/fasttext.py
piskvorky: `black`, probably. But my preferences = standard PEP8: blocks of "built-in" - "3rd party" - "local" imports (in that order).

piskvorky: Plus my editor (sublime + pylint) highlights mixed up imports, even though I still have to reorder them manually (if I choose to).

gensim/models/word2vec.py
piskvorky: Alright, I'll remove it. Let's keep this in mind for the migration guide.
Can you craft a more noob-friendly error message for KeyedVectors?

piskvorky:
Yeah thanks. I think I created a ticket for ScanNN when it came out, although I cannot seem to find it now.

piskvorky:
> I personally feel it's OK (and even beneficial) for fresh gensim releases to require fairly-recent releases of those libraries

Yes, upgrading numpy is not a problem. I'd expect our tests to fail though. I'm pretty sure we were testing against the oldest lib, explicitly, but with all the recent changes to the testing framework (azure etc) it's possible that fell through the cracks. We should put it back if possible, @mpenkov WDYT?

piskvorky:
One more idea: once we have the migration guide in place, use it on the *other* tutorials (beyond `run_annoy.py`), to validate the guide is complete.

The expectation is that all should run smoothly, no errors or deprecation warnings. The added value is that we refresh the generated notebooks & rst, since I expect many of the outputs & timings will change with 4.0.0.

docs/src/gallery/tutorials/run_annoy.py
gojomo: We're *not* recalculating for any specific request of a single vector - that does the minimum, and would be a quickie option ((b) above) to improve this code. 

But if a caller asks for `vectors_norm`. it's now calculated & returned (without being cached). Arguably, `vectors_norm` was barely even part of the public API, as it wasn't guaranteed to even exist if a caller hadn't triggered `init_sims()` yet. I woudn't expect the problematic pattern here, requesting `vectors_norm` repeatedly, to be very common, especially when callers who just need single vectors have the `get_vector(..., use_norm=True)` option. 

I suppose we could (a) cache it once requested, to be more like the old behavior, but then we have to add more logic for properly handling other invalidations of this cache elsewhere, and explicitly discarding it at times (like serialization); or (b) eliminate the backward-compatible `@getter` property, either leaving it blank (to generate cryptic error), or stub an explanatory error message. In that latter case, a new distinctly-name utility method could be added, `calculate_unit_normed_vectors()`, that avoids creating the impression this is a simple access of an already-existing property.

gensim/models/word2vec.py
gojomo: I don't think noobs will have been likely to accessing `.vocab` directly, unless it's in cut & pasted code. And other than the existing pointers to the names of new properties/methods offering equivalent functionality, I don't have ideas for more noob-friendly verbiage to put in the actual error text. (I don't think they'll care about the rationales, and I don't have enough insight into why they might be calling it to include any mini-tutorials on how to update their code.)

docs/src/gallery/tutorials/run_annoy.py
piskvorky: I'm in favour or removing it = hard fail with a clear message (again, noob-friendly = prescription for "what to do next").

-1 on pretending we're a simple dict, when there's a substantial calculation happening behind the scenes.

gensim/models/word2vec.py
piskvorky: Hmm. I hoped it might be something as simple as `model.init_sims() no longer exists; call model.kw.init_sims() instead`? 
What's the complication?

piskvorky:
@mpenkov I'll go ahead and merge this, because I'm worried about getting out of sync again. But please still review when you have a moment.

gensim/models/word2vec.py
gojomo: Ah, for some reason I got this thread confused with the `.vocab` discussion, & didn't see a path to any message better than the (already-4-lines-of-text) message accompanying the `NotImplementedError` there. Yes, if you're OK with a hard-fail on uses of `init_sims()`, such a message makes sense. And it could be on `KeyedVectors`, too, pointing users of any variant of that method to the info that (1) they probably no longer need to do init_sims; (2) single normed vectors are easy to get via `get_vector()`; (3) if they want to get a big array of all unit-normed vectors, some (to be added distinctively-named) method is available; (4) if they want to ensure norms are fresh after performing atypical out-of-band vector tampering,  call `fill_norms()`.

docs/src/gallery/tutorials/run_annoy.py
gojomo: Sure, `.vectors_norm` should be a hard-fail with helpful error message, similar to the one for `init_sims()`. I can add these for both as part of other touch-up work, such as that in #2892.

gensim/models/word2vec.py
piskvorky: Yes, that's exactly it.

I thought we'd put that info directly into the error message, but perhaps a link to some permanent URL (Github wiki?) gives us more flexibility, at the cost of decreased user convenience, for very chatty errors.

docs/src/gallery/tutorials/run_annoy.py
piskvorky: Alright. I'll add some tentative fix here (so I don't merge "known to be bad" code), merge this PR, and then let's continue with this in another PR, such as #2892.

piskvorky: @gojomo pushed now. Let me know if OK and I'll merge this PR, to be continued elsewhere.

gensim/models/keyedvectors.py
```diff
@@ -519,9 +518,17 @@ def rank(self, key1, key2):
...
-    # backward compatibility; some would be annotated `@deprecated` if that stacked with @property/.setter
     @property
     def vectors_norm(self):
+        raise ValueError(
+            "The vectors_norm attribute became a get_normed_vectors() method in Gensim 4.0.0. "
+            "See https://github.com/RaRe-Technologies/gensim/wiki/Migrating-from-Gensim-3.x-to-4#init_sims"
+        )
+
+    def get_normed_vectors(self):
+        # TODO: what's the way for users to get from a matrix index (integer) to the
```
piskvorky, line 529: @gojomo FYI

gensim/models/keyedvectors.py
```diff
@@ -1454,7 +1464,7 @@ def save_word2vec_format(self, fname, fvocab=None, binary=False, total_vec=None,
             if not (i == val):
                 break
             index_id_count += 1
-        keys_to_write = chain(range(0, index_id_count), store_order_vocab_keys)
+        keys_to_write = itertools.chain(range(0, index_id_count), store_order_vocab_keys)
```
gojomo, line 1469: Elsewhere, I think I've followed a growing convention I've seen elsewhere to consider `itertools` as a package with a 'well known standard abbreviation' of `it`, so `import itertools as it`. I'd prefer it if gensim's project guide specifically recommended this for a small set of packages: `np`, `pd`, `it`, and maybe `pkl`, as it matches wider practices & simplifies import management. (That's especially true for numpy & itertools, where the prior practice of naming exactly which functions were used meant a jump-around just-typing burden of changing the imports every time some single function was used or removed-from-use.) But whatever project style can be unambiguously declared somewhere authoritative is OK.

gensim/models/fasttext.py
```diff
@@ -1264,14 +1263,14 @@ def save(self, *args, **kwargs):
         kwargs['ignore'] = kwargs.get('ignore', ignore_attrs)
         super(FastTextKeyedVectors, self).save(*args, **kwargs)
 
-    def get_vector(self, word, use_norm=False):
+    def get_vector(self, word, norm=False):
```
gojomo, line 1266: `use_norm` was an unfortunate choice of parameter-name which I believe dated back to #1777,  & I considered changing it, but didn't want to have to change all uses elsewhere & add another compatibility-burden on users. 

But if you're entertaining breaking changes:

(1) not sure `norm` is a good name either - it makes it sound like you're requesting the vector's (scalar) norm/length;
(2) even supporting this via an optional extra parameter on a dual-use method strikes me as silly/opaque, too. Instead, have `__getitem__` be the implementation of the default behavior, add a separate explicitly-named method for the unit-normed vector, like perhaps `get_unit_vector(key_or_int)`, no optional parameters;
(3) more aggressively, maybe let users **toggle** the model to give either unit-vecs, or raw-vecs, as the default `__getitem__` behavior, but have both `get_unit_vector()` and `get_raw_vector()` alt options for when a caller needs to specify. Then the desire you've expressed elsewhere to have unit-normed access, via plain dict-bracket-accesses, is available when toggled-on, but the raw data is never destroyed, preventing a class of errors/regrets. I'd put this thought aside as a somewhat complicated "maybe in the future" idea, but if you're OK to break `get_vector()` signature, & want to prioritize unit-normed access, maybe it's worth making that change now.

gensim/models/keyedvectors.py
piskvorky: Yes. The idea here was to fuse `unit_normalize_all()` with `get_normed_vectors()`: have a single function which returns a new KV instance, normalized.

I think that's cleaner than several in-situ state modifiers/getters, but it's true the use-case is hypothetical.

gensim/models/keyedvectors.py
piskvorky: No preference from me. This one I changed merely because I saw two side-by-side imports `import itertools; from itertools import chain`, which looked weird, so I merged them into a single `import itertools`.

gensim/models/fasttext.py
piskvorky: TBH I find the idea of producing a new standalone (normed) KV instance cleaner than toggles.

I'll merge here lest we get a monster PR again, let's revisit this in one of your other PRs.

gensim/models/keyedvectors.py
piskvorky: > get_vector(key_or_int, use_norm=True) returns an individual normed vector, by key or raw integer slot

Hmm, I don't see that in the code. How does the "raw integer slot" option work?

piskvorky:
@mpenkov some of the tests unrelated to this PR (I think) started failing: https://travis-ci.org/github/RaRe-Technologies/gensim/jobs/713448950#L862

I disabled them for now and opened a ticket to look into it: #2908.

gensim/models/keyedvectors.py
piskvorky: I removed the `int` option from the docstring, because I don't think it exists.

mpenkov:
> * CC @mpenkov is this expected? I thought we're testing against the **oldest** supported version of our dependencies.

No, that's not the case. When you `pip install gensim` in a clean environment, pip will pick the latest versions of the packages to install. All the CI builds occur in a clean environment, so in this case the pin was too loose.

mpenkov:
> Also, how do you check the Circle CI results? Clicking on the "Details" link here brings up a page which asks me to sign in with Github, with read&write access to my repos, which is not happening. How do you view the actual CI results?

@piskvorky For me, I just click on "Details" and I can see the results for the relevant build.

piskvorky:
Weird. Somebody else will have to inspect the Circle CI results then, I cannot access them.

PR #2901, opened 2020-07-27
Fix AttributeError in WikiCorpus

### LINKED ISSUE #2744: AttributeError: 'WikiCorpus' object has no attribute 'input'
#### AttributeError: 'WikiCorpus' object has no attribute 'input'

I am using WikiCorpus to read and process the wikipedia dump. However when I try to iterate  over  getstream() output, I get the error that there is no attribute input. Indeed, there is no input attribute but there is a fname attribute.

```python
path = "backend/de_wiki/dewiki-latest-pages-articles.xml.bz2"
wiki = WikiCorpus(path)
for stream in wiki.getstream():
    print(stream)
    break
```

On the other hand, this fixes the problem, but I guess would be better if it gets fixed in the source code if the attribute missing is the problem.

```python
setattr(wiki, "input", wiki.fname)
```

#### Versions

Please provide the output of:

```python
Linux-x86_64-with-debian-9.11
Python 3.7.0 (default, Oct  9 2018, 10:31:47) 
[GCC 7.3.0]
NumPy 1.17.3
SciPy 1.3.2
gensim 3.8.0
FAST_VERSION 1
```

jenishah:
Fix for : #2744 

`WikiCorpus` subclasses `TextCorpus`.
Renaming the attribute `fname` of  `WikiCorpus` to `input` fixes the `AttributeError`: 'WikiCorpus' object has no attribute 'input'`

### DIFF
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -620,7 +620,7 @@ class WikiCorpus(TextCorpus):
         Unless a dictionary is provided, this scans the corpus once, to determine its vocabulary.
 
         """
-        self.input = fname
+        self.fname = fname
         self.filter_namespaces = filter_namespaces
         self.filter_articles = filter_articles
         self.metadata = False
@@ -638,6 +638,10 @@ class WikiCorpus(TextCorpus):
             self.dictionary = Dictionary(self.get_texts())
         else:
             self.dictionary = dictionary
+
+    @property
+    def input(self):
+        return self.fname
 
     def get_texts(self):
         """Iterate over the dump, yielding a list of tokens for each article that passed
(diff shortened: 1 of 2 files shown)


piskvorky:
Thanks! @mpenkov is overwhelmed right now but he'll be with you eventually, to review & merge :)

gensim/corpora/wikicorpus.py
```diff
@@ -620,7 +620,7 @@ def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), diction
         Unless a dictionary is provided, this scans the corpus once, to determine its vocabulary.
 
         """
-        self.fname = fname
+        self.input = fname
```
piskvorky, line 623: Should we rename the param too? That would break API (unless we keep both `input` and `fname`) but maybe a little cleaner. @mpenkov WDYT?

mpenkov:
Looks good to me. Thank you for the contribution, @jenishah !

jenishah:
> Looks good to me. Thank you for the contribution, @jenishah !

It was a good learning for me. Thanks to you :)

Issue #2894, opened 2020-07-24
Gensim Doc2Vec model Segmentation Faulting for Large Corpus


mohsin-ashraf:
#### Problem description

What are you trying to achieve? 
I was trying to train a doc2vec model on a corpus of 10M (10 millions) documents for my dataset roughly having a length of ~5000 words on average. The idea was to generate a semantic search index on these documents using the doc2vec model.

What is the expected result? 
I was expecting it to be completed successfully as I tested for the smaller dataset. On a smaller dataset of size 100K documents, it worked fine and I was able to do basic benchmarking for the search index, which successfully passed the criteria.

What are you seeing instead?
When I started training on the 10M dataset. After building the vocabulary the training of the doc2vec model stoped and resulted in **Segmentation Fault**.

#### Steps/code/corpus to reproduce

Include full tracebacks, logs, and datasets if necessary. Please keep the examples minimal ("minimal reproducible example").
Here is the [link](https://github.com/mohsin-ashraf/reproduce-gensim-bug) to the example to reproduce it. It uses the following libraries (unfortunately I could not make a virtual env. due to some issues.) other than mentioned below.
[RandomWords](https://pypi.org/project/RandomWords/)

Attached is the logging file.
[logging_progress.log](https://github.com/RaRe-Technologies/gensim/files/4970118/logging_progress.log)

#### Versions

Please provide the output of:

```python
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import gensim; print("gensim", gensim.__version__)
from gensim.models import word2vec;print("FAST_VERSION", word2vec.FAST_VERSION)
```

```
Linux-4.15.0-45-generic-x86_64-with-Ubuntu-18.04-bionic
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
NumPy 1.19.0
SciPy 1.5.1
gensim 3.8.3
FAST_VERSION 1
```

[Here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/gensim/AK4ydLclz5g/bQsKVEUKCQAJ) is the google group thread for a detailed discussion

mohsin-ashraf:
For pinning down the exact number over which it always segmentations faults and below which it always works, I am trying some other experiments as well. For now, it works fine for 5M documents and gives the expected results.

[...]

PR #2902, opened 2020-07-27
Ref #2894 fixing segmentation fault for doc2vec

mohsin-ashraf:
PR for #2894 I was unable to install `test` scripts (`pip install -e .[test]`) and could not test it on my PC. Please review the PR and let me know if I need to change something before its gets meged.

**Also I see some other potential script that could have the same problem, for example [word2vec_corpusfile.pyx](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec_corpusfile.pyx)**

### DIFF
--- a/gensim/models/doc2vec_corpusfile.pyx
+++ b/gensim/models/doc2vec_corpusfile.pyx
@@ -50,20 +50,20 @@ from gensim.models.word2vec_corpusfile cimport (
 
 DEF MAX_DOCUMENT_LEN = 10000
 
-cdef int ONE = 1
+cdef long long ONE = 1
 cdef REAL_t ONEF = <REAL_t>1.0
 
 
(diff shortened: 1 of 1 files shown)


mohsin-ashraf:
Any instructions on what I can do to resolve these circle ci failures?

piskvorky:
Your modified file doesn't even compile, see https://dev.azure.com/rare-technologies/gensim-ci/_build/results?buildId=202&view=logs&j=3761f2ce-4b39-5e3c-4414-067c73e601fb&t=85f7bcdc-b957-5525-0201-363867e07f0c&l=350 . You cannot change the types *that* mechanically, the signatures and data structures need to match.

If you're not familiar with C code it's probably best to wait until we get it. Although I cannot promise any ETA, we're busy with preparing for 4.0.0 now.

PR #2904, opened 2020-07-28
Unify LDA hyperparameter documentation

### LINKED ISSUE #2903: "Wrong" default parameter for [sum]alpha for LdaMallet
The [original Mallet](https://github.com/mimno/Mallet/blob/3bda6c73b2ff90d2e91a35dc6e9d604fbea8248e/src/cc/mallet/topics/tui/TopicTrainer.java#L152) default parameters are 
--alpha DECIMAL
  SumAlpha parameter: sum over topics of smoothing over doc-topic distributions. [alpha_k = [this value] / [num topics]](https://github.com/mimno/Mallet/blob/12487de1aa6433bdcf5af0ee0a17b368e64c7acf/src/cc/mallet/topics/ParallelTopicModel.java#L181)
  Default is 5.0
--beta DECIMAL
  Beta parameter: smoothing parameter for each topic-word. beta_w = [this value]
  Default is 0.01

The default value provided by gensim is confusing, as could be seen here:
https://stackoverflow.com/questions/52989568/lda-gensim-mallet-documentation-on-alpha

https://github.com/RaRe-Technologies/gensim/blob/27bbb7015dc6bbe02e00bb1853e7952ac13e7fe0/gensim/models/wrappers/ldamallet.py#L78

jonaschn:
Make documentation for alpha & eta consistent for AuthorTopicModel, LdaModel and LdaMulticore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Changes
 
 - LsiModel: Only log top words that actually exist in the dictionary (PR [#3091](https://github.com/RaRe-Technologies/gensim/pull/3091), [@kmurphy4](https://github.com/kmurphy4))
 
+### Documentation
+
+- Fix hyperlink for doc2vec tutorial (PR [#3123](https://github.com/RaRe-Technologies/gensim/pull/3123), [@AdityaSoni19031997](https://github.com/AdityaSoni19031997))
+
 ## 4.0.1, 2021-04-01
 
 Bugfix release to address issues with Wheels on Windows:
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -21,7 +21,7 @@ and Phrases and their Compositionality. In Proceedings of NIPS, 2013"
 <https://papers.nips.cc/paper/5021-distributed-representations-of-words-and-phrases-and-their-compositionality.pdf>`_.
 
 For a usage example, see the `Doc2vec tutorial
-<https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/doc2vec-lee.ipynb>`_.
+<https://radimrehurek.com/gensim/auto_examples/tutorials/run_doc2vec_lee.html#sphx-glr-auto-examples-tutorials-run-doc2vec-lee-py>`_.
 
 **Make sure you have a C compiler before installing Gensim, to use the optimized doc2vec routines** (70x speedup
 compared to plain NumPy implementation, https://rare-technologies.com/parallelizing-word2vec-in-python/).


gensim/models/wrappers/ldamallet.py
```diff
@@ -118,7 +126,16 @@ def __init__(self, mallet_path, corpus=None, num_topics=100, alpha=50, id2word=N
             raise ValueError("cannot compute LDA over an empty collection (no terms)")
         self.num_topics = num_topics
         self.topic_threshold = topic_threshold
-        self.alpha = alpha
+        if alpha is not None:
+            # Mallet's alpha parameter is the SumAlpha parameter
+            # (sum over topics of smoothing over doc-topic distributions. alpha_k = [SumAlpha] / [num topics])
+            # To keep the same interface with gensim, calculate the SumAlpha:
+            self.alpha = alpha * num_topics
```
piskvorky, line 133: I'd prefer to do it the other way around: keep `alpha` gensim-style internally, and convert to SumAlpha only at the point of Mallet API = when calling Mallet.

I think it's safer wrt expectations of gensim users. Is there a good reason to do it the other way around?

gensim/models/wrappers/ldamallet.py
```diff
@@ -87,8 +88,15 @@ def __init__(self, mallet_path, corpus=None, num_topics=100, alpha=50, id2word=N
...
-            Alpha parameter of LDA.
+        alpha : float, optional
+            Alpha parameter of LDA (scalar for a symmetric prior over document-topic probability).
+            Default: 5.0 / num_topics
+        use_symmetric_alpha : bool, optional
+            Only optimize the concentration parameter of the prior over document-topic distributions.
+            (This may reduce the number of very small, poorly estimated topics,
+            but may disperse common words over several topics.)
+        eta : float, optional
+            Eta parameter of LDA (scalar for a symmetric prior over topic-word probability).
```
piskvorky, line 99: Where does the `0.01` default come from?

jonaschn: It is the Mallet default behaviour (see [here](https://github.com/mimno/Mallet/blob/024a64ff9caf3f54586c5f70399da001f208305b/src/cc/mallet/topics/ParallelTopicModel.java#L93)):
`public static final double DEFAULT_BETA = 0.01;`

piskvorky: I see. And this one is not a "sum", unlike alpha?

jonaschn: You are right, as mentioned here #2903
Beta parameter: smoothing parameter for each topic-word. beta_w = [this value]

piskvorky: Alright. So please change `alpha` to be stored internally as non-sum too (convert to sum only when needed) and we can merge. Thanks.

gensim/models/wrappers/ldamallet.py
jonaschn: No, you are totally right. 
It is indeed safer for gensim users.
I addressed your feedback and substantially improved the documentation.

jonaschn:
@piskvorky I encorporated your feedback.

Unfortunately, the documentation build is failing. I am not familiar with sphinx and don't know how to fix this warning.

```
Warning, treated as error:
/home/circleci/gensim/gensim/models/atmodel.py:docstring of gensim.models.atmodel.AuthorTopicModel:29:Block quote ends without a blank line; unexpected unindent.
make: *** [Makefile:37: html] Error 2
make: Leaving directory '/home/circleci/gensim/docs/src'
ERROR: InvocationError for command /usr/bin/make -C docs/src clean html (exited with code 2)
```

jonaschn:
As gensim dropped the support for the Mallet wrapper, I removed this part of the PR.
You can find the improved and fixed (!) version of the Mallet wrapper for LDA and Labeled LDA here:
https://github.com/jonaschn/gensim/tree/labeled-lda

Issue #2840, opened 2020-05-14
Implement position-dependent weighting to fastText


Witiko:
In Section 2.2 of the [2017 “Advances” paper by Mikolov et al.][advances], a position-dependent weighting is introduced to the input context vector computation in the fastText CBOW unsupervised model. On [Common Crawl][], this extension led to 5 point improvement in the total accuracy on the English word analogy task.

I am proposing to add this extension to Gensim's implementation of fastText. Despite the usefulness of the extension, it has not been introduced to facebook's code base (https://github.com/facebookresearch/fastText/issues/445), so Gensim could be the first publicly available implementation.

In Table 4, row CBOW of the [2018 “Learning” paper by Mikolov et al.][learning] shows the word analogy task results of CBOW fastText models for 10 languages trained on Wikipedia. The CBOW fastText models use the position-dependent weighting extension and the default parameters described in Section 4.3 of the [2017 “Enriching” paper by Bojanowski et al.][enriching]: hash table bucket size 2*10^6, 5 epochs, 300 vector dimensions, negative sampling loss with 5 negative samples, initial learning rate 0.05, sampling threshold 10^-4, and window size 5. The correctness of our implementation can be checked by reproducing the results.

 [advances]: https://arxiv.org/pdf/1712.09405.pdf#page=2
 [learning]: https://arxiv.org/pdf/1802.06893.pdf#page=4
 [common crawl]: https://commoncrawl.org/
 [enriching]: https://arxiv.org/pdf/1607.04606.pdf#page=4

piskvorky:
Sure, why not. The motivation is clear and shouldn't disrupt users.

I haven't looked into these papers (or perhaps forgot already) but what is the expected impact on performance (speed)? Why **wouldn't** you want to use this weighting, as a user?

Witiko:
> What is the expected impact on performance (speed) and code complexity?

For all positions in the context window ([-c, ..., −1, 1, ..., c]), you need to maintain a separate 2c × D weight matrix with *position vectors*. The position vectors need to be stored with the model. For each forward pass of the CBOW, you need to compute the average of the Hadamard (entrywise) product between the context word vectors and the context word position vectors instead of just the average of the context word vectors. For each backward pass, you also compute and apply the gradient to the position vectors.

Since 2c is usually small (10 position vectors), the impact on both speed and model storage size should be negligible.

> Why wouldn't you want to use this weighting, as a user?

According to the papers, the weighting should be a net improvement.

[...]

Issue #299, opened 2015-03-03
LdaMallet breaks with Large number of documents (>400k)


chiragnagpal:
I have been trying to use the Mallet wrapper in gensim for topic modelling tweets, however, when the corpus has more than 400k documents, LdaMallet breaks

piskvorky:
How do you know? Can you specify "breaks", ideally with a concrete log + error message?

chiragnagpal:
![screenshot from 2015-03-04 08 09 07](https://cloud.githubusercontent.com/assets/4943802/6477166/08e410aa-c246-11e4-987a-9584cc55250a.png)

I initially thought of it to be a memory issue, so increased the size of /tmp and also tried changing the prefix location to a physical disk. 

That helps, as I am able to train my model, however in the step, wherein gensim tries to extract the topics from the MALLET file, it again shows an error, 'too many values to unpack'

![screenshot from 2015-03-04 09 10 31](https://cloud.githubusercontent.com/assets/4943802/6477871/75190052-c24e-11e4-97eb-a5db44067109.png)

cscorley:
Just to cross post, there is also a thread on this happening on the mailing list: https://groups.google.com/forum/#!topic/gensim/uEZiu9hbMYg

I'll be looking into fixing this problem this weekend!

piskvorky:
Thanks Christopher!

cscorley:
I think I nailed the problem in the mailing list, have a look. It likely comes down to tokens having whitespace in them (e.g., tab, non-breaking space, etc). Either make sure you're decoding the source data and tokenizing it correctly first (e.g., it could be in encoded in something not UTF-8), or don't build bigrams with spaces! :)

However, the fix, on Gensim's end, is to only split on the single space character only, and not any whitespace.

cscorley:
If you don't mind, @chiragnagpal, try using the new `gensim.models.wrappers.LdaMallet` on develop branch. Note that it's been moved to a new submodule. I believe it should solve the problem of reading in the file.

chiragnagpal:
@cscorley  thnx man! ur a lifesaver! I used a better tokeniser and now its working like a charm!

cscorley:
Great! Glad to hear its working for you now :beers:

Going to close this issue.

PR #2905, opened 2020-07-28
Implement position-dependent weighting to fastText

Witiko:
This PR adds support for position-dependent weighting for fastText CBOW with negative sampling, as discussed in issue #2840. The `FastText` constructor receives a new boolean parameter `pdw`, which enables the position-dependent weighting when set to 1. The changes currently aren't optimized for speed and aim at the smallest possible change.


Witiko:
The code currently segfaults immediately after building the vocab. @gojomo, I would appreciate a quick glance over the changes to see if there is something obvious I have missed before I start debugging the Cython code.

gojomo:
I don't feel familiar enough with what *should* be happening to proofread the kinds of indexing-operations that are likely at fault here, so I'd suggest getting right to the deeper debugging. 

A couple style comments: 

* opaque abbreviations like `pdw` are generally disfavored, even though some standard-practices from elsewhere (eg `sg` & `hs`) are tolerated. If FB's own FastText used `pdw` as a parameter, maybe it could be used here for consistency, but otherwise something more descriptive would be preferred, especially at the Python/public-API level. Maybe either the full expansion, 'position_dependent_weighting`, or something more-imperative like `learn_position_weights`?

* `l` as a loop index is disfavored for confusion with `i`/etc. (I even think I've seen the project auto-testing flake8 style rules reject it, though it seems to have passed in your cython code.)

Sorry my review isn't more substantive.

Witiko:
Thank you for the review.

> I don't feel familiar enough with what _should_ be happening to proofread the kinds of indexing-operations that are likely at fault here, so I'd suggest getting right to the deeper debugging with `gdb` or `cygdb`, etc. (And I've often hoped some simple errors could be caught, debugging cython code, by changing the `boundscheck` to on until debugged... but that's never found a real issue for me. Still, maybe worth a quick try.)

Since I am not too experienced with `cygdb` and the diff from `develop` is currently minimal, I am thinking it may be faster to debug via `printf` first. Indexing is likely at fault, but `c.syn0_positions` may also point to the wrong thing, I'll need to investigate.

> A couple style comments:
> 
>   * opaque abbreviations like `pdw` are generally disfavored, even though some standard-practices from elsewhere (eg `sg` & `hs`) are tolerated. If FB's own FastText used `pdw` as a parameter, maybe it could be used here for consistency, but otherwise something more descriptive would be preferred, especially at the Python/public-API level. Maybe either the full expansion, `position_dependent_weighting`, or something more-imperative like `learn_position_weights`?

I was thinking that if this parameter ever appear's in FB's code base, `pdw` is what it would have been named. Other than that, I agree that a human-readable parameter name would be preferred.

>   * `l` as a loop index is disfavored for confusion with `i`/etc. (I even think I've seen the project auto-testing flake8 style rules reject it, though it seems to have passed in your cython code.)

Good point. Depending on the font, `l` can also be mistaken for `1` in array indexing, so I suppose it's best to stay away.

> Sorry my review isn't more substantive.

It's plenty substantive, thanks. I have experience with C and C++ programming, but I am new to Cython.

Witiko:
The issue was that I assumed that `c.vocab_lockf[c.indexes[m] % c.vocab_lockf_len]` and `c.ngrams_lockf[c.subwords_idx[m][o] % c.ngrams_lockf_len]` were vectors, while they were just scalars.

I updated the code, renamed the parameter `pdw` to `position_dependent_weights`, and the variable `l` to `o` as suggested by @gojomo. Currently, I am training a model to see if I can reproduce the experimental results from the original paper.

Witiko:
Note to self: Dense positional weights are a poor fit for Hogwild. Single worker may be required for optimal rate of convergence.

Witiko:
It also seems that unless we reduce the gradients to the positional vectors, both positional and word vectors tend to NaN. I am currently exploring this issue using positional weight scalars as opposed to positional weight vectors, because the former is too computationally to quickly iterate.

When I update positional weights by their full gradients (44e1d836), we end up with both positional weights and word vectors full of NaNs. When I update positional weights by a mean gradient across all dimensions (32331151), meaningful positional weight vectors are trained (`[1.07, 1.05, 1.10, 1.28, 2.29, 2.38, 1.33, 1.08, 1.00, 0.99]`), although the accuracy on the word analogy task is not improved (70.88% as opposed to 72.19 % without position-dependent weighting).

gojomo:
Analogy scores aren't everything - I've seen cases where a word2vec model was measurably better on the real classification task for which it was needed, even though it was worse for analogies. But, I do see the original PDW paper claimed an improvement on analogies. 

However, it looks like the only specific PDW-vs-non-PDW results they report (Table 2 in Section 4) are:

* on the giant 630B words common-crawl dataset - requiring many days of training. (Are you using a similarly large dataset?)
* after **6 passes** of Phrase-combination are applied – so while after phrasing, there's a PDW improvement, there's no PDW-vs-non-PDW comparison without phrasing. (Perhaps its biggest benefit is somewhat synergistic with phrase-combinations?)
* using `window=5` in the non-PDW cases, but `window=15` in the PDW case. To me that raises the question of whether the window-enlargements might be a contributor (or complete cause) of any evaluation improvements

If you're testing just a single scalar per position, rather than a vector (that might allow dimensions of the vectors to spcialize in 'near' or 'far' influences), it strike me you may not be testing the same effect at all, and I wouldn't necessarily expect the same results. (And, an analogy difference of just ~1.31% might just be run-to-run jitter.) 

Looking at the weights your experiment learned, it seems only the 'next-door' words have very-different weights (>2x). Perhaps if there is a benefit, simply adopting some simple weighting that's more-near-word-heavy than the linear-dropping via the `reduced_windows` effect, will offer the benefit without a full optimization on the weights? (Also: it appears your PR doesn't replace the `reduced_windows` effect, so these weights are *additive to* the effect of such window-clipping. I can't tell if that was the intent of the original paper, or not.)

Witiko:
> However, it looks like the only specific PDW-vs-non-PDW results they report (Table 2 in Section 4) are:
> 
> - on the giant 630B words common-crawl dataset – requiring many days of training. (Are you using a similarly large dataset?)

Not at the moment. I am using English Wikipedia as the dev set, so that I can quickly iterate over different implementations. After the implementation has been finished, I can evaluate on full Common Crawl.

> - after **6 passes** of Phrase-combination are applied – so while after phrasing, there's a PDW improvement, there's no PDW-vs-non-PDW comparison without phrasing. (Perhaps its biggest benefit is somewhat synergistic with phrase-combinations?)

I would expect the weighting to offer similar benefits with and without phrasing, but we should also apply the phraser for the final evaluation, of course.

> - using `window=5` in the non-PDW cases, but `window=15` in the PDW case. To me that raises the question of whether the window-enlargements might be a contributor (or complete cause) of any evaluation improvements.

I should try using `window=15` for both PDW and non-PDW, so that we can verify your hypothesis.

> If you're testing just a single scalar per position, rather than a vector (that might allow dimensions of the vectors to spcialize in 'near' or 'far' influences), it strike me you may not be testing the same effect at all and I wouldn't necessarily expect the same results.

I am testing just a single scalar per position, since the full position-dependent weighting is currently too slow to quickly iterate over different implementations.

> an analogy difference of just ~1.31% might just be run-to-run jitter.

True, but that still means that there is no improvement at best, i.e. using just a single scalar per position seems to offer no benefit over no position-dependent weighting (with `window=5`).

> Looking at the weights your experiment learned, it seems only the 'next-door' words have very-different weights (>2x). Perhaps if there is a benefit, simply adopting some simple weighting that's more-near-word-heavy than the linear-dropping via the reduced_windows effect, will offer the benefit without a full optimization on the weights? (Also: it appears your PR doesn't replace the reduced_windows effect, so these weights are additive to the effect of such window-clipping. I can't tell if that was the intent of the original paper, or not.)

I believe the reduced windows have been core property of both word2vec and fastText since [the 2013 paper](https://arxiv.org/pdf/1301.3781.pdf#page=4).

Witiko:
For speedy results, I have reduced the dev set to the 2017 English Wikipedia (3.5B words, ca 6% of Common Crawl) over *just a single epoch* instead of five as the dev set and I measured the speed and the English word analogy task total accuracies for different Git branches and configurations. Take the results with a grain of salt, your mileage may vary with a larger training set. Full results are available in the [attached Jupyter notebook][notebook].

The results are reported for the `develop` branch before #2698, the `develop` branch after #2698, the [`positional-dependent-scalars`][scalar] branch at commit bd2596aa, which implements position-dependent scalar weights, and the [`positional-dependent-vectors`][vector] branch at commit 56376990, which implements position-dependent weight vectors as described by [the 2017 “Advances” paper][advances].

### Training Speed and Word Analogy Task Accuracies

Below are the English word analogy task total accuracies of the resulting word vectors:

|  | branch `develop`, before #2698 | branch `develop`, after #2698 | branch [`positional-dependent-scalars`][scalar], `pdw=0` | branch [`positional-dependent-vectors`][vector], `pdw=0` | branch [`positional-dependent-scalars`][scalar], `pdw=1` | branch [`positional-dependent-vectors`][vector], `pdw=1` |
|--|--|--|--|--|--|--|
| `window=5`  | 65.63% | 65.40% | 66.27% | 65.95% | 64.95% |   67.96%   |
| `window=15` | 31.88% | 60.98% | 61.67% | 61.67% | 62.22% | **69.75%** |

Below are the training durations on a dedicated machine with 252G RAM, 32 Intel Xeon X7560 2.26 GHz CPU cores, and OpenBLAS:

|  | branch `develop`, before #2698 | branch `develop`, after #2698 | branch [`positional-dependent-scalars`][scalar], `pdw=0` | branch [`positional-dependent-vectors`][vector], `pdw=0` | branch [`positional-dependent-scalars`][scalar], `pdw=1` | branch [`positional-dependent-vectors`][vector], `pdw=1` |
|--|--|--|--|--|--|--|
| `window=5`  | 2h 05m 55s | 1h 58m 32s | 2h 06m 02s | 1h 59m 06s | 2h 03m 21s |    8h 29m 44s   |
| `window=15` | 2h 26m 22s | 2h 23m 56s | 2h 20m 25s | 2h 25m 33s | 2h 28m 23s | **11h 05m 16s** |

The first four columns on the left are a *successful sanity check*: both accuracy and speed seem comparable between branch `develop` both before and after #2698, and with both the [`positional-dependent-scalars`][scalar] and [`positional-dependent-vectors`][vector] branches when position-dependent weighting is switched off. Interstingly, branch `develop` before #2698 achieves accuracy of only 31.88% with `window_size=15`, whereas all the other branches achieve accuracy of >60%; any idea what the cause could be, @gojomo? I trained three different models to make sure, but all received comparable accuracies: 31.27%, 32.40%, and 31.88%.

The fifth column shows that *positional weight vectors can not be replaced with positional weight scalars*. Although the speed is competitive with branch `develop`, training positional weight scalars seems to offer no benefit to the accuracy.

The sixth column shows that *our implementation of positional weight vectors is correct*, since it leads to a ca 5% improvement in accuracy as reported in [the 2017 “Advances” paper by Mikolov et al.][advances]. This improvement is *unrelated to phrasing*, which Mikolov et al. used and we did not. The improvement is *not solely caused by using a larger window*: Without positional weight vectors, using `window=15` consistently decreases the accuracy compared to `window=5`. With positional weight vectors, accuracy is still improved by ca 3% when using `window=5`.

Currently, using positional weight vectors leads to 5–6× slowdown, and training for 5 epochs will produce a more accurate model (73%). Therefore, a further speed-up is required to make positional weight vectors practically useful. In ffad3b59, I have already drafted a BLAS implementation that uses [dsbmv][] to implement Z += aX ⊙ Y. Hopefully, this will speed things up a bit.

 [dsbmv]: http://www.netlib.org/lapack/explore-html/d7/d15/group__double__blas__level2_ga5c7ca036c788c5fd42e04ade0dc92d44.html
 [scalar]: https://github.com/Witiko/gensim/tree/bd2596aa
 [vector]: https://github.com/Witiko/gensim/tree/56376990
 [advances]: https://arxiv.org/pdf/1712.09405.pdf#page=3 (Advances in Pre-Training Distributed Word Representations)
 [notebook]: https://github.com/Witiko/gensim/blob/01e20060/docs/notebooks/fasttext-position-weights.ipynb

### Positional Scalar Weights and Weight Vectors

Below is a plot of the trained positional scalar weights:

positional-scalar-weights

Below is a plot of the trained positional weight vectors:

positional-weight-vectors

As expected, the magnitude of both the positional vectors and scalar weights increases around the center word, showing that short-term dependencies are more reliable for prediction. As @gojomo noted, some word vector features seem specialized for specific positions now.

gojomo:
Wow, super-thorough & interesting!

I can't yet think of any reason why analogies-with-`window=15` would be *so* bad pre-#2698, but then roughly comparable with smaller windows after. While there were lots of fixes, few affected the algorithm & I can't think of any that would have had a big interaction with `window`. If the newer code was more anomalous, this would be a glitch I'd want to dig further into, but as it's the older code, such digging is too low priority. 

Running with a single epoch is certainly defensible with a giant corpus – where any word's training examples throughout the  corpus, early & late, are equally diverse. But, if your metrics get noticeably better in any trial when running for more epochs, the model was likely still far-from-converged at the end of the one epoch, so all trials deserve a conditional/contingent asterisk – & it's possible converged-model-vs-converged-model would show larger tangible differences. (Ideally we'd be able to use loss-tallying to ascertain the actual point where training stops helping, but that feature remains quite broken, especially for large training runs.) 

That the usual pattern is for *smaller* windows to outperform on analogies is interesting, until the vector-based PDW offers a reversal. But that makes me wonder: might even smaller than `window=5` even better? Might the PDW in some way just be learning to (mostly) use an even-smaller window? 

I still wonder if in fact the original research disabled the random `reduced_windows` scaling when trying PDW. (Leaving both in place seems to apply 2 different styles of distance-weighting; further, it means oftentimes the longer-distance weights aren't even used - even if the underlying data had enough context. (For example, in the `window=15` case, if in fact the weights mean words-further-than-5 often have value, still 1/3 of the time, that potential training-info is being thrown away without even considering/tuning the weights. If the longer-positions are worth learning about, why often ignore them?) Similarly, I wonder about any interaction with the `sample` parameter. It discards many instances of common words - making effective windows shorter (and the training text *less* like natural text), often changing the relative positions of words that often appear with common words. It might have a strong interaction with PDW. 

 If the PDW vectors can be trained with little overhead after optimizations, it's an interesting option. (If the overhead is negligible, maybe a new default.) But also, what if *most* runs, or at least on similar kinds of data, wind up with similar PDW vectors? Maybe just assigning a "best practices" set of static PDW vectors at the beginning, untuned by training, could enable most the the benefit at little cost? (Maybe even some extreme choices would create some more-interpretable trained vectors? For example, dimensions 0-99 are overweight-trained by 4 nearest words; dimensions 100-199 overweight-trained by 4 *furthest* words; dimensions 200-299 equally trained by all context words.)

Witiko:
> That the usual pattern is for _smaller_ windows to outperform on analogies is interesting, until the vector-based PDW offers a reversal. But that makes me wonder: might even smaller than `window=5` even better? Might the PDW in some way just be learning to (mostly) use an even-smaller window?

We may try additional values of `window` (1, 2, 3, 4, 30). However, the average value of a positional weight vector feature is 0.8 (0.7 and 0.67 for the features of vectors at positions 1 and 30, respectively). Therefore, it does not seem that using a large window with PDW would be discarding more distant words. What seems to be happening is that we are effectively increasing the capacity of the model: Instead of learning just a single vector for every word, `2 * window` different word vectors are effectively learnt.

> I still wonder if in fact the original research disabled the random `reduced_windows` scaling when trying PDW. (Leaving both in place seems to apply 2 different styles of distance-weighting; further, it means oftentimes the longer-distance weights aren't even used - even if the underlying data had enough context. (For example, in the `window=15` case, if in fact the weights mean words-further-than-5 often have value, still 1/3 of the time, that potential training-info is being thrown away without even considering/tuning the weights. If the longer-positions are worth learning about, why often ignore them?) Similarly, I wonder about any interaction with the `sample` parameter. It discards many instances of common words - making effective windows shorter (and the training text _less_ like natural text), often changing the relative positions of words that often appear with common words. It might have a strong interaction with PDW.

We may try disabling `reduced_window` and try different values of the `sample` parameter. The `sample` parameter always seemed like a speed optimization that prevents overtraining on common words: I don't expect much interaction with PDW.

> If the PDW vectors can be trained with little overhead after optimizations, it's an interesting option. (If the overhead is negligible, maybe a new default.) But also, what if _most_ runs, or at least on similar kinds of data, wind up with similar PDW vectors? Maybe just assigning a "best practices" set of static PDW vectors at the beginning, untuned by training, could enable most the the benefit at little cost?

This would require further study. Different position-dependent weights may be trained for different languages and for different genres of text.

> (Maybe even some extreme choices would create some more-interpretable trained vectors? For example, dimensions 0-99 are overweight-trained by 4 nearest words; dimensions 100-199 overweight-trained by 4 _furthest_ words; dimensions 200-299 equally trained by all context words.)

By mistake, I only trained 75 dimensions of the positional weight vectors out of 300 in one of the runs (`memcpy` in 1586756e only copied `c.size` bytes instead of `cython.sizeof(REAL_t) * c.size`). As a result, I received an improvement in accuracy from 69.75%* to 74.58%*. This may also be worth investigating.

gojomo:
> Pre-trained positional weight vectors are an intriguing option, but users will require arbitrarily many positional weight vectors with arbitrary dimensionality, depending on `window` and `vector_size`. I am not sure how we would tackle that.

Many runs (with alternate window/dimensionality/corpus) may reveal some repeating rules-of-thumb, from the observed distributions of trained vectors. And because initial randomization makes it arbitrary whether it happens to be dimension #0 or #299 that overweights/underweights a certain window ranges. (That is: the N dimensions of any one model could be reordered based on PDW interactions before comparison/investigation for patterns or reuse as a fixed, specified parameter.)

Witiko:
> Many runs (with alternate window/dimensionality/corpus) may reveal some repeating rules-of-thumb, from the observed distributions of trained vectors. And because initial randomization makes it arbitrary whether it happens to be dimension #0 or #299 that overweights/underweights a certain window ranges. (That is: the N dimensions of any one model could be reordered based on PDW interactions before comparison/investigation for patterns or reuse as a fixed, specified parameter.)

We may try different initializations for the positional word vectors, lock them, train word vectors, and evaluate on the word analogy task. Even if we do not provide pre-trained positional weight vectors, better initialization may lead to faster convergence. However, I worry that unless we evaluate on a wide range of languages, text genres, and tasks, the general usefulness of our custom initialization will be dubious.

Witiko:
I am sorry to say that using [?sbmv][] to implement Z += aX ⊙ Y in 3d49f4c6 did not speed things up compared to 5637699, but slowed them down ~20× instead (more accurate measurements to be added after the training has finished).

In retrospect, this is not entirely surprising: [?sbmv][] is a level 2 BLAS matrix-vector operation, which I used as a vector-vector operation. As a result, fewer vectorization options are available to the compiler, leading to slower code. Sadly, there does not seem to exist a better fit for the Hadamard product in standard BLAS, although there does exist the [v?Mul][] operation in the Math Kernel Library.

 [v?mul]: https://software.intel.com/en-us/node/470448
 [?sbmv]: http://www.netlib.org/lapack/explore-html/d6/d30/group__single__blas__level2_gad8eb1004e944af8112a6e69f668ac53b.html

Witiko:
Next experiments should take place this week, I am still waiting for 3d49f4c to finish.

> Many runs (with alternate window/dimensionality/corpus) may reveal some repeating rules-of-thumb, from the observed distributions of trained vectors.

To see if the trained positional weight vectors can be modeled as normal distribution samples, I standardized all 30 positional weight vectors from 5637699 and I plotted the theoretical quantiles of the standard normal distribution (x-axis) against the empirical quantiles of the standardized features (y-axis):

pplot-2x

We see some left skew in the positional weight vectors and the D’Agostino-Pearson normality test outright rejects the normality hypothesis (p-value = 0). Nevertheless, modeling the data as normal may still be useful for initialization. Below is a plot of the trained positional weight vectors from 5637699 (left) and the positional weight vectors generated from 30 normal distributions (right):

comparison-2x

The stark difference between the two is likely because the trained feature weights also seem to be a continuous function of position:

comparison-2x-feature

Therefore, instead of modeling the individual positional 300-vectors as random samples, perhaps we should instead model the individual feature 30-vectors as noisy, translated and scaled density functions of a (normal) distribution.

Witiko:
In e03537ce, I rewrote the Hadamard product as an inline function without BLAS and I split `fasttext_fast_sentence_cbow_neg` into two functions, so that we can minimize branching in loops. This leads to more than 2–3× speed-up compared to 5637699:

Below are the English word analogy task total accuracies of the resulting word vectors:

|  | branch `develop`, before #2698 | branch `develop`, after #2698 | branch [`positional-dependent-scalars`][scalar], `pdw=0` | branch [`positional-dependent-vectors`][vector], `pdw=0` | branch [`positional-dependent-scalars`][scalar], `pdw=1` | branch [`positional-dependent-vectors`][vector], `pdw=1` |
|--|--|--|--|--|--|--|
| `window=5`  | 65.63%<sup>*</sup> | 65.40%<sup>*</sup> | 66.27%<sup>*</sup> | 65.52%<sup>*</sup> | 64.95%<sup>*</sup> |   70.37%<sup>*</sup>   |
| `window=15` | 31.88%<sup>*</sup> | 60.98%<sup>*</sup> | 61.67%<sup>*</sup> | 61.60%<sup>*</sup> | 62.22%<sup>*</sup> | **71.01%<sup>*</sup>** |

Below are the training durations on a dedicated machine with 252G RAM, 32 Intel Xeon X7560 2.26 GHz CPU cores, and OpenBLAS:

|  | branch `develop`, before #2698 | branch `develop`, after #2698 | branch [`positional-dependent-scalars`][scalar], `pdw=0` | branch [`positional-dependent-vectors`][vector], `pdw=0` | branch [`positional-dependent-scalars`][scalar], `pdw=1` | branch [`positional-dependent-vectors`][vector], `pdw=1` |
|--|--|--|--|--|--|--|
| `window=5`  | 2h 05m 55s | 1h 58m 32s | 2h 06m 02s | 2h 06m 33s | 2h 03m 21s |    3h 08m 05s   |
| `window=15` | 2h 26m 22s | 2h 23m 56s | 2h 20m 25s | 2h 29m 28s | 2h 28m 23s | **4h 59m 27s** |

This is still not practically useful, but we are getting there:

- `pdw=0`, `window=5`, `epochs=3`:
    - Training duration:  4h 34m 36s
    - English word analogy task, total accuracy: 71.04%<sup>*</sup>
- `pdw=1`, `window=15`, `epochs=1`:
    - Training duration:  4h 59m 27s
    - English word analogy task, total accuracy: 71.01%<sup>*</sup>

I don't expect to get much faster than this with full position-dependent weighting, but I'll be happy for suggestions on further optimizations. Additional speed-up can be gained by relaxing the requirements and using either pre-trained positional weight vectors, as suggested by @gojomo, or using fewer dimensions for positions, [as discussed above][above].

 [above]: https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-671090321
 [scalar]: https://github.com/Witiko/gensim/tree/bd2596aa
 [vector]: https://github.com/Witiko/gensim/tree/e03537ce

Witiko:
I am close to finishing the parameter optimization experiments on the dev set: `position_dependent_vector_size`, `size`, `window`, and `sample`. After I have finished and reported the results, I will train a model with the best parameters on Common Crawl and we'll see what the results are. Since we may easily get a SOTA on the English Analogy task out of the experiment, would you have the time and inclination to co-author a paper on this, @gojomo?

gojomo:
Looking forward to the results! 

I'd be happy to help improve any paper with some review/comments. But given the overwhelming amount of work & writing that you've done, maybe that's more of an 'acknowledgement' role than 'co-author'? (We can consider further as things develop - I'm not especially familiar with the customs for such collaboration/attribution.)

Witiko:
Here are the English word analogy task accuracies<sup>*</sup> for values {0, 5, 10, …, 295, 300} of parameter `position_dependent_vector_size`, which I have added in commit 0c9a8be3. The `position_dependent_vector_size` parameter controls the dimensionality of the positional weight vectors. Only the first `position_dependent_vector_size` word and n-gram vectors are weighted, whereas the rest are not weighted and forward- and backward-passed using the BLAS `saxpy` function for speed.

index

These results show that we can reap most of the benefits with just ten positional weight vectors and get the best accuracy with ca ⅓ of the word and n-gram vector dimensionality at a fraction of the speed cost. Needless to say that this makes the position-dependent weighting highly practically useful.

> I'd be happy to help improve any paper with some review/comments. But given the overwhelming amount of work & writing that you've done, maybe that's more of an 'acknowledgement' role than 'co-author'? (We can consider further as things develop - I'm not especially familiar with the customs for such collaboration/attribution.)

@gojomo I will appreciate any review and comments after I have drafted the paper. Since you were involved in the experimental design, I think it would be appropriate that you are one of the authors. For the pull request, it should be sufficient to show that we can reproduce the results of [Mikolov et al. (2017)](https://arxiv.org/pdf/1712.09405.pdf#page=2) on the English word analogy task at practical speed. For the paper, we should evaluate on more tasks (word similarity / relatedness, language modeling, …).

gojomo:
That's very interesting! A few questions:

Is `position_dependent_vector_size=0` definitively the same, in algorithm & run-time, as if this feature were not available at all?

Do you think the decreasing accuracy with more parameters may be a symptom of undertraining given the 1-epoch model?

Do I understand correctly that if `vector_size=100`, & `position_dependent_vector_size=10`, then there's an *extra* 10d vector for each window-position - 30 in the case of `window=15`? Then, that's applied to just part (such as the first 10d) of the usual per-word projection vectors, for forward-propagation, then only backpropagation-adjusted from that same part? 

(I wonder if any one, or some average of, the PDW position vectors should also be applied to the final word-projection-vectors when using them outside of positional contexts, as in the post-training evaluations of words alone.)

Witiko:
> That's very interesting! A few questions:
> 
> Is `position_dependent_vector_size=0` definitively the same, in algorithm & run-time, as if this feature were not available at all?

It is. If you look at the upper table in https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-670828783 and https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-677827927, you can see that for `window_size=15`, both `master` and our implementation with position-dependent weighting switched off received around 61%<sup>*</sup> accuracy on the English word analogy task, which is consistent with the plot in https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-687100878.

> Do you think the decreasing accuracy with more parameters may be a symptom of undertraining given the 1-epoch model?

If the decreasing accuracy is the result of undertraining, then the right tail of the plot should flip and the accuracy should start increasing as we add more epochs. If most word and n-gram vector features are position-independent, which is what I believe, then full position-dependent weighting only introduces noise and the right tail of the plot should not change much as we add more epochs.

> Do I understand correctly that if `vector_size=100`, & `position_dependent_vector_size=10`, then there's an _extra_ 10d vector for each window-position - 30 in the case of `window=15`? Then, that's applied to just part (such as the first 10d) of the usual per-word projection vectors, for forward-propagation, then only backpropagation-adjusted from that same part?

That's correct.

> (I wonder if any one, or some average of, the PDW position vectors should also be applied to the final word-projection-vectors when using them outside of positional contexts, as in the post-training evaluations of words alone.)

That struck me also. Are you aware of any applications of the word embeddings, where this would make sense?

Witiko:
I am currently running similar experiments for the `vector_size`, `window`, and `sample` parameters. My hope is that most of these will not be dependent on the positional weighting and we can skip them from the subsequent experiments to limit the scope and duration. Afterwards, I am going to rerun the experiment with `position_dependent_vector_size` from with more expochs and also with phrasing, so that we can verify if the decreasing accuracy in https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-687100878 is actually a consequence of undertraining and start moving closer to the configuration that we will use in the final evaluation on Common Crawl.

Witiko:
I am currently quite happy with the code, but there is [one ugly hack](https://github.com/RaRe-Technologies/gensim/blob/0c9a8be3bdf7f2f13d5e74f70c28a4cbe3ff2b9a/gensim/models/fasttext_inner.pyx#L405) that we need to get rid of before the final experiment and the merge: The positional weight vectors are initialized to ones, which corresponds to vanilla fastText weighting, and then they are updated by SGD. However, when using the standard SGD, positional, word, and n-gram vectors tend to NaN, likely because of the initialization to ones, which enlarges the gradients. To hotfix this, I decreased the learning rate for positional vectors by a factor of `vector_size`:

```cython
    inv_count = ONEF / c.size

    n = j - i + c.window
    for m in range(j, k):
        if m == i:
            continue
        memcpy(c.neu1, &c.syn0_positions[n * c.pdw_size], c.pdw_size * cython.sizeof(REAL_t))
        our_hadamard_product(
            c.pdw_size, inv_count,
            &c.syn0_vocab[c.indexes[m] * c.size],
            c.work,
            c.neu1)

        [...]

        memcpy(&c.syn0_positions[n * c.pdw_size], c.neu1, c.pdw_size * cython.sizeof(REAL_t))
        n += 1
```
A possible solution is to initialize positional, word, and n-gram vectors to a standard normal distribution. A minor detail is that the product distribution of standard normal distributions is not normal, although it still has zero mean and unit variance:

index

I am not sure how this will affect the training, but it's something to keep in mind. Sadly, neither [Mikolov et al. (2017)](https://arxiv.org/pdf/1712.09405.pdf#page=2) nor [Mnih and Kavukcuoglu (2013)](https://papers.nips.cc/paper/5165-learning-word-embeddings-efficiently-with-noise-contrastive-estimation.pdf#page=3) discuss initialization.

Alternatively, we could initialize the positional vectors and the first `position_dependent_vector_size` dimensions of word and n-gram vectors with [a distribution whose product distribution is standard normal](https://math.stackexchange.com/questions/3817393/a-distribution-whose-product-distribution-is-standard-normal), i.e. d<sub>p</sub> ⊙ u<sub>t + p</sub> ~ N(0, 1). This would be perfect, but I don't know which distribution would satisfy this.

Witiko:
@gojomo Here are the English word analogy task accuracies<sup>*</sup> for the values {100, 200, …, 700, 800} of parameter `vector_size` with and without position-dependent weighting:

index

To me, the results to indicate that position-dependent weighting and vector dimensionality are independent with respect to the English word analogy task. The decreasing accuracy is likely the result of dividing the learning rate positional weight vectors as described in https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-688421837.

Witiko:
@gojomo Here are the English word analogy task accuracies<sup>*</sup> (red curves) and training durations (blue curves) for the values {1, 2, 3, …, 9, 10, 15, 20, 25, 30, 40, 50} of parameter `window` with (dashed curves) and without (solid curves) position-dependent weighting. In both cases, we are using the reduced windows.

index

This is a crucial result that shows the main contribution of position-dependent weighting: Whereas vanilla fastText cannot cope with larger windows, position-dependent weighting allows fastText to take advantage of window sizes up to 10 (although this is likely specific to English), resulting in the five point advantage in accuracy<sup>\*</sup>, and can cope with window sizes up to 50 without loss of accuracy<sup>\*</sup>. Notice also that unlike [Mikolov et al. (2017)](https://arxiv.org/pdf/1712.09405.pdf#page=2), who suggest using window of size 15, we receive the best results with window of size 10, which is a more practical default with respect to speed.

Witiko:
@gojomo Here are the English word analogy task accuracies<sup>*</sup> (red curves) and training durations (blue curves) for the values {10<sup>-6</sup>, 10<sup>-5</sup>, 10<sup>-4</sup>, 10<sup>-3</sup>} of parameter `sample` with (dashed curves) and without (solid curves) position-dependent weighting.

index_

To me, the results indicate that position-dependent weighting and sampling frequency are independent with respect to the English word analogy task. The default value of 10<sup>-5</sup> seeps to produce the best results.

Witiko:
I am currently removing the gradient clipping hack discussed in https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-688421837. However, initializing the weights properly is more difficult than it would seem. Untill now, I initialized the positional weights to ones, so that the product of the word/n-gram weights and positional weights keeps the same initial distribution as vanilla fastText (recall that with position-dependent weighting, the hidden layer weights are a product of two weight matrices). However, this makes the gradients of word/n-gram vectors too large and makes them approach NaN. As a hotfix, I divided the gradient updates by vector_size, which fixed things, but now we need a less arbitrary and more permanent solution to initialization.

Currently, [Gensim is initializing the word and n-gram vectors with Uniform(-1 / vector_size, 1 / vector_size) distribution](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/fasttext.py#L1336). This is consistent with [facebook's fastText](https://github.com/facebookresearch/fastText/blob/master/src/fasttext.cc#L711), but always seemed an odd choice, since weights are typically initialized with a normal distribution. Additionally, it makes the initialization a real headache: The normal distribution [can easily be factorized](https://math.stackexchange.com/a/3818660/133843) to create two independent random variables whose product is normal:

2-factors of N(0, 1)

However, there seems to be no such nice result for the uniform random distribution. The product of two independent sqrt(a) * Beta(1.2, 0.5)-distributed random variables [seems roughly Uniform(0, a)-distributed](https://ravshansk.com/articles/uniform-distribution.html):

2-factors of Uniform(0, 1)

Alas, shifting to Uniform(-a, a) would require additional translations after every product of word/n-gram weights and positional weights in the forward- and backward-pass and seems generally messy.

@piskvorky @gojomo Given the above, would it be acceptable if I initialized word/n-gram vectors with the N(0, 1) / vector_size distribution instead of the Uniform(-1 / vector_size, 1 / vector_size) distribution when position-dependent weighting is disabled, and with the 2-factor of the N(0, 1) / vector_size distribution when position-dependent weighting is enabled? Obviously, I would test that this does not deteriorate the performance before merge.

piskvorky:
> @piskvorky @gojomo Given the above, would it be acceptable if I initialized word/n-gram vectors with

I have no intuition on the impact of Gauss vs Uniform (vs Beta) initialization on word2vec performance, sorry.

More broadly – can you give a TL;DR of the user benefits please? Something pithy, to potentially go in the release notes / tweet.

You've done a tremendous amount of work @Witiko and I'm sure it'll be an interesting paper; I see two paths forward:

1. This is better than the current algo, for the majority of common training scenarios. We merge and change Gensim's default to this.
2. This improves a fringe use-case, or has serious caveats, so to be only used in "special circumstances".

   In this case the PR is better served as a standalone "contributed repo", extending Gensim, I'm -1 on merging yet another "maybe" switch. We're moving Gensim in the direction of less switches, more compact core, less cognitive load.

Thanks!

Witiko:
> More broadly – can you give a TL;DR of the user benefits of this PR please? Something pithy, to potentially go in the release notes / tweet.

This PR contains the first public implementation of fastText's positional weighting from the [the 2017 “Advances” paper by Mikolov et al.](https://arxiv.org/pdf/1712.09405.pdf#page=2). With no extra training time, we gain 5% on the Analogy task scores and can now [exploit larger context windows](https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-691714727).

It needs more time in the oven, though, see the TODO list in the original post. Any deadlines I should aim for if this should appear in Gensim 4.0?

> You've done a tremendous amount of work @Witiko and I'm sure it'll be an interesting paper; I see two paths forward:
> 
> 1. This is better than the current algo, for the majority of common training scenarios. We merge and change Gensim's default to this.
>
> 2. This improves a fringe use-case, or has serious caveats, so to be only used in "special circumstances".
>    In this case the PR is better served as a standalone "contributed repo", extending Gensim, I'm -1 on merging yet another "maybe useful" switch. We're moving Gensim in the direction of fewer switches, more compact core, less cognitive load.

So far it seems like case 1, but we have only tested on the English word analogy task in this PR.

piskvorky:
5% extra accuracy for no extra training time sounds great. Any gotchas? Why would some app want to **not** use this?

Witiko:
@piskvorky The only apparent gotcha is the speed. If we use positional weights with the same dimensionality as the word/n-gram vectors, the training time doubles (compare the leftmost and the rightmost duration in [the graph](https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-687100878)). This is in part because BLAS does not support the pointwise (Hadamard) product that is used in positional weighting. We could further optimize the speed if we also used the MKL. However, using only a couple of positional features (50 out of 300) seems to give all the benefits (see the left part of [the graph](https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-687100878)) with next to no slow-down.

piskvorky:
I saw that graph and the difference between "0" and "50" on the X-axis seems to be 1h on the Y-axis.

Which doesn't match "at the same training time" – that's why I ask for a TL;DR clarification. What am I missing?

Witiko:
You can get most of the benefit (+10% word analogy task accuracy) just by training 10-dimensional positional weights with negligible added cost (tens of minutes). With 50-dimensional positional weights, you get additional 5% at 1 additional hour training time, which is significantly more than you would get for corresponding increase in the amount of training data. However, we only evaluated on the English Wikipedia so far. We expect smaller gains with larger datasets that were used to train SOTA fastText models (Common Crawl).

Witiko:
@piskvorky Does this sound like something that you'd like to be the new default fastText for Gensim? I have paused the experiments for the moment, but I can finish the implementation and rerun the evaluation on a larger dataset (to reproduce the SOTA results) in ca one week if we'd like to merge soon.

piskvorky:
Yeah replacing the default is a scary operation, but if the numbers bear it, let's do it.

I'd like to see some more experiments first though – different tasks on different datasets, not just analogy-on-Wikipedia. Ideally with more TL;DR graphs and observations, a more direct head-to-head comparison against the existing default.

Are there any additional parameters to tweak? Are the results sensitive to them? If so, what defaults + what guidance do we offer users for setting them?

Witiko:
> Are there any additional parameters to tweak? Are the results sensitive to them? If so, what defaults + what guidance do we offer users for setting them?

We have done experiments to see how the other parameters of fastText react to position-dependent weighting:

- Word vector dimensionality and position-dependent weighting seem independent. https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-690680917
- Sampling frequence and position-dependent weighting seem independent. https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-692189892
- Position-dependent weighting copes better with larger context windows. https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-691714727

The experiments use only the small Wikipedia dataset and do not use proper initialization https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-688421837, so they are not perfectly conclusive. I am planning to rerun them with the final implementation and a larger dataset used for training SOTA fastText models ([C4](https://www.tensorflow.org/datasets/catalog/c4)).

> I'd like to see some more experiments first though – different tasks on different datasets, not just analogy-on-Wikipedia. Ideally with more TL;DR graphs and observations, a more direct head-to-head comparison against the existing default.

That's on my wishlist as well. I have a couple of interested postgraduates that are looking to evaluate on a number of other tasks, but this will realistically take a couple of months.

Witiko:
@piskvorky > Are there any additional parameters to tweak? Are the results sensitive to them? If so, what defaults + what guidance do we offer users for setting them?

We have conducted experiments to see how the other parameters of fastText react to position-dependent weighting:

- Word vector dimensionality and position-dependent weighting seem independent. https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-690680917
- Sampling frequence and position-dependent weighting seem independent. https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-692189892
- Position-dependent weighting copes better with larger context windows. https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-691714727

The experiments use only the small Wikipedia dataset and do not use proper initialization https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-688421837, so they are not perfectly conclusive. I am planning to rerun them with the final implementation and a larger dataset used for training SOTA models ([C4](https://www.tensorflow.org/datasets/catalog/c4)).

> I'd like to see some more experiments first though – different tasks on different datasets, not just analogy-on-Wikipedia. Ideally with more TL;DR graphs and observations, a more direct head-to-head comparison against the existing default.

That's on my wishlist as well. I have a couple of interested postgraduates that are looking to evaluate on a number of other tasks, but this will realistically take a couple of months. I we're taking shortcuts, we could be satisfied with reproducing [the SOTA word analogy results on several languages](https://arxiv.org/pdf/1712.09405.pdf#page=2).

piskvorky:
> We could make position-dependent weighting the new default later after a sufficient additional evaluation on different tasks has been conducted.

That sounds like a plan. Thanks for your hard work, I appreciate the thought and effort into this.

Witiko:
In a4baef46, I have implemented random sampling from factorized normal distribution as `matutils.normal_factors` together with 15 unit tests in `test.test_matutils`. This is the first step for fixing the initialization as discussed in https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-698049085. Example below:

```python
>>> from gensim.matutils import normal_factors
>>> import matplotlib.pyplot as plt
>>> import numpy as np
>>> 
>>> rand_obj = np.random.default_rng(seed=42)
>>> X, Y = normal_factors(k=2, size=10**7, rand_obj=rand_obj)
>>> 
>>> fig, (ax1, ax2, ax3) = plt.subplots(1, 3, figsize=(15, 4), constrained_layout=True)
>>> fig.suptitle('Random variable probability density functions')
>>> for ax, s, variable in [(ax1, X, '$X$'), (ax2, Y, '$Y$'), (ax3, X * Y, '$XY$')]:
>>>     ax.hist(s, 'auto', density=True)
>>>     ax.set_title('Variable {}'.format(variable))
>>>     ax.set_xlim(-1.96, 1.96)
>>>     ax.set_xlabel('value')
>>>     ax.set_ylabel('probability density')
>>>     ax.grid()
>>> plt.show()
```

example code output

Perhaps we can get `matutils.normal_factors` to the scipy codebase later, so that we don't have to maintain what is a generally useful but only recently discovered equation. [(Pinelis, 2018)](https://arxiv.org/abs/1803.09838) However, it's only 31 LOC, so it's not too inconvenient if we don't.

gojomo:
Unless some alternate initialization gave improved results – such as faster convergence – I wouldn't deviate from the convention used in word2vec papers/code or FastText papers/code.

Witiko:
@gojomo I would also rather keep the uniform initialization, but that's difficult with positional weighting, see https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-698049085. The effect of switching from uniform to normal initialization should not impact the speed or accuracy, but it will be tested. Note that the initialization has never been discussed in the papers, so it's a minor implementation detail.

gojomo:
Our standard has been matching the reference implementations to the maximum extent possible, rather than just the papers.

Witiko:
@gojomo In that case, I can either

1. keep the uniform initialization when position-dependent weighting is disabled, and use the factors of the normal distribution to initialize positional, word, and n-gram vectors when the position-dependent weighting is enabled, or
2. try to obtain a reference implementation from the authors of [the 2017 “Advances” paper by Mikolov et al.](https://arxiv.org/pdf/1712.09405.pdf#page=2) and follow their initialization.

Would that be acceptable?

Witiko:
@gojomo To reinforce the argument for changing the initialization: in CBOW, context word embeddings are averaged and then passed to the hidden layer. Whatever the initial distribution (normal or uniform), the mean tends to the normal distribution by the central limit theorem. Therefore, the initial distribution is less important than the mean and variance of the initial random samples, which affect the mean and variance of what the hidden layer sees.

Witiko:
@gojomo 

> a quick timeit spot check indicates .normal() about 4x slower than .random() - which is likely to be noticeable to users with large models

Thank you for the measurements. I wonder, compared to training time, would the difference in initialization time be significant?

> If the distribution is unimportant, then faithfully matching the actual `word2vec.c`/`fasttext` practices that Gensim claims to re-implement in Python would seem the dominant consideration. [...] ask on the list or a forum, "why is Gensim using a differently-distributed initialization?" I don't find, "it makes no difference either way, so we chose to be different" a very satisfying answer.

To be clear, the switch to normal distribution is not "to be different", but because no reference implementation is available and the uniform distribution is difficult to factorize as discussed in https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-698049085. We need to factorize, because positional weighting multiplies word and positional vectors before the hidden layer, and it is *their product* that needs to have the proper distribution.

> `word2vec.c` uses uniform random over (-0.5/fanout, 0.5/fanout], while FastText uses uniform random over (-1.0/fanout, 1.0/fanout].

Interesting! I can't say I know why the choice was made, but it is likely to decrease the initial effective learning rate in word2vec compared to fastText.

gojomo:
Wouldn't initializing the *new* positional weights however they need to be initialized be sufficient? Then, word-vector/ngram-vector initialization can remain conventional. Separately, even in the absence of a PDW reference implementation, can you ask the PDW paper authors what they actually did, and whether they changed the word-vector initialization in any way?

Witiko:
@gojomo 

> Wouldn't initializing the new positional weights however they need to be initialized be sufficient?

That's what the current code does: word and n-gram vectors are initialized with uniform distribution and positional vectors are initialized to ones. However, this lleads to exploding gradients of word and n-gram vectors. Unless we arbitrarily decrease the gradients (see https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-688421837), the word and n-gram vectors tend to NaN.

> Can you ask the PDW paper authors what they actually did, and whether they changed the word-vector initialization in any way?

I have asked on October 5, but I have received no response yet.

Witiko:
We held [a seminar session](https://is.muni.cz/el/fi/podzim2020/PV212/index.qwarp?prejit=5595948) regarding the position-dependent weighting with our research group yesterday.   
So far, initializing with the square of normal distribution seems to be the best option, see 1:25:17 -- 1:50:00.

Witiko:
@gojomo I bring good news: I got rid of [the gradient clipping hack](https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-688421837) *without changing the initialization*, i.e. we are keeping the vanilla fastText initialization for vector and n-gram weights and we initialize the positional vectors to ones. The exploding gradients were related to improper normalization of gradient updates to positional weights, which I have fixed in 94a57ffc. Since the gradient clipping hack decreased the learning rate for positional weights, we see another 5% increase in the English word analogy task accuracy<sup>*</sup>:

|  | branch [`positional-dependent-vectors`][vector], `pdw=0`  | branch [`positional-dependent-vectors`][vector], `pdw=1`, before 94a57ffc | branch [`positional-dependent-vectors`][vector], `pdw=1`, after 94a57ffc |
|--|--|--|--|
| `window=5`  | 65.52%<sup>*</sup> | 70.37%<sup>*</sup> | 74.14%<sup>*</sup>
| `window=15` | 61.60%<sup>*</sup> | 71.01%<sup>*</sup> | **75.02%**<sup>*</sup>

I am currently re-running [the `vector_size` experiment](https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-690680917), which was the one most affected by the gradient clipping hack. After that, I will rerun [the `position_dependent_vector_size` experiment](https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-691714727), which could catapult us to **80%**<sup>*</sup> accuracy, i.e. within 8% of the current English word analogy SOTA *without phrasing and using just the English Wikipedia corpus (4% of Common Crawl)*. After that, I'll be trying to reproduce the phrasing, which should give us another 5% bump, but this may take a while due to missing parameters in [the 2017 “Advances” paper by Mikolov et al](https://arxiv.org/pdf/1712.09405.pdf#page=2) as [discussed in the Gensim group](https://groups.google.com/u/2/g/gensim/c/D3F0wePSRzM). Finally, training on Common Crawl should get us past the English word analogy SOTA.

 [vector]: https://github.com/Witiko/gensim/tree/e03537ce

piskvorky:
That looks awesome. Do I understand correctly the "improper normalization" is present in current Gensim too? How about FB's fastText?

Witiko:
> That looks awesome. Do I understand correctly the "improper normalization" is present in the current Gensim too?
> How about FB's fastText?

The improper normalization is only related to this PR and the positional vectors, i.e. there should be no issue with either the current Gensim or FB's fastText.

Witiko:
I re-ran [the `vector_size` experiment](https://github.com/RaRe-Technologies/gensim/pull/2905#issuecomment-690680917), which was the one most affected by the gradient clipping hack. The gradient clipping hack decreased the learning rate proportionately to the `vector_size`, so the model would not converge for larger values of `vector_size` see the results before (above) and after removing the gradient clipping hack (below):

92793880-22fc2e00-f3af-11ea-842f-1b19d8f9c996
index

Witiko:
@piskvorky @gojomo In two upcoming articles from our research group, we have [compared the English word analogy task performance of the positional weighting using different initializations](https://www.overleaf.com/read/srzrhgyssynf):

analogy-task

We have also [confirmed the usefulness of positionally-weighted word embeddings for text classification and language modeling](https://www.overleaf.com/read/pgxhwnbnvprq):

text-classification
language-modeling


These results are still using only the positional word vectors trained on the English Wikipedia without phrasing, so they are not the SOTA. Nevertheless, they lllustrate the general usefulness of positional weighting across a number of different NLP tasks.

piskvorky:
@Witiko could you check https://groups.google.com/g/gensim/c/5mAeWrQN7lg/m/7Ul-uMJWBAAJ please? (FYI, in case you're not getting the mailing notifications). Thanks!

Witiko:
The work from this pull request is currently available in the [witiko/gensim@pine](https://github.com/witiko/gensim/tree/pine) fork of Gensim 3.8.3 and through the high-level [MIR-MU/PInE](https://github.com/MIR-MU/pine) library. The architecture and the conducted experiments are available in our admittedly clickbaity preprint [“When FastText Pays Attention: Efficient Estimation of Word Representations using Constrained Positional Weighting”](https://arxiv.org/abs/2104.09691) on arXiv.

Witiko:
The work from this draft has been accepted as [a journal paper][1], which was co-authored by @piskvorky and is to appear in [J.UCS 28:2][4]. As I wrote above, we have produced [the high-level PInE library][2], which was used for the experiments and which uses [a fork of Gensim 3.8.3][3] for model training. If there is interest, we can rebase the fork on top of the current `develop` branch.

 [1]: https://arxiv.org/abs/2104.09691v5
 [2]: https://github.com/MIR-MU/pine
 [3]: https://github.com/witiko/gensim/tree/pine
 [4]: https://www.jucs.org/jucs_journal.html

PR #2907, opened 2020-07-30
Fix doc2vec crash for large sets of doc-vectors

### LINKED ISSUE #2894: Gensim Doc2Vec model Segmentation Faulting for Large Corpus

gojomo:
Fixes small triggering case for #2894, by using `long long` instead of `int` for a value that (via multiplication) eventually determines an array index that needs to reach further than 2^31 positions forward, for large sets of doc-vectors in the `corpusfile` mode. 

Fixes #2894.

### DIFF
--- a/gensim/models/doc2vec_corpusfile.pyx
+++ b/gensim/models/doc2vec_corpusfile.pyx
@@ -162,7 +162,7 @@ def d2v_train_epoch_dbow(
     cdef int sent_idx, idx_start, idx_end
 
     cdef vector[string] doc_words
-    cdef int _doc_tag = start_doctag
+    cdef long long _doc_tag = start_doctag
 
     init_d2v_config(
         &c, model, _alpha, learn_doctags, learn_words, learn_hidden, train_words=train_words,
@@ -302,7 +302,7 @@ def d2v_train_epoch_dm(
     cdef REAL_t count, inv_count = 1.0
 
     cdef vector[string] doc_words
-    cdef int _doc_tag = start_doctag
+    cdef long long _doc_tag = start_doctag
 
     init_d2v_config(
         &c, model, _alpha, learn_doctags, learn_words, learn_hidden, train_words=False,
@@ -455,7 +455,7 @@ def d2v_train_epoch_dm_concat(
     cdef int sent_idx, idx_start, idx_end
 
     cdef vector[string] doc_words
-    cdef int _doc_tag = start_doctag
+    cdef long long _doc_tag = start_doctag
 
     init_d2v_config(
         &c, model, _alpha, learn_doctags, learn_words, learn_hidden, train_words=False,
@@ -469,6 +469,17 @@ def d2v_train_epoch_dm_concat(
             effective_words = 0
 
             doc_words = input_stream.read_sentence()
+
+            # FIXME? These next 2 lines look fishy to me (gojomo). First, skipping to
+            # 'total_documents' (end) seems it'd do nothing useful. Second, assigning
+            # into what is typically a count (`doctag_len`) from a boolean test is
+            # sketchy, even if in the current limitations of this mode (corpus_file)
+            # only '1' is a workable value. But, this code seems to pass at least
+            # one real has-some-function test (test_dmc_hs_fromfile), and this mode
+            # is rarely used, & I haven't written this code & would prefer to see the
+            # whole duplicate-logic of corpus_file mode removed in favor of an approach
+            # with less duplication. So I'm not sure anything is broken & it's far from
+            # a near-term priority - thus leaving this note.
             _doc_tag = total_documents
             c.doctag_len = _doc_tag < c.docvecs_count
 


gojomo:
The failing CI run is in some doc-building step of `run_annoy.py` that was failing before this PR.

piskvorky:
Can you run `cd docs/src; make clean html`? I'm pretty sure `run_annoy` is working, but maybe I forgot to refresh the auto-generated files with `make`.

gensim/models/doc2vec_corpusfile.pyx
```diff
@@ -469,8 +469,8 @@ def d2v_train_epoch_dm_concat(
             effective_words = 0
 
             doc_words = input_stream.read_sentence()
-            _doc_tag = total_documents
-            c.doctag_len = _doc_tag < c.docvecs_count
+            _doc_tag = total_documents  # WTF? skipping to end no matter what start_doctag passed in?
```
piskvorky, line 472: `FIXME` easier to grep than `WTF`. Plus a ticket, so we don't forget to clean before release if indeed critical.

gojomo:
> Can you run `cd docs/src; make clean html`? I'm pretty sure `run_annoy` is working, but maybe I forgot to refresh the auto-generated files with `make`.

(1) Should the docs-build really be dependent on developers manually running make, and committing such auto-generated files to source control? Seems like a mess.

(2) I tried, but after installing `sphinx` then `sphinx-gallery` then a 2-years-old `sphinxcontrib-napoleon` (which should have been obsoleted by a 5-years-old sphix-1.3.0 release), it still wasn't working. Seems like the docs require a whole bunch of bunch of precisely-pinned old packages, per `requirements-docs.txt`:

```
Sphinx==2.2.0
sphinx-gallery==0.4.0
sphinxcontrib-napoleon==0.7
sphinxcontrib-programoutput==0.15
memory-profiler==0.55.0
Pyro4==4.77
annoy==1.16.2
scikit-learn==0.21.3
plotly==4.2.1
nltk==3.4.5
testfixtures==6.10.0
statsmodels==0.10.1
pyemd==0.5.1
pandas==0.25.2
```

So, I'll let whoever designed and/or broke this handle it.

gensim/models/doc2vec_corpusfile.pyx
gojomo: I'm not sure if it's critical. It just looks sketchy, & at a glance it's hard to imagine it doing anything - but this mode does in fact seem to pass at least 2 tiny functional tests (`TestDoc2VecModel.test_dmc_hs_fromfile` & `TestDoc2VecModel. test_dmc_neg_fromfile`). So maybe it's fine. I just want to leave some warning to anyone looking at it in the future. (It's unlikely to be me, as I strongly think the `corpus_file` duplicated-paths should be discarded in favor of some other more maintainable approach using far less code-duplication.)

piskvorky:
I also tried with the latest `sphinx` in #2899, but it was failing with obscure errors (for which google didn't show me much). Something with "duplicate labels" IIRC.

After downgrading to `sphinx==2.4.4` (from `setup.py`, I didn't even notice `requirements-docs.txt`) it worked though. I don't think I installed anything else. @mpenkov is that "napoleon" package etc really needed?

The tutorials can take a long time (=> time out in CI) but yes, having them refresh automatically would be much better than doing it manually. I'll push the regenerated docs here and handle the rest separately, not to hijack this PR.

piskvorky:
I couldn't push the fix directly into your branch, so I merged it into `develop`.

@gojomo Please merge `develop` into here and if all green, we're good to merge here as well.

mpenkov:
@piskvorky I don't think we need napoleon. It's for using Google-style docstrings, and we don't use them.

mpenkov:
> > Can you run `cd docs/src; make clean html`? I'm pretty sure `run_annoy` is working, but maybe I forgot to refresh the auto-generated files with `make`.
> 
> (1) Should the docs-build really be dependent on developers manually running make, and committing such auto-generated files to source control?

We currently expect people to auto-generate the files for parts of the documentation that they touched. If they don't, then the CI tries to do it for them. I think this is what happened with the annoy docs. It looks like part of the autogenerated files were not updated:

```
git diff --summary 4c1b3f7 4c1b3f7~ --name-status | cat
M       docs/src/auto_examples/tutorials/run_annoy.py
M       docs/src/gallery/tutorials/run_annoy.py
M       gensim/models/keyedvectors.py
M       gensim/similarities/annoy.py
```

The run_annoy.rst file didn't get updated. I think we can implement a script to automatically check for such omissions and block the PR. This would ensure authors push complete documentation.

> Seems like a mess.

Consider the alternatives:

1. The CI could generate the auto-generated files itself. This would take hours for each commit.
2. We could leave documentation out of CI. We would no longer be able to tell if our documentation is correct.
3. Can you think of any others?

gojomo:
My preferred alternative would be for no auto-generated artifacts that risk taking hours to auto-generate be in the core project, as *either* canonical original sources *or* compiled-outputs.

Lengthy auto-generation steps don't mix well with the main compile-&-unit-test role of the main CI, which is good to run on every-commit-and-PR. Witness, for example, the default auto-fail after 10m of no-evident progress here. 

As far as I can tell, any code here which purports to build-what's-missing-on-the-CI-box-even-if-it-takes-hours could never have worked, given typical CI service timeouts. 

Could it be made to work, with longer timeouts or some keep-alive progress-logging? Maybe, but it strikes me as misguided to try. A CI build-test cycle that "usually" gives an OK/bad result in under 20m, but sometimes takes hours/days before even giving an "all-OK", loses much of the value of the auto-CI – it has high-variance automated steps in the wrong place. This is especially the case if something like a tiny typo change in some documentation could trigger the long-cycle. 

So: any documentation (like demo tutorials) that specifically require active automated testing, including potentially "hours" of runtime, should move elsewhere.

mpenkov:
> My preferred alternative would be for no auto-generated artifacts that risk taking hours to auto-generate be in the core project, as _either_ canonical original sources _or_ compiled-outputs.

I think that's a reasonable preference. We can achieve it by making sure (via a script) that people who touch documentation push complete changes, as I mentioned earlier.

piskvorky:
I see no real alternative either. Doesn't moving parts of our docs + testing elsewhere simply shift the problem around? That new place will still need testing, whenever things change. No difference. Except code + docs in one place makes more sense conceptually IMO.

gojomo:
"Code + human written docs" in the same place makes sense. 

"Code + human written docs + giant-demos-that-take-tens-of-hours-or-multiple-days-to-run" together makes no sense. It's mixing things with wildly different contribution/testing rhythms together, to the detriment of both – a pain for both long-term maintainers & casual contributors.

piskvorky:
All docs (tutorials included) are human-written.

Demos and tutorials are an important *subset* of the documentation. That's also why we render them into multiple formats & offer as Python scripts, HTML, notebooks.

But since there's no other option on the table, this all sounds very academic and hypothetical. Or maybe you're talking about something else than me and @mpenkov?

gojomo:
If a demo/tutorial requires hours (or even tens-of-minutes) to run, it no longer fits the mould of something that can be part of a rapid CI build/test cycle. It's a different kind of docs, that should be part of a different system. Maybe, a different tutorials sub-project.

The code that you changed that broke the build seems to have been triggering something on the buildbox – lengthy regeneration of derivative content – that (1) by reasonable standards, isn't the kind of derivation that should even be attempted in a CI build; but also (2) **could never have worked**, given the usual CI build timeouts - so there were bad assumptions about the build process coded in. 

I think a proper sense of what goes where would've prevented that failure, and this tangent conversation, and would also prevent similar failures in the future. So the other option is to adopt some principles that would help:

* **Builds should take minutes not hours** - so as a matter of build/test on the main project, require fast/stable CI operations and make ease-of-contribution a priority. Don't design-in any steps where a small desirable change  (like changing a trivial bit of text or variable name in docs) could require "hours" of extra build-time. 
* **Only commit masters, not developer-machine-created-autogenerated-outputs, to source-control**. The project CI cycle also shouldn't depend on developers autogenerating artifacts on their myriad dev-configs, potentially for hours, either. That's an extra barrier to contributions, and risks reliance on idiosyncratic undocumented/non-source-controlled aspects of developer working configurations. If, as a requirement of having realistic data-science tutorials, there are necessarily ome long-running demos that require auto-generation/auto-testing, those deserve to be silo'd to a separate build/release process where the cycles fit their needs, that's neither limited by the main-project fast-CI cycles, nor throws a monkey-wrench into them. For example, this might be a separate 'gensim-tutorials' project.

This is distantly-related to the the prior on-again-off-again practice of committing cython's C/C++ output into source control. That was also a hard-to-maintain, error-prone situation with too much reliance on manual developer steps.

mpenkov:
I think we can continue the documentation discussion elsewhere (for example, recently opened #2916).

This PR is good to merge once the build failure is taken care of. @gojomo Can you please merge develop?

https://github.com/RaRe-Technologies/gensim/pull/2907#issuecomment-667886628

gojomo:
I've added an extensive in-code comment to capture thoughts about https://github.com/RaRe-Technologies/gensim/pull/2907/files#r463932887 .

And, to keep this a simple few-lines, 1-file, 1-commit PR, I've chosen to rebase rather than adding another intervening merge-commit. (I wouldn't normally do that on a PR already in MRG-requested state, with comments, but everything from that history will be captured in the new single commit.)

PR #2910, opened 2020-08-03
[MRG] Refresh docs for run_annoy tutorial

piskvorky:
Refreshed auto-generated docs with `cd docs/src && make clean html`, after fixes in #2899.


piskvorky:
Hm, now it's the word2vec model failing some sanity checks (unrelated to changes in this PR):

Screen Shot 2020-08-03 at 10 04 57

I'll restart, but the flaky tests are still a problem.

Issue #2886, opened 2020-07-18
Clean up keras code


mpenkov:
But looking more closely at both the tested method & test code:

This utility function doesn't really belong cluttering up the `KeyedVectors` class: it's a fairly trivial bit of data-translation, very specific to some narrow uses, with an extra package requirement. It could & should be a in a keras-specific submodule that's only imported by people using keras, and could take the `KeyedVectors` instance as an argument (rather than `self`). 

Or, in the spirit of #2852 ("reduce gensim surface area"), it could be moved to a separate package (`gensim-extras`) - so it doesn't encumber testing/development of core gensim, or removed in favor of a tiny bit of example code, "here's how you convert gensim KeyedVectors into a keras Embedding", in docs, an example notebook, a 'recipes' webpage, or a blog post. It's 2-6 lines of trifling code, but its test failure due to a 'cosmic ray' from another project has soaked up a bunch of attention/time from 4 people who could be making other more important gensim improvements. 

Looking at the `test_keras_integration.py` code, it also failed because it was insufficiently focused on the gensim functionality.  It's not checking if the return value is just a translation of gensim values into a legal/sensible `Embedding` - it's doing some full-cycle `Word2Vec` training and some full-cycle `keras` training. It's one of the 20 slowest tests in our whole test suite – sure, only ~9 seconds, but of negligible value usually over the many hundreds/thousands of times it's triggered as a result of other changes, and of negative value in this case. 

(Gensim has a bunch of tests that take a lot of time but don't test much real, just completion-without-error, and often which deeply test inner structural details rather than meaningful results, and thus serve as a drag on sensible refactorings when they trigger test failures that require maintenance without tangible benefits to developer or gensim end-users.)

To advance #2852, I'd copy the existing method code to a new wiki page on "using gensim with keras", leave a stub pointing users there, and remove the functionality/tests from gensim so there will be no further maintenance cost.

_Originally posted by @gojomo in https://github.com/RaRe-Technologies/gensim/pull/2869#issuecomment-651354811_

gojomo:
Keras-related tests across multiple PRs have again started failing due to unrelated-to-the-PR upstream version changes. Here's an example of 2 tests currently failing on Python 3.7 & Python 3.8:

```
=================================== FAILURES ===================================
____________ TestKerasWord2VecWrapper.testEmbeddingLayer20NewsGroup ____________
[...]

Issue #2873, opened 2020-07-06
Further focus/slim keyedvectors.py module


gojomo:
Pre-#2698, `keyedvectors.py` was 2500+ lines, including functionality over-specific to other models, & redundant classes. Post-#2698, with some added generic functionality, it's still over 1800 lines. 

It should shed some other grab-bag utility functions that have accumulated, & don't logically fit inside the `KeyedVectors` class. 

In particular, the evaluation (analogies, word_ranks) helpers could move to their own module that takes a KV instance as an argument. (If other more-sophisticated evaluations can be contributed, as would be welcome, they should also live alongside those, rather than bloating `KeyedVectors`.)

The `get_keras_embedding` method, as its utilit is narrow to very specific uses, and is conditional on a not-necessarily install package, could go elsewhere too – either a kera-focused utilities module, or even just documentation/example code about how to convert to/from keras from `KeyedVectors. 

Some of the more advanced word-vector-**using** calculations, like 'Word Mover's Distance' or 'Soft Cosine SImilarity', could move to method-specific modules that are then better documented/self-contained/optimized, without bloating the generic 'set of vectors' module. (They might be more discoverable, there, as well.)

And finally, some of the existing calculations could be unified/streamlined (especially the two variants of `most_similar()`, and some of the steps shared by multiple operations). My hope would be the modules is eventually <1000 lines.

piskvorky:
@gojomo do you see this as essential for 4.0.0 = API breaking?

Or can we leave it for a later release?

gojomo:
This is low-risk and not-hard, but also low-priority. 

Making the decision that some things should be relocated would be nicer to do in 4.0.0, along with other "update your imports/code/function-names" changes, but could wait.

piskvorky:
@mpenkov do you feel like taking this up, reshuffling / organizing KeyedVectors?

I already resolved the `get_keras_embedding` part in #2937.

gojomo:
My 1st thoughts would be:

* move evaluation code to its own module
* move WMD calcs to its own module (even if a passthrough method remains here indefinitely); the options/optimizations for WMD might improve more beneficially in its own space (and evolve in parallel with other extensions like 'soft cosine')

Next but lower-priority, some of the other utility methods may be amenable to more reuse. (Perhaps, the `cosmul` as an option to `most_similar`, or some methods being redefined in terms of fewer more-central operations.)

[...]

Issue #2852, opened 2020-06-10
Reduce gensim surface area


mpenkov:
Are there subpackages/submodules that we're not maintaining anymore, and could remove?

- [x] summarization
- [x] HDP
- [x] wordrank
- [x] dependency on pattern
- [x] various wrappers (incl. sklearn) and others.
- [x] simserver documentation
- [x] viz

The goal is to reduce the maintenance burden of the project.

piskvorky:
@mpenkov any other modules to be removed?

I'd like to finalize this even before the pre-release, so the documentation is up-to-date. Thanks.

piskvorky:
`PoincareModel` is another such candidate. Not touched by the recent refactoring and upgrades, it's probably out of date and useless.

mpenkov:
@piskvorky I had a look at Poincare model. Yes, it's old and hasn't changed much recently, but it's fairly well tested (I eyeballed the [tests](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/test/test_poincare.py) and they make sense to me, and they all pass). Do you really want to remove it?

piskvorky:
No, if it's good, we can keep it.

I don't remember many user complaints / bug reports about the Poincare model, although that may be simply because nobody uses it.

mpenkov:
@piskvorky Perhaps we can remove just the visualization subpackage? That doesn't look like it belongs in a library like gensim

https://github.com/RaRe-Technologies/gensim/tree/develop/gensim/viz

piskvorky:
`viz` seems to contain a single module, with two functions related to Poincare embeddings. So if we keep Poincare, maybe move those two functions to its module? (without introducing new dependencies, so import-on-demand)

Or perhaps move those functions to the particular documentation / tutorial that needs `viz`? Or remove them altogether? I really don't know the context, sorry. But it's a silly subpackage for sure.

mpenkov:
The viz stuff is untested and requires plotly as a dependency (which is in setup.py, meaning all installs include it). My suggestion is to remove viz altogether, and remove the dependency. Visualization is too use case-specific, everyone will want to do it differently anyway.

piskvorky:
Yes, but why is it there? Used by some tutorial / notebook / guide?

mpenkov:
As far as I can tell, it isn't used or mentioned anywhere outside of the automatically generated documentation.

piskvorky:
Hm, double weird. I guess we can just remove it then.

[...]

PR #2911, opened 2020-08-03
Added "tensorflow.keras" compatibility in "KeyedVectors.get_keras_embedding" function.

### LINKED ISSUE #2842: KeyedVectors importing keras instead of tensorflow.keras?
Hello there,

I got a super quick question regarding the function `get_keras_embeddings` of KeyedVectors.py. Is there a reason behind requiring keras package to be installed (specifically referring to [these lines](https://github.com/RaRe-Technologies/gensim/blob/2360459e0014f5db8fc587005933a6399efab435/gensim/models/keyedvectors.py#L1415-L1419)) instead of going through tensorflow 2.0 API (i.e., `import tensorflow.keras`)?

andrewsapw:
Since TensorFlow 2.0+, Keras is the part of TensorFlow library, so I added "from tensorflow.keras.layers import Embedding" as possible import to "Embedding" layer (the old way to import is allowed too) in "get_keras_embedding" function.

Fixes #2842 and closes #2729.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -1605,17 +1605,23 @@ class KeyedVectors(utils.SaveLoad):
         Raises
         ------
         ImportError
-            If `Keras <https://pypi.org/project/Keras/>`_ not installed.
+            If `Keras <https://pypi.org/project/Keras/>`_ or
+            `TensorFlow 2.0+ <https://pypi.org/project/tensorflow/>`_ not installed.
 
         Warnings
         --------
-        Current method work only if `Keras <https://pypi.org/project/Keras/>`_ installed.
+        Current method work only if `Keras <https://pypi.org/project/Keras/>`_ or
+        `TensorFlow 2.0+ <https://pypi.org/project/tensorflow/>`_ installed.
 
         """
         try:
             from keras.layers import Embedding
         except ImportError:
-            raise ImportError("Please install Keras to use this function")
+            try:
+                from tensorflow.keras.layers import Embedding
+            except ImportError:
+                raise ImportError("Please install Keras or TensorFlow 2.0+ to use this function")
+
         weights = self.vectors
 
         # set `trainable` as `False` to use the pretrained word embedding


piskvorky:
Thanks! A duplicate of #2729 (and #2842) but those are stale, so let's continue here.

gojomo:
For the reasons listed in #2886, and in the spirit of #2873 (slim this module) & #2852 (reduce gensim surface area), I'd prefer to remove this one-line-of-recipe code from `KeyedVectors`.

It could easily live as a utility function elsewhere, or just be part of some example code in some doc/tutorial/notebook. 

It's not inherent to the `KeyedVectors` type, and conditional-importing inside a method is a pretty bad 'code smell' indicating this is out-of-place.

piskvorky:
That is true, though mostly orthogonal to this PR.

@andrewsapw as a user, where would you expect this function to live, were would you look for it first?

andrewsapw:
> 
> 
> That is true, though mostly orthogonal to this PR. We want both {modernize import, move fnc elsewhere}.
> 
> @andrewsapw as a user, where would you expect this function to live, were would you look for it first?
> What's your use-case when calling it?
I agree that `KeyedVectors` if pretty big, and because of that it's not always obvious how to use it. On the other hand, this function is actual helpful for newbies, so mb `gensim.utils` is the right place for such small functions.
Regarding tutorials, it's hard to navigate there: plenty of notebooks and all kind of files (.png, .gif, .md and so on) in one place, so i think it can be reorganized to be more friendly to new people, maybe I can help with that.

piskvorky:
Are you talking about the offical tutorials, https://radimrehurek.com/gensim/auto_examples/index.html ?

Or what .png, .gif files?

andrewsapw:
I meant https://github.com/RaRe-Technologies/gensim/tree/develop/docs/notebooks

gojomo:
If it needs a module, maybe a new `gensim.keras` or `gensim.integrations.keras`. But really, since the functionality is just...

https://github.com/RaRe-Technologies/gensim/blob/4b7e372cf289b1bfa154db28a042618f96e91226/gensim/models/keyedvectors.py#L1619-L1627

...putting it in a "example of using gensim word-vectors with Keras" wiki page, or demo-notebook, would be just as understandable/discoverable as the existing utility function.

Issue #2743, opened 2020-02-02
Word2vec: total loss depends on number of workers


tsaastam:
#### Problem description

The word2vec implementation requires a workaround, as detailed in #2735, to correctly report the total loss per epoch. After doing that though, the next issue is that the total loss reported seems to vary depending on the number of workers.

#### Steps/code/corpus to reproduce

This is my code:

    class MyLossCalculatorII(CallbackAny2Vec):
        def __init__(self):
            self.epoch = 1
            self.losses = []
            self.cumu_loss = 0.0
            self.previous_epoch_time = time.time()

        def on_epoch_end(self, model):
            loss = model.get_latest_training_loss()
            norms = [linalg.norm(v) for v in model.wv.vectors]
            now = time.time()
            epoch_seconds = now - self.previous_epoch_time
            self.previous_epoch_time = now
            self.cumu_loss += float(loss)
            print(f"Loss after epoch {self.epoch}: {loss} (cumulative loss so far: {self.cumu_loss}) "+\
                  f"-> epoch took {round(epoch_seconds, 2)} s - vector norms min/avg/max: "+\
                  f"{round(float(min(norms)), 2)}, {round(float(sum(norms)/len(norms)), 2)}, {round(float(max(norms)), 2)}")
            self.epoch += 1
            self.losses.append(float(loss))
            model.running_training_loss = 0.0

    def train_and_check(my_sentences, my_epochs, my_workers=8, my_loss_calc_class=MyLossCalculatorII):
        print(f"Building vocab...")
        my_model: Word2Vec = Word2Vec(sg=1, compute_loss=True, workers=my_workers)
        my_model.build_vocab(my_sentences)
        print(f"Vocab done. Training model for {my_epochs} epochs, with {my_workers} workers...")
        loss_calc = my_loss_calc_class()
        trained_word_count, raw_word_count = my_model.train(my_sentences, total_examples=my_model.corpus_count, compute_loss=True,
                                                            epochs=my_epochs, callbacks=[loss_calc])
        loss = loss_calc.losses[-1]
        print(trained_word_count, raw_word_count, loss)
        loss_df = pd.DataFrame({"training loss": loss_calc.losses})
        loss_df.plot(color="blue")
    #    print(f"Calculating accuracy...")
    #    acc, details = my_model.wv.evaluate_word_analogies(questions_file, case_insensitive=True)
    #    print(acc)
        return loss_calc, my_model

[...]

Issue #2735, opened 2020-01-26
Word2vec: cumulative loss maxes out at 134217728.0


tsaastam:
#### Cumulative loss of word2vec maxes out at 134217728.0

I'm training a word2vec model with 2,793,404 sentences / 33,499,912 words, vocabulary size 162,253 (words with at least 5 occurrences).

Expected behaviour: with `compute_loss=True`, gensim's word2vec should compute the loss in the expected way.

Actual behaviour: the cumulative loss seems to be maxing out at `134217728.0`:

    Building vocab...
    Vocab done. Training model for 120 epochs, with 16 workers...
    Loss after epoch 1: 16162246.0 / cumulative loss: 16162246.0
    Loss after epoch 2: 11594642.0 / cumulative loss: 27756888.0

    [ - snip - ]

    Loss after epoch 110: 570688.0 / cumulative loss: 133002056.0
    Loss after epoch 111: 564448.0 / cumulative loss: 133566504.0
    Loss after epoch 112: 557848.0 / cumulative loss: 134124352.0
    Loss after epoch 113: 93376.0 / cumulative loss: 134217728.0
    Loss after epoch 114: 0.0 / cumulative loss: 134217728.0
    Loss after epoch 115: 0.0 / cumulative loss: 134217728.0

And it stays at `134217728.0` thereafter. The value `134217728.0` is of course exactly `128*1024*1024`, which does not seem like a coincidence.

#### Steps to reproduce

My code is as follows:

    class MyLossCalculator(CallbackAny2Vec):
        def __init__(self):
            self.epoch = 1
            self.losses = []
            self.cumu_losses = []

        def on_epoch_end(self, model):
            cumu_loss = model.get_latest_training_loss()
            loss = cumu_loss if self.epoch <= 1 else cumu_loss - self.cumu_losses[-1]
            print(f"Loss after epoch {self.epoch}: {loss} / cumulative loss: {cumu_loss}")
            self.epoch += 1
            self.losses.append(loss)
            self.cumu_losses.append(cumu_loss)

    def train_and_check(my_sentences, my_epochs, my_workers=8):
        print(f"Building vocab...")
        my_model: Word2Vec = Word2Vec(sg=1, compute_loss=True, workers=my_workers)
        my_model.build_vocab(my_sentences)
        print(f"Vocab done. Training model for {my_epochs} epochs, with {my_workers} workers...")
        loss_calc = MyLossCalculator()
[...]

Issue #2617, opened 2019-10-01
Fix, improve, complete 'training loss' computation for *2Vec models


gojomo:
Word2Vec training-loss isn't quite yet the epoch-based loss most would expect – as pending PR #2135 might address – but also `Doc2Vec` and `FastText` should offer functional, analogous reporting, and the docs should make clear what this loss is good for (monitoring training progress) and what it's not good for (assessing overall model fitness for downstream tasks). 

(Loss for `Doc2Vec` looks like it might be there due to inherited interfaces, and was requested along with `Word2Vec` as in #1272, but that request was closed as a duplicate of #999, which wound up only implementing it for `Word2Vec`.)

gojomo:
In addition to adding loss tallying where it's missing (FastText, Doc2Vec)...

To address the potential multithreading issues (#2743), each thread should have its own loss-tally, only combined safely at the end of an epoch. 

To address the precision issue of #2735, wider types should be used as appropriate. But also tallying the loss from a single call/batch into a local var first, before adding to a larger running total (that's potentially much larger and thus in lower-precision ranges of the floating-point implementation) could also help. And splitting tallying per thread, as above, could help as well. 

Ensuring there's an easy way to get a loss summary from a single training-batch (or non-training inference in the case of Doc2Vec) might offer new/improved ways of doing a "does this text match this model's expectation" calculation, which might enable new uses (and/or replacing the old 'scoring' feature @mataddy added to some Word2Vec modes long ago). 

Potentially even offering a way to tally loss *per word* (or other model aspect), if low-overhead, could also enable new insight into whether different parts of a model are relatively undertrained compared to others, or warnings when parts of a model are updated a lot without any updates to others (in the case of incremental training), or even dynamic choice of learning-rate per epoch or per word (as in Adagrad/etc). 

Having loss-tracking really working might also allow a mode that avoids any explicit/fixed choice of `epochs`, just a "train 'til converged" option.

fabriciorsf:
O loss value doesn't change over epochs.

gojomo:
> O loss value doesn't change over epochs.

You've not provided enough context to know what you're talking about. Are you confiming things already discussed in this and other issues? Reporting something new? What models/methods were you using?

PR #2922, opened 2020-08-24
[early WIP] Fix/rationalize loss-tallying

gojomo:
PR to eventually address loss-tallying issues: #2617, #2735, #2743. Early tinkering stage.


gojomo:
Changes so far in `Word2Vec`: 

* using `float64` for all loss tallying 
* resetting tally to 0.0 per epoch - but remembering history elsewhere for duration of current `train()` call
* micro-tallying into a per-batch value rather than the global tally
* then, adding to global tally rather than replacing it

Though the real goal is sensible loss-tallying across all classes, I think these small changes already remedy #2735 (float32 swallows large loss-values) & #2743 (worker losses clobber each other). 

An oddity from looking at per-epoch loss across a full run: all my `hs` runs have shown **increasing** loss every epoch, which makes no sense to me. And yet, the models at the end have moved word-vectors to more useful places (thus passing our minimal sanity-tests). I don't think my small changes could have caused this oddity (but maybe); I suspect something pre-existing in HS-mode loss-tallying is the real reason. When I have a chance I'll compare to the loss patterns for similar-modes/similar-data in something like the Facebook FastText code, that also reports running loss.

gojomo:
Training FB `fasttext` (HS, CBOW, no-ngrams `./fasttext cbow -verbose 5 -maxn 0 -bucket 0 -lr 0.025 -loss hs -thread 3 -input ~/Documents/Dev/gensim/enwik9 -output enwik9-cbow-nongrams-lr025-hs`) shows decreasing loss reports through the course of training, as expected and unlike the strangely-increasing per-epoch loss our code (at least in this PR) reports. But, final results on a few quick `most_similar` ops seem very similar. So something remains odd about our loss reporting, especially in HS mode.

gojomo:
As a point of comparison, Facebook's `fasttext` reports an "average loss", divided over some trial-count, like so:

```
(base) gojomo@Gobuntu-2020:~/Documents/Dev/fasttext/fastText-0.9.2$ time ./fasttext cbow -verbose 5 -maxn 0 -bucket 0 -lr 0.025 -loss hs -thread 3 -input ~/Documents/Dev/gensim/enwik9 -output enwik9-cbow-nongrams-lr025-hs
Read 142M words
Number of words:  847816
Number of labels: 0
Progress:  39.8% words/sec/thread:  431099 lr:  0.015052 avg.loss:  5.263475 ETA:   0h 5m31s
Progress:  45.4% words/sec/thread:  429306 lr:  0.013645 avg.loss:  4.725245 ETA:   0h 5m 1s
Progress:  58.6% words/sec/thread:  426932 lr:  0.010339 avg.loss:  3.865230 ETA:   0h 3m50s
Progress: 100.0% words/sec/thread:  422384 lr:  0.000000 avg.loss:  2.483185 ETA:   0h 0m 0s
```

Gensim should probably collect & report 2Vec-class training loss in a comparable way, so that numbers on algorithmically-analogous runs are broadly similar, for familiarity to users & as a cross-check of whatever it is we're doing.

piskvorky:
+1 on matching FB's logic. What is "trial-count", and is the average taken over words or something else?

gojomo:
Unsure; their c++ (with a separate class for 'loss') is different enough from our code that I couldn't tell at-a-glance & will need to study it a bit more.

piskvorky:
@gojomo cleaning up the loss-tallying logic still very much welcome. We're planning to make a Gensim release soon – whether this PR gets in now or later, it will be a great addition.

gojomo:
These changes would likely apply, & help a bit in `Word2Vec`, with just a little adaptation to current `develop`. I could take a look this week & wouldn't expect any complications.

But getting consistent loss-tallying working in `Doc2Vec` & `FastText`, & ensuring a similar calculation & roughly similar loss magnitudes with other libraries (mainly Facebook FastText), would require more, & hard-to-estimate, effort. We kind of need someone who both – (1) needs it; & (2) can get deep into understanding the code – to rationalize the whole thing. 

Never figured out why our `hs` mode reports growing loss despite the model improving as expected on other checks.

PR #2923, opened 2020-08-24
Eventually better-handle loading of unsupported-mode FastText files

gojomo:
I noticed our code for loading FT models shoehorns some of FastText's not-supported-in-gensim modes into our model shapes, without showing any error/warning. This might lead to weird behavior down the road, & should probably be improved to be more explicit about what's happening. For now, this is just a code-comment noting the existing not-quite-right behavior.

### DIFF
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -1038,6 +1038,12 @@ def _load_fasttext_format(model_file, encoding='utf-8', full_model=True):
         window=m.ws,
         epochs=m.epoch,
         negative=m.neg,
+        # FIXME: these next 2 lines read in unsupported FB FT modes (loss=3 softmax or loss=4 onevsall,
+        # or model=3 supervised) possibly creating inconsistent gensim model likely to fail later. Displaying
+        # clear error/warning with explanatory message would be far better - even if there might be some reason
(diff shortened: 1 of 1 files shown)


mpenkov:
I think it's better to capture this as a separate issue instead of a TODO. @piskvorky WDYT?

piskvorky:
Well yes, and I see this PR as that ticket.

The current `FIXME` comment is a stub, to highlight the issue – or is this meant to be merged as-is @gojomo ?

gojomo:
This could be merged as-is just to capture the knowledge/acknowledgment of the issue near the code. (I created this because I'd looked a bit into this, & wrote the comment, & didn't want the info lost – but also didn't want this side-matter distracting from the other more functional work-in-progress in other PRs.) Sure, there can be an issue too!

mpenkov:
Opened https://github.com/RaRe-Technologies/gensim/issues/3179, so this PR can be closed, because:

- Nobody ever reads TODOs/FIXMEs in code, so merging this offers little benefit
- Having this an an issue makes it easier for people to contribute fixes. New contributors typically look for new, easy issues to work on as opposed to existing PR, which look like existing work has already been done and needs additional understanding.

PR #2926, opened 2020-08-28
Change num_words to topn in dtm_coherence

### LINKED ISSUE #2925: Change parameter used in dtm_coherence() (in DTM wrapper) to avoid persistent warning
Within the DTM wrapper, using `dtm_coherence()` always produces a warning:

> "The parameter `num_words` is deprecated, will be removed in 4.0.0, use `topn` instead."

Although the function works, obviously the intention is to deprecate the parameter at some point. 

This can be tracked back to `show_topic()`, which is where `num_words` has switched to `topn`. It is a very simple fix to just change the parameter used in the `dtm_coherence()` function accordingly. 

PR incoming with fix.

MeganStodel:
Fixes #2925 

This is a very small change to make `dtm_coherence()` use `topn` instead of `num_words` when calling `show_topic()`. This then avoids triggering an error every time the function is used.

### DIFF
--- a/gensim/models/wrappers/dtmmodel.py
+++ b/gensim/models/wrappers/dtmmodel.py
@@ -604,7 +604,7 @@ class DtmModel(utils.SaveLoad):
         """
         coherence_topics = []
         for topic_no in range(0, self.num_topics):
-            topic = self.show_topic(topicid=topic_no, time=time, num_words=num_words)
+            topic = self.show_topic(topicid=topic_no, time=time, topn=num_words)
             coherence_topic = []
             for prob, word in topic:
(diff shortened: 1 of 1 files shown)


piskvorky:
> This then avoids triggering an error every time the function is used.

Does that happen? The warning should be issued only once, not every time the function is used.

MeganStodel:
> The warning should be issued only once (once per whole application run), not every time the function is used.

My experience of using the function within notebooks is that it is issued every time `dtm_coherence()` is used - I'm not sure if it would differ is used in other contexts. This fix should still avoid that happening if it's once or repeatedly though.

Issue #2928, opened 2020-08-31
The type information about job_queue in word2vec is wrong


lunastera:
#### Problem description

`_worker_loop` and `_job_producer` say that the `job_queue` element is a `(list of object, dict of (str, int))` type, when in fact it appears to be a `(list of object, float)` type.
Is this a statement for a future change, or is it a mistake?

#### Steps/code/corpus to reproduce

```python
def _worker_loop(self, job_queue, progress_queue):
        """Train the model, lifting batches of data from the queue.

        This function will be called in parallel by multiple workers (threads or processes) to make
        optimal use of multicore machines.

        Parameters
        ----------
        job_queue : Queue of (list of objects, (str, int))
            A queue of jobs still to be processed. The worker will take up jobs from this queue.
            Each job is represented by a tuple where the first element is the corpus chunk to be processed and
            the second is the dictionary of parameters.
```

but

```python
def _job_producer(self, data_iterator, job_queue, cur_epoch=0, total_examples=None, total_words=None):
        ...
        next_job_params = self._get_job_params(cur_epoch)
        job_no = 0

        for data_idx, data in enumerate(data_iterator):
            data_length = self._raw_word_count([data])

            # can we fit this sentence into the existing job batch?
            if batch_size + data_length <= self.batch_words:
                # yes => add it to the current job
                job_batch.append(data)
                batch_size += data_length
            else:
                job_no += 1
                job_queue.put((job_batch, next_job_params))
```

In `_get_job_params`

```python
def _get_job_params(self, cur_epoch):
        """Get the learning rate used in the current epoch.

        Parameters
        ----------
        cur_epoch : int
            Current iteration through the corpus

        Returns
        -------
        float
            The learning rate for this epoch (it is linearly reduced with epochs from `self.alpha` to `self.min_alpha`).

        """
        alpha = self.alpha - ((self.alpha - self.min_alpha) * float(cur_epoch) / self.epochs)
        return alpha
```

#### Versions

```python
Darwin-19.6.0-x86_64-i386-64bit
Python 3.7.4 (default, Jan 24 2020, 20:34:38)
[Clang 11.0.0 (clang-1100.0.33.16)]
Bits 64
NumPy 1.19.1
SciPy 1.5.2
gensim 4.0.0.dev0
```

gojomo:
[...]

PR #2931, opened 2020-09-02
Clear up job queue parameters in word2vec

lunastera:
Fixes: #2928

Small `job_queue` related changes.

- Changed type information from `(list of object, (str, int))` (or `(list of object, (str, float))`) to `(list of object, float)`.
- The `_get_job_params` and `_update_job_params` are not properly named and have been replaced by `_get_current_alpha` and `_update_alpha`, respectively.
- Names like `job_params` have been replaced with `alpha` .
- The `_update_job_params` had `job_params` passed to it, which has been removed because it wasn't being used.

PS:
- Since `_get_current_alpha` and `_update_alpha` do essentially the same thing, they have been merged into `_get_next_alpha`.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -1045,10 +1045,10 @@ class Word2Vec(utils.SaveLoad):
 
         Parameters
         ----------
-        job_queue : Queue of (list of objects, (str, int))
+        job_queue : Queue of (list of objects, float)
             A queue of jobs still to be processed. The worker will take up jobs from this queue.
             Each job is represented by a tuple where the first element is the corpus chunk to be processed and
-            the second is the dictionary of parameters.
+            the second is the floating-point learning rate.
         progress_queue : Queue of (int, int, int)
             A queue of progress reports. Each report is represented as a tuple of these 3 elements:
                 * Size of data chunk processed, for example number of sentences in the corpus chunk.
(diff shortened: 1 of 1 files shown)


lunastera:
The py36-win test didn't pass, so I put back the unused argument and only changed the names.
Why is this argument only needed in the py36-win environment?

gojomo:
If you look at the actual test failure in the logs, it was not related to your changes. It's a test with some randomness that's occasionally failed, so that test-run was unlucky, and the later succeeding run was lucky, neither affected by your changes. (I just added some changes that may make it less flaky in #2930 - if you re-base on current `develop` it should minimize further distracting unrelated errors.)

gojomo:
Looks like an improvement! But, having now looked at this more closely, the two different (current/update) methods seem redundant: they're both doing essentially the same thing. Want to take a try at going further, unifying them into one `_get_next_alpha()` method, that (at the outset) just considers current-epoch-progress as `0.0`, & maybe even has simpler but consistent calculation-logic?

gensim/models/word2vec.py
```diff
@@ -1359,12 +1359,12 @@ def _get_job_params(self, cur_epoch):
...
         return alpha
 
-    def _update_job_params(self, job_params, epoch_progress, cur_epoch):
+    def _update_alpha(self, alpha, epoch_progress, cur_epoch):
         """Get the correct learning rate for the next iteration.
 
         Parameters
         ----------
-        job_params : dict of (str, obj)
+        alpha : float
```
mpenkov, line 1367: If this is an unused parameter, and the method is internal, perhaps it would be safe to get rid of the parameter?

lunastera:
@gojomo I see! I understand about the test failing.
Also, thanks for the further specific suggestions for improvement! That's certainly true, so I'll fix this as well.

gensim/models/word2vec.py
lunastera: I will include a fix here as well.

gojomo:
Looks good to me! (I'll leave it to either @mpenkov or @piskvorky to do final approval/application.)

piskvorky (approved):
Looks good.
Do any of the [tutorials](https://github.com/RaRe-Technologies/gensim/tree/develop/docs/src) need updating after this API change? Did you / could you run a grep please?

lunastera:
Thanks for the confirmation! I did a grep on the changed variables and functions, but I didn't find any changes.

piskvorky:
Awesome. Thanks :)

PR #2935, opened 2020-09-08
Fix "generator" language in word2vec docs

### LINKED ISSUE #2934: Word2Vec docs indicate sents can be a generator, but it can't
#### Problem description

Tried to pass a generator as the `sentences` argument when training Word2Vec, as suggested [in the docs](https://radimrehurek.com/gensim/models/word2vec.html#usage-examples:~:text=sentences%20can%20be%20a%20generator). But then I get this error:

    TypeError: You can't pass a generator as the sentences argument. Try a sequence.

```

def get_sents():
    for ii in range(100):
        yield ['hi']

sents = get_sents()
model = Word2Vec(sents, size=100, window=5, min_count=1, workers=4)
```

#### Versions

Please provide the output of:

```python
import struct; print("Bits", 8 * struct.calcsize("P"))
```
```
Linux-5.8.5-arch1-1-x86_64-with-glibc2.2.5
Python 3.8.5 (default, Jul 27 2020, 08:42:51) 
[GCC 10.1.0]
Bits 64
NumPy 1.19.1
SciPy 1.4.1
gensim 3.8.3
FAST_VERSION 1
```

polm:
Docs say you can use a generator as the first argument, but you can't.

The tempfile path was also unused, so that's been removed.

Fixes #2934.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -39,18 +39,15 @@ Initialize a model with e.g.:
 
 .. sourcecode:: pycon
 
-    >>> from gensim.test.utils import common_texts, get_tmpfile
+    >>> from gensim.test.utils import common_texts
     >>> from gensim.models import Word2Vec
-    >>>
-    >>> path = get_tmpfile("word2vec.model")
     >>>
     >>> model = Word2Vec(common_texts, size=100, window=5, min_count=1, workers=4)
     >>> model.save("word2vec.model")
 
-The training is streamed, meaning `sentences` can be a generator, reading input data
-from disk on-the-fly, without loading the entire corpus into RAM.
+`common_texts` is a list of lists of strings, where each list is a document.
 
-It also means you can continue training the model later:
+If you save the model you can continue training it later:
 
 .. sourcecode:: pycon
 
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -45,7 +45,14 @@ Initialize a model with e.g.:
     >>> model = Word2Vec(common_texts, size=100, window=5, min_count=1, workers=4)
     >>> model.save("word2vec.model")
 
-`common_texts` is a list of lists of strings, where each list is a document.
+
+The training is streamed, so `sentences` can be an iterable, reading input data
+from disk on-the-fly. This lets you avoid loading the entire corpus into RAM.
+However, note that because the iterable must be re-startable, `sentences` must
+not be a generator. For an example of an appropriate iterator see
+:class:`~gensim.models.word2vec.BrownCorpus`,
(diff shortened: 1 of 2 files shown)


gensim/models/word2vec.py
```diff
@@ -39,18 +39,15 @@
...
     >>> from gensim.models import Word2Vec
     >>>
-    >>> path = get_tmpfile("word2vec.model")
-    >>>
     >>> model = Word2Vec(common_texts, size=100, window=5, min_count=1, workers=4)
     >>> model.save("word2vec.model")
 
-The training is streamed, meaning `sentences` can be a generator, reading input data
-from disk on-the-fly, without loading the entire corpus into RAM.
+`common_texts` is a list of lists of strings, where each list is a document.
```
piskvorky, line 48: This is not good – the streaming is essential. We've had many instances where people read the docs and assume the input **must** be a list, which is not the case.

Please keep the streaming (iterable) language in, just drop the generator.

piskvorky (changes requested):

gensim/models/word2vec.py
polm: Ah OK, I'll fix that.

polm:
For reference, it looks like people have been confused about this for years :/ 

https://stackoverflow.com/questions/34166369/generator-is-not-an-iterator

piskvorky:
Yes. There was a point in history where a generator was really accepted (and it still is in other Gensim models), which doesn't help.

Thanks for fixing.

piskvorky:
@mpenkov can you check that the updated docs render OK? I still cannot access CircleCI, it's asking me for full access to my Github account + my soul.

piskvorky (approved):

mpenkov:
Yes, the generated docs seem to be rendering just fine here: https://3588-1349775-gh.circle-artifacts.com/0/documentation/html/index.html

gensim/models/word2vec.py
```diff
@@ -39,18 +39,22 @@
...
-    >>> path = get_tmpfile("word2vec.model")
-    >>>
     >>> model = Word2Vec(common_texts, size=100, window=5, min_count=1, workers=4)
     >>> model.save("word2vec.model")
 
-The training is streamed, meaning `sentences` can be a generator, reading input data
-from disk on-the-fly, without loading the entire corpus into RAM.
 
-It also means you can continue training the model later:
+The training is streamed, so `sentences` can be an iterable, reading input data
```
mpenkov, line 49:
```suggestion
The training is streamed, so ``sentences`` can be an iterable, reading input data
```

mpenkov:
Merged. Thank you for your contribution @polm !

PR #2937, opened 2020-09-08
Remove Keras dependency

### LINKED ISSUE #2842: KeyedVectors importing keras instead of tensorflow.keras?

### LINKED ISSUE #2908: Re-enable tests for Keras integration
The Keras tests started failing on py3.7 + 3.8 in #2899:

https://travis-ci.org/github/RaRe-Technologies/gensim/jobs/713448950#L862

None of the changes in that PR seem related to these errors, so probably some TF / Keras weirdness again.

**Task**: Look into it and either fix, or rip out that entire Keras submodule. One of the contributed / never properly vetted features. Or Keras is that unstable… same difference.

piskvorky:
Keras / Tensorflow is a wonky dependency that was messing up our tests. At the same time, as noted by @gojomo in https://github.com/RaRe-Technologies/gensim/pull/2911#issuecomment-668825647, Gensim only ever uses Keras in a single function: 3 lines total to export trained word2vec vectors into a Keras layer.

This PR removes Keras from Gensim altogether. Instead, the same 3 lines were moved to FAQ, to be used (and maintained, updated) by those who actually need them:
https://github.com/RaRe-Technologies/gensim/wiki/Recipes-&-FAQ#q13-how-do-i-export-a-trained-word2vec-model-to-keras

Fixes #2911. 
Fixes #2842. 
Fixes #2908.


gojomo:
Glad to see this addressed! 

If there's likely to be example code using the old method, it could be retained as a stub but throw something like a `NotImplementedError` with a message pointing people at the wiki answer. 

I'd make the help a separate wiki page devoted to "using Gensim Word2Vec with Keras". That's easier to create a stable, long-lived link to. (Then if some other questions become more frequent, this bit of help doesn't need to stay "Q13" forever.) 

Also, each additional question in a single long FAQ about some random (not-really-frequently-asked) topic makes the info that really frequently-asked harder to skim over or edit/expand. (Long pages with mixtures of dissimilar info are much less useful than focused-topic pages.) And, a dedicated Keras help page might then grow to include any other related Keras-specific hints, far more easily than one section of a long page.

gojomo:
Also: the notebook at <https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/keras_wrapper.ipynb> could be updated with the recipe code, and/or moved (perhaps to some other broader-themed set of demos/cookbook exmaples), and/or deleted - as it's currently still using the removed method.

piskvorky:
> the notebook at https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/keras_wrapper.ipynb

That's weird. I did a grep and didn't find this. Thanks for pointing out this notebook. @mpenkov what's our plan with this directory as a whole?

A separate wiki page makes sense, this is not really a "frequently asked question". I'll create it.

piskvorky:
I tried to update that notebook, but failed to install tensorflow. It's giving me some `Symbol not found: _SecKeyCopyExternalRepresentation` error, and googling shows it's not really supported on my Mac.

I'm not going to upgrade my laptop just to install that, so unless someone steps up to update (and test) `keras_wrapper.ipynb`, I'll simply remove it from the gensim repo.

piskvorky:
Done: https://github.com/RaRe-Technologies/gensim/wiki/Using-Gensim-Word2Vec-with-Keras-and-Tensorflow

mpenkov:
> > the notebook at https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/keras_wrapper.ipynb
> 
> That's weird. I did a grep and didn't find this. Thanks for pointing out this notebook. @mpenkov what's our plan with this directory as a whole?

I'm not sure there is a legit plan for that subdir. It appears to be mostly unmaintained notebooks (with some exceptions, I remember @Witiko did some good work there). I think the plan should be something like:

1. Gradually remove stuff that is stale (as you've done in this PR)
2. Migrate working notebooks to the sphinx gallery, where it makes sense
3. Keep the notebooks as-is if they are a bad fit for the gallery

@gojomo WDYT?

mpenkov (approved):

gojomo:
I like the idea of a grab-bag of different demo/tutorial notebooks, demonstrating different techniques, including projects that takes hours/days to run. I prefer the original IPYNB format to the extra complexities & limitations of the Sphinx presentation. 

Such notebooks could be better organized/named, with some standard text or code boilerplate that helps identify the versions where they were last run or any extra assumptions (like other packages) they might depend upon. 

I may be mistaken, but I had thought that previously, even a `pip install gensim` would replicate the `docs` & `docs/notebooks` into an environment's gensim install directory - and thought that was also a good idea, so every user had those locally for reference/experimentation. That doesn't seem to be the case now. 

Such demo/tutorial IPYNBs could be moved to another github project.

Issue #2887, opened 2020-07-19
Measure performance of gensim 4.0.0 vs previous versions


mpenkov:
Not every 1-line decision; just ones that are in inner loops of hot-spot code.

Definitely a big TODO: compare performance before/after.

_Originally posted by @piskvorky in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0MzQ5Mjk1NTk1/timeline/more_items_

Fixes #2960.

piskvorky:
This link is also broken. @mpenkov this way of creating tickets seems more trouble than worth, with the context missing.

mpenkov:
I'll take it up with github support. It's convenient, but only when it works.

mpenkov:
@piskvorky From github support:

> This might be an uncaught edge case on our end. I have raised this up with our engineering team to investigate further.

I'll keep my eyes open for the problem in case it recurs.

piskvorky:
Some Word2vec measurements here: https://github.com/RaRe-Technologies/gensim/pull/2939#issuecomment-691729995

piskvorky:
Comparing current `develop` (at ea87470e) against 3.8.3:
(identical training params, identical HW, text9 corpus, measured with [benchmark.py](https://gist.github.com/piskvorky/85015d52254df77cd23ca43bc2e8322d))

### fasttext 3.8.3

training on a 124301826 raw words (88163974 effective words) took 107.3s, 821794 effective words/s
2:53.89 elapsed
4318564k max mem

### fasttext develop

training on a 124301826 raw words (88166519 effective words) took 96.4s, 914282 effective words/s
2:17.43 elapsed
1318592k max mem

### word2vec 3.8.3

training on a 124301826 raw words (88162276 effective words) took 52.3s, 1684982 effective words/s
1:41.36 elapsed
373612k max mem

### word2vec develop

training on a 124301826 raw words (88166114 effective words) took 50.0s, 1762436 effective words/s
1:13.82 elapsed
348060k max mem

PR #2939, opened 2020-09-09
[MRG] *2Vec SaveLoad improvements

piskvorky:
Continued from #2892 (subsumes #2892):

1. Move model serialization fixups in `save()` / `load()` into `_save_specials()`/`_load_specials()`, to better handle cases when one `SaveLoad` contains another.
2. Code style & py3 migration fixes.
3. Resolved mysterious bloat in the pickle file of pre-4.0-dev Word2Vec models.

This PR does **not** include additional changes around serialization:

- ensure there's at least one full test load of gensim-3.8.3 FastText/Word2Vec/Doc2Vec models that might not yet be tested
- double-checking mmap options are functionally tested across 2Vec models, & fixing anything that might turn up
- delete all tests of loading no-longer-supported older models – ideally, everything earlier than gensim-3.8.3 - and then potentially discarding support code thus no-longer needed

These will come in separate PRs later. In particular, we have to decide what to do with compatibility for models trained and stored pre-3.8.3.


piskvorky:
> resolve whatever mysterious bloat is in the pickle file of pre-4.0-dev Word2Vec models

@gojomo I checked the pickle sizes and they seem as expected. Do you have a reproducing example?

One potential pitfall is that `SaveLoad` stores the main pickle under `protocol=2`, whereas "native" pickle in Python 3 protocol 3 upward (3, 4, or 5). Could that be the difference you saw?

gensim/models/keyedvectors.py
```diff
@@ -395,9 +395,8 @@ def add_one(self, key, vector):
         Warning: using this repeatedly is inefficient, requiring a full reallocation & copy,
         if this instance hasn't been preallocated to be ready fro such incremental additions.
 
-        returns: actual index used TODO: other param docs
+        returns: actual index used FIXME: other param docs
```
mpenkov, line 398: I think it's better to make GitHub issues for TODOs and FIXMEs:

- Unless we're explicitly grepping for TODO/FIXME/whatever, this thing is essentially invisible to developers.
- This text will appear in our publicly readable docs. It's not a good look.

gensim/models/word2vec.py
```diff
@@ -131,23 +131,17 @@
 from collections import defaultdict, namedtuple
 from types import GeneratorType
 import threading
-import itertools
+import itertools as it
```
mpenkov, line 134: -1 on this.

Pros: it makes things slightly more brief.
Cons: it's not a universally accepted abbreviation or convention, so it can make the code more difficult to read.

The pros don't outweigh the cons in my opinion.

gensim/models/keyedvectors.py
piskvorky: In general yes. Explicitly grepping for FIXMEs will be a crucial task before 4.0.0 though – I suspect there will still be several.

(Not) appearing in public docs is a feature – we never want FIXMEs in any release code.

gensim/models/word2vec.py
```diff
@@ -1837,49 +1825,65 @@ def load(cls, *args, rethrow=False, **kwargs):
...
+        if not hasattr(self, 'ns_exponent'):
+            self.ns_exponent = 0.75
+        if self.negative and hasattr(self.wv, 'index_to_key'):
+            self.make_cum_table()  # rebuild cum_table from vocabulary
+        if not hasattr(self, 'corpus_count'):
+            self.corpus_count = None
+        if not hasattr(self, 'corpus_total_words'):
+            self.corpus_total_words = None
+        if not hasattr(self.wv, 'vectors_lockf') and hasattr(self.wv, 'vectors'):
+            self.wv.vectors_lockf = getattr(self, 'vectors_lockf', np.ones(1, dtype=REAL))
```
mpenkov, line 1851: Why do we need to use getattr here? We already know that those attrs don't exist from the conditional on the line above.

```suggestion
            self.wv.vectors_lockf = np.ones(1, dtype=REAL)
```

gensim/models/word2vec.py
piskvorky: I agree. See the original PR for a discussion with @gojomo – he prefers `it` because it's more idiomatic.

gojomo: I increasingly see `it` as a common alias for `itertools`, and as the use of the individual functions becomes more prevalent/natural/assumed, having to either – (1) add/remove each function by name in imports as they move in/out of use; or (2) prepend `itertools` everywhere– both add cognitive load & repetitive typing compared to just adopting `import itertools as it`.

Similarly, a lot of our older code imports each function used from `numpy` individually - which given the absolute dominance of the `import numpy as np` convention throughout our domain, just added maintenance overhead & extra-indirection in reading code. I've been gradually using `np.` instead, but the transition is not complete.

gojomo:
> > resolve whatever mysterious bloat is in the pickle file of pre-4.0-dev Word2Vec models
> 
> @gojomo I checked the pickle sizes and they seem as expected. Do you have a reproducing example?
> 
> One potential pitfall is that `SaveLoad` stores the main pickle under `protocol=2`, whereas "native" pickle in Python 3 protocol 3 upward (3, 4, or 5). The `protocol=2` files are larger. Could that be the difference you saw?

When I last ran the exact same `Word2Vec` job (on something like `text8` or `text9`) in 3.8.3 vs `develop`, the gensim `.save()` file from `develop` was significantly larger, suggestive of something unnecessary being serialized. Do you not see that in your tests?

piskvorky:
OK, I'll double check tmr.

piskvorky:
Yes, I can replicate on `text8`:

### 3.8.3

27034172 bytes in pickle (after `training on a 66515395 raw words (49739954 effective words) took 96.0s, 518211 effective words/s`).

### 4.0.0

68765086 bytes in pickle (after `training on a 66515395 raw words (49737330 effective words) took 79.3s, 627379 effective words/s`).

Timings are for #2887; trained with `word2vec.Word2Vec(word2vec.LineSentence('/tmp/enwik8'), size=100, window=5, min_count=2, workers=8)`.

gensim/models/word2vec.py
mpenkov: > both add cognitive load & repetitive typing compared to just adopting import itertools as it

Yes, it's additional cognitive load, but only for one person - the writer of the code, and only once - when the code is written. The alias adds cognitive load for many people - whoever needs to read the code, and for many times over - because code is read far more often than it is written.

gojomo: I believe it's harder to read without consistent alias use, as well. 

In the case where the functions are imported into the module scope (as happens in gensim code of a certain vintage, especially `from itertools import chain` but there have been others), the function name may not be unambiguously from `itertools` - so a reader may have to search for the definition or import to understand, and adding/removing usages is onerous. 

The case where they're always preceded by the full `itertools.` prefix is somewhat better, but sometimes that causes common combined to more-often wrap lines, requiring the reader to parse indention-conventions. Consistently using `it.` everywhere only requires the reader to realize one thing – "this codebase uses it=itertools convention" – and from that moment on, code is both compact & readable.

gojomo:
I'd especially suspect the backward-compatibility `@property` aliases `index2word` or `index2entity`, though even if pickle thinks those need serialization I'm not sure why they'd not use a proper ref. 

I also just noticed that the `KeyedVectors._load_specials()` tries a `pop` of `index2word` twice where one attempt should be `index2entity`, though that couldn't be related to save-expansion. And, the back-compatibility setter for `vectors_norm` should probably trigger a loud error, like its getter (or the way getter & setter for `.vocab` errors). Perhaps `index2word` and `index2entity` should also loud-error to trigger old-code-updates as with `.vocab` and `.vectors_norm`.

piskvorky:
This is the core of it:

```python
In [88]: type(model.wv.index_to_key[0])
Out[88]: numpy.str_
```

I'll check where "numpy strings" creep in.

piskvorky:
OK I think I found it. Have to run now, will clean up the `pop` & `index2word` etc later.

piskvorky:
@gojomo fixed in 06aef751. Please review – there's still one outstanding FIXME left in `keyedvectors.py`, I wasn't sure how you meant `# FIXME: Update to better match & share code with most_similar()`.

gensim/models/word2vec.py
piskvorky: I'm also in favour of explicit `itertools.xyz`, not a fan of 2-letter acronyms. Especially ones that coincide with common English words.

How strongly do you feel about this @gojomo ?

piskvorky:
I'm still not sure about superficial breaking changes like `size` => `vector_size` or `index2word` => `index_to_key`.

Do we really have to annoy users with that? I'll add that to the list for "IRL discussion", probably best handled & decided live.

gojomo:
> I'm still not sure about superficial breaking changes like `size` => `vector_size` or `index2word` => `index_to_key`. Do we really have to annoy users with that?
> 
> I'll add that to the list for "IRL discussion", probably best handled & decided live.

Previous discussion of `vector_size`, https://github.com/RaRe-Technologies/gensim/pull/2698#discussion_r445749873 , where I thought this concern was explained and settled. 

#1777 changed some, but not all, uses of `size` for vector-dimensionality to `vector_size`. Having both in use indefinitely & having code of different eras or closely-related classes prefer one or the other is worse than either one. I thought `size` was OK, but it does invite confusion with other things that might be a `size`, like total count of vectors, so I went with `vector_size` everywhere.

The `blah2blah` naming was always of questionable Pythonicity - no underscores? Abbreviating 'to' as '2'? 

`KeyedVectors` is in its core intentionally **not** word-specific – both when used for purposes like `Doc2Vec` where the lookup keys aren't 'words', and even when specifically used by the `Word2Vec` algorithm because many times, the user's 'words' aren't truly words in the training domain. So for consistency with the class name, and to lessen the dissonance of always referring to non-word things as 'words', a more generic name was appropriate. 

But #1777 & other contemporaneous work starting using 'entity' as the generic term in lots of places, over the objections of [myself](https://github.com/RaRe-Technologies/gensim/pull/1957#issuecomment-373142528) & [other reviewers](https://github.com/RaRe-Technologies/gensim/pull/1777#issuecomment-353003664). ('Entity' is incredibly generic, and doesn't match the class name: it's `KeyedVectors`, highlighting the central fact that they are accessed by unique keys, not `EntityVectors`. Neither the key nor the vector is vividly described as an 'entity'.)

The incomplete & ill-considered shift toward `enitity` also meant the implementing classes had a mix of properties named **either** `index2word` or `index2entity` - and in that confusion, some classes made one the alias of the other, and in some cases a superclass used one as the actual property while the subclass used another, leading to duplication/confusion where some instances had both properties, with one an empty-list.  It was very poor design, implementation, and review. 

Using `index_to_key` and `key_to_index` is Pythonic, consistent across all users of the class, and consistent with the class-name `KeyedVectors`. Also, it emphasizes the essential inverse relationship between `index_to_key` and `key_to_index` (formerly the messy `vocab` dict): they are exactly each other's reverse-mappings, and always between keys and int indexes. (While not yet guaranteed, ideally the keys could be not just strings but any valid, hashable Python objects, such as tuples or ints.)

gojomo:
I'd also note that where it's practical & the underlying structure remains unchanged (an ordered list of lookup keys), the latest changes retain aliases for backward-compatible access: 

https://github.com/RaRe-Technologies/gensim/blob/08a61e5eb6395ca4f03fa0bee9b2845f983fc3de/gensim/models/keyedvectors.py#L552-L566

Retaining `size` as an alias in `__init__()` methods would be possible but more onerous, as every method signature would then include redundant names for the same parameter, and every init method would include temporary boilerplate for using one if the other isn't present. "Ripping off the bandaid" now, at the same time as other changes, but still just the small step of a few mechanistic parameter replacements, is no worse, & probably better, than doing it later. 

The biggest pain to any advanced API users is likely to be the removal of the `.vocab` dict, where each value was an object amenable to arbitrary expansion. The `get_vecattr()`/`set_vecattr()` mechanism offers that same potential, in a more compact & performant way, but is sufficiently different that learning/adopting the new system will require more attention to old code than just a search-and-replace names. I also expect a slightly larger 'bugtail' around that feature, as it's likely that out-in-wild patterns-of-use that we don't have covered in unit tests will exercise things in unexpected ways. But it's likely to be somewhat more capable programmers who are doing that, and the wins in generality/speed/memory-efficiency justify the one-time pain, in my opinion.

piskvorky:
> the latest changes retain aliases for backward-compatible access:

I actually changed this in recent commits, following:

> Perhaps index2word and index2entity should also loud-error to trigger old-code-updates as with .vocab and .vectors_norm.

I was thinking whether to keep those in there or not… but with so many breaking changes, I also figured it is best to "rip off the bandaid" rather than introduce deprecations (again).

> The biggest pain to any advanced API users is likely to be the removal of the .vocab dict,

Can you put the rationale into https://github.com/RaRe-Technologies/gensim/wiki/Migrating-from-Gensim-3.x-to-4 ? Even if we discussed it somewhere already, it'll be useful to have all the changes and their back-story in one place. Thanks!

piskvorky:
> the wins in generality/speed/memory-efficiency [of `expandos`] justify the one-time pain, in my opinion.

If `expandos` is backed by numpy arrays, won't it suffer from the same memory issues as `numpy.str_`? Numpy arrays are great for primitive types (ints, floats, etc), but not the best choice for arbitrary objects.

gojomo:
If you've added hard-fails for those aliases, which I think is alright, that's in some still-pending PR. 

I'll add an overview of the rationale on the wiki page. 

I was unfamiliar with that limitation on `string`-typed numpy arrays! I'd seen `object`-typed numpy arrays, & had assumed Python strings were similarly flexibly handled, without fixed-length allocations. The type-imputation code should be changed to work around that, probably by using `object` type. The most prevalent extra attributes *are* simple numeric types (like ints or floats). In the other cases I can think of, they may also be ragged-sized other arrays of various types, as with HS-mode `codes`/`points`. I can't think of a current use where an actual string is needed, but the intent is that it should be supported – and I suspect both the ragged-sized arrays & strings will just best be served by using `object` type instead.

piskvorky:
Yes, `object` dtype should be safer wrt memory, that stores just the references. OTOH we'll lose the type safety of "all array values are of the same type"… but I guess nobody expects that from Python anyway.

> If you've added hard-fails for those aliases, which I think is alright, that's in some still-pending PR.

I added them to this PR:
https://github.com/RaRe-Technologies/gensim/pull/2939/files#diff-741f69634ff2e1b2c4ce544e0d82a1e6R616

piskvorky:
@mpenkov ping – please review & merge, so we can move on.

gensim/models/word2vec.py
piskvorky: @gojomo ping. Unless this is vital for you I'll revert to spelling out `itertools` in full, I concur with @mpenkov.

gensim/models/fasttext.py
```diff
@@ -850,25 +850,15 @@ def load(cls, *args, **kwargs):
...
-        # fixup mistakenly overdimensioned gensim-3.x lockf arrays
-        if len(model.wv.vectors_vocab_lockf.shape) > 1:
-            model.wv.vectors_vocab_lockf = ones(1, dtype=REAL)
-        if len(model.wv.vectors_ngrams_lockf.shape) > 1:
-            model.wv.vectors_ngrams_lockf = ones(1, dtype=REAL)
-        if hasattr(model, 'bucket'):
-            del model.bucket  # should only exist in one place: the wv subcomponent
-        if not hasattr(model.wv, 'buckets_word') or not model.wv.buckets_word:
-            model.wv.recalc_char_ngram_buckets()
+        return super(FastText, cls).load(*args, rethrow=True, **kwargs)
```
gojomo, line 853: Do we really need such stub methods that only calls the same superclass method with the same arguments? That's already the default which occurs if no method is present. By my understanding, doc-comment tools like Sphinx will, in their current versions, already propagate superclass API docs down to subclasses.

The only thing that's varying is the comment, and while it expresses a different expected-type from the superclass, in practice that doc may be misleading: I **think** (but have not recently checked) that these `SaveLoad` `.load()` methods can return objects that **may not be** what the caller expects. They return **the class that's in the file**, not the class-that-`.load()`-was-called-on.

If so, it might be a worthwhile short-term step as soon as 4.0.0 – for limiting the risk of confusion & requirement for redundant/caveat-filled docs – to **deprecate the practice of calling *SpecificClass*.load(filename) entirely**, despite its common appearance in previously-idiomatic gensim example code. Instead, either (1) call it only on class `SaveLoad` itself, to express that the only expectation for the returned type is that it's a `SaveLoad` subclass; (2) promote load functionality to model-specific top-level functions in each relevant model – a bit more like the `load_facebook_model()` function for loading Facebook-FasttText-native-format models – which might themselves do some type-checking, so any docs which imply they return a certain type are true; (3) just make one `utils.py` generic `load()`, perhaps with an optional class-enforcement parameter, and encourage its use. 

Despite the effort involved in making such changes, they could minimize duplicated code/comments & avoid some unintuitive gotchas in the current `SaveLoad` approach.

gensim/models/fasttext.py
```diff
@@ -1202,12 +1192,49 @@ def __init__(self, vector_size, min_n, max_n, bucket):
...
+        :class:`~gensim.models.fasttext.FastTextKeyedVectors`
+            Loaded model.
+
+        See Also
+        --------
+        :meth:`~gensim.models.fasttext.FastTextKeyedVectors.save`
+            Save :class:`~gensim.models.fasttext.FastTextKeyedVectors` model.
+
+        """
+        return super(FastTextKeyedVectors, cls).load(fname_or_handle, **kwargs)
```
gojomo, line 1213: Same issue as above.

gensim/models/keyedvectors.py
```diff
@@ -1814,15 +1887,15 @@ def _load_word2vec_format(cls, fname, fvocab=None, binary=False, encoding='utf8'
...
-    """Alias for `KeyedVectors.load_word2vec_format(...)`"""
+    """Alias for :meth:`~gensim.models.keyedvectors.KeyedVectors.load_word2vec_format`."""
     return KeyedVectors.load_word2vec_format(*args, **kwargs)
 
 
 def pseudorandom_weak_vector(size, seed_string=None, hashfxn=hash):
-    """Get a 'random' vector (but deterministically derived from seed_string if supplied).
+    """Get a random vector, derived deterministically from `seed_string` if supplied.
+
+    Useful for initializing KeyedVectors that will be the starting projection/input layers of _2Vec models.
```
gojomo, line 1897: FYI/ this per-slot pseudorandom initialization is no longer used in #2944's changes, and thus this method will likely disappear entirely.

gensim/models/word2vec.py
gojomo: It's not vital to me either way. I find `it` a helpful convention, I've seen it in a growing number of places, and I think that `itertools` is central-enough to idiomatic Python programming that I expect this conventions' prevalence to grow over time, eventually reaching the same status as `np`. But it's ok to wait/follow here rather than lead!

gensim/models/word2vec.py
```diff
@@ -1841,49 +1908,65 @@ def load(cls, *args, rethrow=False, **kwargs):
...
+        Returns
+        -------
+        float
+            Current training loss.
+
+        """
+        return self.running_training_loss
+
 
 class BrownCorpus(object):
```
gojomo, line 1971: Not a high-priority at all, but it'd be more sensible for such a tutorial/testing utility corpus to be implemented elsewhere - maybe under `/test/` or some other data- or doc- related module – rather than in `gensim.models.word2vec`.

gojomo:
Happy to see the more-specific and -tangible `add_vector`/`add_vectors` renamings, & improved doc-comments! My inline comments aren't things that should hold this up, just related concerns to discuss/address at some point.

gensim/models/fasttext.py
piskvorky: `SpecificClass.load()` is there to allow custom logic overrides, IIRC. To clean up backward compatibility from loaded objects etc. So exactly your item `(2) promote load functionality to model-specific top-level functions in each relevant model`?

Note I didn't make any changes to this code in this PR, the stubs must be coming from elsewhere (older).

gensim/models/keyedvectors.py
piskvorky: OK. Let's merge this ASAP, I fear divergence. CC @mpenkov.

gensim/models/word2vec.py
piskvorky: I agree. This belongs under `gensim.corpora` – not sure why it's here.

Although I vaguely recall some discussion around this, a long time ago… I wish the code had more motivating (as opposed to technical) comments!

gensim/models/fasttext.py
piskvorky: > unintuitive gotchas in the current SaveLoad approach

What are they? I was under the impression `save`/`load` was straightforward. Were users complaining?

mpenkov (commented):
Reviewed this together with @piskvorky earlier today.

In general, I think the changes are good to merge, but there are several things that could be better.

I think we should document `expandos` in code for the benefit of future developers and maintainers. So, add a multi-line code comment explaining what it is, what it does, and why it does things this way.

The same thing goes for `specials`.

It looks like the above information is available from comments in this PR (and others), but it's highly unlikely that anyone wanting to learn more about `expandos` and `specials` will know to look there.

gensim/models/word2vec.py
piskvorky: Created a ticket for this: https://github.com/RaRe-Technologies/gensim/issues/2956

gensim/models/word2vec.py
piskvorky: OK. Renamed to `itertools` for now, in da8847a0.

piskvorky:
My notes from our call with @mpenkov ; FYI @gojomo :

# Decisions

1. Support loading of pre-3.8.3 models?
  - Misha: better: a stand-alone script to upgrade models to the latest version… rather than class methods doing the work on each load
  - Misha: also in favour of limited backward compatibility loading (just 3.8.3)
  - Radim: in favour of keeping existing compatibility code; why remove it? pros/cons unclear
  => **decision**: talk to Gordon, make the costs (pros/cons) more explicit.
2. `itertools`vs `it`?
  - prefer full name unless critical to Gordon, but w/e.
3. Remove subpar packages, contributed modules, wrappers?
   - yes; Misha to review & axe
   - what about NMF? semi-ready, but not much confidence, and its author no longer interested in finishing/maintaining.
      - options: 1. axe it; 2. invest effort & finish & fix; 3. do nothing, keep as-is => **decision** keep 4.0.0, Radim looks at option (2) later after 4.0.0

# Actions

- **G**: Finish the model serialization tickets: mmap, save_facebook bug
- **R**: Finish all docs:
   - migration guide (wiki)
   - release notes
   - docstrings
   - migrate & re-run tutorials and guides
   - proof-read generated API docs for readability: we don't want stubs or obsolete/compatibility classes and methods like `word2vec.Word2VecVocab`, `word2vec.Word2VecTrainables` etc.
      - esp. with the new website style: correct syntax highlighting, headings, sections
   - internal code style: hanging indents, trailing commas, FIXME vs TODO, etc
- **M**: Finish structure cleanup: remove subpackages and modules we don't want, incl. their docs

@mpenkov @gojomo did I forget something?

gensim/models/fasttext.py
piskvorky: Moved to a separate ticket, so this doesn't get lost: #2957.

piskvorky:
> I think we should document expandos in code for the benefit of future developers and maintainers. So, add a multi-line code comment explaining what it is, what it does, and why it does things this way.

@mpenkov I improved the comments in 782f7ff9e0cdd53832f93300b712a19b95782608 . I didn't want to start a new PR for that, so it's a part of #2954 , where I'm fixing docs anyway.

gojomo:
It's a bit awkward to discuss release todos/criteria deep in a PR that's already been merged/closed. I'd suggest creating a specific 'rollup' issue that's for meta-discussion, and in/out decisions, around the "4.0.0" release. 

My main hope would be to get more eyes on the big changes via one or more "not yet for everyone" test releases, as further described in new issue #2963.

PR #2944, opened 2020-09-12
[MRG] fix save_facebook_model failure after update-vocab & other initialization streamlining

### LINKED ISSUE #2853: save_facebook_model() - AssertionError
#### Problem description
I am trying to save the trained model of fasttext using the new save_facebook_model function.
I was unable to do it so because an assertionError arises in the code line:
`assert vocab_n == len(model.wv.vocab)`
 The vocabulary of my model is of 2000264:
```
len(model.wv.vocab)
2000264
```
I tried with a model with a vocabulary of 4500 and it worked. So I guess there is a limitation in that. But the error message did not tell any of that.

```
from gensim.models.fasttext import save_facebook_model

save_facebook_model(model,'own_fasttext_model_pretrained.bin')

AssertionError                            Traceback (most recent call last)
<ipython-input-201-0a3c1c458b74> in <module>
      2 from gensim.models.fasttext import load_facebook_model, load_facebook_vectors,save_facebook_model
      3 
----> 4 save_facebook_model(model,'own_fasttext_model_pretrained.bin')

/opt/conda/lib/python3.7/site-packages/gensim/models/fasttext.py in save_facebook_model(model, path, encoding, lr_update_rate, word_ngrams)
   1334     """
   1335     fb_fasttext_parameters = {"lr_update_rate": lr_update_rate, "word_ngrams": word_ngrams}
-> 1336     gensim.models._fasttext_bin.save(model, path, fb_fasttext_parameters, encoding)

/opt/conda/lib/python3.7/site-packages/gensim/models/_fasttext_bin.py in save(model, fout, fb_fasttext_parameters, encoding)
    666     if isinstance(fout, str):
    667         with open(fout, "wb") as fout_stream:
--> 668             _save_to_stream(model, fout_stream, fb_fasttext_parameters, encoding)
    669     else:
    670         _save_to_stream(model, fout, fb_fasttext_parameters, encoding)

/opt/conda/lib/python3.7/site-packages/gensim/models/_fasttext_bin.py in _save_to_stream(model, fout, fb_fasttext_parameters, encoding)
    629 
    630     # Save words and ngrams vectors
--> 631     _input_save(fout, model)
    632     fout.write(struct.pack('@?', False))  # Save 'quot_', which is False for unsupervised models
    633 

/opt/conda/lib/python3.7/site-packages/gensim/models/_fasttext_bin.py in _input_save(fout, model)
    573 
    574     assert vocab_dim == ngrams_dim
--> 575     assert vocab_n == len(model.wv.vocab)
    576     assert ngrams_n == model.wv.bucket
    577 

AssertionError: 
```

#### Versions

```
Python 3.7.6 | packaged by conda-forge | (default, Jan  7 2020, 22:33:48) 
[GCC 7.3.0] on linux
Linux-4.9.0-12-amd64-x86_64-with-debian-9.12
Python 3.7.6 | packaged by conda-forge | (default, Jan  7 2020, 22:33:48) 
[GCC 7.3.0]
NumPy 1.18.1
SciPy 1.4.1
gensim 3.8.3
FAST_VERSION 0
```

### LINKED ISSUE #2943: Issue with calculation of `new_words` and `pre_exist_words`
https://github.com/RaRe-Technologies/gensim/blob/bb947b38a23f10c40948ec6876bc0a850eafbcf6/gensim/models/word2vec.py#L602

`new_words` and `pre_exist_words` are initialized by reference as follows:
```python
new_words = pre_exist_words = []
```
This has a result every time an append happens to one of the two, both variables change because they refer to the same underlying list. Therefore, the calculation of `new_words` and `pre_exist_words` is not correct.

The solution would simply be:
```python
new_words = []
pre_exist_words = []
```

This has downstream implications. The calculation of `retain_words` (`retain_words = new_words + pre_exist_words`)  is also not correct because of that.

gojomo:
Fixes #2853 & fixes #2943. 

Refactors a bunch of `FastText`/`Word2Vec`/`Doc2Vec` initialization & update code for clarity & less-duplication. Each of these classes now share the faster bulk style of vector-random-initalization previously only `FastText` used, a slight change in prior per-vector seeding behavior. 

Seems to fix the test code for triggering ancient `Doc2Vec` update crash in #1019 - but there may still be problems there, and there's been no design/testing for `build_vocab(..., update=True)` on `Doc2Vec`, so current status there is still "not supported".

There was some purported (but possibly not useful, and not covered by unit-tests) `mmap_path`-specifying code in `KeyedVectors` that's been stubbed out here with FIXMEs/comments, pending intended (#2955) repair/rationalization of that functionality before 4.0.0 final release.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -286,6 +286,9 @@ class Doc2Vec(Word2Vec):
 
         self.vector_size = vector_size
         self.dv = dv or KeyedVectors(self.vector_size, mapfile_path=dv_mapfile)
+        # EXPERIMENTAL lockf feature; create minimal no-op lockf arrays (1 element of 1.0)
+        # advanced users should directly resize/adjust as desired after any vocab growth
+        self.dv.vectors_lockf = np.ones(1, dtype=REAL)  # 0.0 values suppress word-backprop-updates; 1.0 allows
 
         super(Doc2Vec, self).__init__(
             sentences=corpus_iterable,
@@ -330,11 +333,9 @@ class Doc2Vec(Word2Vec):
         self.wv.norms = None
         self.dv.norms = None
 
-    def reset_weights(self):
-        super(Doc2Vec, self).reset_weights()
-        self.dv.resize_vectors()
-        self.dv.randomly_initialize_vectors()
-        self.dv.vectors_lockf = np.ones(1, dtype=REAL)  # 0.0 values suppress word-backprop-updates; 1.0 allows
+    def init_weights(self):
+        super(Doc2Vec, self).init_weights()
+        self.dv.resize_vectors(seed=self.seed)
 
     def reset_from(self, other_model):
         """Copy shareable data structures from another (possibly pre-trained) model.
@@ -359,7 +360,7 @@ class Doc2Vec(Word2Vec):
         self.dv.key_to_index = other_model.dv.key_to_index
         self.dv.index_to_key = other_model.dv.index_to_key
         self.dv.expandos = other_model.dv.expandos
-        self.reset_weights()
+        self.init_weights()
 
     def _do_train_epoch(self, corpus_file, thread_id, offset, cython_vocab, thread_private_mem, cur_epoch,
                         total_examples=None, total_words=None, offsets=None, start_doctags=None, **kwargs):
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -82,8 +82,6 @@ from gensim.models.keyedvectors import KeyedVectors, pseudorandom_weak_vector
 
 logger = logging.getLogger(__name__)
 
-SEED_DIVERSIFIER = 7919  # 1000th prime
-
 try:
     from gensim.models.doc2vec_inner import train_document_dbow, train_document_dm, train_document_dm_concat
 except ImportError:
@@ -337,7 +335,8 @@ class Doc2Vec(Word2Vec):
 
     def init_weights(self):
         super(Doc2Vec, self).init_weights()
-        self.dv.resize_vectors(seed=self.seed + SEED_DIVERSIFIER)  # don't use identical rnd stream as words
+        # to not use an identical rnd stream as words, deterministically change seed (w/ 1000th prime)
+        self.dv.resize_vectors(seed=self.seed + 7919)
 
     def reset_from(self, other_model):
         """Copy shareable data structures from another (possibly pre-trained) model.
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -82,6 +82,8 @@ from gensim.models.keyedvectors import KeyedVectors, pseudorandom_weak_vector
 
 logger = logging.getLogger(__name__)
 
+SEED_DIVERSIFIER = 7919  # 1000th prime
+
 try:
     from gensim.models.doc2vec_inner import train_document_dbow, train_document_dm, train_document_dm_concat
 except ImportError:
@@ -335,7 +337,7 @@ class Doc2Vec(Word2Vec):
 
     def init_weights(self):
         super(Doc2Vec, self).init_weights()
-        self.dv.resize_vectors(seed=self.seed)
+        self.dv.resize_vectors(seed=self.seed + SEED_DIVERSIFIER)  # don't use identical rnd stream as words
 
     def reset_from(self, other_model):
         """Copy shareable data structures from another (possibly pre-trained) model.
--- a/gensim/models/fasttext.py
+++ b/gensim/models/fasttext.py
@@ -261,7 +261,7 @@ from numpy import ones, vstack, float32 as REAL
 
 import gensim.models._fasttext_bin
 from gensim.models.word2vec import Word2Vec
-from gensim.models.keyedvectors import KeyedVectors
+from gensim.models.keyedvectors import KeyedVectors, prep_vectors
 from gensim import utils
 from gensim.utils import deprecated
 try:
@@ -432,7 +432,10 @@ class FastText(Word2Vec):
             bucket = 0
 
         self.wv = FastTextKeyedVectors(vector_size, min_n, max_n, bucket)
-        self.wv.bucket = bucket
+        # EXPERIMENTAL lockf feature; create minimal no-op lockf arrays (1 element of 1.0)
+        # advanced users should directly resize/adjust as desired after any vocab growth
+        self.wv.vectors_vocab_lockf = ones(1, dtype=REAL)
+        self.wv.vectors_ngrams_lockf = ones(1, dtype=REAL)
 
         super(FastText, self).__init__(
             sentences=sentences, corpus_file=corpus_file, workers=workers, vector_size=vector_size, epochs=epochs,
@@ -441,29 +444,6 @@ class FastText(Word2Vec):
             min_count=min_count, sample=sample, sorted_vocab=sorted_vocab,
             null_word=null_word, ns_exponent=ns_exponent, hashfxn=hashfxn,
             seed=seed, hs=hs, negative=negative, cbow_mean=cbow_mean, min_alpha=min_alpha)
-
-    def prepare_weights(self, update=False):
-        """In addition to superclass allocations, compute ngrams of all words present in vocabulary.
-
-        Parameters
-        ----------
-        update : bool
-            If True, the new vocab words and their new ngrams word vectors are initialized
-            with random uniform distribution and updated/added to the existing vocab word and ngram vectors.
-        """
-        super(FastText, self).prepare_weights(update=update)
-        if not update:
-            self.wv.init_ngrams_weights(self.seed)
-            # EXPERIMENTAL lockf feature; create minimal no-op lockf arrays (1 element of 1.0)
-            # advanced users should directly resize/adjust as necessary
-            self.wv.vectors_vocab_lockf = ones(1, dtype=REAL)
-            self.wv.vectors_ngrams_lockf = ones(1, dtype=REAL)
-        else:
-            self.wv.update_ngrams_weights(self.seed, self.old_vocab_len)
-            # EXPERIMENTAL lockf feature; create minimal no-op lockf arrays (1 element of 1.0)
-            # advanced users should directly resize/adjust as necessary
-            self.wv.vectors_vocab_lockf = ones(1, dtype=REAL)
-            self.wv.vectors_ngrams_lockf = ones(1, dtype=REAL)
 
     def _init_post_load(self, hidden_output):
         num_vectors = len(self.wv.vectors)
@@ -484,85 +464,6 @@ class FastText(Word2Vec):
             self.syn1neg = hidden_output
 
         self.layer1_size = vector_size
-
-    def build_vocab(self, corpus_iterable=None, corpus_file=None, update=False, progress_per=10000,
-                    keep_raw_vocab=False, trim_rule=None, **kwargs):
-        """Build vocabulary from a sequence of sentences (can be a once-only generator stream).
-        Each sentence must be a list of unicode strings.
-
-        Parameters
-        ----------
-        corpus_iterable : iterable of list of str, optional
-            Can be simply a list of lists of tokens, but for larger corpora,
-            consider an iterable that streams the sentences directly from disk/network.
-            See :class:`~gensim.models.word2vec.BrownCorpus`, :class:`~gensim.models.word2vec.Text8Corpus`
-            or :class:`~gensim.models.word2vec.LineSentence` in :mod:`~gensim.models.word2vec` module for such examples.
-        corpus_file : str, optional
-            Path to a corpus file in :class:`~gensim.models.word2vec.LineSentence` format.
-            You may use this argument instead of `sentences` to get performance boost. Only one of `sentences` or
-            `corpus_file` arguments need to be passed (not both of them).
-        update : bool
-            If true, the new words in `sentences` will be added to model's vocab.
-        progress_per : int
-            Indicates how many words to process before showing/updating the progress.
-        keep_raw_vocab : bool
-            If not true, delete the raw vocabulary after the scaling is done and free up RAM.
-        trim_rule : function, optional
-            Vocabulary trimming rule, specifies whether certain words should remain in the vocabulary,
-            be trimmed away, or handled using the default (discard if word count < min_count).
-            Can be None (min_count will be used, look to :func:`~gensim.utils.keep_vocab_item`),
-            or a callable that accepts parameters (word, count, min_count) and returns either
-            :attr:`gensim.utils.RULE_DISCARD`, :attr:`gensim.utils.RULE_KEEP` or :attr:`gensim.utils.RULE_DEFAULT`.
-            The rule, if given, is only used to prune vocabulary during
-            :meth:`~gensim.models.fasttext.FastText.build_vocab` and is not stored as part of the model.
-
-            The input parameters are of the following types:
-                * `word` (str) - the word we are examining
-                * `count` (int) - the word's frequency count in the corpus
-                * `min_count` (int) - the minimum count threshold.
-
-        **kwargs
-            Additional key word parameters passed to
-            :meth:`~gensim.models.word2vec.Word2Vec.build_vocab`.
-
-        Examples
-        --------
-        Train a model and update vocab for online training:
-
-        .. sourcecode:: pycon
-
-            >>> from gensim.models import FastText
-            >>> sentences_1 = [["cat", "say", "meow"], ["dog", "say", "woof"]]
-            >>> sentences_2 = [["dude", "say", "wazzup!"]]
-            >>>
-            >>> model = FastText(min_count=1)
-            >>> model.build_vocab(sentences_1)
-            >>> model.train(sentences_1, total_examples=model.corpus_count, epochs=model.epochs)
-            >>>
-            >>> model.build_vocab(sentences_2, update=True)
-            >>> model.train(sentences_2, total_examples=model.corpus_count, epochs=model.epochs)
-
-        """
-        if not update:
-            self.wv.init_ngrams_weights(self.seed)
-        elif not len(self.wv):
-            raise RuntimeError(
-                "You cannot do an online vocabulary-update of a model which has no prior vocabulary. "
-                "First build the vocabulary of your model with a corpus "
-                "by calling the gensim.models.fasttext.FastText.build_vocab method "
-                "before doing an online update."
-            )
-        else:
-            self.old_vocab_len = len(self.wv)
-
-        retval = super(FastText, self).build_vocab(
-            corpus_iterable=corpus_iterable, corpus_file=corpus_file, update=update, progress_per=progress_per,
-            keep_raw_vocab=keep_raw_vocab, trim_rule=trim_rule, **kwargs)
-
-        if update:
-            self.wv.update_ngrams_weights(self.seed, self.old_vocab_len)
-
-        return retval
 
     def _clear_post_train(self):
         """Clear the model's internal structures after training has finished to free up RAM."""
@@ -1112,7 +1013,7 @@ def save_facebook_model(model, path, encoding="utf-8", lr_update_rate=100, word_
 
 
 class FastTextKeyedVectors(KeyedVectors):
-    def __init__(self, vector_size, min_n, max_n, bucket):
+    def __init__(self, vector_size, min_n, max_n, bucket, count=0, dtype=REAL):
         """Vectors and vocab for :class:`~gensim.models.fasttext.FastText`.
 
         Implements significant parts of the FastText algorithm.  For example,
@@ -1158,12 +1059,12 @@ class FastTextKeyedVectors(KeyedVectors):
 
         """
         super(FastTextKeyedVectors, self).__init__(vector_size=vector_size)
-        self.vectors_vocab = None  # fka syn0_vocab
-        self.vectors_ngrams = None  # fka syn0_ngrams
-        self.buckets_word = None
         self.min_n = min_n
         self.max_n = max_n
         self.bucket = bucket  # count of buckets, fka num_ngram_vectors
+        self.buckets_word = None  # precalculated cache of buckets for each word's ngrams
+        self.vectors_vocab = np.zeros((count, vector_size), dtype=dtype)  # fka (formerly known as) syn0_vocab
+        self.vectors_ngrams = None  # must be initialized later
         self.compatible_hash = True
 
     @classmethod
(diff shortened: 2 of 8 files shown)


gojomo:
The paths for first, or post-vocab-expansion, allocation/initialization of `vectors`/`vectors_vocab` were more complex & redundant than necessary - streamlining into fewer methods also cleared up the discrepancy causing the `save_facebook_model` error.

There's a chance this also resolves very-old Doc2Vec-segfault #1019 - the test code there doesn't trigger a segfault anymore, but maybe it was other recent changes that fixed it, or maybe it's just become harder to trigger. 

Making the `Word2Vec` & `Doc2Vec` vector-randomizations share the `FastText` code-path means they all now do the randomization in one bulk call, as FT has been doing, instead of the vector-at-a-time-seeded-by-key that W2V/D2V were doing. This should be faster.

But, at the moment the prior support for memmapped vector-arrays hasn't yet been maintained in this refactor, and the `.uniform()` call used (following the FT precedent) will 1st generate oversized `np.float64`numbers. Fixing those is next.

piskvorky (changes requested):

gensim/models/word2vec.py
```diff
@@ -1392,17 +1389,27 @@ def _raw_word_count(self, job):
...
+            raise TypeError("Either one of corpus_file or corpus_iterable value must be provided")
+        if corpus_file is not None and corpus_iterable is not None:
+            raise TypeError("Both corpus_file and corpus_iterable must not be provided at the same time")
+        if corpus_iterable is None and not os.path.isfile(corpus_file):
+            raise TypeError("Parameter corpus_file must be a valid path to a file, got %r instead" % corpus_file)
+        if corpus_iterable is not None and not isinstance(corpus_iterable, Iterable):
+            raise TypeError("The corpus_iterable must be an iterable of list, got %r instead" % corpus_iterable)
+        if corpus_iterable is not None and isinstance(corpus_iterable, GeneratorType) and passes > 1:
+            raise TypeError(
+                f"Using a generator as corpus_iterable can't support {passes} passes. Try a re-iterable sequence.")
```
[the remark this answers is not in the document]
gojomo: I proposed dropping py3.5 support last December in #2713; you offered enthusiastic support; all py3.5 builds/testing was turned off in January & the issue closed. And, official CPython 3.5 reached its complete EOL a couple days ago, and won't even receive critical security patches.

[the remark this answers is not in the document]
piskvorky: I'm very much for dropping 3.5, yes (f-strings being one of the main reasons).

That hasn't happened yet though; @mpenkov WDYT?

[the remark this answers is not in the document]
gojomo: If either of you think gensim development code has been supporting Py3.5, you shouldn't have approved the January items (#2713 & #2715) that removed Py3.5 from the testing infrastructure. If it's not being tested, it's unlikely to be working. 

More generally, you should just make the reasonable blanket commitment, now & forevermore, that Gensim releases never support a Python that's officially EOL, to avoid constant re-discussion like this when the bulk of the Python world has moved on.

I personally would go even further, and just decide forevermore not to have new releases support any Python that will be EOL within the next 12 months. Laggard projects/organizations can always choose to use an older release (or perform backport work on their own dime). And more generally, orgs that have difficulty upgrading their deployments to stay within ~3 years of the Python feature/fix frontier have deeper problems, that should be solved by highly-paid consultants, not by open-source developers volunteering to make their own maintenance/testing/coding tasks harder.

Alternatively, Gensim could commit to the reasoned and documented policy described in Numpy's [NER29](https://numpy.org/neps/nep-0029-deprecation_policy.html) - which (coincidentally) recommended dropping Py3.5 support for releases starting in December 2019 (about when we did) and recommends dropping Py3.6 support for releases starting in November 2020.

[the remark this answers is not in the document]
piskvorky: 4.0.0 will be the first release to NOT support Python 2.7 (never mind Python 3.5).

What we do with Py3 support going forward is up to us. Not supporting EOLed Pythons makes sense to me (with docs + `setup.py` updated to match).

I don't think we're making much use of any new Python 3.x features either – the switch is more externally-enforced. Once the "no more Python 2" river has been crossed, the version makes little difference in practice IMO beyond sysops.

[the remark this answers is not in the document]
gojomo: Well, I concur with your decisions 9 months ago to drop Python 3.5, and find baffling the re-consideration of the decision as if the prior discussion & tangible actions, which I've linked to for reference, never even happened. Yes, `setup.py` should be updated to sync with the prior actions, too.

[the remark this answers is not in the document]
piskvorky: I checked `setup.py` + official docs and saw Py3.5+. I simply forgot about those old CI PRs, don't read too much into it.

I hope to clean up the overall docs structure a bit, to go with 4.0.0: an explicit split into "for users" / "for contributors" (incl. this release policy, code style, docs style…); deduplicate content from wiki vs sphinx vs github README; better content on the new main page (website index).

gojomo:
When the larger #2939 is merged, I propose rebasing this on that & applying - it fixes the original goal, and then some. 

But, mmap functionality is currently broken in `KeyedVectors`, and likely other 2Vec classes as well - though in ways not covered by tests. That would best be addressed in a separate focused issue/PR - for which I've started #2955.

piskvorky:
#2939 is merged now. @gojomo will you be able to finish this & the broken `mmap`?

I'd love to release this week. Only the documentation items in #2960 (for me) and this ticket (for you, hopefully) are left now.

Plus the discussion around supporting pre-3.8.3 models, but it's best to do that live – see email.

gojomo:
Subject to the provisos...

* this either breaks or simply abandons the pretense that mmapping works for KeyedVectors instances (which will be fixed in a separate PR)
* I had to disable the `pytest-rerunfailures` modules (which I don't think was active) to work around https://github.com/pytest-dev/pytest-rerunfailures/issues/128

...this is ready-for-final-review & merging!

gensim/models/doc2vec.py
```diff
@@ -286,6 +286,9 @@ def __init__(self, documents=None, corpus_file=None, vector_size=100, dm_mean=No
 
         self.vector_size = vector_size
         self.dv = dv or KeyedVectors(self.vector_size, mapfile_path=dv_mapfile)
+        # EXPERIMENTAL lockf feature; create minimal no-op lockf arrays (1 element of 1.0)
+        # advanced users should directly resize/adjust as desired after any vocab growth
```
mpenkov, line 290: Do we have any other documentation for such advanced users, or is this it?

gensim/models/fasttext.py
```diff
@@ -465,29 +466,6 @@ def __init__(self, sentences=None, corpus_file=None, sg=0, hs=0, vector_size=100
             null_word=null_word, ns_exponent=ns_exponent, hashfxn=hashfxn,
             seed=seed, hs=hs, negative=negative, cbow_mean=cbow_mean, min_alpha=min_alpha)
 
-    def prepare_weights(self, update=False):
```
mpenkov, line 468: These deleted methods are part of the public API. If we remove them, then shouldn't we also update documentation that refers to them?

gensim/models/fasttext.py
```diff
@@ -1135,7 +941,7 @@ def save_facebook_model(model, path, encoding="utf-8", lr_update_rate=100, word_
 
 
 class FastTextKeyedVectors(KeyedVectors):
-    def __init__(self, vector_size, min_n, max_n, bucket):
+    def __init__(self, vector_size, min_n, max_n, bucket, count=0, dtype=REAL):
```
mpenkov, line 944: We should document these new parameters.

gensim/models/fasttext.py
```diff
@@ -1338,63 +1144,21 @@ def get_vector(self, word, norm=False):
             else:
                 return word_vec
 
-    def init_ngrams_weights(self, seed):
-        """Initialize the vocabulary and ngrams weights prior to training.
+    def resize_vectors(self, seed=0):
+        """Make underlying vectors match 'index_to_key' size; random-initialize any new rows.
 
-        Creates the weight matrices and initializes them with uniform random values.
+        Unlike in superclass, the 'vectors_vocab' array is of primary importance, with
```
mpenkov, line 1150: I'm not sure if this comment is docstring-worthy. What important information does it convey to the reader? What decision can they now make based on that information?

gensim/models/fasttext.py
```diff
@@ -1338,63 +1144,21 @@ def get_vector(self, word, norm=False):
...
+        'vectors' derived from it. And, the ngrams_vectors may need allocation."""
```
mpenkov, line 1151: This is particularly unhelpful. What does "may need allocation" mean here? It is too ambiguous. Does it mean the user should allocate it? Again, what are we trying to convey here?

gensim/models/keyedvectors.py
```diff
@@ -323,34 +323,16 @@ def get_vecattr(self, key, attr):
...
-        return range(prev_count, target_count)
+    def resize_vectors(self, seed=0):
+        """Make underlying vectors match index_to_key size; random-initialize any new rows."""
 
-    def randomly_initialize_vectors(self, indexes=None, seed=0):
-        """Initialize vectors with low-magnitude random vectors, as is typical for pre-trained
-        Word2Vec and related models.
+        target_shape = (len(self.index_to_key), self.vector_size)
+        self.vectors = prep_vectors(target_shape, prior_vectors=self.vectors, seed=seed)
+        # TODO: support memmap?
```
mpenkov, line 331: In my opinion, if this is a future idea, we should add it as a ticket. If this is dead code, we should remove it. Adding TODOs like this decreases the quality of the code.

gensim/models/keyedvectors.py
```diff
@@ -1902,3 +1884,18 @@ def pseudorandom_weak_vector(size, seed_string=None, hashfxn=hash):
     else:
         once = utils.default_prng
     return (once.random(size).astype(REAL) - 0.5) / size
+
+
+def prep_vectors(target_shape, prior_vectors=None, seed=0, dtype=REAL):
+    """Return a numpy array of the given shape. Reuse prior_vectors values instance or values
+    to extent possible. Initialize new values randomly if requested."""
```
mpenkov, line 1891: This seems to be a public API function. We should document the parameters so they appear in the documentation.

piskvorky (commented):
Is the PR description up to date? W.r.t. functionality contained within, and issues fixed by it.

PR titles and descriptions is what we have to go with for release notes. Plus that's what interested users see when they click through for more details on a change.

gensim/models/doc2vec.py
gojomo: This is it! It's only ever existed as an advanced experiment most users should ignore, and only users comfortable using the source code as documentation should consider using.

gensim/models/fasttext.py
gojomo: Aren't API docs autogenerated from doc-comments, so deleting the method/doc-comment causes the matching docs to change? 

Also, while this superficially looked like a public API (in method name and doc-comment)I don't think it was actually a good idea for anyone to call except via other more prominent methods. It's an internal step that must be called only after a bunch of specific other internal things are done. (It was only called at the end of `build_vocab()` variants, and only appropriate to call for models that had already had all the earlier vocab-establishing steps done just before.)

gensim/models/fasttext.py
gojomo: I'll doc here and in superclass as part of the mapfile work that also touches them.

gensim/models/fasttext.py
gojomo: Ultimately, because users can and may access the `.vectors`/`.vectors_vocab` directly, this is worthwhile to note this dichotomy somewhere in the doc-comments – but yes, the API mention should be at a different level, with other notes in source. 

(This dichotomy, while necessary, is also a bit of a pain to navigate in the mmap-support, as in KV class it's definitely `.vectors` that must be live disk-backed, while in the FTKV class the purely-derived `.vectors` is not even saved separately – so an mmap-load wouldn't necessarily have a file to use. But still, the typical use-case for mmapping – many processes using the same model – would still want to share a `.vectors`.)

gensim/models/fasttext.py
gojomo: Yes, this looks like slop from other spliced-together code comments during the re-org.

gensim/models/keyedvectors.py
gojomo: On this particular case, yes - there's already a pending issue (#2955) for making mmap stuff work as well or better than it used to, before 4.0.0, and these are some of the exact lines where the code is almost certain to change/expand, so this is just temporary ugliness to draw attention here until that's done. 

More generally: I think there are many, many smaller potential improvements or "maybe this could/should change" observations that are much better highlighted with code comments, including intentions like "TODO MAYBE" or "FIXME?", than issues. They may not deserve any current attention, or limited triage bandwidth, via appearing in lists-of-issues. They may only make sense in the context of exact micro-decisions in the code that are a pain to link/restate in issue prose elsewhere. (And thus, they'll be lost if the standard is, "go to Github in a web browser & make a proper issue".) But, they become intensely relevant if and only if someone else (or the original author) is actually reviewing the same lines-of-code.

gensim/models/keyedvectors.py
gojomo: It ultimately may not be, pending the mmap work & other initialization clean-up, which might also jostle the internal names a bit. (At the moment, this is *only* called from `resize_vectors()` which may be the preferable public-entry point, because outside callers are less-likely to have a `prior_vectors`.)

gensim/models/fasttext.py
mpenkov: Yes, the API docs are generated, but what about tutorials/notebooks that call those methods?

piskvorky: Can you please grep the code for other docstrings (maybe in other modules) + notebooks that may link to these methods? I don't think any do, but it's a valid concern.

piskvorky:
@gojomo is it still the case this doesn't fix #2853? Is the PR description up-to-date w.r.t. all the changes within?

gensim/models/fasttext.py
gojomo: A recursive grep of the project directory for the names of the 4 methods removed here, `reset_weights`, `update_weights`, `init_ngrams_weights`, and `update_ngrams_weights` found no external mentions of these on `FastText`. (There's still unaffected mentions, & local implementations, of `prepare_weights` on `Word2Vec` & `Doc2Vec` - but those are independent & I think I'd like to remove those in initialization-cleanup of those classes soon, too.)

gojomo:
Title & top-comment updated to describe what's addressed. Any lingering TODOs/doc-comment/naming stuff will be addressed in a followup PR related to #2955 or others.

gojomo:
So, that one distant test-failing is a real head-scratcher, as noted in the code-comment I've added in 4de243e: 

```python
    rng = np.random.default_rng(seed=seed)  # use new instance of numpy's recommended generator/algorithm               
    # FIXME: `uniform` passes all tests, but generates temporary double-sized np.float64 array,                         
    # then cast-down ito right-sized np.float32, which means momentary 3x RAM usage on the model's                      
    # largest structure (often GB in size)                                                                              
    new_vectors = rng.uniform(-1.0, 1.0, target_shape).astype(dtype)
    # Meanwhile, this alternative, which by docs/reasoning/visual-inspection should be equivalent                       
    # while never creating the unneeded oversized np.float64 array, passes all *2Vec class                              
    # functional tests, but mysteriously (but reliably!) fails one obscure barely-sensible test                         
    # of a fringe downstream functionality: `TestBackMappingTranslationMatric.test_infer_vector`.                       
    # I'd adjust or jettison that test entirely *except* that the failure is *so* reliable, and                         
    # *so* mysterious, that it may be warning of something very subtle. So for now, very briefly,                       
    # sticking with the RAM-wasteful-but-all-tests-passing approach above, TODO debug/fix ASAP.                         
    # new_vectors = rng.random(target_shape, dtype=dtype)  # [0.0, 1.0)                                                 
    # new_vectors *= 2.0  # [0.0, 2.0)                                                                                  
    # new_vectors -= 1.0  # [-1.0, 1.0)                                                                                 
    new_vectors /= vector_size
```

So, in order to get this PR mergeable, I've reverted the code to what works-but-is-inefficient, with that blaring comment commitment to remedy as part of the next PR. (The inefficiency is what `FastText` has been doing all along, and what passes all tests, so no little harm to develop, but still something I'd like to fix before 4.0.0-final.)

gensim/models/keyedvectors.py
```diff
@@ -1918,3 +1912,33 @@ def pseudorandom_weak_vector(size, seed_string=None, hashfxn=hash):
...
+    # of a fringe downstream functionality: `TestBackMappingTranslationMatric.test_infer_vector`.
+    # I'd adjust or jettison that test entirely *except* that the failure is *so* reliable, and
+    # *so* mysterious, that it may be warning of something very subtle. So for now, very briefly,
+    # sticking with the RAM-wasteful-but-all-tests-passing approach above, TODO debug/fix ASAP.
+    # new_vectors = rng.random(target_shape, dtype=dtype)  # [0.0, 1.0)
+    # new_vectors *= 2.0  # [0.0, 2.0)
+    # new_vectors -= 1.0  # [-1.0, 1.0)
+    new_vectors /= vector_size
+    new_vectors[0:prior_vectors.shape[0], 0:prior_vectors.shape[1]] = prior_vectors
+    print(f"NEW {new_vectors[0:2]}")
```
piskvorky, line 1943:
```suggestion
```

piskvorky (changes requested):

gensim/models/keyedvectors.py
gojomo: Oops! Fixed via amend/force-push.

gojomo:
On further investigation, the flimsiness of the failing test is the real problem, and it was only passing in the 1st place via lucky seeding. (Over a range of seeds, either code alternative makes that test fail 20-30% of the time.) So I'll disable that test, & created #2977 to eventually – not at all urgently – clean up the related classes/docs.

piskvorky:
Thanks! Merge will have to wait for @mpenkov , but I already branched off here for my tickets, so no more force-pushing please (normal merges are OK).

gensim/models/keyedvectors.py
mpenkov: OK, but this stuff will show up in the docs, right?

I can think of several better ways forward:

- Make the docstring a code-comment so it doesn't show up in the docs
- Mark the function as internal

gensim/models/word2vec.py
```diff
@@ -964,8 +953,8 @@ def train(
         Parameters
         ----------
         corpus_iterable : iterable of list of str
-            The `sentences` iterable can be simply a list of lists of tokens, but for larger corpora,
-            consider an iterable that streams the sentences directly from disk/network.
+            The `corpus_iterable` can be simply a list of lists of tokens, but for larger corpora,
```
mpenkov, line 956: In general, Sphinx expects literals to be in double backticks.

```suggestion
            The ``corpus_iterable`` can be simply a list of lists of tokens, but for larger corpora,
```

There is no behavior for single backticks. We can customize that, but haven't, as far as I understand.

gensim/test/test_translation_matrix.py
```diff
@@ -102,17 +102,21 @@ def test_translation_matrix(self):
         transmat = model.train(self.train_docs[:5])
         self.assertEqual(transmat.shape, (8, 8))
 
-    def test_infer_vector(self):
+    def disabled_test_infer_vector(self):
```
mpenkov, line 105: I think it's better to use the built-in mechanism to skip tests.

```suggestion
    @unittest.skip("skip reason")
    def test_infer_vector(self):
```

mpenkov (commented):
Left some minor comments. Please have a look @gojomo .

gensim/models/doc2vec.py
```diff
@@ -82,6 +82,8 @@
 
 logger = logging.getLogger(__name__)
 
+SEED_DIVERSIFIER = 7919  # 1000th prime
```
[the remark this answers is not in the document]
gojomo: Historically, I didn't perceive a strong public/private distinction in Gensim code, though I've seen the leading-`_` convention growing recently. Yes, this is wholly internal to `Doc2Vec` at the moment, and only used in one place, so the literal number could be used that one place instead with a code comment to explain its role (instead of its variable name).

gensim/models/keyedvectors.py
```diff
@@ -204,6 +204,18 @@ def __init__(self, vector_size, count=0, dtype=REAL, mapfile_path=None):
...
+        vector_size : int
+            Intended number of dimensions for all contained vectors.
+        count : int, optional
+            If provided, vectors wil be pre-allocated for at least this many vectors. (Otherwise
+            they can be added later.)
+        dtype : type, optional
+            Vector dimensions will default to `np.float32` (AKA `REAL` in some Gensim code) unless
+            another type is provided here.
+        mapfile_path : string, optional
+            TODO: UNDER CONSTRUCTION / SUBJECT TO CHANGE - pending mmap work
```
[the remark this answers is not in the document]
gojomo: Per my comments 4 days ago, I'm not proposing any of these comments become part of the permanent codebase - they're all just temporary until #2955 is addressed, to prevent this PR from growing to become a diverse grab-bag of things unrelated to its title and initial focus. As @piskvorky has already started building on this in another PR, and I've got other functional things that depend on this, polishing comments here just risks making the tangle of PRs-on-PRs worse.

gensim/models/keyedvectors.py
```diff
@@ -337,34 +349,16 @@ def get_vecattr(self, key, attr):
...
-        return range(prev_count, target_count)
+    def resize_vectors(self, seed=0):
+        """Make underlying vectors match index_to_key size; random-initialize any new rows."""
 
-    def randomly_initialize_vectors(self, indexes=None, seed=0):
-        """Initialize vectors with low-magnitude random vectors, as is typical for pre-trained
-        Word2Vec and related models.
+        target_shape = (len(self.index_to_key), self.vector_size)
+        self.vectors = prep_vectors(target_shape, prior_vectors=self.vectors, seed=seed)
+        # TODO: support memmap & cleanup
```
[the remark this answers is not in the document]
gojomo: Intent is to fix before 4.0.0 via #2955.

gensim/models/word2vec.py
gojomo: OK, but this was just fixing the previous parameter name, leaving the prior-for-who-knows-how-long single-backtick formatting in place. Will commit your double-backticking.

gensim/test/test_translation_matrix.py
gojomo: Addressed in latest commit.

gensim/models/doc2vec.py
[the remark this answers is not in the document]
gojomo: I went ahead an inlined the seed-offset, as it's only done in one place.

gojomo:
For some reason, github isn't showing any option to reply to @mpenkov's comment about `prep_vectors`. To that:

* it's likely to change as part of #2955 / #2975 further initialization cleanup
* in its final form, it *might* still be a useful vector-resize-and-reinitialize routine for outside use, so visibility in API docs may not be any concern

So, I'd say: let's see how it settles in that other work.

gensim/models/keyedvectors.py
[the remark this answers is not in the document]
piskvorky: I personally use `FIXME` for things that *must* never see the light of day. (and mark even temporary debug constructs as `FIXME`, and explicitly grep for `FIXME` before a merge / release)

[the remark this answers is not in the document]
gojomo: I see that you added a note about `FIXME`/`TODO` to <https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#unresolved-problems> as a project guideline after this was last discussed – thanks! – so I've ensured all the loose ends are marked with `FIXME` (and "before 4.0.0") in the latest commit.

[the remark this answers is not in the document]
piskvorky: Thanks! Yes, I cleaned up & simplified that page further, in preparation for the 4.0 release.

piskvorky (approved):

gojomo:
Does the approval mean it's OK for me to push the Github 'rebase & merge' button?

piskvorky:
OK from me. Don't know about @mpenkov . Definitely no rebase though – just merge!

gojomo:
I'm just looking at the default button in the Github interface (which I suspect results in the cleanest trunk history), and might not foul any history elsewhere... nor foul other similar merges). But: if you or @mpenkov can just merge this as soon as it's OK, that'll be quickest - in my opinion, this PR has long been functionally done, and I'd prefer for simplicity to base the followup PR off `develop`.

piskvorky:
Okay. Merging here – @mpenkov please add any additional comments here / into @gojomo 's new PR.

PR #2953, opened 2020-09-20
Clarify random state in ldamulticore.py

sezanzeb:
The results of ldamulticore models were different with similar random_states, and given more passes the results became more similar, but never equal.

The only difference seems to be the order of how results are pushed to the result_queue.

LdaModel works as expected, the topics are exactly equal (a difference of 0) with the same random_state.

Since I found this rather irritating at first, I thought that maybe this should be reflected in the docstring.

I'd like to properly validate that first though, or maybe you know that this is the reason which ends up saving time.

### DIFF
--- a/gensim/models/ldamulticore.py
+++ b/gensim/models/ldamulticore.py
@@ -160,7 +160,8 @@ class LdaMulticore(LdaModel):
         minimum_probability : float, optional
             Topics with a probability lower than this threshold will be filtered out.
         random_state : {np.random.RandomState, int}, optional
-            Either a randomState object or a seed to generate one. Useful for reproducibility.
+            Either a randomState object or a seed to generate one. Useful for reproducibility. Note, that
+            results can vary slightly due to the nature of the multiprocessing queue.
         minimum_phi_value : float, optional
             if `per_word_topics` is True, this represents a lower bound on the term probabilities.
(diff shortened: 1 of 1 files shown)


piskvorky:
Yes, with multiprocessing (and multithreading), training results are no longer 100% reproducible. This is because of OS scheduling of processes (threads), which introduces non-determinism.

A docstring to say as much is welcome, thanks!

gensim/models/ldamulticore.py
```diff
@@ -160,7 +160,8 @@ def __init__(self, corpus=None, num_topics=100, id2word=None, workers=None,
         minimum_probability : float, optional
             Topics with a probability lower than this threshold will be filtered out.
         random_state : {np.random.RandomState, int}, optional
-            Either a randomState object or a seed to generate one. Useful for reproducibility.
+            Either a randomState object or a seed to generate one. Useful for reproducibility. Note, that results can
```
piskvorky, line 163:
```suggestion
            Either a randomState object or a seed to generate one. Useful for reproducibility.
            Note that results can still vary due to non-determinism in OS scheduling of the worker processes.
```

piskvorky (changes requested):

piskvorky (approved):

PR #2959, opened 2020-09-26
added test for sync_state

sezanzeb:


gensim/test/test_ldamodel.py
```diff
@@ -42,6 +43,25 @@ def setUp(self):
         self.class_ = ldamodel.LdaModel
         self.model = self.class_(corpus, id2word=dictionary, num_topics=2, passes=100)
 
+    def testSyncState(self):
```
gojomo, line 46: This is a question for @piskvorky / @mpenkov - our existing unit-test methods are a mix of `camelCase` and `snake_case` - somewhat owing to the way Python's `unittest` module names were largely imported from Java, and thus Python conventions around unit-testing classes/methods have been a mix. But, we should probably strongly prefer one or the other for new methods, and gradually bring old methods into conformance. I prefer `snake_case`, with the only exception being methods that specifically override a `unittest` `camelCase` precedent - so that all methods we ourselves write follow the single Python convention.

piskvorky: That's exactly right. I also prefer `snake_case`, code has `camelCase` due to historic reasons only.

sezanzeb: The documentation has it the way gojomo prefers: https://docs.python.org/3/library/unittest.html `setUp` and `test_*`

I'm going to rename the tests and push them to this PR then

mpenkov: Yeah, method names should be `snake_case` everywhere except where it is unavoidable, like the infamous `setUp` and friends.

sezanzeb: https://github.com/RaRe-Technologies/gensim/pull/3033

mpenkov:
Merged. Thank you @sezanzeb !

PR #2966, opened 2020-09-30
Build/test on Python 3.9

gojomo:
Python 3.9.0 is expected within a week: https://lwn.net/SubscriberLink/831783/44f4191759587fb7/

I'd tried a 3.9 build target a while back - but errors in other build-related packages made it fail. It might work now, and it'd be good to verify whether our build/unit-tests all work in Python 3.9. 

So, here the Travis CI & Azure pipeline builds get added Python "3.9-dev" targets, to see if all's well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,11 @@ matrix:
     - python: '3.6'
       env: TOXENV="flake8,flake8-docs"
 
+    - python: '3.9-dev'
+      env:
+        - TOXENV="py39-linux"
+      dist: focal
+      
     - python: '3.8'
       env:
         - TOXENV="py38-linux"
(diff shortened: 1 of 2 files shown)


gojomo:
Failing on Linux because there's no scypy wheel for py3.9, and a local build can't find necessary libraries. I expect some time soon official scypy wheels will appear.

Failing on Windows because the latest Python they have is 3.8.5.

FredHappyface:
is this something that I would be able to take a look at as it would be great to get the 3.9 Wheels on pypi.org?

piskvorky:
@gojomo are you able to merge `develop` (where we got rid of Travis) & trigger the CI build again?

Perhaps the dependencies have caught up with py3.9 in the meantime.

gojomo:
As adding py39 targets is just a few lines in the build spec files, I'd suggest trying a new PR off `develop` for a clean/simple history.

Issue #2960, opened 2020-09-28
Update documentation for 4.0.0


piskvorky:
Internal checklist so I don't forget something:

- [x] Rewrite the landing page copy for the new website.
- [x] Restructure top-level TOC tree for the new website.
- [x] Make all `rst` docs up-to-date for the new website, check hyperlinks, update stats & copy.
- [x] Go over rendered API docs and check everything's up to date and makes sense: formatting, descriptions, notes, hyperlinks.
- [x] Migrate & regenerate all Tutorials and How-tos.
- [x] Write the [Migration guide](https://github.com/RaRe-Technologies/gensim/wiki/Migrating-from-Gensim-3.x-to-4).
- [x] Write the 4.0.0 release notes.
- [x] Improve the [Developer page](https://github.com/RaRe-Technologies/gensim/wiki/Developer-page) with Gensim code-style: hanging indents, trailing commas, FIXME vs TODO, line length, etc.

piskvorky:
Fixes go into PR #2954.

piskvorky:
@mpenkov release notes below. Can you please add (auto-generate?) the `🔴 Bug fixes`  / `👍 Improvements`
/ `📚 Tutorial and doc improvements` / `⚠️ Deprecations` sections?

Next up, I'll fill in the performance numbers & finish the Migration guide, and we're ready to release 4.0.0beta. I don't think any of the other tickets are blocking (CC @gojomo).

## 1. Tweet

------

So, Gensim 4.0 is nearly here. Faster models, less RAM, prettier documentation.

Want to help? Install the pre-release and let us know!

pip install --pre --upgrade gensim

<link to the pre-release notes>

--------

## 2. Release notes (maybe also a blog post?)

--------

**⚠️ Gensim 4.0 contains breaking API changes! See the [Migration guide](FIXME) to migrate your existing Gensim 3.x code.**

Gensim 4.0 is a major release with lots of performance improvements, robustness and a [new website](FIXME new docs – kept side-by-side with 3.8.3 until the full release?) (finally!).

Main highlights (see also *👍 Improvements* below):

* Improved the most popular Gensim algorithms: [fastText](FIXME), [word2vec](FIXME), [doc2vec](FIXME), [phrases](FIXME). We did this:
  a) For efficiency. E.g. fastText & word2vec model training is now **10% faster** while using **FIXME % less RAM**. Phrases are more than 2x faster.
[...]

Issue #2883, opened 2020-07-14
Mmap model shared between process , memory use is not stable


raphaelauv:
#### Problem description

I'm exposing a word2vec model with Gunicorn sync workers.( inside flask )

_(and I run this gunicorn process inside one docker container )_

With 1 worker there is no problem. But with multiples workers ( forks from the main gunicorn process and with preload activated ) :

- mmap work well at statup , the memory use is almost the same than if I'm using only one process.

- But when I start to query the API ( each call execute a  `indexer.model.wv.most_similar` ) with a loadtest program, the memory usage grow until it stabilize.

This is with 8 sync workers (memory is growing):
![Screenshot from 2020-07-15 01-25-44](https://user-images.githubusercontent.com/10202690/87486375-7c5c1100-c63b-11ea-9f1f-ce835237dbc2.png)

 with only one worker the lines stay perfectly flat : 
![Screenshot from 2020-07-15 01-43-00](https://user-images.githubusercontent.com/10202690/87486905-a19d4f00-c63c-11ea-87f7-de820c8d49bd.png)

#### Steps/code/corpus to reproduce

Load a word2vec Gensim model with mmap activated and fork N times , then query the model in each subprocesses.

The problem is  already in stackoverflow , but no solution proposed : 

https://stackoverflow.com/questions/51616074/sharing-memory-for-gensims-keyedvectors-objects-between-docker-containers/51722607#51722607

What I tried :

This is the loading code of the model
```Python
from gensim.similarities.index import AnnoyIndexer
from gensim.models import KeyedVectors
from gensim.similarities import index as gensim_index
from my_project.patch_annoy import annoy_load

gensim_index.AnnoyIndexer.load = annoy_load # monkeypatch to put self.index.load(fname, prefault=True)

indexer = AnnoyIndexer()
indexer.load(LOCAL_MODEL_PATH)
wv = KeyedVectors.load(LOCAL_VECTORS_NORMED, mmap='r')
indexer.model = wv
indexer.model.wv.vectors_norm = indexer.model.wv.vectors
```

Also i get this message when I load the previously normalized ( `init_sims(replace=True)` ) model is  :

`setting ignored attribute vectors_norm to None`

#### Versions

```
Linux-5.0.0-38-generic-x86_64-with-Ubuntu-19.04-disco
Python 3.7.3 (default, Oct  7 2019, 12:56:13) 
[GCC 8.3.0]
NumPy 1.17.1
SciPy 1.4.1
[...]

PR #2968, opened 2020-09-30
[MRG] Migrate tutorials & how-tos to 4.0.0

piskvorky:
Part of #2960.


piskvorky:
@gojomo I went over our guides & tutorials, and all necessary migrations are in 502b654a. It went pretty smoothly, I only wasn't sure about doc2vec (see the one FIXME in diff). Overall migration shouldn't be a major issue, I'm happy about that.

I did see some uses of `build_vocab()` in the tutorials. These might need another touch after your #2944 is merged (maybe), I'll keep that in mind.

gensim/similarities/annoy.py
```diff
@@ -156,6 +156,7 @@ def build_from_word2vec(self):
 
     def build_from_doc2vec(self):
         """Build an Annoy index using document vectors from a Doc2Vec model."""
+        # FIXME docvecs now = ? What's the replacement for docvecs.count, index_to_doctag, etc?
```
gojomo, line 159: `.docvecs` = `.dv` (for compact consistency to its sibling `.wv`)
`.index_to_doctag` = `.index_to_key` (shared class means lookup-value is always 'key')
`dv.count` = `len(dv)` (which worked, & likely would have been clearer, in earlier code too)

piskvorky: I figured it out just now, thanks. Will push an update shortly = all migrations complete.

gensim/similarities/annoy.py
```diff
@@ -156,6 +156,7 @@ def build_from_word2vec(self):
...
         docvecs = self.model.docvecs
         labels = [docvecs.index_to_doctag(i) for i in range(docvecs.count)]
         return self._build_from_model(docvecs.vectors_docs_norm, labels, self.model.vector_size)
```
gojomo, line 162: There's no `doc2VecKeyedVectors.vectors_docs_norm` anymore, either as a doc-vectors-specific set of norms, nor as an usually-realized large array of all normed vectors. Instead, options include: (1) requesting normed vectors one-by-one with `dv.get_vector(key, norm=True)`; (2) `dv.get_normed_vectors()` which actually calcs the whole dense array on demand (without caching it for future requests) if truly needed.

I suspect `_build_from_model()`, below, may not be as important as `build_from_keyed_vectors()`. If it exists and needs the normed vectors (& I'm not sure it should force norming on all users, sometimes euclidean distances on raw vectors win on certain tasks!), it could unit-norm them itself, or let the user know they should pre-norm if that's what they know they want.

gojomo: Also re: `build_from_keyedvectors()` below - it may want to avoid instantiating the whole `get_normed_vectors()` - to use much less peak memory, but perhaps at cost of slower looping over each individually. (Whether that makes any different given the one-at-a-time `add_item()` further below, unsure.) I fear this class may in many cases be an 'attractive nuisance', which people assume is a good/necessary improvement since it's there, and we have a demo, but often isn't worth the overhead/complexity. See, for example, recent memory issue #2883 resolved satisfactorily just by dropping the complexifying use of AnnoyIndexer.

piskvorky: Yes, already fixed.

piskvorky:
Should be done now. I'll remove my internal test hacks & regenerate all docs next, so there will be lots of autogenerated `auto_examples` files soon.

My migration notes:

**word2vec**
- .vocab => .index_to_key
- size=100 => vector_size=100
- model.analogies() => model.wv.evaluate_word_analogies()
- index2word => .index_to_key

**doc2vec**
- .docvecs => .dv
- .index_to_doctag => .index_to_key
- dv.count = len(dv)

**fasttext**
- shenanigans around `word in model` => `word in model.wv.key_to_index`

piskvorky:
Actually, before I push the autogenerated `auto_examples` – do you want to review too @mpenkov ? After the push, the PR will be a mess of files.

piskvorky:
The regenerated docs are now available at https://radimrehurek.com/gensim_redesign/

piskvorky:
OK I'll merge, need to continue with other tickets. @mpenkov please let me know if you see anything wrong here whenever you get to it. I'll fix ex-post.

Issue #2956, opened 2020-09-24
Move BrownCorpus from word2vec to gensim.corpora


piskvorky:
Not a high-priority at all, but it'd be more sensible for such a tutorial/testing utility corpus to be implemented elsewhere - maybe under `/test/` or some other data- or doc- related module – rather than in `gensim.models.word2vec`.

_Originally posted by @gojomo in https://github.com/RaRe-Technologies/gensim/pull/2939#discussion_r493820305_

rishabhvarshney14:
Can I work on this?

piskvorky:
Yes :) 

We'll want to retain an alias in word2vec, so that people's code that relies on the current location continues to work. But the code itself should live under `gensim.corpora`.
Same with other such corpus classes that are currently under `gensim.models` – I believe there were several.

Thanks!

gojomo:
The same goes for the `Text8Corpus` alongside it - if that class survives at all. Probably, `LineSentence` should be able to handle not just "more than X tokens" lines but also "all tokens on one line" - in which case, it can subsume what `Text8Corpus` was doing. (And then, maybe it lives in a more sensible place, like utils/corpora, instead of as a tag-on under `word2vec`?

rishabhvarshney14:
@piskvorky I am thinking about moving BrownCorpus to another file named browncorpus.py in corpora also there is TaggedBrownCorpus in doc2vec which I will move to browncorpus.py along with BrownCorpus and will do it for other utility corpus like Text8Corpus. Is this what should be done? Please correct me if I am wrong.

piskvorky:
Yes – plus add aliases to the old locations, so code that relies on the original location continues to work. Make sure the files have proper headers and documentation.

Do we have unit tests for those classes? If not, could you add them too please? Thanks.

gojomo:
As these classes haven't had unit-tests, I wouldn't burden the simple housekeeping step (moving them to a more logical place) with the requirement to add unit tests for the 1st time.

[...]

Issue #2948, opened 2020-09-16
rename 8.6 tag to 0.8.6


pabs3:
#### Problem description

There is a tag "8.6" in the repository that is between 0.8.7 and 0.8.5 in terms of when it was created, but it is missing the 0 at the start.

#### Steps/code/corpus to reproduce

```
git clone https://github.com/RaRe-Technologies/gensim.git
cd gensim
git tag | grep -v '^[0-3]'
```

Or load the github tags page:

https://github.com/RaRe-Technologies/gensim/tags?after=0.8.9

#### Versions

0.8.6

#### Suggested fix

```
git tag 0.8.6 8.6
git tag -d 8.6
git push --delete origin 8.6
git push --tags
```

mpenkov:
Done, thank you for including a suggested fix!

pabs3:
@mpenkov looks like this weird 8.6 tag has re-appeared, although the 0.8.6 tag still exists. Could you delete the 8.6 tag again and try to figure out why it came back?

mpenkov:
@piskvorky @gojomo Could you guys please check that your local clones do not have an 8.6 tag? I suspect one of us has this tag lying around somewhere, and pushes it upstream occasionally.

Nobody else is able to push to upstream (right?) so I think it's one of us, anyway. I'll check my other machines for traces of this tag, too.

pabs3:
@mpenkov @piskvorky @gojomo I noticed that the bogus 8.6 tag has been reintroduced, please remove it again. The correct `0.8.6` tag is still present though. Please run these commands again on all of the systems where gensim developers push to the repo:

```
git tag -d 8.6
git push --delete origin 8.6
```

mpenkov:
@pabs3 Why does this matter to you?

pabs3:
The `8.6` tag is clearly incorrect and it is strange that it keeps showing up despite getting deleted.

[...]

PR #2970, opened 2020-10-01
Move copus to corpora

rishabhvarshney14:
Solves #2956

I have moved the following classes from word2vec and doc2vec
- BrownCorpus
- TaggedBrownCorpus
- LineSentence
- PathLineSentences
- TaggedDocument
- TaggedLineDocument
- Text8Corpus

I have also changed the old location of these files in the comments (but not everywhere like in tests which still import from the last location).

I hope this the correct method of this.


gensim/corpora/linesentence.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2013 Zygmunt Zając <zygmunt@fastml.com>
```
mpenkov, line 4: @piskvorky Are these copyright statements legit?

rishabhvarshney14: I have taken it from [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/corpora/csvcorpus.py) .

gensim/corpora/_mmreader.c
```diff
@@ -0,0 +1,10902 @@
+/* Generated by Cython 0.29.14 */
```
gojomo, line 1: This and other `.c` files created by Cython compilation should *not* be added to version control. (In fact, if we think no hand-authored `.c` files will be part of the project, a patern excluding them could be added to `.gitignore`.)

gensim/corpora/linesentence.py
gojomo: But, even if it was accurate there, this code is from many places & many authors, so it's not appropriate to attribute it all to another person in past time. 

In a few cases where many have contributed, I've changed the leading comment to read like (but it should probably say '2020' upon updates):

https://github.com/RaRe-Technologies/gensim/blob/683cebef27a0bf5e9af39d9b5b0a522e62f20862/gensim/models/doc2vec.py#L4-L5

Still, it'd be good to have a project standard for how authorship/copyright is declared per file. Saying less makes more sense than saying more that isn't true.

gensim/corpora/taggeddocument.py
```diff
@@ -0,0 +1,80 @@
...
+# Licensed under the GNU LGPL v2.1 - http://www.gnu.org/licenses/lgpl.html
+
+"""This module contains the classes of TaggedDocument and TaggedLineDocument."""
+
+from collections import namedtuple
+
+from gensim import utils
+
+
+class TaggedDocument(namedtuple('TaggedDocument', 'words tags')):
```
gojomo, line 14: `TaggedDocument` is closely-aligned with `Doc2Vec` as a description of the 'expected shape' of its training-items. It should stay in that class. 

The utility class `TaggedLineDocument` could move with other corpora-utility iterables.

gojomo:
I don't think all these small utility iterable classes each deserve their own files/modules. (I especially think this because I'm looking forward to unifying `LineSentence` & `Text8Corpus` into one more-generic class that does them both, per my comment on #2956.) 

It'd make more sense for them to share a `gensim.corpora.utils` home, or perhaps tag onto `gensim.corpora.textcorpus` (if that weren't already 700 lines & focused on a slightly-different corpus style). The cleanup/movement here may need some organization preference judgement from @piskvorky.

gensim/corpora/linesentence.py
piskvorky: Agreed with @gojomo . I plan to clarify the developer instructions as part of #2948.

TL;DR: authorship cannot be transferred in many legislations (including EU), it's for life. In order for us (maintainers) to be able to publish & modify the code, we need full assignment of all IP rights from the contributor. Which is what we've been [implicitly doing](https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#legal) but perhaps should make more explicit.

gojomo: While I've seen more complicated processes, and the State of the Art may have advanced since I last looked, my non-lawyerly hunch is that a low-effort system could be: (1) There's a very-simple rights-assignment statement in the repo; (2) To accept someone's contributions, they must 1st (either before or in their 1st PR) include the addition of their Github username to a list of developers granting such assent, at the bottom of that statement. Thus the same mechanism of contribution collects overt/authenticated expressions of assignment, while building a master-list of contributors as a side-effect.

rishabhvarshney14: Sorry I did not know about these things. 
I have changed it to 
``` 
 # Author: Gensim Contributors 
 # Copyright (C) 2020 RaRe Technologies s.r.o.
```
Sorry again.

gensim/corpora/_mmreader.c
rishabhvarshney14: Should I also add header files(.h) in gitignore?

mpenkov: No, some header files are real, like `wrap_stdint.h`.

rishabhvarshney14:
@gojomo I have moved all the classes in corpora.utils which also include TaggedDocument because TaggedBrownCorpus and TaggedLineDocument needed TaggedDocument. 

If I put TaggedDocument in doc2vec it causes Importing error since corpora.utils imported by doc2vec (for TaggedBrownCorpus and TaggedLineDocument) and doc2vec imported by corpora.utils (for TaggedDocument). 

This can be fixed by importing TaggedBrownCorpus and TaggedLineDocument after defining TaggedDocument like:
```
class TaggedDocument (...)
from corpora.utils import TaggedBrownCorpus, TaggedLineDocument
```
Is there anything that I missed?

gensim/corpora/utils.py
```diff
@@ -0,0 +1,284 @@
...
+
+import logging
+import os
+import itertools
+from collections import namedtuple
+
+from gensim import utils
+
+try:
+    from gensim.models.word2vec_inner import MAX_WORDS_IN_BATCH
```
gojomo, line 16: There isn't truly any dependency of these utility classes on some other-module Cython optimizations. sp this seems off in terms of direction-of-dependency, fragility, exception raised, etc.

Perhaps instead: a reference value is defined with a comment here, then used both below *and* imported from here into the other modules that want to be in-sync on this arbitrary cap? EG: 

```
#: Shared capped size, in count of words/tokens, for texts that should not be arbitrarily long
MAX_WORDS=10000
```

gojomo:
Good point about the order-of-imports, I suppose that means TaggedDocument & TaggedLineDocument should stay together, here. 

I prefer this current grouping of these utility-readers/format-classes in `gensim.corpora.utils` to their previous locations, or individual files for each, but it would be good to hear others' confirmation that this fits their preferences, too. 

I would however prefer to re-order them so that they're in roughly-related groupings, and roughly in order-of-prominence. In my mind, that'd be:

    LineSentence
    PathLineSentence
    Text8Corpus  # (but hopefully subsumed into LineSentence someday soon)
    TaggedDocument
    TaggedLineSentence
    BrownCorpus
    TaggedBrownCorpus

gojomo:
Regarding `LineSentence`, here's some code I was playing with that I think provides a superset of `LineSentence` & `Text8Corpus` functionality - in that it can iterate over an existing whitespace-tokenized file, and enforce a max-tokens per text, and will still use bounded memory (and emit max-tokens texts) from a single-giant-line file like `text8` or `text9`. It doesn't necessarily need to be part of this PR, but could be considered afterwards:

```python
class TokenLines(object):
    """
    Iterable over already-whitespace-tokenized texts in a line-oriented file.
    
    (Subsumes functionality of former LineSentence & Text8Corpus classes.)
    """
    def __init__(self, source, max_sentence_length=10000, limit=None, max_line_length=1024**2, encoding='utf-8'):
        self.source = source  # file or path
        self.max_sentence_length = max_sentence_length  # split any texts with more tokens
        self.limit = limit  # don't read more than this many lines
        self.max_line_length = max_line_length  # max readline; also once tokens collectively longer, stop extending
        
    def __iter__(self):
        try:
            self.source.seek(0)  # if it seems to be a file, assume directly readable
            reader = self.source
        except AttributeError:
            reader = utils.open(self.source, 'r', encoding=encoding)
            
        with reader:
            line_no = 0
            tokens = []
            slop= ''
            
            line = reader.readline(self.max_line_length)
            while line or tokens:
                tokens += line.split()
                
                # detect if line maybe not complete, mis-token or mid-line
                if len(line) >= self.max_line_length:
                    line_maybe_truncated = line[-1] != '\n'
                    token_maybe_truncated = tokens and line.endswith(tokens[-1])
                else:
                    line_maybe_truncated = token_maybe_truncated = False
                
                # while (1) tokens overlong; or (2) not at all truncated; or (3) chars overlong = emit text
                while (tokens 
                       and ((len(tokens) > self.max_sentence_length) 
                            or (not (line_maybe_truncated or token_maybe_truncated))
                            or (sum(len(t) for t in tokens) >= self.max_line_length))):
                    text, tokens = tokens[0:self.max_sentence_length], tokens[self.max_sentence_length:]
                    yield text
                    line_no += 1
                    
                if self.limit and line_no >= self.limit:
                    break

                # read next
                line = reader.readline(self.max_line_length)

                # sew-up potential token fragment
                if token_maybe_truncated:
                    tokens, slop = tokens[:-1], tokens[-1]
                    line = slop + line
```

gensim/corpora/utils.py
```diff
@@ -0,0 +1,282 @@
...
+import logging
+import os
+import itertools
+from collections import namedtuple
+
+from gensim import utils
+
+logger = logging.getLogger(__name__)
+
+# Shared capped size, in count of words/tokens, for texts that should not be arbitrarily long
```
gojomo, line 17: Note that you may specifically need to use a "#: ` to start this comment for it to be picked up by Sphinx/autodoc tools as a bit of documentation for the variable assignment. (See: https://stackoverflow.com/a/20227174/130288)

rishabhvarshney14: I have fixed that thank you.

PR #2972, opened 2020-10-03
Remove wrappers and other cruft

mpenkov:
Removed `gensim/models/wrappers`, associated tests and documentation entirely.

Removed `gensim/sklearn_api` entirely.

Removed simserver.rst - seems like something really ancient (e.g. https://radimrehurek.com/gensim/simserver.html).

@piskvorky Some questions:

- Is there anything else worth removing?
- What should we do about the coherence model tests? They are useless without the wrappers.


piskvorky:
Thanks! Can you add a summary into the Migration wiki too please? So I don't forget to describe it.

piskvorky:
Aaaah, here it is. This PR wasn't labeled as "Milestone 4.0" nor linked to its issue #2852, so I couldn't find it. But I knew there was more :)

> Is there anything else worth removing?

Yes, for sure. My dilemma is that I'm not sure whether I dislike some modules simply because I never used them (~a bit of NIH), or because they're objectively useless.

Although my intuition leans heavily toward the latter.

Going by mailing list to group the problematic modules:
1. Top candidates that people report issues with (=try to use and fail) are HDP, summarization, dtm and the mallet wrapper. **So either fix (unrealistic for us unfortunately) or remove those**.
2. More obscure wrappers with occasional reports are sklearn, pattern. **Just remove.**
3. Never-heard-user-reports-from: pivoted normalization, poincare, ?more? (not top of my mind because not discussed). **Just remove.**

> What should we do about the coherence model tests? They are useless without the wrappers.

I'm not sure, would have to look closer what that does.

mpenkov:
From my point of view, the problem is that I don't have a full understanding of what needs to go and what needs to stay. Once that gets decided, I can execute relatively quickly.

Of course, making the big decision is the hardest part of this task.

piskvorky:
OK. Not blocking for 4.0.0beta.

Issue #2973, opened 2020-10-06
phrases.export_phrases() doesn't yield all bigrams


o-nc:
Hallo and thank you for this tool,

phrases.export_phrases() doesn't yield all bigrams when some are part of a bigger n-gram.

If I create a Phrases object with 

`phrases = gensim.models.phrases.Phrases(sentences, min_count=1, threshold=10, delimiter=b' ', scoring='default')`

on the following two sentences

New York City has the largest population of all the cities in the United States .
Every year, many travelers come to the United States to visit New York City .

`print(dict(phrases.export_phrases(sentences)))` only returns {b'New York': 11.5, b'United States': 11.5}. It should also return {b'York City': 11.5} however.

line 187 of phrases.py should probably be changed to `last_uncommon = word` . It fixes the problem on my side and seems to be what the code was intended to be.

Thank you,
Olivier NC

macOS-10.14.5-x86_64-i386-64bit
Python 3.8.5 (v3.8.5:580fbb018f, Jul 20 2020, 12:11:27) 
[Clang 6.0 (clang-600.0.57)]
Bits 64
NumPy 1.19.2
SciPy 1.5.2
gensim 3.8.3
FAST_VERSION 0

piskvorky:
That module has weird docs written in broken English. E.g. `common_terms : list of object – List of common terms, they receive special treatment.`… WTF?

@mpenkov I'll try to clean up, otherwise another candidate for axing IMO.

@o-nc hopefully I'll get to your question too during the cleanup.

piskvorky:
@o-nc if `New York` phrase is detected in the text `New York City`, then `York City` cannot be detected (the `York` token is already "used").

If you wish to detect multi-word phrases (3grams, 4grams, etc), run Phrases multiple times over itself.

Check out #2976 for a faster & cleaner re-implementation of phrase detection. Feel free to use that, feedback welcome. That PR will be a part of the 4.0.0 Gensim release, planned for this month.

gojomo:
See also #1719 for discussion of the same overlapping-bigram issue, how the "greedy" left-to-right nature of the current code causes it, & some partial ideas for how it might be changed to report the 'better' pairings. (Though, in your test case, as they're perfectly tied, you'd still not get both without yet another choice to return some constituent unigrams more than once, which might surprise more users.)

PR #2976, opened 2020-10-08
[MRG] Refactor phrases

piskvorky:
Following #2973, I set out to improve the documentation for our phrases module. While at it, I saw many inefficiencies and cumbersome constructs there, so ended up with a larger refactor.

Benefits of this PR:
- significantly less code (790 vs 941 lines – despite adding many comments)
- significantly cleaner code (got rid of the maze of call indirections, weird OOP inheritance)
- significantly faster (42s vs 81s on text8 = almost double!)

Not in this PR:
- Compiled code for performance. Although I did locate all hot-spot code into one function, so cythonizing should be both straightforward and highly effective (a single tight loop).

I also added migration code for `load()`, so old models continue to work.


piskvorky:
More detailed timing info, all measured on `text8`:

| action         | develop                                    | this PR               |
|----------------|--------------------------------------------|-----------------------|
| create Phrases | 36.6s                   | 27.3s                    |
| apply Phrases  | 81s                   | 42.8s |
| export Phraser | 53.4s | 10.9s            |
| apply Phraser  | 57.1s   | 22s            |

@gojomo @mpenkov seems better across the board; please review and let's include this in 4.0.0.

gensim/models/phrases.py
```diff
@@ -554,237 +610,73 @@ def add_vocab(self, sentences):
...
+        word_a_cnt = self.vocab[word_a]
+        if word_a_cnt <= 0:
+            return None, None
+
+        word_b_cnt = self.vocab[word_b]
+        if word_b_cnt <= 0:
+            return None, None
+
+        phrase = self.delimiter.join([word_a] + in_between + [word_b])
+        # XXX: Why do we care about *all* phrase tokens? Why not just score the start+end bigram?
```
piskvorky, line 644: @alexgarel @gojomo do you remember why we score on `word_a + in_between + word_b`, instead of just `word_a word_b`? What's the point of including the common words in the calculation?

gojomo: I'd guess that tracks back to feature #1258, with an abortive implementation in #1567, arriving in #1568. It sounded useful to me in some situations, but hard enough to understand, and twisty enough in its effect on existing code, that I suggested a couple times it be a separate class, rather than options on `Phrases`.

gojomo: (And relatedly: my hunch is this has no effect unless that non-default option is activated, because `in_between` is then always empty.)

piskvorky: Your hunch is correct – `common_words` is empty by default.

I'm leaving the logic of `#['eye', 'of', 'the', 'beholder']` vs `#['eye', 'beholder']` unchanged. I don't know if anyone's using `common_words` at all, but if they are, let's avoid surprises.

gojomo: Re: defaults

Despite my own perspective as an English monglot, I'd hate to change a language-agnostic algorithm, that previously had no English-specific initialization, to become English-specific – even in a major version bump, with lots of warning. (Providing a handy switch/presets that makes enabling an English-specific mode, and/or other languages, sounds useful though.)

piskvorky:
@mpenkov @gojomo Gentlemen, I'll merge this now for the sake of simplicity. Your review comments are still welcome; I'll address them in one of the upcoming PRs.

gensim/models/phrases.py
piskvorky: Addressed in #2979 .

PR #2979, opened 2020-10-10
[MRG] Skip common English words in phrases

### LINKED ISSUE #1465: Better support for evaluating threshold settings in models.phrases.Phrases
One of the challenges when learning bigrams from a new corpus is determining the right threshold for your scoring to accept or reject a potential bigram. Standard approaches involve taking a list of gold-labeled bigrams created by humans, ranking all bigrams in your corpus by their score, and determining a score threshold based on a comparison to the gold labels. (for one example, see [this paper](https://svn.spraakdata.gu.se/repos/gerlof/pub/www/Docs/npmi-pfd.pdf)

Right now, doing this with built in functionality in models.Phrases requires running export_phrases on your whole corpus. For an especially large corpus, this could mean a lot of wasted time waiting for export_phrases to run. It also means getting some strange results--if you set your threshold very low, a strong bigram in your corpus may not be output by export_phrases if the first word in the strong bigram is especially rare and the strong bigram is proceeded by a limited set of words (you'd get a bunch of lower scoring bigrams with the first word of the strong bigram as the second word of some weaker bigrams).

There should be a method that only traverses the vocab dictionary and returns something that shows the scores for the bigrams in the corpus. This would be both faster than export_phrases and would ensure that all bigrams (that exceed some threshold) have their score output. I have code to do something like this, and I'm happy to contribute it. (Although it might make sense to wait until after #1464 and maybe #1446 are finalized.)

### LINKED ISSUE #2520: Bug in models.phraser.original_scorer
#### Problem description

There seems to be a bug in models.phraser.original_scorer. `min_count` is defined as "Minimum collocation count threshold", which is a bit ambiguous, but this argument actually corresponds to `Phrases.min_count`, which is defined as:

```
min_count : float, optional
            Ignore all words and bigrams with total collected count lower than this value.
```
Given this definition, line 692 is seemingly incorrect:

```
return (bigram_count - min_count) / worda_count / wordb_count * len_vocab
```

When bigram_count == min_count, this will return zero, meaning that either the definition of min_count is wrong, or this formula is incorrect.

#### Versions

I'm using gensim 3.7.3, but this issue seems to go back all the way to the original commit of the phrases module back in 0.10.3.

piskvorky:
Follow up from https://github.com/RaRe-Technologies/gensim/pull/2976#discussion_r502769899 : add a list of English "common words" to skip during phrase construction.

These English words are an optional switch; the default doesn't change.

**Backward incompatible change**: this PR renamed the `common_terms` parameter of `gensim.models.phrases.Phrases()` to `connector_words`, with the same meaning.

Fixes #2520. Fixes #1465.

### DIFF
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -463,6 +463,7 @@ class Phrases(_PhrasesTransformation):
             connector words in the middle.
 
             **If your texts are in English, set** ``connector_words=phrases.ENGLISH_CONNECTOR_WORDS``.
+
             This will cause phrases to include common English articles, prepositions and
             conjuctions, such as `bank_of_america` or `eye_of_the_beholder`.
 
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -21,7 +21,7 @@ Examples
 
     >>> from gensim.test.utils import datapath
     >>> from gensim.models.word2vec import Text8Corpus
-    >>> from gensim.models.phrases import Phrases, ENGLISH_COMMON_TERMS
+    >>> from gensim.models.phrases import Phrases, ENGLISH_CONNECTOR_WORDS
     >>>
     >>> # Create training corpus. Must be a sequence of sentences (e.g. an iterable or a generator).
     >>> sentences = Text8Corpus(datapath('testcorpus.txt'))
@@ -31,7 +31,7 @@ Examples
     ['computer', 'human', 'interface', 'computer', 'response', 'survey', 'system', 'time', 'user', 'interface']
     >>>
     >>> # Train a toy phrase model on our training corpus.
-    >>> phrase_model = Phrases(sentences, min_count=1, threshold=1, common_terms=ENGLISH_COMMON_TERMS)
+    >>> phrase_model = Phrases(sentences, min_count=1, threshold=1, connector_words=ENGLISH_CONNECTOR_WORDS)
     >>>
     >>> # Apply the trained phrases model to a new, unseen sentence.
     >>> new_sentence = ['trees', 'graph', 'minors']
@@ -75,10 +75,10 @@ logger = logging.getLogger(__name__)
 
 NEGATIVE_INFINITY = float('-inf')
 
-# Set of common English words. Tokens from this set are "ignored" during phrase detection:
-# 1) Phrases may not start with these words.
+# Words from this set are "ignored" during phrase detection:
+# 1) Phrases may not start nor end with these words.
 # 2) Phrases may include any number of these words inside.
-ENGLISH_COMMON_TERMS = frozenset(
+ENGLISH_CONNECTOR_WORDS = frozenset(
     " a an the "  # articles; we never care about these in MWEs
     " for of with without at from to in on by "  # prepositions; incomplete on purpose, to minimize FNs
     " and or "  # conjunctions; incomplete on purpose, to minimize FNs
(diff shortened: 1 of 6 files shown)


piskvorky:
I still cannot access CircleCI – @mpenkov can you check what it's complaining about? Thanks.

piskvorky:
10 random phrases that were discarded (left column) vs newly introduced (right column) by this PR. Trained on a toy corpus 
of `text8`, default settings:

| set(develop) - set(this PR) | set(this PR) - set(develop) |
|-----------------------|--------------------------------------------|
| crash_in | died_of_a_stroke |
| the_remainder | from_abram |
| prefixed_by | prefectures_of_greece |
| the_serfs | new_creative |
| of_rakis | conquered_by_the_assyrian |
| releasing_a | stream_of_knowledge |
| an_inventive | tend_to_consider |
| overheard_by | war_in_iraq |
| smoothly_at | destruction_of_troy |
| an_offshore | four_preludes |

gojomo:
I'm not sure `text8` (small, weird) is a great reference dataset for evaluating this feature. But I can believe that most users of `Phrases` on English-text will benefit from enabling this `common_words` auto-expansion of the original `Phrases` algorithm beyond just bigrams. 

Still, the case against making it the silent default is:

* users following the exact same high-level interface with the same training text as in Gensim 0.x through 3.8.3 shouldn't be surprised by a different set of phrases being created, especially phrases of more-than-2-unigrams
* explicit is better than implicit, and algorithms that aren't inherently linked to a specific language shouldn't be language-specialized without clear user opt-in

If this previously-underappreciated feature is now liked enough to recommend more widely, I believe a better way to do that would be to update docs/tutorials/change-notes to highlight the availability of the `ENGLISH_COMMON_TERMS` preset, & use it where appropriate, with discussion of when it is or isn't appropriate.

piskvorky:
I agree. I'll make this optional.

piskvorky:
@mpenkov please review & merge (after checking that circleci fail).

gensim/models/phrases.py
```diff
@@ -455,24 +455,26 @@ def __init__(
             #. "default" - :func:`~gensim.models.phrases.original_scorer`.
             #. "npmi" - :func:`~gensim.models.phrases.npmi_scorer`.
         common_terms : set of str, optional
-            Set of "stop words" to include in bigram phrases, without affecting their scoring.
-            Allows detection of expressions like `bank_of_america` or `eye_of_the_beholder`.
+            Set of "stop words" that may be included within a phrase, without affecting its scoring.
```
gojomo, line 457: While analogous & overlapping in some ways with traditional 'stop words', these aren't exactly the same thing. (Among other things, verbs and pronouns don't seem wise to include in `ENGLISH_COMMON_TERMS`.) So, to actually call them 'stop words' seems misleading (& perhaps prone to user errors, such as supplying some larger set instead, like say NLTK's 100-plus word set). 

Similarly, it occurs to me that this feature/role is inconsistently referred-to as 'common terms' here but 'common words' elsewhere. Perhaps the 'word'/'term' variance should be settled one way or the other, and maybe even 'common' is the wrong word for this particular function. 

If this were still identical to the behavior of the Elasticsearch feature that once inspired it, or if the optimal values to provide were exactly the most 'common' words, perhaps the current name makes sense. But to the extent the algorithm/use has changed, and the most actually 'common' terms aren't necessarily the tight values to provide, making a distinct name could make sense. 

Perhaps, 'connector' words? Or 'filler' words? Or 'subsidiary' words? (They're only considered for phrases via attachment to non-subsidiary words.)

piskvorky: True. I'll rename to `filler_words`, and remove any mentions of "stop words".

piskvorky: @gojomo done in 75caa93, I went with `connector_words`.

It's a mouthful, but captures the purpose better than `filler_words` (which to me implies these words carry no meaning, which is not true).

piskvorky:
Merging to keep branches clean. @mpenkov review welcome as always.

Issue #2942, opened 2020-09-10
Segfault when training doc2vec


Paul-E:
#### Problem description

When attempting to train doc2vec, gensim segfaults.

#### Steps/code/corpus to reproduce

I run the code:
```
import faulthandler
import gensim
faulthandler.enable()
model = gensim.models.doc2vec.Doc2Vec(corpus_file = "yelp_tripadvisor_linesentence.txt", vector_size=250, min_count=10, epochs=40, workers = 5)
```
I get the output:
```
Fatal Python error: Segmentation fault

Current thread 0x00007f2d9effd700 (most recent call first):
  File "/home/paul/.local/lib/python3.8/site-packages/gensim/models/doc2vec.py", line 431 in _do_train_epoch
  File "/home/paul/.local/lib/python3.8/site-packages/gensim/models/base_any2vec.py", line 172 in _worker_loop_corpusfile
  File "/usr/lib/python3.8/threading.py", line 870 in run
  File "/usr/lib/python3.8/threading.py", line 932 in _bootstrap_inner
  File "/usr/lib/python3.8/threading.py", line 890 in _bootstrap

Thread 0x00007f2d9f7fe700 (most recent call first):
  File "/home/paul/.local/lib/python3.8/site-packages/gensim/models/doc2vec.py", line 431 in _do_train_epoch
  File "/home/paul/.local/lib/python3.8/site-packages/gensim/models/base_any2vec.py", line 172 in _worker_loop_corpusfile
  File "/usr/lib/python3.8/threading.py", line 870 in run
  File "/usr/lib/python3.8/threading.py", line 932 in _bootstrap_inner
  File "/usr/lib/python3.8/threading.py", line 890 in _bootstrap

Thread 0x00007f2d9ffff700 (most recent call first):
  File "/home/paul/.local/lib/python3.8/site-packages/gensim/models/doc2vec.py", line 431 in _do_train_epoch
  File "/home/paul/.local/lib/python3.8/site-packages/gensim/models/base_any2vec.py", line 172 in _worker_loop_corpusfile
  File "/usr/lib/python3.8/threading.py", line 870 in run
  File "/usr/lib/python3.8/threading.py", line 932 in _bootstrap_inner
  File "/usr/lib/python3.8/threading.py", line 890 in _bootstrap

Thread 0x00007f2da48df700 (most recent call first):
[...]

Issue #2938, opened 2020-09-09
Fix deprecations in SoftCosineSimilarity


piskvorky:
When running our CI test suite, I see an array of deprecation warnings:

https://dev.azure.com/rare-technologies/gensim-ci/_build/results?buildId=287&view=logs&jobId=f9575ddc-dec8-54e6-9d26-abb8bdd9bed7&j=f9575ddc-dec8-54e6-9d26-abb8bdd9bed7&t=180156a9-2bf9-537d-c84a-ef9e808c0367

Some are from gensim, some from scipy:
<img width="1440" alt="Screen Shot 2020-09-09 at 09 16 00" src="https://user-images.githubusercontent.com/610412/92567026-635b8f00-f27d-11ea-9d5f-c56f3ba7c08b.png">

@Witiko could you please have a look? Is it something your existing PR already addresses?

If not, can you please fix those? Thanks.

piskvorky:
…and we're also still using the deprecated `smart_open.smart_open` apparently:

```
gensim/test/test_api.py::TestApi::test_load_dataset
gensim/test/test_api.py::TestApi::test_multipart_load
  d:\a\1\s\.tox\py37-win\lib\site-packages\smart_open\smart_open_lib.py:254: UserWarning: This function is deprecated, use smart_open.open instead. See the migration notes for details: https://github.com/RaRe-Technologies/smart_open/blob/master/README.rst#migrating-to-the-new-open-function
    'See the migration notes for details: %s' % _MIGRATION_NOTES_URL
```

piskvorky:
And more:

```
gensim/test/test_datatype.py: 2 warnings
gensim/test/test_doc2vec.py: 14862 warnings
gensim/test/test_fasttext.py: 96 warnings
gensim/test/test_keyedvectors.py: 20 warnings
gensim/test/test_word2vec.py: 84 warnings
  D:\a\1\s\gensim\models\keyedvectors.py:1478: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
    fout.write(utils.to_utf8(prefix + str(key)) + b" " + row.tostring())
```

Seems to be coming from https://github.com/python-pillow/Pillow/commit/71c95c8e5f3bba1845444a246d04646825e6bab3 , which I don't even know how/where we're using.

The whole CI log could do with a careful inspection.

Witiko:
[...]

Issue #2929, opened 2020-09-01
ImportError: cannot import name 'logsumexp' on KeyedVectors


luisaheise:
####Problem description

I'm unable to import KeyedVectors.

I guess replacing 
`from scipy.misc import logsumexp`
 to
`from scipy.special import logsumexp`
on the source code should fix it, but I'm not quite sure.

#### Steps/code/corpus to reproduce

from gensim.models import KeyedVectors

#### Versions

Linux-5.3.0-61-generic-x86_64-with-Ubuntu-18.04-bionic
Python 3.6.5 (default, Apr  1 2018, 05:46:30) 
[GCC 7.3.0]
Bits 64
NumPy 1.19.1
SciPy 1.5.2

piskvorky:
Can you add the full stacktrace + gensim version?

luisaheise:
gensim==2.0.0
![image](https://user-images.githubusercontent.com/43735789/91854426-1b63b900-ec3a-11ea-9b96-e242ce4c9e84.png)

luisaheise:
I just noticed that this error only happens in a jupyter notebook

piskvorky:
Gensim 2.0.0 is more than 3 years old. I recommend upgrading.

PR #2981, opened 2020-10-17
Update changelog for 4.0.0 release

### LINKED ISSUE #2960: Update documentation for 4.0.0

mpenkov:
Ideally, we update the changelog after each merged PR, so that it reflects the most recent changes.

In real life, I found that many PRs weren't mentioned in the log. I wrote a quick script to go through all merged PRs since the last release, and then manually added them to each relevant changelog section. They're currently ordered by their PR number, to make it easier to spot duplicates, but I imagine ordering them in order of decreasing significance may be more helpful to the users. @piskvorky I'll leave that part to you, if you don't mind.

Closes #2960.


piskvorky:
Nice! How about closed tickets?

release/generate_changelog.py
```diff
@@ -0,0 +1,38 @@
+"""Generate changelog entries for all PRs merged since the last release."""
+import requests
+
+get = requests.get('https://api.github.com/repos/RaRe-Technologies/gensim/releases')
+get.raise_for_status()
+
+last_release_date = get.json()[0]['published_at'][:10]
+
+
+def g():
```
piskvorky, line 10: Better name. How about:

```suggestion
def iter_prs(newer_than):
```

release/generate_changelog.py
```diff
@@ -0,0 +1,38 @@
+"""Generate changelog entries for all PRs merged since the last release."""
```
piskvorky, line 1: Needs the standard Gensim file header.

release/generate_changelog.py
```diff
@@ -0,0 +1,38 @@
...
+import requests
+
+get = requests.get('https://api.github.com/repos/RaRe-Technologies/gensim/releases')
+get.raise_for_status()
+
+last_release_date = get.json()[0]['published_at'][:10]
```
piskvorky, line 7: Is this guaranteed to be sorted? What's `[0]` and `[:10]`? Some code comments please.

release/generate_changelog.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def g():
+    page = 1
+    done = False
+
+    while not done:
+        get = requests.get(
+            'https://api.github.com/repos/RaRe-Technologies/gensim/pulls',
+            params={'state': 'closed', 'page': page},
```
piskvorky, line 17: Is this guaranteed to be sorted by `created_at`?

release/generate_changelog.py
```diff
@@ -0,0 +1,38 @@
...
+        pulls = get.json()
+        if not pulls:
+            break
+
+        for i, pr in enumerate(pulls):
+            if pr['created_at'] < last_release_date:
+                done = True
+                break
+
+            if pr['merged_at'] and pr['merged_at'] > last_release_date:
```
piskvorky, line 29: How does this work when a PR is merged on the same date as `last_release_date`? What's the resolution of `last_release_date`?

CHANGELOG.md
```diff
@@ -7,23 +7,42 @@ This release contains a major refactoring.
...
 * No more wheels for x32 platforms (if you need x32 binaries, please build them yourself).
-  (__[menshikh-iv](https://github.com/menshikh-iv)__, [#6](https://github.com/RaRe-Technologies/gensim-wheels/pull/6))
-* Speed up random number generation in word2vec model (PR [#2864](https://github.com/RaRe-Technologies/gensim/pull/2864), __[@zygm0nt](https://github.com/zygm0nt)__)
-* Fix deprecations in SoftCosineSimilarity (PR [#2940](https://github.com/RaRe-Technologies/gensim/pull/2940), __[@Witiko](https://github.com/Witiko)__)
-* Remove Keras dependency (PR [#2937](https://github.com/RaRe-Technologies/gensim/pull/2937), __[@piskvorky](https://github.com/piskvorky)__)
-* Bump minimum Python version to 3.6 (PR [#2947](https://github.com/RaRe-Technologies/gensim/pull/2947), __[@gojomo](https://github.com/gojomo)__)
+  __[menshikh-iv](https://github.com/menshikh-iv)__, [#6](https://github.com/RaRe-Technologies/gensim-wheels/pull/6)
+* Refactor ldamulticore to serialize less data [#2300](https://github.com/RaRe-Technologies/gensim/pull/2300), __[@horpto](https://github.com/horpto)__
```
piskvorky, line 12: Hm, I remember already fixing inarticulate PR titles like this one. How did the script come up with this again?

piskvorky:
OK, I updated & cleaned up the CHANGELOG. I didn't check whether the PR list was complete though.

@mpenkov please check my review above when you have a moment, then we can merge. Thanks.

piskvorky:
@mpenkov Where is the `🔴 Bug fixes` section for 4.0.0? Did you remove it on purpose?

piskvorky:
I added back the 🔴 Bug fixes section, but need the above info on what modules / functionality was removed, to complete the CHANGELOG.

gensim/models/keyedvectors.py
```diff
@@ -1520,45 +1520,52 @@ def save_word2vec_format(
...
+                    vout.write(f"{prefix}{word} {self.get_vecattr(word, sort_attr)}\n")
 
         logger.info("storing %sx%s projection weights into %s", total_vec, self.vector_size, fname)
         assert (len(self.index_to_key), self.vector_size) == self.vectors.shape
 
-        # after (possibly-empty) initial range of int-only keys,
-        # store in sorted order: most frequent keys at the top
+        # After (possibly-empty) initial range of int-only keys,
+        # store in sorted order: most frequent keys at the top.
+        # TODO: is this still relevant in 4.0? What is this about?
```
piskvorky, line 1549: @gojomo is this still relevant? What comment goes here?

piskvorky:
@gojomo I fixed a bunch of FIXMEs in 52acca5f85a5779ada3bb641fb67eb0e5fbf0f69, and demoted others to TODO because not really blocking. Can you have a look – anything already resolved / outdated?

gensim/models/keyedvectors.py
gojomo: This is set up so that *if* this instance has taken advantage of (any number of initial) literal int index keys to avoid having a larger `key_to_index` dict, those indexes are preserved (because changing them would be disruptive), before the rest of the key-looks-up-index indirectly keys (which don't need to preserve exact indexes across saves). 

(This optimized alternative was added for `Doc2Vec`, where doc-tags generally *don't* have a frequency order, and often can be just natural-ordinal numbers. In some plausible-but-extreme scenarios, like tens/hundreds-of-millions of doctags, this could save a notable amount of memory.)

piskvorky: So it's a just backward compatibility thing? No possibility of new models (created in 4.0) being affected?

If so, I'll move it to some migration / load code path. That "optimized Doc2Vec code path with ints" I thought (hoped) we got rid of during the refactor. Way more trouble than worth.

tox.ini
```diff
@@ -13,7 +13,7 @@ show-source = True
 [flake8-rst]
 filename = *.rst *.py
 max-line-length = 120
-ignore = F821  ; TODO remove me when all examples in docstrings will be executable
+ignore = F821, F401  ; TODO remove me when all examples in docstrings will be executable
```
gojomo, line 16: Is the intent of adding `F401` here to ignore *all* unused imports *everywhere*? (That seems a big change from the past.)

piskvorky: This should apply only to docstrings (flake8-rst), not "actual" code.

It's actually working around a bug where one import in fasttext.py was actually used but still triggered this error, mysteriously.

Let me add a better comment here (and maybe report that bug upstream).

piskvorky: I went down the rabbit hole of debugging this. flake8-rst (added in PR #2192) is quite flaky, doesn't even work with up-to-date versions of flake8: https://gitlab.com/pycqa/flake8/-/issues/641

I'm inclined to remove flake8-rst altogether, rather than waste more time on this.

release/generate_changelog.py
mpenkov: I can't find any guarantees anywhere, but in practice, yes, it seems to be.

release/generate_changelog.py
mpenkov: I increased the resolution to use the full timestamp instead of just the date.

release/generate_changelog.py
```diff
@@ -0,0 +1,38 @@
...
+            'https://api.github.com/repos/RaRe-Technologies/gensim/pulls',
+            params={'state': 'closed', 'page': page},
+        )
+        get.raise_for_status()
+        pulls = get.json()
+        if not pulls:
+            break
+
+        for i, pr in enumerate(pulls):
+            if pr['created_at'] < last_release_date:
```
piskvorky, line 25: How does this work? PRs created before the last release do not appear in the changelog?

piskvorky (changes requested):

mpenkov:
> @mpenkov Where is the `🔴 Bug fixes` section for 4.0.0? Did you remove it on purpose?
> 
> Checking that it wasn't deleted by accident (with PRs already in it?).

I didn't remove anything - I only added to the changelog in [my commit](https://github.com/RaRe-Technologies/gensim/pull/2981/commits/1b1aec4e2359ddba9ca42fea031ebb78635464af).

piskvorky:
I hijacked this PR to add some house-keeping fixes again: code style, docs, formatting, etc… things I saw while scanning the Gensim code base before this merge = beta release.

gensim/models/keyedvectors.py
gojomo: It's not just backward-compatibility - this optimization is still active & available to anyone who uses only plain ascending int doc tags in `Doc2Vec`. 

The original motivation was I had a target use where there were tens-of-millions of docs, and the RAM savings from avoiding the strings and the full dict was noticeable given the 16GB or 32GB machine I had to work with. (It's possible that the same sort of trick might be helpful for some outlier Word2Vec uses, too – perhaps in bioinformatics or log/network-analysis – but I haven't actually tried it.)

The code's been simplified somewhat, in that `.index_to_key` will always have entries for these direct-int slots. (There had been a further complication in earlier code trying to minimize or eliminate the list-of-keys - but the list overhead for ints seems much smaller than the dict overhead for strings.)

mpenkov:
@piskvorky I updated the changelog script. Please have a look.

I also skimmed through issues that aren't linked to a PR, but didn't find much that belongs in the change log (aside from a pair of issues). The candidates are below, please have a look.

---

### :question: Closed issues

TODO: move each issue to its appropriate section or delete if irrelevant

* [#2990](https://github.com/RaRe-Technologies/gensim/issues/2990): No explanation of the languages supported by Gensim various modules
* [#2988](https://github.com/RaRe-Technologies/gensim/issues/2988): `word2vec.doesnt_match` numpy vstack deprecation warning 
* [#2985](https://github.com/RaRe-Technologies/gensim/issues/2985): Bug report of gensim official webpage
* [#2973](https://github.com/RaRe-Technologies/gensim/issues/2973): phrases.export_phrases() doesn't yield all bigrams
* [#2948](https://github.com/RaRe-Technologies/gensim/issues/2948): rename 8.6 tag to 0.8.6
* [#2942](https://github.com/RaRe-Technologies/gensim/issues/2942): Segfault when training doc2vec
* [#2938](https://github.com/RaRe-Technologies/gensim/issues/2938): Fix deprecations in SoftCosineSimilarity
* [#2929](https://github.com/RaRe-Technologies/gensim/issues/2929): ImportError: cannot import name 'logsumexp' on KeyedVectors
* [#2928](https://github.com/RaRe-Technologies/gensim/issues/2928): The type information about job_queue in word2vec is wrong
* [#2920](https://github.com/RaRe-Technologies/gensim/issues/2920): Gensim's word2vec has a loss of 0 from epoch 1?
* [#2918](https://github.com/RaRe-Technologies/gensim/issues/2918): intersect_word2vec_format lock-factor is not triggered when the data is not in the binary format
* [#2915](https://github.com/RaRe-Technologies/gensim/issues/2915): docvecs.most_similar() does not return cosine similarity = 1 for same document vector
* [#2914](https://github.com/RaRe-Technologies/gensim/issues/2914): logs are printed without values
* [#2906](https://github.com/RaRe-Technologies/gensim/issues/2906): Add 32/64-bit reporting to issue template
* [#2900](https://github.com/RaRe-Technologies/gensim/issues/2900): TypeError: summary() got an unexpected keyword argument 'sentenses'
* [#2895](https://github.com/RaRe-Technologies/gensim/issues/2895): LdaMallet eta parameter 
* [#2893](https://github.com/RaRe-Technologies/gensim/issues/2893):  _pickle.UnpicklingError: invalid load key, '\xd0'.
* [#2889](https://github.com/RaRe-Technologies/gensim/issues/2889): Document the X2Vec refactoring in the change log.
* [#2888](https://github.com/RaRe-Technologies/gensim/issues/2888): Clean up variable names
* [#2883](https://github.com/RaRe-Technologies/gensim/issues/2883): Mmap model shared between process , memory use is not stable
* [#2881](https://github.com/RaRe-Technologies/gensim/issues/2881): KeyedVectors.load_word2vec_format('w2v.bin',binary=False) raise Value Error
* [#2877](https://github.com/RaRe-Technologies/gensim/issues/2877): Finetuning gensim doc2vec
* [#2876](https://github.com/RaRe-Technologies/gensim/issues/2876): corpus.mallet returned non-zero exit status 1.
* [#2870](https://github.com/RaRe-Technologies/gensim/issues/2870): Error 
* [#2858](https://github.com/RaRe-Technologies/gensim/issues/2858): How To Save/Plot Poincare Embedding Model After Every Epoch?
* [#2857](https://github.com/RaRe-Technologies/gensim/issues/2857): Type mismatch in utils.py: use an iterator as an iterable
* [#2856](https://github.com/RaRe-Technologies/gensim/issues/2856): KeyError: "word '...' not in vocabulary" 20-newsgroups
* [#2855](https://github.com/RaRe-Technologies/gensim/issues/2855): mallet wrapper return non-zero exit status 126
* [#2850](https://github.com/RaRe-Technologies/gensim/issues/2850): AttributeError: 'Doc2VecTrainables' object has no attribute 'vectors_lockf'
* [#2847](https://github.com/RaRe-Technologies/gensim/issues/2847): doc2vec - infer_vector gives different values on every run
* [#2845](https://github.com/RaRe-Technologies/gensim/issues/2845): id2word returns weird numbers instead of words with get_topic_terms
* [#2844](https://github.com/RaRe-Technologies/gensim/issues/2844): Conflicts between hyperparameters for negative sampling? 
* [#2843](https://github.com/RaRe-Technologies/gensim/issues/2843): Conflicts between hyperparameters for negative sampling?
* [#2835](https://github.com/RaRe-Technologies/gensim/issues/2835): Can't get attribute 'callback' on <module '__main__' (built-in)> when calling load()
* [#2834](https://github.com/RaRe-Technologies/gensim/issues/2834): api.load of fasttext vectors returns Word2VecKeyedVectors 
* [#2827](https://github.com/RaRe-Technologies/gensim/issues/2827): module 'gensim.models.doc2vec' has no attribute 'FAST_VERSION'
* [#2826](https://github.com/RaRe-Technologies/gensim/issues/2826): can't install gensim to conda because of python version 3.8
* [#2820](https://github.com/RaRe-Technologies/gensim/issues/2820): Prepare gensim 3.8.3
* [#2771](https://github.com/RaRe-Technologies/gensim/issues/2771): Get rid of tox progress bars in CI
* [#2756](https://github.com/RaRe-Technologies/gensim/issues/2756): "Pattern library is not installed" -- this still is a problem
* [#2747](https://github.com/RaRe-Technologies/gensim/issues/2747): Rackspace account discontinued 
* [#2746](https://github.com/RaRe-Technologies/gensim/issues/2746): Word2Vec ns_exponent cannot be changed from default
* [#2724](https://github.com/RaRe-Technologies/gensim/issues/2724): load_facebook_model memory footprint
* [#2667](https://github.com/RaRe-Technologies/gensim/issues/2667): FastTextKeyedVectors.word_vec(...,use_norm=True) nonsensically summing ngrams norms
* [#2658](https://github.com/RaRe-Technologies/gensim/issues/2658): Can you add compute Loss feature (as seen in Word2Vec model) in the fasttext Implementation too. It will help us in sending a callback to monitor the loss at each epoch.
* [#2593](https://github.com/RaRe-Technologies/gensim/issues/2593): Memory error while processing with mm.serialize() method
* [#2548](https://github.com/RaRe-Technologies/gensim/issues/2548): Drop Py2 support
* [#2537](https://github.com/RaRe-Technologies/gensim/issues/2537): FastText inconsistent normalization in adjust_vectors method vs word_vec
* [#2502](https://github.com/RaRe-Technologies/gensim/issues/2502): High RAM usage when loading FastText Model on Google Colab
* [#2140](https://github.com/RaRe-Technologies/gensim/issues/2140): Phrases keeps learned vocabs as bytestring
* [#1900](https://github.com/RaRe-Technologies/gensim/issues/1900): Seems Word2VecKeyedVectors.get_keras_embedding should take care of 'mask_zero'
* [#1506](https://github.com/RaRe-Technologies/gensim/issues/1506): Make `Phrases` model stopword-aware to prevent non-adjacent pairings
* [#801](https://github.com/RaRe-Technologies/gensim/issues/801): WARNING : supplied example count did not equal expected count

piskvorky:
Yes, I don't see anything release-note-worthy in there either.

piskvorky (approved):

gensim/models/keyedvectors.py
piskvorky: Needs way better documentation in the code then. I thought the KeyedVectors keys are keys, and all code (that I saw or wrote) relies on that.

What are the actual invariants?

I'm strongly in favour of discontinuing this optimization in 4.0.

piskvorky:
@mpenkov thanks, LGTM!

mpenkov:
> OTOH, it makes sense to dump all closed (better: resolved) tickets too, for completeness, somewhere at the bottom. WDYT?

I guess so, but:

- It's a lot of noise with relatively little signal (most interesting stuff gets covered by the PRs)
- This information is publicly available to anyone who _does_ need it
- Including and maintaining that information separately requires effort, and it may not be worth it given the relatively small payoff

In my opinion, seeing the list of issues is helpful for us when we're authoring release notes, but that's about it. We already have a tool that helps us do that (the new generator script) so any further work in this direction will bring diminishing returns IMHO.

PR #3000, opened 2020-11-17
Tidy up KeyedVectors.most_similar() API

### LINKED ISSUE #2998: KeyedVectors.most_similar() API has a confusing wrinkle
There is a shorthand written into the API for `KeyedVectors.most_similar()` that can result in confusion:

https://github.com/RaRe-Technologies/gensim/blob/a28a3e79f4c4cbd97106fc9b35034707295994cd/gensim/models/keyedvectors.py#L745-L747

I imagine this reduces more confusion than it causes, but it can cause the unexpected result that 
```python
model.mv.most_similar(positive="dog")
```
works as expected, but 
```python
model.wv.most_similar(negative="dog")
```
raises a potentially confusion exception.  While neither is "correct" according to the docs, this result can be unexpected (see https://twitter.com/quinnanya/status/1318398096041078784).

It seems like it would be straightforward to either accept a `KEY_TYPES` argument in *either* case, or to catch the error earlier and return a more helpful error message.

simonwiles:
Fixes #2998. 

This PR:
* allows passing a single string-key as the `negative` argument
* allows passing a single vector as the `positive` or `negative` argument
* allows passing single values (string-keys or vectors) for both `positive` and `negative` arguments in the same call
* modifies `most_similar_cosmul()` to match this behaviour as well.

The API is now, imo, a little more straightforward in that it simply wraps single values in a list where appropriate.

There's a comment in `most_similar_cosmul` at line ~981:
```
# TODO: Update to better match & share code with most_similar()
```
but I'm not sure whether the changes in the last commit completely address this or not, so I've left the comment in place.

I don't think these changes necessitate an update to the documentation, which has always specified that lists should be passed to these methods -- this is just a defensive pattern that preserves the shortcut that's in production and makes the API more resistant to violating the principle of least astonishment :)


simonwiles:
w/r/t the remove of `and not negative` in 475b013: It doesn't affect the ability to pass `positive` as a positional parameter (the `most_similar('dog')` case), which presumably is invoked in the wild all over the place and so oughtn't to be broken, and it enables calling, e.g. `most_similar(positive='dog', negative='cat')` (or the same with single vectors), which seems appropriate.

I presume the original intention was to guard against someone calling `most_similar('word1', 'word2')` with the expectation that they would both be interpreted as `positive`?  Perhaps?  Not sure.  In any event, the only result of `and not negative` was that `most_similar('word1', 'word2')` was parsed as `most_similar(['w', 'o', 'r', 'd', '1'], ['w', 'o', 'r', 'd', 2'])` instead of `most_similar(['word1'], ['w', 'o', 'r', 'd', 2'])`, so I figure this is an improvement.

(In fact, in the case of the tweet I cited, it was only the good fortune that the key `'h'` wasn't in the model that caused the error and prompted the call for help -- later in the thread the researcher remarks that they'd been using `negative='word'` for some time and never seen a problem before; results were being returned based on the parsing as `negative=['w', 'o', 'r', 'd']` and they'd been taken to be legitimate.)

gensim/models/keyedvectors.py
```diff
@@ -742,10 +742,14 @@ def most_similar(
             clip_start = 0
             clip_end = restrict_vocab
 
-        if isinstance(positive, KEY_TYPES) and not negative:
-            # allow calls like most_similar('dog'), as a shorthand for most_similar(['dog'])
+        if isinstance(positive, KEY_TYPES + (ndarray,)):
+            # allow passing a single string-key or vector for the positive argument
```
gojomo, line 746: Is there a risk that something that may have previously worked, like an ndarray of multiple vectors, will stop working here, with the extra `list` wrapping?

gojomo:
Thanks for this, sorry it's taken so long to review. A few test-cases that verify/demo the new possibilities (and ensure that the line-comment I made about a possible lost-possibility isn't the case) would make this an easy merge.

gensim/models/keyedvectors.py
```diff
@@ -985,10 +989,14 @@ def most_similar_cosmul(self, positive=None, negative=None, topn=10):
 
         self.fill_norms()
 
-        if isinstance(positive, str) and not negative:
-            # allow calls like most_similar_cosmul('dog'), as a shorthand for most_similar_cosmul(['dog'])
+        if isinstance(positive, KEY_TYPES + (ndarray,)):
+            # allow passing a single string-key or vector for the positive argument
             positive = [positive]
 
+        if isinstance(negative, KEY_TYPES + (ndarray,)):
```
mpenkov, line 996: You're duplicating the key types manipulation in 4 places.

Please do something like:

```
key_types = KEY_TYPES + (ndarray, )
if instance(positive, key_types):
    ...
```

mpenkov:
@simonwiles Ping. Are you able to complete this PR?

mpenkov:
Note to self:

- Before merging, this PR needs tests
- Minor code cleanup possible (see suggestions above)

gensim/models/keyedvectors.py
```diff
@@ -191,6 +191,8 @@
 
 KEY_TYPES = (str, int, np.integer)
 
+_EXTENDED_KEY_TYPES = (str, int, np.integer, np.ndarray)
+
```
gojomo, line 195: This doesn't match the style of the immediately-preceding module constant: shouldn't they both be underscore-preceded, or neither? (And: I prefer to be sparing with leading-underscores, only where there's a strong implication of "internal only".)

gensim/models/keyedvectors.py
gojomo: Still a concern here.

gojomo: Still a concern here.

gensim/models/keyedvectors.py
mpenkov: OK, I've made them both internal.

gensim/models/keyedvectors.py
mpenkov: @gojomo @piskvorky How do you suggest we resolve these concerns?

mpenkov: @gojomo @piskvorky Ping. This is blocking the release.

gojomo: Check if ndarray-of-multiple-vectors used to work as any of these arguments, and if so, if it's now broken. If so, then decide whether to fix to retain old behavior for that usage, or just add to release-note or doc-comment as rare case no longer supported.

piskvorky: I checked and `model.most_similar(positive=np.array([vec1, vec2]))` indeed does work in `develop`. And will likely stop working here.

mpenkov: OK. Finally had a chance to look at this. Please have a look at let me know if I've missed anything.

gojomo:
Slightly-larger #3188 may make sense to apply 1st, then this.

gensim/models/keyedvectors.py
```diff
@@ -189,7 +189,28 @@
...
-KEY_TYPES = (str, int, np.integer)
+_KEY_TYPES = (str, int, np.integer)
+
+_EXTENDED_KEY_TYPES = (str, int, np.integer, np.ndarray)
+
+
+def _ensure_list(value):
+    #
+    # Ensure that the specified value is a list.  Sometimes users pass a single
+    # value when they should really pass a list containing value.
```
gojomo, line 200: Slightly off to say users "should" pass a list - when docs/examples have always encouraged other forms (& a single string key is probably the most common usage). Instead I'd say simply: "Ensure that the specified value is wrapped in a list, for those supported cases where we also accept a single key or vector."

mpenkov:
@piskvorky Is #3188 in scope for the current release? It isn't mentioned anywhere in the associated milestone or project. If not, my preference is to:

1. Merge this PR
2. Finally release 4.1
3. Move on with other PRs

Please let me know your thoughts.

piskvorky:
My preference is to get in all API-breaking changes ASAP. Looking at #3188, it seems backward-compatible, so can wait potentially.

Of course, merging both this and #3188 is ideal, but I see @gojomo added new (good) review comments to #3188. So if #3188 would mean blocking the release, we can release without.

gensim/models/keyedvectors.py
```diff
@@ -189,7 +189,28 @@
...
 
 
-KEY_TYPES = (str, int, np.integer)
+_KEY_TYPES = (str, int, np.integer)
+
+_EXTENDED_KEY_TYPES = (str, int, np.integer, np.ndarray)
+
+
+def _ensure_list(value):
+    #
```
piskvorky, line 198: Why not a standard `""""` docstring, for these two lines? (the rationale below can stay as comment)

Sphinx doesn't render docs for `_underscore_methods`, right?

mpenkov: No particular reason. I'll make it into a docstring.

Yes, this docstring won't end up in the docs, because the function is _internal.

mpenkov:
Merged. Thank you @simonwiles !

PR #3005, opened 2020-11-26
Add epochs in transform function of sklearn wrapper for doc2vec

rogozinushka:
Hello friends. I want to speak about the sklearn wrapper for doc2vec.

The `infer_vector` function has the `epochs` parameter, but the `transform` function in the sklearn wrapper does not. Sometimes the `epochs` parameter in the doc2vec initialization is not equal to the` epochs` parameter in the `infer_vector` function, but the sklearn wrapper uses the `epochs` from the `fit` function. We can add the `transform_epochs` parameter to the `fit` function in the sklearn wrapper. This parameter is stored in the `transform_epochs` variable and is used in the `transform` function where the `infer_vectors(epochs=transform_epochs))` is called.

### DIFF
--- a/gensim/sklearn_api/d2vmodel.py
+++ b/gensim/sklearn_api/d2vmodel.py
@@ -40,7 +40,7 @@ class D2VTransformer(TransformerMixin, BaseEstimator):
                  dv_mapfile=None, comment=None, trim_rule=None, vector_size=100, alpha=0.025, window=5,
                  min_count=5, max_vocab_size=None, sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
                  hs=0, negative=5, cbow_mean=1,
-                 hashfxn=hash, epochs=5, sorted_vocab=1, batch_words=10000):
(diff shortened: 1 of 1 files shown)


piskvorky:
Hello @rogozinushka , we're discontinuing 3rd party wrappers in Gensim, including for sklearn.

If you need those, please copy out the wrapper code (for now at [d2vmodel.py](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/sklearn_api/d2vmodel.py)) and modify and use it as needed yourself. The wrapper code is quite trivial anyway.

CC @mpenkov .

rogozinushka:
It is sad. The sklearn wrapper for doc2vec can be useful for sklearn transformers and pipelines. For example, the first steps consist of text cleaning and the last step is sklearn wrapper for doc2vec.

piskvorky:
Yes, that's the idea! That's why we added these wrappers.

But there's no one to maintain them, so it's best the wrappers continue live as external / contributed packages, not a part of core Gensim.

mpenkov:
@rogozinushka Thank you for your contribution, but the relevant wrappers have been removed, so this PR should be closed.

PR #3006, opened 2020-11-26
Make Wikipedia metadata accessible

kumar-neelabh:
Let the users have metadata (e.g. title) if they need it. Added an argument in WikiCorpus init() to specify if metadata is needed. Previously, it was set to False and could not be toggled.

### DIFF
--- a/gensim/corpora/wikicorpus.py
+++ b/gensim/corpora/wikicorpus.py
@@ -576,7 +576,7 @@ class WikiCorpus(TextCorpus):
     """
     def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), dictionary=None,
                  filter_namespaces=('0',), tokenizer_func=tokenize, article_min_tokens=ARTICLE_MIN_WORDS,
-                 token_min_len=TOKEN_MIN_LEN, token_max_len=TOKEN_MAX_LEN, lower=True, filter_articles=None):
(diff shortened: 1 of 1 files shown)


gensim/corpora/wikicorpus.py
```diff
@@ -576,7 +576,7 @@ class WikiCorpus(TextCorpus):
     """
     def __init__(self, fname, processes=None, lemmatize=utils.has_pattern(), dictionary=None,
                  filter_namespaces=('0',), tokenizer_func=tokenize, article_min_tokens=ARTICLE_MIN_WORDS,
-                 token_min_len=TOKEN_MIN_LEN, token_max_len=TOKEN_MAX_LEN, lower=True, filter_articles=None):
+                 token_min_len=TOKEN_MIN_LEN, token_max_len=TOKEN_MAX_LEN, lower=True, filter_articles=None, metadata=False):
```
piskvorky, line 579: All parameters must be documented.

Also, this will probably fail the code style check (line too long).

piskvorky (changes requested):

gensim/corpora/wikicorpus.py
kumar-neelabh: Okay, closing this pull request and creating another after making required changes.

piskvorky: You could have applied the changes to this PR. Pull requests are editable.

PR #3012, opened 2020-12-18
remove pattern dependency

### LINKED ISSUE #2697: Cannot use gensim 3.8.x when `nltk` package is installed
#### Problem description

> What are you trying to achieve? What is the expected result? What are you seeing instead?

In my script i'm trying to `import gensim.models.keyedvectors` and also import another package, that requires `nltk` package internally. Whenever i have NLTK installed in the same virtualenv (i'm not using virtualenv, but a docker image actually) - the gensim model fails to import.

```
# pip list | grep -E 'gensim|nltk'
gensim                        3.8.1

# pip install nltk
Processing /root/.cache/pip/wheels/96/86/f6/68ab24c23f207c0077381a5e3904b2815136b879538a24b483/nltk-3.4.5-cp36-none-any.whl
Requirement already satisfied: six in /usr/local/lib/python3.6/site-packages (from nltk) (1.13.0)
Installing collected packages: nltk
Successfully installed nltk-3.4.5

# pip list | grep -E 'gensim|nltk'
gensim                        3.8.1
nltk                          3.4.5

# python
Python 3.6.8 (default, Jun 11 2019, 01:16:11)
[GCC 6.3.0 20170516] on linux
  File "/usr/local/lib/python3.6/site-packages/gensim/__init__.py", line 5, in <module>
    from gensim import parsing, corpora, matutils, interfaces, models, similarities, summarization, utils  # noqa:F401
  File "/usr/local/lib/python3.6/site-packages/gensim/corpora/__init__.py", line 14, in <module>
    from .wikicorpus import WikiCorpus  # noqa:F401
  File "/usr/local/lib/python3.6/site-packages/gensim/corpora/wikicorpus.py", line 539, in <module>
    class WikiCorpus(TextCorpus):
  File "/usr/local/lib/python3.6/site-packages/gensim/corpora/wikicorpus.py", line 577, in WikiCorpus
  File "/usr/local/lib/python3.6/site-packages/gensim/utils.py", line 1614, in has_pattern
    from pattern.en import parse  # noqa:F401
  File "/usr/local/lib/python3.6/site-packages/pattern/text/en/__init__.py", line 61, in <module>
    from pattern.text.en.inflect import (
  File "/usr/local/lib/python3.6/site-packages/pattern/text/en/__init__.py", line 80, in <module>
    from pattern.text.en import wordnet
  File "/usr/local/lib/python3.6/site-packages/pattern/text/en/wordnet/__init__.py", line 57, in <module>
    nltk.data.find("corpora/" + token)
  File "/usr/local/lib/python3.6/site-packages/nltk/data.py", line 673, in find
    return find(modified_name, paths)
  File "/usr/local/lib/python3.6/site-packages/nltk/data.py", line 660, in find
    return ZipFilePathPointer(p, zipentry)
  File "/usr/local/lib/python3.6/site-packages/nltk/compat.py", line 228, in _decorator
    return init_func(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/nltk/data.py", line 506, in __init__
    zipfile = OpenOnDemandZipFile(os.path.abspath(zipfile))
  File "/usr/local/lib/python3.6/site-packages/nltk/compat.py", line 228, in _decorator
    return init_func(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/nltk/data.py", line 1055, in __init__
    zipfile.ZipFile.__init__(self, filename)
  File "/usr/local/lib/python3.6/zipfile.py", line 1131, in __init__
    self._RealGetContents()
  File "/usr/local/lib/python3.6/zipfile.py", line 1198, in _RealGetContents
    raise BadZipFile("File is not a zip file")
zipfile.BadZipFile: File is not a zip file
```

#### Versions

```python
 python
Python 3.6.8 (default, Jun 11 2019, 01:16:11)
[GCC 6.3.0 20170516] on linux
Linux-5.0.0-050000rc8-generic-x86_64-with-debian-9.11
Python 3.6.8 (default, Jun 11 2019, 01:16:11)
[GCC 6.3.0 20170516]
NumPy 1.17.4
SciPy 1.3.3
gensim 3.8.1
FAST_VERSION 1
```

mpenkov:
- Fix #2697


gensim/scripts/segment_wiki.py
```diff
@@ -278,9 +278,6 @@ def __init__(self, fileobj, min_article_character=200, processes=None,
             Minimal number of character for article (except titles and leading gaps).
         processes : int, optional
             Number of processes, max(1, multiprocessing.cpu_count() - 1) if None.
-        lemmatize : bool, optional
```
piskvorky, line 281: Should we keep some sort of note? This change will break people's existing code, so a helpful message would be… helpful :)

setup.py
```diff
@@ -327,13 +321,14 @@ def run(self):
...
 #
-CYTHON_STR = 'Cython==0.29.14'
+CYTHON_STR = 'Cython==0.29.21'
 
 install_requires = [
     NUMPY_STR,
     'scipy >= 0.18.1',
     'smart_open >= 1.8.1',
     "dataclasses; python_version < '3.7'",  # pre-py3.7 needs `dataclasses` backport for use of `dataclass` in doc2vec.py
+    'six',
```
piskvorky, line 331: What for? Looks like a regression, we already got rid of six in previous PRs IIRC.

gensim/scripts/segment_wiki.py
gojomo: Yes, a no-longer-supported failure message if `lemmatize=True` - perhaps pointing to external lemmatization options – would seem polite.

setup.py
mpenkov: gensim wouldn't build without it, so I added it as a temporary workaround.

Upon further digging, we haven't removed it completely:

```
$ grep -R "from six" gensim
gensim/models/word2vec_corpusfile.pyx:from six import iteritems
gensim/models/word2vec_corpusfile.cpp: * from six import iteritems
gensim/models/word2vec_corpusfile.cpp: * from six import iteritems
```

plus the one occurrence I removed in this PR.

gensim/scripts/segment_wiki.py
mpenkov: This would go in the change log, right?

docs/src/scripts/make_wiki_online.rst
```diff
@@ -1,5 +1,5 @@
 :mod:`scripts.make_wiki_online` -- Convert articles from a Wikipedia dump
-=========================================================================
+=================================================================================
```
mpenkov, line 2:
```suggestion
=========================================================================
```

setup.py
piskvorky: Aha, I guess I only searched in `*.py` when I did my purging. I'll add a commit here for the `.pyx` and `.cpp` too, so that we're really six-free.

mpenkov: I handled the pyx already.

The cpp is autogenerated, so we shouldn't touch it.

mpenkov:
Is there anything left to do here? I think this may be ready to merge.

gensim/scripts/segment_wiki.py
```diff
@@ -278,22 +278,24 @@ def __init__(self, fileobj, min_article_character=200, processes=None,
...
-            Otherwise, use simple regexp tokenization.
         filter_namespaces : tuple of int, optional
             Enumeration of namespaces that will be ignored.
         include_interlinks: bool
             Whether or not interlinks should be included in the output
 
         """
+        if lemmatize is not None:
+            raise NotImplementedError(
+                'The lemmatize parameter is no longer supported. '
```
piskvorky, line 289:
```suggestion
                'The lemmatize parameter is no longer supported since Gensim 4.0.0. '
```

gensim/scripts/segment_wiki.py
```diff
@@ -278,22 +278,24 @@ def __init__(self, fileobj, min_article_character=200, processes=None,
...
+                'If you need to lemmatize, use e.g. <https://github.com/clips/pattern>.'
```
piskvorky, line 290:
```suggestion
                'If you need to lemmatize, use e.g. https://github.com/clips/pattern to preprocess your corpus before submitting it to Gensim.'
```

piskvorky:
Yes, looks ready, but all tests are failing.

mpenkov:
Those failures are unrelated. In between me making this PR one week ago, and the build triggered by your changes to segment_wiki.py, there has been a new sklearn release and they deprecated one of the modules we were relying on.

```
$ python
Python 3.8.5 (default, Jul 21 2020, 10:48:26)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
>>> import sklearn.datasets.base
/Users/misha/envs/gensim/lib/python3.8/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.datasets.base module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.datasets. Anything that cannot be imported from sklearn.datasets is now part of the private API.
  warnings.warn(message, FutureWarning)
```

I'll disable the failing tests and open a ticket to deal with the problem properly.

mpenkov:
https://github.com/RaRe-Technologies/gensim/issues/3016

gojomo:
This is another case where for the sake of maintaining the tests a trustworthy indicator of whether the code is ready for release, I would not be so quick to disable them, even with a note-linking-to-a-pending-issue. 

If previous functionality is now broken due to outside library changes – and also broken on `develop`, not just this PR branch where it's 1st noticed – the tests *should* remain failing until either (a) a true fix/update is made; or (b) an explicit decision is made/announced/added-to-release-notes that such functionality is no longer supported (or temporarily not supported).

If test-disablement is chosen, because that functionality can be ignored for a while, it'd be best done in a standalone PR that's applied to `develop` separately, rather than mixed with a different set of in-process changes.

piskvorky:
Yes – @mpenkov can you please look into removing sklearn, as the proper fix, independent of this PR?

mpenkov:
Sure, I'll have a look at it, most likely in the New Year.

I disabled the tests because it was least bad alternative out of the following list:

1) Do not merge until tests are properly fixed (I don't know when that will be,, and chances are this PR will be stale by then)
2) Merge with broken tests (this is bad form, because now other contributors will trip over the same failures)
3) Disable unit tests temporarily

If you disagree with my decision, I can re-enable the tests in a separate PR. It's simple.

gojomo:
> 2. Merge with broken tests (this is bad form, because now other contributors will trip over the same failures)

If this PR's changes caused the test-failures (even via bumping dependency versions), I'd agree that merging it would be a problem. But, I'm pretty sure `develop` without this PR would also be failing, given the upstream changes. The `develop` trunk only looks 'healthy' because no build/test has been triggered, and this PR was only "unlucky" to be the 1st experiencing the side-effects of external changes. 

Given that, merging this PR, with that known-but-unrelated problem, wouldn't create any *new* problems for others, so I think (2) would be OK. With regard to passing tests, I think the precise policy could be: "any PR shouldn't break any tests that were working before the PR - but it doesn't have to fix everything that's currently broken".

PR #3023, opened 2021-01-11
Remove wrappers

mpenkov:


piskvorky:
Thanks! What's the relationship between this PR and #2972?

mpenkov:
Ah, it's redundant. I think the only thing useful here is the Cython version bump.

mpenkov:
Thanks for reminding me about #2972, I'll take it from there.

piskvorky:
This PR also removed the sklearn wrappers (which is desirable), while #2972 didn't.

mpenkov:
Yeah, that part was trivial. I did this while in transit, which is why I neglected to check for #2972.

PR #3035, opened 2021-01-30
Update repos before trying to install gdb

janaknat:
I was able to reproduce the behaviour with the test failing. Updating the repo before trying to install gdb helped.

Build log: https://github.com/janaknat/gensim/actions/runs/523516055


mpenkov:
Good work! Looks like the workflow completes successfully now.

PR #3055, opened 2021-02-27
rm gensim.viz submodule

mpenkov:
Partially address #2852


piskvorky (approved):

piskvorky (commented):
How about removing plotly too? Still needed?

mpenkov:
Good point, let's remove it.

mpenkov:
Added to https://github.com/RaRe-Technologies/gensim/wiki/Migrating-from-Gensim-3.x-to-4

PR #3057, opened 2021-03-01
modify comment in LdaModel.inference() to be more explicit

yocen:
One comment  in LdaModel.inference() confused me:
`# The optimal phi_{dwk} is proportional to expElogthetad_k * expElogbetad_w.`
And I think it should modified to this so the comment would be more accurate:
`# The optimal phi_{dwk} is proportional to expElogthetad_k * expElogbetad_kw.`

Reasons:
1.
According to Algorithm 1 on paper:
Online Learning for Latent Dirichlet Allocation, NIPS 2010 <http://www.cs.princeton.edu/~mdhoffma>, 
phi_dwk is proportional to exp(Elogtheta_dk + Elogbeta_kw).
2.
`phi_{dwk}` is a scalar,  `expElogthetad_k ` is a scalar, and `expElogbetad_w `is a vector with K elements,  so  `expElogthetad_k * expElogbetad_w` will return a vector with K elements, which does not match the type of `phi_{dwk}`. If we use `expElogthetad_k * expElogbetad_kw` instead, in which `expElogbetad_kw `is a scalar, it matches.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -687,7 +687,7 @@ class LdaModel(interfaces.TransformationABC, basemodel.BaseTopicModel):
             expElogthetad = expElogtheta[d, :]
             expElogbetad = self.expElogbeta[:, ids]
 
-            # The optimal phi_{dwk} is proportional to expElogthetad_k * expElogbetad_w.
+            # The optimal phi_{dwk} is proportional to expElogthetad_k * expElogbetad_kw.
             # phinorm is the normalizer.
             # TODO treat zeros explicitly, instead of adding epsilon?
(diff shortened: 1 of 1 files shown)


mpenkov (approved):
Thank you for pointing this out to us!

jonaschn:
@yocen Thanks for pointing that out. 
@olavurmortensen implemented an efficient version of computing the phinorm for the Author-Topic Model.
I am not sure if the phinorm is the same for LDA and the author-topic model.
If this is the case it would make sense to "backport" @olavurmortensen idea to the LDAModel.

https://github.com/RaRe-Technologies/gensim/blob/2feef89a24c222e4e0fc6e32ac7c6added752c26/gensim/models/atmodel.py#L377

PR #3059, opened 2021-03-02
Fix race condition in FastText tests

sleepy-owl:
This is in relation to #3050 

I found two main problems which were causing the tests to fail intermittently when run in parallel. I found this problem in at least 4 tests during my initial analysis.

1. `get_tmpfile(x)` only creates the file name `/tmp/x` which is not unique, and can be overwritten when tests run in parallel
2. `temporary_file` will not create temporary directory, when absolute path is provided (by `get_tmpfile`)

Hence, when running these tests in parallel, they write to the same file, e.g., `/tmp/gensim_fasttext.tst` causing them to fail

In this PR, I fix this by 
1. Creating a temporary directory to be used by `get_tmpfile`
2. Using only `temporary_file` instead of `temporary_file(get_tmpfile(...))` 

These changes fix all the flaky failures reported previously and most likely many others. If these changes makes sense, I can extend this fix to `test_word2vec.py` and `test_doc2vec.py` which also have the 2nd problem described above.

Let me know what you guys think. Thanks!

### DIFF
--- a/gensim/test/test_fasttext.py
+++ b/gensim/test/test_fasttext.py
@@ -107,7 +107,7 @@ class TestFastTextModel(unittest.TestCase):
         self.assertRaises(TypeError, model.train, corpus_file=sentences, total_examples=1, epochs=1)
 
     def test_training_fromfile(self):
-        with temporary_file(get_tmpfile('gensim_fasttext.tst')) as corpus_file:
+        with temporary_file('gensim_fasttext.tst') as corpus_file:
             utils.save_as_line_sentence(sentences, corpus_file)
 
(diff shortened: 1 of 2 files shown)


sleepy-owl:
@mpenkov @piskvorky

piskvorky:
Oh this is great work. Thanks so much @sleepy-owl, the flakiness has been bugging us for years!

mpenkov:
Merged. Good work @sleepy-owl , and congrats on your first contribution to gensim! Thank you and keep up the good work.

By all means, if you find similar issues in other parts of the code base, please make a PR. Cheers!

sleepy-owl:
Thanks, folks! Will try to resolve more such issues :)

PR #3060, opened 2021-03-04
Record lifecycle events in Gensim models

### LINKED ISSUE #2863: Keep "lifecycle log" in models
## Why

When Gensim users report issues with models, we have trouble deciding what they actually did, what version they used, etc. They're often confused themselves.

This leads to protracted communication => increased support effort => slower issue resolution and more annoyance all around. Not good.

## What

Keep an internal "lifecycle log" inside each model. This could be as simple as an internal list-of-strings attribute, e.g. `word2vec.lifecycle = ["init() created on 2020-06-20 15:49:03 UTC with Gensim 3.8.3 on Python 3.6", "train() called on 2020-06-20 15:49:04 UTC", "load() called on 2020-06-29 8:7:58 UTC", …]`. We would serialize this log on `save()`, and ask users to provide its value when investigating an issue.

We probably want to keep this simple and human-readable, both in the API and log content. Not to over-engineer to start with – not attempting a full model recreation from the log or anything, I can see how this could turn into a rabbit hole.

*Originally conceived by @gojomo in https://github.com/RaRe-Technologies/gensim/pull/2698#issuecomment-646929210.*

piskvorky:
Fixes #2863. See the discussion there for motivation.


piskvorky:
@gojomo @mpenkov is this what you had in mind?

If OK, I'll add something similar to all other classes / models as well.

gensim/models/word2vec.py
```diff
@@ -481,6 +484,7 @@ def build_vocab(
         report_values = self.prepare_vocab(update=update, keep_raw_vocab=keep_raw_vocab, trim_rule=trim_rule, **kwargs)
         report_values['memory'] = self.estimate_memory(vocab_size=report_values['num_retained_words'])
         self.prepare_weights(update=update)
+        self.add_lifecycle_event("build_vocab", update=update, trim_rule=str(trim_rule))
```
mpenkov, line 487: Should we log some (or all) the parameters here?

gojomo: As a minimal internal log, I'd think *all* parameters could be unwieldy, but a major switch like `update` makes sense, and even (where not too hard) some very-high-level indicators of the magnitude of the operation, such as elapsed time and before/after model vocabulary size.

gensim/__init__.py
```diff
@@ -4,11 +4,12 @@
 
 """
 
+__version__ = '4.0.0beta'
```
gojomo, line 7: Tangential comment: As a matter of labeling-correctness, '4.0.0beta' is already released and frozen, so adding this label now can't possibly be a correct description of code-from-this-branch. What I've seen as a useful process before - unless some sort of auto process is auto-labeling the code tree on every test build - is to immediately promote this identifier to some ".dev"/etc extension of the last version, or preview of the next-expected-version, so that a precise release label like this only ever exists in the repo for the brief moment that it's true, and the code is feeding a build-for-release.

mpenkov: Yes, .dev1, .dev2, etc builds are helpful.

piskvorky: You're right. Should have been incremented right after the beta release according to [our own instructions](https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#update-gensim-version-to-report-the-next-release).

I'll update it in this PR.

gensim/models/word2vec.py
piskvorky: Yes – we want to record some parameters, but not all.

If you have concrete ideas what should be recorded where, please push a suggestion / commit. Different events call for different semantics, so I don't think this can be automated.

mpenkov:
Do we need to document how to use this stuff? Or is this for us gensim developers only?

piskvorky:
What documentation are you thinking of? (beyond the docstring I put in there)

Yes, I conceive this event log as mostly for us. But that may be just my lack of imagination. What public uses / documentation do you envision? CC @gojomo

mpenkov:
Yeah, putting this in the issue template would help, because it would save us asking them what they did to the model.

That would be sufficient for doco purposes, IMHO.

piskvorky:
Okay. Done.

Issue #2967, opened 2020-09-30
Adopting a (narrow) backward-compatibility standard; implications for 4.0.0


gojomo:
From #2939:

# Decisions

1. Support loading of pre-3.8.3 models?
   - Misha: better: a stand-alone script to upgrade models to the latest version… rather than class methods doing the work on each load
   - Misha: also in favour of limited backward compatibility loading (just 3.8.3)
   - Radim: in favour of keeping existing compatibility code; why remove it? pros/cons unclear
   => **decision**: talk to Gordon, make the costs (pros/cons) more explicit.

_Originally posted by @piskvorky in https://github.com/RaRe-Technologies/gensim/pull/2939#issuecomment-698247945_

piskvorky:
I'll read all this when I find some time. Until then, I'm strongly in favour of supporting old models, as far back as possible. 

Changes I'd propose would be toward making compatibility easier going forward: a dedicated "model migration script", pickling additional metadata inside models, better testing etc.

gojomo:
I made a detailed writeup & top-level issue because I'd prefer not to make a single-case decision, with some mumbled live-call compromises of the moment, where the full reasoning isn't captured and thus soon forgotten. 

I'd like to fix this as a source of bad code, technical debt, and drag on future contributions, not just rush out 4.0.0. I'd rather not have either me, nor any other contributor, waste their time on backward-compatibility code that is very likely to **never even run**, on the unproven theory that someone, somewhere *might* still be using years-old version A, and iffy feature B, and need to bring that into `gensim-4.0.0` for unknown reasons.

gojomo:
TLDR pros/cons of 3-point policy I proposed above:

Pros:
* we can clean up the sloppy mess that's the `gensim/test/test_data` directory of poorly-labeled files, and delete many of the old test-methods that (while most are passing) only provide superficial, not actual, assurance we can load 1.x, 2.x, early 3.x models
* we can do a `dev`, `alpha`, or `beta` prerelease ASAP
* we spend more time on actually-known-to-be-pain-points confusing/inefficient APIs, or requested features, rather than hypothetically-required backward-compatibility - right now and in every future release cycle
* users who hit rare loading issues & see the policy clearly understand what efforts they should exert - be that keeping up with releases or trying other processes - to update or replace their older models

Cons:
* I'm not sure of any in enough detail to describe them

piskvorky:
[...]

Issue #2848, opened 2020-05-26
Consider: dropping custom SaveLoad in favor of Pickle v.5 or joblib.dump? 


gojomo:
[PEP 574](https://www.python.org/dev/peps/pep-0574/) defines pickle-version-5, which supports alternate serialization of things like large numpy arrays. It's natively available in Python 3.8, with a PyPI backport [pickle5](https://pypi.org/project/pickle5/)  that works in 3.5, 3.6, and 3.7. 

I'm not yet certain it allows the same mmap-on-load possible with `gensim.utils.SaveLoad`, but I suspect it might. 

Similarly, `joblib`'s [dump()](https://joblib.readthedocs.io/en/latest/generated/joblib.dump.html) supports serializing large objects with large `numpy` arrays, and it what `sklearn`/etc uses for large models. Its matching [load()](https://joblib.readthedocs.io/en/latest/generated/joblib.load.html#joblib.load) has an `mmap_mode` option. 

Gensim should consider whether either of these could be a superior alternative to `gensim.utils.SaveLoad` – in terms of being in sync with Python & related projects, & minimizing gensim idiosyncratic code. If so, a transition goal cold be set in some future release, with perhaps just one version supporting both modes (if continued forward-migration of old models is to be supported).

piskvorky:
Native pickle is certainly worth a consideration, as long as it supports mmap. Mmap for IPC data sharing is important to Gensim.

`joblib` I'd be more careful about. I remember trying it out when it "appeared" (a long time ago…) and it wasn't a viable alternative to our `save`/`load`. Replacing idiosyncratic code *that works*, with another idiosyncratic code that might or might not work, is not a top of our priority list.

gojomo:
While I threw out this idea as a longer-term concept, looking now at #2957 & #1851, & with mmap-improvements (#2955) & back-compatibility standards also on my mind, some quick wins/upgrades may be quite easy. 

Potentially all of these could be helpful:

* `gensim.utils` declares a new pref for Pickle protocol=4 (or even =5 with `pickle5` backport), and uses that as default for `pickle()`/`.save()` from now on
[...]

PR #3065, opened 2021-03-07
Default to pickle protocol 4 when saving models

### LINKED ISSUE #1851: Support streaming models split into multiple files from S3 / GCS
Streaming small d2v models from s3 bucket works fine. Simply insert the s3 address into model.load. e.g. model.load('s3://<bucket>/<path>). However, when the model gets bigger and is split into multiple files all files except the main model file cannot be loaded. These other files are loaded by numpy and not smart_open. 

The essential part of my code is

```
def load_model(model_file):
    return Doc2Vec.load(model_file)

# infer 
def infer_docs(input_string, model_file, inferred_docs=5):
    model = load_model(model_file)
    processed_str = simple_preprocess(input_string, min_len=2, max_len=35)    
    inferred_vector = model.infer_vector(processed_str)
    return model.docvecs.most_similar([inferred_vector], topn=inferred_docs)
```
Trying to load from s3 on a bugger model yields: 

```
[INFO]  2018-01-21T20:44:59.613Z    f2689816-feeb-11e7-b397-b7ff2947dcec    testing keys in event dict
[INFO]  2018-01-21T20:44:59.614Z    f2689816-feeb-11e7-b397-b7ff2947dcec    loading model from s3://data-d2v/trained_models/model_law
[INFO]  2018-01-21T20:44:59.614Z    f2689816-feeb-11e7-b397-b7ff2947dcec    loading Doc2Vec object from s3://data-d2v/trained_models/model_law
[INFO]  2018-01-21T20:44:59.650Z    f2689816-feeb-11e7-b397-b7ff2947dcec    Found credentials in environment variables.
[INFO]  2018-01-21T20:44:59.707Z    f2689816-feeb-11e7-b397-b7ff2947dcec    Starting new HTTPS connection (1): s3.eu-west-1.amazonaws.com
[INFO]  2018-01-21T20:44:59.801Z    f2689816-feeb-11e7-b397-b7ff2947dcec    Starting new HTTPS connection (2): s3.eu-west-1.amazonaws.com
[INFO]  2018-01-21T20:45:35.830Z    f2689816-feeb-11e7-b397-b7ff2947dcec    loading wv recursively from s3://data-d2v/trained_models/model_law.wv.* with mmap=None
[INFO]  2018-01-21T20:45:35.830Z    f2689816-feeb-11e7-b397-b7ff2947dcec    loading syn0 from s3://data-d2v/trained_models/model_law.wv.syn0.npy with mmap=None
[Errno 2] No such file or directory: 's3://data-d2v/trained_models/model_law.wv.syn0.npy': FileNotFoundError
  File "/var/task/handler.py", line 20, in infer_handler
    event['input_text'], event['model_file'], inferred_docs=10)
  File "/var/task/infer_doc.py", line 26, in infer_docs
    model = load_model(model_file)
  File "/var/task/infer_doc.py", line 21, in load_model
    return Doc2Vec.load(model_file)
  File "/var/task/gensim/models/word2vec.py", line 1569, in load
    model = super(Word2Vec, cls).load(*args, **kwargs)
  File "/var/task/gensim/utils.py", line 282, in load
    obj._load_specials(fname, mmap, compress, subname)
  File "/var/task/gensim/models/word2vec.py", line 1593, in _load_specials
    super(Word2Vec, self)._load_specials(*args, **kwargs)
  File "/var/task/gensim/utils.py", line 301, in _load_specials
    getattr(self, attrib)._load_specials(cfname, mmap, compress, subname)
  File "/var/task/gensim/utils.py", line 312, in _load_specials
    val = np.load(subname(fname, attrib), mmap_mode=mmap)
  File "/var/task/numpy/lib/npyio.py", line 372, in load
    fid = open(file, "rb")
FileNotFoundError: [Errno 2] No such file or directory: 's3://data-d2v/trained_models/model_law.wv.syn0.npy'
```
My use case is to serve the models on a AWS lambda. My current workarond is to download all model files to a local folder and then load the model from the local folder which is rather slow

piskvorky:
Gensim has saved models with protocol=2 for years. This protocol has a limitation on the maximum object size, and is not as efficient as later protocols.

This PR changes the default protocol to 4. It's [available from py3.4+](https://docs.python.org/3/library/pickle.html#data-stream-format); Gensim currently supports 3.5+, so that's fine.

Gensim release 4.0 seems the right place for this. Old environments (python < 3.4) won't be able to load models created with Gensim 4.0+, but because we made so many changes in 4.0, this is the place to cut the support anyway.

Fixes #1851.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes
 
 - fix RuntimeError in export_phrases (change defaultdict to dict) (PR [#3041](https://github.com/RaRe-Technologies/gensim/pull/3041), [@thalishsajeed](https://github.com/thalishsajeed))
 - Record lifecycle events in Gensim models (PR [#3060](https://github.com/RaRe-Technologies/gensim/pull/3060), [@piskvorky](https://github.com/piskvorky))
+- Default to pickle protocol 4 when saving models (PR [#3065](https://github.com/RaRe-Technologies/gensim/pull/3065), [@piskvorky](https://github.com/piskvorky))
 
 ## 4.0.0beta, 2020-10-31
 
--- a/gensim/similarities/annoy.py
+++ b/gensim/similarities/annoy.py
@@ -84,7 +84,7 @@ class AnnoyIndexer():
                 raise ValueError("Only a Word2Vec, Doc2Vec, FastText or KeyedVectors instance can be used")
             self._build_from_model(kv.get_normed_vectors(), kv.index_to_key, kv.vector_size)
 
-    def save(self, fname, protocol=2):
+    def save(self, fname, protocol=utils.PICKLE_PROTOCOL):
         """Save AnnoyIndexer instance to disk.
 
         Parameters
--- a/gensim/similarities/nmslib.py
+++ b/gensim/similarities/nmslib.py
@@ -85,6 +85,7 @@ try:
 except ImportError:
     raise ImportError("NMSLIB not installed. To use the NMSLIB indexer, please run `pip install nmslib`.")
 
+from gensim import utils
 from gensim.models.doc2vec import Doc2Vec
 from gensim.models.word2vec import Word2Vec
 from gensim.models.fasttext import FastText
@@ -141,7 +142,7 @@ class NmslibIndexer():
             else:
                 raise ValueError("model must be a Word2Vec, Doc2Vec, FastText or KeyedVectors instance")
 
-    def save(self, fname, protocol=2):
+    def save(self, fname, protocol=utils.PICKLE_PROTOCOL):
         """Save this NmslibIndexer instance to a file.
 
         Parameters
--- a/gensim/utils.py
+++ b/gensim/utils.py
@@ -567,7 +567,7 @@ class SaveLoad:
         compress, suffix = (True, 'npz') if fname.endswith('.gz') or fname.endswith('.bz2') else (False, 'npy')
         return compress, lambda *args: '.'.join(args + (suffix,))
 
-    def _smart_save(self, fname, separately=None, sep_limit=10 * 1024**2, ignore=frozenset(), pickle_protocol=2):
+    def _smart_save(self, fname, separately=None, sep_limit=10 * 1024**2, ignore=frozenset(), pickle_protocol=4):
         """Save the object to a file. Used internally by :meth:`gensim.utils.SaveLoad.save()`.
 
         Parameters
@@ -595,8 +595,9 @@ class SaveLoad:
         """
         compress, subname = SaveLoad._adapt_by_suffix(fname)
 
-        restores = self._save_specials(fname, separately, sep_limit, ignore, pickle_protocol,
-                                       compress, subname)
+        restores = self._save_specials(
+            fname, separately, sep_limit, ignore, pickle_protocol, compress, subname,
+        )
         try:
             pickle(self, fname, protocol=pickle_protocol)
         finally:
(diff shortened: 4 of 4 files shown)


gojomo:
Simple enough, but per some ideas floated in #2848...
* could consider jumping to version 5 (& use `pickle5` backport when in Python <3.8)
* could declare Gensim's preferred version in a central named variable that's then reused - other than just a default-parameter literal in a variety of places
* there seem to be many refs to `pickle` or `import pickle` in other files - should all be checked to ensure using now preferred-version. (in particular: the utility `pickle()` function also declared in `utils.py` around line 1400 should also use now-preferred version.)
* if saving-separately is no longer oft-needed to work around pickle limitations, it could be default-off *unless* people need separate files for other reasons (like memory-mapping). this is a bigger change in expected-behavior, that would need a prominent release-note bullet, but I've seen enough people lose/break models when they *don't* realize they're spread among multiple files, to think that the proper *default* is one-file, and only experts who need/explicitly-specify separate numpy files should get them.

piskvorky:
Yes, perhaps it's time to revisit why we needed `save`/`load` at all. Why not just pickle?

1. To work around pickle size limitations. **No longer relevant**.
2. To allow memory-mapping large arrays, saving RAM. **Maybe resolved with protocol=5? I haven't really checked**.
3. To apply class-specific post-load operations (e.g. upgrade of model attributes on load). **Could be bypassed by having a dedicated upgrade script, as per https://github.com/RaRe-Technologies/gensim/issues/2967#issuecomment-710688660.
4. Persisting models to non-local storages (e.g. S3). **Can still be done by `pickle.dump(file-like object opened by smart_open)`**.

So, all reasons for `SaveLoad`'s existence may be gone now. Python has finally caught up.

We have to keep `SaveLoad` anyway for compatibility of course.

I'll look into protocol=5 and other places it's used, good point.

piskvorky:
Using an external library like [pickle5](https://github.com/pitrou/pickle5-backport) seems a hassle. Unless we actually take advantage of `protocol=5`, I'd prefer to keep the default at `protocol=4`.

gojomo:
My understanding is that pickle v5 has the hooks needed for the sort of custom-separate-serialization of things like numpy arrays, but it's not yet automatic/standard. So it's not a simple matter of "use it". But, if we thought using it soon in that manner was likely, doing a one-hop to v5 now, rather than hop-to-v4 now then hop-to-v5 soon after, might be a way to minimize interim states and catch any potential gotchas sooner rather than later. (AFAICT, 'pickle5' is an official backport by the same Python core contributor who wrote the v5 PEP and Python3.8+ implementation, so it should have fewer risks than other arbitrary external libs.) 

Even without 'upgrade-scripts', old-object-cleanup hooks that are analgous to `SaveLoad` 'specials'/etc seem possible via pickle - and if we relied on those instead, we might stay closer to other standard Python practices. If we think that's the eventual future direction, we needn't break anything that works and is familiar right away - but we would want to discourage further growth/complexity in `SaveLoad` - so something like the recent lifecycle logging could move to a different superclass/mix-in.

piskvorky:
Reasonable points. But this will need someone to code it up – at least as a proof of concept. It's not completely trivial work.

I fear I won't have the bandwidth for this – certainly not in time for 4.0.0. And frankly, if you do have some extra bandwidth for Gensim @gojomo, I'd much rather you revisited the [tickets for 4.0.0 Milestone](https://github.com/RaRe-Technologies/gensim/milestone/3).

mpenkov (approved):

PR #3071, opened 2021-03-10
Issue 3064, idf formula of BM25 corrected. 

### LINKED ISSUE #3064: Add +1 inside the math.log in the idf computation of bm25.
https://github.com/RaRe-Technologies/gensim/blob/e889fa3d45a406cabbc7e180fa9a8ee3f76ac6f0/gensim/summarization/bm25.py#L137

As it is right now, you can end up with negative idf values. It's counter intuitive. The right formula for idf computation (from [https://en.wikipedia.org/wiki/Okapi_BM25]) is: 
Screenshot from 2021-03-05 16-41-53

You just need to add +1 inside the log.

Also, you won't need the warning in `if self.average_idf < 0:` anymore.

jeromebedard12:
Fixes #3064.

The idf computation of BM25 model has been corrected (according to <https://en.wikipedia.org/wiki/Okapi_BM25>). With this modification, a word can't have a negative idf value. So no more counter intuitive scores for documents that have too many frequent words.

Since we can't have negative idf values anymore, there's no need for the epsilon parameter. Hence, it has been removed from BM25.

2 missing dependencies have also been added to `requirements_docs.txt`.

### DIFF
--- a/gensim/summarization/bm25.py
+++ b/gensim/summarization/bm25.py
@@ -32,7 +32,6 @@ Data:
 -----
 .. data:: PARAM_K1 - Free smoothing parameter for BM25.
 .. data:: PARAM_B - Free smoothing parameter for BM25.
-.. data:: EPSILON - Constant used for negative idf of document in corpus.
 
 """
 
@@ -47,7 +46,6 @@ from ..utils import effective_n_jobs
 
 PARAM_K1 = 1.5
 PARAM_B = 0.75
-EPSILON = 0.25
 
 logger = logging.getLogger(__name__)
 
@@ -69,7 +67,7 @@ class BM25(object):
         List of document lengths.
     """
 
-    def __init__(self, corpus, k1=PARAM_K1, b=PARAM_B, epsilon=EPSILON):
+    def __init__(self, corpus, k1=PARAM_K1, b=PARAM_B):
         """
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


mpenkov:
NB. If we remove epsilon parameter, things will break for existing code.

jeromebedard12:
I didn't see any broken code, all unit tests passed on my side.

mpenkov:
> I didn't see any broken code, all unit tests passed on my side.

It's possible that none of the unit tests are passing the epsilon keyword parameter to the get_bm25_weights function.

get_bm25_weights is a publicly accessible function, so it is possible that code like this exists _somewhere_ in someone's product:

```python
result = get_bm25_weights(corpus, n_jobs=-1, epsilon=epsilon)  # for some arbitrary value of epsilon
```

If we simply remove the epsilon parameter from the get_bm25_weights function, then the above code will crash with a TypeError (unexpected keyword argument). Our options are:

1. Warn people through the release notes that the function no longer has an epsilon parameter
2. Continue accepting the epsilon parameter with some benign default alue and warn people through warnings.warn / logging.warn if they pass a non-default value

@piskvorky I think either option is fine, but 2 is marginally better. WDYT?

piskvorky:
If the formula had been wrong all along (??? who authored this?), then option 1. is better IMO. Just Tear off the bandaid. And 4.0 is the perfect place for that.

If this is rather some gradual improvement, an algo that used to need X but no longer does, then 2. is more appropriate.

mpenkov:
@piskvorky Ah, this submodule got removed entirely in 4.0.0 (it belongs to the summarization subpackage).

@jeromebedard12 Thank you for your effort, but the functionality in this PR is no longer part of gensim (see [link]) for details. I'm sorry for not letting you know about this sooner, and hope you can continue to contribute to gensim in the future.

[1]: https://github.com/RaRe-Technologies/gensim/wiki/Migrating-from-Gensim-3.x-to-4#iv-removal-of-deprecations-and-unmaintained-modules

Issue #3067, opened 2021-03-09
Update WMD documentation


piskvorky:
A user reported a documentation issue on the mailing list: https://groups.google.com/g/gensim/c/8nobtm9tu-g.

The report shows two problems:

1) Something changed with `wmdistance` between 3.8 and 4.0 that is not properly captured in the [Migration notes](https://github.com/RaRe-Technologies/gensim/wiki/Migrating-from-Gensim-3.x-to-4).

2. The [WMD tutorial](https://radimrehurek.com/gensim_4.0.0/auto_examples/tutorials/run_wmd.html#normalizing-word2vec-vectors) contains instructions that are now outdated in 4.0: 

   > When using the wmdistance method, it is beneficial to normalize the word2vec vectors first, so they all have equal length. To do this, simply call model.init_sims(replace=True) and Gensim will take care of that for you.

   …And then our own tutorial logs a `WARNING : destructive init_sims(replace=True) deprecated & no longer required for space-efficiency`, which looks silly.

piskvorky:
@gojomo do you remember the motivation for forcing normalized vectors in WMD? I vaguely remember some discussion.

I checked the code and the way the normalization is done is not great. The key vectors are normalized over and over again, in a nested loop => bad performance.

gojomo:
There's discussion in #1094 - but there's no clear case for either way there. (Non-normalized seems closer to the original implementation; some secondhand testimony suggests norming might give better results.) It's hard to think of why you'd want to use *euclidean* distance on *non-normalized* vectors, as their magnitudes could then contribute a lot of "difference" that isn't correlated with the more typical cosine-distance-comparison. So I suspect default to unit-normed vectors is better-grounded & better-performing, but don't have an empirical case either way. 

The discussion as part of the `KeyedVectors` big-refactor was left kind of hanging as to what the proper default/configurability should be: https://github.com/RaRe-Technologies/gensim/pull/2698#discussion_r445812840

piskvorky:
Okay, thanks @gojomo. So basically we don't have a strong intuition / evidence one way or the other.

@mattkoehne since you're the user, what makes more sense to you, or gives better results?

I'm thinking we could add a `norm` parameter to `wmdistance` and pass it through = leave the decision with the user. 
Although that feels kinda weak… principled defaults are better.

gojomo:
[...]

PR #3073, opened 2021-03-14
[MRG] Make WMD normalization optional

piskvorky:
*Follow-up from #3067:*

This PR introduces an optional `KeyedVectors.wmdistance(norm=True/False)` parameter. Users can now choose whether to normalize their embedding vectors in each `wmdistance` call or not.

Prior behaviour:
* `gensim < 4.0`: no normalization; user had to normalize all vectors explicitly manually (destructive operation).
* `gensim == 4.0.0beta`: normalize with each call = `norm=True` hardwired.
* `gensim > 4.0.0beta` (this PR): default `norm=True`, optional `norm=False`.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -836,7 +836,7 @@ class KeyedVectors(utils.SaveLoad):
         """
         return self.most_similar(positive=[vector], topn=topn, restrict_vocab=restrict_vocab)
 
-    def wmdistance(self, document1, document2):
+    def wmdistance(self, document1, document2, norm=True):
         """Compute the Word Mover's Distance between two documents.
 
         When using this code, please consider citing the following papers:
@@ -854,6 +854,9 @@ class KeyedVectors(utils.SaveLoad):
             Input document.
         document2 : list of str
             Input document.
+        norm : boolean
+            Normalize all word vectors to unit length before computing the distance?
+            Defaults to True.
 
         Returns
         -------
@@ -873,7 +876,6 @@ class KeyedVectors(utils.SaveLoad):
             If `pyemd <https://pypi.org/project/pyemd/>`_  isn't installed.
 
         """
-
         # If pyemd C extension is available, import it.
         # If pyemd is attempted to be used, but isn't installed, ImportError will be raised in wmdistance
         from pyemd import emd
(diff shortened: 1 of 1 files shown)


gensim/models/keyedvectors.py
```diff
@@ -916,11 +915,11 @@ def wmdistance(self, document1, document2):
...
                     continue
 
-                # Compute Euclidean distance between unit-normed word vectors.
+                # Compute Euclidean distance between (potentially unit-normed) word vectors.
                 distance_matrix[i, j] = distance_matrix[j, i] = np.sqrt(
-                    np_sum((self.get_vector(t1, norm=True) - self.get_vector(t2, norm=True))**2))
+                    np_sum((self.get_vector(t1, norm=norm) - self.get_vector(t2, norm=norm))**2))
 
-        if np_sum(distance_matrix) == 0.0:
+        if abs(np_sum(distance_matrix)) < 1e-8:
```
gojomo, line 922: This looks like a fix or partial fix for a different issue. I'm not sure, but if all pairwise vocab distances are 0.0 (or sufficiently close?), isn't a return value of 0.0 more likely to match intuitions about what the distance between "essentially identical" texts should be?

And, if there is some actual range of tiny-but-not-technically-zero distances that either (a) confuse  EMD/PyEMD; or (b) justify clipping-the-result-to-zero, are we sure `1e-8` is the well-grounded threshold, based on some reasoning about the source of such noise? Or should the threshold be highlighted as something more tentative, or even configurable?

gojomo:
Other than my line-comment highlighting some possibilities around the zero-case threshold & default-return, the core default-but-optional norming looks good!

gensim/models/keyedvectors.py
gojomo: (Separate thought: if it is only true `0.0` that causes `emd` problems, maybe the better solution is to replace zeros with tiny noise nonzero values?)

piskvorky: This wasn't a fix for any particular issue. Just when I visited the code now, I saw a bunch of weird constructs there – this one of them (comparing a float for exact equality, bad smell).

I'm not aware of any related user report or bugs.

gojomo: Aha - my unconfirmed suspicion had been one of the other recent WMD problem reports might have been due to unexpected/excessive 0-distances, so changes here seemed potentially related. Still, seems odd that zeros should generate a loud warning and max (`inf`) distance, rather than just `0.0`. (It looks to me like this code might report `inf` as the WMD between `['foo', 'bar']` & `['bar', 'foo']` instead of `0.0`.) And, I'd be reluctant to change an absolute-equality test to a less-than-some-threshold test unless I knew for sure a test case where that was required.

piskvorky: Yeah, I'm not worried treating distance matrices whose total sum is <0.00000001 as 0.0.

If there's a bug with infinite WMD distances let's treat that separately. This change really is unrelated.

gojomo: But: you've changed the behavior for tiny-distances returning a tiny-WMD to instead returning `inf` - without any illustrative test cases. (That looks to me like it took an existing non-intuitive behavior/error in the code and expanded it to apply to **more** cases.) On a simple, "if it ain't broken don't fix it" analysis, why take that risk?

piskvorky: The previous behaviour was broken – comparing a sum of many floats against a fixed value cannot be right.

I agree the discontinuity around `inf` is weird. The algo, and its edge cases, deserve a thorough check. Can you investigate?

PR #3077, opened 2021-03-16
fix various documentation warnings

mpenkov:
I encountered the following while building under Sphinx 3.5.2 on MacOS:

docstring of gensim.corpora._mmreader.MmReader.num_docs:1:duplicate object description of gensim.corpora._mmreader.MmReader.num_docs, other instance in corpora/_mmreader, use :noindex: for one of them

gensim/gensim/utils.py:docstring of gensim.utils.chunkize:9:Inline interpreted text or phrase reference start-string without end-string.

gensim/docs/src/_matutils.rst:document isn't included in any toctree


piskvorky:
> I encountered the following while building under Sphinx 3.5.2 on MacOS:

https://github.com/RaRe-Technologies/gensim/blob/338ef330dea97c90c3180a9b570be9d0c9cef302/setup.py#L303-L306

mpenkov:
Yeah, I think that comment is out of date.

I'll try removing the pin, but it's possible it will cause a massive rebuild of everything, which isn't what we want.

PR #3084, opened 2021-03-20
Fix README badge link

piskvorky:
The README badge for "monthly Gensim downloads" links to a page that no longer works. Not sure what happened to it.

So instead, this PR changes that link to "Gensim downloads in general", which still works.

Also fixed some unrelated code style issues I spotted randomly.


piskvorky:
@mpenkov are the CI fails something I should worry about?

mpenkov:
Nah, those are wheel builds, they have been disabled for a while.

I was thinking about using them for the release, but there appears to be a bit more work to be done there, so I disabled them again.

PR #3103, opened 2021-04-05
Add early stop to on_epoch_end

gaffney2010:
This change allows the user to pass in a callback where on_epoch_end may return EARLY_STOP, which would cause the modeling to stop early.  This would preserve all existing code, because existing on_epoch_end functions return None.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -201,6 +201,8 @@ from gensim.models.keyedvectors import KeyedVectors, pseudorandom_weak_vector
 from gensim import utils, matutils
 
 
+EARLY_STOP = "EARLY_STOP_TOKEN"
+
 logger = logging.getLogger(__name__)
 
 try:
@@ -1073,7 +1075,8 @@ class Word2Vec(utils.SaveLoad):
             job_tally += job_tally_epoch
 
             for callback in callbacks:
-                callback.on_epoch_end(self)
+                if callback.on_epoch_end(self) == EARLY_STOP:
+                    break
 
         # Log overall time
         total_elapsed = default_timer() - start


mpenkov:
What's the use case?

gaffney2010:
Early stopping is a common approach to prevent overfitting.  I'm using the callback to compute recall@k on a test set with each epoch.  I want to stop the algorithm if recall@k converges - stops improving or worsens for a few consecutive epochs.

mpenkov:
@gojomo @piskvorky What do you think? Feels like something extremely hacky to do inside a callback.

gojomo:
My personal intuition is that for a shallow, SGD-optimized NN like in our *2Vec models, it's better theoretically-grounded to prevent overfitting with model shrinking than early-stopping. 

*BUT*, I've seen very-wise people working in deeper NNs who defend the practice there, so it seems OK to offer as an experimental option for those who might find a benefit. 

It might be a natural thing to add alongside cleaning up the buggy/inconsistent mess that is tour current loss-tallying/reporting, but if stopping according to project-specific external evaluations, which themselves would necessarily happen in a callback, offering it as a callback-controlled option makes sense, too. 

The use of the single-purpose flag-value string-token feels a bit fragile/odd to me. If the return value is special & this is the only purpose, it could just be a `True`/`False` for whether to early-stop. Some thought should be given as to whether a callback in a multi-callback situation should be able to cancel later callbacks, or just, when requesting a stop, stop further epochs. 

Alternatively, maybe the count-of-remaining-epochs (or even something else that may be interesting to offer user-control over, like next epoch's `alpha`) is instead controlled by some in-model hot state, and any callback that wants to change the next-epoch's behavior just mutates that state during the callback. That'd even open the possibility of callback-evaluations *extending* training epochs, or accelerating/decelerating/reversing learning-rate decay, when trends in either internal training loss or external evaluations suggest that might be helpful. 

So: I'm not averse to offering this functionality, but don't love this single-purpose special-casing, and would prefer to see some related issues (especially around loss-tracking) considered together to maybe clean up some longstanding sore points while also addressing this need.

gaffney2010:
These are interesting points.

I'm not sure agree about the True/False (I think the EARLY_STOP token is self-documenting and harder to misuse).  In a multi-callback situation, we'd probably want to run all the callbacks - I hadn't thought of that.

Your idea of an in-model hot state is a good one.  That kind of exists already in self.epochs.  I wonder if the better solution here is to change
```
for cur_epoch in range(self.epochs):
```

to something like:
```
cur_epoch_ct = 0
while True:
  if cur_epoch_ct >= self.epochs:
    # This pattern allows self.epochs to change in a callback
    continue
  ...
  cur_epoch_ct += 1
```

This would give control to the user writing the callbacks to make this change without cluttering the code for people who don't want to use it.

gojomo:
Yes, something like `cur_epoch_ct` that coudl be mutated by (one or more) callbacks is what I was thinking about. It cold potentially be wrapped up into a distinct well-named data-object describing both the reportable & mutable state of the *current* training-call-in-progress. (That is, it might also include current- or next-alpha, running-training-loss, etc.) Perhaps, a small dataclass? (Supported in Py3.7+, with an official backport for Py3.6.)

piskvorky:
@gaffney2010 are you able to implement @gojomo's suggestions above?

mpenkov:
@piskvorky Guess that's a "no" then...

piskvorky:
Looks like it :(

PR #3116, opened 2021-04-14
Fix bug where saved Phrases model did not load its connector_words

aloknayak29:
Bug description: If we have saved the trained phrases model in this version i.e gensim version 4.0.0 or 4.0.1, while giving non empty connector_words. And If we try to load the saved model in this version itself. Then the connector_words of the loaded phrases model will be wrongly assigned empty frozen set. This is a functional bug and will result in some of the words not getting grouped in ngram

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes
 
 - LsiModel: Only log top words that actually exist in the dictionary (PR [#3091](https://github.com/RaRe-Technologies/gensim/pull/3091), [@kmurphy4](https://github.com/kmurphy4))
 - [#3115](https://github.com/RaRe-Technologies/gensim/pull/3115): Make LSI dispatcher CLI param for number of jobs optional, by [@robguinness](https://github.com/robguinness))
+- fix bug when loading saved Phrases model (PR [#3116](https://github.com/RaRe-Technologies/gensim/pull/3116), [@aloknayak29](https://github.com/aloknayak29))
 
 ### Documentation
 
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -396,8 +396,7 @@ s        """
                 model.connector_words = model.common_terms
                 del model.common_terms
             else:
-                logger.warning('older version of %s loaded without common_terms attribute, '
-                               'setting connector_words to an empty set', cls.__name__)
+                logger.warning('loaded older version of %s, setting connector_words to an empty set', cls.__name__)
                 model.connector_words = frozenset()
 
         if not hasattr(model, 'corpus_word_count'):
--- a/gensim/models/phrases.py
+++ b/gensim/models/phrases.py
@@ -391,15 +391,14 @@ s        """
                     raise ValueError(f'failed to load {cls.__name__} model, unknown scoring "{model.scoring}"')
 
         # common_terms didn't exist pre-3.?, and was renamed to connector in 4.0.0.
-        if hasattr(model, "common_terms"):
-            model.connector_words = model.common_terms
-            del model.common_terms
-        else:
-            logger.warning(
-                'older version of %s loaded without common_terms attribute, setting connector_words to an empty set',
-                cls.__name__,
-            )
-            model.connector_words = frozenset()
+        if not hasattr(model, "connector_words"):
+            if hasattr(model, "common_terms"):
+                model.connector_words = model.common_terms
+                del model.common_terms
+            else:
+                logger.warning('older version of %s loaded without common_terms attribute, '
+                               'setting connector_words to an empty set', cls.__name__)
+                model.connector_words = frozenset()
 
         if not hasattr(model, 'corpus_word_count'):
             logger.warning('older version of %s loaded without corpus_word_count', cls.__name__)
--- a/gensim/test/test_phrases.py
+++ b/gensim/test/test_phrases.py
@@ -305,41 +305,42 @@ class TestPhrasesModel(PhrasesCommon, unittest.TestCase):
 
 
 class TestPhrasesPersistence(PhrasesData, unittest.TestCase):
-
     def test_save_load_custom_scorer(self):
         """Test saving and loading a Phrases object with a custom scorer."""
+        bigram = Phrases(self.sentences, min_count=1, threshold=.001, scoring=dumb_scorer)
         with temporary_file("test.pkl") as fpath:
-            bigram = Phrases(self.sentences, min_count=1, threshold=.001, scoring=dumb_scorer)
             bigram.save(fpath)
             bigram_loaded = Phrases.load(fpath)
-            test_sentences = [['graph', 'minors', 'survey', 'human', 'interface', 'system']]
-            seen_scores = list(bigram_loaded.find_phrases(test_sentences).values())
 
-            assert all(score == 1 for score in seen_scores)
-            assert len(seen_scores) == 3  # 'graph minors' and 'survey human' and 'interface system'
+        test_sentences = [['graph', 'minors', 'survey', 'human', 'interface', 'system']]
+        seen_scores = list(bigram_loaded.find_phrases(test_sentences).values())
+
+        assert all(score == 1 for score in seen_scores)
+        assert len(seen_scores) == 3  # 'graph minors' and 'survey human' and 'interface system'
 
     def test_save_load(self):
         """Test saving and loading a Phrases object."""
+        bigram = Phrases(self.sentences, min_count=1, threshold=1)
         with temporary_file("test.pkl") as fpath:
-            bigram = Phrases(self.sentences, min_count=1, threshold=1)
             bigram.save(fpath)
             bigram_loaded = Phrases.load(fpath)
-            test_sentences = [['graph', 'minors', 'survey', 'human', 'interface', 'system']]
-            seen_scores = set(round(score, 3) for score in bigram_loaded.find_phrases(test_sentences).values())
 
-            assert seen_scores == set([
-                5.167,  # score for graph minors
-                3.444  # score for human interface
-            ])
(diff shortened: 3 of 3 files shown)


gensim/models/phrases.py
```diff
@@ -391,15 +391,14 @@ def load(cls, *args, **kwargs):
...
-                cls.__name__,
-            )
-            model.connector_words = frozenset()
+        if not hasattr(model, "connector_words"):
+            if hasattr(model, "common_terms"):
+                model.connector_words = model.common_terms
+                del model.common_terms
+            else:
+                logger.warning('older version of %s loaded without common_terms attribute', cls.__name__)
+                logger.warning('setting connector_words to an empty set')
```
piskvorky, line 400: Please keep the warning as one line.

piskvorky (changes requested):
Sorry I didn't understand the motivation for this PR. What is the problem, why is this needed?

gensim/models/phrases.py
aloknayak29: One line warning was making the line 121 characters long and a flake test was failing, so I had to keep it in into two lines warning like all other already existing warnings e.g lines 404 and 405

aloknayak29:
A reproducible example will include training data as well. Although an short example of bug will be, If you train a bigram and trigram model, and apply them on words like ["united, "states", "of", "america"], then this will output ['united_states_of_america']

`trigram_phrases_model[bigram_phrases_model[['united', 'states', 'of', 'america']]]`
`['united_states_of_america']`
 
But if you these models and load them and then apply them on words like ["united, "states", "of", "america"], then this will output

`['united_states', 'of', 'america']`

piskvorky:
Please add a unit test that fails before this PR, and succeeds after. That should demonstrate the issue clearly.

aloknayak29:
Do you need the unit test changes in the same commit or different commit over this?

gensim/test/test_phrases.py
```diff
@@ -378,9 +379,11 @@ def test_save_load_custom_scorer(self):
     def test_save_load(self):
         """Test saving and loading a FrozenPhrases object."""
         with temporary_file("test.pkl") as fpath:
-            bigram = FrozenPhrases(Phrases(self.sentences, min_count=1, threshold=1))
+            bigram = FrozenPhrases(Phrases(self.sentences, min_count=1, threshold=1,
```
piskvorky, line 382: [Hanging indent](https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#code-style) please.

piskvorky (changes requested):
Thanks a lot! One minor code style fix and we're good to go.

gensim/test/test_phrases.py
aloknayak29: code style fix updated in the commit

piskvorky (approved):

gensim/test/test_phrases.py
```diff
@@ -321,9 +321,10 @@ def test_save_load_custom_scorer(self):
     def test_save_load(self):
         """Test saving and loading a Phrases object."""
         with temporary_file("test.pkl") as fpath:
-            bigram = Phrases(self.sentences, min_count=1, threshold=1)
+            bigram = Phrases(self.sentences, min_count=1, threshold=1, connector_words=self.connector_words)
```
mpenkov, line 324: Why are we changing the behavior of an existing, passing test?

Should we not be adding a new test case here?

aloknturing: I wanted to create a separate test, But after reading this whole file gensim/test/test_phrases.py, I found that this test function i.e "test_save_load" should ideally be capturing this bug. Give its name "test_save_load" I assumed that this test was designed to capture the saving & loading of the whole model and this test shouldn't pass if saved connector words are not loaded. Adding it separately will increase the number of lines of code and will increase the test execution time. Another thought can be to remove this test altogether i.e not testing loading & saving of connector words. Let me know I can change the PR.

aloknayak29: I wanted to create a separate test, But after reading this whole file gensim/test/test_phrases.py, I found that this test function i.e "test_save_load" should ideally be capturing this bug. Give its name "test_save_load" I assumed that this test was designed to capture the saving & loading of the whole model and this test shouldn't pass if saved connector words are not loaded. Adding it separately will increase the number of lines of code and will increase the test execution time. Another thought can be to remove this test altogether i.e not testing loading & saving of connector words. Let me know I can change the PR.

mpenkov: > Adding it separately will increase the number of lines of code and will increase the test execution time.

The test is less than 10 lines long, so duplicating them isn't a huge issue. That leaves execution time.

How long does the test take to run?

> Another thought can be to remove this test altogether i.e not testing loading & saving of connector words.

What would be the benefit of removing this test? Isn't it testing a valid case?

gensim/models/phrases.py
mpenkov: I think what @piskvorky means here is that you should output a single _warning_ only. You're outputting two warnings here, where one would do (the warning can occupy multiple lines in code, as in the original code).

gensim/test/test_phrases.py
aloknturing: I couldn't understand your first point. After reading your first point, It seems that you have misunderstood that this bug is loading an old model. This bug occurs when we save and load the model in the same version itself. I mean to say that "test_save_load_old_model" would not describe this bug related test. string like "test_save_load_with_connector_words" can describe this bug
I agree with all of the rest of your points. 
It will take the same time as test  "test_save_load"
Yes its testing a valid case

aloknayak29: I couldn't understand your first point. After reading your first point, It seems that you have misunderstood that this bug is loading an old model. This bug occurs when we save and load the model in the same version itself. I mean to say that "test_save_load_old_model" would not describe this bug related test. string like "test_save_load_with_connector_words" can describe this bug
I agree with all of the rest of your points.
It will take the same time as test "test_save_load"
Yes its testing a valid case

mpenkov: > string like "test_save_load_with_connector_words" can describe this bug

Sure, if that's a more suitable name, please use that instead.

gensim/models/phrases.py
piskvorky: Yes, that's what I meant, thanks. If other parts of the code do it badly (split a single warning message into multiple isolated warnings), feel free to fix them too :)

gensim/test/test_phrases.py
aloknayak29: Updated PR with new test method and no behaviour changing of an existing, passing test

gensim/models/phrases.py
mpenkov: @aloknayak29 Please have a look at this.

gensim/test/test_phrases.py
```diff
@@ -385,6 +394,15 @@ def test_save_load(self):
                 bigram_loaded[['graph', 'minors', 'survey', 'human', 'interface', 'system']],
                 ['graph_minors', 'survey', 'human_interface', 'system'])
 
+    def test_save_load_with_connector_words(self):
```
mpenkov, line 397: How is this different to the test on line 335?

gensim/models/phrases.py
aloknayak29: Yes, converted it to a single line warning

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Changes
 
 - LsiModel: Only log top words that actually exist in the dictionary (PR [#3091](https://github.com/RaRe-Technologies/gensim/pull/3091), [@kmurphy4](https://github.com/kmurphy4))
 - [#3115](https://github.com/RaRe-Technologies/gensim/pull/3115): Make LSI dispatcher CLI param for number of jobs optional, by [@robguinness](https://github.com/robguinness))
+- fix bug when loading saved Phrases model (PR [#3116](https://github.com/RaRe-Technologies/gensim/pull/3116), [@aloknayak29](https://github.com/aloknayak29))
```
piskvorky, line 8: Don't we have a script that spits out a formatted list of all PRs and tickets automatically, during a release?

mpenkov: We do, but I've been keeping track of them as we merge them in smart_open, and I've found that way to be more manageable (less risk of missing PRs), so I'm trying out the same approach with gensim.

piskvorky: OK, no problem. One advantage of the script is consistency in formatting – see for example the change here :)

mpenkov: Yeah, I've got a script that handles these changes, but multiple versions of it were floating around my work/home machines. It's now committed to the repo, so once we nail down the format of the entries, the formatting should no longer be an issue.

mpenkov: https://github.com/RaRe-Technologies/gensim/blob/develop/release/annotate_pr.py

piskvorky: I was talking about this script: https://github.com/RaRe-Technologies/gensim/blob/develop/release/generate_changelog.py

mpenkov: Yes. Once the two scripts output stuff in the same format, the consistency problem will be solved, right?

piskvorky: I may be misunderstanding, but why do we need two scripts? `generate_changelog.py` seems to be doing what we need = generate formatted CHANGELOG. When / why would we use `annotate_pr.py`?

mpenkov: I'm using annotate_pr.py for each PR (since a few week ago). I'll use generate_changelog.py for the next release. I'll keep whichever ends up being the most convenient - for now, it seems to be the former, but time will tell.

mpenkov:
Thank you @aloknayak29 !

PR #3117, opened 2021-04-14
Ensure next_index available when loading old stored KeyedVectors models

### LINKED ISSUE #3114: AttributeError: 'KeyedVectors' object has no attribute 'next_index'
#### Problem description

I try to add new vectors to a given model.

Relevant traceback:
```
  File "d:\projekte\foo\source\backend\venv\lib\site-packages\gensim\models\keyedvectors.py", line 458, in add_vector
    target_index = self.next_index
AttributeError: 'KeyedVectors' object has no attribute 'next_index'
```

My code:
```python
class Foo:
    def __init__(self) -> None:
        self._model: KeyedVectors = gensim.models.KeyedVectors.load('C:\foo.bar')

   def is_word_known(word: str, language: str) -> bool:
       pass

    def add_word(self, word: str, synonym: str, language: str) -> None:
        if not self.is_word_known(word=synonym, language=language):
            raise ValueError(f'Word "{word}" cannot be added since synonym "{synonym}" is not known.')

        if self.is_word_known(word=word, language=language):
            raise ValueError(''Word "{word}" cannot be added since it is already known by the model.')

        vector = self._model.get_vector(key=word)
        self._model.add_vector(key=word, vector=vector)
```

#### Versions

```
Windows-7-6.1.7601-SP1
Bits 64
NumPy 1.20.2
SciPy 1.6.2
gensim 4.0.1
FAST_VERSION 0
```

gojomo:
Fix #3114.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -258,7 +258,7 @@ class KeyedVectors(utils.SaveLoad):
         # fixup rename of vocab into map
         if 'key_to_index' not in self.__dict__:
             self._upconvert_old_vocab()
-        # ensure older instances have next_index 
+        # ensure older instances have next_index
         if not hasattr(self, 'next_index'):
             self.next_index = len(self)
 
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -258,6 +258,9 @@ class KeyedVectors(utils.SaveLoad):
         # fixup rename of vocab into map
         if 'key_to_index' not in self.__dict__:
             self._upconvert_old_vocab()
+        # ensure older instances have next_index 
+        if not hasattr(self, 'next_index'):
+            self.next_index = len(self)
 
     def _upconvert_old_vocab(self):
         """Convert a loaded, pre-gensim-4.0.0 version instance that had a 'vocab' dict of data objects."""


mpenkov:
Do we want to add some unit tests here, to validate the solution and prevent regressions?

gojomo:
> Do we want to add some unit tests here, to validate the solution and prevent regressions?

Yes, per my comment on #3114: https://github.com/RaRe-Technologies/gensim/issues/3114#issuecomment-821862153 - but not sure when I'd be able to get to it, so just dropped this 1st quick stab at a maybe-fix here, as a building block for others.

piskvorky:
Looks like the first *and* the last stab :)

Do we merge as-is, or keep waiting for someone to continue? Or close. @gojomo @mpenkov

gojomo:
I think this is sufficiently straightforward & low-risk it can be added without a unit test. We still have giant gaps in rigorous/complete testing of backward-compatibility goals – a well-labeled, growing library of older models whose loadability into newer versions is automatically checked has been discussed, & (once such processes 1st established) wouldn't be too hard to maintain, & would have caught this (& other) issues earlier. (See related ideas in #2967.) But getting that started (& ideally then using the improved rigor to clean out all the junk test files no longer relevant) seems a bigger task than anyone is currently able to prioritize.

PR #3128, opened 2021-04-28
Materialize and copy the corpus passed to SoftCosineSimilarity

Witiko:
SoftCosineSimilarity currently does not perform any indexing and keeps a reference to the corpus passed to the constructor. This [can cause confusion](https://groups.google.com/g/gensim/c/3ksqzgCTWxI) if a corpus is lazy and requests for documents are only resolved when reading the corpus. In general, subclasses of the SimilarityABC interface are expected to materialize and copy (i.e. index) the corpus rather than just reference it.

Some classes, such as WmdSimilarity, do not conform to this expectation. However, these classes also don't work with traditional BoW corpora and should be considered novelties/outliers. SoftCosineSimilarity, on the other hand, is very similar to the core SparseMatrixSimilarity class, and should satisfy the same invariants even if they are generally unspoken.

This pull request makes SoftCosineSimilarity materialize and copy the corpus passed to the constructor using the `list()` built-in. This is a minimal change with no change of interface and little chance of regressions. A more thorough solution would be to index the corpus to a sparse CSC matrix. However, this would likely [break the interface](https://groups.google.com/g/gensim/c/3ksqzgCTWxI/m/QZZEv0r5AgAJ), which may be undesirable so soon after 4.0.0.

### DIFF
--- a/gensim/similarities/docsim.py
+++ b/gensim/similarities/docsim.py
@@ -937,7 +937,7 @@ class SoftCosineSimilarity(interfaces.SimilarityABC):
         """
         self.similarity_matrix = similarity_matrix
 
-        self.corpus = corpus
+        self.corpus = list(corpus)
         self.num_best = num_best
         self.chunksize = chunksize
         self.normalized = normalized
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changes
 ## Unreleased
 
 - LsiModel: Only log top words that actually exist in the dictionary (PR [#3091](https://github.com/RaRe-Technologies/gensim/pull/3091), [@kmurphy4](https://github.com/kmurphy4))
+- Materialize and copy the corpus passed to SoftCosineSimilarity (PR [#3128](https://github.com/RaRe-Technologies/gensim/pull/3128), [@Witiko](https://github.com/Witiko))
 - [#3115](https://github.com/RaRe-Technologies/gensim/pull/3115): Make LSI dispatcher CLI param for number of jobs optional, by [@robguinness](https://github.com/robguinness))
 
 ### Documentation


gensim/similarities/docsim.py
```diff
@@ -937,7 +937,7 @@ def __init__(self, corpus, similarity_matrix, num_best=None, chunksize=256, norm
         """
         self.similarity_matrix = similarity_matrix
 
-        self.corpus = corpus
+        self.corpus = list(corpus)
```
piskvorky, line 940: This needs a change in the documentation too IMO – a big warning that this is happening, the implications.

piskvorky (changes requested):

gensim/similarities/docsim.py
Witiko: Isn't this arguably the expected behavior? The documentation of SoftCosineSimilarity does not make it clear what happens to the constructur, but isn't it assumed that the corpus is read, indexed, and stored inside the SimilarityABC class?

Witiko: If this is not the expected behavior, then there is no grounds for this PR in the first place.

piskvorky: On second thoughts, you're right.
It is a reasonable expectation that the indexing process will a) traverse the corpus, and b) grow in in RAM with corpus size.

piskvorky (approved):

mpenkov (commented):
Looks good to me. Thank you @Witiko !

sug4ndh:
Hi, 

I have an update to a problem that I reported here https://groups.google.com/g/gensim/c/3ksqzgCTWxI. Not sure if this affects this PR merge. I noticed that the problem reported in the google group thread does not occur if I generate corpus and dictionary afresh before I train word2vec or fasttext models and then subsequently generate SoftCosineSimilarity index. However, if used a saved corpus and dictionary to train fasttext or word2vec model and then generate SoftCosineSimilarity index, I always get the problem that "No such file or directory: 'corpus.mm' if corpus.mm is not present in dir from where I am executing the program. If I copy corpus.mm to the dir from where I am executing the program then no error appears.

Witiko:
@sug4ndh Hello, this should solve you problem. As a workaround, you can create SoftCosineSimilarity index with list(corpus) instead of corpus. Then, you should be able to also save and subsequently reload your SoftCosineSimilarity index. Cheers!

PR #3133, opened 2021-05-05
Update link to Hoffman paper (online VB LDA)

jonaschn:
The previous link was Matthew Hoffman's Google Scholar profile or not the official one.
Use full author names in the first occurrence and first author only afterward.
Furthermore, use latex symbols for original parameter names from Hoffman et al.

### DIFF
--- a/gensim/models/ldamodel.py
+++ b/gensim/models/ldamodel.py
@@ -14,9 +14,12 @@ for online training.
 
 The core estimation code is based on the `onlineldavb.py script
 <https://github.com/blei-lab/onlineldavb/blob/master/onlineldavb.py>`_, by
-`Matthew D. Hoffman, David M. Blei, Francis Bach:
-Online Learning for Latent Dirichlet Allocation, NIPS 2010
-<https://papers.neurips.cc/paper/2010/file/71f6278d140af599e06ad9bf1ba03cb0-Paper.pdf>`_.
+Matthew D. Hoffman, David M. Blei, Francis Bach:
+`'Online Learning for Latent Dirichlet Allocation', NIPS 2010`_.
+
+.. _'Online Learning for Latent Dirichlet Allocation', NIPS 2010: online-lda_
+.. _'Online Learning for LDA' by Hoffman et al.: online-lda_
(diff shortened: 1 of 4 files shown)


mpenkov (changes requested):
Thank you @jonaschn for picking up on this.

Is it possible to include the hyperlink only once, e.g. in the module docstring, and then reference it from within the function docstrings? I think RST allows something like this.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks

I'd rather avoid having the same long hyperlink (or hyperlinks, if we're showing more than one) appear in multiple places.

gensim/models/ldamodel.py
```diff
@@ -14,9 +14,12 @@
...
 The core estimation code is based on the `onlineldavb.py script
 <https://github.com/blei-lab/onlineldavb/blob/master/onlineldavb.py>`_, by
-`Matthew D. Hoffman, David M. Blei, Francis Bach:
-Online Learning for Latent Dirichlet Allocation, NIPS 2010
-<https://papers.neurips.cc/paper/2010/file/71f6278d140af599e06ad9bf1ba03cb0-Paper.pdf>`_.
+Matthew D. Hoffman, David M. Blei, Francis Bach:
+`'Online Learning for Latent Dirichlet Allocation', NIPS 2010`_.
+
+.. _'Online Learning for Latent Dirichlet Allocation', NIPS 2010: online-lda_
+.. _'Online Learning for LDA' by Hoffman et al.: online-lda_
```
mpenkov, line 21: I think defining it one file and then referencing it from other files should work. Defining the same label in multiple files is likely to raise and error or even a warning when building the docs.

Have you tried building these with Sphinx?

jonaschn: I tried defining the link reference in ldamodel.py only but this does not work.
I build the docs using `make -C docs/src html` and it works without errors or warnings.

mpenkov:
Thank you for your contribution @jonaschn !

mpenkov (approved):

PR #3143, opened 2021-05-15
replace _mul function with explicit casts

mpenkov:
Follow up from #3136

### DIFF
--- a/gensim/models/word2vec_inner.pyx
+++ b/gensim/models/word2vec_inner.pyx
@@ -71,10 +71,6 @@ cdef void our_saxpy_noblas(const int *N, const float *alpha, const float *X, con
     for i from 0 <= i < N[0] by 1:
         Y[i * (incY[0])] = (alpha[0]) * X[i * (incX[0])] + Y[i * (incY[0])]
 
-cdef long long _mul(const np.uint32_t a, const int b) nogil: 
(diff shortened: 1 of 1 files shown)


piskvorky:
@mpenkov can you remove the (empty) file [16](https://github.com/RaRe-Technologies/gensim/blob/develop/16)? Also added in #3136, probably by mistake.

gojomo:
I personally find this inline, direct approach clearer - so, I prefer it to the `_mul()` function approach. 

But also, I'd suggested previously a `cdef`/`ctypedef` might possibly make it even more compact.  Thinking more about it, it strikes me that best of all could be to explicitly reuse an existing standard numpy type. For example, replacing `<long long>` (which may still have risks of varying meaning across platforms) with `<np.uint64_t>` might best both communicate intent & also ensure uniform behavior across our supported platforms. (If it breaks, it'll break the same everywhere these same `numpy` arrays are being used, which is preferable.)

gojomo:
Would the more-explicit/more-portable `<np.uint64_t>` not have worked?

piskvorky:
I like the explicit width version too.

Writing the comment seems comparable effort to writing the commit – could you help @mpenkov out here? Thanks.

gojomo:
The change would best be done by someone set up to verify it fixes the test case, which I won't be set up to do anytime soon.

Issue #3137, opened 2021-05-10
Minimum required numpy version needs to be bumped to 1.17.0


shakhat:
#### Problem description

gensim is declared to work with numpy >= 1.11.3, however it doesn't work with any version below 1.17.0.

#### Steps/code/corpus to reproduce

Install gensim and non-recent version of numpy:

```bash
$ pip install "numpy<1.17.0" gensim
```

Run python interpreter with the following code
```python
>>> import gensim
```

Observed error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/private/tmp/.venv/lib/python3.7/site-packages/gensim/__init__.py", line 11, in <module>
    from gensim import parsing, corpora, matutils, interfaces, models, similarities, utils  # noqa:F401
  File "/private/tmp/.venv/lib/python3.7/site-packages/gensim/parsing/__init__.py", line 4, in <module>
    from .preprocessing import (remove_stopwords, strip_punctuation, strip_punctuation2,  # noqa:F401
  File "/private/tmp/.venv/lib/python3.7/site-packages/gensim/parsing/preprocessing.py", line 26, in <module>
    from gensim import utils
  File "/private/tmp/.venv/lib/python3.7/site-packages/gensim/utils.py", line 62, in <module>
    default_prng = np.random.default_rng()
AttributeError: module 'numpy.random' has no attribute 'default_rng'
```

The solution is to bump the minimum version of numpy to be at least 1.17.0: https://github.com/RaRe-Technologies/gensim/blob/develop/setup.py#L318

#### Versions

```
>>> import numpy; print("NumPy", numpy.__version__)
NumPy 1.16.6
>>> import scipy; print("SciPy", scipy.__version__)
SciPy 1.6.3
```

mpenkov:
I think this is the discussion here: https://github.com/RaRe-Technologies/gensim/issues/3085

We did it for gensim-wheels, but not for gensim proper. It's a slightly different problem. I think we can bump the minimum version to 1.17.0 as mentioned in this ticket.

piskvorky:
Fixed in #3156.

PR #3156, opened 2021-05-24
Update Numpy minimum version to 1.17.0

PrimozGodec:
Based on the discussion in https://github.com/RaRe-Technologies/gensim/issues/3137 I am updating the minimum NumPy version to 1.17.0. It is required since otherwise users who install Gensim in the environment with an older NumPy version get `AttributeError: module 'numpy.random' has no attribute 'default_rng'` error.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -315,7 +315,7 @@ docs_testenv = core_testenv + distributed_env + visdom_req + [
     'pandas',
 ]
 
-NUMPY_STR = 'numpy >= 1.11.3'
+NUMPY_STR = 'numpy >= 1.17.0'
 #
 # We pin the Cython version for reproducibility.  We expect our extensions
 # to build with any sane version of Cython, so we should update this pin


piskvorky (approved):
Thanks!

Numpy 1.11.3 is almost five years old; 1.17.0 is two years old. I guess we can do this – recent versions are more relevant because more people update to cutting edge in the ML domain. WDYT @mpenkov @gojomo ?

mpenkov (approved):
No objections from me.

gojomo:
[NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) implies the Numpy project has already (as of January 2021) dropped support for every version earlier `numpy-1.17`.

 It'd be logical to never do a formal release that specifies an expired-support version of Numpy - so I'd be fine bumping to 1.17 ASAP.

We may even want to go further, and always ensure our releases specify a Numpy version that has at least N months of support remaining, where N could be 3, 6, or even 12.

piskvorky:
Okay, merged. Congrats on your first Gensim contribution @PrimozGodec ⭐ 

@gojomo is there a way to automate that? If the process relies on us (or users) submitting version edits to `setup.py` (like this PR), we're bound to forget.

gojomo:
I don't know of a standard way to integrate a project's "oldest supported" or "oldest supported as of a certain date" version - it seems those expectations are still somewhat informally specified, as through that 'NEP', rather than rigorously specified. 

I could imagine some new convention where a project, at a well-known & relatively-secure URL under its management, might maintain text fragments specifying key transitions of support. For example, perhaps `https://numpy.org/earliest-supported.ver` might have just the string, `numpy=1.17.5  # released 2021-01-01; end-support 2021-07-25`. Similarly, paths `/latest-supported.ver` & other variants might include useful strings, allowing downstream projects like us (or projects downstream of us) to auto-fetch recommended versions of things before releases or testing. (Maybe there is something like tis or some other level of labeling on PyPI that might provides such hints - but I don't recall coming across it.)

Barring something that reliable, adding to a checklist for new releases a quick review of *major* prerequisites (like `numpy`/`scipy`) for changed supported-status.

piskvorky:
Thanks, makes sense. Could you add that step to the release checklist please (with a link to this ticket, for context)?

gojomo:
Where is a release checklist, if any? (A search of the repo reveals no files ith 'checklist' in them.)

mpenkov:
https://github.com/RaRe-Technologies/gensim/wiki/Developer-page

gojomo:
I've added an item with the gist-of-the-check at: https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#prepare - review & adjust as desired.

Since our earlier discussion, I came across scipy's `oldest-supported-numpy` meta-package: https://github.com/scipy/oldest-supported-numpy

It seems more focused on being a build-prereq for wheels, not an end-user-install-time prereq - so I'm unsure whether it changes our concerns here.... but it's interesting to know about, as a formalized way to express upstream-dependency lifecycle states.

mpenkov:
Yeah, we use oldest-supported-numpy for building gensim wheels for releases, but as you've stated, they solve a slightly different problem than the one mentioned by the OP.

Issue #2283, opened 2018-12-03
SECURITY: api.load() recklessly downloads & runs arbitrary python code


gojomo:
The `api.load()` utility will grab fresh Python code, in the form  of an `__init__.py` file inside the Github project `gensim-data` download area, and then run it at the user's machine. 

This form of dynamic code loading & execution:

* violates user expectations: requesting a dataset should not run arbitrary new code that was never explicitly installed on the user's machine. Further, there's no indication in the `api.load()` docs that this could occur.
* creates a severe security risk: if a bad-faith actor obtains the ability to edit the `gensim-data` Github "releases" files, they can cause arbitrary new code to run on gensim users' machine, when those users use `api.load()`. The users wold think, "I'm still using this old, assumed-safe-through-widespread-use gensim-X.Y.Z version" – but they'd be running arbitrary all-new code from over the network. It's hard to tell who has `gensim-data` project rights. It's also not clear that anyone would quickly notice edits/changes there.

Further, these `__init__.py` files in the `gensim-data` releases aren't even in version-control – instead, they're uploaded as 'assets' via the Github releases interface. (There's no file-size need to do this; the `__init__.py` I reviewed, for `wiki-english-20171001` at <https://github.com/RaRe-Technologies/gensim-data/releases/download/wiki-english-20171001/__init__.py>, is just a tiny shim and I imagine most other such files are, as well. It's code, it should be versioned.)

That they are not in version-control makes them hard to review through normal means (such as browsing the Github website), and raises the possibility they could be changed to something malicious, and then back again, without anyone noticing or it being visible in any persistent logs. 

Recommendations: 

The `api.load()` mechanism should be immediately redesigned to not load any new code over the network – and the developer guidelines for gensim & associated projects should make it clear such dynamic loading of code outside normal package-installation processes (like `pip`) is unacceptable. 

If supporting new datasets requires dataset-specific code, that code should go through normal collaboration/version-control/release procedures, waiting for a new `pip`-installable `gensim` (or other new supporting project) explicit release before running on users' computers. 

ATTN: @menshikh-iv @piskvorky @chaitaliSaini

gojomo:
> * If a bad actor gains access to the repo, we're screwed anyway (python packages always execute arbitrary code)

[...]

PR #3157, opened 2021-05-25
New KeyedVectors.vectors_for_all method for vectorizing all words in a dictionary

Witiko:
When searching for similar word embeddings using the `KeyedVectors.most_similar()` method, we often have a dictionary that limits the number of words that we would like to consider and, for subword models such as FastText that enable word vector inference, also expands the number of words that we would like to consider:

119530938-c6497380-bd83-11eb-8129-9d56a4270dad

This is also true in document similarity measures that use word embeddings as a source of word similarity, such as the Soft Cosine Measure. In the Soft Cosine Measure, the first step is the construction of a word similarity matrix. The word similarity matrix models a dictionary that will often be different from the vocabulary of the word embeddings. The word similarity matrix is sparse and uses the `topn` parameter of the `KeyedVectors.most_similar()` method to control how many closest words for each word will be considered. However, if the overlap between our dictionary and the vocabulary of the word embeddings is small, the `KeyedVectors.most_similar()` method will consistently return fewer than `topn` closest words from the dictionary and the matrix will be much more sparse than it would have otherwise been. This leads to a lack of control and possibly weaker models.

## Proposed solution
The solution @gojomo and I discussed in https://github.com/RaRe-Technologies/gensim/pull/3146#issuecomment-844000455 is to have a `KeyedVectors.vectors_for_all(words)` method that would take an iterable of words (a dictionary) and produce a new `KeyedVectors` object that would contain vectors only for the requested words. In subword models such as FastText, vectors for words outside the vocabulary would be inferred. This would guarantee that all `topn` words retreived by the `KeyedVectors.most_similar()` method originated from our dictionary.

Here is an example usage, which this PR also adds to the documentation:

``` python
from gensim.test.utils import common_texts, datapath
from gensim.models.word2vec import LineSentence

corpus = common_texts
model = FastText(corpus, vector_size=20, min_count=1)  # train word-vectors on a corpus
different_corpus = LineSentence(datapath('lee_background.cor'))
dictionary = Dictionary(different_corpus)
words = [word for word, count in dictionary.most_common()]  # construct vocabulary on a different corpus
word_vectors = model.wv.vectors_for_all(words)  # remove OOV word-vectors and infer vectors for new words
assert len(dictionary) == len(word_vectors)  # all words from our vocabulary received their word-vectors
```


Witiko:
As a syntactic sugar, we may recognize the `Dictionary` class and allow just `model.wv.vectors_for_all(dictionary)` to be specified.

gojomo:
As before, appreciate general utility of this method.

Regarding name, my 1st impression is that `fit()` is too generic, but maybe could be convinced if there's little else that could be confused with it. (But: mightn't someone get the impression that might do some fine-tuning?)

Perhaps `vector_subset()`? 

I don't think always sorting is a good default, given that sets of word-vectors are most-often sorted in decreasing frequency, which proves useful when users want just a head-subset of results, and often gives a noticeable cache-warmness benefit in typical lookup patterns (with all the frequent words clustered near the front). It could instead be never (let the caller specify exactly the order they want) or a non-default option-to-lexical-sort. (Similarly, I suspect a uniqueness-check could be left to the caller, with whatever oddness being created by a nonunique list-of-keys being survivable.)

Also: given the `vecattr` mechanism for adding typed extra info per key into a `KeyedVectors` instance - most often, word- frequencies – it may be interesting for such info to be preserved in this subset operation, either by default or (if the space/copy-over-time is too much to bear) only on-request.

Witiko:
> As before, appreciate general utility of this method.

Thank you. In contrast to the original proposal, I moved the method from `FastTextKeyedVectors` to `KeyedVectors`, because I realized that removing unwanted words is just as benefitial and it makes sense to fit word2vec word vectors to a dictionary even though some words from the dictionary may not receive a word vector.

> Regarding name, my 1st impression is that `fit()` is too generic, but maybe could be convinced if there's little else that could be confused with it. (But: mightn't someone get the impression that might do some fine-tuning?)
> 
> Perhaps `vector_subset()`?

If we only took a subset of existing word vectors, then `vector_subset()` would be a perfect match. However, I think that `fit()` is better at communicating that we also infer new word vectors. This inference almost requires a single forward pass of the fastText model, so I don't think the name promises more than it delivers.

> I don't think always sorting is a good default, given that sets of word-vectors are most-often sorted in decreasing frequency, which proves useful when users want just a head-subset of results, and often gives a noticeable cache-warmness benefit in typical lookup patterns

That is a good point. I was aiming at reproducibility, since we are enforcing uniqueness using `set()`, which randomizes the order. However, we can easily just filter out duplicates without chaning the original ordering:

```python
>>> from collections import OrderedDict
>>> list(OrderedDict.fromkeys([3, 2, 2, 3, 2, 1]))
[3, 2, 1]
```

> Similarly, I suspect a uniqueness-check could be left to the caller, with whatever oddness being created by a nonunique list-of-keys being survivable.

Since we are already doing a single pass over the entire iterable and we want to know the size of the vocabulary before we start building the `KeyedVectors` object, it seems reasonable to remove the duplicates and check inferrability, so that we don't overallocate memory. Do you see any utility in not removing the duplicates?

> Also: given the `vecattr` mechanism for adding typed extra info per key into a `KeyedVectors` instance - most often, word- frequencies – it may be interesting for such info to be preserved in this subset operation, either by default or (if the space/copy-over-time is too much to bear) only on-request.

At the moment, the lack of this functionality is explicitly noted in the docstring. Since we are not only subsetting word vectors, but also possibly inferring new ones, there is no guarantee that all word vectors in the resulting `KeyedVectors` object will have some attributes even if all word vectors in the original object had them, which seems quite confusing. Perhaps we should also add a parameter that would control whether new word vectors may be inferred.
```python
def fit(self, dictionary: Union[Iterable, Dictionary], allow_inference : bool = True, copy_vecattrs : bool = False):
    # ...
    if copy_vecattrs:
        for attr in self.expandos:
            try:
                val = self.get_vecattr(key, attr)
                kv.set_vecattr(key, attr, val)
            except KeyError:
                continue
    return kv
```

Witiko:
@gojomo 

> As before, appreciate general utility of this method.

Thank you. In contrast to the original proposal, I moved the method from `FastTextKeyedVectors` to `KeyedVectors`, because I realized that removing unwanted words is just as benefitial and it makes sense to fit word2vec word vectors to a dictionary even though some words from the dictionary may not receive a word vector.

> Regarding name, my 1st impression is that `fit()` is too generic, but maybe could be convinced if there's little else that could be confused with it. (But: mightn't someone get the impression that might do some fine-tuning?)
> 
> Perhaps `vector_subset()`?

If we only took a subset of existing word vectors, then `vector_subset()` would be a perfect match. However, I think that `fit()` is better at communicating that we may also infer new word vectors. This inference (almost) requires a single forward pass of the fastText model, so I don't think the name promises more than it delivers from the `sklearn` perspective. From the language perspective, the method shrinks-and-expands the `KeyedVectors` object so that it *fits* the dictionary.

> I don't think always sorting is a good default, given that sets of word-vectors are most-often sorted in decreasing frequency, which proves useful when users want just a head-subset of results, and often gives a noticeable cache-warmness benefit in typical lookup patterns

That is a good point. I was aiming at reproducibility, since we are enforcing uniqueness using `set()`, which randomizes the order. However, we can easily just filter out duplicates without reordering:

```python
>>> from collections import OrderedDict
>>> list(OrderedDict.fromkeys([3, 2, 2, 3, 2, 1]))
[3, 2, 1]
```

> Similarly, I suspect a uniqueness-check could be left to the caller, with whatever oddness being created by a nonunique list-of-keys being survivable.

Since we are already doing a single pass over the entire iterable and we want to know the size of the vocabulary before we start building the `KeyedVectors` object, it seems reasonable to remove the duplicates and check inferrability, so that we don't overallocate memory. Do you see any utility in not removing the duplicates?

> Also: given the `vecattr` mechanism for adding typed extra info per key into a `KeyedVectors` instance - most often, word- frequencies – it may be interesting for such info to be preserved in this subset operation, either by default or (if the space/copy-over-time is too much to bear) only on-request.

At the moment, the lack of this functionality is explicitly noted in the docstring. Since we are not only subsetting word vectors, but also possibly inferring new ones, there is no guarantee that all word vectors in the resulting `KeyedVectors` object would have vector attributes even if all word vectors in the original `KeyedVectors` object had them, which seems confusing. Perhaps we should also add a parameter that would control whether new word vectors may be inferred. If new word vectors may not be inferred, then this gives us a stronger guarantee about the presence of vector attributes.
```python
def fit(self, dictionary: Union[Iterable, Dictionary], allow_inference : bool = True, copy_vecattrs : bool = False):
    # ...
    if copy_vecattrs:
        for attr in self.expandos:
            try:
                val = self.get_vecattr(key, attr)
                kv.set_vecattr(key, attr, val)
            except KeyError:
                continue
    return kv
```

Witiko:
> If we only took a subset of existing word vectors, then `vector_subset()` would be a perfect match. However, I think that `fit()` is better at communicating that we may also infer new word vectors. This inference (almost) requires a single forward pass of the fastText model, so I don't think the name promises more than it delivers from the `sklearn` perspective. From the language perspective, the method shrinks-and-expands the `KeyedVectors` object so that it _fits_ the dictionary.

It is true that `fit()` strongly implies in-place modification of a mutable object, which is something our method does not satisfy.

gensim/models/keyedvectors.py
```diff
@@ -1722,7 +1722,10 @@ def vectors_for_all(self, keys: Iterable) -> 'KeyedVectors':
             Vectors for all the given keys.
 
         """
-        vocabulary = sorted(set(filter(lambda key: key in self, keys)))
+        if isinstance(keys, Dictionary):
+            vocabulary = keys.token2id
+        else:
+            vocabulary = sorted(set(filter(lambda key: key in self, keys)))
```
piskvorky: Avoid `filter` and `map`, use comprehension (idiomatic Python).

gensim/models/keyedvectors.py
```diff
@@ -1696,8 +1696,8 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
             msg=f"merged {overlap_count} vectors into {self.vectors.shape} matrix from {fname}",
         )
 
-    def vectors_for_all(self, keys: Iterable) -> 'KeyedVectors':
-        """Produces vectors for all keys in a given iterable.
+    def vectors_for_all(self, keys: Union[Iterable, Dictionary]) -> 'KeyedVectors':
```
piskvorky, line 1700: Why do we single out `Dictionary`?

Accepting only a sequence of strings seems cleaner API. Ppl who have `Dictionary` can simply pass `Dictionary.token2id`, no?

gensim/models/keyedvectors.py
```diff
@@ -1695,6 +1697,44 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
             msg=f"merged {overlap_count} vectors into {self.vectors.shape} matrix from {fname}",
         )
 
+    def vectors_for_all(self, keys: Union[Iterable, Dictionary]) -> 'KeyedVectors':
+        """Produces vectors for all given keys.
```
piskvorky, line 1701: Docstrings in imperative mode (PEP8).

```suggestion
        """Produce vectors for all given keys.
```

piskvorky (changes requested):

gensim/models/keyedvectors.py
Witiko: Thank you for the feedback. As per our discussion with @gojomo in https://github.com/RaRe-Technologies/gensim/pull/3157#issuecomment-848081847, I have removed sorting and replaced this part with deduplication:
```
>>> from collections import OrderedDict
>>> list(OrderedDict.fromkeys([3, 2, 2, 3, 2, 1]))
[3, 2, 1]
```
This is still not too idiomatic, but very succint. Perhaps a comment `  # deduplicates keys` would make a maintainer's life easier.

gensim/models/keyedvectors.py
piskvorky: I see what you mean, but am -1 on complicating the API with two input types. IMO that increases the cognitive load and makes people look up the docs, rather than the opposite.

Where one code path merges into the other via a simple `input = input.some_attribute`, that can (should) be done by the caller. A cleaner / leaner API is preferable.

piskvorky: That's not what I see in this PR though:

```python
vocabulary = [key for key in keys.token2id if key in self]
```

Witiko: Indeed, and it's likely not even what we want. As discussed with @gojomo in https://github.com/RaRe-Technologies/gensim/pull/3157#issuecomment-848081847, we likely want to order the dictionary by decreasing frequency for cache locality. Nevertheless, this is a good reason to special-case `Dictionary`, since it's not easy to get this right.

Witiko: I will convert this to draft, since there are pending changes.

Witiko:
@piskvorky, @gojomo Following the discussion, the `vectors_for_all()` method now:

- Supports `Dictionary` objects, which it sorts by decreasing collection frequency for cache-warmness.
- Supports copying vector attributes with `copy_vecattrs=True`.
- Supports disabling the inference of new vectors `allow_inference=False`, so that we can have a guarantee that a) we will always receive a subset of the vectors and b) all vectors will have vector attributes defined when `copy_vecattrs=True`.
- Adds five new unit tests.

Unless we want to rename the method (@gojomo suggested `vector_subset()`, I suggested `fit()`, `materialize()` or `vectorize()`, but `vector_subset()` and `fit()` are both misleading and I don't prefer `materialize()`, or `vectorize()` over `vectors_for_all()`), or remove the special support for `Dictionary` objects (suggested by @piskvorky, but since the use of `Dictionary` will be common and getting a properly sorted iterable out of `Dictionary` is difficult, I vote we keep it), this PR should now be ready for review.

Witiko:
Unless we want to rename the method (@gojomo suggested `vector_subset()`, I suggested `fit()`, `materialize()`, or `vectorize()`, but `vector_subset()` and `fit()` are both misleading as discussed in https://github.com/RaRe-Technologies/gensim/pull/3157#issuecomment-848706184 and https://github.com/RaRe-Technologies/gensim/pull/3157#issuecomment-849172789 and I don't prefer `materialize()` or `vectorize()` over `vectors_for_all()`), or remove the special support for `Dictionary` objects (suggested by @piskvorky, but since the use of `Dictionary` will be common and getting a properly sorted iterable out of `Dictionary` is tricky, I vote we keep it), this PR should now be ready for review.

gensim/models/keyedvectors.py
Witiko: Is this solution OK for you?

gensim/models/keyedvectors.py
Witiko: Is this solution OK for you?

gensim/models/keyedvectors.py
```diff
@@ -1695,6 +1697,79 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
...
+            else:
+                return key in self.key_to_index
+
+        if isinstance(keys, Dictionary):
+            vocab = sorted(keys.cfs.items(), key=lambda x: (-x[1], x[0]))  # sort by decreasing collection frequency
+            vocab = (keys[key] for key, _ in vocab)
+        else:
+            vocab = keys
+
+        vocab = filter(is_key_defined, vocab)  # remove undefined keys
```
piskvorky, line 1754: Please avoid `filter` / `map` (anti-pattern), prefer comprehension.

In this case, the (trivial) inner function seems useless:
```suggestion
        vocab = [key in self if allow_inference else key in self.key_to_index for key in vocab]
```

gensim/models/keyedvectors.py
```diff
@@ -1695,6 +1697,79 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
...
+
+        for key in vocab:  # produce and index vectors for all the given keys
+            weights = self[key]
+            _add_word_to_kv(kv, None, key, weights, len(vocab))
+            if not copy_vecattrs:
+                continue
+            for attr in self.expandos:  # copy vector attributes to the new KeyedVectors object
+                try:
+                    val = self.get_vecattr(key, attr)
+                    kv.set_vecattr(key, attr, val)
```
piskvorky, line 1768: No need for the unused temporary var:

```suggestion
                    kv.set_vecattr(key, attr, self.get_vecattr(key, attr))
```

piskvorky (changes requested):
Thanks for the clean up.

I'm still not convinced special-casing Dictionary is worth complicating the API, but I don't feel that strongly about it. @gojomo @mpenkov please review.

gensim/models/keyedvectors.py
gojomo: I don't personally mind the `map`/`filter` style in Python, but if something can be done in a suitably compact 1-liner, and/or not resort to custom functions defined elsewhere (even just a few lines away), I prefr the comprehension version for all-in-one-place clarity.

gensim/models/keyedvectors.py
```diff
@@ -1695,6 +1697,79 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
...
+                        copy_vecattrs: bool = False) -> 'KeyedVectors':
+        """Produce vectors for all given keys as a new :class:`KeyedVectors` object.
+
+        Notes
+        -----
+        The keys will always be deduplicated. For optimal performance, you should not pass entire
+        corpora to the method. Instead, you should construct a
+        :class:`~gensim.corpora.dictionary.Dictionary` object out of your corpus first:
+
+        >>> import gensim.downloader as api
```
gojomo, line 1711: Because I consider the post-installation loading & execution of arbitrary new, unversioned source code to be a dangerously reckless process we should eliminate, for the reasons in #2283, I never like to see the use of `gensim.downloader` encouraged.

gojomo:
I see mostly downside from special handling of `Dictionary`: 
* complicates signature type
* adds type-switched branches
* complicates docs & leads to current doc-comment example where naive users may think they *need* to use `Dictionary`
* adds cross-module dependency on `gensim.corpora.dictionary`

If main motication is helping user get a properly freqeuncy-sorted iterable, maybe as that as a utility method on `Dictionary`? Then using a plain-iterable (or even `list`) API here is a 1-liner for those who are starting with a `Dictionary`.

gensim/models/keyedvectors.py
```diff
@@ -1695,6 +1697,79 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
...
+        >>> from gensim.test.utils import common_texts
+        >>>
+        >>> model = api.load('word2vec-google-news-300')
+        >>> corpus = common_texts
+        >>> dictionary = Dictionary(corpus)  # construct a vocabulary out of your corpus
+        >>> word_vectors = model.wv.vectors_for_all(dictionary)  # create word-vectors for words in your corpus
```
gojomo, line 1717: Without `Dictionary`, this sort of example could be:
```suggestion
        >>> word_counts = Counter(itertools.chain.fron_iterable(corpus))  # count words in corpus
        >>> words_by_freq = [k for k, v in word_counts.most_common()]
        >>> word_vectors = model.wv.vectors_for_all(words_by_freq)  # create word-vectors for words in your corpus
```

gensim/models/keyedvectors.py
Witiko: The use of `gensim.downloader` is pervasive throughout auto-examples and docstrings. I would suggest making a batch change to all of these when there is an agreement on how to resolve #2283. What would you suggest?

Witiko:
I see `Dictionary` as a core class of gensim. However, if the policy is to keep most of the communication between gensim classes in standard library objects, then adding a `Dictionary` method that turns the dictionary into an iterable of words in the decreasing order of frequency is definitely an option.

gensim/models/keyedvectors.py
Witiko: Fixed in 26baf6d3.

gensim/models/keyedvectors.py
Witiko: Fixed in 98c070e9.

gensim/models/keyedvectors.py
Witiko: I added the `most_common` method to the `Dictionary` class together with four unit tests and I incorporated your above suggestion in a4590c10.

Witiko:
I removed the special-casing of `Dictionary` in a4590c10 and a9907377. This change adds the `Dictionary.most_common` method, which closely follows the `Counter.most_common` method from the standard library. I updated the examples and added four unit tests for this new method.

gensim/models/keyedvectors.py
mpenkov: For the sake of simplicity, if it's possible to avoid using the downloader by using an inline corpus, or something readily available, then the documentation would be better without the downloader.

On top of the problems that @gojomo mentioned, using the downloader instantly makes the docstring non-doctest-able, because downloading those datasets can take a long time.

gensim/models/keyedvectors.py
```diff
@@ -1695,6 +1697,64 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
...
+            To ensure that *all* the produced vectors will have vector attributes assigned,
+            you should set `allow_inference=False`.
+
+        Returns
+        -------
+        keyedvectors : :class:`~gensim.models.keyedvectors.KeyedVectors`
+            Vectors for all the given keys.
+
+        """
+        vocab = (key for key in keys if key in (self if allow_inference else self.key_to_index))  # drop undefined keys
```
mpenkov, line 1740: This comprehension feels a little busy to me. I think this would be more readable:

```suggestion
        if allow_inference:
            vocab = list(OrderedDict.fromkeys((key for key in keys if key in self)))
        else:
            vocab = list(OrderedDict.fromkeys((key for key in keys if key in self.key_to_index)))
```

If we're planning to use type checking later, we should avoid redefining variables to be of different type. For example, in the original code, you start with vocab being a generator, but then turn it into a list. Mypy will complain about that. It's better to either use a different variable, or avoid reassignment altogether as I did in the suggestion above.

gensim/models/keyedvectors.py
```diff
@@ -1695,6 +1697,64 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
...
+        vocab = list(OrderedDict.fromkeys(vocab))  # deduplicate keys
+
+        datatype = self.vectors.dtype
+        kv = KeyedVectors(self.vector_size, len(vocab), dtype=datatype)  # preallocate new KeyedVectors object
```
mpenkov, line 1744: I don't think the comment adds any value to the code. It's stating _what_ you're doing, which should already be obvious from the code itself. You also don't need to declare a datatype variable, you're only using it once.

```suggestion
        kv = KeyedVectors(self.vector_size, len(vocab), dtype=self.vectors.dtype)
```

gensim/models/keyedvectors.py
```diff
@@ -1695,6 +1697,64 @@ def intersect_word2vec_format(self, fname, lockf=0.0, binary=False, encoding='ut
...
+
+        for key in vocab:  # produce and index vectors for all the given keys
+            weights = self[key]
+            _add_word_to_kv(kv, None, key, weights, len(vocab))
+            if not copy_vecattrs:
+                continue
+            for attr in self.expandos:  # copy vector attributes to the new KeyedVectors object
```
mpenkov, line 1751: Again, the comment is saying something that should be obvious from the code. You can make it even more obvious by doing:

```python
if copy_vecattrs:
    for attr in self.expandos:
        ...
```

gensim/test/test_corpora_dictionary.py
```diff
@@ -359,6 +359,30 @@ def test_patch_with_special_tokens(self):
         self.assertNotIn((1, 1), d.doc2bow(corpus_with_special_tokens[0]))
         self.assertIn((1, 1), d.doc2bow(corpus_with_special_tokens[1]))
 
+    def test_most_common_with_n(self):
+        texts = [['human', 'human', 'human', 'computer', 'computer', 'interface', 'interface']]
+        d = Dictionary(texts)
+        expected = [('human', 3), ('computer', 2)]
+        self.assertEqual(d.most_common(n=2), expected)
```
mpenkov, line 366: We use pytest, so please use the simpler assertion syntax, here and everywhere.

```suggestion
        assert d.most_common(n=2) == expected
```

mpenkov (changes requested):
@Witiko Left you some comments. Please have a look.

gensim/corpora/dictionary.py
```diff
@@ -689,6 +690,32 @@ def load_from_text(fname):
                 result.dfs[wordid] = int(docfreq)
         return result
 
+    def most_common(self, n: Optional[int] = None, include_counts: bool = True) -> List[str]:
```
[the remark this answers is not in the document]
Witiko: > What's your stance on type annotations?

Although not intended for me, this has been discussed by @gojomo and I in the originating thread, see https://github.com/RaRe-Technologies/gensim/pull/3146#issuecomment-844164000.

> The return type isn't entirely correct [...]

Thank you, that escaped my attention.

gensim/models/keyedvectors.py
Witiko: > This comprehension feels a little busy to me. I think this would be more readable:

We had a more readable version in the course of the review: https://github.com/RaRe-Technologies/gensim/pull/3157#discussion_r642614592 I find it preferable to this code, which features heavy duplication.

gensim/test/test_corpora_dictionary.py
Witiko: Forgive my ignorance. Is this really equivalent from the viewpoint of informative error messages?

gensim/corpora/dictionary.py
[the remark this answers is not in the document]
piskvorky: > @piskvorky What's your stance on type annotations?

I find them distracting, making the code less readable. I personally don't need / use them.

But I can see how Python type annotations might help some people. So I'm not opposed to someone adding them to Gensim (although that will result in a negative to me personally).

gensim/models/keyedvectors.py
piskvorky: Since we're in the bike shed territory, that whole `OrderedDict` approach seems unnecessarily opaque to me. That's just for deduplication that keeps the keys in the original order, right? Can you remind me, why do we even want that?

And assuming we do:

```python
        # Pick only the keys that actually exist & deduplicate them.
        vocab, seen = [], set()
        for key in keys:
            if key not in seen:
                seen.add(key)
                if key in (self if allow_inference else self.key_to_index):
                    vocab.append(key)
```

And assuming we don't:

```python
# Pick only the keys that actually exist & deduplicate them.
vocab = list(set(key for key in vocab if key in (self if allow_inference else self.key_to_index)))
```

Witiko: @piskvorky To keep high-frequency terms clustered together for cache-warmness, see https://github.com/RaRe-Technologies/gensim/pull/3157#issuecomment-848081847. The first version with a set is nice and readable.

piskvorky: >  for cache-warmnes

Is that even measurable? Sounds like a premature optimization to me.

gensim/corpora/dictionary.py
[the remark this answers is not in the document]
Witiko: Type annotations would also help us catch type errors in the CI before they happen. From my experience, it is a big help and moves Python much closer to the "when it compiles, it works" territory. Type checkers such as mypy or ptype can actually infer the types, so writing them out explicitly in the code is just for the human eye. The fact that they make the life of IDE users easier ranks pretty low on my list.

gensim/models/keyedvectors.py
Witiko: @gojomo profiled the code of word2vec and fastText in the past, so I assumed his suggestion is based on this experience.

piskvorky: OK, thanks. I have no strong feeling either way, any of the two options (or even the current PR code) is fine with me.

gojomo: It's conventional, but not required, for assignment of words-to-ordered-positions to put the most-frequent words early. This makes it easy to limit operations to the most-common words - for space or time efficiency, or because the long-tail of rare words tend to have poorer-quality vectors, etc. But also: I've seen mention in papers, and observed long-ago in my own tests, that realistic workloads (including training itself) go faster with frequency-sorted vector assignment - by ensuring the most-looked-up/modified vectors are all in the same few always-in-cache pages. 

So, while it's not a hard requirement to always do/maintain such ordering, we should try not to inadvertently make choices that (a) ruin/scramble that beneficial ordering, when it already exists; or (b) override a users' desired ordering, for these benefits, or anything else the user might want to test.

gensim/test/test_corpora_dictionary.py
gojomo: Also: overwhelming weight-of-existing-unit-test-code (in this file and elsewhere) seems to prefer the assertion-methods over the `assert` keyword.

gensim/models/keyedvectors.py
piskvorky: Yes, whichever. Just put a better comment there please @Witiko, explaining the rationale (see my example above), so someone doesn't inadvertently remove this later thinking it serves no purpose.

gensim/test/test_corpora_dictionary.py
mpenkov: > Is this really equivalent from the viewpoint of informative error messages?

Yes, pytest takes care of that.

> overwhelming weight-of-existing-unit-test-code (in this file and elsewhere) seems to prefer the assertion-methods over the assert keyword.

Most tests were written before gensim started using pytest in 2017. The trade-off here is test readability vs consistency with an older test framework (unittest has it's pros and cons, but we've been using pytest for years, so let's not debate them here). My opinion is that for newer tests, readability is more important than consistency.

gojomo: Aha. Well, if keyword-style has some advantages under `pytest` such that it's now preferred style, that could be declared on the project style guidelines, including some hint about if/when older usages should be changed, and whether matching existing style of a method/class or new-style should be preferred.

gensim/models/keyedvectors.py
Witiko: Gotcha!

gensim/corpora/dictionary.py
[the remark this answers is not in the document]
Witiko: @mpenkov 

> Is it necessary to have this conditional behavior? Can we not always return the tuples, and get the user to unpack them, if necessary? What is the benefit of having this additional parameter in the API?

The benefit is convenience, but it's true that it makes the API a little complicated. Let's get rid of `include_counts` then, I don't even mind.

Witiko:
Thank you again for the many thoughtful suggestions raised in the reviews. I will implement them in bulk once I have drafted the demo that I promised @piskvorky in https://github.com/RaRe-Technologies/gensim/pull/3146#issuecomment-850688035.

gensim/models/keyedvectors.py
Witiko: I removed the Gensim downloader from the examples in 32c14c5d7ecc0577bfc59e1c3cf9e96bbb3b9bc9 and 712ee6164e18215092859d5e4611d8ccb06578a9.

gensim/corpora/dictionary.py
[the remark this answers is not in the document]
Witiko: I removed `include_counts` in 9acbcba25a8981c68acdf0ec41514b2895d6c071 and b8625a5360ec6482bddf4ebaaa9f567cf3fd6120.

gensim/test/test_corpora_dictionary.py
Witiko: I have added the pytest assertion syntax in 4aacad27.

gensim/models/keyedvectors.py
Witiko: Fixed in a86522c6.

gensim/models/keyedvectors.py
Witiko: Fixed in ebc276d1.

gensim/similarities/termsim.py
```diff
@@ -409,20 +431,19 @@ class SparseTermSimilarityMatrix(SaveLoad):
     >>> from gensim.models import Word2Vec
     >>> from gensim.similarities import SoftCosineSimilarity, SparseTermSimilarityMatrix, WordEmbeddingSimilarityIndex
     >>> from gensim.similarities.index import AnnoyIndexer
-    >>> from scikits.sparse.cholmod import cholesky
     >>>
     >>> model = Word2Vec(common_texts, vector_size=20, min_count=1)  # train word-vectors
     >>> annoy = AnnoyIndexer(model, num_trees=2)  # use annoy for faster word similarity lookups
```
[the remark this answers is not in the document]
Witiko: Fixed in 3bf7f339.

Witiko:
I have implemented all suggestions from the reviews.

gensim/test/test_fasttext.py
```diff
@@ -850,6 +850,54 @@ def obsolete_testLoadOldModel(self):
...
+
+        smaller_distance = np.linalg.norm(
+            vectors_for_all['an out-of-vocabulary word']
+            - vectors_for_all['another out-of-vocabulary word']
+        )
+        greater_distance = np.linalg.norm(
+            vectors_for_all['an out-of-vocabulary word']
+            - vectors_for_all['responding']
+        )
+        self.assertGreater(greater_distance, smaller_distance)
```
mpenkov, line 880:
```suggestion
        assert greater_distance > smaller_distance
```

gensim/test/test_fasttext.py
```diff
@@ -850,6 +850,54 @@ def obsolete_testLoadOldModel(self):
...
+        ]
+        vectors_for_all = self.test_model.wv.vectors_for_all(words, allow_inference=False)
+
+        expected = 3
+        predicted = len(vectors_for_all)
+        assert expected == predicted
+
+        expected = self.test_model.wv['responding']
+        predicted = vectors_for_all['responding']
+        self.assertTrue(np.allclose(expected, predicted))
```
mpenkov, line 899:
```suggestion
        assert np.allclose(expected, predicted)
```

gensim/test/test_keyedvectors.py
```diff
@@ -39,6 +39,43 @@ def test_most_similar(self):
...
+        ]
+        vectors_for_all = self.vectors.vectors_for_all(words)
+
+        expected = 3
+        predicted = len(vectors_for_all)
+        assert expected == predicted
+
+        expected = self.vectors['conflict']
+        predicted = vectors_for_all['conflict']
+        self.assertTrue(np.allclose(expected, predicted))
```
mpenkov, line 59:
```suggestion
        assert np.allclose(expected, predicted)
```

gensim/test/test_fasttext.py
```diff
@@ -850,6 +850,54 @@ def obsolete_testLoadOldModel(self):
...
+        ]
+        vectors_for_all = self.test_model.wv.vectors_for_all(words)
+
+        expected = 5
+        predicted = len(vectors_for_all)
+        assert expected == predicted
+
+        expected = self.test_model.wv['responding']
+        predicted = vectors_for_all['responding']
+        self.assertTrue(np.allclose(expected, predicted))
```
mpenkov, line 870:
```suggestion
        assert np.allclose(expected, predicted)
```

gensim/test/test_fasttext.py
mpenkov:
```suggestion
        assert greater_distance > smaller_distance
```

gensim/test/test_fasttext.py
mpenkov:
```suggestion
        assert np.allclose(expected, predicted)
```

gensim/test/test_keyedvectors.py
mpenkov:
```suggestion
        assert np.allclose(expected, predicted)
```

mpenkov (approved):
Looks good! Thank you @Witiko !

PR #3169, opened 2021-06-09
Implement `shrink_windows` argument for Word2Vec.

### LINKED ISSUE #3159: Enable to master the window size variations during training
**Context:** I am trying to use `gensim` on a corpus in which the order of words has no importance at all. So my idea is to give the constructor of the class `Word2Vec` a large window size, so that it trains each word with the full rest of the sentence (the iterable I give as parameter `sentences` to the `Word2Vec` constructor yields sentences in which word order is irrelevant).

I noticed [here](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/word2vec_inner.pyx#L572) in the code that during training, the window size was reduced of a size drawn uniformly at random. If I get it right, it feels like saying, if words are too far from one another, sometimes they carry less sense together. And at any rate, adding some randomness very often gives better results in machine learning in general. In general I agree with these two ideas.
- However, I find it too bad that this is hard-coded, and that the user cannot choose to keep the window fix-sized, with the size they enter. 
-  Another cool option would be to be able to input a minimal and a maximal size, instead of the current behaviour where the window size varies from the full length provided by the user to 0.

**Propositions:**
 From these observations I propose to do one or all of the following changes:
- Add a parameter to choose whether or not the window size randomly reduces. This could be something like `change_window_size`, set to `True` by default to be consistent with the current behaviour. 
- Add parameters for lower and higher bounds of the final window size. Note that if these bounds are equal, this amounts to no random reduction at all and this replaces the behaviour of the previous proposition.
- [Bonus] We could also imagine that instead of drawing sizes uniformly at random, the user could choose the probability distribution. But this is more of an idea I'm throwing into the air, rather than one I'm ready to implement right now. I don't know if it is relevant enough, considering the code changes it may imply.

What do you think of this ? If you're ok with it, I plan on doing the modifications myself.

M-Demay:
Fixes #3159.
The idea is to allow the user to use a window with fixed-size, contrary to the current default behaviour which is to pick the window size in [1, `window`] uniformly at random. We want this alternative behaviour for a project.
We did it by adding a boolean parameter `reduced_windows`. For compatibility, we put it at the end of the arguments list, default to `False`. 
The user can set it directly in the constructor of the class `Word2Vec`, or at any further step (for example, to retrain the model in a second step), since we propagated it to the downstream calls.


gensim/models/word2vec_inner.pyx
```diff
@@ -71,10 +71,10 @@ cdef void our_saxpy_noblas(const int *N, const float *alpha, const float *X, con
     for i from 0 <= i < N[0] by 1:
         Y[i * (incY[0])] = (alpha[0]) * X[i * (incX[0])] + Y[i * (incY[0])]
 
-cdef long long _mul(const np.uint32_t a, const int b) nogil: 
+cdef long long _mul(const np.uint32_t a, const int b) nogil:
     """Safe multiplication of ints with explict typecasting"""
     return <long long>a * <long long>b
-
```
piskvorky, line 77: Nice catch! Trailing whitespace should never have been merged. CC @mpenkov .

gensim/models/word2vec_inner.pyx
```diff
@@ -570,7 +573,11 @@ def train_batch_sg(model, sentences, alpha, _work, compute_loss):
             break  # TODO: log warning, tally overflow?
 
     # precompute "reduced window" offsets in a single randint() call
-    for i, item in enumerate(model.random.randint(0, c.window, effective_words)):
+    if reduced_windows:
+        window_size = model.random.randint(0, c.window, effective_words)
+    else:
+        window_size = [0] * effective_words
```
piskvorky, line 579: Why `0`?

gensim/models/word2vec_corpusfile.pyx
```diff
@@ -225,7 +226,10 @@ cdef void prepare_c_structures_for_batch(
 
     # precompute "reduced window" offsets in a single randint() call
     for i in range(effective_words[0]):
-        reduced_windows[i] = random_int32(next_random) % window
+        if do_reduced_windows:
+            reduced_windows[i] = random_int32(next_random) % window
```
piskvorky, line 230: This doesn't look like `[1, window]`. Or rather, if `reduced_windows` is expected to contain "window minus one", shouldn't your new code path be `reduced_windows[i] = window - 1`?

piskvorky (changes requested):

gensim/models/word2vec.py
```diff
@@ -240,7 +240,7 @@ def __init__(
             max_vocab_size=None, sample=1e-3, seed=1, workers=3, min_alpha=0.0001,
             sg=0, hs=0, negative=5, ns_exponent=0.75, cbow_mean=1, hashfxn=hash, epochs=5, null_word=0,
             trim_rule=None, sorted_vocab=1, batch_words=MAX_WORDS_IN_BATCH, compute_loss=False, callbacks=(),
-            comment=None, max_final_vocab=None,
+            comment=None, max_final_vocab=None, reduced_windows=True
```
piskvorky, line 243: Please add a trailing comma at the end of enumerations:
https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#code-style

piskvorky (changes requested):

gensim/models/word2vec.py
```diff
@@ -345,6 +345,9 @@ def __init__(
             :meth:`~gensim.models.word2vec.Word2Vec.get_latest_training_loss`.
         callbacks : iterable of :class:`~gensim.models.callbacks.CallbackAny2Vec`, optional
             Sequence of callbacks to be executed at specific stages during training.
+        reduced_windows : bool, optional
+            If True, the window size is uniformly sampled from {1, `window`}
+            during training. Otherwise, it is fixed to `window`.
```
gojomo, line 350: I would extend this comment with more explanation, eg:

    If True, the effective window size is uniformly sampled from  [1, `window`] 
    for each target word during training, to match the original word2vec algorithm's 
    approximate weighting of context words by distance. Otherwise, the effective 
    window size is  always fixed to `window`.

gensim/models/word2vec.py
```diff
@@ -910,12 +915,14 @@ def _do_train_epoch(
         if self.sg:
             examples, tally, raw_tally = train_epoch_sg(
                 self, corpus_file, offset, cython_vocab, cur_epoch,
-                total_examples, total_words, work, neu1, self.compute_loss,
+                total_examples, total_words, work, neu1,
+                self.compute_loss, self.reduced_windows,
```
gojomo, line 919: In general, if the parameter `window` wasn't being passed-along in a widened method-signature - because the inner implementation method was able to read it from elsewhere, then this new closely-related `reduced_windows` parameter probably should be consulted the same way. So: I don't think any method that didn't already have `window` in its signature shouild be adding `reduced_windows`.

gensim/models/word2vec.py
```diff
@@ -941,20 +948,26 @@ def _do_train_job(self, sentences, alpha, inits):
...
 
     def train(
-            self, corpus_iterable=None, corpus_file=None, total_examples=None, total_words=None,
-            epochs=None, start_alpha=None, end_alpha=None, word_count=0,
-            queue_factor=2, report_delay=1.0, compute_loss=False, callbacks=(),
-            **kwargs,
+            self, corpus_iterable=None, corpus_file=None, total_examples=None,
+            total_words=None, epochs=None, start_alpha=None, end_alpha=None,
+            word_count=0, queue_factor=2, report_delay=1.0, compute_loss=False,
+            reduced_windows=True, callbacks=(), **kwargs,
```
gojomo, line 970: Per above: this parameter shouldn't need to be here. (Further: if it were here, it shouldn't override a possible user-chosen value of `False` from elsewhere.)

gensim/models/word2vec_corpusfile.pyx
```diff
@@ -186,7 +186,8 @@ cdef void prepare_c_structures_for_batch(
         vector[vector[string]] &sentences, int sample, int hs, int window, long long *total_words,
         int *effective_words, int *effective_sentences, unsigned long long *next_random,
         cvocab_t *vocab, int *sentence_idx, np.uint32_t *indexes, int *codelens,
-        np.uint8_t **codes, np.uint32_t **points, np.uint32_t *reduced_windows) nogil:
+        np.uint8_t **codes, np.uint32_t **points, np.uint32_t *reduced_windows,
+        bint do_reduced_windows) nogil:
```
gojomo, line 190: GIven that there's an existing variable, `reduced_windows`, which is an array of the precomputed sampled effective windows (something simply called `b` in the original `word2vec.c` source code IIRC), we definitely should *not* be naming the top-level toggle parameter the same thing almost everywhere else, but then change that toggle to `do_reduced_windows` here. That's a recipe for unnecessary confusion. The top-level param should be a name that is (1) different; (2) consistent. Perhaps a change as small as `reduce_windows` (imperative sense) is enough; but for more contrast & descriptiveness, maybe the parameter should be named `shrink_windows`.

gojomo: Also: our other de facto boolean toggles - like `hs` and `sample` here – have otherwise been passed as type `int`, not `bint`. Even if `bint` might be better, I'd prefer consistency.

gensim/models/word2vec_corpusfile.pyx
gojomo: The *effective* window winds up being from `1` to `window` (inclusive), because these `0` to `window - 1` values get offset against a full `window` sum, per the indexing math just a few lines down: https://github.com/RaRe-Technologies/gensim/blob/6afb3ff9bcaa4f5faef05a4223c324852b2e4c22/gensim/models/word2vec_inner.pyx#L589-L594

(I suspect it's this way because it was mimicking something in the original C code - it could be clearer if `reduced_windows` values became the literal, actually-used window size, by adding 1, changing the later indexing to match.)

gensim/models/word2vec_inner.pyx
gojomo: Per above, the legacy summing logic means a `0` here leaves the full specified `window` in use. (You can essentially read the `reduced_windows` array as a list of values to 'reduce' the window by, with a 'reduction' of `0` leaving the full window in place.)

gensim/models/word2vec_corpusfile.pyx
piskvorky: That's what I thought – hence my off-by-one warning for the new code path.

gensim/models/word2vec_inner.pyx
piskvorky: Oh, OK. This variable is terribly named, then.

gojomo: But also, there's no need to create this list-of-0-values just to feed the next loop. The `if reduced_windows` branch should let each path fully initialize `c.reduced_windows`. EG:

```python
    if reduced_windows:
        for i, item in enumerate(model.random.randint(0, c.window, effective_words)):
            c.reduced_windows[i] = item
    else:
        c.reduced_windows[:] = 0
```

gensim/models/word2vec_corpusfile.pyx
pandrey-fr: For the variable-size window, the rng code is the same as in the current `develop` branch.
But you are right, there is a mistake here: in the case when we want fixed-size windows, it should be `reduced_windows[i] = 0`.
This change will be part of an upcoming commit.

gensim/models/word2vec.py
pandrey-fr: Will do; sorry about that.

gensim/models/word2vec_inner.pyx
pandrey-fr: To be fair, the text editor did this!

gensim/models/word2vec_inner.pyx
pandrey-fr: That was indeed a bad variable name choice on my behalf; sorry.
The next commit will incorporate the suggested change.

gensim/models/word2vec.py
pandrey-fr: Okay, we will use this suggested docstring.

gensim/models/word2vec.py
pandrey-fr: I agree for all "backend" methods and functions and have corrected them.
I have a slight hesitation regarding the `train` method (see discussion below).

gensim/models/word2vec.py
pandrey-fr: I have mixed feeling, influenced by the fact that `compute_loss` is an example of a parameter overridden here in spite of its being also declared at instantiation. So I think it might be useful to leave the parameter open for overriding here - but if you disagree I will of course take it off.
At the very least I agree with the fact that it should not override the preset value if left to a default `None` value - a behaviour corrected in the upcoming commit.

gensim/models/word2vec_corpusfile.pyx
pandrey-fr: We (M-Demay and I) used this name based on the discussion in the associated issue, but we both agree that is was not such a great choice. The next commit goes for `shrink_windows`.
(and ok for using `int` rather than `bint`)

gensim/models/word2vec.py
gojomo: The current haphazard attempts at computing running loss are a big mess - see for example #2617 & other issues referenced therein. So I'd not use this `compute_loss` parameter as a model of good practice. Notably, also, loss-computation *might* involve an extra overhead people would only sometimes want to pay across multiple `train()` callss, and shouldn't change algorithmic behavior at all. But for enabling/disabling variable-windows, that *does* change the algorithm's operation, and I can't really imagine a case where someone would want to try it both ways in the same model, on subsequent `.train()` calls. So having this novel algorithmic parameter here – but not many others! –  doesn't fit.

pandrey-fr: Thank you for the detailed argument; this makes perfect sense and the last commit on Friday removes the parameter from `train`.

pandrey-fr:
Tests are failing on old models' reloading due to the added `shrink_windows` attribute. I will try to understand where we need to edit the code to add the attribute, although I do not quite understand how old models' reloading work - apparently it does not run things through the current version's object instantiation process? Otherwise the newly-introduced parameter should be added to the loaded models (with a default value of True that implements the same behaviour as the legacy one).

gojomo:
In general the right place to path older models, upon load, to ensure they have all currently-required properties is `_load_specials()` - and a new parameter should be assigned the value that would preserve the prior behavior. See for example all the current check-and-add-if-missing logic: 

https://github.com/RaRe-Technologies/gensim/blob/7b6e73a7dd2bf6e74c6b7099c02fe49d282156ea/gensim/models/word2vec.py#L1955

Separately: it occurs to me that it should be checked that the new option will work in `Word2Vec`'s closely-related subclasses, `Doc2Vec` and `FastText`, as well. (It will likely require similar changes in their `.pyx` files, and perhaps some attention to their `__init__()` methods.)

pandrey-fr:
Thank you for clarifying things about old models' reloading; it is now fixed.

I also deployed the parameter to `Doc2Vec` and `FastText` (both exposing it as part of their Python API and implementing similar tweaks to the backend pyx files as for word2vec). Otherwise the behaviour would have stayed coherent with the legacy `shrink_windows=True`, but it does feel right to have it properly implemented by children models!

On my Linux machine all tests passed, although I am wondering whether it would be appropriate to tweak those as well, so that some test models may also run with the non-default `shrink_windows=False` parameter? Apart from the code compiling I did check manually that it worked for Word2Vec but not for the other models. Not that I have actual doubts at this point about the implementation working (although I may of course be mistaken), but perhaps adding tests would be suited for future maintainability.
Do you have any opinion on that matter, and if you do want tests to be added, could you please tell me what you would expect? The simplest seems to be to add specific tests just to check that instantiation/training runs with the non-default parameter value, but I can't think of a simple way to additionally assert that the expected backend behaviour is indeed implemented.

gojomo:
Yes, it'd be good to ensure at least one test method, for each class, tries the non-default `shrink_windows=False` code path. 

If there were an easy way to test that it's having the desired effect, that'd be great - but it's a subtle thing to test. Its effects on final vectors aren't easily characterizable. The tangible external effect will be: training runs about twice as long. But, running some training twice, once each way, timing it, and asserting one was "around" twice as long seems a bit flaky & low-value to me, compared to the amount of runtime it'd add to the automated testing. 

Because of that, I think careful review of the source for intent, along with the old path definitely still working as before, and the new path not erroring, is enough for the unit-testing of this.

pandrey-fr:
Alright, I just pushed three additional commits:

(1) Backend changes; as it turns out, the previously-suggested use of `c.reduced_windows[:] = 0` instead of a for loop is not working (it compiled but failed); I also tried using `c.reduced_windows.fill(0)` but it turns out that in the pyx files, the arrays are treated as lists instead of numpy arrays, so either approach failed. I switched back to for loops; it adds a few lines of (readable) code but it does work.

(2) Additional tests, following @gojomo's recommendations. I actually added 4 tests to both Word2Vec, Doc2Vec and FastText so as to test all underlying code branches (cbow and skipgram, on an iterable corpus or on a corpus file). If you would rather remove some of these to limit the extra testing time, feel free to trim some off.

(3) As you suggested, I added the "experimental" status line to the parameter's docstring in all three models.

On the side, I note that:
* Most Doc2vec tests labeled to be running from a corpus file look like they actually use `list_corpus` instead. I might be wrong, but it might be worth taking a look and fixing things (in a separate PR) if needed.
* FastText tests could use some refactoring. To avoid copy/pasting long blocks of redundant code, I tweaked some of the existing ones to simply call them a second time passing `shrink_window=False` in the ones that this PR adds.

piskvorky:
> I switched back to for loops

This is desirable, thanks. The for-loop is clearer and I see you also made it more efficient (less looping).

gensim/models/doc2vec_inner.pyx
```diff
@@ -364,12 +364,13 @@ def train_document_dbow(model, doc_words, doctag_indexes, alpha, work=None,
     c.document_len = i
 
     if c.train_words:
+        # single randint() call avoids a big thread-synchronization slowdown
         if model.shrink_windows:
-            # single randint() call avoids a big thread-synchronization slowdown
             for i, item in enumerate(model.random.randint(0, c.window, c.document_len)):
                 c.reduced_windows[i] = item
         else:
-            c.reduced_windows[:] = int(0)
```
gojomo: Sorry if I had previously suggested a syntax that didn't work or didn't make sense (as the underlying structure is MAX_DOCUMENT_LENGTH long) here!

pandrey-fr: No problem, the suggestion made sense as we could have expected the object to be a proper numpy array :)
Then it's just good to see that unit tests are indeed useful!

gensim/models/doc2vec_inner.pyx
```diff
@@ -365,8 +365,12 @@ def train_document_dbow(model, doc_words, doctag_indexes, alpha, work=None,
 
     if c.train_words:
         # single randint() call avoids a big thread-synchronization slowdown
-        for i, item in enumerate(model.random.randint(0, c.window, c.document_len)):
-            c.reduced_windows[i] = item
+        if model.shrink_windows:
```
mpenkov, line 368: This is a bit repetitive.

Wouldn't it be better to define a model.initialize_reduced_windows method that does this, and call it when needed?

```python
def initialize_reduced_windows(self, ...):
  if model.shrink_windows:
    for i, item in ...:
        ...
  else:
    ...
```

That way, if we find a logic error in the initialization code, we don't have to remember to fix it in a half a dozen other places (e.g. https://github.com/RaRe-Technologies/gensim/pull/3169/files#r649243075).

We should probably also do a bounds check on c.reduced_windows to ensure that it's at least document_len elements long.

gensim/models/word2vec.py
```diff
@@ -941,20 +950,26 @@ def _do_train_job(self, sentences, alpha, inits):
         work, neu1 = inits
         tally = 0
         if self.sg:
-            tally += train_batch_sg(self, sentences, alpha, work, self.compute_loss)
+            tally += train_batch_sg(
```
mpenkov, line 953: This previously fit on one line. What is the benefit of this change?

Same thing for the below.

gensim/models/word2vec.py
```diff
@@ -941,20 +950,26 @@ def _do_train_job(self, sentences, alpha, inits):
...
     def _clear_post_train(self):
         """Clear any cached values that training may have invalidated."""
         self.wv.norms = None
 
     def train(
-            self, corpus_iterable=None, corpus_file=None, total_examples=None, total_words=None,
-            epochs=None, start_alpha=None, end_alpha=None, word_count=0,
-            queue_factor=2, report_delay=1.0, compute_loss=False, callbacks=(),
-            **kwargs,
+            self, corpus_iterable=None, corpus_file=None, total_examples=None,
```
mpenkov, line 969: If this change is purely cosmetic, please roll it back.

gensim/test/test_fasttext.py
```diff
@@ -533,6 +533,18 @@ def test_sg_hs_training_fromfile(self):
                 overlap_count, 2,
                 "only %i overlap in expected %s & actual %s" % (overlap_count, expected_sims_words, sims_gensim_words))
 
+    def test_cbow_hs_training_fixedwindowsize(self):
```
mpenkov, line 536: You can avoid this repetitiveness with parameterized tests (see my suggestion above).

gensim/test/test_fasttext.py
```diff
@@ -497,12 +497,12 @@ def test_sg_hs_training(self):
             overlap_count, 2,
             "only %i overlap in expected %s & actual %s" % (overlap_count, expected_sims_words, sims_gensim_words))
 
-    def test_sg_hs_training_fromfile(self):
+    def test_sg_hs_training_fromfile(self, shrink_windows=True):
```
mpenkov, line 500: https://docs.pytest.org/en/6.2.x/example/parametrize.html

```suggestion
    @pytest.mark.parametrize('shrink_windows', [True, False])
    def test_sg_hs_training_fromfile(self, shrink_windows):
```

mpenkov (changes requested):
Looking good! I'm late to the party, so I see most of the issues have been ironed out. I did manage to find a few nitpicks and left you some comments. Please have a look.

gensim/models/word2vec.py
pandrey-fr: Sorry, this comes from an intermediate version when a parameter had been added, which has then been deleted again. I will restore the previous version (same below).

gensim/test/test_fasttext.py
pandrey-fr: Nice! I did not know about this pytest feature. I will definitely deploy it to reduce the redundant code.

gensim/models/doc2vec_inner.pyx
pandrey-fr: I agree with you on the potential for refactored code that would ease future maintenance and updates. That being said, I actually built on the existing, repetitive code, so in my humble opinion this could be handled as a distinct PR (this one would add a feature, while a second one would enhance the validated, working code).

I also think that your suggestion of a `model.initialize_reduced_windows` is not entirely suitable here:
If I am not mistaken, in this code, `c` is _not_  `model`: it is a special C structure (of a different class depending on the trained model class: word2vec, doc2vec or fasttext), and the `reduced_windows` PyArray is initialized to have the proper size, so there should not be a need to do a bounds check.
I do not know how we could implement a common method (or function) that would also fill the array with either deterministic or random values; but I do agree with you that it could be worth it.

To be honest I lack experience with cython, so that I do not feel entirely at ease with diving into this refactoring. I am willing to give it a try at some point if you want, but would prefer it to be a distinct effort (and PR) from the current one - partly because Mathis and I are pushing the `shrink_windows` feature due to our needing it for a current research project at work, while contributing to refactoring the code base would be something I would (willingly) do on my time off.

gensim/test/test_fasttext.py
pandrey-fr: I pushed two commits making use of that feature. The first one only applied to the FastText tests, as they were the one you attached this review to. The second one applies to the other two test sets (wor2vec and doc2vec); if you feel like the latter should not have been changed, we can discard this commit.

pandrey-fr: Update: actually, `pytest.mark.parametrize` does not work as expected on methods (although it does on functions). I will try to find a way to make it work, otherwise I will undo the last two commits.

pandrey-fr: Oh, in fact it probably will not work due to the test classes being `unittest.TestCase` subclasses.

https://docs.pytest.org/en/6.2.x/unittest.html
> The following pytest features do not work [in unittest.TestCase subclasses], and probably never will due to different design philosophies: Fixtures (except for autouse fixtures, see below); Parametrization; Custom hooks;

pandrey-fr: I am revoking the last two commits as a consequence.

gensim/models/doc2vec_inner.pyx
gojomo: Alas, there's a lot of pre-existing cut & paste duplication between all these related algorithms & modes that could be refactored. I think that'd be OK to note-and-defer-for-later, as either an issue or a FIXME comment in the code.

gensim/test/test_fasttext.py
mpenkov: Thanks for pointing that out. I didn't know pytest parameterization is incompatible with unittest.TestCase.

Luckily, in this particular case, the unittest.TestCase class isn't doing anything. I'll get parameterization working in a separate commit.

gensim/models/doc2vec_inner.pyx
mpenkov: OK, we can deal with this separately.

mpenkov:
Merged. Thank you @M-Demay !

gensim/test/test_fasttext.py
pandrey-fr: Well, we both learned something :)

PR #3170, opened 2021-06-10
Fixed typo on Levenshtein for warning message at import

M-Demay:
Fixed typo on 'Levenshtein' for warning message at import when Levenshtein is not installed.


piskvorky:
Thanks for the fix, but we got rid of the external python-Levenshtein package altogether in #3146, including this message. So closing here.

PR #3174, opened 2021-06-15
Fix a bug when upgrading phraser from gensim 3.x to 4.0

### LINKED ISSUE #3173: Gensim 4.0 loading Phraser trained from Gensim 3.x
#### Problem description

Gensim 4.0 cannot load Phraser model from Gensim 3.x

```python
gemsim.model.load("saved_phraser.pkl") 
```
Will not be able to load phraser trained in gensim 3.x due to a bug in source code. See below.

#### Versions
This line of code from 4.0.1 (and also current development branch) need to be fixed:
https://github.com/RaRe-Technologies/gensim/blob/4.0.1/gensim/models/phrases.py#L367

```python
model.phrasegrams = {
  str(model.delimiter.join(component), encoding='utf8'): score
  for key, val in phrasegrams.items()
}
```
The existing code will only keep the first phrase. To upgrade and load all phrases, it should be replaced with:

```python
model.phrasegrams = {
  str(model.delimiter.join(key), encoding='utf8'): val
  for key, val in phrasegrams.items()
}
```

emgucv:
Fixes #3173 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Changes
 
 * [#3116](https://github.com/RaRe-Technologies/gensim/pull/3116): Fix bug where saved Phrases model did not load its connector_words, by [@aloknayak29](https://github.com/aloknayak29)
 * [#3136](https://github.com/RaRe-Technologies/gensim/pull/3136): fix indexing error in word2vec_inner.pyx, by [@bluekura](https://github.com/bluekura)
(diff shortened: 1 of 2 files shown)


mpenkov (approved):

mpenkov:
Thank you @emgucv !

piskvorky (approved):

PR #3176, opened 2021-06-18
Eliminate obsolete step parameter from doc2vec infer_vector and similarity_unseen_docs

### LINKED ISSUE #3158: doc2vec's infer_vector has `epochs` and `steps` input parameters - `steps` not in use
Refer to doc2vec.py, `infer_vector` function seems to be using `epochs` for the number of iterations and `steps` is not in used.

However, in the `similarity_unseen_docs` function, `steps` is used when calling the infer_vector function.

image

rock420:
Fix #3158 

Eliminate the step parameter from the function `infer_vector()` , `similarity_unseen_docs()` and corresponding `test` function.  (modification in the code and function documentation)
No other code/docs/examples using the outdated `step` parameter.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -581,13 +581,13 @@ class Doc2Vec(Word2Vec):
         """
         return 60 * len(self.dv) + 140 * len(self.dv)
 
-    def infer_vector(self, doc_words, alpha=None, min_alpha=None, epochs=None, steps=None):
+    def infer_vector(self, doc_words, alpha=None, min_alpha=None, epochs=None):
         """Infer a vector for given post-bulk training document.
 
         Notes
         -----
         Subsequent calls to this function may infer different representations for the same document.
-        For a more stable representation, increase the number of steps to assert a stricket convergence.
+        For a more stable representation, increase the number of epochs to assert a stricket convergence.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


gensim/models/doc2vec.py
```diff
@@ -581,13 +581,13 @@ def estimated_lookup_memory(self):
...
 
-    def infer_vector(self, doc_words, alpha=None, min_alpha=None, epochs=None, steps=None):
+    def infer_vector(self, doc_words, alpha=None, min_alpha=None, epochs=None):
         """Infer a vector for given post-bulk training document.
 
         Notes
         -----
         Subsequent calls to this function may infer different representations for the same document.
-        For a more stable representation, increase the number of steps to assert a stricket convergence.
+        For a more stable representation, increase the number of epochs to assert a stricket convergence.
```
piskvorky, line 590:
```suggestion
        For a more stable representation, increase the number of epochs to assert a stricker convergence.
```

piskvorky (changes requested):

gensim/models/doc2vec.py
rock420: Done the typo fix.

gojomo:
Looks good to me. 

Will raise new exceptions for user code using `steps` as parameter to `infer_vector()` or `similarity_unseen_document()` that was previously using the default silently. 

As the previous behavior (to both accept `steps` and to ignore it) was buggy, I think that's OK in a `4.1` minor or even `4.0.x` bugfix release, with proper mention in release notes – but other possible standards for what counts as a 'breaking change' might disagree.

piskvorky (approved):

mpenkov (approved):
Looks good. Thank you for your contribution!

PR #3180, opened 2021-06-22
Isolate generic preprocessing functions

### LINKED ISSUE #3171: make `remove_stopwords()` behavior more consistent
(triggered by SO question: https://stackoverflow.com/questions/67944732/using-my-own-stopword-list-with-gensim-corpora-textcorpus-textcorpus/67951592#67951592)

Gensim has two `remove_stopwords()` functions with similar, but slightly-different behavior that risks confusing users. 

`gensim.parsing.preprocessing.remove_stopwords` takes a space-delimited string, and always consults the *current* value of `gensim.parsing.preprocessing.STOPWORDS` (including any user reassignments). 

By contrast, `gensim.corpora.textcorpus.remove_stopwords` takes a list-of-tokens, allows the specification of an alternate list-of-stopwords, but if allowed to use its 'default' stopwords, uses only the value of `gensim.parsing.preprocessing.STOPWORDS` that was captured when the function was defined (which could miss later user redefinitions).

Avoiding the reuse of identical function names that take different argument-types would help avoid user/code-reviewer confusion. Also, capturing the value of a variable that might change leads to confusing function-behavior: the token version could just as easily consult `gensim.parsing.preprocessing.STOPWORDS` on each call (as the string version already does).

rock420:
Fixes #3171

PR Description - 
 1. All the preprocessing functions from `gensim.corpora.textcorpus` and `gensim.corpora.lowcorpus` have been moved to `gensim.parsing.preprocessing` module.
 2. Made behavior of both `remve_stopwords` function consistent.
 3. Added test cases for these preprocessing functions.

- [x] check for PEP8
- [x] successfully passed all unit tests


gensim/corpora/textcorpus.py
```diff
@@ -44,93 +44,13 @@
 
 from gensim import interfaces, utils
 from gensim.corpora.dictionary import Dictionary
-from gensim.parsing.preprocessing import STOPWORDS, RE_WHITESPACE
+from gensim.parsing.preprocessing import remove_stopword_tokens, remove_short_tokens, \
```
piskvorky, line 47: Please use this style of line continuation:

```suggestion
from gensim.parsing.preprocessing import (
    remove_stopword_tokens, remove_short_tokens, …,
    …,
)
```

gensim/corpora/textcorpus.py
```diff
@@ -177,12 +97,12 @@ class TextCorpus(interfaces.CorpusABC):
 
     The default preprocessing consists of:
 
-    #. :func:`~gensim.corpora.textcorpus.lower_to_unicode` - lowercase and convert to unicode (assumes utf8 encoding)
+    #. :func:`~gensim.parsing.preprocessing.lower_to_unicode` - lowercase and convert to unicode (assumes utf8 encoding)
     #. :func:`~gensim.utils.deaccent`- deaccent (asciifolding)
-    #. :func:`~gensim.corpora.textcorpus.strip_multiple_whitespaces` - collapse multiple whitespaces into a single one
+    #. :func:`~gensim.parsing.preprocessing.strip_multiple_whitespaces`- collapse multiple whitespaces into a single one
```
piskvorky, line 102: Is this change intentional?

gensim/parsing/preprocessing.py
```diff
@@ -89,8 +92,32 @@ def remove_stopwords(s):
...
 
     """
+    if stopwords is None:
+        stopwords = STOPWORDS
     s = utils.to_unicode(s)
-    return " ".join(w for w in s.split() if w not in STOPWORDS)
+    return " ".join(w for w in s.split() if w not in stopwords)
+
+
+def remove_stopword_tokens(tokens, stopwords=None):
```
piskvorky, line 101: Looks like a duplicate of `remove_stopwords` – do we need both?

piskvorky (changes requested):
Thank you for the cleanup! I left some minor style comments in the review.

gensim/corpora/textcorpus.py
rock420: Yes. As the TextCorpus class now using the `strip_multiple_whitespaces()` function from the `preprocessing` module, I have changed that line in documentation also.

piskvorky: I mean the deleted space.

piskvorky:
@mpenkov how do you move tickets and PRs to your columns, in https://github.com/RaRe-Technologies/gensim/projects/9?

gensim/parsing/preprocessing.py
rock420: Yes both the functions are quite similar. The only difference is that `remove_stopwords()` take a string as an argument and also return string while the `remove_stopword_tokens()` take a list of tokens as argument and return a list.  (same thing for `strip_short()` and `remove_short_tokens()`)

`remove_stopwords()` is what widely used across the gensim package. But `remove_stopword_tokens()` is used only in the `textcorpus` module as it first tokenizes the text and then using preprocessing filters.

Though we can remove these duplicate functions by doing some minor changes (doing the preprocessing filters first and then use the tokenizer) in the `Textcorpus` class.

piskvorky: Ah sorry, I completely missed that difference.

It's probably OK then, if the signatures differ. Does it make sense to implement one in terms of the other? To keep it at least a little DRY.

gensim/corpora/textcorpus.py
rock420: sorry, I got it wrong.

Yes, I had changed it intentionally. It was giving PEP8 error -
`E501 line too long (121 > 120 characters)`

Should I keep that space?

piskvorky: Of course.

gensim/parsing/preprocessing.py
rock420: >Does it make sense to implement one in terms of the other? 

Yes, we can do that to minimize the repetition.

gensim/corpora/textcorpus.py
rock420: Done

gensim/corpora/textcorpus.py
rock420: Done the fixing as per suggestion.

piskvorky (approved):
Thanks a lot!

mpenkov:
> @mpenkov how do you move tickets and PRs to your columns, in https://github.com/RaRe-Technologies/gensim/projects/9?

Sorry for the late reply. First, add the ticket to the project (see Projects on the right side of this PR screen, next to milestones, etc.) and then select a column to add to.

mpenkov:
@rock420 Thank you for cleaning this up! I've merged your changes.

@piskvorky Do we want to include this in the change log for 4.1.0? If yes, please edit the CHANGELOG.md on develop HEAD.

piskvorky:
> @piskvorky Do we want to include this in the change log for 4.1.0? If yes, please edit the CHANGELOG.md on develop HEAD.

Yes. Let's do it as part of the release, to avoid editing `develop` directly:

```

### :warning: Removed functionality & deprecations

* [#3189](https://github.com/RaRe-Technologies/gensim/pull/3180): Move  preprocessing functions from gensim.corpora.textcorpus and gensim.corpora.lowcorpus to gensim.parsing.preprocessing, by [@rock420](https://github.com/rock420)
```

PR #3182, opened 2021-06-23
Fix error message when Doc2Vec does not receive corpus_file or corpus iterable

blainedietrich:
The error message thrown when you do not include `corpus_file` or `corpus_iterable` to the doc2vec model is incorrect, stating that you can set the `documents` value.  This change simply makes the error message more accurate so others going through a gensim 3.x to 4.0 migration don't run into issues if they were previously using `documents`.

### DIFF
--- a/gensim/models/doc2vec.py
+++ b/gensim/models/doc2vec.py
@@ -487,7 +487,7 @@ class Doc2Vec(Word2Vec):
 
         """
         if corpus_file is None and corpus_iterable is None:
-            raise TypeError("Either one of corpus_file or documents value must be provided")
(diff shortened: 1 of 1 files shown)


piskvorky (approved):
Thank you! Are there any other places that still refer to `documents`?

blainedietrich:
> Thank you! Are there any other places that still refer to `documents`?

Of course!  My team didn't notice any other instances of `documents`, this message was the only one giving us issues.

piskvorky:
@mpenkov good to merge or not?

mpenkov (approved):

mpenkov:
Merging. Thank you for your patience!

Issue #3015, opened 2020-12-21
Add convenience `get_sentence_vector()`-like methods for FastText, other models


gojomo:
Per <https://stackoverflow.com/questions/65397810/whats-the-equivalent-to-get-sentence-vector-for-gensims-fasttext>, the official Python FastText wrapper offers a `get_sentence_vector()` convenience method to averaging the word-vectors of a text. 

Gensim could offer something similar, for FastText & other models. Though, it should perhaps have a more generic name & clear docs that this is just one simple way to create a text-vector from a bunch of words.

Rlopezra:
I like to take a shot on this issue.

piskvorky:
@Rlopezra go for it. Please check the https://github.com/RaRe-Technologies/gensim/wiki/Developer-page first, and let us know how your PR evolves.

gojomo:
@Rlopezra If you are note a regular user of Gensim's word-vector models, please review other relevant issues/discussion-threads/questions to get an idea of what users want this for.

Rlopezra:
@gojomo I'll have an idea of what's being asked but I'll read on  relevant issues/discussion-threads/questions to get a better understanding.

AdityaSoni19031997:
Is this still being considered as a wishlist? I can try to spend some time to fix this;
Thanks.

gojomo:
Yes.

rock420:
I can take this issue if no one is working on it.
>  please review other relevant issues/discussion-threads/questions to get an idea of what users want this for.

Some observations why users want this feature (in the context of finding document similarity)  - 
1.  While doc2vec is a good option for this but in the case of small corpus or shorter documents, it becomes hard for doc2vec algorithm to generate some meaningful embedding for docs. Averaging the word2vec of words in the doc outperform in some cases like that.
2.  Availability of various pre-trained word2vec models, makes it easier to get some useful embedding for a doc by just using the avg of word vectors and in some cases outperform a doc2vec model trained on the corpus.

I think, it will be good if gensim can also provide an option for `TF-IDF` weighted avg vector with this function, as it works better by reducing the influence of common words.

gojomo:
I like the ideo of providing all those options, so long as the method names & doc-comments properly set expectations. 

[...]

PR #3188, opened 2021-07-06
Add get_sentence_vector() to FastText and get_mean_vector() to KeyedVectors

rock420:
Fixes: #3015 

This PR provides gensim support for `get_sentence_vector()` method. It also implements `get_mean_vector()` under `keyedVectors` as a general method to get the average wordvecs from a list of keys.

Tasks to complete - 
- [x] Implement `get_mean_vector()` under `keyedVectors`.
- [x] Implement `get_sentence_vector()` under `FastTextKeyedVectors`.
- [x] Refactor any existing method which performs an average itself.


rock420:
```py
all_keys, mean = set(), []
for key, weight in positive + negative:
   if isinstance(key, ndarray):
      mean.append(weight * key)
   else:
      mean.append(weight * self.get_vector(key, norm=True))
      if self.has_index_for(key):
          all_keys.add(self.get_index(key))
```

@gojomo, @piskvorky I am trying to refactor the `most_similar` function. The `key` in this function can be a numpy array or str or int but in the current implementation of `get_mean_vector`, it can support only str or int. 
Do you also want to support a list of numpy arrays in get_mean_vector() or use the function only in the case of str, int key in most_similar()?

piskvorky:
For existing functions, let's keep backward compatibility = support the same input types.

For new functions, we have freedom. I have no strong preference either way. Supporting the same input types makes sense (consistency) but if it's too ugly / too much work, we can simplify the function signature.

rock420:
> Supporting the same input types makes sense (consistency) 

 Yeah, we can do this by some simple changes in `get_mean_vector`.

rock420:
@piskvorky, I have updated the PR. 
But some checks are failing because of an error while updating sbt in Linux.

piskvorky:
Something to do with `sbt`. I've seen the error before, but not sure what it's about. @mpenkov do you know why we need scala at all?

I restarted the jobs now, I've seen that help previously.

mpenkov:
Not sure. I've merged develop into this PR, let's see if that helps.

gensim/models/fasttext.py
```diff
@@ -1059,7 +1059,10 @@ def __contains__(self, word):
             False
 
         """
-        return True
+        if self.bucket == 0:  # check for the case when char ngrams not used
```
gojomo, line 1062: The above doc-comment 'note' should be updated to reflect this new behavior & its rationale.

gensim/models/keyedvectors.py
```diff
@@ -746,27 +806,28 @@ def most_similar(
...
+            else:
+                keys.append(item[0])
+                weight.append(item[1])
+        for item in negative:
+            if isinstance(item, KEY_TYPES + (ndarray,)):
+                keys.append(item)
+                weight.append(-1.0)
+            else:
+                keys.append(item[0])
+                weight.append(item[1])
```
gojomo, line 823: This seems a bit verbose (& may need further refactoring to gel with PR #3000 also pending).  It also looks like it's expending effort to retain an (undocumented) option of specyfing the `postiive` or `negative` list with `(key, weight)` tuples - which may be interesting, and even in use somewhere, but if it's a live/maintained capability should be mentioned in doc-comment. 

Maybe the two loops can be merged by concatenating `positive` & `negative` earlier? Maybe the repetitive branches can be minimized by pre-composing `weights`?

gensim/models/keyedvectors.py
```diff
@@ -746,27 +806,28 @@ def most_similar(
...
-                mean.append(weight * key)
-            else:
-                mean.append(weight * self.get_vector(key, norm=True))
-                if self.has_index_for(key):
-                    all_keys.add(self.get_index(key))
-        if not mean:
-            raise ValueError("cannot compute similarity with no input")
-        mean = matutils.unitvec(array(mean).mean(axis=0)).astype(REAL)
+        mean = self.get_mean_vector(keys, weight, pre_normalize=True, ignore_missing=False)
+        mean = matutils.unitvec(mean).astype(REAL)
```
gojomo, line 827: If `get_mean_vector()` had a `norm` option for (post-)norming of its result, this line (& one below) could be subsumed above.

gensim/models/keyedvectors.py
```diff
@@ -1035,7 +1096,7 @@ def rank_by_centrality(self, words, use_norm=True):
         if not used_words:
             raise ValueError("cannot select a word from an empty list")
         vectors = vstack([self.get_vector(word, norm=use_norm) for word in used_words]).astype(REAL)
-        mean = matutils.unitvec(vectors.mean(axis=0)).astype(REAL)
+        mean = matutils.unitvec(self.get_mean_vector(vectors)).astype(REAL)
```
gojomo, line 1099: Another norming that might make sense to pull into a `get_mean_vector()` option.

gensim/models/keyedvectors.py
rock420: We can concatenate the `positive` and `negative` to reduce for loop, but I am not sure how to minimize the if, else branches as we need to check for every element if it's of type `KEY_TYPES` or `tuple`.

rock420: I am thinking of refactoring the above one like this to reduce one `for` loop. Is there any other way I can further improve this one?
```py
keys = []
weight = np.concatenate((np.ones(len(positive)), -1.0*np.ones(len(negative))))
for idx,item in enumerate(positive+negative):
    if isinstance(item, KEY_TYPES + (ndarray,)):
        keys.append(item)
    else:
        keys.append(item[0])
        weight[idx] = item[1]
```

piskvorky:
@rock420 the related PRs have been merged. We're planning a new release soon, I'd love to get this merged – can you please merge `develop` into your PR & resolve conflicts?

@gojomo OK to merge from your side?

gensim/models/fasttext.py
piskvorky: @rock420 how about:

```
This method **always** returns True with char ngrams, because of the way FastText works.
```

rock420: Updated the doc-comment

rock420:
@piskvorky, I have rebased the branch with `develop` and resolved all the conflicts.

gojomo (approved):

piskvorky:
LGTM – @mpenkov did you figure out why some tests are failing? I'd like to get a "green light" before merging. Thanks.

mpenkov:
Merging. Thank you for your contribution and your patience @rock420 !

stelmath:
If I may make an observation: as a user of Facebook's FastText library, I was looking for the equivalent of Facebook's `get_sentence_vector()` in Gensim. This brought me to this PR. However, I spotted a difference in the type of input between Gensim's and Facebook's implementation that might be lurking and confusing for users that are primed with using Facebook's implementation. 

Here is what I mean:
Facebook's [get_sentence_vector()](https://github.com/facebookresearch/fastText/blob/a20c0d27cd0ee88a25ea0433b7f03038cd728459/python/fasttext_module/fasttext/FastText.py#L123) docstring:

>         Given a string, get a single vector represenation. This function
>         assumes to be given a single line of text. We split words on
>         whitespace (space, newline, tab, vertical tab) and the control
>         characters carriage return, formfeed and the null character.

Gensim's [get_mean_vector()](https://github.com/RaRe-Technologies/gensim/blob/eeb7e8662d5350efe68fa14db08b02d273735af9/gensim/models/keyedvectors.py#L461) docstring (`get_sentece_vector()` is based on `get_mean_vector()`):

> Get the mean vector for a given list of keys.
>         Parameters
>         ----------
>         keys : list of (str or int or ndarray)
>             Keys specified by string or int ids or numpy array.

So, when I first tried out Gensim's `get_sentence_vector()` I did not dive into the function definition, and I naively assumed it would expect the same input as Facebook's `get_sentence_vector()`, since Gensim's implementation was inspired by Facebook's implementation. So, I passed something like `"A sample sentence"` into Gensim's `get_sentence_vector()` which results in calling `self.get_vector()` on each **character** of my input string, instead of each **token** of it. Perhaps some type hinting at the function definition would help catch this inconsistency. Thank you!

piskvorky:
We definitely want to accept tokens – like everywhere else in Gensim.

+1 on providing a fat warning when users provide a string where a list of strings was expected. IIRC we already do such "input sanity checks" in other places, so should be straightforward to extend them to `get_sentence_vector()` also.

PR #3141, opened 2021-05-13
Update link for online LDA paper

dymil:
The old link was broken, as Blei moved from Princeton to Columbia. The code link is also broken, but, alas, I don't know anywhere it's still alive.

PR #3190, opened 2021-07-11
Fix broken external link for LDA implementation

ahaya3776:
The old link was broken since Blei moved from Princeton to Columbia, as mentioned in #3141. It seems there was LDA implementation in C, and I fixed the link.


piskvorky:
Thanks!

I'm not a fan of updating our docs every time someone changes departments though. Can you think of a better way to handle this?

Google shows me https://github.com/blei-lab/lda-c . Looks semi-official… maybe we can link to that?

ahaya3776:
Thanks for the comment, I fixed the link to be more permanent.

docs/src/auto_examples/core/sg_execution_times.rst
```diff
@@ -5,10 +5,10 @@
 
 Computation times
 =================
-**00:03.242** total execution time for **auto_examples_core** files:
+**02:36.161** total execution time for **auto_examples_core** files:
 
 +--------------------------------------------------------------------------------------------------------------+-----------+---------+
-| :ref:`sphx_glr_auto_examples_core_run_corpora_and_vector_spaces.py` (``run_corpora_and_vector_spaces.py``)   | 00:03.242 | 48.2 MB |
+| :ref:`sphx_glr_auto_examples_core_run_corpora_and_vector_spaces.py` (``run_corpora_and_vector_spaces.py``)   | 02:36.161 | 40.1 MB |
```
piskvorky, line 11: This is weird. From 3 seconds to 2.5 minutes? What happened?

mpenkov: I'm not sure. I ran the build again on my linux dev machine (the previous one was on MacOS) and the time is more appropriate (5s).

mpenkov:
Merged. Thank you @ahaya3776 !

Issue #3181, opened 2021-06-22
Mismatch get_coherence_per_topic and get_coherence for single topic


silviatti:
#### Problem description

Hi! I am using Gensim to compute the NPMI coherence for each of my topics. I used the method `get_coherence_per_topic()` and also `get_coherence()` (in this case, just passing a list with a single topic), and I noticed that the coherences per topic do not match with the ones returned by `get_coherence()` of the corresponding topics. In my understanding, the NPMI of a topic should be independent of the number of topics or of the other input topics. 
This happens also with the other c_* coherences, not with the UMASS version. 

Thank you!

#### Steps/code/corpus to reproduce

```python
from gensim.test.utils import common_texts, common_dictionary
from gensim.models.ldamodel import LdaModel
from gensim.models.coherencemodel import CoherenceModel

topics = [
    ['human', 'computer', 'system', 'interface'],
    ['graph', 'minors', 'trees', 'eps']
]

cm = CoherenceModel(topics=topics, texts=common_texts, coherence='c_npmi', 
                    dictionary=common_dictionary)
coherence = cm.get_coherence_per_topic()  
print(coherence) # got [0.23583958321789514, -0.24456941091456053]

cm_topic0 = CoherenceModel(topics=[topics[0]], texts=common_texts, 
                           coherence='c_npmi', dictionary=common_dictionary)
coherence_topic0 = cm_topic0.get_coherence()  
print(coherence_topic0) # expect this to be == coherence[0] but got -0.14624062517782566

cm_topic1 = CoherenceModel(topics=[topics[1]], texts=common_texts, 
                           coherence='c_npmi', dictionary=common_dictionary)
coherence_topic1 = cm_topic1.get_coherence()  
print(coherence_topic1) # expect this to be == coherence[1] but got -0.31633310918174923

```

#### Versions

Linux-5.4.104+-x86_64-with-Ubuntu-18.04-bionic
Python 3.7.10 (default, May  3 2021, 02:48:31) 
[GCC 7.5.0]
Bits 64
NumPy 1.19.5
SciPy 1.4.1
gensim 3.8.3
FAST_VERSION 1

PR #3197, opened 2021-07-21
Fix computation of topic coherence

silviatti:
Fixes ##3181.

**Problem description and how I solved it**

Before this PR, the accumulator used for the computation of topic coherence discarded the documents that do not contain words that belong to the input topics, called "not relevant" ([see in the code](https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/topic_coherence/text_analysis.py#L303)). Although this may seem computationally convenient, it has an impact on the computation of the topic coherence (c_v, c_npmi and c_uci). In fact, the number of documents that the accumulator computes is part of the formula for the computation of the direct confirmation measures. 

As a result, discarding "irrelevant" documents makes the resulting topic coherence **not comparable** with the coherences computed over topics with different words (this could result in a different number of total documents). To solve this, I consider all the documents of the corpus in the counting of the number of documents. This PR should solve issue #3181.

**Motivating example**

I'll explain the fairness issue with an example, for the sake of simplicity. Suppose we have two topic models A and B. Each topic model outputs two topics, the first topic of each model (A1 and B1) happens to be the same. We want to compare their coherences in a fair way. We expect the coherence of A1 and B2 to be the same.  

Topic A1: ['human', 'computer']
Topic A2: ['system', 'interface']

Topic B1: ['human', 'computer']
Topic B2: ['graph', 'minors']

Corpus of documents to compute the occurrences: 
d1: 'human computer system'
d2: 'human interface'
d3: 'graph minors trees'

_Before the PR:_
- Considering topic model A: 
Document d3 ('graph minors trees') is discarded by the accumulator because it doesn't contain words that are in the top-words of A1 or A2. So the total number of documents (num_docs) is 2. Let us compute the probability of the occurrence of the word "human" (this should be part of the direct confirmation measure. I just show the example of "human" but could be applied to other words. It is just to show that the results are different even if they shouldn't be). P(human) would be:
`P(human)= number of occurrences of "human" / number of total "relevant" docs = 2/2`
	
- Considering topic model B:
Now document d3 is not discarded in the counts because it contains words that are in the top-words of B2, so the total number of documents is 3. Then P(human) is:
`P(human) = number of occurrences of "human" / number of total "relevant" docs = 2/3`
	
P(human), along with all the other word probabilities and co-occurrence probabilities, will be then used for the computation of topic coherence (in particular, the direct confirmation measure). We will then get two different coherences for the topics A1 and B1, even if they are the same topic in practice. This shouldn't be expected. The coherence of the same topic should be the same. 

Code for comparing the coherences of A1 and B1:

```python 
from gensim.models.coherencemodel import CoherenceModel
from gensim.corpora.dictionary import Dictionary

topics_A = [['human', 'computer'], ['system', 'interface']]
topics_B = [['human', 'computer'], ['graph', 'minors']]

corpus = ['human computer system', 'human interface', 'graph minors trees']
texts = [doc.split() for doc in corpus]
dictionary = Dictionary(texts)

cm_A = CoherenceModel(topics=topics_A, texts=texts, coherence='c_npmi', dictionary=dictionary)
coherence_A = cm_A.get_coherence_per_topic()
print(coherence_A[0]) #coherence of topic A1 is 2.885326251991536e-12

cm_B = CoherenceModel(topics=topics_B, texts=texts, coherence='c_npmi', dictionary=dictionary)
coherence_B = cm_B.get_coherence_per_topic()
print(coherence_B[0]) #coherence of topic B1 is 0.3690702464322811
```

_After the PR:_
- Considering topic model A: 
Document d3 is not discarded in the countings because all the documents are considered, so P(human) is:
`P(human) = number of occurrences of "human" / number of total docs = 2/3`
	
- Considering topic model B:
Document d3 is not discarded, so P(human) is:
`P(human) = number of occurrences of "human" / number of total docs = 2/3`

Now the coherences of A1 and B1 are the same (same code above to test it).


silviatti:
Hello, do you have any plans to consider this PR and the related issue? Many people use gensim to estimate the coherence of the topics and I believe it's important to get accurate and reliable scores.

Thanks :)
@piskvorky

piskvorky:
Thanks for the clear and articulate PR!

Sorry we've been swamped with life, and to be honest I forgot about this PR. It's great you pinged us!

I'll try to review. CC @mpenkov WDYT?

gensim/topic_coherence/text_analysis.py
```diff
@@ -300,18 +300,10 @@ def accumulate(self, texts, window_size):
...
-                    for w in text), dtype=dtype, count=len(text))
-
-    def text_is_relevant(self, text):
-        """Check if the text has any relevant words."""
-        for word in text:
-            if word in self.relevant_words:
-                return True
-        return False
+            yield np.fromiter((
+                self.id2contiguous[self.token2id[w]] if w in self.relevant_words
```
piskvorky, line 304: What happens when there are no relevant words in the text at all? (Why was the `text_is_relevant` check there in the first place?)

Technical note: can you please split up this statement, it is hard to read. E.g. let's give the inner generator its own named variable:

```python
ids = (self.id2contiguous[self.token2id[w]] if w in self.relevant_words else self._none_token for w in text)
yield np.fromiter(ids, …)
```

I realize the ugliness comes from the original code (that you just copied), but it's externally contributed code so let's clean it up while we're at it. Thanks!

piskvorky: Actually, looking at that whole `fromiter()` construct – what on earth is that? Why not simply:

```python
ids = [self.id2contiguous[self.token2id[w]] if w in self.relevant_words else self._none_token for w in text]
yield np.array(ids, dtype=dtype)
```

piskvorky (changes requested):
I don't really understand the `relevant_words` business, and the existing code looks a bit shady to me overall.

So I only did a surface review – as long as your changes don't break the corner case you're replacing ("no `relevant_words` at all"), and helps your cause, I think were good to merge. Thanks!

gensim/topic_coherence/text_analysis.py
aaronjbecker: @piskvorky np.fromiter() seems to be faster than allocating the entire list and calling np.array(). To investigate, I set up [this Jupyter notebook](https://gist.github.com/aaronjbecker/362c1b88368bccd501a077e0515542f1) to simulate the use case in this text_analysis module and found that using np.fromiter with a generator instead of np.array with a list results in about a ~10-15% speedup. 

The relative performance gain increases with document count and decreases with document length (proportional to share of total work devoted to array creation). I feel like a 10-15% runtime difference is noticeable for coherence evaluation on large corpora?

Readability can still be improved--you can bind the ids variable to a generator expression instead of putting the generator directly in the fromiter call w/o any performance impact. I think a compromise makes sense here--create a separate ids variable for clarity, as a generator, but keep np.fromiter for speed?

@silviatti sorry to hijack your commit thread, wanted to help move this along if possible. I'm curious to see whether this change resolves a difference in Gensim's coherence scores vs. another topic modeling library that I use...

piskvorky: Personally I'd go with readability instead of a 15% speed increase. But it's a minor point, both constructs seem fine.

@silviatti can you answer my questions above? We're planning a new release soon, I'd love to get this clean-up merged in. Thanks.

piskvorky: @silviatti ping.

silviatti:
Hello,  
I improved the readability of the code by splitting line 303 into two, still preserving `np.fromiter()`. 

If a topic is composed of only words that are not in the dictionary, that section of code is not even reached. In fact, CoherenceModel(topics, ...) would raise an Exception (so before the computation of the coherence). I added an additional test for this case. 

Apologize for the late reply. I will have more time to work on this issue in the next days, in case you request other changes.

PR #3230, opened 2021-09-11
Adding lifecycle configuration

mpenkov:
You may have noticed that the gensim-wheels S3 bucket is starting to fill up: http://gensim-wheels.s3-website-us-east-1.amazonaws.com/

Here's some bucket config to delete wheels older than 90 days, to prevent cost buildup.

### DIFF
--- a/continuous_integration/BucketLifecycleConfiguration.json
+++ b/continuous_integration/BucketLifecycleConfiguration.json
@@ -1,0 +1,10 @@
+{
+  "Rules": [
+    {
+      "Expiration": {"Days": 30},
+      "Filter": {"Prefix": ""},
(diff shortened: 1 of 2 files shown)


piskvorky (approved):
Thanks! How does the deployment work, is adding the JSON under `continuous_integration` enough (some github magic)? Or should I do something, as the S3 bucket owner?

mpenkov:
You don't need to do anything, I already set this policy. The command is in the readme file:

    aws --profile smart_open s3api put-bucket-lifecycle-configuration --bucket gensim-wheels --lifecycle-configuration file://continuous_integration/BucketLifecycleConfiguration.json

Issue #3226, opened 2021-09-01
numpy 1.19.2 incompatible with gensim 4.1.0


martinobertoni:
#### Problem description

When importing gensim I get the following error

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mbertoni/software/miniconda3/envs/test/lib/python3.7/site-packages/gensim/__init__.py", line 11, in <module>
    from gensim import parsing, corpora, matutils, interfaces, models, similarities, utils  # noqa:F401
  File "/home/mbertoni/software/miniconda3/envs/test/lib/python3.7/site-packages/gensim/corpora/__init__.py", line 6, in <module>
    from .indexedcorpus import IndexedCorpus  # noqa:F401 must appear before the other classes
  File "/home/mbertoni/software/miniconda3/envs/test/lib/python3.7/site-packages/gensim/corpora/indexedcorpus.py", line 14, in <module>
    from gensim import interfaces, utils
  File "/home/mbertoni/software/miniconda3/envs/test/lib/python3.7/site-packages/gensim/interfaces.py", line 19, in <module>
    from gensim import utils, matutils
  File "/home/mbertoni/software/miniconda3/envs/test/lib/python3.7/site-packages/gensim/matutils.py", line 1024, in <module>
    from gensim._matutils import logsumexp, mean_absolute_difference, dirichlet_expectation
  File "gensim/_matutils.pyx", line 1, in init gensim._matutils
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```

#### Steps/code/corpus to reproduce
```
conda create --name=test python=3.7 -y
conda install -y numpy==1.19.2
pip install gensim
```

#### Versions

Linux-5.11.0-25-generic-x86_64-with-debian-bullseye-sid
Python 3.7.11 (default, Jul 27 2021, 14:32:16) 
[GCC 7.5.0]
Bits 64
NumPy 1.19.2
SciPy 1.7.1

GuillemGSubies:
Same error here, any plans to solve it soon?

subbuvidyasekar:
I also have the same issue.
umap package requires numpy 1.19.2.

It works for me initially. it didnt work from 29th August in importing gensim

piskvorky:
Possible duplicate of #3097.

@mpenkov I thought we fixed this in #3095 (Gensim 4.0.1). Is this some regression?

[...]

PR #3231, opened 2021-09-13
[FIX 3226] Attempt to resolve numpy hell

menshikh-iv:


piskvorky:
> Instead of manual pinning of numpy version everywhere, use oldest-supported-numpy everywhere (on every python and platform)

Wouldn't that preclude us from using any new API functions from numpy? (new = introduced since `oldest-supported-numpy`)

If so, that doesn't sound like a desirable solution.

For example, this PR introduces a regression: https://github.com/RaRe-Technologies/gensim/pull/3220 fixed `default_rng`, this PR removes it again.

gensim/models/word2vec.py
```diff
@@ -1956,8 +1956,7 @@ def _load_specials(self, *args, **kwargs):
         if not hasattr(self.wv, 'vectors_lockf') and hasattr(self.wv, 'vectors'):
             self.wv.vectors_lockf = np.ones(1, dtype=REAL)
         if not hasattr(self, 'random'):
-            # use new instance of numpy's recommended generator/algorithm
-            self.random = np.random.default_rng(seed=self.seed)
+            self.random = np.random.RandomState(self.seed)
```
piskvorky, line 1959: Regresses https://github.com/RaRe-Technologies/gensim/issues/2782.

piskvorky (changes requested):

gensim/models/word2vec.py
menshikh-iv: fixed ab53e62

menshikh-iv:
> Wouldn't that preclude us from using any new API functions from numpy? (new = introduced since oldest-supported-numpy)

If we'll have tests for oldest & newest numpy in CI, these will be detected automatically

.github/workflows/build-wheels.yml
```diff
@@ -162,6 +86,18 @@ jobs:
         install_run ${{ matrix.PLAT }}
         echo ::endgroup::
     #
+    # Guarantee maximum compatibility with older numpy versions and prevent regression of #3226
+    #
+    - name: Build and Install Wheels against oldest-supported-numpy (Multibuild)
```
mpenkov, line 91:
```suggestion
    - name: Test wheel against oldest-supported-numpy (Multibuild)
```

gensim/utils.py
```diff
@@ -57,8 +57,13 @@
 )
 """An exception that gensim code raises when Cython extensions are unavailable."""
 
-#: A default, shared numpy-Generator-based PRNG for any/all uses that don't require seeding
-default_prng = np.random.default_rng()
+
+def default_rng(seed=42):
```
mpenkov, line 61:
```suggestion
#
# Numpy versions older than 1.17.0 use an older method for pseudo-random number generation.
# The older method (RandomState) is available everywhere but has performance issues on some platforms.
# The newer method (default_rng) is preferable, when it is available.
#
def default_rng(seed=42):
```

setup.py
```diff
@@ -315,7 +315,7 @@ def run(self):
     'pandas',
 ]
 
-NUMPY_STR = 'numpy >= 1.17.0'
+NUMPY_STR = 'oldest-supported-numpy'
```
mpenkov, line 318:
```suggestion
#
# The numpy version here is especially significant because it affects our ability to build
# wheels that are backwards-compatible with older numpy versions. If restrict this to
# e.g. numpy>=1.17.0, then our wheel building process becomes significantly more
# complicated. Related issue: #3266.
# 
NUMPY_STR = 'oldest-supported-numpy'
```

gensim/models/word2vec.py
mpenkov: > Wouldn't that preclude us from using any new API functions from numpy? (new = introduced since oldest-supported-numpy)

It doesn't have to. Where we can easily replicate the new API functions, we should do so (as @menshikh-iv has done with gensim.utils.default_rng function). Where we can't (so far, there are no such cases, but there may be in the future), we could do something like:

```python
def np_cool_new_function(*args, **kwargs):
    try:
        fn = np.cool_new_function
    except AttributeError:
        raise RuntimeError("your numpy version is too old, please upgrade")
    else:
        return fn(*args, **kwargs)
```

piskvorky: Perhaps. Although this smells of re-implementing dependency resolution internally, which would be extra hell.

But I guess as long as `setup.py` requires an actual working version of numpy (not just `oldest-supported-numpy`), it should be OK.

But requiring some dependency version in `setup.py`, and then raising RuntimeErrors suggesting that version is too old, would not be OK.

mpenkov: > But I guess as long as setup.py requires an actual working version of numpy (not just oldest-supported-numpy), it should be OK.

The problem with going that way is that it prevents us from building against oldest-supported-numpy. We then have to reinvent _that_ particular wheel in our build system infrastructure.

piskvorky: Let's see if I got this right:

|                    |                      | numpy API (setup.py) |                   |
|--------------------|----------------------|----------------------|-------------------|
|                    |                      | oldest-supported-numpy | 1.17.0            |
| numpy ABI (wheels) | oldest-support-numpy | this #3231                | current `develop` |
|                    | 1.17.0               | ?                    | desired, https://github.com/RaRe-Technologies/gensim/issues/3226#issuecomment-917364050           |

Building against the oldest numpy *we actually support* makes more sense to me than building against oldest-supported-numpy and then throwing RuntimeError because it's too old.

@mpenkov what exactly was the problem with that?

mpenkov: It's not exactly code, it's [build system configuration](https://github.com/RaRe-Technologies/gensim/blob/96add7f315735e9579608111a0d1fc72df8e4e0b/.github/workflows/build-wheels.yml#L24-L108).

We're essentially replicating [this](https://github.com/scipy/oldest-supported-numpy/blob/master/setup.cfg) with slightly different business logic.

piskvorky: OK. Reading the comment in your `.yml` link, that sounds exactly like what we want = the "desired" state. Why does that not work?

mpenkov:
Closing this, because we've moved on.

I also recall a problem with oldest-supported-numpy, but can't remember what it is at this stage.

@menshikh-iv Thank you for your effort and for guiding us through these numpy issues.

menshikh-iv:
:pray:

PR #3236, opened 2021-09-16
Change build NumPy versions for MacOS according to oldest-supported-numpy

PrimozGodec:
Made according to the conversation in: https://github.com/RaRe-Technologies/gensim/issues/3226#issuecomment-920854099

It fixes wrongly set NumPY build versions which prevent Gensim to work on macOS with Numpy <1.20

I tried to build wheels with those versions and it seems it works correclty.


mpenkov (approved):

mpenkov:
Thank you!

PR #3247, opened 2021-10-11
Sparse2Corpus: update __getitem__ to work on slices, lists and ellipsis

PrimozGodec:
Currently, Sparse2Corpus's `__getitems__` only work on integer indexes and return a sparse representation for only one document.

Since the underlying structure is a CSC matrix which allows more options in `__getitems__` I suggest upgrading the current implementation with options supported for the CSC matrix. I suggest that `__getitems__` returns the Sparse2Corpus object with a selected subset of data. Supported subsets are slice, list, ellipsis, int. 

My change includes:
1. More options for selecting part of the data
2. Returning Sparse2Corpus object instead of a manmade list. I can support that decision with the fact that the underlying object (CSC matrix) have the same behaviour (returns the same type). The same holds for NumPy for example. 

If we agree we should also discuss how to deprecate old behaviour and introduce the new one.

This PR is not final. When we agree that the solution is OK I will polish the code and add some tests.


piskvorky:
> I suggest that __getitems__ returns the Sparse2Corpus object with a selected subset of data. Supported subsets are slice, list, ellipsis, int.

Fancy slicing & indexing is an interesting idea, thanks. Can you post some supporting motivation? What made you look at this functionality? What's your use-case?

I'm definitely -1 on breaking existing primary interfaces / API contracts of Gensim though, such as the expectation that a transformed document is a list / plain sequence of `(feature_id, weight)` pairs. Or is your change backward compatible?

mpenkov (approved):

mpenkov:
Merging. Thank you for your effort and your patience!

PrimozGodec:
No worries. Thank you for considering my PR.

PR #3259, opened 2021-10-27
Fixed error message when raising to negative exponent #3232 

### LINKED ISSUE #3232: Negative exponent with value -1 (minus one) raises error
#### Problem description

I try to vary the value of the negative exponent parameter. When I use a value of -1, training works fine, but prediction afterwards with infer_vector does not work and raises the error "ValueError: Integers to negative integer powers are not allowed."

This is due to the following line: https://github.com/RaRe-Technologies/gensim/blob/266a01455ade51a93a08dba5950e87b4d98e0724/gensim/models/word2vec.py#L836

Here, numpy does not raise an integer by the power of another, but negative integer.

I guess this could be solved by converting the exponent to a float in this case?

tchan102:
Fixed #3232 
In gensim/models/word2vec.py, we changed the line , we type-casted ns_exponent in  def make_cum_table(self, domain=2**31 - 1) to a float to allow raising integers to negative exponent.

### DIFF
--- a/gensim/models/word2vec.py
+++ b/gensim/models/word2vec.py
@@ -833,11 +833,11 @@ class Word2Vec(utils.SaveLoad):
         train_words_pow = 0.0
         for word_index in range(vocab_size):
             count = self.wv.get_vecattr(word_index, 'count')
-            train_words_pow += count**self.ns_exponent
+            train_words_pow += count**float(self.ns_exponent)
         cumulative = 0.0
         for word_index in range(vocab_size):
             count = self.wv.get_vecattr(word_index, 'count')
-            cumulative += count**self.ns_exponent
+            cumulative += count**float(self.ns_exponent)
             self.cum_table[word_index] = round(cumulative / train_words_pow * domain)
         if len(self.cum_table) > 0:
             assert self.cum_table[-1] == domain


lelandbatey:
Suggested edits #1; improve the description of the PR.

> Fixes #3232
> Fixed the bug where a using a negative integer exponent within a Doc2Vec model causes the `"ValueError: Integers to negative integer powers are not allowed."` to be raised.  Fixed by always casting the exponent to a float.

README.md
```diff
@@ -43,7 +43,7 @@ Features
 
 If this feature list left you scratching your head, you can first read
 more about the [Vector Space Model] and [unsupervised document analysis]
-on Wikipedia.
+on Wikipedia.
```
lelandbatey, line 46: In the future I'd suggest leaving out, or removing at the end, unnecessary whitespace changes in documentation.

gensim/similarities/fastss.c
```diff
@@ -0,0 +1,8416 @@
+/* Generated by Cython 0.29.23 */
```
lelandbatey, line 1: This file seems like it's an auto-generated file which isn't related to this PR. If it is unrelated, then this file should be removed. If this file is related and required for this PR, then the reasons why should be explained in the description of this PR.

PR #3263, opened 2021-11-07
Remove commented out pytest-rerunfailures test dependency

pabs3:
pytest-rerunfailures is now compatible with pytest 6.1.0 and later.

Reverts: commit 8687e7f68afb9f9251e69039e04280b6ae1e7bdc.
Reported-in: https://github.com/pytest-dev/pytest-rerunfailures/issues/128
Fixed-by: https://github.com/pytest-dev/pytest-rerunfailures/pull/129

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -267,7 +267,6 @@ visdom_req = ['visdom >= 0.1.8, != 0.1.8.7']
 core_testenv = [
     'pytest',
     'pytest-cov',
-#    'pytest-rerunfailures',  # disabled 2020-08-28 for <https://github.com/pytest-dev/pytest-rerunfailures/issues/128>
     'mock',
     'cython',
     'testfixtures',


mpenkov:
What's the benefit of re-running failures?

In my opinion, if a test keeps blinking, it's probably a bad test. We should improve the test so that it does not blink instead of re-running it and hoping that we get lucky on one of the attempts.

@pabs3 What are your thoughts?

pabs3:
I'm not sure about the benefit, I just figured it was used for a reason
before and only got disabled due to a bug somewhere else and that bug
got fixed so now it could get enabled again.

What you say sounds correct to me. If it isn't useful then I can turn
this PR into a commit dropping the commented out dependency.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

mpenkov:
@menshikh-iv @piskvorky What do you guys think? Can we get rid of this and run tests once only?

piskvorky:
Do we still experience flaky tests? How bad is it – if not bad, then no reason to automate "re-runs" IMO.

mpenkov:
I don't think it's bad. One of the wheel builds fails occasionally, but there are ways around that other than automatic re-runs (plus the re-runs don't always help).

menshikh-iv:
If I remember correctly, rerun plugin help us to avoid "blinking" test failures, if we didn't see these tests now - we can just don't use this plugin.

> there are ways around that other than automatic re-runs 
@mpenkov just curious, would you mind to add details?

mpenkov:
In order of decreasing preference, we could:

- Eliminate the external factors that cause the test to blink (are we forgetting to mock or otherwise nail down something?)
- Rewrite the test so that it does not blink (why do these test blink, anyway? Where are we dealing with non-deterministic code?)
- Mark the test as an expected failure under certain conditions
- Disable the test
- Remove the test

gojomo:
Personally, I'd rather we default to *no* automatic rerunning of tests. It slows down detection of reliable failures, & might hide other stability issues better solved another way.

pabs3:
The consensus seems to be to not re-add pytest-rerunfailures, so I have added a commit removing the commented out dependency.

pabs3:
Can someone re-run the checks? Seems like one of them timed out and another I cannot tell why the failure happened.

piskvorky:
The wheel-build failures are likely unrelated. @mpenkov is looking into it – thanks for your PR!
The new release is coming soon now :)

mpenkov:
Merged. Thank you @pabs3 !

PR #3264, opened 2021-11-07
Detect when a fasttext executable is available in PATH

pabs3:
Also check that fasttext exists in FT_HOME and is executable.

This is useful when using a distro like Debian that
has a package of fasttext available to install.

### DIFF
--- a/gensim/test/test_fasttext.py
+++ b/gensim/test/test_fasttext.py
@@ -8,6 +8,7 @@ import io
 import logging
 import unittest
 import os
+import shutil
 import subprocess
 import struct
 import sys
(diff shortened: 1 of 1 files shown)


pabs3:
Can someone re-run the checks? Seems like they timed out.

mpenkov:
@piskvorky I'm going to merge this as-is because we use the facebook executable in multiple places during the tests. It's easier to use the executable than to store half a dozen models as test data.

mpenkov:
@pabs3 Thank you for your work!

gensim/test/test_fasttext.py
```diff
@@ -44,7 +45,8 @@
 BUCKET = 10000
 
 FT_HOME = os.environ.get("FT_HOME")
-FT_CMD = os.path.join(FT_HOME, "fasttext") if FT_HOME else None
+FT_CMD = shutil.which("fasttext", path=FT_HOME) or \
```
piskvorky, line 48: Single line please.

PR #3271, opened 2021-11-26
Added new ValueError in place of assertion error for no model data provided in lsi model

### LINKED ISSUE #3270: Update: assert self.projection.u is not None, "decomposition not initialized yet"
So I am also receiving the error below (one that was previously reported, but closed for lack of information  #2278 ) - but the good news is I have a working minimal example that replicates it. It seems to be more an error handling issue than anything, as it requires you to set the input to dictionary to the empty list (which is invalid syntax).

Error:
assert self.projection.u is not None, "decomposition not initialized yet"

To replicate:
```
from gensim import corpora, models, similarities

texts = []
dictionary = corpora.Dictionary(texts)
corpus = []
lsi = models.LsiModel(corpus, id2word=dictionary, num_topics=2)
index = similarities.MatrixSimilarity(lsi[corpus])
```

mark-todd:
Added warning to lsi model for initialising a model with no data

Fixes #3270.


gensim/models/lsimodel.py
```diff
@@ -489,7 +489,8 @@ def add_documents(self, corpus, chunksize=None, decay=None):
             chunksize = self.chunksize
         if decay is None:
             decay = self.decay
-
+        if corpus == []:
+            warnings.warn('LsiModel.add_documents() called but no documents provided, is this intended?')
```
piskvorky, line 493: We want standard logging (`logger.warn()`).

piskvorky (changes requested):

gensim/models/lsimodel.py
```diff
@@ -489,7 +489,8 @@ def add_documents(self, corpus, chunksize=None, decay=None):
             chunksize = self.chunksize
         if decay is None:
             decay = self.decay
-
+        if corpus == []:
```
piskvorky, line 492:
```suggestion
        if not corpus:
```

piskvorky (changes requested):

gensim/models/lsimodel.py
mark-todd: Perhaps "if len(corpus) < 0" - if not corpus would also catch corpus = False etc (I know, unlikely, but would be very confusing), and corpus=True would pass no issue. Also I think corpus can very feasibly be None which would have the same issue

piskvorky: All these cases ought to result in the same warning IMO.

piskvorky:
@mpenkov could you please check why one of the builds is failing? Thank you!

gensim/models/lsimodel.py
piskvorky: @mark-todd can you fix this? We're going to be releasing a new version of Gensim soon, and this PR could go in if finished.

mark-todd: Ah sure ok - sorry this slipped my mind - when would you need it by?

mark-todd:
I've now corrected the file according to your suggested modifications - hope this is all ok!

gensim/models/lsimodel.py
```diff
@@ -70,7 +70,6 @@
 
 from gensim import interfaces, matutils, utils
 from gensim.models import basemodel
-
```
piskvorky, line 73: Keep the blank line between imports and global assignments please.

piskvorky (changes requested):

gensim/models/lsimodel.py
```diff
@@ -70,7 +70,6 @@
 
 from gensim import interfaces, matutils, utils
 from gensim.models import basemodel
```
piskvorky, line 72:
```suggestion
from gensim.models import basemodel

```

gensim/models/lsimodel.py
```diff
@@ -489,7 +489,8 @@ def add_documents(self, corpus, chunksize=None, decay=None):
             chunksize = self.chunksize
         if decay is None:
             decay = self.decay
-
+        if not corpus:
```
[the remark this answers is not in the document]
mark-todd: So I initially suggested `if corpus == []`, but in a previous revision you suggested `if not corpus` - personally I still think the first option is the safest, but if you have any other ideas I'm open :)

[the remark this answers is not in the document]
piskvorky: Hmm, I probably didn't realize a `scipy.sparse` path was possible :)

But `corpus == []` would not work either – that will fail to detect empty generators, empty scipy.sparse arrays etc.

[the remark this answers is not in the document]
piskvorky: Blah, that will silently eat away one document if `corpus` is a generator. Not good, scratch that.

Checking for emptiness in a general case sounds pretty complicated. How about we allow some false positives:

```
def is_empty(corpus):
    """Is the corpus (an iterable or a scipy.sparse array) empty?"""
    if scipy.sparse.issparse(corpus):
       return corpus.shape[1] > 0  # by convention, scipy.sparse documents are columns
    if isinstance(gen, types.GeneratorType):
       return False  # don't try to guess the content of generators, to avoid reading irretrievable elements
    try:
        first_doc = next(iter(corpus))  # list, numpy array etc
        return True  # first document exists => not empty
    except Exception:
        return False
```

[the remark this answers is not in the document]
mark-todd: This looks ok, but something just occurred to me - does `__len__` have to be defined for any corpus? If not, maybe this would be a good requirement for a later version, as I feel like this would be useful in many places.

If it's defined we can just say:
```py
if len(corpus) == 0:
````
which works certainly for numpy, scipy.sparse I presume, and lists
...also if len has to be defined for generators and iterators

If not, I suppose something like this:
```py
def is_empty(corpus):
     try:
            length = len(corpus)
     except TypeError:
            return False
     return length == 0
```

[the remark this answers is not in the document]
mark-todd: Also, I think the bools in your example are the wrong way round, perhaps this:


```py
def is_empty(corpus):
    """Is the corpus (an iterable or a scipy.sparse array) empty?"""
    if scipy.sparse.issparse(corpus):
       return corpus.shape[1] > 0  # by convention, scipy.sparse documents are columns
    if isinstance(gen, types.GeneratorType):
       return False  # don't try to guess emptiness of generators, may lose elements irretrievably
    try:
        first_doc = next(iter(corpus))  # list, numpy array etc
        return False # first document exists => not empty
    except Exception:
        return True
```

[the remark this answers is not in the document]
piskvorky: No, `len` doesn't work on scipy.sparse, nor on generators.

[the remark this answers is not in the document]
mark-todd: >  No, len doesn't work on scipy.sparse, nor on iterables in general.

Ah shame about scipy.sparse - in terms of generators in general, this is true - I suppose I meant something like this:
```py
from typing import TypeVar, Generator, Generic
YieldType = TypeVar("YieldType")
SendType = TypeVar("SendType")
ReturnType = TypeVar("ReturnType")

class AbstractCorpusGenerator(Generic[YieldType, SendType, ReturnType], Generator[YieldType, SendType, ReturnType]):
      def __len__(self) -> int:
             raise NotImplementedError
```
...but this is well outside the scope of this PR - just a little food for thought :)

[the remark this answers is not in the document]
mark-todd: > No, `len` doesn't work on scipy.sparse, nor on iterables in general.
> 
> > Also, I think the bools in your example are the wrong way round, perhaps this:
> 
> You're right – needs to be switched, in multiple places:
> 
> ```python
> def is_empty(corpus):
>     """Is the corpus (an iterable or a scipy.sparse array) empty?"""
>     if scipy.sparse.issparse(corpus):
>        return corpus.shape[1] == 0  # by convention, scipy.sparse documents are columns
>     if isinstance(gen, types.GeneratorType):
>        return False  # don't try to guess emptiness of generators, may lose elements irretrievably
>     try:
>         first_doc = next(iter(corpus))  # list, numpy array etc
>         return False # first document exists => not empty
>     except Exception:
>         return True
> ```

Also yeah good spot on the other inequality - I think we can go with this version then if you're happy?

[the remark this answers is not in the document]
mark-todd: Oh one final tweak
```py
def is_empty(corpus):
    """Is the corpus (an iterable or a scipy.sparse array) empty?"""
    if scipy.sparse.issparse(corpus):
       return corpus.shape[1] == 0  # by convention, scipy.sparse documents are columns
    if isinstance(gen, types.GeneratorType):
       return False  # don't try to guess emptiness of generators, may lose elements irretrievably
    try:
        next(iter(corpus))  # list, numpy array etc
        return False # first document exists => not empty
    except Exception:
        return True
```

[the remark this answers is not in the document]
piskvorky: Nah, the original was more explicit = self-documenting = clearer.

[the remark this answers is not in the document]
mark-todd: Eh I guess - normally I would avoid assigning variables that aren't used, although the performance is minimal I suppose. 

Also, just realised, any other error than stop iteration (e.g. the corpus cannot be made iterable) should return False - only "StopIteration" tells us it's an empty iterator:

```py
def is_empty(corpus):
    """Is the corpus (an iterable or a scipy.sparse array) empty?"""
    if scipy.sparse.issparse(corpus):
       return corpus.shape[1] == 0  # by convention, scipy.sparse documents are columns
    if isinstance(gen, types.GeneratorType):
       return False  # don't try to guess emptiness of generators, may lose elements irretrievably
    try:
        first_doc = next(iter(corpus))  # list, numpy array etc
        return False # first document exists => not empty
    except StopIteration:
        return True
    except Exception:
        return False
 ```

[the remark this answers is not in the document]
Comment (piskvorky) [reply gensim/models/lsimodel.py:492]: You're right. Much better (safer) this way :)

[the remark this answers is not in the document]
Comment (mark-todd) [reply gensim/models/lsimodel.py:492]: Ok cool - I'll add this to the PR :)

**Comment by mark-todd** [comment]
Not sure why it's still saying one change requested - the import line missing issue should now be resolved

File: gensim/models/lsimodel.py [hunk 6]
```diff
@@ -482,14 +483,28 @@ def add_documents(self, corpus, chunksize=None, decay=None):
         If the distributed mode is on, each chunk is sent to a different worker/computer.
 
         """
+        def is_empty(corpus):
```
piskvorky, line 486: Let's put this in `utils.py`. Rather than an inline (anonymous) function.

piskvorky (changes requested):

gensim/models/lsimodel.py
piskvorky: @mark-todd can you move this function to `utils.py` please?

mark-todd: Yep tis done - sorry about the delay

gensim/utils.py
```diff
@@ -2084,3 +2084,17 @@ def effective_n_jobs(n_jobs):
     elif n_jobs < 0:
         n_jobs = max(multiprocessing.cpu_count() + 1 + n_jobs, 1)
     return n_jobs
+
+def is_empty(corpus):
+    """Is the corpus (an iterable or a scipy.sparse array) empty?"""
+    if scipy.sparse.issparse(corpus):
+        return corpus.shape[1] == 0  # by convention, scipy.sparse documents are columns
+    if isinstance(corpus, types.GeneratorType):
```
piskvorky, line 2092: Undefined `types`?

gensim/utils.py
```diff
@@ -2084,3 +2084,17 @@ def effective_n_jobs(n_jobs):
...
+        return False  # don't try to guess emptiness of generators, may lose elements irretrievably
+    try:
+        first_doc = next(iter(corpus))  # list, numpy array etc
+        return False # first document exists => not empty
```
piskvorky, line 2096:
```suggestion
        return False  # first document exists => not empty
```

gensim/models/lsimodel.py
```diff
@@ -482,14 +483,16 @@ def add_documents(self, corpus, chunksize=None, decay=None):
         If the distributed mode is on, each chunk is sent to a different worker/computer.
 
         """
+
```
piskvorky, line 486:
```suggestion
```

gensim/models/lsimodel.py
```diff
@@ -70,7 +71,7 @@
 
 from gensim import interfaces, matutils, utils
 from gensim.models import basemodel
-
+from gensim.utils import is_empty
```
piskvorky, line 74:
```suggestion
from gensim.utils import is_empty

```

piskvorky (changes requested):
We follow PEP8 & PEP257 in code style: no blank line after docstring; blank line after import group; more than one space before comment etc.

Plus `types` doesn't seem to be defined in `utils.py`. Does this code really work, did you try it?

gensim/utils.py
mark-todd: Yep sorry missed this but have corrected - and yeah I've run the test suite now and seems ok

piskvorky:
Looks good, thanks!

@mpenkov what's the best way to shut up flake8 about this?

Screen Shot 2022-03-18 at 13 03 04

I'd prefer to have the variable there and named – makes the code much easier to read and understand. I don't really care that the variable is unused.

Other than that LGTM.

piskvorky (approved):

mpenkov:
Merging. Thank you for your contribution and your patience @mark-todd !

PR #3273, opened 2021-12-04
fix vector representation in example

sashaaKr:
In `Core Concepts` documentation there is an error for vector representation:
```
For example, assume we have a dictionary containing the words ['coffee', 'milk', 'sugar', 'spoon']. A document consisting of the string "coffee milk coffee" would then be represented by the vector [2, 1, 0, 0] where the entries of the vector are (in order) the occurrences of “coffee”, “milk”, “sugar” and “spoon” in the document
```
Vector [2, 1, 0, 0] is not correct and should be [1, 1, 0, 0],
This pr provide fix for this issue


piskvorky:
What makes you think so?

sashaaKr:
@piskvorky pardon, my bad, for some reason I was thinking that string example is `"coffee milk"`
Closing.

PR #3274, opened 2021-12-04
Migrate setup.py from distutils to setuptools

### LINKED ISSUE #3272: Migrate setup.py from distutils to setuptools
distutils is going away in a future version of Python

https://docs.python.org/3/whatsnew/3.10.html#distutils-deprecated

geojacobm6:
Fixes #3272 

Migrate setup.py from distutils to setuptools

distutils is going away in a future version of Python

https://docs.python.org/3/whatsnew/3.10.html#distutils-deprecated


mpenkov:
Good work, thank you!

PR #3275, opened 2021-12-05
Update word2vec.py

Sangram-Rout:
Added from error     total_words += len(sentence)
TypeError: object of type 'map' has no len()


piskvorky:
Sentences shouldn't be a `map` (?), so this is likely a user error.

PR #3279, opened 2021-12-22
Add the FastSS and Levenshtein modules to docs

piskvorky:
I tried to find our new levenshtein module via google, and failed.

Upon investigation it turns out it's not documented at all (not API pages), so google doesn't even know about.

This PR adds the two missing module doc files, plus adds them to the top-level API ref listing.


piskvorky:
@mpenkov could you please check what's up with the failing "Build wheels" test?

mpenkov (approved):
Looks good, merging.

PR #3299, opened 2022-02-26
Enable test_word2vec_stand_alone_script by using sys.executable for python

pabs3:


gensim/test/test_word2vec.py
```diff
@@ -1162,15 +1164,17 @@ def test_path_line_sentences_one_file(self):
...
-#         cmd = 'python -m gensim.scripts.word2vec_standalone -train ' + datapath('testcorpus.txt') + \
-#               ' -output vec.txt -size 200 -sample 1e-4 -binary 0 -iter 3 -min_count 1'
-#         output = check_output(cmd, stderr=PIPE)
-#         self.assertEqual(output, '0')
-# #endclass TestWord2VecScripts
+
+class TestWord2VecScripts(unittest.TestCase):
+    def test_word2vec_stand_alone_script(self):
+        """Does Word2Vec script launch standalone?"""
+        cmd = [sys.executable, '-m', 'gensim.scripts.word2vec_standalone', '-train', datapath('testcorpus.txt')] + \
```
piskvorky, line 1171:
```suggestion
        cmd = [
            sys.executable, '-m',
            'gensim.scripts.word2vec_standalone', '-train',
            datapath('testcorpus.txt')],
            '-output vec.txt -size 200 -sample 1e-4 -binary 0 -iter 3 -min_count 1'.split(),
        ]
```

piskvorky (changes requested):
Code style.

pabs3:
I note that this newly enabled test is failing, but I have no idea what the error is about, any thoughts?

gensim/test/test_word2vec.py
piskvorky: Sure, thanks.

piskvorky:
@pabs3 did you find out the cause of the `0`-vs-nothing difference?

gensim/test/test_word2vec.py
```diff
@@ -1162,15 +1164,21 @@ def test_path_line_sentences_one_file(self):
...
+            sys.executable, '-m', 'gensim.scripts.word2vec_standalone',
+            '-train', datapath('testcorpus.txt'),
+            '-output', 'vec.txt', '-size', '200', '-sample', '1e-4',
+            '-binary', '0', '-iter', '3', '-min_count', '1',
+        ]
+        output = check_output(args=cmd, stderr=subprocess.PIPE)
+        self.assertEqual(output, '0')
+
+
+# endclass TestWord2VecScripts
```
piskvorky, line 1181: Let's get rid of this too. Such closing-class comments are not really a part of our code style in Gensim.

piskvorky (changes requested):

pabs3:
I haven't found time to investigate the 0-vs-nothing difference, sorry. 

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

gensim/test/test_word2vec.py
pabs3: Removed. Only kept it because it was present before and the test above also has such a comment.

pabs3:
I did some digging just now and I can't find a time when `gensim.scripts.word2vec_standalone` ever output an integer like the test suggests it does. Back in 2016 the test for it was failing and got disabled in commit 055cd7d5 in order to make a release and it was originally added with the check for `0` in PR #538.

pabs3:
I've switched the test to check for `b''` instead of `'0'`.

piskvorky:
Hm, that PR was closed, not merged.

In any case, if you fixed a bug that's great, thanks!

piskvorky (approved):

piskvorky:
Thanks again @pabs3 !

PR #3307, opened 2022-03-18
[MRG] Documentation fixes

### LINKED ISSUE #3260: Documentation should explain how to run tests
The [Testing Gensim](https://radimrehurek.com/gensim/#testing) section of the documentation doesn't explain how to run tests.

Running ```python -m pytest``` fails:
```
============================= test session starts ==============================
platform freebsd13 -- Python 3.8.12, pytest-4.6.11, py-1.9.0, pluggy-0.13.1
rootdir: /disk-samsung/freebsd-ports/textproc/py-gensim/work-py38/gensim-4.1.2
plugins: forked-1.0.2, cov-2.9.0, rerunfailures-10.1, timeout-1.4.2, xdist-1.32.0, mock-1.10.4, hypothesis-6.23.1, asyncio-0.10.0, localserver-0.5.0
collected 0 items / 44 errors

==================================== ERRORS ====================================
_______________ ERROR collecting gensim/test/test_aggregation.py _______________
gensim/matutils.py:1346: in <module>
    from gensim.corpora._mmreader import MmReader  # noqa: F401
E   ModuleNotFoundError: No module named 'gensim.corpora._mmreader'

/usr/local/lib/python3.8/site-packages/py/_path/local.py:704: in pyimport
    __import__(modname)
gensim/__init__.py:11: in <module>
    from gensim import parsing, corpora, matutils, interfaces, models, similarities, utils  # noqa:F401
gensim/corpora/__init__.py:6: in <module>
    from .indexedcorpus import IndexedCorpus  # noqa:F401 must appear before the other classes
gensim/corpora/indexedcorpus.py:14: in <module>
gensim/interfaces.py:19: in <module>
gensim/matutils.py:1348: in <module>
    raise utils.NO_CYTHON
E   RuntimeError: Compiled extensions are unavailable. If you've installed from a package, ask the package maintainer to include compiled extensions. If you're building Gensim from source yourself, install Cython and a C compiler, and then run `python setup.py build_ext --inplace` to retry.
___________________ ERROR collecting gensim/test/test_api.py ___________________
gensim/matutils.py:1346: in <module>
    from gensim.corpora._mmreader import MmReader  # noqa: F401
E   ModuleNotFoundError: No module named 'gensim.corpora._mmreader'

/usr/local/lib/python3.8/site-packages/py/_path/local.py:704: in pyimport
    __import__(modname)
gensim/__init__.py:11: in <module>
    from gensim import parsing, corpora, matutils, interfaces, models, similarities, utils  # noqa:F401
gensim/corpora/__init__.py:6: in <module>
    from .indexedcorpus import IndexedCorpus  # noqa:F401 must appear before the other classes
gensim/corpora/indexedcorpus.py:14: in <module>
gensim/interfaces.py:19: in <module>
gensim/matutils.py:1348: in <module>
    raise utils.NO_CYTHON
```

### LINKED ISSUE #3291: LsiModel ( & maybe others) should generate docs for essential special-methods (like `__getitem__` etc)
Answering someone's LSI question, I noticed that none of the Gensim docs around LSI point out that you need to use bracket-indexed access to get the LSI topics for a BoW doc - not the class doc, or the tutorial. A user would need to know or intuit that's the Gensim convention from elsewhere. 

Without understanding the auto-doc config exactly, I suspect `LsiModel` needs some config file like the one in `poincare.rst`:

https://github.com/RaRe-Technologies/gensim/blob/develop/docs/src/models/poincare.rst

...which enabled doc-generation for special-members. Perhaps, it'd even be appropriate to toggler generation of such docs globally, if possible, given the centrality of the bracket-accessing to some of the Gensim classes' APIs. If that's not possible/wise, a survey of which classes – (a) implement specials that are essential to thei use; & (b) don't yet have doc-comments, or generate web docs for those methods – should be done with the oversights fixed.

piskvorky:
Various pre-release fixes for https://github.com/RaRe-Technologies/gensim/milestone/5:

* Fixes #3291
* Fixes #2809
* Fixes #3260


piskvorky:
@gojomo the doc2vec runs finished:



Unlike in the original notebook, the DBOW results no longer seem superior to DM. Both are reasonable, so I changed the conclusions and wording in the notebook accordingly. Perhaps the previously pronounced difference was due to some bug in the earlier versions of Gensim (the notebook seems quite old).

The next step is to run the notebook front-to-back, commit & merge here.

gojomo:
Looks good overall. My main suggestion would be to add a parallel demonstration of a 3rd mode: pure DBOW (leaving the default `dbow_words=0` in place). 

IIRC, the very idea of mixing DBOW with interleaved (& largely-analogous) skipgram word-training traces back to this paper, in its innocent aside, "We also show a simple yet effective trick to improve Paragraph Vector. In particular, we observe that by jointly training word embeddings, as in the skip gram model, the quality of the paragraph vectors
is improved." In fact, I think I added the `dbow_words` option to make it possible to match this paper. 

But the plain DBOW is also very fast, & very competitive on many tasks, *if* you just want full-doc vectors (without word-vectors). And the fact that the word-vectors still exist in that plain mode, but remain untrained random-initialized junk, is an occasional gotcha that it'd be useful to note. (My old IMDB-based Doc2Vec demo showed the nonsense results, when doing word-vector most-similars on a plain-DBOW model, to highlight this point.)

In particular, the reason your existing DBOW+W vs DM comparison shows a significant speed advantage for DM is entirely the extra word-training. With a corpus of M words, in one epoch, plain DBOW does M innnermost-loop micro-predictions, each `doc_vector -> word`. Plain DM also does M innermost-loop micro-predictions, each `mean(doc_vector, contextword0_vector, ..., contextwordN_vector) -> word`. But DBOW+W does `M + (window * M)` innermost micro-predictions: both the plain DBOW predictions, *plus* an average of `window` more predictions, per target-word, of skip-gram form `word-vector -> word`. 

It's likely that DBOW alone would train fastest, *and* do reasonably well on all the tasks that work only with articles. Even if not adding a "pure DBOW" trial, it'd be good to be clear that what you're trying is the special DBOW plus skip-gram introduced by the 'Document Embeddings' paper, not plain DBOW. 

I notice you're using a non-default `window=8` value – though the paper doesn't seem to specify what value they're using. You might as well leave the default `window=5` in place. That'll make the DM mode a teensy bit faster, with fewer words averaged to context vector, and the DBOW+W mode noticeably faster, probably about 33% faster, as it directly reduces the raw number of `word-vector -> word` predictions. This improvement could mean that adding a 3rd plain-DBOW model still takes less training time than the existing 2. (Though, holding all 3 in RAM might be a challenge.) 

From my eyeballing of the paper's declared vector-sizes, it looks like they tried 100, 300, 1000, & 10000 – & reported the 10000-dimensional (!) doc-vectors as best-performing. 10k dimensions are probably overkill & impractical on your dev laptop... but their 10000 wasn't that much better than their 1000. You could probably manage that – especially if perhaps discarding more smaller articles (so there aren't 10M+ total doc-vectors to train). You could also aim for 300 dimensions – very common in word-vector work, and one of the sizes they tested. If you go for 200 not matching anything the paper tried, I'd make a comment in the notebook noting that you're doing that for memory efficiency, & it's still sufficient to demo the benefits, though projects with adequate resources may find it useful to try larger dimensionalities when copious training data is available. 

Asuming your M1 CPU reports 8 cores, it's possible that `cores -1` (7) or `cores-2` (6) workers would offer even better training throughput. (Running a train just for a few minutes, for the reported words-rate in the log to stabilize, is enough to search the space:  this throughput remains roughly stable through long trains.) 

On a corpus this large, a non-default `sample=1e-05` (or even `sample=1e-06`) could halve training time, or more, with no harm to quality – or even improvement, by diluting (via aggressive removal) the influence of the most-common words. (I suspect the model paper may have used such a more-aggressive-than-default value, because I think some early iterations of `word2vec.c` used `1e-05` rather than the later, more modest `1e-04` default that eventually prevailed.) 

Note also that `max_final_vocab` dynamically sets a `min_count` such that the surviving total of words is *no more than* the target. (It won't hit the target exactly, if that'd require arbitrarily saving some words of frequency *n* but not others.) So, you're only winding up with `effective_min_count=27` & 894,446 surviving words. As I suspect the paper authors just chose a `min_count` that got them under 1M words, using `max_final_vocab=1000000` might be more in spirit with the paper, and wind up with a closer number of surviving words. 

Finally, the paper seems to declare that they used HS mode, which you could match by specifying `hs=1, negative=0` to the models. I'm not sure what effect that might have on training speed or end-evaluations. Generally negative-sampling does relatively better (same benefits in less training time) as vocabularies grow larger, but I'm not sure where this corpus/demo falls in terms of potential 'sweet spots'. Since negative-sampling is the default, & does well, & seems to be used more commonly on serious projects, it seems OK to use it here. But if trying to match the paper as closely as possible in all declared/practical aspects, you'd want to use HS.

piskvorky:
> I notice you're using a non-default `window=8` value – though the paper doesn't seem to specify what value they're using. You might as well leave the default `window=5` in place. 

I didn't touch the pre-existing model settings. I assumed, given the notebook's explicit goal of "replicate the paper", that these were already set appropriately. So in addition to poor language and poor code, the setup was also incorrect? That's sad.

But at least a chance to clean up properly now… Snowballing from the initial user report of `No str.decode in python3`.

> You could also aim for 300 dimensions – very common in word-vector work, and one of the sizes they tested. If you go for 200 not matching anything the paper tried, I'd make a comment in the notebook noting that you're doing that for memory efficiency, & it's still sufficient to demo the benefits, though projects with adequate resources may find it useful to try larger dimensionalities when copious training data is available.

OK (dtto).

> Asuming your M1 CPU reports 8 cores, it's possible that `cores -1` (7) or `cores-2` (6) workers would offer even better training throughput. (Running a train just for a few minutes, for the reported words-rate in the log to stabilize, is enough to search the space: this throughput remains roughly stable through long trains.)

Dtto. Your suggestion makes good sense, I'll set the worker threads to 8 or 9; I use my laptop for other things too during the training anyway.

Regarding all the following:

> Looks good overall. My main suggestion would be to add a parallel demonstration of a 3rd mode: pure DBOW (leaving the default `dbow_words=0` in place).

> It's likely that DBOW alone would train fastest, _and_ do reasonably well on all the tasks that work only with articles. Even if not adding a "pure DBOW" trial, it'd be good to be clear that what you're trying is the special DBOW plus skip-gram introduced by the 'Document Embeddings' paper, not plain DBOW.

> On a corpus this large, a non-default `sample=1e-05` (or even `sample=1e-06`) could halve training time, or more, with no harm to quality – or even improvement, by diluting (via aggressive removal) the influence of the most-common words. (I suspect the model paper may have used such a more-aggressive-than-default value, because I think some early iterations of `word2vec.c` used `1e-05` rather than the later, more modest `1e-04` default that eventually prevailed.)

> Note also that `max_final_vocab` dynamically sets a `min_count` such that the surviving total of words is _no more than_ the target. (It won't hit the target exactly, if that'd require arbitrarily saving some words of frequency _n_ but not others.) So, you're only winding up with `effective_min_count=27` & 894,446 surviving words. As I suspect the paper authors just chose a `min_count` that got them under 1M words, using `max_final_vocab=1000000` might be more in spirit with the paper, and wind up with a closer number of surviving words.

> Finally, the paper seems to declare that they used HS mode, which you could match by specifying `hs=1, negative=0` to the models. I'm not sure what effect that might have on training speed or end-evaluations. Generally negative-sampling does relatively better (same benefits in less training time) as vocabularies grow larger, but I'm not sure where this corpus/demo falls in terms of potential 'sweet spots'. Since negative-sampling is the default, & does well, & seems to be used more commonly on serious projects, it seems OK to use it here. But if trying to match the paper as closely as possible in all declared/practical aspects, you'd want to use HS.

May I tempt you with a commit fix @gojomo? Clearly you're more knowledgable here so you set the params and then I run the cells.

Let me know because otherwise I'll adapt what I can & merge, I don't want to stall the release, the notebook isn't that vital.

Although after the clean up, maybe we could promote it to a tutorial @mpenkov, give it more visibility? Are the long running times a problem?

piskvorky:
@gojomo I changed the settings as suggested (300 dim, more subsampling, default window etc) & re-ran the notebook.

Good news: the `WARNING : Effective 'alpha' higher than previous training cycles` disappeared. It was indeed an artifact of re-launching the cells ad-hoc, corrupting the state somehow.

Bad news: the results seem significantly worse. They no longer match the paper, nor the accompanying text.

I'm starting to see why the original notebook author chose the settings they did. It's probably a result of some laborious experimentation, left undocumented. Which to be fair matches the original paper too, which seems to have been done in a similar (undocumented) manner. A meta-match.

gojomo:
It surprises me that a `window=8` vs `window=5` change (or `sample=1e-05` vs `sample=1e-04`) creates noticeable degradation in demo-able results.

I can take a look & test/apply some other tweaks this week - but given the long cycles involved in full runs, even if it's just a couple hours of poking around, it'll likely be 4+ days of wall clock time.

gojomo:
Thinking a bit more about it: 

That the 100d model "looks better" suggests to me the larger 300d model might've needed more `epochs` to fully converge. Checking back to the original paper, they're even coy about exactly how many epochs they used, saying only "we trained paragraph vectors over at least 10 epochs of the data". So they may have used more, especially if their running-epoch-loss readout (which Gensim sadly still doesn't have in `Doc2Vec`) helped better-hint to them when more epochs could help. 

So my vague hunch is that to the extent a smaller `window` & more-aggressive `sample` save lots of training time, achieving the best results probably requires then giving some training-effort back via more `epochs`. But of course it'd take a very time-consuming, more-extensive parameter search to be sure. 

To the extent the 100d model is going to be faster-to-train & smaller, and seems sufficient to demonstrate the vivid results, we might as well focus on that, with side commentary noting that when time/resources permit, even larger models can perform better on rigorous evaluations, as the paper's {300, 1000, 10000} figure implies.

(More generally, another reason for some drift in results, or other entries surging past the paper's, could easily be the emergence of newer acts in the later Wikipedia dumps we're using. So if matching-as-closely-as-possible were a top priority, using a similar vintage dump could be considered. But I think using the `latest` is a better basis for users present needs – & if we or they are going to spend the effort downloading/preprocessing a 10-20GB dataset, most would rather have the `latest` around than some arbitrary dump from a decade+ ago.)

piskvorky:
Several DM runs have stalled (no progress, no CPU used). @gojomo is this a known bug, have you seen it before?

Screen Shot 2022-03-22 at 20 31 32

Restarting usually helps, the issue appears in maybe 20% runs.

Only in DM mode training btw, never in DBOW

gojomo:
Never seen that. Suggestive of a queueing issue where either expected items are lost, or the cross-thread signaling has gone awry. I'd look in the console where the notebook-server/kernel is running to see if some sort of odd/fatal error happened in a worker thread preventing normal progress.

piskvorky:
I don't see anything out of ordinary in the server console:

Screen Shot 2022-03-23 at 08 14 46

The cell "stopped" between 1:08 and 1:30 AM.

This is weird. The issue has become pretty much 100%, so I haven't been able to finish the notebook the last couple of days. I wonder if it's some HW issue.

gojomo:
My first run of a version of your "100d" notebook also hit the DM-mode hang, at the end of the very-1st epoch, waiting forever for 4 of the threads to report they were finished. My environment is MacOS 12.3 / Python 3.9 on a pre-M1 MacBook Pro - so it's not unique to the M1-compiled-versions you're using. That this is now so quick to reproduce makes me suspect it's a regression. I still suspect worker threads are silently dying.

piskvorky:
I retrained on a Linux machine, several times, and everything worked fine.

So I suspect it's some Mac issue. Because the issue appears both pre-M1 (you) and on M1 (me), it's likely something related to the OS. Maybe the MacOS BLAS, I remember it has some quirks.

Either way, I'm done here. @gojomo please review and I'll merge and we'll release Gensim.

PR #3309, opened 2022-03-20
Respect encoding when reading binary keyed vectors

alhoo:
Current implementation fails to read keyed vectors that have iso-8859-1
encoding in the words when encoded in binary format. An example of this
type of a file can be seen in the turkuNLP finnish embeddings:

http://dl.turkunlp.org/finnish-embeddings/finnish_s24_skgram.bin

This file is quite trivial to load by passing the encoding to the vector
loading function. It is also logical that when user asks

KeyedVectors.load_word2vec_format(filename, binary=True, encoding='iso-8859-1')

The library would try to load the file assuming that the matrix is in
binary format and the words are encoded using iso-8859-1 encoding.


piskvorky:
Makes sense, thanks.

If this works and all tests pass let's include this PR in the upcoming release @mpenkov .

piskvorky:
One test failed, can you check please @alhoo?

Issue #3224, opened 2021-08-31
Can't add vector to pretrained fasttext model via .add_vector


om-hb:
#### Problem description

I'm trying to add a new vector to a pretrained fasttext model via `.add_vector`. However, it seems like the vector is not added if I check via `.has_index_for`.

#### Steps/code/corpus to reproduce

```
>>> import gensim.models.fasttext as fasttext
>>> import numpy as np
>>> ft_model =  fasttext.load_facebook_vectors("fastText/cc.en.300.bin")
>>> ft_model.has_index_for("testtest")
False
>>> ft_model.add_vector("testtest", np.zeros((300,)))
2000000
>>> ft_model.has_index_for("testtest")
False
>>> ft_model.index_to_key[2000000]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
```

#### Versions

```
Windows-10-10.0.19041-SP0
Python 3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]
Bits 64
NumPy 1.20.3
SciPy 1.6.2
gensim 4.0.1
FAST_VERSION 1
```

faustusdotbe:
Hi! Have you found a solution yet? I encounter a same issue with a 4.x-trained (cf https://github.com/RaRe-Technologies/gensim/issues/3114) Word2Vec:

When adding new vectors:
- using `m.wv.add_vector(word, vector)` works, **BUT** after saving to disk and reloading, 
    - `m.wv.get_vector(w)` does return the vector
    - `m.wv.has_index_for(w)` does return True
    - `m.wv.most_similar(w)` returns  `ValueError: operands could not be broadcast together with shapes (115026,) (115025,)`

- using `m.wv.add_vectors(list_of_words, list_of_numpy_arrays, replace  = True)`, works, **BUT** after saving to disk, the model does not load:

```
Traceback (most recent call last):
  File "/home/simon/w2v/word2vec-utils/out/models/test_test.py", line 3, in <module>
    m2 = gensim.models.KeyedVectors.load("model_32_5_word2vec_sanitised.model")
  File "/home/simon/w2v/word2vec-utils/env2/lib/python3.9/site-packages/gensim/utils.py", line 486, in load
    obj._load_specials(fname, mmap, compress, subname)
[...]

PR #3320, opened 2022-04-08
Fix `KeyedVectors.add_vectors()` error when use most_similar

s2terminal:
Fixing one of the issues in https://github.com/RaRe-Technologies/gensim/issues/3224

If I call `most_similar()` before doing `add_vectors()`  and then call `most_similar()` again after doing `add_vectors()`, I get a `ValueError: operands could not be broadcast together with shapes`.
This error occurs because `len(vectors)` and `len(vectors.norms)` do not match.

```python
import numpy

model = Word2Vec(sentences=[
                            ["this", "is", "test1"],
                            ["that", "is", "test2"],
], vector_size=2, min_count=1)

print(model.wv.most_similar("test1", topn=1)) #=> [('test2', 0.9941185712814331)]

model.wv.add_vectors(["test3"], [numpy.array([0.5, 0.5])])

print(model.wv.most_similar("test1", topn=1)) #=> ValueError: operands could not be broadcast together with shapes (6,) (5,) 
```

To resolve this error, I have used fill_norms to match `len(vectors)` and `len(vectors.norms)`.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -614,6 +614,8 @@ class KeyedVectors(utils.SaveLoad):
             for attr, extra in extras:
                 self.expandos[attr][in_vocab_idxs] = extra[in_vocab_mask]
 
+        self.fill_norms()
+
     def __setitem__(self, keys, weights):
         """Add keys and theirs vectors in a manual way.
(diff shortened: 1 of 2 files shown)


mpenkov:
Needs a test. The example from the issue description is probably good enough.

Issue #3321, opened 2022-04-10
Improve models dump


dariocurr:
#### Problem description

I think you could improve the disk footprint of the save and load mechanism 

#### Steps/code/corpus to reproduce

In order to save space you could do something like the following, avoiding to save the whole `random_state` but just its seed

```python
model.random = np.nan
model.random_seed = model.random.get_state()[1][0]
```

and then when loading

```python
model.random = np.random.RandomState(seed=model.random_seed)
```

#### Versions

Please provide the output of:

```python
Linux-5.13.0-39-generic-x86_64-with-glibc2.31
Python 3.9.12 | packaged by conda-forge | (main, Mar 24 2022, 23:22:55) 
[GCC 10.3.0]
Bits 64
NumPy 1.22.3
SciPy 1.8.0
gensim 4.1.2
FAST_VERSION 0
```

piskvorky:
Hi, can you expand the motivation section? What would be the advantage of your suggestion, what's the problem?

gojomo:
The savings would have to be significant to justify adding extra lines of pre-save/post-load special handling. 

For the given example, replacing a `RandomState` with just its original `seed`, it looks like the savings might be 2-3KB:

    len(pickle.dumps(np.random.RandomState()))  # 2788 in my Python 3.9.5 test

But the models are often hundreds of megabytes or gigabytes in size, making this a negligible savings.

But the cost is a few lines of extra code – and, as shown, a slight behavior change. (in my test, a `RandomState` started again from the `seed` so saved will *not* give the same stream of future randoms as a pickled-then-restored `RandomState`. That risks breaking user expectations. 

There might be other optimizations possible in model saving, but this example shows trivial gains, for non-zero risks. 

More valuable work would be looking into whether advances in the `pickle` protocol, or some other non-custom-to-Gensim serialization, can replace Gensim's custom steps entirely, a possibility raised by #2848.

piskvorky:
Yeah sounds like a non-issue – expanded motivation would be great @dariocurr. What prompted this?

dariocurr:
We want to reduce the disk footprint as much as possible when storing a model, and in our implementation we implemented the few lines above.

The problem is the **disk footprint**.

I agree that it's a very small improvement but it's something, and this is just one case I found digging through your code. I don't know if other, more compelling improvements could be made. 

About the following:

[...]

PR #3322, opened 2022-04-10
default isort && default black

dariocurr:
Hi, I am Dario, a developer at [BuildNN](https://github.com/buildnn).

I tried to give a look at the code to solve https://github.com/RaRe-Technologies/gensim/issues/3321 but i found the code hard to read. I just gave a default run of [isort](https://github.com/PyCQA/isort) and [black](https://github.com/psf/black) to ease contributions

I would like to fix the plenty of [flake8](https://github.com/PyCQA/flake8) inconsistencies as well but I don't know if it's appreciated

I guess it's easier and safer to run them by yourself instead of checking thousands of lines, but if you would like to solve all flake8's inconsistencies just say it


piskvorky:
Hi. I'm not opposed in  principle but this PR doesn't seem to match the [coding style we find practical](https://github.com/RaRe-Technologies/gensim/wiki/Developer-page#code-style). E.g. I see no end trailing comma at the end of lists. Is it possible to fix this?

Also we run automated `flake8` tests as part of CI and this PR seems to trigger a lot of errors.

gojomo:
The trend towards opinionated auto-formatting, and `black` in Python, is interesting. I was considering raising it as a future possibility, to align Gensim with larger trends, & forestall some of the digressions/learning-curves involved with getting new devs sync'd to project-style. 

But, there's a bit of up-front cost in adjusting habits, and the big reformat diffs in the history that might make other changes hard to follow (like any big change-of-style transition). The biggest factor is how comfortable the existing devs are to the new practices. 

Like @piskvorky, the default-to-double-quotes is a bit different than I'd pick, & initially disorienting, but I understand the reasons for it & it's far from a deal-breaker. Line-lengths are actually one of the areas that, despite `black`'s strong opinions, it lets you change. (Though, rigorous application of some of its other line-breaking conventions rules may make its default 88-char limit less confining.) And `black` will respect, & even add, trailing commas when such series are already, or necessarily, multiline. 

So as a matter of going-with-the-overall-open-source-Python-community-practices, I'd be mildly in favor of moving towards `black`-enforced formattting - but not necessarily ASAP, or in one big jump. 

It'd be best to start right *after* a fairly major, & stable, release – so that there's more time for devs to adjust to the altered-style before any other fixes/release are urgent. While there's a good case to "get it all over with" in one leap, we could also consider "dipping our toes in" with an incremental, module-by-module switchover - perhaps starting with modules that have historically received fewer updates, to iron out any unexpected interactions with flake/CI/doc-comments/etc. Then, with comfort, set a target for a full (& applied-on-commit, enforced-by-CI) switchover. 

Some of the edge-case auto-reformatting might be better with a human touch, or review, that still reaches `black`-conformance. For example, at...

https://github.com/RaRe-Technologies/gensim/blob/79a8d09f6f0720d359a9663207274829801c60b9/gensim/models/keyedvectors.py#L749

...a comment that was just barely over the 88-char limit broke things onto 3 lines, with extra parens - whereas changing the line-ending comment to a line-before comment, sparing the parens, keeping the code on 1 line would strike me as a better style. Still, that's pretty minor (and most such cases might disappear if we alloed a longer line-length), so might not justify the extra effort of human review before going forward.

piskvorky:
That's a good point. The strict line length discourages inline comments. Especially if the comment is inserted (weirdly) at the *closing* bracket – I saw a lot of that in the diff. We'd need to switch all inline comments into standalone-comment-line-before-the-code-line.

No big deal IMO.

So if the `black` settings allow for it, and if we can reasonably automate the style checks / formatting (git hooks?) in CI, we can go ahead.

But seeing how much time we all "have" lately (I had another kid), I'd probably be in favour of a single full PR switch = task done. I doubt any of us will have the capacity / attention span for incremental tinkering and checking and re-checking, even if that would be ideal.

dariocurr:
I understand all of your points and agree with you. I am glad that my PR has started an internal discussion.
As I said it is easier and safer for you to change on your own, either step by step or in one big jump.

Yes, the main point is.
> The biggest factor is how comfortable the current developers are with the new practices.

Also, I opened this PR and raised this discussion because:
1. I haven't looked at the code-style wiki section (my bad) (maybe mentioning it in the main README, over [here](https://github.com/RaRe-Technologies/gensim/blob/develop/CONTRIBUTING.md) might be helpful)
2. I couldn't find any specification in the `setup.cfg` file.

I guess the latter could be something useful to, for example at least:
- specify the line-length you use
- specify the flake8's inconsistencies that you want to ignore (i.e [E402](https://www.flake8rules.com/rules/E402.html#:~:text=All%20module%20level%20imports%20should,in%20between%20module%20level%20imports.))

piskvorky:
Let's see what @mpenkov thinks but I think the discussion has reached its conclusion :-)

@dariocurr If you can make `black` a) split lines at 120 chars, b) insert trailing commas we're good to go.

Flake8 I'm not sure about, I think we have a few exceptions like "this variable is unused but we still give it a name, because that makes the code easier to read / understand". But that's on a case-by-case basis, I don't know that we have any general "rule exceptions".

mpenkov:
This discussion about code formatting has come up multiple times before, like [here](https://github.com/RaRe-Technologies/gensim/pull/2712#discussion_r364076359).

Personally, I dislike automatic code formatters like black and never use them, for two reasons. First, the cost-to-benefit ratio is not worth it. Sure, with black, your code looks consistent, but it consistently looks like crap (sure, you can tweak some parameters, but I still don't like the output). To me, the main benefit of black seems to end flame wars about code style, but gensim never suffered from that particular problem. Like @piskvorky, it's also difficult for me to see how cosmetic issues made the code so difficult to work with that it prompted this PR. If the code is _that_ crap, then it's unlikely that formatting alone will fix it. So, I don't see the practical benefit of reformatting our entire code base, especially at this stage in the project's lifecycle.

Second, I typically put effort into formatting my code to make it readable, and I don't appreciate some program stomping over that it reformatting it due to some arbitrary rules. That said, the vast majority of the gensim code wasn't written by me, so this is more of a personal point than an argument against applying black to gensim.

In the end, it's @piskorky's call. If you _do_ go ahead with this, my recommendation would be to try it out on a reasonably-sized module like keyedvectors.py first, and then see if you like the results before applying it to the entire repo. Like @gojomo said, definitely do this _after_ the next major release, because there'll be havoc and it'll take time to bring things back to equilibrium. On top of that, you probably want to merge existing PRs first, because it may be difficult to do once the develop branch has been reformatted.

dariocurr:
I don't know what to say. This is certainly your call. I raised this discussion because since `gensim` is an open-source project, in my humble opinion, it would be nice to have a common standard that is easy to use, 

@piskvorky black obviously supports custom line length, I should look into the issue of trailing commas instead. I could try to meet your needs, but I think you should decide what the standards are first.

By the way, what about import sorting?

mpenkov:
> By the way, what about import sorting?

Sure, that makes sense. I think we're currently splitting the imports into three sections:

1. Standard library
2. Non-standard library
3. Gensim submodules

and sorting each section, but this isn't enforced anywhere by linters or otherwise.

PR #3323, opened 2022-04-13
Allow overriding the Cython version requirement

pabs3:
Use an environment variable for this since it is often easier to set in a
build wrapper rather than trying to override command-line options in the
right layer of a multi-layer build wrapper and it also requires a lot less
code to do the override.

This will be useful for using alpha versions of Cython or old versions of
Cython provided by the distros or specific versions that fix certain bugs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -320,6 +320,9 @@ NUMPY_STR = 'numpy >= 1.17.0'
 #
 CYTHON_STR = 'Cython==0.29.28'
 
+# Allow overriding the Cython version requirement
+CYTHON_STR = os.environ.get('GENSIM_CYTHON_REQUIRES', CYTHON_STR)
+
 install_requires = [
     NUMPY_STR,
     'scipy >= 0.18.1',


piskvorky:
> useful for using alpha versions of Cython or old versions of Cython provided by the distros or specific versions that fix certain bugs

Hi @pabs3 , what are you working on, what's the motivation here? The PR looks simple enough but could use some context.

pabs3:
I have packaged gensim for Debian and the version of Cython available
in Debian unstable is often not the same version as the one in the
gensim setup.py. I patched out the version pinning in the Debian
package but I think it would be better to have an official override
mechanism so that updating the patch is no longer needed.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

piskvorky (approved):

mpenkov (approved):

mpenkov:
Merged, thank you!

PR #3329, opened 2022-04-22
Check gallery up to date as part of CI

### LINKED ISSUE #2916: Add script to verify that autogenerated docs are up to date
I see no real alternative either. Doesn't moving parts of our docs + testing elsewhere simply shift the problem around? That new place will still need testing, whenever things change. No difference.

Except code + docs in one place (=now) makes more sense conceptually IMO.

Yes, scripts / hooks that verify docs are in-sync, preferably at PR merge-time, would be nice.

_Originally posted by @piskvorky in https://github.com/RaRe-Technologies/gensim/pull/2907#issuecomment-671317949_

mpenkov:
Been meaning to do this for a while (almost 2 years!).

When people update the gallery docs, it's possible (likely) they'll forget to update the cached gallery files. This script runs as part of the CI to notify them when they need to rebuild the gallery themselves.

Fix #2916


piskvorky (approved):
That docstring with motivation is great, thanks.

docs/src/auto_examples/tutorials/run_lda.py
```diff
@@ -245,7 +245,7 @@ def extract_documents(url='https://cs.nyu.edu/~roweis/data/nips12raw_str602.tgz'
 iterations = 400
 eval_every = None  # Don't evaluate model perplexity, takes too much time.
 
-# Make an index to word dictionary.
+# Make a index to word dictionary.
```
piskvorky, line 248: Looks like a regression.

docs/src/auto_examples/tutorials/run_lda.py
```diff
@@ -278,7 +278,7 @@ def extract_documents(url='https://cs.nyu.edu/~roweis/data/nips12raw_str602.tgz'
 # methods on the blog at http://rare-technologies.com/lda-training-tips/ !
 #
 
-top_topics = model.top_topics(corpus)
+top_topics = model.top_topics(corpus) #, num_words=20)
```
piskvorky, line 281: Also undesirable. We should probably revert whatever commit brought in those documentation changes.

mpenkov: OK, will revert.

I think the cache on develop may be stale.

PR #3344, opened 2022-05-05
Switch to Cython language level 3

pabs3:
Python 2 is not supported by gensim, so
switching to language level 3 should be fine.

Silences a warning from Cython:

FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release!

### DIFF
--- a/gensim/_matutils.pyx
+++ b/gensim/_matutils.pyx
@@ -1,6 +1,7 @@
 #!/usr/bin/env cython
 # coding: utf-8
 # cython: embedsignature=True
+# cython: language_level=3
 
 from __future__ import division
 cimport cython
--- a/gensim/corpora/_mmreader.pyx
+++ b/gensim/corpora/_mmreader.pyx
@@ -1,5 +1,6 @@
 # Copyright (C) 2018 Radim Rehurek <radimrehurek@seznam.cz>
 # cython: embedsignature=True
+# cython: language_level=3
 
 """Reader for corpus in the Matrix Market format."""
 import logging
--- a/gensim/models/doc2vec_corpusfile.pyx
+++ b/gensim/models/doc2vec_corpusfile.pyx
@@ -1,4 +1,5 @@
 #!/usr/bin/env cython
+# cython: language_level=3
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: cdivision=True
--- a/gensim/models/doc2vec_inner.pxd
+++ b/gensim/models/doc2vec_inner.pxd
@@ -1,5 +1,6 @@
 #!/usr/bin/env cython
 # distutils: language = c++
+# cython: language_level=3
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: cdivision=True
--- a/gensim/models/doc2vec_inner.pxd
+++ b/gensim/models/doc2vec_inner.pxd
@@ -16,7 +16,7 @@
 import numpy as np
 cimport numpy as np
 
-from word2vec_inner cimport REAL_t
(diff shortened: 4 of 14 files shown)


piskvorky:
Thanks. Where is that warning shown? If some extensions are being built with an incorrect language level, that sounds dangerous.

pabs3:
The warning is shown during the build of the Debian package of gensim ([amd64 log](https://buildd.debian.org/status/fetch.php?pkg=gensim&arch=amd64&ver=4.2.0%2Bdfsg-1&stamp=1651727914&raw=0)). Side note: the log also shows various GCC warnings, some test warnings and a warning about setup.py being deprecated within the Python community.

Unfortunately based on the CI results for this PR, it looks like the Cython code in gensim isn't yet compatible with Cython language version 3, so I'm going to mark this PR as draft and work on fixing the issues as I get time.

pabs3:
Hmm, maybe I should instead mark those files as language level 2 for now, to silence the warnings.

piskvorky:
`language_level` is about the level of Python (2 vs 3), right? Not Cython, which works with either.

So we should have `language_level=3` everywhere in Gensim. Unless I'm missing something.

pabs3:
I thought so too, but the CI fails when I add that everywhere.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

piskvorky:
Hmm, the errors complain about syntax stuff in the Cython code like `gensim/models/fasttext_inner.pyx:187:9: 'REAL_t' is not a type identifier`. Which is unrelated to py2/3 IMO, weird. Will need a deeper look.

piskvorky:
The problem seems to stem from this code:

https://github.com/RaRe-Technologies/gensim/blob/5a19dd0407577001bcc177166582da7a5a4e5f70/gensim/models/word2vec_inner.pxd#L125-L126

which seems indeed syntactically broken. It comes from 3c3506d51a2caf6b890de3b1b32a8b85f7566ca5 in #2127.

Weird. How did Gensim ever compile after this, since 4 years ago? I don't get it. CC @persiyanov @mpenkov.

piskvorky:
My hypothesis: error related to compilation order.

1. `fasttext_inner.pyx` depends on `word2vec_inner.pyx`
2. But with `language_level=3`, the `fasttext_inner.pyx` gets cythonized first => failure.
3. Under `language_level=2`, `word2vec_inner.pyx` is cythonized first and everything works.

Checking my hypothesis now :)

piskvorky:
OK I think I figured it out. The issue is with relative imports (`from word2vec_inner import …`). After changing them to absolute (`from gensim.models.word2vec_inner import …`) the compilation works.

@pabs3 I'll push the fix to your branch, OK?

piskvorky:
Hm, I get a permission error when I try to push to your branch directly (usually possible for maintainers across forks but maybe you disabled this).

And github doesn't let me open a PR against your fork (doesn't even list it??). So can you please apply 0e70635675f6ef50a0052b1c2de1c697d3a70e59 manually @pabs3? Cheers.

pabs3:
I usually disable maintainer edits of my patches; had
some suboptimal experiences with that in the past.

I think it would be best to merge your patch first, since otherwise git
bisects will have to skip my patch since it will be unbuildable, but if
your patch goes in first, then both patches will be bisect buildable.

Anyway, I've added your patch to the top of my branch too.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

mpenkov (approved):

mpenkov:
Thank you @pabs3 !

PR #3359, opened 2022-06-22
Fix bug that prevents loading old models

### LINKED ISSUE #3357: How to load 3.8.3 gensim word2vec model by new version 4.2.0 gensim model
## Problem description

Hi! I have been using gensim 3.8.1 and want to use the latest version, gensim 4.2.0.

I have been able to load models made with older gensim versions, but when I changed to the latest version, I can no longer load them.
This is the error code.

```zsh
Exception has occurred: AttributeError
Can't get attribute 'Vocab' on <module 'gensim.models.word2vec' from '/PATH'>
```

I would like to know if there is some way to load older models.

## Versions

python 3.8.9
gensim 4.2.0

funasshi:
Fixes #3357.

# Problem
the model created by gensim 3.8.3 cannot be loaded by load function of gensim 4.2.0 and Attribute error occurs.

# How to solve 
import Vocab in word2vec.py


piskvorky (approved):

gensim/models/word2vec.py
```diff
@@ -197,7 +197,7 @@
 import numpy as np
 
 from gensim.utils import keep_vocab_item, call_on_class_only, deprecated
-from gensim.models.keyedvectors import KeyedVectors, pseudorandom_weak_vector
+from gensim.models.keyedvectors import KeyedVectors, pseudorandom_weak_vector, Vocab
```
mpenkov, line 200: Is this really enough? Can you please add a test that successfully loads an older gensim model?

mpenkov (changes requested):

gensim/models/word2vec.py
mpenkov: Also, this import appears unused to flake8, so we should add a directive to ignore it.

gensim/models/word2vec.py
```diff
@@ -200,6 +200,11 @@
 from gensim.models.keyedvectors import KeyedVectors, pseudorandom_weak_vector
 from gensim import utils, matutils
 
+#
+# This unused import is "somehow" required to fix #3357
```
piskvorky, line 204: Explain the reason clearly – this comments not helpful.

piskvorky (changes requested):

gensim/models/word2vec.py
```diff
@@ -200,6 +200,11 @@
 from gensim.models.keyedvectors import KeyedVectors, pseudorandom_weak_vector
 from gensim import utils, matutils
 
+#
+# This import is required by pickle to load models stored by Gensim < 4.0, such as Gensim 3.8.3.
+#
```
piskvorky, line 205: No need for this blank line – the comment above relates directly to the line below.

piskvorky (changes requested):

piskvorky (approved):

mpenkov:
Not sure why the changes are causing tests to consistently fail. Will need to look into it later.

mpenkov:
Thank you @funasshi !

PR #3361, opened 2022-07-05
Using efficient np.linalg.multi_dot 

maldil:
It is more succinct and effective to refactor code to use `np.linalg.multi_dot` rather than numerous `np.dot` routines.
What do you think about this change that has practical value?

### DIFF
--- a/gensim/test/svd_error.py
+++ b/gensim/test/svd_error.py
@@ -62,7 +62,7 @@ def rmse(diff):
 
 
 def print_error(name, aat, u, s, ideal_nf, ideal_n2):
-    err = -np.dot(u, np.dot(np.diag(s), u.T))
+    err = -np.linalg.multi_dot([u, np.diag(s), u.T])
     err += aat
     nf, n2 = np.linalg.norm(err), norm2(err)
     print(


gojomo:
Can you quantify what the benefit might be, perhaps with a demo showing the improvement?

piskvorky:
Where does Gensim multiply more than 2 matrices (so that multidot has an effect)?

Like @gojomo, a demonstration of the benefits would be nice.

gojomo:
But where would the proposed replacement make any noticeable difference to a Gensim user? 

Is the `print_error()` function that's changed in the PR a major contributor to the runtime of the `svd_error.py` script? How much would a representative usage of that script be sped up by this change?

piskvorky:
Sorry, without a clear demonstration of its benefit, I'm -1 on introducing changes.

If there's a compelling use-case, such as a speed-up when training a new model, or calculating document similarities, please post it here.

mpenkov:
Thank you for your effort @maldil . I see this change has pros and cons.

Pros: more efficient
Cons: more obscure 

Given that efficiency is not a prime concern of the affected code, I think at this stage it's best not to merge this change, because the pros outweigh the cons.

Issue #3101, opened 2021-04-03
Segfault running run-core-concepts-py against first lines of Shakespeare's sonnets


jorendorff:
*(Hi! Thanks for putting so much effort into the tutorials! Gensim looks amazing and I'm looking forward to experimenting more. Happy to help debug--anything you need. I know Python, CPython, and C/C++ pretty well, and I can drive GDB.)*

`run-core-concepts.py` crashes with `Segmentation fault (core dumped)` when I replace `text_corpus` with a list of first lines of Shakespeare's sonnets.

#### Steps/code/corpus to reproduce

1. Go to <https://radimrehurek.com/gensim/auto_examples/core/run_core_concepts.html>, scroll to the bottom, click "Download Python source code". Move the downloaded script into a new, empty directory.

2. cd to that directory, `python3 -m venv venv; source venv/bin/activate; pip install gensim`.

3. Note that `python run_core_concepts.py` now works fine, up until the end where it tries to import `matplotlib`.

4. Now edit `run_core_concepts.py` and replace the `text_corpus` with this:

    ```python
    text_corpus = [
        'From fairest creatures we desire increase,',
        'When forty winters shall besiege thy brow,',
        'Look in thy glass and tell the face thou viewest',
        'Unthrifty loveliness, why dost thou spend',
        'Those hours, that with gentle work did frame',
        "Then let not winter's ragged hand deface,",
        'Lo! in the orient when the gracious light',
        "Music to hear, why hear'st thou music sadly?",
        "Is it for fear to wet a widow's eye,",
        "For shame deny that thou bear'st love to any,",
        "As fast as thou shalt wane, so fast thou grow'st",
        'When I do count the clock that tells the time,',
        'O! that you were your self; but, love, you are',
        'Not from the stars do I my judgement pluck;',
        'When I consider every thing that grows',
        'But wherefore do not you a mightier way',
        'Who will believe my verse in time to come,',
        "Shall I compare thee to a summer's day?",
    ]
    ```

5. Note that `python run_core_concepts.py` now crashes; the output ends with

    ```console
     [(0, 1), (7, 1), (11, 1), (12, 1), (14, 1)],
     [(3, 1), (6, 1), (12, 1)],
     [(7, 1), (11, 1), (13, 1)],
     [(14, 1)],
     [(1, 1), (12, 1)]]
    []
    Segmentation fault (core dumped)
    ```

The crash occurs on the line:

```python
sims = index[tfidf[query_bow]]
```

[...]

PR #3373, opened 2022-08-01
dynamic number of features for tf-idf

Utopiah:
avoid core dump on larger user dataset, as discussed in https://github.com/RaRe-Technologies/gensim/issues/3101


piskvorky:
Thank you!

PR #3381, opened 2022-08-17
clarify runtime expectations

gojomo:
Motivated by discussion thread: https://groups.google.com/g/gensim/c/ZnuJkxWT6ps/m/x5pXQgU8AAAJ

Related: it seems there's an exact duplicate of this file in project-path `/gensim/docs/src/**auto_examples**/tutorials`. I *think* that (& sibling plain `.py` files there) are just automatic copies of these canonical versions, so I haven't applied the change there. (And perhaps, even though we want to commit the other sphinx-build outputs there, we don't want the duplicate `.py` sources there?)


docs/src/gallery/tutorials/run_doc2vec_lee.py
```diff
@@ -215,9 +215,15 @@ def read_corpus(fname, tokens_only=False):
...
-# minutes, so use optimized Gensim with BLAS if you value your time.
+# In the usual case, where Gensim installation found a BLAS library for optimized
+# bulk vector operations, this training on this tiny 300 document, ~60k word corpus 
+# should take just a few seconds. (More realistic datasets of tens-of-millions
+# of words or more take proportionately longer.) If for some reason a BLAS library 
+# isn't available, training uses a fallback approach that takes 60x-120x longer, 
+# so even this tiny training will take minutes rather than seconds. (And, in that 
+# case, you should also notice a warning in the logging letting you know there's 
+# something worth fixing.) So, be sure your installation uses the BLAS-optimized 
+# Gensim if you value your time.
```
piskvorky, line 226: A concrete set of instructions for how to actually perform such check would be helpful. E.g. through the `FAST_VERSION` attribute.

I'd also personally drop the brackets, but I realize that's your writing style :)

piskvorky:
Thanks! I'm also not sure what the "canonical" vs generated paths are, deferring to @mpenkov .

mpenkov:
Code under auto_examples is essentially a copy. We _do_ keep it under version control (reasons below), but you shouldn't have to change it manually yourself. Instead, rebuild the gallery, and Sphinx will update those files for you. You can then add them to the PR.

The reason why we keep these duplicate files under version control is to keep Sphinx happy. If those files are missing, then Sphinx will attempt to rebuild the gallery from the source files, by running each of them individually. This takes time (hours). We want to avoid that, because we want our documentation to build relatively quickly from a fresh clone of the repo.

mpenkov:
@gojomo @piskvorky Our CI now [shows](https://github.com/RaRe-Technologies/gensim/runs/7944055617?check_suite_focus=true) you what needs to be done:

```
The gallery cache appears stale.

Rebuild the documentation using the following commands from the gensim root subdirectory:

    pip install -e .[docs]
    make -C docs/src html

and then run `git add docs/src/auto_examples` to update the cache.

Stale files: ['docs/src/auto_examples/tutorials/run_doc2vec_lee.py']
```

This has been around for a while, but other linting issues with the PR were masking it.

PR #3382, opened 2022-08-19
Added flsamodel.py and added a dependency

ERijck:
FLSA-W is a new and state-of-the-art topic modeling algorithm (https://ieeexplore.ieee.org/abstract/document/9660139, ). I have added flsamodel.py, a script with which the FuzzyTM package is called with parameters and methods similar to ldamodel.py. In the setup.py file, I have added FuzzyTM as a dependency.

### DIFF
--- a/gensim/models/flsamodel.py
+++ b/gensim/models/flsamodel.py
@@ -8,6 +8,7 @@ import itertools
 from FuzzyTM import FLSA_W, FLSA
 import gensim
 
+
 class FlsaModel:
     """
     Class to initialize and train fuzzy topic models with methods similar
--- a/gensim/models/flsamodel.py
+++ b/gensim/models/flsamodel.py
@@ -1,0 +1,208 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Jul 11 17:05:39 2022
+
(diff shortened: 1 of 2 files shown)


gojomo:
Given that all the real work is done in the `FuzzyTM` module, & that looks pretty easy for any user to import/use directly, I don't see much value in this small `FlsaModel` wrapper class - but it does add extra costs/risks via coupling/extra-dependencies. 

To the extent some of the corpus-handling helper methods might make it easier for people using Gensim models to also compare results from `FLSA` or `FLSA_W`:

* could those helpers be generic utilities that don't require importing anything from `FuzzyTM`?
* could those helpers live in `FuzzyTM` package instead, possibly even providing a Gensim-shaped workalike model from there (without even explicitly importing Gensim code)?

mpenkov:
On top of what gojomo said, FuzzyTM is also very new: https://github.com/ERijck/FuzzyTM

So I think this PR is at best premature.

ERijck:
I have a new version with an extra model (FLSA-W) without a FuzzyTM and sparsesvd dependency. Hence, I will close this pull request and open a new one.

PR #3384, opened 2022-08-28
NLP functions for pre-processing text in the English language in a pandas dataframe including the removal of stopwords.

biswaroopmitra:
Hello. This would be my first contribution towards an open-source project.

**Background to the idea**

Pre-processing text utilises multiple steps that include, amongst other steps, the removal of special characters and punctuations, stemming/lemmatization as well as the removal of stopwords based on the language in use.

For the pre-processing of text data in the English language in a pandas dataframe to involve fewer lines of code, it may be a good idea if these steps of pre-processing could be clubbed together into a function that could be called when necessary. This could help by not having to write multiple lines of code every time we need to pre-process text-based data especially in a dataframe.

`gensim.utils.simple_preprocess()` provides tokens after pre-processing text in the English language. However, there seems to be the scope for new functions to be developed that involve iterating through text in a dataframe and include:
1. the removal of stopwords in the English language
1. rejoining the words obtained after the above step 

The above are some of the steps that the functions would perform. Also, adding `tqdm` to the above mix to monitor the progress of the iterations could be useful.

**Output data structure, IDE used, and dataset**

The new functions return their outputs in the form of a list. I have worked on and tested the new functions using Colab on a dataset obtained from [Kaggle](https://www.kaggle.com). The dataset's file has been attached for reference.
[E-Commerce Reviews.csv](https://github.com/RaRe-Technologies/gensim/files/9440045/E-Commerce.Reviews.csv)


gojomo:
It's theoretically-possible that Gensim could benefit from some additional utility preprocessing functions, for common use-cases. But, such preprocessing is not a main focus of the project – which tends to leave such choices to user code, with just a few minimal utilities to assist in tutorials/tests/demos. So, the case for inclusion of new options would need to be pretty strong. 

Also, the desirable form of any such new capability would be highly-usable/well-documented/broadly-applicable code, in a logical place (perhaps `gensim.util` or `gensim.parsing.preprocessing`), with some matching test/demo code. It would also not be `pandas`-specific functionality – as Gensim doesn't require or assume data is coming from `pandas` datasctructures. (Such datastructures also add overhead & complexity that's often burdensome for the large text corpora often used with Gensim.) 

A new notebook as appears in this PR, alone at the top-level of the project, without rationale/docs/etc, & with a couple `pandas`-specific functions, would not be consistent with the project's usual layout/approach/aims.

So I suggest you return to the "drawng-board", and if you think there's a gap in what Gensim provides, explain better the need, and adapt any contributions to fit the existing Gensim layout/standards.

piskvorky:
What @gojomo said :)

It's a good effort @biswaroopmitra but not a good fit for Gensim in the current form.

biswaroopmitra:
@gojomo and @piskvorky Thank you for the feedback.

PR #3363, opened 2022-07-08
Implement numpy hack in setup.py to enable install under Poetry

jaymegordo:
Fixes https://github.com/RaRe-Technologies/gensim/issues/3362

Confirmed poetry install works on M1 mbp with this fix.

`poetry add git+ssh://git@github.com:jaymegordo/gensim.git#fix_3362`

PR #3386, opened 2022-09-05
Enable install under poetry

simonseed:
Implement numpy hack in setup.py to enable install under Poetry #3363

This is the same as https://github.com/RaRe-Technologies/gensim/pull/3363 but on the 3.8.3 branch

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,15 @@ class CustomBuildExt(build_ext):
         build_ext.finalize_options(self)
         # Prevent numpy from thinking it is still in its setup process:
         # https://docs.python.org/2/library/__builtin__.html#module-__builtin__
-        __builtins__.__NUMPY_SETUP__ = False
+        try:
+            __builtins__.__NUMPY_SETUP__ = False
+        except:
+            try:
+                # For python 3
+                import builtins
+                builtins.__NUMPY_SETUP__ = False
+            except:
+                print("Skipping numpy hack; if gemsum installation fails, try installing numpy first")
 
         import numpy
         self.include_dirs.append(numpy.get_include())


setup.py
```diff
@@ -108,7 +108,15 @@ def finalize_options(self):
         build_ext.finalize_options(self)
         # Prevent numpy from thinking it is still in its setup process:
         # https://docs.python.org/2/library/__builtin__.html#module-__builtin__
-        __builtins__.__NUMPY_SETUP__ = False
+        try:
+            __builtins__.__NUMPY_SETUP__ = False
+        except:
+            try:
+                # For python 3
```
piskvorky, line 115: Gensim does not support python 2 any more.

piskvorky (changes requested):

setup.py
simonseed: This is a PR against 3.8.3 which does support Python 2. I am happy to remove Python 2 support from the PR if that is an issue. The problem this solves is, install with Poetry (Python 3) fails for the the same reason as reference in   https://github.com/RaRe-Technologies/gensim/pull/3363

piskvorky: Yes, correct of course.

simonseed: I know it's not supported. But the Poetry install suddenly stopped working - same as in the develop branch. And for those of us that are still using it. This small PR fixes the issue.

mpenkov: I think the best we can do is fix this on develop so future versions won't be affected.

Releasing a bugfix for 3.8.3 requires effort, the volume of which isn't justified, IMHO.

mpenkov:
Closing, will not fix.

Issue #3388, opened 2022-09-28
FastTextKeyedVectors.add_vectors is not adding vectors


globba:
#### Problem description

I have been trying to create a `FastTextKeyedVectors` and adding vectors to it using either `add_vector` or `add_vectors` but the methods are not adding anything. After looking at the implementation of those methods, I think there is an error while checking if a key has already been added.

#### Steps/code/corpus to reproduce

I create a `FastTextKeyedVectors` using the defaults used by the `FastText` model, then try to add vectors to it using `add_vector` or `add_vectors`:

```
wv = FastTextKeyedVectors(vector_size=2, min_n=3, max_n=6, bucket=2000000)
wv.add_vector("test", [0.5, 0.5])
print(wv.key_to_index)
>> {}
print(wv.index_to_key)
>> []
print(wv.vectors)
>> []

wv.add_vectors(["test"], [[0.5, 0.5]])
print(wv.key_to_index)
>> {}
print(wv.index_to_key)
>> []
print(wv.vectors)
>> []
```

`wv.key_to_index`, `wv.index_to_key` and `wv.vectors` are all empty.

`FastTextKeyedVectors` is a child of `KeyedVectors` where the `add_vector/s` methods are implemented. `add_vector` does a few checks then calls `add_vectors`.
In `add_vectors`, there is an `in_vocab_mask`, which a list of booleans indicating if a key is already present in the KeyedVectors.

```
in_vocab_mask = np.zeros(len(keys), dtype=bool)
        for idx, key in enumerate(keys):
            if key in self:
                in_vocab_mask[idx] = True
```

Since Gensim 4.0, `key in wv` will always return True with FastText by design. The proper way of checking if a key exists is by calling `key in wv.key_to_index` (See https://github.com/RaRe-Technologies/gensim/wiki/Migrating-from-Gensim-3.x-to-4#10-check-if-a-word-is-fully-oov-out-of-vocabulary-for-fasttext)

So replacing the above code by
```
in_vocab_mask = np.zeros(len(keys), dtype=bool)
        for idx, key in enumerate(keys):
            if key in self.key_to_index:
                in_vocab_mask[idx] = True
```
seems to fix the issue.

```
wv = FastTextKeyedVectors(vector_size=2, min_n=3, max_n=6, bucket=2000000)
wv.add_vectors(["test"], [[0.5, 0.5]])
print(wv.key_to_index)
>> {'test': 0}
print(wv.index_to_key)
>> ['test']
[...]

PR #3389, opened 2022-09-29
Fixed FastTextKeyedVectors handling in add_vector 

globba:
Fixes issue #3388

Since Gensim 4.0,`'key' in FastTextKeyedVectors` always returns True by design.
This resulted in vectors not being added when performing `FastTextKeyedVectors.add_vectors()` due to the way that `in_vocab_mask` was being created.
The proper way to check if a key already exists is with `'key' in FastTextKeyedVectors.key_to_index`.

### DIFF
--- a/gensim/models/keyedvectors.py
+++ b/gensim/models/keyedvectors.py
@@ -592,7 +592,7 @@ class KeyedVectors(utils.SaveLoad):
 
         in_vocab_mask = np.zeros(len(keys), dtype=bool)
         for idx, key in enumerate(keys):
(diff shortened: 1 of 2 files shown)


mpenkov (approved):

piskvorky:
@gojomo please review so we can merge.

mpenkov:
@gojomo Ping

gojomo:
Looks good to me!

mpenkov:
Thank you @globba !!

Issue #3395, opened 2022-10-26
Support Python 3.11


djstrong:
#### Problem description

gensim does not install on Python 3.11.

#### Steps/code/corpus to reproduce

While installing gensim by `pip install gensim`:
```
      gensim/models/word2vec_inner.c: In function ‘__Pyx_AddTraceback’:
      gensim/models/word2vec_inner.c:457:62: error: dereferencing pointer to incomplete type ‘PyFrameObject’ {aka ‘struct _frame’}
        457 |   #define __Pyx_PyFrame_SetLineNumber(frame, lineno)  (frame)->f_lineno = (lineno)
            |                                                              ^~
      gensim/models/word2vec_inner.c:12424:5: note: in expansion of macro ‘__Pyx_PyFrame_SetLineNumber’
      12424 |     __Pyx_PyFrame_SetLineNumber(py_frame, py_line);
            |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      error: command '/usr/bin/gcc' failed with exit code 1
```

piskvorky:
Please fill in the issue template fully, instead of deleting it.

acul3:
i also encountered it

any plans to support 3.11?

acul3:
here stacktrace when install from the source :

```
/Users/samsulrahmadani/env/lib/python3.11/site-packages/setuptools/installer.py:27: SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.
  warnings.warn(
WARNING: The wheel package is not available.
running install
/Users/samsulrahmadani/env/lib/python3.11/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/Users/samsulrahmadani/env/lib/python3.11/site-packages/setuptools/command/easy_install.py:144: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
running bdist_egg
running egg_info
creating gensim.egg-info
writing gensim.egg-info/PKG-INFO
writing dependency_links to gensim.egg-info/dependency_links.txt
writing requirements to gensim.egg-info/requires.txt
writing top-level names to gensim.egg-info/top_level.txt
writing manifest file 'gensim.egg-info/SOURCES.txt'
Compiling gensim/models/word2vec_inner.pyx because it changed.
[...]

PR #3402, opened 2022-11-08
add support for python 3.11

acul3:
Update Cython, numpy ,scipy version to support python 3.11 

#3395 #3401

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -318,20 +318,20 @@ docs_testenv = core_testenv + distributed_env + visdom_req + [
     'pandas',
 ]
 
-NUMPY_STR = 'numpy >= 1.17.0'
+NUMPY_STR = 'numpy >= 1.23.4'
 #
 # We pin the Cython version for reproducibility.  We expect our extensions
 # to build with any sane version of Cython, so we should update this pin
 # periodically.
 #
-CYTHON_STR = 'Cython==0.29.28'
+CYTHON_STR = 'Cython==0.29.32'
 
 # Allow overriding the Cython version requirement
 CYTHON_STR = os.environ.get('GENSIM_CYTHON_REQUIRES', CYTHON_STR)
 
 install_requires = [
     NUMPY_STR,
-    'scipy >= 0.18.1',
+    'scipy >= 1.9.3',
     'smart_open >= 1.8.1',
 ]
 


setup.py
```diff
@@ -318,20 +318,20 @@ def run(self):
...
-CYTHON_STR = 'Cython==0.29.28'
+CYTHON_STR = 'Cython==0.29.32'
 
 # Allow overriding the Cython version requirement
 CYTHON_STR = os.environ.get('GENSIM_CYTHON_REQUIRES', CYTHON_STR)
 
 install_requires = [
     NUMPY_STR,
-    'scipy >= 0.18.1',
+    'scipy >= 1.9.3',
```
piskvorky, line 334: This looks like the very latest version of scipy.

We want the *oldest* version (that works) here, because otherwise we make life difficult for users who don't have the very latest.

Could you please find which is the first version of numpy & scipy to support 3.11?

piskvorky (changes requested):

piskvorky:
@mpenkov do we need to update the CI versions too?

setup.py
acul3: > Could you please find which is the first version of numpy & scipy to support 3.11?

just check(install on local..pip)
the last version of scipy and numpy are 1.9.2 and 1.23.0 (not very far from the lastest)

maybe we can add special condition just for python 3.11 ?
something like overriding version(numpy,scipy) like the cython https://github.com/RaRe-Technologies/gensim/blob/c93eb0b647ec9de7cfcb43c74c37f295f2944821/setup.py#L330

piskvorky: According to https://devguide.python.org/versions/, the currently oldest supported Python (before end-of-life) is Python 3.7. So we can ignore Pythons older than that (3.6 etc).

Now looking at SciPy's compatibility matrix https://docs.scipy.org/doc/scipy/dev/toolchain.html :

Screen Shot 2022-11-08 at 10 22 27

It seems for 3.11 we need Scipy 1.9.0 (or .1?) and Numpy 1.18.5. Which doesn't match your `the oldest version that support 3.11 for scipy and numpy are 1.9.2 and 1.23.0 ` – can you verify please?

> maybe we can add special condition just for python 3.11 ?

I don't think so. Gensim is a library used by ML developers, so we can afford to update its dependencies. We just shouldn't do it without a reason. @mpenkov WDYT?

mpenkov: Yes, I think that's fine. The versions you suggested also make sense to me.

mpenkov:
Yes, the CI stuff will need an update. That's probably the bulk of the work for this PR. @acul3 Are you able to take care of this? We need to:

- [ ] Add Py3.11 build for Linux and Windows
- [ ] Make sure all the other builds are passing

setup.py
acul3: ah thank you for detail and correction

my missunderstanding
i refer to this issue https://github.com/scipy/scipy/issues/16851#issuecomment-1271363490 for oldest "wheel" scipy 1.9.2 and https://numpy.org/news/ for numpy


> If the matrix above is correct, I'd go with Scipy>=1.9.0 and Numpy>=1.18.5.

just tried with this version and  it works

acul3:
sure @mpenkov . i will try work on this today

acul3:
@mpenkov @piskvorky 

python 3.11 ubuntu build test failed due 
error on `pip install nmslib`

there is still open issue on nmslib repo related about this(maybe):
https://github.com/nmslib/nmslib/issues/519#issuecomment-1214458081

some ref also:
https://github.com/RaRe-Technologies/gensim/issues/3260#issuecomment-1102793697

any advice?

piskvorky:
@acul3 could you disable that failing test (with nmslib) when running the tests under py3.11?

nmslib is not critical to most users of Gensim, it's just an extra, no need to block the release. We're planning to release soon (next week or so) and I'd love to get support for Python 3.11 in!

.github/workflows/build-wheels.yml
```diff
@@ -202,24 +217,19 @@ jobs:
         echo ::group::Set up dependencies
         python --version
         python -c "import struct; print(struct.calcsize('P') * 8)"
-
         python -m pip install -U pip setuptools wheel wheelhouse_uploader ${{ env.BUILD_DEPENDS }}
         echo ::endgroup::
-
```
piskvorky, line 208: Any reason to remove the existing blank lines?

Here and in many other places in this PR.

setup.py
```diff
@@ -272,15 +272,14 @@ def run(self):
...
 
-if not (sys.platform.lower().startswith("win") and sys.version_info[:2] >= (3, 9)):
+if not sys.platform.lower().startswith("win") and sys.version_info[:2] < (3, 11):
     core_testenv.extend([
-        'POT',
-        'nmslib',
-    ])
+            'POT',
+            'nmslib',
+        ])
```
piskvorky, line 279: Bad indent – why this change?

.github/workflows/build-wheels.yml
```diff
@@ -48,7 +48,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10','3.11']
```
piskvorky, line 51:
```suggestion
        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
```

piskvorky (changes requested):
Thanks for the fixes! The whitespace reformatting looks sus, otherwise seems OK.

.github/workflows/build-wheels.yml
acul3: i'm sorry

forgot to check the formatting
change it now

.github/workflows/build-wheels.yml
```diff
@@ -82,7 +82,7 @@ jobs:
         - os: ubuntu-latest
           manylinux-version: 2010
           python-version: 3.8
-          build-depends: numpy==1.17.3
+          build-depends: numpy==1.18.5
```
mpenkov, line 85: Why this change? For backwards compatibility, we want to build against the _oldest_ possible numpy version that has wheels for Python 3.8. I can see that:

- [1.17.3](https://pypi.org/project/numpy/1.17.3/#files): *does* have wheels for Py3.8
- [1.17.2](https://pypi.org/project/numpy/1.17.3/#files): *does not* have wheels for Py3.8

So isn't 1.17.3 the right choice here?

setup.py
mpenkov: @piskvorky @acul3 Does this mean that 3.6 and 3.7 are no longer supported? If so, we need to update setup.py accordingly.

acul3: > the currently oldest supported Python (before end-of-life) is Python 3.7. So we can ignore Pythons older than that (3.6 etc).

hmm it seems python 3.7 still need supported according to this argument

mpenkov: From the table above, it looks like it's impossible to support both 3.7 and 3.11 using a single version of scipy, numpy, etc.

So, our options are:

1. Use different versions for Py3.11
2. Drop 3.7 support now (3.7 reaches EOL in 6 months anyway)
3. Wait before 3.7 reaches EOL, and then add support for 3.11

Personally, I think 2) is best. 1) requires a fair bit of effort for very little value (6 more months of 3.7 support, and after 6 months the value of that effort goes down to zero). We can do a major version bump. If people really need support for 3.7, they can continue to use the current version of gensim.

@piskvorky thoughts?

piskvorky: Yes, agreed. Now that Python versions go away frequently, there isn't much expectation to support older ones (not as much as when py2.7 was supported for ages).

mpenkov:
@acul3 Can you please have a look at my changes?

acul3 (commented):
look goods to me

mpenkov:
Merged. Thank you @acul3!

Issue #3368, opened 2022-07-25
CoherenceModel does not finish with computing


PrimozGodec:
#### Problem description

When computing coherence scores, it newer finishes with computing on a bit bigger dataset. Run the code below (with the provided dataset) to reproduce. 

#### Steps/code/corpus to reproduce

```python
    with open("coherence-bug.pkl", "rb") as f:
        model, tokens = pickle.load(f)

    print("conherence")
    print(datetime.now())
    t = time.time()
    cm = CoherenceModel(model=model, texts=tokens, coherence="c_v")
    coherence = cm.get_coherence()
    print(time.time() - t)
```
[coherence-bug.pkl.zip](https://github.com/RaRe-Technologies/gensim/files/9178910/coherence-bug.pkl.zip)

#### Versions

The bug appears on Gensim version 4.2, but it does not happen on 4.1.2

macOS-10.16-x86_64-i386-64bit
Python 3.8.12 (default, Oct 12 2021, 06:23:56) 
[Clang 10.0.0 ]
Bits 64
NumPy 1.22.3
SciPy 1.8.1
gensim 4.2.1.dev0
FAST_VERSION 0

piskvorky:
@silviatti could you check this one please? #3197 was the only change in CoherenceModel.

@PrimozGodec could you interrupt the stuck computation with ctrl-c and post the traceback? Thanks.

PrimozGodec:
Thank you for your fast response. Here is the traceback.
```
Process AccumulatingWorker-1:
Process AccumulatingWorker-2:
Process AccumulatingWorker-3:
Traceback (most recent call last):
  File "/Users/primoz/miniconda3/envs/orange3/lib/python3.8/multiprocessing/process.py", line 318, in _bootstrap
    util._exit_function()
  File "/Users/primoz/miniconda3/envs/orange3/lib/python3.8/multiprocessing/util.py", line 360, in _exit_function
    _run_finalizers()
  File "/Users/primoz/miniconda3/envs/orange3/lib/python3.8/multiprocessing/util.py", line 300, in _run_finalizers
    finalizer()
  File "/Users/primoz/miniconda3/envs/orange3/lib/python3.8/multiprocessing/util.py", line 224, in __call__
    res = self._callback(*self._args, **self._kwargs)
  File "/Users/primoz/miniconda3/envs/orange3/lib/python3.8/multiprocessing/queues.py", line 195, in _finalize_join
    thread.join()
[...]

PR #3406, opened 2022-11-18
Path Coherence Model to correctly handle empty documents

PrimozGodec:
Fixes https://github.com/RaRe-Technologies/gensim/issues/3368
Before https://github.com/RaRe-Technologies/gensim/pull/3197 empty documents were skipped, after changes coherence fails if any document without tokens is in the list.

Fixing to also work when documents without tokens are in the list.


mpenkov:
Thank you @PrimozGodec !

PR #3415, opened 2022-12-15
Fix backwards compatibility bug in Word2Vec

### LINKED ISSUE #3413: Can not open and score a Word2vec model generated in v3.8.0 in v4.2.0
#### Problem description
Hello,

I'm trying to score a sentence in a script using gensim v4.2.0, with a model trained in v3.8.0. However, I meet the error `AttributeError: 'Word2Vec' object has no attribute 'syn1'`

Link to the model file: https://filetransfer.io/data-package/DIMOegMO#link

```python
from gensim.models import Word2vec
from gensim.utils import SaveLoad

sv = SaveLoad()
model = sv.load('test.mdl')
model.score(['test'], total_sentences=1)
```

#### Versions

```python
Linux-5.4.0-135-generic-x86_64-with-Ubuntu-18.04-bionic
Python 3.6.9 (default, Nov 25 2022, 14:10:45) 
[GCC 8.4.0]
>>> import struct; print("Bits", 8 * struct.calcsize("P"))
Bits 64
NumPy 1.18.3
SciPy 1.4.1
gensim 4.2.0
FAST_VERSION 1
```

Thanks a lot !

mpenkov:
Fix #3413


piskvorky (approved):

gensim/models/word2vec.py
```diff
@@ -1986,9 +1986,6 @@ def _load_specials(self, *args, **kwargs):
             for a in ('hashfxn', 'layer1_size', 'seed', 'syn1neg', 'syn1'):
                 if hasattr(self.trainables, a):
                     setattr(self, a, getattr(self.trainables, a))
-            if hasattr(self, 'syn1'):
-                self.syn1 = self.syn1
```
gojomo, line 1990: The actual intent here was to move `self.trainables.syn1` back to `self.syn1` to undo the `.trainables` complexification. So I don't think *just* deleting will fix user's raised issue. May need to be:

    if hasattr(self.trainables, 'syn1'):
        self.syn1 = self.trainables.syn1

There's then no need to `del` anything in this branch, as the whole `self.trainables` is dropped outside on the next line.

piskvorky: That's already done in the loop above.

mpenkov: OK, so perhaps there's more we can do in the test, because it should fail (not pass, like it does currently). What do we need to do?

piskvorky: I think the test should pass now. Why do you think it should fail?

mpenkov: Because of this:

> So I don't think just deleting will fix user's raised issue. 

But after re-reading the thread, it seems like the above comment is resolved, so we're good to go, right @gojomo ?

gensim/test/test_word2vec.py
```diff
@@ -275,6 +275,13 @@ def test_persistence(self):
         self.assertTrue(np.allclose(wv.vectors, loaded_wv.vectors))
         self.assertEqual(len(wv), len(loaded_wv))
 
+    def test_persistence_backwards_compatible(self):
+        """Can we still load a model created with an older gensim version?"""
+        path = datapath('model-from-gensim-3.8.0.w2v')
+        model = word2vec.Word2Vec.load(path)
+        x = model.score(['test'])
+        assert x is not None
```
gojomo, line 283: Should probably be a `self.assertSomething()` call, rather than a language `assert` for both consistency with other tests & rare case where someone may have disabled language-`assert`s via command-line. And, relying on the (maybe not possible to support indefinitely) `score()` function to probe model completeness, when that's not declared as tested in method name, could create maintenance headaches at some later date.

gensim/models/word2vec.py
gojomo: Aha, I'd missed the prior attribute move-over. If it fixes the user problem, looks good! But made a separate coment about the details of the new test.

PR #3437, opened 2023-01-20
[WIP] Clean up of FlsaModel: fixing bugs + formatting + efficiency

piskvorky:
Fixes #3423. Supersedes #3435, #3436.

This is still work-in-progress and needs finishing up. Namely:

1. Missing user-friendly docstrings and overall model motivation: what is this, who should use it? What do the various parameters mean?
2. As input, accept standard streaming corpora in the bag-of-words (BoW) format. Drop all the in-memory handling of the entire corpus in RAM as "list of list of strinks" and "scipy DOK matrix", that doesn't scale.
3. Complete the cleanup of the code formatting that I started. Especially use more helpful error messages in ValueErrors, showing what values are expected vs what the user supplied.
4. Related to that, focus all the parameter validation to a single place in code = the module entrypoints where users pass in these parameters. Currently the checks (even the same checks?) appear in multiple places, even in internal methods, where we should be in control of what the input values are, so we're doublechecking ourselves which makes no sense.

PR #3438, opened 2023-01-25
My flsa fixes

ERijck:
Fixes points 1,3 and 4 from #3437. 

More specifically, I have:

- Addressed the FIXMEs. I removed the ones I fixed and answered the ones `ERijck:` I did not fix.
- Fixed points 1,3, and 4. 
- Removed the method to get topic embeddings. It is not logical to have it in the `flsamodel.py`.

Now, we still need to fix point 2. It is unclear how to perform singular value decomposition on a bow representation. However, I suspect you do this in the `lsimodel.py` as well, so I will see how it is done there and rewrite the code.

### DIFF
--- a/gensim/models/flsamodel.py
+++ b/gensim/models/flsamodel.py
@@ -7,7 +7,7 @@
 """
 Fuzzy topic models.
 
-FlsaModel contains three topic modeling algorithms: FLSA, FLSA-W and FLSA-E. 
+FlsaModel contains three topic modeling algorithms: FLSA, FLSA-W and FLSA-E.
 
 1. FLSA - https://link.springer.com/article/10.1007/s40815-017-0327-9
 2. FLSA-W - https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=9660139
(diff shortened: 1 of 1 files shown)


gensim/models/flsamodel.py
```diff
@@ -623,7 +675,21 @@ def _create_partition_matrix(
             method='fcm',
         ):
         """
-        Cluster the projected data.
+        Use fuzzy clustering on the projected data.
+
+        The three clustering methods are:
+            1. fcm (Fuzzy C-Means)
+                (http://bitly.ws/zjs6)
```
piskvorky, line 682: Please use full canonical URLs. Avoid shorteners and any obfuscation (security risk).

gensim/models/flsamodel.py
```diff
@@ -1434,11 +1320,12 @@ def __init__(
             min_count=1,
             window=5,
             vector_size=20,
-            workers=4,
         ):
 
         self.model = ...   # FIXME what is this?
         self.word_embedding = ...  # FIXME what is this?
+        #ERijck: this way users can do post-analysis. But storing them as an attribute
+        #       is not crucial.
```
piskvorky, line 1328: OK, but what does the `...` do?

It's Python's special syntax for ellipsis, or slicing, but I'm not familiar with using it for attribute values.

gensim/models/flsamodel.py
```diff
@@ -1473,18 +1358,15 @@ def get_word_embedding(
...
         """
         self.model = Word2Vec(
             sentences=data,
             vector_size=vector_size,
             window=window,
             min_count=min_count,
-            workers=workers,
         )
         # FIXME is the whole Word2Vec model really needed? Why are we storing it as an attribute?
+        #ERijck: This way users can do a post analysis. However, it is not crucial and can be removed.
```
piskvorky, line 1369: OK. Best add a comment to that effect: 

`# Storing the whole trained model is not needed. We're only doing it so users can do a post analysis. Could be removed.`

gensim/models/flsamodel.py
```diff
@@ -359,7 +392,8 @@ def _create_sparse_global_term_weights(
                 word_to_index,
             )
         else:
-            raise ValueError('Invalid word weighting method')
+            raise ValueError(f'Unsupported word_weighting {word_weighting}')
+        #ERijck: This ValueError can be omitted, as it is an internal method. Do you agree?
```
piskvorky, line 396: Up to you. In general, input validation should only happen at public API points, where user sends in some data. 

And even there, the "pythonic" way is to rely on duck typing, and not require specific classes or methods.

gensim/models/flsamodel.py
```diff
@@ -955,19 +892,18 @@ def show_topics(
...
             )
         if not isinstance(index_to_word, dict):
-            raise TypeError("Please feed the algorithm 'index_to_word' as a dict")
+            raise TypeError("Please feed a dict to 'index_to_word'")
         if not isinstance(num_words, int) or num_words <= 0:
             raise TypeError("Please use a positive int for 'num_words'.")
         if prob_word_given_topic.shape[0] < prob_word_given_topic.shape[1]:
             raise ValueError("'prob_word_given_topic' has more columns then rows,",
                              " probably you need to take the transpose.")  # FIXME What? Why?
+            #ERijck: A model with more topics than words makes no sense.
```
piskvorky, line 906: Maybe, but that's not an error. At most a warning.

piskvorky (changes requested):
This is great, the documentation makes much more sense now.

I left a few minor comments; the only bigger step left is to accept streaming input = avoid materializing the whole corpus in-RAM. I'll try to get to it this weekend, to assist you.

gensim/models/flsamodel.py
mpenkov: It's a built-in ellipsis constant. People use it as a placeholder (TODO).

```
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0] on linux
>>> x = ...
>>> x
Ellipsis
```

piskvorky: Yes, that's what I wrote. But why is it here?

piskvorky: If they're really "required" public attributes with deferred initialization, use simply `self.model = None`. No need for special ellipsis.

Plus maybe add a comment about why these attributes are not initialized right away. Unless it's obvious (sorry I haven't checked the code in detail).

ERijck:
I have just fixed the small comments you left:

image

However, when I push this the same as how I did before, I get the following error:

image

When I try to pull the branch, git seems up to date already:

image

Which commands can I use to push my latest commit: 7864ebc8cbb5f54095b2b90d6b72255c915433c3?

piskvorky:
@ERijck we'll be releasing the hotfix soon, which only removes the `setup.py` dependency. Can you finish up this PR please?

Your error above means you rebased your PR, instead of doing simple commit, so the changes are "non-fast-forward". What did you do exactly? Maybe some graphic git tool messed up your repository. Prefer the command line.

ERijck:
Hi @piskvorky, thanks for your message. I will look into this with a colleague soon.

piskvorky:
@ERijck any plans to finish this?

I'm thinking it's best if you put this (fixed) module into a separate package, separate from gensim. It's useful on its own, but looks like it's not ready for gensim yet. The maintenance effort to support this module seems too steep for us.

ERijck:
@piskvorky, I fixed the points on March 21st. So that only step 2 (avoid materializing the whole corpus in RAM) is left. 
For this, I was waiting for your input to finish, but I understand that this takes too much time and effort for you.

What needs to change so that it is ready for Gensim? Is it correct that you want to use a BOW representation using integers to represent a document rather than a list of strings? If so, I can implement these changes. If you want other changes, could you briefly describe what you want/how I can do that? Then, I will work on these implementations.

PR #3447, opened 2023-03-03
remove unused dependency, handle ImportError

mpenkov:
1. Get rid of FuzzyTM import (it's completely unnecessary)
2. Delay import of pyfume until it's actually used, and then provide instructions for installing it

I also got rid of the flsamodel extra from setup.py, because it's not necessary to introduce another level of indirection here.

Bandaid hot-fix PR until #3437 is properly resolved.

### DIFF
--- a/gensim/models/flsamodel.py
+++ b/gensim/models/flsamodel.py
@@ -5,15 +5,17 @@ Created on Thu Oct 27 11:04:27 2022
 @author: 20200016
 """
 
-import math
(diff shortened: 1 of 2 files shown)


danieldaeschle:
This is needed heavily for us! FuzzyTM brings some GPL licensed code which we must avoid. Please merge this asap and create a new release. That would be helpful for us!

piskvorky (approved):

PR #3430, opened 2023-01-16
ci: enable arm64/aarch64 wheel builds

nikaro:
I gave a try at enabling builds for these architectures but did not succeed for linux/aarch64. What seems to happen (i may be wrong) is that the ARM docker images used by multibuild cannot be run on Linux amd64 platforms (the one used by GitHub runners), even with `--platform linux/arm64` flag. Cf. https://github.com/nikaro/gensim/actions/runs/3933985826

I cannot allocate more time try to fix this so i will left this PR as draft here, if someone want to continue the work on it or take inspiration from it (or close it) feel free to do it.

May be rewriting the wheels build CI from scratch using [cibuildwheel](https://github.com/pypa/cibuildwheel) could be easier, i did not test but to me it looks cleaner that the mess of bash scripts that multibuild is.

Refs https://github.com/RaRe-Technologies/gensim/issues/2994 https://github.com/RaRe-Technologies/gensim/issues/3425 https://github.com/RaRe-Technologies/gensim/issues/3341

PR #3448, opened 2023-03-04
ci: enable arm64/aarch64 wheel builds

nikaro:
Migrate wheel build system from [multibuild](https://github.com/multi-build/multibuild) to [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) to enable builds for ARM64 OSes.

Follow-up of https://github.com/RaRe-Technologies/gensim/pull/3430 ping @mpenkov

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -13,352 +13,54 @@ jobs:
   linters:
     uses: ./.github/workflows/linters.yml
 
-  multibuild:
-    timeout-minutes: 35
+  build_wheels:
+    needs: linters
+    name: Build wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     defaults:
       run:
         shell: bash
-
-    needs: [linters]
-
     strategy:
       fail-fast: false
       matrix:
-        include:
-        #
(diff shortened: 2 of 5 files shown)


mpenkov (approved):

mpenkov:
@nikaro Any idea why uploading takes 1.5 hours (in comparison to building, which takes 20 min)?

https://github.com/RaRe-Technologies/gensim/actions/runs/4370071051/jobs/7644594637

nikaro:
That's weird, i don't see why the Windows upload takes so much time for seven files when it takes only ~20 minutes for [eight files under Ubuntu](https://github.com/RaRe-Technologies/gensim/actions/runs/4370071051/jobs/7644594561#step:5:148). And this step did not change 🤔 

I wonder if a separate and unique upload job (that would grab the artifacts for all three platforms) wouldn't be more efficient.

mpenkov:
Might also be simpler to rewrite the upload code using boto3... what do you think?

nikaro:
Actually, GitHub runners come with [AWS CLI installed](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#cli-tools) so it should be even more straightforward to just use `aws s3 sync ./wheelhouse s3://gensim-wheels` (cf. https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html).

In the meantime, i pushed this [PR](https://github.com/RaRe-Technologies/gensim/pull/3454), and it [works](https://github.com/RaRe-Technologies/gensim/actions/runs/4384212183) and has the advantage also to provide the wheels as GitHub artifacts (see at the bottom of the workflow summary page). Replacing `wheelhouse-uploader` by `aws s3 sync` should be very simple, but i don't have an AWS playground to test it.

PR #3454, opened 2023-03-10
ci: move wheels upload into its own job

nikaro:
Since https://github.com/RaRe-Technologies/gensim/pull/3448 Windows wheels upload takes more than 1h. Let's try to do one upload (on Ubuntu) for all platforms.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -47,21 +47,42 @@ jobs:
           CIBW_TEST_REQUIRES: pytest testfixtures mock
           CIBW_TEST_SKIP: cp38* cp39* cp310* *_aarch64 *_arm64 *_universal2
 
+      - name: Upload wheels as artifacts
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: wheels-${{ matrix.os }}
+          path: wheelhouse/*.whl
+
+  upload:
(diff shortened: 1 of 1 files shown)


nikaro:
Copy/paste my comment from another PR of what could be an improvement:

> Actually, GitHub runners come with [AWS CLI installed](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#cli-tools) so it should be even more straightforward to just use `aws s3 sync ./wheelhouse s3://gensim-wheels` (cf. https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html).
>
> In the meantime, i pushed this [PR](https://github.com/RaRe-Technologies/gensim/pull/3454), and it [works](https://github.com/RaRe-Technologies/gensim/actions/runs/4384212183) and has the advantage also to provide the wheels as GitHub artifacts (see at the bottom of the workflow summary page). Replacing `wheelhouse-uploader` by `aws s3 sync` should be very simple, but i don't have an AWS playground to test it.

mpenkov:
OK, thanks for this. I'll merge and then see if anything further needs to be done.

nikaro:
There is definitely something strange with Windows wheels i guess: https://github.com/RaRe-Technologies/gensim/actions/runs/4384850019/jobs/7677703653

PR #3459, opened 2023-03-14
Replace HTTP with HTTPS in enwiki URLs

### LINKED ISSUE #3458: HTTP link provided instead of HTTPS
The text8 corpus link in the[ documentation](https://radimrehurek.com/gensim/models/word2vec.html) is http when it could be https. That's all.

Holmes5:
Replacing some http with https URLs; apart from being generally more secure, the browser complains about http download links from an https website. I have tested that the links and `wget` commands work.

resolves #3458


piskvorky:
Thank you @Holmes5 !

PR #3462, opened 2023-03-18
Doc fixes and separate workflow for building docs via CI

pabs3:
This set of changes fixes the current docs build failures in GitHub workflows
and makes it easier for contributors to submit docs changes, by doing the docs
build for them and exporting the changes to the prebuilt docs to a patch,
which they can then incorporate into their pull request.

- Use new `smart_open` compression parameter instead of `ignore_ext` when possible
- Download the NLTK WordNet data before using the WordNet lemmatizer
- Add missing documentation build dependency scikit-learn
- Timeout GitHub workflow documentation build after two hours instead of 10 min
- Upload the changes to the generated docs to GitHub artifacts


piskvorky (approved):
Thank you as always Paul!

.github/workflows/tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
   docs:
     name: build documentation
-    timeout-minutes: 10
+    timeout-minutes: 120
```
mpenkov, line 15: This is a significant increase. Why is this necessary?

.github/workflows/tests.yml
```diff
@@ -51,6 +51,29 @@ jobs:
...
+            git add .
+            branch="$GITHUB_HEAD_REF ($GITHUB_REF_NAME)"
+            git commit -m "Import rebuilt documentation for branch $branch"
+            git format-patch @^
+            git bundle create prebuilt-docs-changes.bundle @^...@
+            git reset --mixed @^
+            git diff --exit-code --stat @
+          fi
+
+      - name: Upload prebuilt docs changes
```
mpenkov, line 69: What do we do with these artifacts after the run is complete?

.github/workflows/tests.yml
pabs3: One of the runs for the update-links branch took over an hour, so I thought I would bump up the limit a bit.

https://github.com/RaRe-Technologies/gensim/actions/runs/4454218032/jobs/7823263904

pabs3: I see now that the build from today only took a minute, so I'll drop the timeout bump commit.

https://github.com/RaRe-Technologies/gensim/actions/runs/4459623150/jobs/7832189964

pabs3: Done.

mpenkov: I think the problem here is that the build _can_ take hours in theory. Some of the documentation examples take a really long time to build: that's why we cache their build output. So if somebody updates those examples, a rebuild will require a long time, and the build documentation action will time out.

We're kind of stuck between a rock and a hard place here.

1. We don't want to run these build steps for hours _all the time_ (e.g. for every commit)
2. We _do_ want to make it easier for people to contribute to documentation, which occasionally requires a long time to run

I think the best way to compromise between the above would be to move this build-and-stash-artifacts step to a separate action that people would trigger explicitly via a comment. We could set a long timeout (e.g. 2 hours, longer if necessary). We could restrict the range of people who can issue that trigger to e.g. project maintainers (if possible).

What do you think?

pabs3:
Sounds like a reasonable plan.

Do we know how long the build takes when starting from scratch
after you delete all of the cache from the git repository?

I don't know GitHub workflows well enough to implement the plan myself.

So perhaps you can merge the existing docs fixes commits,
then modify the workflows to implement the plan?

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

.github/workflows/tests.yml
pabs3: Sounds like a reasonable plan.

Do we know how long the build takes when starting from scratch
after you delete all of the cache from the git repository?

I don't know GitHub workflows well enough to implement the plan myself.

So perhaps you can merge the existing docs fixes commits,
then modify the workflows to implement the plan?

mpenkov: > So perhaps you can merge the existing docs fixes commits,
> then modify the workflows to implement the plan?

I'd rather modify the workflow first, then merge. That way, we won't break the existing workflow.

mpenkov:
@pabs3 Can you please enable commits from maintainers to this branch? I'd like to push some changes before merging.

pabs3:
Enabled maintainer commits.

pabs3:
Thanks for merging.

Looks like this was squashed when merging, I strongly suggest next time
using rebase merging or branch merging instead, because that preserves
the commits as they were, in this case they were deliberately separated
into logical commits, which helps later with git bisects and reviews.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

mpenkov:
My personal preference is to squash and merge. There's a wide range of contributors, and everyone does things at their own granularity: some people make lots of small commits, others a few large one. The squash brings everything to the same granularity: one PR, one commit. So, all PRs from the past several years have been squashed into a single commit before merging.

You make a good point about reviews and bisections, but we rarely do those after code has been merged in.

I know this preference is not universal (e.g. @piskvorky has mentioned he has a slight preference for not squashing) but there hasn't been an acute reason to change so far.

piskvorky:
Exactly right – I prefer branch merging too. But in case of gensim, w/e, no practical difference in reality.

pabs3:
I can see using squash+rebase merges is useful when contributors are
submitting messy PRs with lots of fixup commits. Personally I would ask
such folks to instead do logical commits and clean up their PRs with
rebase, have them force-push their fixed PR branches and then either do
rebase merges or fast-forward merges.

The other thing I don't like about squash merges is the commit message
ends up being very long with a ton of different concerns listed. When
there are lots of fixup commits, that means a lot of clutter too,
unless you manually clean that up locally before pushing.

I've been bisecting an infinite loop due to some kind of threading race
and or memory corruption in a test in another project recently and the
giant squash merges there really make it a lot harder to figure out
where the problem is coming from and which parts of the commit are
relevant. I'm thinking about bisecting the different hunks of the
commit to figure it out, but there are no tools to do that.

Anyways, thanks again for merging the patches, I'll move on now :)

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #3470, opened 2023-05-10
git rm gensim/models/flsamodel.py

mpenkov:


piskvorky:
@mpenkov so we're axing the module? Either way, let me merge the FLSA PRs first.

In case someone wants to continue in the future, so that they have the "fixed" version, with the various improvement from the recent PRs by me and @ERijck.

mpenkov:
Yes, I think it's best to remove this. It's completely new code, and we don't have the resources to maintain it (pace of existing PRs related to this module is extremely slow).

Let me know when it's ready to torch.

piskvorky:
> Are you open to me fixing this issue and adding it to Gensim once ready?

Yes, I am. The code is in much better shape already, after our recent PRs. If you can get its data representation in shape (same streaming input as the rest of Gensim, not everything in RAM) then I'll be happy to include FLSA.

For now, can you please fix https://github.com/RaRe-Technologies/gensim/pull/3437#issuecomment-1541707161? So we can merge that PR, it will be a starting point for any future revivals. Thanks!

PR #3475, opened 2023-05-23
Fix inconsistent documentation for LdaSeqModel #3474

### LINKED ISSUE #3474: Inconsistent documentation for LdaSeqModel
#### Problem description

The code description of how to use `LdaSeqModel` does not match what the code example actually does. On https://radimrehurek.com/gensim/models/ldaseqmodel.html we have:

image

according to the code example we have 9 documents, with 2, 4, and 3 documents respectively in each time slice,  while the text above claims otherwise. Also, the text is grammatically incorrect. 

#### Versions

`gensim==4.3.1`

rsokolewicz:
As mentioned in #3474, there is an inconsistency in a code example and the description text above it. Currently on the documentation it reads:

image

with this pull request, this will be changed to:

image

fixes #3474

### DIFF
--- a/gensim/models/ldaseqmodel.py
+++ b/gensim/models/ldaseqmodel.py
@@ -22,7 +22,7 @@ TODO: The next steps to take this forward would be:
 Examples
 --------
 
-Set up a model using have 30 documents, with 5 in the first time-slice, 10 in the second, and 15 in the third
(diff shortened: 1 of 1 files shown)


mpenkov:
Thank you @rsokolewicz !

PR #3524, opened 2024-04-03
Import deprecated scipy.linalg.triu from numpy.triu instead

### LINKED ISSUE #3525: Gensim broken with SciPy 1.13.0
Gensim is broken with [SciPy 1.13.0](https://github.com/scipy/scipy/releases/tag/v1.13.0) which was released today.

This is due to the [removal of scipy.linalg.triu](https://github.com/scipy/scipy/pull/18021)

```
.venv/lib/python3.11/site-packages/gensim/__init__.py:11: in <module>
    from gensim import parsing, corpora, matutils, interfaces, models, similarities, utils  # noqa:F401
.venv/lib/python3.11/site-packages/gensim/corpora/__init__.py:6: in <module>
    from .indexedcorpus import IndexedCorpus  # noqa:F401 must appear before the other classes
.venv/lib/python3.11/site-packages/gensim/corpora/indexedcorpus.py:14: in <module>
.venv/lib/python3.11/site-packages/gensim/interfaces.py:19: in <module>
.venv/lib/python3.11/site-packages/gensim/matutils.py:20: in <module>
    from scipy.linalg import get_blas_funcs, triu
E   ImportError: cannot import name 'triu' from 'scipy.linalg
```

Luffy610:
The scipy package latest released has removed the triu function due to its support already provided in numpy package 

Diff from the scipy library :
https://github.com/scipy/scipy/commit/5721702ee1471bdcbde0b33b850d91a9b4727d23#diff-6482beada610f60c724811cf77561f2bde8fd44ed2349939e505d28746a6d078

Removing the triu import from scipy library and getting it instead from numpy 

Fixes #3525.

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -20,6 +20,10 @@ from scipy.stats import entropy
 from scipy.linalg import get_blas_funcs
 from scipy.linalg.lapack import get_lapack_funcs
 from scipy.special import psi  # gamma function utils
+try:
+    from numpy import triu
+except ImportError:
+    from scipy.linalg import triu
 
 
 logger = logging.getLogger(__name__)
@@ -1127,7 +1131,7 @@ def qr_destroy(la):
     qr, tau, work, info = geqrf(a, lwork=work[0], overwrite_a=True)
     del a  # free up mem
     assert info >= 0
-    r = np.triu(qr[:n, :n])
+    r = triu(qr[:n, :n])
     if m < n:  # rare case, #features < #topics
         qr = qr[:, :m]  # retains fortran order
     gorgqr, = get_lapack_funcs(('orgqr',), (qr,))


gensim/matutils.py
```diff
@@ -1127,7 +1127,7 @@ def qr_destroy(la):
     qr, tau, work, info = geqrf(a, lwork=work[0], overwrite_a=True)
     del a  # free up mem
     assert info >= 0
-    r = triu(qr[:n, :n])
+    r = np.triu(qr[:n, :n])
```
piskvorky, line 1130: Does this work for all versions of scipy and numpy that Gensim supports?

Luffy610: yes I see the numpy has this method added way long back in 1.13 and now the stable version is 1.26, for the existing users who are using scipy as 1.12.0 and below won't have any problem, but for users using latest scipy would need this fix

mpenkov: It's probably best to conditionally import triu from np, falling back to scipy if np.triu is unavailable.

Luffy610: so try and except for this if try block np.triu fails then in except import the triu from scipy and the execute the triu function ?

mpenkov: At module scope:

```python
try:
    from numpy import triu
except ImportError:
    from scipy.linalg import triu
```

Luffy610: Ohh I see, Thanks let me update the PR :)

Luffy610:
Done updated the PR

piskvorky:
@mpenkov this fix seems important. Another maintenance release of Gensim may be timely anyway – drop unsupported Python versions, add new Python versions, etc.

WDYT? How about a release session, one of these days?

mpenkov:
Sure, we need to work out what's going on with the Windows builds, first. That's been on my todo list for a while.

mpenkov (approved):

mpenkov (approved):

unlimitedsola:
The failing Windows builds seem to be caused by https://github.com/pypa/cibuildwheel/issues/1740.
Looks like https://github.com/piskvorky/gensim/pull/3523 solves the issue as all workflows are finished successfully.

mpenkov:
Thank you for pointing that out @unlimitedsola

gensim/matutils.py
nickodell: @Luffy610 Did a quick look at when NumPy added this functionality. It seems to have been in roughly 2004 or 2005.

https://github.com/numpy/numpy/commit/2d0b625d58a48525ded47c4d184ad3837e71c316#diff-fc980d30a327648402874bd5daed1e37c96153a4a8fa542d4471e0f45dfadf9fR103

(The reason why all of the files have SciPy in the name is that this release pre-dates the split between SciPy and NumPy in 2006. That's also why SciPy has a function that does the exact same thing as NumPy.)

For context, the first NumPy release which was compatible with Python 3 was 1.5.0, in 2010. So if Gensim requires that its users have Python 3, then they definitely have `np.triu`.

philicious:
I'd like to express interest in a new release with this fix :)

PR #3529, opened 2024-04-23
use newer unittest.mock everwhere

a-detiste:
this is the same pattern as already seen in `test_poincare.py`


mpenkov:
I think even simpler `from unittest import mock` is safe.

That submodule has been around since Python 3.3. There is no way it won't be available.  The existing behavior is a hangover from the days when we supported Python 2.7.

a-detiste:
https://github.com/piskvorky/gensim/blob/645c2d191b7076331c9b0f160f027cb9b3400b15/gensim/test/test_parsing.py#L34 -> I studied this one :-)

a-detiste:
Changed as requested

> I think even simpler `from unittest import mock` is safe.
> 
> That submodule has been around since Python 3.3. There is no way it won't be available. The existing behavior is a hangover from the days when we supported Python 2.7.

mpenkov:
Thank you @a-detiste !

PR #3538, opened 2024-06-14
removed scipy from build-only dependencies

filip-komarzyniec:
Resolves: https://github.com/piskvorky/gensim/issues/3536

In addition to removing scipy from _build-system.requires_ table I've also updated the README file. 

Please refer to the comments for some additional doubts I have regarding more changes that might need to be introduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,12 +50,11 @@ Installation
 ------------
 
 This software depends on [NumPy], a Python package for
-scientific computing. You must have it installed prior to installing
-gensim.
-
-It is also recommended you install a fast BLAS library before installing
-NumPy. This is optional, but using an optimized BLAS such as MKL, [ATLAS] or
-[OpenBLAS] is known to improve performance by as much as an order of
+scientific computing. Please bear in mind that building NumPy from source
+(e.g. by installing gensim on a platform which lacks NumPy .whl distribution)
+is a non-trivial task involving [linking NumPy to a BLAS library].  
+It is recommended to provide a fast one (such as MKL, [ATLAS] or
+[OpenBLAS]) which can improve performance by as much as an order of
 magnitude. On OSX, NumPy picks up its vecLib BLAS automatically,
 so you don’t need to do anything special.
 
@@ -175,6 +174,7 @@ BibTeX entry:
   [Vector Space Model]: https://en.wikipedia.org/wiki/Vector_space_model
   [unsupervised document analysis]: https://en.wikipedia.org/wiki/Latent_semantic_indexing
   [NumPy]: https://numpy.org/install/
+  [linking NumPy to a BLAS library]: https://numpy.org/devdocs/building/blas_lapack.html
   [ATLAS]: https://math-atlas.sourceforge.net/
   [OpenBLAS]: https://xianyi.github.io/OpenBLAS/
   [source tar.gz]: https://pypi.org/project/gensim/
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ on Wikipedia.
 Installation
 ------------
 
-This software depends on [NumPy and Scipy], two Python packages for
-scientific computing. You must have them installed prior to installing
+This software depends on [NumPy], a Python package for
+scientific computing. You must have it installed prior to installing
 gensim.
 
 It is also recommended you install a fast BLAS library before installing
@@ -69,7 +69,9 @@ Or, if you have instead downloaded and unzipped the [source tar.gz]
 package:
 
 ```bash
-    python setup.py install
+    tar -xvzf gensim-X.X.X.tar.gz
+    cd gensim-X.X.X/
+    pip install .
 ```
 
 For alternative modes of installation, see the [documentation].
@@ -172,7 +174,7 @@ BibTeX entry:
   [documentation and Jupyter Notebook tutorials]: https://github.com/RaRe-Technologies/gensim/#documentation
   [Vector Space Model]: https://en.wikipedia.org/wiki/Vector_space_model
   [unsupervised document analysis]: https://en.wikipedia.org/wiki/Latent_semantic_indexing
-  [NumPy and Scipy]: https://scipy.org/install/
+  [NumPy]: https://numpy.org/install/
   [ATLAS]: https://math-atlas.sourceforge.net/
   [OpenBLAS]: https://xianyi.github.io/OpenBLAS/
   [source tar.gz]: https://pypi.org/project/gensim/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,6 @@ requires = [
     # is 1.18.5, remove the line when they increase oldest supported Numpy for this platform
     "numpy==1.18.5; python_version=='3.8' and platform_machine not in 'arm64|aarch64'",
     "oldest-supported-numpy; python_version>'3.8' or platform_machine in 'arm64|aarch64'",
-    "scipy",
     "setuptools",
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -49,8 +49,8 @@ on Wikipedia.
 Installation
 ------------
 
-This software depends on [NumPy and Scipy], two Python packages for
-scientific computing. You must have them installed prior to installing
+This software depends on [NumPy], a Python package for
+scientific computing. You must have it installed prior to installing
```
filip-komarzyniec, line 53: Does it really have to be installed prior to installing `gensim` having in mind that a proper _numpy_ version is already specified in _build-system.requires_ table? Meaning it will be downloaded automatically during _gensim_ build phase.

If it does not --- what should we do with regard to the next paragraph about BLAS lib? I think it will not be customisable for any architecture that a _numpy_ .whl build exists for.

mpenkov: Yes, I think the sentence starting "You must" is incorrect. Numpy will get installed automatically if it already isn't there.

With BLAS, I think we just leave that note as-is. You can't customize it as part of the gensim install. If you want to mess with BLAS, then you have to do it yourself, _before_ installing numpy and gensim.

Right @piskvorky ?

piskvorky: I forgot how exactly it worked back then. IIRC numpy was needed for the installation (numpy actively used during install, down to its C API). Maybe scipy too. I believe that is the intent behind *"You must have [numpy] installed prior to installing gensim"*.

If `pip` installs numpy automatically before the gensim install itself starts, that's great. Some people don't use `pip`, I don't know all the deployment methods people use (from source, conda, etc) and whether they all install dependencies first.

In any case that was 15 years ago, the current ecosystem might work differently, and none of this is needed.

filip-komarzyniec: As I see it, the problem of linking Numpy to any BLAS lib is a matter to worry about only during building Numpy from source. 
This can occur only if: 
  * one wants to customise his Numpy build in a non typical way. In this case I believe such person knows what he's doing so this note is not really valuable for him, 

  * one installs gensim which in turn downloads appropriate NumPy build (but src distribution instead of a .whl). Then the pip (or I believe any package manager) would try building the NumPy from source and implicitly linking an existing BLAS lib. 
  
What's more, a NumPy installed as a gensim's dependency will not be (in some cases) the same NumPy as used for building gensim. Pip by default builds a library in a separate temporary virtual env where it installs build-only dependencies.

I've redacted the paragraph. Please review whether it sounds sensible .

pyproject.toml
```diff
@@ -5,11 +5,11 @@ requires = [
     # order of magnitude slower, so avoid it for now.
     #
     "Cython>=0.29.32,<3.0.0",
-    # oldest supported Numpy for this platform is 1.17 but the oldest supported by Gensim
-    # is 1.18.5, remove the line when they increase oldest supported Numpy for this platform
-    "numpy==1.18.5; python_version=='3.8' and platform_machine not in 'arm64|aarch64'",
+    # oldest supported Numpy for non-'arm64|aarch64' architectures is 1.17.*
+    # as well as for 'Windows' on arm64. The oldest supported Numpy by Gensim
+    # is 1.18.5. Remove below line when they increase the oldest supported Numpy for those platforms
+    "numpy==1.18.5; python_version=='3.8' and platform_machine != 'aarch64' and (platform_machine != 'arm64' or platform_system == 'Windows')",
```
mpenkov, line 11: What is the motivation behind changing this?

filip-komarzyniec: Looking at the meta package [_oldest-supported-numpy_](https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg) I see that `numpy<1.18.5` will be downloaded for python 3.8 and Windows on arm64 (apart from all non-arm64 and non-aarch64 platforms). 

So I believe this check was incomplete resulting in installing unsupported version of Numpy for python3.8 arm64 Windows users.

mpenkov: @piskvorky Do we want to deal with this now? It hasn't been reported yet, so it may not actually be a problem. It's been a while since I've messed with oldest-supported-numpy, so I'm not 100% sure this fix is needed and safe.

filip-komarzyniec: I understand the reasoning. 
Should I revert this change then?

mpenkov: Yes, let's revert it. Feel free to make a separate PR with that change so we can review it later, after the current release is out.

filip-komarzyniec: I'll do as advised, thanks!

mpenkov (approved):
Good work, thank you!

filip-komarzyniec:
I see the builds failing on tests step (due to different scipy versions being installed for a given python version). 

My PR should not affect anything more than library building step (which passes successfully each time). 
Is there anything that can be done in order to fix the tests (for example should I merge my changes with any specific other branch fixing scipy runtime dependency?)

mpenkov:
Very strange. I see your previous commits built fine. I will investigate.

juhoinkinen:
@mpenkov and @filip-komarzyniec, I tried [restricting scipy versions to <1.14 in setup.py](https://github.com/juhoinkinen/gensim/commit/76b54199dc301883f57e2fce192b14364ee4ba3c), and this makes the CICD tests pass, see [this run](https://github.com/juhoinkinen/gensim/actions/runs/9953342285). 

Not sure why scipy 1.13.1 got installed for Python 3.8 and 3.9 in the first place ([this logs line](https://github.com/piskvorky/gensim/actions/runs/9695920583/job/26785100142#step:9:33)), and scipy 1.14.0 for Python 3.10+ ([this logs line](https://github.com/piskvorky/gensim/actions/runs/9695920583/job/26785100282#step:9:30)). I checked that:
- pipdeptree showed that there is a second-degree-dependency of scipy via visdom, but it has no [version constraints](https://github.com/fossasia/visdom/blob/79c0360264eef8b784ea43a34d482efe124411db/setup.py#L42),
- in PyPI there is just the [source distribution of visdom](https://pypi.org/project/visdom/#files), so no Python version effects from there either.


Anyway, I wonder could this scipy version constraint <1.14 be applied or is there something more to it?

(I took a look because our [CICD pipeline](https://github.com/NatLibFi/Annif/actions/workflows/cicd.yml) is not working due to Gensim installation issues.)

mpenkov:
I think I see what's happening here:

- We don't need scipy to build wheels
- We **do** need scipy to test the wheels; more specifically 1.13.1 or below, because future versions got rid of sparsetools

juhoinkinen:
I got the tests to pass by pinning to [numpy<2.0 for Python 3.12 build-system](https://github.com/juhoinkinen/gensim/commit/6ba56820d6b9431cc2dd259f9a187352f8b323b6), see this run: https://github.com/juhoinkinen/gensim/actions/runs/9953367669

 (Ref https://github.com/piskvorky/gensim/issues/3541#issuecomment-2230600527).

mpenkov:
The problem with scipy is fixed, now we have to solve the problem with macos.

https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/

Crosswind:
Let's see if `macos-12` will fix the issue. It may not.
I have found some information on the problem which seems to originate in `pip`. The build uses a pretty outdated version of it.
I have found this https://github.com/pypa/cibuildwheel/issues/1410 and then this PR https://github.com/bloomberg/blazingmq-sdk-python/pull/23 which could be worth a try.

PR #3547, opened 2024-07-18
Pin scipy in installation requirements as well to fix failing tests

### LINKED ISSUE #3544: gensim package import issues
Why pakage import issue 

import gensim
Screenshot from 2024-07-11 00-53-59

Crosswind:
@mpenkov I did some digging to try and understand why the tests started failing suddenly. The important difference is that the successful builds were running the tests against scipy==1.13.1. This is because the version isn't pinned inside setup.py.

This completes the version pinning of scipy started in afbb82ea3446b5462a6ae1f5b02d338ff0170e5a

The tests are currently failing because the build system runs the test against scipy==1.14.0 which doesn't have the required sparsetools (csc_matvecs, specifically) anymore. This commit makes sure that gensim also doesn't test against a version higher than 1.13.1.

Fixes #3544

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -331,7 +331,7 @@ NUMPY_STR = 'numpy >= 1.18.5, < 2.0'
 
 install_requires = [
     NUMPY_STR,
-    'scipy >= 1.7.0',
+    'scipy >= 1.7.0, <= 1.13.1',
     'smart_open >= 1.8.1',
 ]
 


Crosswind:
For those interested on how I got to the conclusion:
I am no expert in building wheels at all but I realized in the workflow logs that for some reason it was still installing scipy 1.14.0 prior to running the tests:
```
Collecting scipy>=1.7.0 (from gensim==4.3.2.dev0)
    Downloading scipy-1.14.0-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (60 kB)
       ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 60.8/60.8 kB 15.3 MB/s eta 0:00:00
```
The `scipy>=1.7.0` got me thinking as I didn't see that in the `pyproject.toml`. Instead, this is specified in `setup.py` which is used during the build process.

One potential improvement would be to remove the `>=1.7.0` as this is also not specified in `pyproject.toml`. This could help avoid confusion in the future and I also believe it is unlikely that pip will find itself in a situation where a version such as 1.8.0 is the highest available version.

Crosswind:
The same fix is already in #3538 so this can be closed.

PR #3548, opened 2024-07-18
Fix import triu

Hoasker:
I'm using scipy 1.13.1 and it works in manual fix. I found scipy in the documentation starting with version 1.12.0, but it appeared in the source code much earlier. The triu function has been moved from scipy.linalg to scipy.sparse. I changed imports:
```
try:
    from numpy import triu
except ImportError:
    from scipy.linalg import triu
```
to `from scipy.sparse import triu` and it worked.

### DIFF
--- a/gensim/matutils.py
+++ b/gensim/matutils.py
@@ -20,10 +20,7 @@ from scipy.stats import entropy
 from scipy.linalg import get_blas_funcs
 from scipy.linalg.lapack import get_lapack_funcs
 from scipy.special import psi  # gamma function utils
-try:
-    from numpy import triu
-except ImportError:
-    from scipy.linalg import triu
+from scipy.sparse import triu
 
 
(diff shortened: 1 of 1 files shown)


Crosswind:
`scipy` suggests to import `triu` from `numpy`. The change to numpy was introduced only a short while back in ad68ee3f105fc37cf8db333bfb837fe889ff74ac and has been working. What exact problem are you trying to fix with this PR?

Hoasker:
> `scipy` suggests to import `triu` from `numpy`. The change to numpy was introduced only a short while back in [ad68ee3](https://github.com/piskvorky/gensim/commit/ad68ee3f105fc37cf8db333bfb837fe889ff74ac) and has been working. What exact problem are you trying to fix with this PR?

triu from numpy for some reason did not fix the problem when run program. And this method fixed it

PR #3552, opened 2024-07-18
Sync PyPi package description with Github README

### LINKED ISSUE #3452: Broken documentation links on Gensim's PyPI page
Links in the `Documentation` section of our official PyPI page are broken (return 404): 
https://pypi.org/project/gensim/#documentation-1

OTOH, links at https://github.com/RaRe-Technologies/gensim#documentation work, so it looks like the PyPI version contains obsolete links that got out of sync.

jicruz96:
closes #3452 

The reason your [PyPi Project Page](https://pypi.org/project/gensim/) is out of sync is because of an out-dated hard-coded description in your setup.py file. To fix it:

* `setup.py` now uses the project README text for `long_description`
* All README image paths have been updated to their respective HTTP urls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,21 +123,21 @@ Adopters
 
 | Company | Logo | Industry | Use of Gensim |
 |---------|------|----------|---------------|
-| [RARE Technologies](https://rare-technologies.com/) | ![rare](docs/src/readme_images/rare.png) | ML & NLP consulting | Creators of Gensim – this is us! |
-| [Amazon](http://www.amazon.com/) |  ![amazon](docs/src/readme_images/amazon.png) | Retail |  Document similarity. |
-| [National Institutes of Health](https://github.com/NIHOPA/pipeline_word2vec) | ![nih](docs/src/readme_images/nih.png) | Health | Processing grants and publications with word2vec. |
-| [Cisco Security](http://www.cisco.com/c/en/us/products/security/index.html) | ![cisco](docs/src/readme_images/cisco.png) | Security |  Large-scale fraud detection. |
-| [Mindseye](http://www.mindseyesolutions.com/) | ![mindseye](docs/src/readme_images/mindseye.png) | Legal | Similarities in legal documents. |
-| [Channel 4](http://www.channel4.com/) | ![channel4](docs/src/readme_images/channel4.png) | Media | Recommendation engine. |
-| [Talentpair](http://talentpair.com) | ![talent-pair](docs/src/readme_images/talent-pair.png) | HR | Candidate matching in high-touch recruiting. |
-| [Juju](http://www.juju.com/)  | ![juju](docs/src/readme_images/juju.png) | HR | Provide non-obvious related job suggestions. |
-| [Tailwind](https://www.tailwindapp.com/) | ![tailwind](docs/src/readme_images/tailwind.png) | Media | Post interesting and relevant content to Pinterest. |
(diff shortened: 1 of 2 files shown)


jicruz96:
unrelated to this PR, I started a conversation in the Google Groups Gensim forum (as recommended by your contribution guidelines when opening new issues) asking if working on adding type hints to gensim would be welcome work, but I haven't gotten any response. Just pinging that again here.

piskvorky:
> setup.py now uses the project README text for long_description

IIRC setup.py (PyPI) used to required the RST format, whereas our README uses the Markdown format. That's why we kept two different descriptions, with just a stub for PyPI.

If that's no longer the case and setup.py supports markdown, that's great. Thanks for the PR.

setup.py
```diff
@@ -342,8 +237,8 @@ def run(self):
     name='gensim',
     version='4.3.2.dev0',
     description='Python framework for fast Vector Space Modelling',
-    long_description=LONG_DESCRIPTION,
-
+    long_description=Path("README.md").read_text(),
```
piskvorky, line 240: What's the default encoding? Better be explicit.

piskvorky (changes requested):

README.md
```diff
@@ -123,21 +123,21 @@ Adopters
...
-| [Search Metrics](http://www.searchmetrics.com/) | ![search-metrics](docs/src/readme_images/search-metrics.png) | Content Marketing | Gensim word2vec used for entity disambiguation in Search Engine Optimisation. |
-| [12K Research](https://12k.com/) | ![12k](docs/src/readme_images/12k.png)| Media |   Document similarity analysis on media articles. |
-| [Stillwater Supercomputing](http://www.stillwater-sc.com/) | ![stillwater](docs/src/readme_images/stillwater.png) | Hardware | Document comprehension and association with word2vec. |
-| [SiteGround](https://www.siteground.com/) |  ![siteground](docs/src/readme_images/siteground.png) | Web hosting | An ensemble search engine which uses different embeddings models and similarities, including word2vec, WMD, and LDA. |
-| [Capital One](https://www.capitalone.com/) | ![capitalone](docs/src/readme_images/capitalone.png) | Finance | Topic modeling for customer complaints exploration. |
+| [RARE Technologies](https://rare-technologies.com/) | ![rare](https://raw.githubusercontent.com/piskvorky/gensim/develop/docs/src/readme_images/rare.png) | ML & NLP consulting | Creators of Gensim – this is us! |
```
piskvorky, line 126: Why the absolute links? The README will now no longer render except with internet.

All the referenced files are local, and IMO should be read as local. This is not a "github README", it's a project README, which currently happens to be hosted on github.

jicruz96: when you publish this package to PyPi, it won't have the images from the docs folder bundled along with it, so when the PyPi page renders the long description text, it won't know what `docs/src/readme_images`, so no images will appear. here's a [stackoverflow question](https://stackoverflow.com/questions/41983209/how-do-i-add-images-to-a-pypi-readme-that-works-on-github) on the subject.

perhaps as a middle ground solution we can run a `.replace(DOC_IMAGES_FOLDER, RAW_GIT_URL)` on the README text in setup.py

jicruz96: ok, I've pushed up an alternate approach. did my best to replicate the existing code/comment style but obviously feel free to request any change or edit the code to your liking.

piskvorky: Ah that makes sense, thank you.

Your solution is great I think, but let's hear from @mpenkov :)

mpenkov (approved):
LGTM

PR #3562, opened 2024-11-07
Numpy 2.0.0 update and removing deprecated scipy modules

hechth:
This PR is an initial attempt at porting the project to numpy 2.0.0 - so far running the tests manually after installing all dependencies seemed fine. I removed the deprecated scipy sparse functions and replaced them with the normal operator versions.

PR #3563, opened 2024-11-08
Numpy2 draft

YoungMind1:
I Wanted to share my the findings here.

- I think Cython upgrade is necessary according to [this](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#the-pyarray-descr-struct-has-been-changed).
- Swapping `except *` and `nogil` in Cython files are due to compiler warnings, shouldn't affect anything.
- The addition of `noexcept` in many functions is due to the acquiring of gil to check exceptions. but in past versions of Cython functions without explicit declaration of `except` would "swallow" exceptions. so this change should not affect anything. related #3488 , [Cython link](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#exception-values-and-noexcept)
- 9 tests are failing, the one I really dug into was `gensim/test/test_word2vec.py::TestWord2VecModel::test_similarities` which should be due to changes in numpy datatype promotions [link](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion)

I don't have the expertise or the time to fix these failing tests, just wanted to give a head start to anyone willing.
cc: @hechth 
related #3562

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -61,19 +61,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - {python: '3.8', os: macos-12}
           - {python: '3.9', os: macos-12}
           - {python: '3.10', os: macos-12}
           - {python: '3.11', os: macos-12}
           - {python: '3.12', os: macos-12}
 
-          - {python: '3.8', os: ubuntu-20.04}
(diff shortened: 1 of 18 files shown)


hechth:
@YoungMind1 thanks a lot, I'll merge your PR into mine locally and continue with the tests since you did complementary things. Thanks a lot!

Issue #3611, opened 2025-04-22
Python 3.13 incompatibilities


cansavvy:
Thanks for putting together this cool package! 

## Summary 

I attempted to install gensim 4.3.3 on my mac computer (and on Ubuntu docker images) that were using Python 3.13 and due to some changes in the Python 3.13

## Details 
It appears to be because gensim's C code is referring to things that have been deprecated in Python 3.13 See here https://docs.python.org/3/deprecations/index.html 

e.g. 
- note: 'ma_version_tag' has been explicitly marked deprecated here
- error: no member named 'ob_digit' in 'struct _longobject'

A truncated version of the log output when I try to install with Python 3.13 
```
        251 | #define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
            |                                                      ^
      gensim/models/word2vec_inner.c:12176:9: warning: 'ma_version_tag' is deprecated [-Wdeprecated-declarations]
       12176 |         __PYX_PY_DICT_LOOKUP_IF_MODIFIED(
             |         ^
      gensim/models/word2vec_inner.c:1697:30: note: expanded from macro '__PYX_PY_DICT_LOOKUP_IF_MODIFIED'
       1697 |         __pyx_dict_version = __PYX_GET_DICT_VERSION(DICT);\
            |                              ^
      gensim/models/word2vec_inner.c:1686:65: note: expanded from macro '__PYX_GET_DICT_VERSION'
       1686 | #define __PYX_GET_DICT_VERSION(dict)  (((PyDictObject*)(dict))->ma_version_tag)
            |                                                                 ^
      /opt/homebrew/opt/python@3.13/Frameworks/Python.framework/Versions/3.13/include/python3.13/cpython/dictobject.h:25:5:
      note: 'ma_version_tag' has been explicitly marked deprecated here
         25 |     Py_DEPRECATED(3.12) uint64_t ma_version_tag;
            |     ^
      /opt/homebrew/opt/python@3.13/Frameworks/Python.framework/Versions/3.13/include/python3.13/pyport.h:251:54:
      note: expanded from macro 'Py_DEPRECATED'
        251 | #define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
            |                                                      ^
      gensim/models/word2vec_inner.c:12812:55: error: no member named 'ob_digit' in 'struct _longobject'
       12812 |             const digit* digits = ((PyLongObject*)x)->ob_digit;
[...]

Issue #3607, opened 2025-04-18
Github Actions Runner Image Deprecations and macOS arch


julianpollmann:
As adressed [here](https://github.com/actions/runner-images/issues/11101), Ubuntu 20.04 runner images were removed for Github actions on April 15, 2025. The CI will therefore not run on these images and fail.

Runner Images should be upgraded to Ubuntu 22.04/24.04.

julianpollmann:
- [ ] Runner Image for Windows is deprecated and should be upgraded to Windows Server 2025. See [this](https://github.com/actions/runner-images/issues/12045)
- [ ] [macOS latest](https://github.com/piskvorky/gensim/blob/develop/.github/workflows/build-wheels.yml#L42C11-L42C41) has dropped support for x86 and only supports arm64. See [this](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners)

PR #3615, opened 2025-06-10
Migration numPy 2.0 and remove deprecated scipy sparsetools functions

julianpollmann:
Hey @gojomo @hechth @piskvorky @mpenkov,
I've created a new PR for the Migration to numPy 2.0 and the removal of deprecated scipy functions. What has been done so far:

- [x] Updated dtypes for functions where tests were failing
- [x] Fixed some numpy assertions alltrue -> all
- [x] **Python 3.8** has been removed from the supported versions
- [x] Fixed deprecated scipy import for csc_matrix
- [x] Fixed deifferent behaviour for argsort, where multiple results on the same level were sorted differently

What needs to be done:
- [ ] Remove NMSLibIndexer since it will only work on numPy 1.0? (See [this comment](https://github.com/piskvorky/gensim/issues/3560#issuecomment-2937318329))
- [ ] For Python 3.13 support some parts of [word2vec.pyx](https://github.com/piskvorky/gensim/blob/develop/gensim/models/word2vec_inner.pyx) need to be adapted (See #3611). However, Python 3.9-3.12 work perfectly. Should 3.13 support incorporated in this PR or in a different one?
- [ ] CI pipeline is failing due to outdated Runners and arch for latest macOS Runner needs to be adapted. (See #3607). I'll try to fix these in a different PR.

Hope we get this done now 💪

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,8 +7,10 @@ requires = [
     "Cython>=0.29.32,<3.0.0",
     # oldest supported Numpy for this platform is 1.17 but the oldest supported by Gensim
     # is 1.18.5, remove the line when they increase oldest supported Numpy for this platform
-    "numpy==1.18.5; python_version=='3.8' and platform_machine not in 'arm64|aarch64'",
-    "oldest-supported-numpy; python_version>'3.8' or platform_machine in 'arm64|aarch64'",
+    # 20240604 GM: testing numpy-2.0.0rc2 which requires python >= 3.9 (to 3.12)
+    "numpy==2.0.0rc2; python_version>='3.9' and platform_machine not in 'arm64|aarch64'",
+    # "oldest-supported-numpy; python_version>'3.8' or platform_machine in 'arm64|aarch64'",
+    "scipy",
     "setuptools",
     "wheel",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ requires = [
     # oldest supported Numpy for this platform is 1.17 but the oldest supported by Gensim
     # is 1.18.5, remove the line when they increase oldest supported Numpy for this platform
     # 20240604 GM: testing numpy-2.0.0rc2 which requires python >= 3.9 (to 3.12)
-    "numpy==2.0.0rc2; python_version>='3.9' and platform_machine not in 'arm64|aarch64'",
+    "numpy==2.0.0; python_version>='3.9' and platform_machine not in 'arm64|aarch64'",
     # "oldest-supported-numpy; python_version>'3.8' or platform_machine in 'arm64|aarch64'",
     "scipy",
     "setuptools",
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -7,8 +7,10 @@ requires = [
     "Cython>=0.29.32,<3.0.0",
```
piskvorky, line 7: Is 3.1.2 any better?

If so let's change this line to 3.1.2, and also update the comment above.

julianpollmann: @piskvorky I removed the upper version bound and tested with 3.1.2. However builds/tests take ages. It seems that Cython >=3 has stricter type checking, making it slower.

piskvorky: I doubt it's type checking that takes ages. Maybe something (a dependency) gets built from source that used to be pulled pre-built before?

julianpollmann: @piskvorky I tried several things, unfortunately no luck. Especially Testing of test_doc2vec.py::TestDoc2VecModel seems to be much slower on Cython 3.1.2.
With Upgrading to 3.1.2 Python 3.13 seems to work however?!

morotti:
>  With Upgrading to 3.1.2 Python 3.13 seems to work however?!

if it's of any help. support for python 3.13 was only added in cython 3.1.x
so you have to update cython to support newer python versions, but old python versions could continue to use an old cython if you want.
https://cython.readthedocs.io/en/latest/src/changes.html

If you need, I think there is an old syntax for setup.py/setup.cfg to specify different versions.

```
 requires = [
    'Cython>=0.29.32,<3.0.0; python_version<"3.13"',
    'Cython>=3.1.2; python_version>="3.13"',
```

piskvorky:
That's an unusually lucid and helpful warning message! Can you please try adding that `noexcept` and see if it makes any difference?

julianpollmann:
> That's an unusually lucid and helpful warning message / hint! Thanks for spotting it.
> 
> Can you please try adding that `noexcept` and see if it makes any difference?

@piskvorky It seems like Cython 3 will check for exceptions and therefore use the GIL. After adding `noexcept` there is a speedup. However I cannot judge, if `noexcept` is appropriate for all `nogil` methods. Would be good, if somebody with more Cython experience can have a look.

julianpollmann:
Hey @piskvorky, could you rerun failed jobs from the CI workflows? There are some http errors, which should disappear.
I updated to Cython 3, since `noexcept` did speed up the builds and this [specifying different Versions](https://github.com/piskvorky/gensim/pull/3615#issuecomment-3137563149) would cause the CI build to fail.
I also disabled Python 3.8 and 3.14 builds.
With that everything should work.

mpenkov:
@piskvorky @gojomo If there are no objections, let's merge and release

piskvorky:
I already OKed above; please merge & let's release!

Thanks to everyone, I know this has been in the oven for a very long time.

PR #3627, opened 2025-10-10
Add builds for Python 3.14

julianpollmann:
Adds experimental support for Python 3.14. To be fixed:
- [ ] #3628


timovdk:
Hi! I was wondering what the progress is on releasing Gensim for Python 3.14, and whether there is anything I can help out with?

PR #3648, opened 2026-03-20
fix: gensim is not longer usable due to pinning scipy <1.14.0

### LINKED ISSUE #3623: gensim is not longer usable due to pinning scipy <1.14.0
hello,

reporting that gensim is no longer installable due to pinning an old scipy<1.14.0

* gensim is no longer installable on recent python versions that only have a recent version of scipy.
* gensim is no longer installable with other recent packages that require scipy 1.14, which is quite a few.

I note that scipy 1.11 1.12 1.13 are crashing on skylake / cascade lake Intel CPUs, they call invalid instructions and crash depending on what matrix operations you do, so the upgrade to scipy 1.14 is quite critical.

the pin is in the setup.py with a comment https://github.com/piskvorky/gensim/blob/master/setup.py#L337

```
install_requires = [
    ...
    # scipy 1.14.0 and onwards removes deprecated sparsetools submodule
    'scipy >= 1.7.0, <1.14.0',
```

JiwaniZakir:
Fixes #3623

The `stochastic_svd` function in `gensim/models/lsimodel.py` relied on the internal `csc_matvecs` function from scipy's sparse module, which was removed from the public API in scipy 1.14+. Replaces the `csc_matvecs` calls with standard matrix multiplication (`y += corpus * o` and `y += chunk * o`), which is functionally equivalent and compatible with all supported scipy versions.


JiwaniZakir:
Closing this PR as it hasn't received maintainer review. Happy to resubmit if there's interest.

PR #3649, opened 2026-03-29
Fixed py3.14 configuration issues

codeandsolder:
chunkize crashed on Linux because default multiprocessing mode changed for forkserver


codeandsolder:
Is anything blocking this PR?

